PR #919, opened 2014-02-25
remove_geom: a function to remove layers from a ggplot2 object

eshilts:
Remove layers from a ggplot2 object.


hadley:
Unfortunately I've now completely gone off this idea - I'd rather people modified their code than modified their plots in this way.

PR #921, opened 2014-02-27
First stub of a length scale

jiho:
It can be used whenever fixed segment lengths are drawn. Examples are given 
with geom_rug but it could also be used with a segment geom that would be 
defined in terms of angle and length rather than in terms of beginning and end 
coords.

The remaining issue is the legend, which currently just draws segments. It 
should draw segments of lengths identical to the main plot but since the two 
viewports (main plot and legend) do not have the same reference, it is hard.

Edit on 2014-02-27: rebased on head + latest version of roxygen


hadley:
Still interested in this?

jiho:
Yes I would love to be able to plot arrow fields in ggplot using this but I have no idea how to fix the legend part.

hadley:
Ah, ok - I think it's fundamentally impossible to draw a legend here because the length is measured in npc, not a physical unit. Legends map elements in the data space to things in physical-graphics space, so a legend here doesn't really make sense. (This is basically the same problem as having text automatically expand the axes).

You could make this work by making the legend scale use physical coordinates (e.g. cm). The obvious disadvantage of this is that the appearance of the plot is not longer independent of its physical size (or even aspect ratio!). But I think this is a sacrifice you have to make if you want a legend.

I'll leave it up to you to either close the PR or modify it to use the cm unit (or ask for more clarification if my explanation doesn't make sense)

hadley:
I'm going to close for now. Feel free to re-open/submit.

PR #924, opened 2014-02-27
Allow to reverse the order of colours in a brewer palette

jiho:
Use the same argument that scale_*_hue uses: direction

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,7 +3,7 @@ ggplot2 0.9.3.1.99
 
 * Add a `direction` argument to brewer scales to be able to reverse the order of
   colours. In many diverging scales, the warm colours get associated with lowest
-  values in an ordered factor.
+  values in an ordered factor. (@jiho, #924)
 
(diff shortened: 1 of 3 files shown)


hadley:
Can you please merge/rebase?  If it still needs the dev version of scales, can you please update the DESCRIPTION?

jiho:
Ideally, I would need you to pull #1076 first. Then I'll rebase (or more probably just completely rewrite) this on top of it. The changes in `scales` are minimal (https://github.com/hadley/scales/pull/36) but should be released in order for this to not depend on a dev version of scales, which is probably unsuitable.

Do you plan on releasing scales also?

hadley:
I'm planning on releasing scales tomorrow.

hadley:
Can you please finish this off now that scales is out? You'll need to add an explicit version requirement in the DESCRIPTION

jiho:
#1139 replaces this one (much simpler thanks to the change in scales and branches from a recent master)

PR #939, opened 2014-04-07
file extension parameter for ggsave() to allow device specification for files without extension

sebkopf:
Thank you for the fantastic ggplots package!
  
This is a suggested implementation for specifying a file extension in ```ggsave``` to allow device matching (pdf, png, jpg, etc.) for filenames without extension. The suggestion is motivated by the problem of using ggsave in a shinyApp for letting the user download a plot (for example addressed on StackOverflow here: http://stackoverflow.com/questions/14810409/save-plots-made-in-a-shiny-app). Because the shiny server provides a temporary file without extension, ggsave can't match to any of the internally defined device functions. I'm not aware of a currently available solution to this problem and although there are several work-arounds, I like the simplicity of ggsave and wanted to suggest this solution (not sure if it's the best way to do it, but it's the most elegant I could think of). This would allow use of ```ggsave``` for a shinyApp ```downloadHandler``` like so:
  
  ```coffee
# In ui.R:
downloadLink('downloadPlot', 'Download')

# In server.R:
output$downloadPlot <- downloadHandler(
  filename = "myplot.pdf",
  content = function(file) {
    ggsave(file, plot = qplot(speed, dist, data = cars), ext = "pdf")
  }
)
```
I'm relatively new to GitHub and this is my second contribution to someone else's project, please forgive if I'm missing some specific conventions or etiquette about forking/pull requests.

Sebastian

### DIFF
--- a/R/save.r
+++ b/R/save.r
@@ -7,7 +7,8 @@
 #'
 #' @param filename File name to create on disk.
 #' @param plot Plot to save, defaults to last plot displayed.
-#' @param device Device to use. By default, extracted from extension.
+#' @param device Device to use (function or any of the recognized extensions,
+#'   e.g. \code{"pdf"}). By default, extracted from filename extension.
 #'   \code{ggsave} currently recognises eps/ps, tex (pictex), pdf, jpeg, tiff,
 #'   png, bmp, svg and wmf (windows only).
 #' @param path Path to save plot to (combined with filename).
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: R/save.r [hunk 1]
```diff
@@ -72,9 +77,22 @@ ggsave <- function(filename = default_name(plot), plot = last_plot(),
...
     ext <- tolower(pieces[length(pieces)])
-    match.fun(ext)
+    match_device(ext)
   }
 
+  match_device <- function(ext) {
+    if(!exists(ext, mode = "function")) {
+      stop("No graphics device defined for the file extension '", ext, "'. ", 
+           "Make sure to specify a filename with supported extension or ",
+           "set the device parameter.")
```
hadley: Can you please add `call. = FALSE` here?

hadley:
Just update this PR by merging/rebasing against master, and re-pushing the branch to github.

hadley:
Are you still interested in this PR?

sebkopf:
Okay, should be all ready to go now (including the requested .call = FALSE addition and some commit cleanup). Thanks for the reminder msg, missed last week's while traveling. And thanks a lot for the rebasing hint to retrigger Travis CI build, that's exactly what I was missing - learned something new again.

hadley:
Thanks!

PR #951, opened 2014-04-24
Fix typo as reported in Documentation Feedback 1162

BrianDiggs:
Resubmission of #847


hadley:
Fixed in like 3 other PRs :/

PR #953, opened 2014-04-25
theme(validate=FALSE) means do not validate_element

tdhock:
This allows support for non-standard theme options, which would close

https://github.com/hadley/ggplot2/issues/938


tdhock:
@wch @hadley can you please take a look at merging this? It is causing us problems in animint.

https://github.com/tdhock/animint/issues/29

tdhock:
actually this was due to a bug in tdhock/ggplot2 which I just fixed (and added tests for)

wch:
I'm a bit reluctant to add an option like this... Could you implement the animint theme options by attaching them as an attribute to the ggplot object? Getting that to play well with the `+` operator may be tricky though.

tdhock:
Dear @wch, thanks for considering my PR, and thanks for your concerns.

Can you please clarify why exactly you are reluctant to add such an option? It seems to me that it is completely backwards compatible with existing ggplots. I even added some unit tests in test-theme.R to make sure that validate=FALSE is respected.

I haven't had time to try implementing it as an attribute of the ggplot object, but thanks for the suggestion.

antongrau:
I would like this merge to go through. I am implementing animint in three R-packages and I need the custom theme functions.

hadley:
Can you please post a clean PR for review? (i.e. unmodified readme & [house style](http://r-pkgs.had.co.nz/r.html#style))

tdhock:
Please review #1121 which includes an unmodified readme and spaces before and after = and + operators.

PR #960, opened 2014-05-05
Corrected field name: colour => color

imanuelcostigan:


hadley:
Thanks - also spotted and fixed myself.

PR #964, opened 2014-05-17
Fixed typo in `geom_density2d` roxygen2 Description

joey711:
"estimatation" --> "estimation"

Only fixed the roxygen header. Would need to rebuild the doc itself.


hadley:
Thanks!

PR #965, opened 2014-05-21
Throw error if lower > upper in cont scales. Fixes #906

### LINKED ISSUE #906: Crash if lower > upper in limits argument of scale_colour_continuous
``` r
ggplot(data.frame(x=1:2, y=2:1, col=1:2), aes(x=x, y=y, col=col)) + 
    geom_point() + 
    scale_colour_continuous(limits=c(1, -1))
```

Does not affect scale_x_continuous, scale_y_continuous.
Affects scale_fill_continuous (confirmed with geom_tile).

tonytonov:


hadley:
I agree that this is probably nicer, but the additional logic around reverse scales is really icky.

tonytonov:
Should I try coming up with some other workaround or we leave it as it is? FWIW, there's an issue discussion here (https://github.com/hadley/ggplot2/issues/906). Though the MWE there no longer causes a crash, just an "empty raster" error.

hadley:
BTW the MWE now works

tonytonov:
Ah, great, I'm deleting the branch then. Thanks!

PR #968, opened 2014-05-24
Fix for bad URL in qplot docs

croach:
The URL for the sample chapter on the qplot function in both the R docs and the man docs were pointing to an old, no longer available URL. I've updated the documentation to point to the latest URL.


hadley:
Thanks!

PR #969, opened 2014-06-01
Coord expand

lionel-:
The difference between setting the limits through a coordinate system or via a scale is central to ggplot but the two procedures are not consistent with each other. Compared to scales, the coord functions lack an expand argument. More importantly, they do not apply the default expansion and the same plot will look different in function of which method was used to set limits.

This pull request implements `xexpand` and `yexpand` arguments for coord functions, with the same default expansion as for the scales.

Since setting the limits through coord is symmetric to setting the limits with scales, it should happen in the same place in the code. Furthermore, this allows to know if the scale is continuous or discrete thanks to the scale object being passed along. To make this possible, I changed the coord object to reflect the scale object structure. Coord objects now contains two lists named `x` and `y`, each containing `expand` and limits. In order to make the difference between passing the whole coord object and passing the x or y component, I added a dot to `coord`. When a function receives a `.coord`, we know it is the x or y component of a coord object.

One difficulty that I encountered while checking that everything was ok, is that coord_polar does not have the x and y components. And it needs special treatment for the default expansion, which depends on theta being equal to "x" or "y".  My solution was to add x and y objects, containing theta, with class set to "polar" so that `coord_expand_defaults` dispatches correctly the .coord object to `coord_expand_defaults.polar`.


hadley:
I'm extremely nervous about introducing a change like this. Even if the behaviour is (arguably) better, it's going to affect a huge amount of existing code.

lionel-:
@hadley 
yes I guess it's better to have the default behaviour to be just as before and optionally expand the coords, so that the change won't affect existing code?

there is actually a bad bug in this patch (IIRC, the regular expand system for scales does not work correctly). I'll probably have to find another implementation approach, but I'm not sure I'll have enough time to work on this until the end of June :/

Are you planning to release soon?

lionel-:
actually it was an easy fix, so it should be good now. The visual tests pass.

hadley:
This is somewhat embarrassing to ask, but how did you run the visual tests? I've forgotten how :/

lionel-:
the instructions are hidden in Winston's wiki ;)
https://github.com/wch/ggplot2/wiki

apparently he wrote some tools to automatically diff the images, but I didn't run them because they rely on git commits and the test results are gitignored. I think they should be included in the repo, and maybe automatically checked by travis if the diffing is reliable enough.

hadley:
I'm a bit torn on this PR. I think it's a reasonable idea, but mimicking existing behaviour makes the code rather complicated. It also makes it harder to explain where to set `expand` - do you do it in the coord or the scale? The benefit also seems rather mild compared to directly expanding the ranges yourself in the call to the coord.

lionel-:
The inconsistent appearance of regular and coord'ed plots has always bugged me, especially since there is no easy way to get the same expansion.

Maybe there is a way to get expansion information from the scales rather as arguments to coord functions? I don't remember why I didn't do it this way. It would make it easier to understand but then the default appearance of coord'ed plots would necessarily be different. Would you prefer such an implementation if it happens to be feasible?

lionel-:
> I don't remember why I didn't do it this way.

Thinking about it, it probably was not a technical issue but just that it seemed more consistent to provide expansion information alongside limits. I think it is straightforward to think about it like this: with a scale, you set `limits`/`expand`, with a coord, you set `xlim`/`xexpand` or `ylim`/`yexpand`.

hadley:
I have some vague memory that for a while there was a `wise` argument that did what you're asking. But it's now gone, and I don't remember why.

lionel-:
Here is the commit introducing `wise` : https://github.com/hadley/ggplot2/commit/67e30b39d0802bb1986f8b80a5247f033d9e4516
And the commit removing it : https://github.com/hadley/ggplot2/commit/37999acf333438543b22c049ab338829fea9f3f0

According to the commit message, the reason for removing it was that the coord'ed plots were always expanded. But this expansion must have been later removed.

hadley:
So it sounds like a better fix would be a unit test to make sure coords are always expanded + always expanding them.

hadley:
Do you want to have a go? Otherwise, because it's bug, I'll do it.

lionel-:
hmm that would be a good default but then it would make it difficult to set precise limits while using coords, for the cases where precision or absolute values are needed...

hadley:
Let's talk this through in an issue.

PR #970, opened 2014-06-02
check.class argument to ggsave

baptiste:
optional by-pass of ggsave checks that an object of class ggplot is being recorded. 
In addition, a grid.draw method is defined, rather than calling print()
directly, to allow other grobs such as modified gtables to be saved to a device with the same convenience.

### DIFF
--- a/R/save.r
+++ b/R/save.r
@@ -22,6 +22,7 @@
 #' @param limitsize when \code{TRUE} (the default), \code{ggsave} will not
 #'   save images larger than 50x50 inches, to prevent the common error of
 #'   specifying dimensions in pixels.
+#' @param check.class logical, checking whether the plot must inherit the ggplot class
 #' @param ... other arguments passed to graphics device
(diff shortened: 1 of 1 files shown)


hadley:
Hmmm, I don't really like adding an additional argument. Any thoughts on alternative ways to check if it's saveable? Maybe we could have?

```{r}
is_gridable <- function(x) UseMethod("is_gridable")
is_gridable.ggplot <- function(x) TRUE
is_gridable.grob <- function(x) TRUE
is_gridable.default <- function(x) FALSE
```

?

hadley:
Done in 7387d39a

baptiste:
Thanks! Any particular reason to have grid.draw.ggplot() calling print.ggplot(), and not the opposite?

hadley:
You're the one who wrote that ;) it probably should be the other way around but it required a bit too much thinking

PR #974, opened 2014-06-15
Facet titles

lionel-:
This implements a `switch_strips` argument to `facet_wrap` and `facet_grid`, which can be set to `x`, `y` or `both` (the latter only for grids). This allows to display the facet strips near the axes. They then act as axes subtitles. The padding between the switched strips and their corresponding axes can be adjusted through a new theme setting.

Displaying the facet titles near the axes especially makes sense when the facet title directly characterize a change in the axis (for example, a transformation, a different measure, etc). This should be typically used in conjunction with a theme without boxes around the strips. I go over this in the updated documentation.

A grid facetting with strips switched to both axes:
facet1

A wrap facetting with free scales and strips switched to the x axis:
facet2

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -38,6 +38,13 @@ ggplot2 1.0.1.9000
 
 * `Geom`, `Stat` and `Position` are now exported, making it easier to create
   new geoms, stats and positions in other packages (#989).
+* `facet_grid()` and `facet_wrap()` gain a `switch` argument that
+  allows the facet titles to be displayed near the axes. They then act
+  as axes subtitles. Can be set to "x", "y" or "both" (the latter only
+  for grids) to control which label strips are switched. (@lionel-)
+
+* `Geom` is now exported, making it easier to create new geoms in other
+  packages (#989).
 
 * New `theme_void()`, which is completely empty. Useful for plots with non
   standard coordinates or for producing numerical drawings with R. (@jiho, #976)
--- a/R/facet-grid-.r
+++ b/R/facet-grid-.r
@@ -27,6 +27,11 @@
 #' @param as.table If \code{TRUE}, the default, the facets are laid out like
 #'   a table with highest values at the bottom-right. If \code{FALSE}, the
 #'   facets are laid out like a plot with the highest value at the top-right.
+#' @param switch By default, the labels are displayed on the top and
+#'   right of the plot. If \code{"x"}, the top labels will be
+#'   displayed to the bottom. If \code{"y"}, the right-hand side
+#'   labels will be displayed to the left. Can also be set to
+#'   \code{"both"}.
 #' @param shrink If \code{TRUE}, will shrink scales to fit output of
 #'   statistics, not raw data. If \code{FALSE}, will be range of raw data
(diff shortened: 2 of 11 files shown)


hadley:
I like the idea. Can you please re-document and merge/rebase so I can do a review on cleaner code? (A bit of commit squashing wouldn't hurt either)

lionel-:
Here you go.

R/theme.r
```diff
@@ -174,6 +174,10 @@ print.theme <- function(x, ...) str(x)
 #'                    (\code{element_text}; inherits from \code{strip.text}) \cr
 #'   strip.text.y     \tab facet labels along vertical direction
 #'                    (\code{element_text}; inherits from \code{strip.text}) \cr
+#'   strip.switch.pad.grid \tab space between strips and axes when strips are switched
+#'                    (\code{unit} \cr
```
hadley: Missing closing parens?

R/facet-grid-.r
```diff
@@ -262,13 +349,23 @@ facet_strips.grid <- function(facet, panel, theme) {
   col_vars <- unique(panel$layout[names(facet$cols)])
   row_vars <- unique(panel$layout[names(facet$rows)])
 
+  dir <- list(r = "r", t = "t")
+  if (!is.null(facet$switch) && facet$switch %in% c("both", "x")) {
+    dir$t = "b"
```
hadley: `<-`

R/facet-grid-.r
```diff
@@ -239,17 +258,85 @@ facet_render.grid <- function(facet, panel, coord, theme, geom_grobs) {
   strips$r$heights <- panels$heights
   strips$t$widths <- panels$widths
 
+  # Check if switch is consistent with grid layout
+  switch_x <- !is.null(facet$switch) && facet$switch %in% c("both", "x")
+  switch_y <- !is.null(facet$switch) && facet$switch %in% c("both", "y")
+  if (switch_x && length(strips$t) == 0) {
+    facet$switch <- if (facet$switch == "both") "y" else NULL
+    switch_x <- FALSE
+    message("Cannot switch x axis strips as they do not exist")
```
hadley: To me, it feels like this should either be silent or a warning

hadley:
Overall, looks good.  A few minor nit-picks inline.

lionel-:
voilà

NEWS
```diff
@@ -38,6 +38,13 @@ ggplot2 1.0.1.9000
 
 * `Geom`, `Stat` and `Position` are now exported, making it easier to create
   new geoms, stats and positions in other packages (#989).
+* `facet_grid()` and `facet_wrap()` gain a `switch` argument that
+  allows the facet titles to be displayed near the axes. They then act
+  as axes subtitles. Can be set to "x", "y" or "both" (the latter only
+  for grids) to control which label strips are switched. (@lionel-)
+
+* `Geom` is now exported, making it easier to create new geoms in other
```
hadley: This got messed up, but I can fix it.

hadley:
Thanks! People are really going to like this feature

PR #975, opened 2014-06-16
Add new function geom_curve

veraanadi:
geom_curve adds curved lines to a plot, similar to geom_segment adding straight lines.


veraanadi:
updated it and created new pull request

PR #983, opened 2014-07-09
Edit warning message to reflect intuitive use

mkcor:
Stacking works fine and as expected whenever `ymin` values are greater than or equal to 0.
And stacking does not really make sense if `ymin < 0`, indeed.

### DIFF
--- a/R/position-stack.r
+++ b/R/position-stack.r
@@ -48,7 +48,7 @@ PositionStack <- proto(Position, {
     }
 
     if (!is.null(data$ymin) && !all(data$ymin == 0))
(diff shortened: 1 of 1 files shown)


hadley:
Can you provide some evidence to back up that statement?

hadley:
This does not look correct to me:

```R
df <- data.frame(x = 1, y = c(1, 3))

ggplot(df, aes(x, ymin = y, y = y + 0.5, ymax = y + 1, group = y))  +
  geom_errorbar(position = "stack")
```

PR #984, opened 2014-07-09
allow ggplot_build to return an additional value: the processed data as ...

xsaintmleux:
...it is prior to calling calculate_stats, to get info about each individual data point (e.g. which panel is it on?)


hadley:
I'd rather always do this than have an additional option.  But why do you need it?

PR #993, opened 2014-08-06
New theme option to put panel/grid on top of data

### LINKED ISSUE #551: Allow to draw gridlines on top of data
When using `geom_tile` or maps, the grid lines are completely obscured by the data plotted on top of them. Sometimes, particularly for maps, it is desirable to still see the grid lines. Using transparency is a workaround but costly in terms of rendering, compatibility (EPS does not support transparency) and legibility.

Ideally, one should be able to specify where to draw grid lines (at which level in the plot). For example, one might want to draw them on top of a map/tile layer but below a points layer.

Another way to look at it is to specify, for each geom/layer wether is it above or below the grid lines.

noamross:
A new theme option, `panel.ontop`, places the background objects (panel, gridlines), on *top* of the
data in the plot if set to `TRUE`.  This allows one to place gridlines over the data.  One would typically also include a blank or transparent panel in the theme.

This is implemented by changes in facet_panels.* functions, changing the ordering of layers depending on the value of `theme$panel.ontop`.  `add_theme` was also changed to allow for logical theme elements.

This is a simple implementation.  A more advanced one would allow gridlines to be placed on top of data while keeping the panel below data.

Updated `theme()` documentation included via roxygen 4.0.1.

Tests passed with the following exception. which I believe is unrelated:

```
Stats : .Error in predLoess(y, x, newx, s, weights, pars$robust, pars$span, pars$degree,  : 
  NA/NaN/Inf in foreign function call (arg 5)
In addition: Warning message:
In max(panels$ROW) : no non-missing arguments to max; returning -Inf
```

Fixes #551


R/theme.r
```diff
@@ -423,7 +426,7 @@ add_theme <- function(t1, t2, t2name) {
       # If x is NULL or element_blank, then just assign it y
       x <- y
     } else if (is.null(y) || is.character(y) || is.numeric(y) ||
-               inherits(y, "element_blank")) {
+               is.logical(y) || inherits(y, "element_blank")) {
```
hadley: What's this change related to?

hadley:
Looks good. Could you please

* [ ] merge/rebase (paying particular attention to NEWS)
* [ ] re-roxygenise
* [ ] add a compelling example somewhere?

hadley:
Thanks!

PR #1002, opened 2014-08-31
created fortify.glm

lselzer:
I created `fortify.glm` because at present `glm` models are handled by `fortify.lm`. This doesn't allow to change the `.fitted` values from `link` (the default), to `response` or `terms`.


hadley:
I've really gone off fortify in favour of the broom package.

PR #1008, opened 2014-09-11
Add a fortify method for x,y,z type lists

jiho:
Those lists are standard but do not have a specific class. There is no way to
make have this method be called on these lists only but great care is taken to
check that the list is indeed well formed before proceeding with the
conversion.

Many functions producing 2D results (interpolation, re-griding of spatial
data, heatmaps, etc.) produce such x, y, z lists because they are directly
plot-able with image() or persp(). This method makes it easy to use ggplot
straight on the output of such functions.


hadley:
I've really gone off fortify as a method - I think broom handles most cases better.

jiho:
Did not know about broom. Looks great! What becomes of `autoplot` then? Do you expect it to depend on broom and call the appropriate `augment` method?

jiho:
Submitted as https://github.com/dgrtwo/broom/pull/56

PR #1010, opened 2014-09-17
typo in help for scale_continuous

jimhester:
missing an s for the scales package.

I did not rebuild the documentation, this fix won't go into effect until then.


hadley:
Fixed elsewhen - thanks!

PR #1011, opened 2014-09-17
stat_function to compute function on data not aesthetic scale

BrianDiggs:
`stat_function` currently computes the function with the transformed
(not original data) values. This causes the values to be wrong if
a different scale, other than continuous, is used.

See http://stackoverflow.com/q/9382032/892313 for an example
of this.


inst/tests/test-stats.r
```diff
@@ -134,3 +134,26 @@ test_that("stat-density2d", {
   expect_true(max(ret$y) > 35)
 
 })
+
+context("stat-function")
```
hadley: Would you mind putting this in a new file?

hadley:
Could you please also 

* [ ] add a bullet to NEWS
* [ ] merge/rebase

BrianDiggs:
I'll work on it this weekend.

BrianDiggs:
Comments addressed.

hadley:
Thanks!

PR #1016, opened 2014-10-03
Jitter for binary data

lionel-:
This is a new position proto for binary data. It bounds the jittering inside the data range, which produces nicer logistic regression plots.

rplot2


NEWS
```diff
@@ -1,3 +1,12 @@
+ggplot2 1.0.0.99
+----------------------------------------------------------------
+
+NEW FEATURES
+
+* `position_innerjitter` allows jittering the data while keeping it
```
hadley: innerjitter or jitterbin? Or `injitter` ?

R/position-jitterbin.r
```diff
@@ -0,0 +1,79 @@
...
+    bound_x <- "x" %in% .$bounded_scales
+    bound_y <- "y" %in% .$bounded_scales
+    if ((bound_x && !length(unique(data$x)) == 2) ||
+        (bound_y && !length(unique(data$y)) == 2))
+      stop("position_jitterbin can only bound binary scales")
+
+    trans_x <- NULL
+    trans_y <- NULL
+    if(.$width > 0) {
+      trans_x <- function(x) {
```
hadley: I think this could probably be much simplified by using `runif()` in both branches, and just adjusting the min and max values.

Alternatively, another parameterisation would be to alway jitter towards the centre (but that might pose just as many problems in other cases)

hadley:
Could you please also merge/rebase?

NEWS
lionel-: `injitter` is nice!

R/position-jitterbin.r
lionel-: doing it as you suggest is still a bit lengthy:
```r
trans_y <- function(x) {
  u <- stats::runif(length(x), -.$height, .$height)
  if (bound_y) {
    range <- range(x, na.rm = TRUE)
    ifelse(x == range[1],
      range[1] + u/2 + max(u/2),
      range[2] - u/2 + min(u/2)
    )
  } else {
    x + u
  }
}
```
so I simplified another way

lionel-:
@hadley done.

I also added an alias `geom_injitter()` in the last commit. If you don't like it, I'll delete the commit, otherwise I'll document the alias and squash it.

hadley:
Now jitterin sounds better to me. Maybe just add a Boolean flag to geom jitter? I don't think it needs its own geom.

lionel-:
> Now jitterin sounds better to me.

Or does it sound like the American version of jittering? ;)

> Maybe just add a Boolean flag to geom jitter? I don't think it needs its own geom.

or allow `position` to take `jitterin` when `width` or `height` are specified, as in the last commit? Then I just need to delete the new alias and it's good to go.

lionel-:
by the way, I have an old almost finished branch where I implement all missing horizontal version of positions and geoms. I was stuck on a very weird #rstats bug in one of the new geoms but now that I've read `adv-r` I might be able to figure it out if you're interested in a PR ;)

hadley:
I'd prefer the boolean argument

lionel-:
here you go

hadley:
Now I'm thinking it would be better to add hjust/vjust arguments to `position_jitter` like the new support in `geom_text()` - they can be set to left/top, middle/center,bottom/right, inward and outward.

Sorry I keep moving the goal posts - let me know if you're too frustrated and I'll do it myself.

lionel-:
but we'd only have middle/center and inward right?

hadley:
I'd implement them all just for completeness. left/center/right would apply to horizontal jittering.  outward is rarely useful, but it seems asymmetric to have inward but not outward

hadley:
Actually, let me just do it once the conversion to R6 is complete. I think the correct way forward is now clear to me, and I can modify `position_jitter()` quickly.

hadley:
Ok, now I don't think it's possible to do what I want, but I think your proposal is a bit too specific for ggplot2.  But you'll soon be able to put geoms in your own packages (code is done, documentation is pending), so would you mind doing that?

lionel-:
for me it's ok.

I have some other stats / geom for plotting fitted models but I don't know if/when I'll get around to publish a package.

PR #1018, opened 2014-10-05
Dark theme

jiho:


hadley:
Thanks!

In the future, can you make sure the news bullet is the first one? (Not under any headings) That makes sure it ends up in the right place even if it takes me a while to merge

PR #1024, opened 2014-10-29
Fortify for Raster* objects

bleutner:
Hi,

I'd like to suggest to expand the functionality of fortify to Raster* objects from the raster package. Since rasters are often very large (>memory) they should be subsampled for efficient plotting. This is implemented in the raster package for calls to raster::plot but it would be nice to have a fortify method for use with ggplot2. 

There are three classes RasterLayer, RasterBrick and RasterStack which can all be subsampled in the same way using a regular grid.The maximal number of pixels to plot can be specified and defaults to 500000.

Fortify returns a data.frame with pixel coordinates and corresponding raster values intended for use with geom_raster.

What do you think?
Benjamin


hadley:
I'm not planning on adding any more fortify methods, since I think broom has done a better job of providing diverse methods.

I'd be interested to see what an equivalent of `geom_map()` would look like though.

bleutner:
ok, thanks for letting me know. I wasn't aware of the broom package. 

Are you planning to deprecate the fortify methods in the future? 
I'm planning to ship the suggested fortify method with my RStoolbox package for now, but might as well name it differently if fortify doesn't have a future.

hadley:
No plans to deprecate

PR #1034, opened 2014-12-09
layer.r: added missing error message newline/spacer

setempler:
Problem: 'stop' does not separate elements in the dots argument, therefore the error message (R/layer.r line 160f) is clumped:
Reproducible example:

```R
f <- function (a) ggplot(data.frame(x=rnorm(10),y=1:10), aes_q(x=quote(x),y=quote(y),color=a)) + geom_point()
f(1) # runs fine
f(1:3) # returns the error message:
Error: Aesthetics must either be length one, or the same length as the dataProblems:1:3

The fix would result in
...
f(1:3)
Error: Aesthetics must either be length one, or the same length as the data
Problems: 1:3
```

### DIFF
--- a/R/layer.r
+++ b/R/layer.r
@@ -159,7 +159,7 @@ Layer <- proto(expr = {
     wrong <- lengths != 1 & lengths != n
     if (any(wrong)) {
       stop("Aesthetics must either be length one, or the same length as the data",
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1055, opened 2015-02-19
typo in annotate docs

Blahah:


hadley:
Thanks - but you have to make the fix in the originating R file, since that's what the Rd files are generated from.

Blahah:
d'oh!

PR #1017, opened 2014-10-05
Default to guide=colourbar for distiller scales

jiho:

PR #1076, opened 2015-03-31
Improvements to scale_*_distiller

jiho:
The order of colors in particular, and the fact that it cannot be easily
reversed (pending a change in package scales) makes scale_*_distiller
almost useless at the moment.

These are simple fixes and should be easy to pull.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -56,6 +56,9 @@ ggplot2 1.0.1
 
 * Fixes to pass `R CMD check --run-donttest` in R-devel.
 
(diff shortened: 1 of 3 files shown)


jiho:
I realised that I already submitted the colourbar aspect as another pull-request #1017. Let me know how you want to handle those.

hadley:
Could you please merge/rebase so that it builds on travis? Otherwise looks good.

hadley:
Thanks!

PR #1077, opened 2015-03-31
Create correct legend for continuous color even if there is only one color

### LINKED ISSUE #943: Empty raster
When I try to do

```R
ggplot(data.frame(x=c(1,2),y=c(1,2),z=c(1.5,1.5)),aes(x=x,y=y,color=z)) +
  geom_point()
```

I get the error "Error in grid.Call.graphics(L_raster, x$raster, x$x, x$y, x$width, x$height, : Empty raster"

I described this issue in a little more detail here http://stackoverflow.com/questions/23003527/weird-ggplot2-error-empty-raster

krlmlr:
fixes #943

I have no idea where to plug in a test for this. Perhaps in the examples?


hadley:
It should probably go in the visual tests, but they're not in very good shape right now.

Would you mind merging/rebasing (mostly to trigger travis)

krlmlr:
Done. What are the main issues, and your further plans concerning visual testing?

GitHub shows diffs for images. Travis could push back any changes detected by the visual differ. This doesn't work for pull requests, so contributors will have to sign up for Travis to use this.

hadley:
The main issues are I don't remember how they're supposed to work (maybe @wch does), and they probably need to be re-thought now that we can rely on travis to automatically run the tests. It may also be able to outsource some of the diffing given that there are a growing number of "visual diffs as service" companies (although they're mostly aimed at web pages)

PR #1082, opened 2015-04-02
Documentation tweaks

joranE:
Some minor changes & clarifications to guides and scales. Main change is the explicit documentation of the inherit.aes, subset and show_guide arguments for geoms, rather than then redirecting people to layer, which has no documentation.

### DIFF
--- a/R/annotation-logticks.r
+++ b/R/annotation-logticks.r
@@ -7,11 +7,12 @@
 #' @param sides a string that controls which sides of the plot the log ticks appear on.
 #'   It can be set to a string containing any of \code{"trbl"}, for top, right,
 #'   bottom, and left.
-#' @param short a \code{unit} object specifying the length of the short tick marks
-#' @param mid a \code{unit} object specifying the length of the middle tick marks.
-#'   In base 10, these are the "5" ticks.
-#' @param long a \code{unit} object specifying the length of the long tick marks.
-#'   In base 10, these are the "1" (or "10") ticks.
+#' @param short a \code{\link[pkg:grid]{unit}} object specifying the length of the 
+#'   short tick marks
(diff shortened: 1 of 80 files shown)


hadley:
I like the basic idea, but I'd like to downplay `subset`, so I'd prefer it not be included by default.  And `inherit.aes` is rather esoteric, and shouldn't be needed by the end user, so I'd rather keep that out too. You would mind updating with that in mind?

(And also could you please squash the commits down to just one or two)

joranE:
Is your concern that `subset` and `inherit.aes` might go away at some point and you don't want them made "official"? The reason I ask is just to clarify whether you would you prefer no mention of the availability of `subset` and `inherit.aes` at all, or whether they could be mentioned under the `...` argument but with a warning that their functionality may change or disappear in the future?

hadley:
They're not going to go away, but equally I don't want people using them.

joranE:
Ok...I hope I navigated all the git squashing correctly...

hadley:
Could you please use the latest version of roxygen?  Squashing looks good

joranE:
Sorry...so many little details to remember! :) Does that look better?

hadley:
Thanks - I really appreciate the improvement due to the extreme fiddliness!  Any interest in doing similar work for the scales? I'm also thinking about documenting more of the stat arguments in the corresponding geom.

joranE:
No problem! I'd be happy to look at the scales and see what I can do. If there's anything specific bothering you about them, that would be helpful to know...

hadley:
Just the same problem as the geoms - that there's a big ... full of important arguments, and you have to go hunting around the docs to figure out the complete set of arguments that a function understands

PR #1084, opened 2015-04-13
Remove tick marks from theme_minimal

tomschloss:
A very minor change: I've removed tick marks from ` theme_minimal() ` as I think that is more in keeping with "minimalism". The black ticks seemed a little visually unnecessary. 

Incidentally, removing black ticks it also aligns minimal to the default plotly theme.

All good if you don't like the idea.

```r 
qplot(data = mtcars, x = mpg, y = wt) + theme_minimal() 
```
Before:
before

After:
after

Tom

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -177,7 +177,8 @@ theme_minimal <- function(base_size = 12, base_family = "") {
       panel.background  = element_blank(),
       panel.border      = element_blank(),
       strip.background  = element_blank(),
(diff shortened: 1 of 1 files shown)


hadley:
I think the numbers now look a bit far away from the axis. Could you tweak that too please?

Also needs a merge/rebase and a bullet point in NEWS

hadley:
Merged by hand - thanks for the patch!

PR #1085, opened 2015-04-15
Add linewidth option geom_path and geom_point.

SeySayux:
This PR adds an aesthetic option to `geom_point` and `geom_path` (and its derivatives, such as `geom_line`) to customize the width of the lines drawn (by passing the `linewidth` property). This is implemented by simply delegating to the `lwd` option of the `grid` library.

### DIFF
--- a/R/geom-path-.r
+++ b/R/geom-path-.r
@@ -167,7 +167,7 @@ GeomPath <- proto(Geom, {
           default.units="native", arrow = arrow,
           gp = gpar(
             col = alpha(colour, alpha)[!end], fill = alpha(colour, alpha)[!end],
-            lwd = size[!end] * .pt, lty = linetype[!end],
+            lwd = size[!end] * linewidth * .pt, lty = linetype[!end],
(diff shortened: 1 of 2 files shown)


hadley:
Why do you need this instead of size?

SeySayux:
I didn't realize the `size` option was available for `geom_path`. What I was specifically looking for was increasing the border thickness of `geom_point` shapes 21-25, which seems to be a common issue:

* http://stackoverflow.com/questions/19506630/control-point-border-thickness-in-ggplot
* http://stackoverflow.com/questions/17677687/change-thickness-of-a-marker-in-ggplot2

I only really added the same option to `geom_path` as an afterthought. If you want, I can resubmit the patch only for `geom_point`.

hadley:
Ah, got it - yes, I'd love a PR that just modified linewidth for geom_point. Maybe call the aesthetic stroke? Or strokewidth? I think you'll also need to update `draw_legend()` and it might be worthwhile thinking about what the default scale for stroke should be.

PR #1092, opened 2015-04-29
Feature/geom schematic

ogdenkev:
This commit adds a new geom, `geom_schematic()` which follows John Tukey's recommendation for balanced schematic plots (see Tukey J. W. (1990) Data-Based Graphics: Visual Display in the Decades to Come. Statistical Science 5(3), 327-339.) This also more closely follows what `bwplot()` does in the `lattice` package and what William Cleveland uses in his book Visualizing Data.

The `geom_schematic()` function started from the `geom_boxplot()` function, and added lines at the upper and lower ends of the whiskers, used a point for the median instead of a line. It also added the aesthetics `whiskers.linetype`, `middle.size`, and `middle.colour` with the default values of `dashed`, `4`, and `NULL`, respectively. The default for `middle.colour` is to use the `colour` aesthetic, but with the option of directly specifying it in the call to `geom_schematic()`. The option for notching was removed, because I didn't think notching would work well with the median as a point.

rplot


hadley:
This just feels a bit too specific for me - and there's quite a lot of code to implement it. I think it would be more appropriate in a separate package.

PR #1096, opened 2015-05-05
Fix how geom_dotplot handles some aes

evolvedmicrobe:
This commit allows geom_dotplot to display color attributes properly
when dots are binned.  Prior to this, if one ran a command such as:

    ggplot(mtcars, aes(x=factor(cyl), y =mpg, fill=factor(carb), group=factor(cyl))) + geom_dotplot(binaxis="y", stackdir="center")

The dots would be positioned on top of each other, because a "double
grouping" was performed where both the fill and the x axis were used to
divide groups.  This change keeps the groups appropriately separated
when the x/y positions are being determined, but still applies separate
coloring to each.

It does this by changing where the fill color is added to the data used
when plotting.  Previously, the data was condensed into bins and a
dataframe where each row represented several dots to be plotted during
the calculate groups step. This condendsed data was then re-expanded
with a constant fill color added for each binned group inside the
reparameterise method.  This commit re-expands the data into one line
per dot in the original calculate method, where the appropriate fill
(or other aes) is still available and can be added back in.

I verified that all the original examples for geom_dotplot still behave
as expected, and all tests currently pass.

### DIFF
--- a/R/geom-dotplot.r
+++ b/R/geom-dotplot.r
@@ -173,10 +173,6 @@ GeomDotplot <- proto(Geom, {
       stackaxismax <- .5
     }
 
-
-    # Fill the bins: at a given x (or y), if count=3, make 3 entries at that x
-    df <- df[rep(1:nrow(df), df$count), ]
-
     # Next part will set the position of each dot within each stack
     # If stackgroups=TRUE, split only on x (or y) and panel; if not stacking, also split by group
     plyvars <- params$binaxis %||% "x"
--- a/R/stat-bindot.r
+++ b/R/stat-bindot.r
@@ -97,6 +97,11 @@ StatBindot <- proto(Stat, {
                         origin = NULL, breaks = NULL, width = 0.9, drop = FALSE,
                         right = TRUE, ...) {
 
+    ## Save the original data, as we may project down to a smaller set of data here
+    ## and we need a way to add back in information that is lost during that projection
+    ## when we "project" back up to plot each point.
+    original_data <- data
+
     # This function taken from integer help page
     is.wholenumber <- function(x, tol = .Machine$double.eps^0.5) {
       abs(x - round(x)) < tol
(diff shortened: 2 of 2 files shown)


evolvedmicrobe:
I gather the failed tests are independent of this PR as several previous ones seem to generate the same error.

evolvedmicrobe:
Also, in case it is useful, some before and after for the command:

    ggplot(mtcars, aes(x=factor(cyl), y =mpg, fill=factor(carb))) + geom_dotplot(binaxis="y", stackdir="center")


before
after

hadley:
@wch could you please review the basic strategy? 

@evolvedmicrobe You'll also need to match the style of the existing code, but there's no point doing that until we're confident that the change as a whole is good to merge.

evolvedmicrobe:
@hadley @wch thanks for taking a look at this.  My hope is really just to commit something that produced the "after" plot rather than the "before" plot, happy to make any coding style (or other) changes to get that done.

wch:
A couple quick thoughts:

* It would be nice if the stacking order were consistent with respect to the group/color. In the example, in the left column of dots, the bottom two groups have different stacking order. I'm guessing they're in order of the y values?
* This may require extra care when using facets - you'd usually not want them to be grouped in this way across facets.
* This is similar to using `stackgroups=TRUE`, but not exactly, since the goal here is to share bin positions only for points that share the same x location (in the same facet), whereas with `stackgroups=TRUE`, the bin positions are shared across all data.

beroe:
Was this closed because it was fixed? And if so, what is the solution. I still get unpredictable spacing using ` ggplot(mtcars, aes(x=factor(cyl), y =mpg, fill=factor(carb))) + geom_dotplot(binaxis="y", stackdir="center",stackgroups = TRUE,binpositions="all")` or the code shown above in the before/after example.

evolvedmicrobe:
This issue was just closed and not fixed.

adempewolff:
Stumbled across this while trying to figure out how to produce something similar to the "after" plot above.  Can see the rational for not wanting to change the default behavior, but it would be nice to have this as an option at least.

vitorbl:
Solution would to feed the fill argument outside aes() into the dotplot layer, but *DATA MUST BE ORDERED FIRST*, firs with respect to your x variable ( factor(cyl) ), secondly with respect to you y variable (mpg). Solution would look a lot like that:
`data <- mtcars[with(mtcars,order(factor(cyl), mpg)),]
ggplot(mtcars, aes(x=factor(cyl), y =mpg, )) + geom_dotplot(binaxis="y", stackdir="center", fill = data$carb)
`
Downside is that you are losing legends, think it can be set manually. It's possible to build a function to address the colors in a very personal way returning strings that reger RGB values or color names.

evolvedmicrobe:
For what's its worth we were able to solve the problem and make a nice publication graphic (Figure 1 here: http://www.pnas.org/content/early/2017/07/12/1703338114.full.pdf).  I love this library, thanks ggplot!

kevinmickey:
This PR was closed, but it appears there is still an open issue: #1745.

PR #1100, opened 2015-05-08
Install Rexamine/stringi from GitHub for Travis

wibeasley:
https://github.com/Rexamine/stringi/issues/155


hadley:
Thanks - fixed after upgrading to modern travis spec

PR #1118, opened 2015-06-10
Example of stat_bin2d with a log-scaled axis

tgs:
I couldn't find an example of this anywhere, and it took me an hour to figure out, so I thought maybe it belonged in the documentation. I'm not sure if anyone else is trying to do this, though.

Here's the image that the added example generates:

semilog

And here's the current help page: <http://docs.ggplot2.org/current/stat_bin2d.html>

### DIFF
--- a/man/stat_bin2d.Rd
+++ b/man/stat_bin2d.Rd
@@ -56,6 +56,11 @@ x <- seq(min(diamonds$carat), max(diamonds$carat), by = 0.1)
 y <- seq(min(diamonds$price), max(diamonds$price), length = 50)
 d + stat_bin2d(breaks = list(x = x, y = y))
 
+# Or with a list of breaks, on a semi-log plot
+log_price_range <- log10(range(diamonds$price))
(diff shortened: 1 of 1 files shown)


hadley:
Why not just use `scale_y_log10()` ?

tgs:
You're right, this doesn't make sense on its own.  The issue I had is that I wanted to set explicit breaks for one dimension (x axis as date binned by month), and so I had to also set the breaks for the other dimension, because stat_bin2d requires that both dimensions of breaks be supplied the same way.  So you have to say `stat_bin2d(breaks=list(x = xbreaks, y = ybreaks))`.  I could have avoided the whole issue if I could have said `stat_bin2d(breaks = list(x = xbreaks, y = NA))`, and had it generate its default breaks for the y axis.  Maybe I should throw out this documentation patch, and write that functionality patch instead.  (or is there some other way I should have done it?)

tgs:
Take a look at https://github.com/hadley/ggplot2/pull/1126 instead?

PR #1126, opened 2015-06-14
Handle stat_bin2d(breaks=(x=NA,y=Something))

tgs:
With this change, you can specify one set of breaks to `stat_bin2d` without setting the other.  This is particularly useful when binning by date in one dimension, and using a transformed scale in the other dimension.  This way, you don't have to figure out how to reconstruct the default breaks for the transformed scale.  For example:

```r
breaks <- list( x = sort(unique(movies$year)) - 0.5, y = NA )
g <- ggplot(movies) + aes(x = year, y = length)
g + stat_bin2d(breaks=breaks) + scale_y_log10()
```

### DIFF
--- a/R/stat-bin2d.r
+++ b/R/stat-bin2d.r
@@ -33,6 +33,10 @@ StatBin2d <- proto2("StatBin2d", Stat,
       y = scale_dimension(scales$y, c(0, 0))
     )
 
+    # is.integer(...) below actually deals with factor input data, which is
+    # integer by now.  Bins for factor data should take the width of one level,
+    # and should show up centered over their tick marks.
+
     # Determine origin, if omitted
     if (is.null(origin)) {
       origin <- c(NA, NA)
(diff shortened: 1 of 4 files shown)


R/stat-bin2d.r
```diff
@@ -80,15 +80,19 @@ StatBin2d <- proto(Stat, {
...
     if (is.null(breaks)) {
-      breaks <- list(
-        seq(origin[1], max(range$x) + binwidth[1], binwidth[1]),
-        seq(origin[2], max(range$y) + binwidth[2], binwidth[2])
-      )
-    } else {
-      stopifnot(is.list(breaks))
-      stopifnot(length(breaks) == 2)
-      stopifnot(all(sapply(breaks, is.numeric)))
+      breaks <- list(x = NA, y = NA)
```
hadley: I think using `list(x = NULL, y = NULL)` makes the code below a bit simpler. Then you can just test `is.null(breaks$x)`

hadley:
Would you consider adding a couple of unit tests? I'd start with something like:

```R
p <- ggplot(df, aes(x, y)) + 
  stat_bin2d()
ggplot_build(p)$data[[1]]
```

R/stat-bin2d.r
tgs: Sure, sounds good.

tgs:
Sure, I'll write some more specific unit tests.  Actually, just got started, and found a bug (is.integer() doesn't work when calculating the binwidth, data$x is 'numeric').  I'll try and fix that too.

tgs:
Ah, the integer/numeric thing seems to be a feature, rather than a bug, of R.  OK.

tgs:
I'm going to open a separate issue for the is.integer thing.

hadley:
I think that's by design: at this point factors are integers and everything else is numeric. It's definitely weird but it's too late to change now

tgs:
AH! That makes sense!

tgs:
I see that man/stat_bin2d.Rd is deleted in master.  Where should documentation go?  Is there any chance of getting this change in the next version of ggplot2?

hadley:
The docs have been merged into `geom_bin2d()` (since they're basically the same).  If you rebase/merge, I can do another round of review and then it should be good to merge.

tgs:
I've merged, and moved my documentation change to geom_bin2d.Rd.  Ready for you to take another look :)

hadley:
Looks good.  Could you please add a bullet to NEWS (copying the existing format) and please rebase (again) now that I've fixed all the build problems?

tgs:
I rebased on master this morning, it looks like the build passes :)

tgs:
Woops, let me fix the documentation once more.

tgs:
OK, fixed.

hadley:
Thanks!

tgs:
Thank you too!

PR #1128, opened 2015-06-14
Complete the set of horizontal Geoms, Stats and Positions

lionel-:
This makes it easy to flip a layout horizontally on a layer basis as opposed to a plot basis (with `coord_flip()`). In addition to being convenient, this allows plots that are not currently possible because `facet_wrap()` does not allow coordinates filpping, or in situations where only some layers should be flipped.


hadley:
I'm not entirely opposed to this idea, but the implementation seems so inelegant.  I wonder if it would be possible to implement a `position_rotate()` that would allow you to use the existing geoms

lionel-:
I know... At the time I wrote this I've played around with the idea of adding an argument to geoms, but IIRC I encountered some problems that I could not easily solve. Since ggplot is not going to evolve much (or is it?) I thought it may be ok to have all this code repetition.

I don't have time to work on another solution however :/

hadley:
I'm going to close this for now, as it just feels a bit too icky (plus no one else has asked for horizontal versions of those geoms).   On the plus side, this issue was thought through much more thoroughly in ggvis, and rotating arbitrary geoms is much much easier.

lionel-:
well the horizontal version of linerange has been asked and is needed for coefficient plots. It's true there is no dire need to merge this since packages and users can implement their own horizontal geoms and colliders (as is currently done e.g. in https://github.com/jaredlander/coefplot/blob/master/R/dodging.r ).

lionel-:
I merely saw this patch as an easy way to make ggplot2 'complete', and in a way it's not ickier than `geom_errorbarh()` or `geom_vline()`/`geom_hline()` currently in the codebase ;)

hadley:
BTW I think we're going to resolve this problem in a more general way - I'm pretty sure we're going to switch to R6 and provide a documented way of adding additional geoms, stats, and position adjustments to ggplot2. Then you can make gghorizontal and we'll both be happy :)

lionel-:
that's good news!

smouksassi:
rplot

There is some plots that cannot be done in ggplot especially when scales are free and when we need rotation see attached done in lattice and it keeps that scales are "free" in ggplot when we do rotation it will keep all levels in all panels

smouksassi:
so I vote for having the possibility to do it !

hadley:
@smouksassi stay tuned for @lionel-'s gghorizon package ;)

lionel-:
@smouksassi here is the package https://github.com/lionel-/gghorizon

If you have time, could you check it with different combinations of geoms, stats and positions in horizontal layers?

smouksassi:
@lionel- Thanks  based on previous posts and requests here is an example:
see how the motivation is to be able to "free" the axis since this cannot
be done with coord_flip.
getting #Error in panel$y_scales[[this_panel$SCALE_Y]] : subscript out of
bounds

require(gghorizon)
require(ggplot2)

set <- factor(rep(1:3, each=50), labels=c("DRUG A","DRUG B","Vehicle"))
subject <- factor(rep(1:6, each=25),
                 labels=c("A","B","C","D","E","F"))
result <- rnorm(150)
test <- data.frame(set, subject, result)

p <- ggplot(test, aes(subject, result))+
  geom_boxplot()+
  facet_grid(. ~ set)
p

p1 <- ggplot(test, aes(subject, result))+
  geom_boxplot()+
  facet_grid(. ~ set, space="free",
                             scales="free_x")

p1 # works but  how to we flip this plot ?

p1 +coord_flip() # wrong output incorrect y axis labels

#trying the new geom geom_boxploth

p2 <- ggplot(test, aes( result,subject))+
  geom_boxploth()+
  facet_grid(set ~.)

p2 # ptoduce the plot but let us try freeing the y axis


p3 <- ggplot(test, aes( result,subject))+
  geom_boxploth()+
  facet_grid(set ~.,scales="free_y")

p3
#Error in panel$y_scales[[this_panel$SCALE_Y]] : subscript out of bounds



sessionInfo()
R version 3.2.3 (2015-12-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United
States.1252
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C

[5] LC_TIME=English_United States.1252

attached base packages:

other attached packages:
[1] ggplot2_2.0.0        gghorizon_0.0.0.9000

 [1] Rcpp_0.12.3       digest_0.6.9      plyr_1.8.3        grid_3.2.3
 [5] R6_2.1.1          gtable_0.1.2      magrittr_1.5      scales_0.3.0
 [9] httr_1.0.0        stringi_1.0-1     reshape2_1.4.1    lazyeval_0.1.10
[13] curl_0.9.4        rmarkdown_0.8.1   labeling_0.3      devtools_1.9.1
[17] tools_3.2.3       stringr_1.0.0     purrr_0.2.0       munsell_0.4.2
[21] yaml_2.1.13       rsconnect_0.4.1.4 colorspace_1.2-6  memoise_0.2.1
[25] htmltools_0.3

lionel-:
thanks. Let's move to https://github.com/lionel-/gghorizon/issues/1 so that Hadley doesn't get notified all the time ;)

PR #1133, opened 2015-06-16
Add stroke option to geom_point.

SeySayux:
This PR adds an aesthetic option `stroke` to `geom_point` that makes it possible to specify the stroke width of point shapes 21-25 (see #1085, I had to resubmit because you can't force push when an issue is closed).


hadley:
Could you please:

* re-document, so travis passes
* add an example of usage to `geom_point()`

SeySayux:
I've added a basic example and made sure all the checks pass without warnings now. Are there any other places where I need to update documentation as well?

hadley:
Thanks!

PR #1134, opened 2015-06-16
Fix all found instances of partial matching

jimhester:
These were found by setting

```
options(warnPartialMatchAttr = TRUE,
    warnPartialMatchDollar = TRUE,
    warn = 2)
```

Using hadley/testthat#269 with `options("testthat.suppress" = NULL)` and running
`devtools::test()` and `vtest::vtest()` until there were no more partial
match messages.

There may still be some instances in the code base, but I think this
finds the vast majority.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,9 @@
 ggplot2 1.0.1.9000
 ----------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


hadley:
Thanks!

PR #1139, opened 2015-06-18
Add access to direction argument of scales::brewer_pal

jiho:
Plus minor example and documentation fixes

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,9 +1,9 @@
+ggplot2 1.0.1.9000
+----------------------------------------------------------------
+
 * Add access to the new `direction` argument of `scales::brewer_pal` and
   simplify changing the order of colours for `scale_*_brewer` and
   `scale_*_distiller` (@jiho, #1139)
-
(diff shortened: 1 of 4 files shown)


NEWS
```diff
@@ -1,3 +1,7 @@
+* Add access to the new `direction` argument of `scales::brewer_pal` and
```
hadley: Sorry I think I was unclear bfore, this should be the first bullet under the version heading

jiho: Oh, OK. I thought this is what I always did so I misunderstood your comment regarding this. I'll correct it.

BTW, do you prefer 1-commit pull requests (in which case I overwrite the changes each time and force push, so I guess line comments like this disappear) or many small commits are OK (in which case the "history" of the PR itself is kept but it gets messier when merged).

hadley: Small commits is fine - if it gets too messy I'll ask you to flatten.

hadley:
Thanks!

PR #1148, opened 2015-06-24
Convert from proto to proto2

wch:
This is just the start of the conversion to R6.

This example works:

```R
ggplot(mtcars, aes(wt, mpg)) + geom_text2(label="x")
```

Objects like `GeomText` are instantiated at run time (in the call to `Layer$new()`), so that if any external packages inherit from, say, `Geom`, they'll inherit from the current version of ggplot2, not whatever version they were built against.

@hadley, currently the `geom_text2` function doesn't separate the `geom_params` from other `params`, because some of the `geom_params` include properties like `label`, `size`, etc. So I'm letting the `LayerR6$initialize` function sort them out, as was done previously. I think it's a good idea to explicitly sort out the geom params from stat params in the `geom_text2` functions if we can, but it'll require a little work.


R/geom-text.r
```diff
@@ -143,6 +143,95 @@ GeomText <- proto(Geom, {
...
+
+    position <- position_nudge(nudge_x, nudge_y)
+  }
+
+  LayerR6$new(
+    data = data,
+    mapping = mapping,
+    stat = stat,
+    geom = GeomTextR6,
+    position = position_nudge(nudge_x, nudge_y),
```
hadley: Can you change this to position please?

wch:
I've completed the conversion. All tests pass! I haven't done visual tests yet, however.

Some things to think about:

* Because this code was originally designed to work with singletons (because that's how proto works), it forces us to either (1) instantiate all the classes immediately, or (2) instantiate them as needed. I went with (2). The drawback to (1) is that you might end up having conflicts with packages that have classes which inherit from classes in ggplot2, if they're built against a different version of ggplot2. The drawback of (2) is that it's a little inefficient and inelegant in places, because we end sometimes up instantiating objects just to invoke a single method from the object. It might be possible to solve this by modifying R6 to have static methods.
* I don't like the existing print methods for stats, geoms, positions, and layers. When I was working on it, I found the default print R6 method to be more useful because it tells you what the object actually is, instead of printing out a confusing code fragment.
* Many of the `geom_xxx` and `stat_xxx` functions didn't have an explicit `inherit.aes` or `show_guide`; those arguments are captured in the `...` and passed to the `params` argument. But this no longer works correctly, so I think we should be explicit about having those args (and others) in every geom and stat. There are other ways to work around it, but I think this is the best option, since it also means they will get documented properly.
* It would be nice to provide cleaner separation of `geom_params`, `stat_params`, and `params`. There were many places I saw where the `geom_xxx` or `stat_xxx` function took params that were obviously for the stat or geom, so we should send those on their way if needed.
* I commented out the bolus/digest code since it didn't seem like it was doing anything, and it used proto.

R/annotation-custom.r
```diff
@@ -44,32 +44,41 @@ NULL
 #' base +
 #'   annotation_custom(grob = g, xmin = 1, xmax = 10, ymin = 8, ymax = 10)
 annotation_custom <- function (grob, xmin = -Inf, xmax = Inf, ymin = -Inf, ymax = Inf) {
-  GeomCustomAnn$new(geom_params = list(grob = grob, xmin = xmin,
-    xmax = xmax, ymin = ymin, ymax = ymax), stat = "identity",
-    position = "identity", data = NULL, inherit.aes = TRUE)
+  Layer$new(
+    data = NULL,
+    stat = "identity",
```
hadley: Would it be better to do `StatIdentity` here?

R/geom-bin2d.r
```diff
@@ -11,23 +11,38 @@
...
 
-  guide_geom <- function(.) "polygon"
+    objname = "bin2d",
 
-  default_stat <- function(.) StatBin2d
-  required_aes <- c("xmin", "xmax", "ymin", "ymax")
-  default_aes <- function(.) {
-    aes(colour = NA, fill = "grey60", size = 0.5, linetype = 1, weight = 1, , alpha = NA)
-  }
+    guide_geom = function() "polygon",
```
hadley: Is objname still needed?

hadley:
That all sounds reasonable. It's a bit hard to comment on such a large PR, so maybe I'll read through it and make a separate list of thoughts (esp. since many aren't necessarily related to your changes)

R/annotation-custom.r
wch: Sure, that sounds reasonable. That reminds me of another general issue which I'll add to my list above.

R/geom-bin2d.r
wch: Probably not. Maybe it could be replaced with a method in the `Geom` class.

wch:
A few more issues I thought of:

* It can be confusing when the user has to pass in a string like `stat="bin"` and other times you use `"StatBin`. And there's also the related `stat_bin`. It's not obvious how all these things are related. We should at the very least clarify in the docs how they're connected.
* Similarly the `guide_geom()` method (and probably others) returns a string, like this: `guide_geom = function() "polygon"`, but it might as well return the object directly, like `guide_geom = function() GeomPolygon`.
* Some members are functions, and others are non-functions, and it's not clear to me why. For example, in [`Stat`](https://github.com/wch/ggplot2/blob/01e94f2/R/stat-.r), `default_aes` is a function, but `required_aes` is a character vector.

wch:
I ran a simple benchmark comparing the CRAN version to this branch. Times are essentially the same:

```R
library(ggplot2)

times <- numeric(0)

for (i in 1:20) {
  times[length(times) + 1] <- system.time({
    print(ggplot(mtcars, aes(wt, mpg)) + geom_point())
    
    print(
        geom_point(colour="black", size = 4.5) +
        geom_point(colour="pink", size = 4) +
        geom_point(aes(shape = factor(cyl)))
    )
    
    print(ggplot(mtcars, aes(wt, mpg, label = rownames(mtcars))) + geom_text())
  })
}


# ggplot2 1.0.1
mean(times)
# [1] 0.25075

# R6 version
mean(times)
# [1] 0.2523
```

hadley:
* I think we should stick to using object rather than strings in the internal code.

* I don't remember why some things are methods and some are fields - I think we should use functions everywhere

* Yes, it's confusing that `stat_bin()` creates a layer - at least this is a bit less confusing in the source now.

* I think as many as possible of the defaults should move out of (e.g.) `StatBin` and into `stat_bin` (but I don't think that can happen with default aesthetics)

wch:
I've improved the way that `super` works.

I've also added `print` and `format` methods for proto2 objects. Proto2 objects can also now take a class. Here's the example of output when printed:

``` 
> C
C <proto2 object>
    getx: function
    x: 2
    super: B <proto2 object>

> print(C, flat=F)
C <proto2 object>
    getx: function
    super: B <proto2 object>
        getx: function
        x: 2
        super: A <proto2 object>
            getx: function
            x: 1
```

Issue #707, opened 2012-10-26
Feature request: Add a 'step' option to geom_ribbon


meow9th:
I'm interested in being able to plot confidence intervals/bands for step functions (such as nonparametric estimated survival curves) using geom_ribbon. I've hacked up a solution using geom_rect, but I thought it would be nice if I could use geom_ribbon instead.

Here is an example using geom_rect:

``` R
library(survival)
aml.fit = survfit(Surv(time, status) ~ 1, data = aml)
aml.gg = do.call(data.frame,
                 summary(aml.fit)[c('time', 'surv', 'lower', 'upper')])
aml.rect = aml.gg[-nrow(aml.gg), ]
aml.rect$timemax = aml.gg$time[-1]
ggplot(aml.gg) +
  geom_step(aes(x = time, y = surv)) +
  geom_rect(data = aml.rect,
            aes(xmin = time, xmax = timemax, ymin = lower, ymax = upper),
            alpha = 0.1)
```

Here is how I would hope to use geom_ribbon instead:

``` R
ggplot(aml.gg) +
  geom_step(aes(x = time, y = surv)) +
  geom_ribbon(aes(x = time, ymin = lower, ymax = upper),
              step = T)
```

kendonB:
+1

kendonB:
Would the solution at https://groups.google.com/forum/?fromgroups=#!topic/ggplot2/9cFWHaH1CPs get accepted?

PR #1149, opened 2015-06-25
Implements stepped ribbon

felasa:
Implements a version of #707


R/stat-stepribbon.r
```diff
@@ -0,0 +1,68 @@
...
+  desc_outputs <- list(
+    x = "stepped independent variable",
+    ymin = "stepped minimum dependent variable",
+    ymax = "stepped maximum dependent variable"
+  )
+  required_aes <- c("x", "ymin", "ymax")
+  default_geom <- function(.) GeomRibbon
+  default_aes <- function(.) aes(x = ..x.., ymin = ..y.., ymax = Inf)
+
+  calculate <- function(., data, scales, yvars = c("ymin", "ymax"), direction = c("hv", "vh"), ...) {
```
hadley: Couldn't this be implemented as part of the existing `stairstep` function?

hadley:
Given that this is needed relatively rarely, and ggplot2 now has an official extension mechanism (soon to be documented), I think this should probably go in a separate package.

joshmyzie2:
Did this get committed anywhere?  It would be nice to have a cadlag step ribbon for time series data.

felasa:
This wasn't commited and the code no longer works after 2.0's ggproto system. But it can easily be adapted and run in a session if needed.

Issue #102, opened 2010-05-21
stat_bin should handle a number of bins in addition to binwidth and breaks


jiho:
Use case: when plotting the histograms of several variables at once, the user wants some control over the fineness of the breaks but cannot provide breaks of binwidth because the units of the variables are very different.

```
library(ggplot2)
d = data.frame(id=1:100, a=runif(100), b=runif(100)*20, c=runif(100)-10)
d = melt(d, id="id")
ggplot(d) + geom_bar(aes(x=value)) + facet_wrap(~variable, scales="free")
```

PR #1158, opened 2015-07-06
`bins` argument to `stat_bin`

tmshn:
As discussed in issue #102 , it will be nicer if `geom_histogram` can handle the number of bins, not the width.
To achieve it, I added new argument `bins` to `stat_bin`, which denotes number of bins.

This is less prior to `breaks` or `binwidth`, so with these arguments, `bins` will not work.


hadley:
Could you please merge/rebase and add a bullet to NEWS?

tmshn:
Done.
Sorry for so many commits :(

hadley:
Thanks!

PR #1160, opened 2015-07-10
Quantile lines on violin plots

DanRuderman:
Violin plots are a nice way to visualize data density, but it is also useful to know where the median and other quantiles lie. Overlaying a box plot can add a lot more than is needed, and it doesn't offer a lot of flexibility in the quantiles to display. This new feature adds horizontal lines to violin plots which mark quantiles specified by the user. In polar plots they are circular arcs, which may or may not be the right behavior.

As an example, this is the visual test:
violin_quantiles

Also fixed a documentation issue in utilities.r

This is my first pull request (and I am new to Git to boot), but hopefully I didn't make too many mistakes in the process.

Thanks
Dan


R/geom-violin.r
```diff
@@ -107,7 +110,19 @@ GeomViolin <- proto2(
       # Needed for coord_polar and such
       newdata <- rbind(newdata, newdata[1,])
 
-      ggname(self$my_name(), GeomPolygon$draw(newdata, ...))
+      if (any(data$is.quantile)) {
```
hadley: I think these should be a parameter of the draw function, rather than an aesthetic.

DanRuderman: Do you mean *is.quantile* vector should be passed as a parameter to the draw function? Or do you mean that the *quantiles* desired should be passed? For the latter, I placed it in the *ydensity* stat because I don't believe the raw data can be accessed from the *geom-violin* draw function. Or can they?

If there is existing code that does what you are referring to and can act as a template for me, please point the way.

thx
dan

hadley:
Thinking about this more, do you expect the quantiles to be derived from the original data, or from the smoothed density?

hadley:
It's tricky to do this right because there's no easy way to get `density()` to estimate at the required quantiles. I think this means we need to it via linear interpolation (which is fine because the errors are likely to be sub-pixel). 

The code in `GeomViolin$draw()` should look something like this:

```R
if (length(draw_quantiles) > 0) {
  dens <- cumsum(data$density) / sum(data$density)
  ecdf <- approxfun(dens, data$y)

  ys <- ecdf(draw_quantiles)
}
```

where `draw_quantiles` is a numeric vector giving the desired quantiles. You'd apply the same idea to find the correct values for xmin and xmax.  Does that make sense?

NeuKon:
I would expect a violin plot to calculate quantiles on the original data. So if one would create a violin plot with quaRtiles, it would be an "ordinary" box plot with a violin shape instead of a rectangle. 

I believe seaborn is also calculating the quantiles on the raw data. (https://github.com/mwaskom/seaborn/blob/f4e777743a93b01ad0499c6f35c2a112161de6ad/seaborn/categorical.py , "draw_quartiles" on line 994)

The violin plot parameter "inner" (http://web.stanford.edu/~mwaskom/software/seaborn/generated/seaborn.violinplot.html), allows to add a boxplot / quaRtiles / ... inside the violin. 

"inner: Representation of the datapoints in the violin interior. If box, draw a miniature boxplot. If quartiles, draw the quartiles of the distribution. If point or stick, show each underlying datapoint. Using None will draw unadorned violins."

I hope this is of any help.

DanRuderman:
The issue of real vs estimated quantiles was my concern and why I went with the real ones, which sort of broke the clean separation of `geom` from `stat`. Since a violin plot is an estimated distribution anyway I can see the logic in using the approximate data quantiles based on that distribution. It has the added advantage that it takes into account the `weight` parameter.

Perhaps another direction would be to pass `y` locations of horizontal lines to be drawn to `geom_violin()` as a parameter, but I don't know if there is a mechanism for `stat_ydensity()` to add these values to the parameter list.

DanRuderman:
This weekend I'll make the changes Hadley suggested.

Point of procedure: do I close this pull request and open a new one?

hadley:
You modify this one by pushing the branch where the changes are

DanRuderman:
OK - I've made the changes and merged in the 280 or so changed files on master. Seems to be working fine. However I'm having some trouble getting the tests to run (using `devtools::test()`). Here is the output I see:

> Utilities : ................
>
> 1. Failure (at test-ggsave.R#26): can't guess if no graphics device open -------

Any help appreciated.

hadley:
You're probably running the tests in the current session which doesn't work. As long as it passes r cmd check you're ok.

DanRuderman:
OK - I think it's ready to go. Thanks for your help.

R/geom-violin.r
```diff
@@ -105,11 +111,32 @@ GeomViolin <- ggproto("GeomViolin", Geom,
...
 
-    ggname("geom_violin", GeomPolygon$draw(newdata, ...))
+    # Draw quantiles if requested
+    if (length(draw_quantiles) > 0) {
+      dens <- cumsum(data$density) / sum(data$density)
+      ecdf <- approxfun(dens, data$y)
+      ys <- ecdf(draw_quantiles)
+
+      # Create list of path grobs for the quantile segments
+      quantile_grob_list <- lapply (ys, function(y_to_match) {
```
hadley: Can you vectorise this so it returns a single grob (i.e. do the data manip inside `lapply()` and then call `GeomPath$draw()` only once? Otherwise it's going to be quite slow. 

I'd also recommend pulling out this code into a separate function. I think it should take `x` and `y` as inputs, and return a data frame of `x1`, `x2`, `y1`, `y2` suitable to be passed to `GeomSegment$draw()` (which I think is a better fit here).

hadley:
Ok, now I think the overall strategy is good, and we can focus on the details. I think pulling out most of the computation into a separate function is important because then you can also add a few unit tests.

DanRuderman:
OK. I added the function to take care of the hard work and a unit test for it. Just one test so far, but could add others if you like. Fixed the multiple grob efficiency issue. Added a sanity check for the supplied `draw_quantiles`.

It's just too bad these violin strings go in the wrong direction.

R/geom-violin.r
```diff
@@ -119,3 +136,22 @@ GeomViolin <- ggproto("GeomViolin", Geom,
...
+# Returns a data.frame with info needed to draw quantile segments.
+create_quantile_segment_frame <- function(data, draw_quantiles) {
+  dens <- cumsum(data$density) / sum(data$density)
+  ecdf <- approxfun(dens, data$y)
+  ys <- ecdf(draw_quantiles) # these are all the y-values for quantiles
+
+  # Create data frame containing quantile segment data, one group per quantile
+  do.call ( rbind, lapply (ys, function(y_to_match) {
+    y_mismatches <- abs(y_to_match-data$y)
+    match_index <- which(y_mismatches==min(y_mismatches))[1]
```
hadley: Wouldn't this be more easily done with `approxfun()`? The aesthetics within a violin are constant, so I don't think you need to do any matching up with rows in the original data

R/geom-violin.r
```diff
@@ -105,11 +111,22 @@ GeomViolin <- ggproto("GeomViolin", Geom,
...
     newdata <- rbind(newdata, newdata[1,])
 
-    ggname("geom_violin", GeomPolygon$draw(newdata, ...))
+    # Draw quantiles if requested
+    if (length(draw_quantiles) > 0) {
+      stopifnot(all(draw_quantiles >= 0) & all(draw_quantiles <= 1))
+
+      quantile_grob <- GeomPath$draw(create_quantile_segment_frame(data, draw_quantiles), ...)
+
+      ggname("geom_violin",
```
hadley: I think you can remove the `do.call()` and now use `grobTree()` directly.

R/geom-violin.r
```diff
@@ -75,7 +81,7 @@ geom_violin <- function(mapping = NULL, data = NULL, stat = "ydensity",
       trim = trim,
       scale = scale
     ),
-    params = list(...)
+    params = list(draw_quantiles = draw_quantiles, ...)
```
hadley: This should go in `stat_params`, not params, because you know that it belongs to the stat (and not the geom)

NAMESPACE
```diff
@@ -1,4 +1,4 @@
-# Generated by roxygen2 (4.1.1): do not edit by hand
+# Generated by roxygen2 (4.1.1.9000): do not edit by hand
```
hadley: Could you please document with released roxygen2? (I know this is a pain and the next version of roxygen will fix it)

hadley:
Looking good ;) Now we're on to the minor code style stuff

R/geom-violin.r
DanRuderman: I'm a little confused by this. How will the draw() function get access to it?

hadley: Ooops, sorry - I mean it should go in `geom_params`

DanRuderman:
OK - I made all the changes you suggested. Let me know if they look ok.

I think I've figured out how to get free R lessons :)

hadley:
Thanks!

PR #1203, opened 2015-07-25
spelling fixes

bbolker:
this is admittedly pretty trivial, but typos bug me ...  I tried to leave the UK/Aus/NZ "-ise" endings alone as well as things like "facetting" (which according to my North American rules should be "faceting") ...


hadley:
Thanks!

PR #1208, opened 2015-07-27
Fix issues with switched facet strips

### LINKED ISSUE #1170: Extra padding is added between facet labels on vertical axis and plot when using also facets on horizontal axis
Using the `facet_grid` option of `switch = "y"` for vertical facets and also using facets on the horizontal axis, I get some padding between the labels on the vertical axis and the plot. This padding is not present if I only plot the vertical facets.

See plots for illustration:

 **vertical facets only**
image

**vertical & horizontal facets**
image

The code for both is identical (except for the facet definition) and will attach it below. I have attempted all the options I could find, to get rid of the padding, with no success. Therefore, I believe this might be a bug, since it would make sense to be able to have no padding between the facet labels and the plot (esp. since there is none if there are no "horizontal" facets).

Using ggplot2 version 1.0.1.9002.

See code bellow: 

```
library(ggplot2)
data("iris")

# add another column to data to facet by in the second example
iris2 <- iris
iris2$Species2 <- rep(c("one","two","three"), 50)

# plot with facet on vertical axis
ggplot(data=iris2, aes(x=Sepal.Length, y=Sepal.Width )) +
  geom_point() +
  facet_grid(Species ~ . , switch = "y") +
  theme(axis.text.y = element_blank(),
        axis.ticks.y = element_blank(),
        axis.ticks.length = unit(0, "cm"),
        axis.ticks.margin = unit(0, "cm"),
        axis.line.y = element_blank(),
        panel.border = element_blank(),
        panel.margin = unit(0, "cm"),
        strip.text.y = element_text(angle = 180),
        strip.switch.pad.grid = unit(0, "cm")
  )

# plot with facets on both axis
ggplot(data=iris2, aes(x=Sepal.Length, y=Sepal.Width )) +
  geom_point() +
  facet_grid(Species ~ Species2 , switch = "y") +
  theme(axis.text.y = element_blank(),
        axis.ticks.y = element_blank(),
        axis.ticks.length = unit(0, "cm"),
        axis.ticks.margin = unit(0, "cm"),
        axis.line.y = element_blank(),
        panel.border = element_blank(),
        panel.margin = unit(0, "cm"),
        strip.text.y = element_text(angle = 180),
        strip.switch.pad.grid = unit(0, "cm")
        )


```

lionel-:
Fixes #1170.

I also adjusted the default padding. Before:

before

After:

after

### DIFF
--- a/R/facet-grid-.r
+++ b/R/facet-grid-.r
@@ -323,8 +323,8 @@ facet_render.grid <- function(facet, panel, coord, theme, geom_grobs) {
       complete <- rbind(center, bottom, z = c(1, 2))
     } else if (switch_y) {
       top <- strips$t
-      top <- gtable_add_cols(top, gt_r$widths, pos = 0)
       top <- gtable_add_cols(top, axes$l$widths, pos = 0)
+      top <- gtable_add_cols(top, gt_r$widths, pos = 0)
 
       center <- cbind(gt_r, axes$l, panels, z = c(3, 2, 1))
       bottom <- axes$b
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -107,8 +107,8 @@ theme_grey <- function(base_size = 12, base_family = "") {
     strip.background =   element_rect(fill = "grey80", colour = NA),
     strip.text.x =       element_text(),
     strip.text.y =       element_text(angle = -90),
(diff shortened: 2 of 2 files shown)


hadley:
Thanks!

Issue #992, opened 2014-08-05
geom_boxplot with continuous x and without group aesthetics


krlmlr:
```
d <- data.frame(x=rep(1:10, 100), y=runif(1000))
ggplot(d) + geom_boxplot(aes(x=x, y=y))
```

will happily produce a single box, while the intention is probably to produce the output of this code:

```
d <- data.frame(x=rep(1:10, 100), y=runif(1000))
ggplot(d) + geom_boxplot(aes(x=x, y=y, group=x))
```

Would a warning be appropriate here -- if `x` is continuous and the `group` aesthetics is missing?

hadley:
Maaaaybe, but I like that there's currently one unifying principle for grouping (i.e. all discrete variable for a group).  But if you feel strongly about, I'd review a PR to add a warning.

PR #1211, opened 2015-07-28
Use group identifier -1 if no group aesthetic is specified

krlmlr:
to allow distinguishing this case (prerequisite for #992)

This implicitly assumes `plyr::id` doesn't ever return 0. Tests in `plyr` test that sequences created by `id` start at 1, so we're probably safe here.

All tests pass, but I'm not sure if this has any other implications I'm not aware of.

### DIFF
--- a/R/utilities-layer.r
+++ b/R/utilities-layer.r
@@ -4,7 +4,8 @@ NO_GROUP <- -1L
 #
 # If the \code{group} variable is not present, then a new group
 # variable is generated from the interaction of all discrete (factor or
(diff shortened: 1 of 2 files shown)


hadley:
How about making it -1, just to make it more clear that it's a special value?

krlmlr:
Done, and also extracted to constant.

Note that the test for d53a584 passed, although it [shouldn't have](https://travis-ci.org/hadley/ggplot2/builds/72998867#L1849). (Tests pass for 21a69dc on my machine.)

PR #1212, opened 2015-07-28
stat_boxplot warns if x is continuous and no group aesthetics is specified

### LINKED ISSUE #992: geom_boxplot with continuous x and without group aesthetics

krlmlr:
- Only if there is more than one x value
- With tests

Fixes #992.

Closes #1213 (=includes it).

### DIFF
--- a/R/stat-boxplot.r
+++ b/R/stat-boxplot.r
@@ -47,6 +47,11 @@ StatBoxplot <- ggproto("StatBoxplot", Stat,
   {
     data <- remove_missing(data, na.rm, c("x", "y", "weight"), name = "stat_boxplot",
       finite = TRUE)
+    if (!is.discrete(data$x) && !has_groups(data) && any(data$x != data$x[1L])) {
+      warning(
+        "Continuous x aesthetic -- did you forget aes(group=...)?",
+        call. = FALSE)
+    }
     data$weight <- data$weight %||% 1
     width <- width %||%  resolution(data$x) * 0.75
 
--- a/R/stat-boxplot.r
+++ b/R/stat-boxplot.r
@@ -47,7 +47,7 @@ StatBoxplot <- ggproto("StatBoxplot", Stat,
   {
     data <- remove_missing(data, na.rm, c("x", "y", "weight"), name = "stat_boxplot",
       finite = TRUE)
-    if (!is.discrete(data$x) && !has_groups(data) && any(data$x != data$x[1L])) {
+    if (is.double(data$x) && !has_groups(data) && any(data$x != data$x[1L])) {
       warning(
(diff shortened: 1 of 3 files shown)


R/stat-boxplot.r
```diff
@@ -45,6 +45,11 @@ StatBoxplot <- ggproto("StatBoxplot", Stat,
   {
     data <- remove_missing(data, na.rm, c("x", "y", "weight"), name="stat_boxplot",
       finite = TRUE)
+    if (!is.discrete(data$x) && data$group[1L] == NO_GROUP && any(data$x != data$x[1L])) {
```
hadley: Did you check this? I think by this point factors have been converted into integers, so the appropriate test is `is.numeric(data$x)`.  And I think `all(data$group == NO_GROUP)` would be clearer

hadley:
Could you please squash the commits?

R/stat-boxplot.r
krlmlr: Thanks. I'll add a test later tonight.

`data$group[1L] == NO_GROUP` is much faster for the (expected) case that the user has specified a group. I'd rather not add an extra slowdown, instead I could add a comment or refactor this into a function. Please confirm.

hadley: Seems unlikely to be a bottleneck - easily understood code would be better

krlmlr:
Squashed commits, all tests pass locally. I hope we can agree on `has_groups()` (#1213).

krlmlr:
The case of facets hasn't been covered yet. I've added a test and included #1236 in this pull request.

R/stat-boxplot.r
```diff
@@ -47,6 +47,11 @@ StatBoxplot <- ggproto("StatBoxplot", Stat,
   {
     data <- remove_missing(data, na.rm, c("x", "y", "weight"), name = "stat_boxplot",
       finite = TRUE)
+    if (!is.discrete(data$x) && !has_groups(data) && any(data$x != data$x[1L])) {
```
hadley: I'm pretty sure the correct test is `is.double()`

krlmlr:
Changed test to `is.double()`. Tests pass locally.

hadley:
Thanks!

PR #1217, opened 2015-07-29
add reference to paper introducing violin plots

tklebel:
Since violin plots are not known in every field of data analysis yet, it could be helpful to include a paper introducing the concept.

Hintze et al. seems to be the first and only paper on this matter: http://amstat.tandfonline.com/doi/abs/10.1080/00031305.1998.10480559


hadley:
Thanks!

PR #1229, opened 2015-07-31
Modify docs of gg-add to match examples

jdagilliland:
The descriptions of the `+` and `%+replace%` operators did not seem to match
the behaviour of the last example on the gg-add docs page; they have been
reworded.

I'm still not sure whether this wording is optimally consistent with how you would like the behaviour of these operators described, but I gave it my best shot. Please let me know if you think this change is worthwhile, or itself in need of changing.


hadley:
Seems better to me too

PR #1230, opened 2015-07-31
Allow snake case arguments

jimhester:
This is perhaps a silly pull request.  It allows all ggplot2 arguments to be specified as snake_case rather than dot.case.

It retains the existing arguments and backwards compatibility.  Both `devtools::test()` and `vtest::vtest()` pass so no existing code should break.

Basically how this works is to add additional formal snake_case arguments for any existing dot.case arguments (with a default value of NULL) and modifies the existing dot.case formals to be `= snake_case %||% (previous default)`.  Therefore the dot.case argument gets the snake_case value if it is set, otherwise it gets it's old default value.  This is done for both normal functions as well as the ggproto class functions.

All of the modifications are done within `.onAttach()`, so there are no additional NOTEs about documentation mismatches from `R CMD check`.

Running `system.time(load_all())` with  `alias_to_snake_case("ggplot2")` included takes 1.390 seconds and commented out takes 0.966 seconds.  I would guess a similar difference would be found on the normal loading times, but I have not explicitly tested it.

I did not add any positive tests for this as I was unsure if it would be accepted, but you can test the functionality using the following code modified from the recent draft chapter programming with ggplot2.

```r
geom_mean <- function(..., bar_params = list(), errorbar_params = list()) {
  params <- list(...)
  bar_params <- modifyList(params, bar_params)
  errorbar_params  <- modifyList(params, errorbar_params)
  
  bar <- do.call("stat_summary", modifyList(
    list(fun_y = "mean", geom = "bar", fill = "grey70"),
    bar_params)
  )
  errorbar <- do.call("stat_summary", modifyList(
    list(fun_data = "mean_cl_normal", geom = "errorbar", width = 0.4),
    errorbar_params)
  )

  list(bar, errorbar)
}

ggplot(mpg, aes(class, cty)) + 
  geom_mean(
    colour = "steelblue",
    errorbar_params = list(width = 0.5, size = 1)
  )
ggplot(mpg, aes(class, cty)) + 
  geom_mean(
    bar_params = list(fill = "steelblue"),
    errorbar_params = list(colour = "blue")
  )
```

### DIFF
--- a/R/stat-summary.r
+++ b/R/stat-summary.r
@@ -142,7 +142,7 @@ StatSummary <- ggproto("StatSummary", Stat,
   {
     data <- remove_missing(data, na.rm, c("x", "y"), name = "stat_summary")
 
-    if (!missing(fun.data)) {
+    if (!is.null(fun.data)) {
       # User supplied function that takes complete data frame as input
       fun.data <- match.fun(fun.data)
       fun <- function(df, ...) {
--- a/R/utilities.r
+++ b/R/utilities.r
@@ -311,3 +311,132 @@ deparse2 <- function(x) {
     paste0(y[[1]], "...")
   }
 }
+
+# convert any formals using dotted argument names to snake case
+snake_case_formals <- function(fmls) {
+
+  dotted_fmls <- grepl("[[:alpha:]]\\.[[:alpha:]]", names(fmls))
+
+  if (any(dotted_fmls)) {
+    dotted_fmls_nms <- names(fmls)[dotted_fmls]
+    underscore_fmls <- rep(list(NULL), length(dotted_fmls_nms))
+    names(underscore_fmls) <- gsub("\\.", "_", dotted_fmls_nms)
+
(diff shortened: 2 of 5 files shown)


jimhester:
Blah it looks like R CMD check is complaining after all (I tested it but perhaps before the replacement formals were working properly).

0dd7a9b uses the global option `ggplot.snake_case` to control whether snake_case arguments are allowed, It is off by default so R CMD check will not complain.

hadley:
Dear Jim,

You scare me.

Regards,

Hadley

:stuck_out_tongue:

hadley:
This just seems like a step too far to me - for better or worse (mostly worse) ggplot2 is stuck with `_` in function names and `.` in arg names.

jimhester:
I figured you might say that, it was a interesting experiment to get working at least.

hadley:
I think you'd get buy in from Dirk if you made it convert to camel case ;)

PR #1261, opened 2015-08-12
add horizontal linerange geom

bbolker:
I believe this is all done (haven't reoxygenized it).  As demonstrated in the example, it's useful when `coord_flip()+ geom_linerange()` is not feasible, e.g. when using `facet_wrap()`.

I'm violating DRY quite a bit here -- it might be better to add an orientation argument to `geom_linerange`?  I'm willing to try that if it's thought to be better, but what's here is reasonably useful as is.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,8 @@ Suggests:
     nlme,
     testthat,
     quantreg,
-    knitr
+    knitr,
+    nlme
 Enhances: sp
 License: GPL-2
 URL: http://ggplot2.org, https://github.com/hadley/ggplot2
--- a/R/geom-linerangeh.r
+++ b/R/geom-linerangeh.r
@@ -17,12 +17,10 @@
 #' @examples
 #' # Generate data: means and standard errors of means for prices
 #' # for each type of cut
-#' require("nlme")
-#' require("reshape2")
-#' dmod <- lmList(price ~ cut|clarity, data=diamonds)
-#' mm <- setNames(melt(unclass(intervals(dmod))),
+#' dmod <- nlme::lmList(price ~ cut|clarity, data=diamonds)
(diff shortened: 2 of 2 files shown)


hadley:
Maybe `orient = c("v", "h")` would be better in `geom_linerange()`?  I think I'd prefer a parameter than more names.  It might be possible to use `flip_labels()` for a fairly dry implementation?

@lionel- would you be interested in working on again for the other vertical only geoms? This would also involve deprecating `geom_errorbarh`

lionel-:
it'd also be great to find a solution for dodging. Maybe the Geoms could have a `orient` field that Positions could query to do the right thing.

I may work on it again next week but could you be more explicit about what you didn't like in #1128 (besides the repeated code)?

hadley:
@lionel- just all the duplicated code. I think with an orient parameter and by using `flip_labels()`, there should be hardly any duplication.

I didn't think about dodging. The easiest approach would probably be an orient param to `position_dodge()`, and then that could get automatically passed on during layer creation, a la `geom_jitter()`. I'd prefer to not break encapsulation by allowing geoms & positions to communicate.

bbolker:
I wasn't aware of #1128, now that I see it it seems to completely supersede what I'm trying to do here (hurrah).  Is there any reason not to just close this PR (and if useful/necessary copy some of the relevant discussion back to the #1128 thread) ?

lionel-:
@bbolker I just uploaded #1265 which implements horizontality through an argument and handles position adjustments. If Hadley is ok with the implementation, I'll try to make the PR as complete as the old one.

bbolker:
bump.  Did this ever happen?  I would love to have that functionality ...

hadley:
I'm not super excited about this (just because it's something I'm unlikely to use) and since you can now create your own geoms outside of ggplot2, I think a package that provided a whole bunch of horizontal geom variants would be reasonable.

PR #1265, opened 2015-08-14
New approach for horizontal layers

lionel-:
So here is a second take at this problem.

I added a `flip` field to Layer objects which is set to `TRUE` whenever a Geom's orientation is not its natural one. Based on this switch we can modify automatically methods and functions in order to swap `x`s and `y`s parameters. This is accomplished with:

* A functional for functions taking a character vector of param names, or a data frame of params
* A factory that transforms layer methods

The factory is very DRY but does not work everywhere. The `draw()` method cannot be handled generically it seems. Therefore I also pass an orientation setting to `geom_params` to adjust the `draw()` methods inside the Geoms themselves.

With this approach, the Positions do not even need to be modified.

If you are ok with this proposal I'll add the other geoms/stats.


hadley:
This just feels a bit too general too me. It looks like code that I'll look at again in a couple of years time, and I'll have no idea how it works. How about something that's called explicitly in the draw functions of those geoms?

lionel-:
I think it needs to be more complicated than that.

For example the `required_aes` need to be flipped at some point before they are checked. The positions are designed to work vertically so the data they get also needs to be flipped, etc.

I don't think it's possible to handle everything only in the draw methods :/

lionel-:
> I don't think it's possible to handle everything only in the draw methods :/

Maybe one way to do it would be to force users to provide a vertical aesthetics specification and then have a `flip` argument. This would resemble how we currently use `coord_flip()` but I don't think this is an intuitive interface.

hadley:
Hmmm, what if it also happened during `reparameterise()`? I think that's called before aesthetics are checked, and conceptually it does somewhat make sense.

lionel-:
reparameterisation occurs after computation of statistics, so we'd need to adjust `compute()` functions as well.

Maybe flipping the data very early in plot-build.r if `Layer$flip` is `TRUE`?

hadley:
Hmmm, I'd really prefer not to add another parameter to layer. Let me keep thinking about this.

bbolker:
bump ... ?

hadley:
For now, I'd say put this in a separate package. I just don't have the brain space to develop a solution I'm happy with for this release.

lionel-:
too bad this `orient` parameter would have been more elegant than adding a whole bunch of new geoms :/

@bbolker I'll try to put together a package within the next month.

bbolker:
@lionel , depending on what's most convenient for you, consider the possibility of contributing it to `ggExtra`? https://github.com/daattali/ggExtra/issues/17

lionel-:
ugh, a gg package that doesn't use snake_case. More seriously, I think it makes sense to bundle all the `h` versions in their own package.

bbolker:
Just to let you know I'd [still be very interested in a horizontal-geom add-on package](http://stackoverflow.com/questions/34643393/horizontal-ggplot2geom-violin-without-coord-flip), if it's still on your to-do list ...

lionel-:
I'll try to get to it shortly :)

lionel-:
@bbolker Here it is: https://github.com/lionel-/gghorizon

If you have time, could you check the package with different combinations of geoms, stats and positions in horizontal layers?

bbolker:
well, geom_violinh works very nicely ... haven't had much more chance yet.

PR #1266, opened 2015-08-14
Rename dlapply() to by_layer()

lionel-:
Clearer function and argument names in plot-build.r.


hadley:
Thanks! Please let me know if you notice any other obvious wins like this.

PR #1272, opened 2015-08-18
fixed typo in error text of geom-path-.r: add "s" to "consist"

bbrewington:
Info shown below is from file ```geom-path-.r```, line 117

**Original error text:** geom_path: Each group consist of only one observation. Do you need to adjust the group aesthetic?
**Updated error text:** geom_path: Each group consist**s** of only one observation. Do you need to adjust the group aesthetic?


hadley:
Thanks!

PR #1278, opened 2015-08-20
Add examples for removing guides

joranE:
Adds some examples to ?guides for selectively suppressing guides.


R/guides-.r
```diff
@@ -29,6 +29,12 @@
 #'  scale_size_discrete(guide = "legend") +
 #'  scale_shape(guide = "legend")
 #'
+#'  # Remove some guides
+#'  p + guides(colour = "none")
```
hadley: I think it would be better to just use "none", and not use `FALSE`

hadley:
Thanks!

PR #1281, opened 2015-08-20
Allow color bar guides for alpha scales

joranE:
This may be a bad idea for reasons I haven't thought of, but someone asked about it recently and it seemed reasonable to me.

I'm a little uncertain that my reversing of `ticks$.label` is the best way to ensure the color bar and labels are aligned in the right direction, but it seemed to work in all the examples I tried.


hadley:
Can you give a couple of examples that show why you need to flip the labels?

joranE:
When I don't flip the labels and run this example:

    DF <- data.frame(
      x = rnorm(100, mean = 0, sd = 1),
      y = rnorm(n = 100, mean = 1, sd = 1),
      alpha = runif(n = 100, min = 0.1, max = 1)
    )

    ggplot(data = DF,aes(x = x, y = y, alpha = alpha)) +
      geom_point() + 
      guides(alpha = guide_colorbar())

The alpha color bar is labelled with 0.75 corresponding to the lighter end and 0.25 to the darker end, but the alpha values closer to one should be darker.

hadley:
I think the problem is actually at line 203:

```R
guide$bar <- data.frame(colour = scale$map(.bar), value = .bar, stringsAsFactors = FALSE)
```

It assumes that the scale produces a colour, but the alpha scale produces a number.  I'd rather you special cased that for alpha scales, rather than messing with the ticks.

hadley:
Intersti

PR #1303, opened 2015-08-28
Fix dulicate xmin, xmax calculation in collide()

has2k1:
1. Copy
2. Paste
3. Modify*

\* Previously incomplete step, now complete.


hadley:
Oops, thanks!

PR #1308, opened 2015-08-31
Make some ggproto objects immutable. Closes #1237

### LINKED ISSUE #1237: Immutable flag for ggproto objects
Would be useful to make it explicit that most ggproto objects (with the exception of ranges) are immutable. Could add extra class that has `$<-.ggimmutable` method defined to throw error

wch:
This makes some ggproto objects in ggplot2 immutable. The exceptions are `Range`, `Scale` and `ScalesList`, because these need to be mutable for ggplot2 to work.

In this implementation, the objects inherit from an `Immutable` ggproto object. It might be slightly more efficient to simply add an `"Immutable"` class attribute to each of the objects (as in `Geom <- ggproto(c("Geom", "Immutable"), ...), but that might be slightly confusing since the S3 "superclasses" would differ from the ggproto superclasses.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,6 +52,7 @@ Collate:
     'utilities.r'
     'aes.r'
     'legend-draw.r'
+    'immutable.r'
     'geom-.r'
     'annotation-custom.r'
     'annotation-logticks.r'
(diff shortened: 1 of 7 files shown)


hadley:
Doesn't this break the `update_geom_defaults()` function?

wch:
Yes. :( We'll need a workaround for that.

hadley:
Lets leave this for now.

PR #1315, opened 2015-09-01
Labels

lionel-:
This adds metadata to the labels: whether they are columns or rows labels, and whether they are generated for a wrap or a grid facet.

I use it to allow `label_bquote()` to take different expressions for rows and columns, and to make the interface of `labeller()` cleaner and more consistent.


stefanedwards:
Basically, your rewriting of `labeller` fixates the labelling to rows and
columns, and only one of each? And the result of a `labeller` can only be
used on facets on the same variables?

I.e., if I were to transpose the facetting, I would also have to swap the
named arguments for the labeller?

How does it handle facetting composites, i.e. `facet_grid(var1+var2 ~
var3)` ?

As it is now, I can define a "global" labeller for all plots.

(As you might guess, I devised the first labeller.)

Stefan McKinnon Høj-Edwards
ph.d. Genetics
+44 (0)776 231 2464
+45 2888 6598
Skype: stefan_edwards

2015-09-01 23:42 GMT+01:00 Lionel Henry <notifications@github.com>:

lionel-:
> How does it handle facetting composites, i.e. `facet_grid(var1+var2 ~ var3)` ?

I think it's cleaner now because you can't do `var1 = labeller1, var2 = labeller2`. You can only attribute one labeller for all composite variables. In this case you'd do `rows = labeller`.

> As it is now, I can define a "global" labeller for all plots.

I didn't think about that, this does sound useful.

So I could allow the key-labeller pairs in `...` on top of the `rows`/`cols` arguments (with some consistency checking). I could restore the original interface. Or I could leave it as it is. Let's hear what @hadley thinks.

stefanedwards:
My opinion on explicitly stating labels for rows and columns
[labeller=labeller(rows=foo,
cols=c(..))] is that it is a step back from the former, automagic
approach [labeller=labeller(var1=foo,
var2=c(..))] as it takes care of all three critiques in my first reply
(easily swap facet variables, use of global labeller, and composites).

How does it handle facetting composites, i.e. facet_grid(var1+var2 ~ var3) ?

I think it's cleaner now because you can't do var1 = labeller1, var2 =
> labeller2. You can only attribute one labeller for all composite
> variables. In this case you'd do rows = labeller.

Could you provide more context here, as I cannot see why this is more
cleaner? In the current setup, it can be done as

facet_grid(var1+var2 ~ var3, labeller=labeller(var1=foo, var2=bar,
var3=umf))


/Stefan


Stefan McKinnon Høj-Edwards
ph.d. Genetics
+44 (0)776 231 2464
+45 2888 6598
Skype: stefan_edwards

2015-09-02 9:26 GMT+01:00 Lionel Henry <notifications@github.com>:

lionel-:
ok, I see what you mean.

Actually the step back occurred in the previous pull request #1127. In the current state of things, you cannot use `labeller()` to attribute different labellers to composite facets, so it seemed more consistent to have the interface deal only with rows and columns.

I'll see what I can do to restore the old behaviour pre-1127.

lionel-:
@stefanedwards do you know what was the purpose of `keep.as.numeric`? It seems it'd be harmless to always coerce as characters. This would allow lookup tables to work with numeric values without having to coerce manually with this argument.

lionel-:
I can imagine that some non-labeller functions could require numeric data, but it's probably a rare occurrence that can be handled with a few more lines of code on the user side.

stefanedwards:
Sorry about your troubles I am causing you. I haven't paid much attention
to the continuous development on ggplot2.

The keep.as.numeric was for those rare cases when a facetting variable was
numeric; I believe I encountered the situation, but it aren't the variables
cast as factors?
Anyway, if the facet variable is numeric/integer, using an integer to fetch
an element from a vector with labels spells trouble. So the names of the
label-vector were coerced to characters, as well were the variable from the
facet so the label was found "by name" and not accidentally "by index".
Hope this makes sense.
The keep.as.numeric then allows the user to overrule this coercing.

/Stefan

Stefan McKinnon Høj-Edwards
ph.d. Genetics
+44 (0)776 231 2464
+45 2888 6598
Skype: stefan_edwards

2015-09-02 16:21 GMT+01:00 Lionel Henry <notifications@github.com>:

lionel-:
ok, it's probably best the interface simpler, always coerce to character and deprecate this argument.

The PR should be ready for review now.

stefanedwards:
I like the intent of the code, now allowing to use named variables. Your
'.default' is a great idea! Good one!
However, I do not have the setup to test the code, so I cannot comment on
whether it actually works.

Stefan McKinnon Høj-Edwards
ph.d. Genetics
+44 (0)776 231 2464
+45 2888 6598
Skype: stefan_edwards

2015-09-02 17:21 GMT+01:00 Lionel Henry <notifications@github.com>:

lionel-:
It'd be great if you could test it. It's easy with devtools:
```r
install.packages("devtools")
devtools::install_github("lionel-/ggplot2", ref = "labels")
```

stefanedwards:
It looks cool! Great work.

There is an unexpected behaviour when dealing with logicals. Building upon
the first example:

p1 <- ggplot(mtcars, aes(x = mpg, y = wt)) + geom_point()

p1 + facet_grid(vs + am ~ gear, labeller = labeller(vs = label_both, am =
label_value)) # Displays 1 or 0 for am.
mtcars$am < with(mtcars, ifelse(am == 0, FALSE, TRUE))

p1 + facet_grid(vs + am ~ gear, labeller = labeller(vs = label_both, am =
label_value)) # Displays 2 or 1 for am.

This is due to lines 305-310. I admit this is my legacy... It was due to
the case when passing a character vector for mapping variables, albeit I
had problems when mapping to logical vectors (as picking element FALSE
behaves quite differently than using a character or integer).


I've made come adjustments by commenting lines 306-308, and add the
following to documentation (inserted into line 413, ca):

#'  # When handling logical values, they are cast as character.

#'  mtcars$am2 <- with(mtcars, ifelse(am == 0, FALSE, TRUE))

#'  p1 <- ggplot(mtcars, aes(x = mpg, y = wt)) + geom_point()

#'  p1 + facet_grid(vs + am2 ~ gear,

#'   labeller = labeller(vs = label_both, am2 = label_value))

#'  tf_labels <- c('FALSE'='No', 'TRUE'='Yes')

#'  p1 + facet_grid(vs + am2 ~ gear,

#'   labeller = labeller(vs = label_both, am2 = tf_labels))

/Stefan

Stefan McKinnon Høj-Edwards
ph.d. Genetics
+44 (0)776 231 2464
+45 2888 6598
Skype: stefan_edwards

2015-09-03 10:51 GMT+01:00 Lionel Henry <notifications@github.com>:

lionel-:
hmm yes... let's clean this further and simply coerce to character, it'll be much more predictable than before.

No need to document this in the `@examples` field since this how default labels are handled too, e.g. in `label_value()`.

hadley:
Is this good to merge?

lionel-:
yup

hadley:
Thanks!

PR #1323, opened 2015-09-03
Fix multiple tiny issues in the position classes

has2k1:
*R/position-.r
   - Wrong call signature

*R/position-fill.r
   - Obsolete function `setup_defaults`, probably a refactoring slip up.
   - Colliding should be in the per panel, rather than per layer.

*R/position-jitter.r
   - Inconsistent call signature name

*R/position-nudge.R
   - Inconsistent call signature name

*R/position-stack.r
   - Make use of new `setup_data` API

### DIFF
--- a/R/position-.r
+++ b/R/position-.r
@@ -66,7 +66,7 @@ Position <- ggproto("Position",
     })
   },
 
-  compute_panel = function(self, data, scales) {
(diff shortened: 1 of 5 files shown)


hadley:
Thanks!!

PR #1327, opened 2015-09-08
Export ggname()

lionel-:
This is useful for creating ggproto objects in other packages.

I think it should be renamed to `set_ggname()`, I can push a commit with the new name if you want.
Also I'd find it clearer if the internal function `empty()` was named `is_empty()`, so we know it's a predicate and not a verb.


hadley:
I don't want to export it yet because I need to rethink how grobs get named. Will probably happen in next release.

PR #1328, opened 2015-09-08
Fixed missing ":" in two places in "extending" vignette

hrbrmstr:
`@inheritParams ggplot2::stat_identity` had a single `:` in two places. Both were fixed.


hadley:
Thanks!

PR #1336, opened 2015-09-11
Extending ggplot2 intro

lindbrook:


hadley:
Thanks! This doc is still pretty rough

PR #1357, opened 2015-09-26
geom_smoke, geom_smoke2, etc.

dkahle:
Here's an interesting idea for an alternative to a density plot.  It's several transparent density plots, of varying bandwidths, super-imposed to create a "smoking" effect.  Looking around for similar graphics, this appears similar to JMP's "shadowgram".  Pics below, they're in the examples.

You might see a better way than doing two geoms, just let me know.  geom_smoke just uses the stat to make a bunch of density estimates and then ..group..'s them.  geom_ribbon won't allow for aesthetics though, so I implemented geom_smoke2, which simply creates a bunch of density layers and so is very slow.  Thoughts?

The stat also likely needs some cleaning...

smoke-basic
smoke2-stacked
smoke2-fill


hadley:
Now that's it's straightforward to define new stats/geoms in other packages, I'd rather keep ggplot2 as streamlined as possible. ccing @rpruim because I know he's interested in this sort of display.

rpruim:
Interesting idea.  And +1 for the new version of ggplot2 that makes creating geoms and stats simpler.

dkahle:
Yea, that seems very reasonable.  Is there a preferred package (repo) for this sort of thing?  I.e., a package of geom's, stat's, etc. external to the base package that you/Winston don't have to manage?  Perusing the "gg" package names on cran, seems like all/most of them contain wrappers for functions that create graphics with existing geoms, not geoms/stats themselves.  That makes sense because in the past it was hard to do that kind of thing, but being as that's no longer the case, do you know of anyone who has started putting another extension together?

rpruim:
I don't know of any yet.  I'm likely going to write a geom/stat for creating ASH plots, but I've not decided where to house it.

Issue #1263, opened 2015-08-13
text orientation in facet_grid is reversed


briatte:
Result with 1.0.1:

![Imgur](http://i.imgur.com/UUNrP78.png)

Same code, result with 1.0.1.9003:

![Imgur](http://i.imgur.com/EWOMrsE.png)

The text orientation of the row facets ("House" and "Senate") is reversed.

The top examples of `facet_grid` will reproduce the issue:

```r
p <- ggplot(mtcars, aes(mpg, wt)) + geom_point()
# With one variable
p + facet_grid(. ~ cyl)
p + facet_grid(cyl ~ .)
```

The minor updates to the themes are great otherwise! I'm seeing smaller ticks (okay), and larger `strip_background` rectangles (very cool).

If I might use this issue to request something, I would also love more space between axis titles and axes. The example figures above use extra end lines to set the space to something more readable: `labs(x = "\nCongress", y = "Modularity\n")`.

baptiste:
Isn't it better though, to have the y-axis title and the facet titles in the same orientation?

polobo:
Not sure about "better" but having the bottom of the letters closer to the data causes the least amount of mental gymnastics for me personally.  Not particularly scientific but there it is.

I think the main reason is that in that orientation the word "House" reads top-to-bottom just like the mind expects while in the "aligned with y-axis title" orientation you have to read the word "House" bottom-to-top which is difficult.

BrianDiggs:
I agree with @polobo that it "feels easier" to read the facet labels on the right when they are going down (beginning of word at top), the opposite of the axis label on the left. However, $n=2$ is not data. Surely someone has done research on this, probably 30 years ago. 

I realize the original choice was probably arbitrary, but I don't seen a reason it should be changed without evidence that it is the right thing to do.

hadley:
Changing it was an accident, I'll flip it back next week. No need to discuss further.

PR #1372, opened 2015-10-15
Strip text rotated opposite dir to axis text (derived themes)

lionel-:
Actually not due to your refactor but to this old regression in #1263

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -198,7 +198,7 @@ theme_light <- function(base_size = 12, base_family = "") {
       ),
       strip.text.y      = element_text(
         colour = "white",
-        angle = 90,
+        angle = -90,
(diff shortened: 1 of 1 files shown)


hadley:
Oh hmmm. That will just break the second issue again

lionel-:
In https://github.com/hadley/ggplot2/commit/3a33d34a7a6046498100496d372ad38bab3a8951 you set `strip.text.y`'s angle back to -90 for `theme_grey()` but not `theme_light()` and `theme_dark()`.

lionel-:
To be clear, the labels were wrong when switched but actually also unswitched, because I was using `theme_light()` ;)

hadley:
Oooooooooh I see. Thanks!

PR #1376, opened 2015-10-16
Handle deprecated labellers

lionel-:


lionel-:
hrm, the test I added fails with R CMD CHECK but not in other contexts. I'm trying to figure out why.

lionel-:
I'm a bit too acquainted with that `%>%` operator. I type it without thinking and I don't see it's there when proofreading code ;)

Should be good now.

hadley:
Thanks!

PR #1414, opened 2015-11-24
Fix swapping of nrow & ncol for vertical facetting

chrismerkord:
The swap code incorrectly resulted in nrow and ncol being equal or both being NULL. Using dummy variables will prevent this.

More details about the problem can be seen in hadley/ggplot2#1279


chrismerkord:
Here is the bug I filed for this: hadley/ggplot2#1417

hadley:
Thanks! Would you mind also add a bullet point to NEWS.md?

Issue #1434, opened 2015-12-17
Improve ggplot error message


hadley:
Currently has "Mapping should be created with aes or aes_string" (should be `aes_()`)

PR #1436, opened 2015-12-17
Fixed 'aes/aes_string' error message

Mellence:
This is a fix to the error message contained in the issue tracker.


Mellence:
Ref #1434

PR #1438, opened 2015-12-21
ggplot2 version should be 2.0.0 not 1.1.0 in vignette?

PeteHaitch:
Doesn't look like there was ever a 1.1.0 release, so I think this should be 2.0.0.


hadley:
Thanks!

PR #1462, opened 2015-12-31
Correct mistakes in the document

codetrainee:
change 
@param t,b,r,l Dimensions of each margin. (To remember order, think trouble).
to
@param t,r,b,l Dimensions of each margin. (To remember order, think trouble).


hadley:
Thanks!

PR #1463, opened 2015-12-31
Make margin property of axis.text more explictly

codetrainee:
As axis.ticks.margin is deprecated, setting margins of axis.text is much implicit. The aim to amend the document is merely to explain the method of using this property more clearly.


hadley:
I think you missed the comment at the top of this file:

```
% Generated by roxygen2: do not edit by hand
% Please edit documentation in R/theme-elements.r
```

Issue #1448, opened 2015-12-23
No way to specify number of bins in stat_density_2d 


ecoRoland:
`stat_density2d` used to pass the ellipses to `stat_contour`. That way one could specify the number of bins.

Pre-v2.0.0:

    m <- ggplot(faithful, aes(x = eruptions, y = waiting)) +
      geom_point() +
      xlim(0.5, 6) +
      ylim(40, 110)
    m + geom_density2d(bins = 3)

 Since `stat_density_2d` doesn't do that anymore, I don't see a way to specify the number of bins. If I'm  not missing something that seems like a regression.

has2k1:
There is an inconsistency in the API. Use `n = 3`.

ecoRoland:
@has2k1 I don't follow. `n = 3` is passed to `kde2d` and does not specify the number of bins.

has2k1:
@ecoRoland, maybe I do not understand what you mean by `bins`, but `n` does partition the grid and that is equivalent to `bins`. In 2nd example from the bottom at http://docs.ggplot2.org/dev/geom_density_2d.html, that is `n=100` which is 100 bins.

MikeWise2718:
Can you replicate this http://stackoverflow.com/questions/19329318/how-to-correctly-interpret-ggplots-stat-density2d effect with the `n=4`  parameter? I couldn't...

has2k1:
Oh. Okay, you cannot do that anymore. The `2.0` clean up means parameters do not fall through any more. I'll submit a PR.

PR #1464, opened 2015-12-31
Fix statdensity2d

has2k1:
Resolves

-  Minor slip up, user supplied h is ignored.
-  #1448, This was a regression due to the `...` refactoring.

I think more refactoring is needed to lift all the `stat$compute_group` parameters into the user API `stat()` function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 2.0.0.9000
 
+## Bug fixes and minor improvements
+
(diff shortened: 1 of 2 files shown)


hadley:
Can you please cite yourself and the issue in the news bullets? i.e. `(#1448, @has2k1)`

has2k1:
I have amended the commit.

hadley:
Thanks!

PR #1478, opened 2016-01-08
Remove superfluous comma in function arguments

jschoeley:
The `element_text` function is called with a comma after its last argument. While this does not produce a bug, it is an irregularity better to be fixed.


hadley:
Thanks. fwiw I wouldn't usually bother with a news bullet for something this small.

PR #1503, opened 2016-01-17
fortify.R: prevent ugly output if model has multiple classes

helix123:
Prevent ugly output if model has multiple classes


hadley:
Thanks - I prefer a slightly different formulation.

Issue #1491, opened 2016-01-14
theme_classic Causes Strange Appearance of Facet Grid


DarioS:
The lack of x-axis and y-axis lines, except for the panels at the bottom and on the left, makes the plot difficult to interpret.

    theme_set(theme_classic())
    ggplot(mtcars, aes(mpg, wt)) + geom_point() + facet_grid(carb ~ cyl)

Perhaps it should be more like `theme_bw` ?

hadley:
I'd be happy to review a PR.

PR #1511, opened 2016-01-23
Panel Border Around Facets

DarioS:
Makes plot more interpretable when there are five or so factor levels. This would solve #1491.

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -275,5 +275,3 @@ theme_void <- function(base_size = 12, base_family = "") {
   )
 }
 
-, size = 0.125),
-      strip.background  = element_rect(fill = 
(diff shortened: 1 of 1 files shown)


R/theme-defaults.r
```diff
@@ -227,7 +227,7 @@ theme_minimal <- function(base_size = 12, base_family = "") {
 theme_classic <- function(base_size = 12, base_family = ""){
   theme_bw(base_size = base_size, base_family = base_family) %+replace%
     theme(
-      panel.border     = element_blank(),
+      panel.border     = element_rect(fill = "transparent"),
```
hadley: Can you please use `fill = NA` to be consistent?

R/theme-defaults.r
```diff
@@ -275,3 +275,5 @@ theme_void <- function(base_size = 12, base_family = "") {
   )
 }
 
+, size = 0.125),
```
hadley: Was this included by accident?

DarioS: Yes.

hadley:
This totally changes the non-facetted appearance, which I think is the most common use case.

DarioS:
It seems impossible to make the theme look nice for both use cases, without user customisation at runtime.

hadley:
In that case, I think its best to stick with current behaviour

PR #1514, opened 2016-01-25
Still label legends and facets in theme_void

jiho:
While theme_void is meant for drawings or plots in which the coordinates
are self describing (like maps for examples), subplots and aesthetics
still need to be described. The legend itself can be removed with
`guide = none` in the scale functions.


hadley:
Can you please add a bullet to NEWS?

jiho:
Done, and I've rebased onto the current master.

Issue #1489, opened 2016-01-14
theme_update(text) fails due to NULL values


oneillkza:
It's possible, using `theme()`, to set individual font elements inline with plotting, eg (adapted from ggplot documentation example):

```{r}
df <- data.frame(gp = factor(rep(letters[1:3], each = 10)), y = rnorm(30))
ggplot(df, aes(x = gp, y = y)) + 
    geom_point() +
    theme(text=element_text(size=16))
```

However, when using theme_update to try to set this globally, it complains about `NULL` values in the `text` element.

```{r}
theme_update(text=element_text(size=16))
ggplot(df, aes(x = gp, y = y)) + 
    geom_point()
```

results in

```
Error in FUN(X[[i]], ...) : 
  Theme element 'text' has NULL property: family, face, colour, hjust, vjust, angle, lineheight, margin, debug
```

I don't know if this is intended, but it seems strange that the behaviour of `theme()` and `update_theme()` is different, and it's frustrating that it isn't possible to set font size globally.

(This is with `ggplot2_2.0.0`.)

wmayner:
I'm having the same problem, though I'm updating the theme manually with 

```r
my.theme = theme_grey() + theme(text=element_text(family='Roboto', size=30))
```

As with OP, the same modification on a ``ggplot`` object works fine; it's only when trying to modify the default theme that this error is raised.

oneillkza:
Interestingly, as a work-around, the following *does* work:

```{r}
theme_set(theme_bw(base_size = 16))
ggplot(df, aes(x = gp, y = y)) + 
    geom_point()
```

hadley:
The base theme settings which are inherited by other settings need to be fully specified. Otherwise nothing gets inherited.

oneillkza:
I see. So the solution for actually updating a single element of an element of a theme is something like:

```{r}
theme_update(text=(theme_grey()+theme(text = element_text(family = "Times")))$text )
```

That's pretty ugly; I'd expect theme_update to pull the base theme settings from the theme being updated, rather than requiring them to all be specified. It's your package, though.

hadley:
@oneillkza there's probably a better way to do it than that, but I hardly ever use themes so I'm the wrong person to ask.

oneillkza:
Thanks Hadley, that's understandable.

Hmm ... looking at the actual definition for `theme_update()`, it seems like the issue is that it uses `%+replace%` instead of `+`. 

```{r}
theme_update <- function(...) {
[...]

PR #1519, opened 2016-01-25
Change behaviour of theme_update to be more like an update

oneillkza:
Per discussion in #1489, this changes `theme_update()` to use the `+` operator, so that unspecified attributes of the element being set are taken from the current theme, rather than `NULL`ing them out. I've added `theme_replace()` with the same behaviour as the current `theme_update()`, in case that behaviour is desired (though I'm hard-pressed to think of a use case).

I've also added some verbiage to the documentation clarifying what these do, along with examples.


hadley:
@wch do you remember the motivation for using `%+replace%` here?

R/theme.r
```diff
@@ -15,19 +21,40 @@
...
 theme_update <- function(...) {
   # Make a call to theme, then add to theme
+  theme_set(theme_get() + theme(...))
+}
+
+#' @rdname theme_update
+#' @export
+
+theme_replace <- function(...) {
+  # Make a call to theme, then add to theme
```
hadley: Can you please remove this comment? (and in the previous fun)

R/theme.r
```diff
@@ -15,19 +21,40 @@
...
 #'  theme(legend.background = element_rect(fill = "white", colour = "white", size = 3))
+#'
 theme_update <- function(...) {
   # Make a call to theme, then add to theme
+  theme_set(theme_get() + theme(...))
+}
+
+#' @rdname theme_update
+#' @export
+
```
hadley: Please remove this empty line

hadley:
Looks good. Just a few minor tweaks and can you please add a bullet point to NEWS?

oneillkza:
Done!

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # ggplot2 2.0.0.9000
 
+* `theme_update()` now uses the `+` operator instead of `%+replace%`, so that
+  unspecified values will no longer be `NULL`ed out. `theme_replace()`
+  preserves the old behaviour if desired (#1519).
```
hadley: Could you please cite your github user name in the brackets?

oneillkza:
Done!

hadley:
Thanks!

PR #1527, opened 2016-01-28
Function as data

thomasp85:
This is a follow up on PR #1486

Summary of changes:

- Add `fortify.function` that returns function unchanged
- Add `layer_data` method to Layer base class that takes the plot data as input and returns the data associated with the layer
- Remove the need for waiver-check in `map_layout`
- Add unit tests for `layer_data` method

Possible additions:
- Documentation of function-as-data functionality somewhere (don't really know the best location for this - any idea?)


R/layer.r
```diff
@@ -101,6 +101,19 @@ Layer <- ggproto("Layer", NULL,
...
+  layer_data = function(self, plot_data) {
+    if (is.waive(self$data)) {
+      return(plot_data)
+    } else if (is.function(self$data)) {
+      data <- self$data(plot_data)
+      if (!is.data.frame(data)) {
+        stop("Data function must return a data.frame")
+      }
+      return(data)
+    }
```
hadley: I think this would be better with no returns and a final else block

R/layer.r
```diff
@@ -101,6 +101,19 @@ Layer <- ggproto("Layer", NULL,
     cat(snakeize(class(self$position)[[1]]), "\n")
   },
 
+  layer_data = function(self, plot_data) {
+    if (is.waive(self$data)) {
+      return(plot_data)
+    } else if (is.function(self$data)) {
+      data <- self$data(plot_data)
+      if (!is.data.frame(data)) {
+        stop("Data function must return a data.frame")
```
hadley: This needs `call. = FALSE`

tests/testthat/test-layer.r
```diff
@@ -32,3 +32,16 @@ test_that("strip_dots remove dots around calculated aesthetics", {
...
+
+test_that("layer_data returns a data.frame", {
+  l <- geom_point()
+  expect_equal(l$layer_data(mtcars), mtcars)
+  l <- geom_point(data = head(mtcars))
+  expect_equal(l$layer_data(mtcars), head(mtcars))
+  l <- geom_point(data = head)
+  expect_equal(l$layer_data(mtcars), head(mtcars))
+  l <- geom_point(data = nrow)
+  expect_error(l$layer_data(mtcars))
```
hadley: Can you please check for the specific error message?

hadley:
This is looking good.  A few points:

* Can you please add a bullet point to NEWS?
* Probably best to document this in `layer()`

thomasp85:
All points have been addressed...

R/layer.r
```diff
@@ -101,6 +128,20 @@ Layer <- ggproto("Layer", NULL,
     cat(snakeize(class(self$position)[[1]]), "\n")
   },
 
+  layer_data = function(self, plot_data) {
+    if (is.waive(self$data)) {
+      data <- plot_data
```
hadley: You can drop the `data <-` in all the list cases

hadley:
One other idea now that I've merged it - what if the fortify happened in `layer_data()` as well?

thomasp85:
Possible error because of unknown data type will first happen when plotting - don't know if that is desirable?

thomasp85:
Should I redo the layer_data or is it merged and accepted and it was just a point for the future?

hadley:
Just a point for the future. I fixed locally

Issue #1548, opened 2016-02-16
annotate() does not support expressions in label argument


dgromer:
This fails:
```r
ggplot(data.frame(x = 1, y = 1), aes(x, y)) +
  geom_point() +
  annotate("text", x = 1, y = 1.1,
           label = expression(paste(Delta, italic("i"))), parse = TRUE)
```

This works:
```r
ggplot(data.frame(x = 1, y = 1), aes(x, y)) +
  geom_point() +
  annotate("text", x = 1, y = 1.1, label = "Delta~italic('i')", parse = TRUE)
```

However, in `labs`, objects of type `expression` don't throw an error.
```r
ggplot(data.frame(x = 1, y = 1), aes(x, y)) +
  geom_point() +
  labs(x = expression(paste(Delta, italic("i"))))
```

hadley:
Yes, annotate is built on top of `geom_text()`

dgromer:
So is there any way to make annotations by passing an expression? Because there are is stuff that cannot be expressed with a string in plotmath syntax (small example: `"Delta~italic('i')"` has a whitespace in between, whereas with `expression(paste(Delta, italic("i")))` it is possible to omit it)

hadley:
Put the paste in the string

dgromer:
Ah yeah, thanks a lot!

PR #1553, opened 2016-02-21
Examples for plotmath expressions in `annotate`

dgromer:
Adding two examples to the documentation of `annotate` about how to pass plotmath expressions.

1. simple plotmath expression
1. mixing plotmath syntax and plain text using `paste`

See #1548.

### DIFF
--- a/R/annotation.r
+++ b/R/annotation.r
@@ -29,6 +29,11 @@
 #'   colour = "red", size = 1.5)
 #'
 #' p + annotate("text", x = 2:3, y = 20:21, label = c("my label", "label 2"))
+#'
+#' p + annotate("text", x = 4, y = 25, label = "italic(R) ^ 2 == 0.75",
+#'   parse = TRUE)
+#' p + annotate("text", x = 4, y = 25,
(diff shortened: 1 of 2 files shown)


R/annotation.r
```diff
@@ -29,6 +29,11 @@
 #'   colour = "red", size = 1.5)
 #'
 #' p + annotate("text", x = 2:3, y = 20:21, label = c("my label", "label 2"))
+#'
+#' p + annotate("text", x = 4, y = 25, label = "italic(R) ^ 2 == 0.75",
+#'   parse = TRUE)
+#' p + annotate("text", x = 4, y = 25,
```
hadley: What does this variation do?

dgromer: It shows, that plain text can be mixed with plotmath expressions.

Using plotmath alone, it would not be possible to trim the leading zero of the r-squared statistic. In Psychology (according to the publication manual of the American Psychological Association) values that can't exceed 1 are typically printed without leading zero.

hadley:
Got it - thanks!

PR #1560, opened 2016-02-26
adding pad param to stat_count

bhive01:
can't compile ggplot2 book or run stat_count examples with freqpoly
```r
 ggplot(diamonds, aes(color, colour = cut)) + 
    geom_freqpoly(aes(group = cut), stat="count")
#Error: Unknown parameters: pad
```

Code added from stat_bin mostly. Unsure how it is incorporated into the calculations in lines 56-71 like in stat_bin. Tested from my branch and is working.

### DIFF
--- a/R/stat-count.r
+++ b/R/stat-count.r
@@ -18,6 +18,7 @@ stat_count <- function(mapping = NULL, data = NULL,
                        geom = "bar", position = "stack",
                        ...,
                        width = NULL,
+                       pad = FALSE,
                        na.rm = FALSE,
                        show.legend = NA,
                        inherit.aes = TRUE) {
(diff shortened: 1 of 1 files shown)


bhive01:
Appears to be an issue from Travis about the lack of the "pad" parameter in the docs for geom_bar.

bhive01:
I tried adding pad=FALSE in geom_bar, but it ended up causing issues with geom_bar that I had with geom_freqpoly:
``` r
ggplot(mpg, aes(x=class)) + geom_bar()
#Error: Unknown parameters: pad
```

I'm sorry, I don't understand the code enough to make this work. I hope this is helpful enough to get you there. :|

PR #1566, opened 2016-03-03
Use "boundary" instead of deprecated "origin" in examples

yutannihilation:
Examples in `geom_histogram()` still use deprecated `origin`. This PR updates the document.


hadley:
Thanks!

PR #1575, opened 2016-03-08
guide_colourbar now respects show.legend

### LINKED ISSUE #1568: Omit legend using `show.legend = FALSE` does not work when aesthetic is continuous
`show.legend = FALSE` works as expected when the fill variable is discrete. However, when the `aes`thetic is mapped to a continuous variable, `show.legend = FALSE` does not drop the legend.

    df <- data.frame(x = 1:3, y = 1)

Discrete `fill` or `color` `aes`:
    
    ggplot(df, aes(x = x, y = y, color = factor(x))) +
      geom_point(size = 20, shape = 21, show.legend = FALSE)
    
    ggplot(df, aes(x = x, y = y, fill = factor(x))) +
      geom_point(size = 20, shape = 21, show.legend = FALSE)

Continuous `fill` or `color` `aes`:

    ggplot(df, aes(x = x, y = y, color = x)) +
      geom_point(size = 20, shape = 21, show.legend = FALSE)

    ggplot(df, aes(x = x, y = y, fill = x)) +
      geom_point(size = 20, shape = 21, show.legend = FALSE)

`ggplot2_2.1.0`

has2k1:
Previously, a colourbar was always drawn whenever there was a
colour/fill mapping to a continuous variable.

Solution is to kill the guide if any has been created but if
no layer uses it.

*Addendum*
There are 3 places in the code where it is decided whether a legend will show up
or not. A better solution would be to fix that. At the moment, this (which added the 3rd)
is the least destructive.

fixes #1568

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,10 @@
 * `x` and `y` scales are now symmetric regarding the list of
   aesthetics they accept: `xmin_final`, `xmax_final`, `xlower`,
   `xmiddle` and `xupper` are now valid `x` aesthetics.
+  
+* `guide_colourbar` now respects the layer parameter `show.legend`.
+  When `show.legend` is set to `FALSE` a colourbar will not be
+  shown. (@has2k1)
 
 # ggplot2 2.1.0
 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks, but I'd rather fix this more generally.

PR #1578, opened 2016-03-09
Show a list of shapes in scale_shape docs

steveharoz:
I often wind up hunting for the indices corresponding to different shapes, and it'd be nice if they were in the docs.

### DIFF
--- a/R/scale-shape.r
+++ b/R/scale-shape.r
@@ -26,11 +26,11 @@
 #' d %+% dsmall
 #'
 #' # Show a list of available shapes
-#' df_shapes <- data.frame(shape = factor(0:24))
-#' ggplot(df_shapes, aes(shape=shape)) +
-#'   geom_point(aes(shape=shape, x=0, y=0), size=5, fill="red") +
-#'   facet_wrap(~shape) +
-#'   scale_shape_manual(values=df_shapes$shape, guide="none")
+#' df_shapes <- data.frame(shape = 0:24)
+#' ggplot(df_shapes, aes(shape = shape)) +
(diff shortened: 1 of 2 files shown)


R/scale-shape.r
```diff
@@ -24,6 +24,13 @@
 #'
 #' # Or for short:
 #' d %+% dsmall
+#'
+#' # Show a list of available shapes
+#' df_shapes <- data.frame(shape = factor(0:24))
+#' ggplot(df_shapes, aes(shape=shape)) +
+#'   geom_point(aes(shape=shape, x=0, y=0), size=5, fill="red") +
```
hadley: Can you please use spaces around `=`?

R/scale-shape.r
```diff
@@ -24,6 +24,13 @@
...
+#'   facet_wrap(~shape) +
+#'   scale_shape_manual(values=df_shapes$shape, guide="none")
```
hadley: I think `scale_shape_identity()` would be simpler

hadley:
This is the code I use in the vignette:

```R
    shapes <- data.frame(
      shape = c(0:19, 22, 21, 24, 23, 20),
      x = 0:24 %/% 5,
      y = -(0:24 %% 5)
    )
    ggplot(shapes, aes(x, y)) + 
      geom_point(aes(shape = shape), size = 5, fill = "red") +
      geom_text(aes(label = shape), hjust = 0, nudge_x = 0.15) +
      scale_shape_identity() +
      expand_limits(x = 4.1) +
      scale_x_continuous(NULL, breaks = NULL) + 
      scale_y_continuous(NULL, breaks = NULL)
```

Probably a bit too complicated to include in an example though

hadley:
I like it! This is more elegant than my approach

Issue #1565, opened 2016-03-02
theme_classic() no axis lines in version 2.1


jscamac:
Shouldn't theme_classic actually have axis lines?
Is the update a mistake?

See theme_classic():
http://docs.ggplot2.org/dev/ggtheme.html

Katiedaisey:
It seems my update to allow access to lower level elements (axis.line.x, panel.grid.major.y, etc) leaves changes to upper level elements unused.  Let me give this some thought.

steveharoz:
Here's a temporary workaround for the axes if anyone needs it until the issue's resolved:
    theme_classic() + theme(
        axis.line.x = element_line(colour = 'black', size=0.5, linetype='solid'),
        axis.line.y = element_line(colour = 'black', size=0.5, linetype='solid'))

jscamac:
Thanks. I already had the work around but thought I should let you know of the changes to the classic_theme.

ustervbo:
I think the problem is related to https://github.com/hadley/ggplot2/issues/1567 and seems to stem from  complete removal of axis lines by default

Katiedaisey:
It has to do with how attributes for elements are inherited.  To remove an attribute, like axis.line in some themes, the current code sets them as element_blank(). However if a subelement, axis.line.x and axis.line.y are added, they inherit element_blank and will not appear. Newly added elements need to inherit from the original object (line, rect, or text) skipping any blank elements but I'm still thinking on how to do that.

bklingen:
Really would appreciate for theme_classic() to work as in version 2.0.0 (and all prior versions), showing the x and y axis line as before instead of drawing no axis at all. (And thanks for the, hopefully temporary, workaround). I have countless shiny applications that depend on the classic theme, and re-programming all of these would be a nightmare.

Katiedaisey:
I think I have this completely fixed.  Would anyone mind testing before I request a pull?

steveharoz:
I pulled from your fork and tried running the example in http://docs.ggplot2.org/current/ggtheme.html
The ticks and labels look messed up, but I'm not sure if it's related.

```
ggplot(mtcars) + 
  geom_point(aes(x = wt, y = mpg, colour=factor(gear))) + 
  facet_wrap(~am) +
  theme_classic()
```
![image](https://cloud.githubusercontent.com/assets/2257540/13652944/9d1c5d2a-e614-11e5-9c92-4a53d43a7a89.png)

Nevertheless, axis lines appear to work:
[...]

PR #1557, opened 2016-02-24
Updates theme-defaults.r (#1555)

Katiedaisey:
addresses issue #1555 where some elements added under certain themes inherit "element_blank" and cannot be updated, e.g.

`ggplot(mtcars, aes(wt, mpg)) + geom_point() + theme_classic() + theme(panel.grid.major.y = element_line(colour = "pink", size = 4))`

update explicitly declares parent element as necessary and children elements as element_blank

Issue #1555, opened 2016-02-22
Panel Grid Aesthetics Can't be Changed for theme_classic


DarioS:
It has no effect for `theme_classic`

    ggplot(mtcars, aes(wt, mpg)) + geom_point() + theme_classic() + theme(panel.grid.major.y = element_line(colour = "pink", size = 4))

but it does work for the other themes

    ggplot(mtcars, aes(wt, mpg)) + geom_point() + theme_bw() + theme(panel.grid.major.y = element_line(colour = "pink", size = 4))

It also doesn't work for the grid lines perpendicular to the x-axis.

Katiedaisey:
Since panel.grid.major.y inherits from panel.grid.major while the later has an element_blank class, it will not inherit the desired element_line. This can easily be confirmed via

`calc_element('panel.grid.major.y', theme_classic() + theme(panel.grid.major.y = element_line(colour = "pink", size = 4)), verbose = T)` 

This will happen whenever a new element inherits from a element that was declared blank in the base theme. For instance, adding tick marks only on x with theme_minimal also fails:

`ggplot(mtcars, aes(wt, mpg)) + geom_point() + theme_minimal() + theme(axis.tick.x = element_line(colour = "pink", size = 4))`

It seems these might actually not work in only three cases in which case I would suggest just updating the two themes, classic and minimal, by declaring the problem children.

    theme_classic2 <- function(base_size = 12, base_family = ""){
      theme_bw(base_size = base_size, base_family = base_family) %+replace%
        theme(
          panel.border     = element_blank(),
          axis.line        = element_line(colour = "black"),
          panel.grid.major = element_line(),
          panel.grid.major.x = element_blank(),
          panel.grid.major.y = element_blank(),
          panel.grid.minor = element_blank(),
          panel.grid.minor.x = element_blank(),
          panel.grid.minor.y = element_blank(),
          strip.background = element_rect(colour = "black", size = 0.5),
          legend.key       = element_blank()
        )
    }

    theme_minimal2 <- function(base_size = 12, base_family = "") {
      # Starts with theme_bw and then modify some parts
      theme_bw(base_size = base_size, base_family = base_family) %+replace%
        theme(
          legend.background = element_blank(),
          legend.key        = element_blank(),
          panel.background  = element_blank(),
          panel.border      = element_blank(),
          strip.background  = element_blank(),
          plot.background   = element_blank(),
          axis.ticks        = element_line(),
          axis.ticks.x     = element_blank(),
[...]

PR #1581, opened 2016-03-10
Inheritance of elements

### LINKED ISSUE #1567: `axis.line.x` and `axis.line.y` work, but not `axis.line`
In 2.1.0, `axis.line.x` and `axis.line.y` work, but `axis.line` does not.

For example, 

This works:
```r
  geom_point() +
  theme(axis.line.x=element_line(),
        axis.line.y=element_line())
```

image

But this doesn't:
```r
  geom_point() +
  theme(axis.line=element_line())
```

image

(The example at http://docs.ggplot2.org/current/theme.html - i.e., `m + theme(axis.line = element_line(size = 3, colour = "red", linetype = "dotted"))` - also fails)

Katiedaisey:
Fixes  #1567

addresses issues #1555, #1557, #1565, #1567, inheritance in elements

previously subelements such as `panel.grid.major.x` were unable to be changed when inheriting from an `element_blank()` element such as  `panel.grid.major`. Update #1557 addressed this, but left the higher element unable to be changed.

`p <- ggplot(mtcars) + geom_point(aes(x = wt, y = mpg, colour=factor(gear))) + theme(axis.line = element_line(colour = "pink", size = 3))`
rplot01

This update allows access to both subelements and higher elements:

`p + theme_classic() + theme(panel.grid.major.x = element_line(colour = "pink", size = rel(4)))`
rplot02

`p + theme_classic() + theme(panel.grid.major = element_line(colour = "pink", size = rel(4)))`
rplot03

and addresses the loss of axes in some themes.

The inheritance scheme necessary for this functionally causes this update to fail one test, however graphical output is as expected.


Katiedaisey:
If all inherited elements are blank, the final subelement will be reassign element_blank. If any inherited elements are not blank, all subelements will retain original element_text, element_line, or element_rect and thus retain the ability to be set independently.

hadley:
I'd like to hold off on this until we have the new visual testing setup complete so that we can ensure we don't introduce any regressions.

hadley:
@Katiedaisey thanks for your work on this! We've decided to go a slightly different route though (as you've seen)

PR #1590, opened 2016-03-23
New geom_(): geom_col()

### LINKED ISSUE #1685: New geom for geom_bar(stat = "identity")
But I don't know what to call it

hrbrmstr:
Fixes #1685 

* Added `geom_col()` which provides an alternative to `geom_bar()` that uses `stat="identity"` by default, requiring the specification of a `y` aesthetic. It also does not have a `binwidth` parameter (the benefits of starting fresh). 
* Updated references & examples in other functions where `geom_bar(stat="identity"…)` was used. Some were kept in since `geom_bar()` still supports `stat_identity()`.
* Substituted geom_col() for geom_bar() in one test to ensure it had test coverage outside the examples
* Updated `NEWS.md`
* Ran `devtools::test()` (all completed successfully)
* Ran `devtools::check()` (no notes or warnings)

I think I may have skimped on the examples in `geom_col()` and will be glad to add ones if deemed necessary. I can pull back any of the other mods as well. I made a choice to have this be a separate function vs have it documented under `geom_bar()` since I think there are enough differences (esp the mandatory `y` aesthetic) to keep it standalone (and it might confuse users to add it under `geom_bar()`).


hadley:
This looks great! 

My one small question is with the default position adjustment - you get a warning if you attempt to display a negative height (which is totally reasonable in this context). Maybe this should be "identity"?  I'm not sure what the right principle to apply is.

hadley:
Actually, let's fix the stacking problem separately: #1691

hadley:
Thanks @hrbrmstr!

PR #1592, opened 2016-03-23
Let Stats and Geoms know about specification of aesthetics and faceting

lionel-:
I am writing a ggplot2 Stat that integrates fitted model into the grammar of graphics. Since it has to link the relevant grouping variables to the model predictors, the Stat requires knowledge of how all relevant aesthetics are specified and what faceting variables are active. For example it needs to know that `colour` has been specified with `factor(cyl)`. I could get this information by overwriting some Layer's method, but it would be much cleaner if that was part of the API.

So this PR proposes to add a `spec` argument to `setup_params()`. This has the downside of breaking the API but on the other hand adding a whole new method just for that does not seem appropriate. I ran a revdep check and, if I didn't miss anything, only four packages are broken by this change. Two of them because they have some custom version of `ggplot_build()` and the other two because they have overwritten `setup_params()` for a custom Stat. All those cases are trivial to fix and if this PR is merged I will send these projects a PR.

Also if Stats get to see the plot specification, we should probably do the same for Geoms. I added a similar `setup_params()` method to Geoms with `data`, `params` and `spec` arguments.

The `spec` object is defined as follows:

```{r}
spec <- function(layer, plot) {
  list(
    facet_spec = facet_spec(plot$facet),
    plot_mapping = plot$mapping,
    layer_mapping = layer$mapping
  )
}

facet_spec <- function(facet) {
  UseMethod("facet_spec")
}

facet_spec.wrap <- function(facet) {
  spec <- list(facets = facet$facets)
  structure(spec, class = c("wrap", "spec"))
}

facet_spec.grid <- function(facet) {
  spec <- list(rows = facet$rows, cols = facet$cols)
  structure(spec, class = c("grid", "spec"))
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,7 +19,8 @@
 
 * `Stat`'s `setup_params()` method now takes a `spec` argument that
   contains information about the specification of aesthetics and
-  facetting. See `Stat`'s documentation. (@lionel-)
+  facetting. See `Stat`'s documentation. `Geom` gains a similar
+  `setup_params()` method. (@lionel-)
 
 # ggplot2 2.1.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,10 @@
 * Minor code formatting issues in examples and function parameters were
   fixed. (@hrbrmstr)
 
+* `Stat`'s `setup_params()` method now takes a `spec` argument that
+  contains information about the specification of aesthetics and
+  facetting. See `Stat`'s documentation. (@lionel-)
+
 # ggplot2 2.1.0
 
 ## New features
--- a/R/geom-.r
+++ b/R/geom-.r
@@ -13,11 +13,11 @@ NULL
 #' implement one or more of the following:
 #'
 #' Compared to \code{Stat} and \code{Position}, \code{Geom} is a little
-#' different because the execution of the setup and compute functions is
-#' split up. \code{setup_data} runs before position adjustments, and
-#' \code{draw_layer} is not run until render time,  much later. This
-#' means there is no \code{setup_params} because it's hard to communicate
-#' the changes.
+#' different because the execution of the setup and compute functions is split
+#' up. \code{setup_params} and \code{setup_data} run before position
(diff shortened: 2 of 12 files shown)


has2k1:
Why not just let the `stat` (and `geom` for that matter) be made aware of the `plot` and `layer` without encapsulating them into a `spec`.

At the end of  the `layer` function in `layer.r`

```r
new_layer <- ggproto("LayerInstance", Layer, ...)
stat$layer <- new_layer
geom$layer <- new_layer
new_layer
```
And also have each `layer` get the plot object attached to it. This could happen when the layer is added to the plot or near the top of `plot_build`.  Already, some `layer` functions need access to `plot` so it gets passed as a parameter, yet it could be available at `self$plot`. How these modifications are done could require changes to the `plot_clone` function to prevent infinite cloning.

With this approach, any other novel `geoms` or `stats` would not require an extension of the `spec` as all the plot information would be available to them.

lionel-:
I now think we shouldn't break compatibility like I did in this PR.

@has2k1 This is good, but I'd add a link to the plot in the Layer rather than in the Stats and Geoms. So you'd access the plot with `self$Layer$plot`.

Another way would be to add a `setup_context(data, params, layer, plot)` method. This is a bit redundant with `setup_params()` but it would not break compatibility. 

In both cases we can still provide a helper function to get the spec from the plot in a structured way.

lionel-:
In the meantime here is a hack to add plot and Layer in the stat_params of a layer:

```r
add_metadata <- function(layer) {
  compute_aesthetics_orig <- layer$compute_aesthetics

  compute_aesthetics_hack <- function(self, data, plot) {
    # Unclassing plot to prevent recursive plot building on print
    metadata <- list(layer = layer, plot = unclass(plot))
    layer$stat_params <- c(layer$stat_params, metadata)
    compute_aesthetics_orig(data, plot)
  }
  layer$compute_aesthetics <- compute_aesthetics_hack

  layer
}
```

hadley:
I'm going to close this for now - can you please open an issue for discussion instead?

PR #1598, opened 2016-03-28
Added new economic theme

JestonBlu:
Created a new theme for classical economic plotting. Its a simplified theme with only the traditional x and y axis outlined. The axis titles are slightly larger and the grid lines have been removed. The idea is to create a basic plot similar to what Economics instructors would manually draw in class.

image


hadley:
Thanks for the contribution, but unfortunately I'm not accepting new themes - I think it's better that they go in new packages now. Maybe the author of ggthemes might be interested?

PR #1621, opened 2016-04-23
Typo in error message while saving large plots

shreyasgm:


hadley:
Thanks! (but someone else got to it before you)

PR #1622, opened 2016-04-26
Fix geom_dotplot y limit calculations

### LINKED ISSUE #1618: geom_dotplot negates free y scale from facet_wrap
When binning the dotplot on the y axis, the `facet_wrap` free y scale is negated.  This is new behavior - is it by design?
```
library(ggplot2)

p1 = ggplot(mtcars, aes(factor(cyl), mpg)) +
	geom_boxplot() +
	facet_wrap(~am, scales = "free_y")
p1
```
rplot
`p1 + geom_dotplot(binaxis = "y")`

rplot01

has2k1:
The computed y limits depend on both the group & panel.
This ensures that the scale trainning does not get wrong
(global y limits) input.

Fixes #1618

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -75,6 +75,9 @@
 * Only one warning is issued when asking for too many levels in 
   `scale_discrete()` (#1674)
 
+* Fixed problem with `geom_dotplot()` when facetting and binning on the
+  y-axis. (#1618, @has2k1)
+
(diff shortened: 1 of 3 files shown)


hadley:
Do you think you could add a unit test too?

hadley:
Bump

has2k1:
Added tests.

PR #1623, opened 2016-04-28
corrected theme inhertiance for legend.key.size

coolbutuseless:
legend.key.size was marked as its own parent. fixed.


hadley:
Thanks!

PR #1624, opened 2016-04-29
Preserve class when adding uneval objects

wch:
This fixes a bug that is the root cause of rstudio/shiny#1178. When adding an `aes()` object to a `ggplot()` object, the `uneval` class is inadvertently dropped. For example:

```R
# When including aes() in the ggplot() call, class is preserved
p <- ggplot(mtcars, aes(wt, mpg))
class(p$mapping)
# [1] "uneval"


# When adding aes() in the ggplot() call, class is lost
p <- ggplot(mtcars) + aes(wt, mpg)
class(p$mapping)
# [1] "list"
```

This PR preserves the class of the object.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,8 @@
 * Minor code formatting issues in examples and function parameters were
   fixed. (@hrbrmstr)
 
+* Class of aesthetic mapping is preserved when adding `aes()` objects. (#1624)
+
 # ggplot2 2.1.0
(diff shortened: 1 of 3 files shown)


hadley:
Could you please add a unit test?

hadley:
And a bullet to NEWS.md please?

hadley:
Thanks!

PR #1625, opened 2016-05-04
Add top margin to default plot.caption

janschulz:
According to an informal survey (@hrbrmstr @juliasilge), this margin is needed every time a plot.caption is used:
https://twitter.com/janschulz/status/727932334218039296

The chosen top margin is the same as for the subtitle, just on the top instead
of the bottom.


hadley:
Thanks!

Issue #1596, opened 2016-03-24
Allow expressions in facet formulas


dgrtwo:
I'd like to be able to put expressions in faceted formulae instead of just column names.

For example, I'd like to be able to facet by a "cut" of a column:

```r
ggplot(mtcars, aes(qsec, mpg)) +
  geom_point() +
  facet_wrap(~ cut(wt, c(0, 2, 4, 6)))
```

rather than needing to do something like:

```r
mtcars$cut <- cut(mtcars$wt, c(0, 2, 4, 6))

ggplot(mtcars, aes(qsec, mpg)) +
  geom_point() +
  facet_wrap(~ cut)
```

DanRuderman:
Great feature. Syntax-wise, how would you handle the overloading of `+` as both an arithmetic operation and a separator in the formula?

hadley:
It would have to be handled the same way as modelling formulas — if you want addition you need to surround in `I()`

DanRuderman:
First attempt in #1722. Will be shelved until some infrastructure changes take place.

thomasp85:
#1722 has now been ported to the new facet implementation in #1633

Issue #1607, opened 2016-04-08
X axis doesn't appear below facet_wrap plot with uneven number of facets


paul4forest:
When drawing a plot with facet_wrap. If there is an uneven number of facets, the x axis is not drawn for facets which do not touch the "floor" of the plotting area. For example 

    ggplot(airquality, aes(x = Day, y = Temp)) +
        facet_wrap(~Month) + 
        geom_line()

![airquality_facet_wrap](https://cloud.githubusercontent.com/assets/5647397/14382482/6cfef90e-fd8e-11e5-98ea-be37fbe4149c.png)

It's possible to add an axis under all plots by setting the scales argument to "free_x". 

    ggplot(airquality, aes(x = Day, y = Temp)) +
        facet_wrap(~Month, scales = "free_x") + 
        geom_line()

![airquatlity_facet_wrap_free_x](https://cloud.githubusercontent.com/assets/5647397/14382542/da62460e-fd8e-11e5-8aa3-6379a7157fac.png)

But it's not possible to add an x axis only for the last dangling plot alone. This axis could be located in two places, right below the dangling plot: 

![airquality_facet_wrap_dangling_above](https://cloud.githubusercontent.com/assets/5647397/14382827/06edbe18-fd91-11e5-9de8-d95a02dbf388.png)

or at the bottom of the plotting area:
 
![airquality_facet_wrap_dangling_below](https://cloud.githubusercontent.com/assets/5647397/14382832/0dbe0f36-fd91-11e5-9377-f01093405ea3.png)

A workaround was given in [this Stackoverflow answer](http://stackoverflow.com/a/13316126/2641825). The author said a suggestion could be submitted here. 2 years after those comments were made on Stackoverflow, no suggestion has been submitted here on github. Considering the number of votes on the workaround answer, the issue still attracts interest. I thought I'd request your feedback. Could this feature be implemented, with minimal change to ggplot2, or is this out of the question?

PR #1633, opened 2016-05-13
facets to ggproto

thomasp85:
@hadley 

This PR is currently empty, but opened now to make a venue for discussing implementation etc.

My goal, besides streamlining ggplot2 OOP usage, is to make facets extensible by third parties in the same way as scales, geoms, etc. are now. For this to work the facet class need to be quite powerful in terms of which responsibilities it has, so I want to check with you before I begin working on it.

As I see it the facet class has the following responsibilities:

1. Get data and return modified data with PANEL column
2. Potentially set range of position scales
3. Get a gList of layers for each panel and lay them out in a gtable
4. Decorate around the facets (both axes, facet headings, etc)

In the end this means that the facet class takes full responsibility of all drawing within the plotting area (title, legends, margins, subtitles, etc should be handled elsewhere)

Mind you, this is before I’ve looked deeply into how facets are currently implemented. There might be additional responsibilities that appears…

How do you feel about the above?


hadley:
That sounds reasonable. I'd recommend starting with a fairly mechanical translation of the current code from S3 to ggproto, and then once that's done correctly, thinking about how to extend the capabilities.

hadley:
Also - could you please shoot me an email? I have a non-GitHub question.

thomasp85:
I have made a fairly mechanical transformation to ggproto. The new classes are still not incorporated into the build process. The most opinionated decision I've made is to make layout a class method instead of free functions (`layout_grid`, `layout_wrap` etc). Further, the `$map` method for wrap and grid were essentially just calling `locate_wrap` and `locate_grid` respectively. The `locate_*` functions has therefor been incorporated directly into `$map`

I'm thinking of incorporating as much as possible of the code in the facet-*.r files into the base `Facet` class so that it is accessible to extension builders. The labeller functions should of course remain as free functions and the facet-labels.r file should probably be renamed to labeller.r or something similar (unless this messes up too much)

thomasp85:
@hadley All code in facet_veiwports.r seems isolated and is not referenced outside the file. Leftover from a refactor?

thomasp85:
This branch now has a complete transition to the ggproto Facet class and subclasses.

The public interface to Facet objects are the `$train`, `$map`, and `$render` functions i.e. these are the methods called during build. I'll look through the codebase and see if part of the facetting code are ripe for refactoring with extensibility in mind

thomasp85:
I'll obviously make sure that the unit tests are updated:-)

R/facet-.r
```diff
@@ -60,3 +63,105 @@ format.facet <- function(x, ...) {
 print.facet <- function(x, ...) {
   cat(format(x, ...), "\n")
 }
+
+
+# Here comes the ggproto rewrite
```
hadley: You can omit comments like this

R/facet-.r
```diff
@@ -60,3 +63,105 @@ format.facet <- function(x, ...) {
...
+  vars = function(self) {
+    stop("Not implemented", call. = FALSE)
+  },
+  print = function(self) {
+    cat("<", class(self)[[1]], ">\n", sep = "")
+    cat("Variables: ", self$vars(), '\n', sep = "")
+  }
+)
+
+# Helpers
```
hadley: By convention can you put `-----` after headers like this? Cmd + Shift + R in RStudio will do it for you.

R/facet-grid-.r
```diff
@@ -160,11 +163,16 @@ facet_grid <- function(facets, margins = FALSE, scales = "fixed", space = "fixed
   # Check for deprecated labellers
   labeller <- check_labeller(labeller)
 
-  facet(
+  # facet(
```
hadley: And don't worry about preserving old code here

hadley:
* Yeah, looks like you can delete `facet-viewports.R`

* I agree that `facet-labels.R` can be renamed

* At this point, I think it would be useful to start documenting the methods. A couple of lines for each one explaining it's chief responsibility would help me remember what they're supposed to do and will identify possibilities for further refactoring.

hadley:
The OO design seems reasonable but let's talk about on vchat.

thomasp85:
Finish cleaning up after mechanical transformation. Old code gone, unit tests rewritten, a few files renamed.

First thought - the result of `train` should be kept in the object if possible, but I'm still not sure if it is. It would reduce complexity of the build though

Will begin documenting the new Facet classes...

thomasp85:
Further, `$layout` should probably be collapsed into `$train`, as the responsibility of train is currently just to call layout and decorate the output

R/facet-null.r
```diff
@@ -7,70 +10,74 @@
...
 
-  layout <- gtable_matrix("layout", all,
-    widths = unit.c(grobWidth(axis_v), unit(1, "null")),
-    heights = unit.c(unit(aspect_ratio, "null"), grobHeight(axis_h)),
-    respect = respect, clip = c("off", "off", "on", "off"),
-    z = matrix(c(3, 2, 1, 4), ncol = 2, byrow = TRUE)
-  )
-  layout$layout$name <- c("axis-l", "spacer", "panel", "axis-b")
+    layout <- gtable_matrix("layout", all,
+                            widths = unit.c(grobWidth(axis_v), unit(1, "null")),
```
hadley: Can you please align so that additional lines of arguments are just two spaces deeper than the function name?

thomasp85: Sure - battling autoformat from RStudio;-)

thomasp85: Actually do you have a rule in terms of when new lines should start at the bracket and when it should start at fixed indent?

thomasp85:
Some more thoughts after the documentation. the panel class is a strange beast of information that captures scales, layout and ranges. Not sure all of this should be swallowed by Facet but definitely the layout information. Its going to take some digging to find where the different elements of panel gets used as it is passed around extensively... Thoughts?

hadley:
Panels are important - I have forgotten exactly what they do, but I think they basically encapsulate everything you need to plot a single panel of data. That includes scales, facets and coordinate systems.

thomasp85:
Yeah, they are definitely important. As I can see they are a vessel that carries the mapping from PANEL to location, the positional scales for each panel as well as axis information for each panel.

My thoughts, but this is probably for another PR, is that the layout is internal data to the Facet class and the axis information is internal information to the Scales class. What we are left with then is a collection of positional scales mapped to each panel?

thomasp85:
@hadley Can you comment on the current implementation? I'll hold of implementing the more demanding facetting until we are satisfied with the interface...

`FacetNull` has been migrated to show a minimal example of a Facet implementation - I've added some additional features to get a feel for how it is to program with facets and what is possible. You are free to disagree with the features as they are easily removed or modified.

If they remain I suggest we rename `facet_null` to `facet_single` as this will now be something users will begin to use for setting different options and `null` is kind of a strange name then...

Expect unit tests to mess up as I begin to migrate the code...

R/facet-.r
```diff
@@ -1,62 +1,190 @@
...
 
-facet_render <- function(facet, panels_grob, coord, theme, geom_grobs)
-  UseMethod("facet_render")
+  train = function(self, data) {
+    self$compute_layout(data, self$params)
+  },
+  map = function(self, data, layout) {
+    self$map_data(data, layout, self$params)
+  },
+  render_below = function(self, data, layout, x_scales, y_scales) {
```
hadley: Reading this afresh, I wonder if `back` and `front` might better suggest the direction of the stacking

thomasp85: I don't have any feelings with regards to that...

R/panel.r
```diff
@@ -27,7 +27,7 @@ new_panel <- function() {
 # @param plot_data the default data frame
 # @return an updated panel object
 train_layout <- function(panel, facet, data, plot_data) {
```
hadley: Should these become methods of `Layout`?

hadley:
Looks good so far.

Maybe `facet_one` instead of `facet_null`? (We'll need to leave `facet_null` with a deprecation warning in case someone is using it)

R/panel.r
thomasp85: panel.R should be deleted - I think I've grabbed everything I needed from there but was holding off for a bit...

R/facet-.r
hadley: Ok, lets change it — back and front are what drawing programs use, e.g. "send to back"

thomasp85:
With regards to renaming I think this should only be done if we want stuff like axis position and secondary axis be part of facetting rather than guides - otherwise `facet_null` will remain the default that no-one calls directly...

thomasp85:
re: axis specification

It could make sense to put it in guides and have the guides always create four axes. If no secondary axes are requested these will be zeroGrobs, but it will mean that facets can always assume the existence of axes on all sides

hadley:
I like that idea

thomasp85:
Hmm... The natural option would be to put it in `guides()` but currently this is only for non-positional guides. I think it would require a rather large rewrite for that to work...

An alternative would be to add a `secondary`argument to `scale_*_continuous` (doesn't make sense for discrete scales IMO) where the user can define the settings for the secondary axis...

This will reach very deep into the underlying code I think, but I think the facet rewrite is the best time to do it...

R/panel.r
has2k1: Getting rid of the `panel` as a code object/file name resolves the concept clash that has existed between `facet` and `panel`.

thomasp85: Yeah, panel now uniquely refers to a single layer stack within the code

smouksassi:
Regarding your numbered items above:
4. Decorate around the facets (both axes, facet headings, etc)
Will you consider extending facets so repeated facet grid across  columns or rows can be merged ? (I see facet_grid strips  similar to table headers)
https://stat.ethz.ch/pipermail/r-help/2008-July/167342.html
Another helpful feature can be to allow coloring of the facet background by the factor level.

thomasp85:
While colouring the strip background seems obvious it would interfere with ggplot2's odeon of one scale per aesthetic. Consider having fill mapped to some factor and then additionally colouring facet strips based on another factor. The viewer would be fooled into thinking the colouring was related (I think)

Regarding your first point I'm not sure I understand your idea. Couldn't you just create a new data column with some levels merged?

hadley:
@smouksassi please don't discuss new features here. I'm going to delete your comments to keep this thread clean

thomasp85:
Here is the preliminary implementation of both axis positioning and secondary axes. I've opted to following your suggestion for the secondary axis format (e.g. ~2*. + 5)

I'll add some more inline documentation once you've had a look through and given it a first nod of approval.

thomasp85:
e.g.

```r
qplot(cyl, mpg, data=mtcars, geom="point") + scale_y_continuous('test', sec.axis = ~.^2)
```

rplot

thomasp85:
I've added documentation for the changes and some inline comments in axis-secondary to clearly show intent

thomasp85:
facet_grid has been transferred to the new system. There has been added a new theme setting: `strip.placement` which controls whether strips are placed in between panels and axes or not. It can take the values `inside` and `outside` and defaults to `outside`. There is also an `.x` and `.y` version to control the placement separately for each strip.

An example showing it in combination with the new axis functionality (not saying this is a good plot):

```r
qplot(cyl, mpg, data = mtcars, geom= 'point') +
  facet_grid(am~gear, switch = 'both') + 
  theme(strip.placement.x = 'outside') + 
  scale_y_continuous(position = 'right', sec.axis = ~.^2)
```

rplot01

thomasp85:
So - at last `facet_wrap` has been moved to the new system!

Something you'll notice is that I've put facet related helpers as member functions of the Facet class so that they do not need to be exported while still being accessible to extension developers - I don't know if you are OK with this pattern - otherwise we'll move them out...

thomasp85:
@hadley should we add a `setup_params` and `setup_data` method to `Facet` in line with `Stat` and `Geom` ?

I can think of times where it would come in handy...

hadley:
Sure - I think consistency is a powerful argument.

thomasp85:
This PR now fixes #1596 and #1607

thomasp85:
@hadley Everything has been converted and the two active facet feature/bugs issue has been solved. Should we feature freeze this PR now and focus on making it mergeble (update unit tests etc) or are there any additional stuff I should add?

hadley:
I'd say freeze features and prepare it for merging.

thomasp85:
I've changed the global figure settings for the vignette to increase the figure size - the former was to small to really look at facetted plots. Are you ok with that @hadley?

thomasp85:
And it passes all checks 🍾🍾🍾

What is left IMO

- [ ] Tweaks to the theming system to support top and right axes better
- [ ] An extensive NEWS bullet

thomasp85:
At least `axis.text` and `axis.label` need to be set separately for each side - do you want ticks etc to also be differentiated?

My suggestion would be to add `axis.text.x.top` that inherits from `axis.text.x` etc but leave `.bottom` and `.left`out as these are implicitly set with the base version - thoughts?

thomasp85:
@hadley - I think the time has come for you to read through the changes and begin to flag stuff that needs attention and try out with a facet extension of your own...

hadley:
Can you make sure this branch is writable to me? (Should be a check box in the PR). Then I can push a few minor changes

hadley:
There are a few problems with the final example in the `facet_wrap()`:

```R
ggplot(economics_long, aes(date, value)) +
   geom_line() +
   facet_wrap(~variable, scales = "free_y", nrow = 2, switch = "x") +
   theme(strip.background = element_blank())
```

* The logic for converting `switch` to `strip.position` needs tweaking
* The example should be updated
* Why is the x axis drawn on the top?
* The y axis labels are left aligned instead of right aligned?

hadley:
Gah, can you merge/rebase post scale constructor changes?

thomasp85:
Do the facet_wrap problems persist?

hadley:
Yeah, I'm still seeing the weirdness.

thomasp85:
Hmm - me too... Strange.

I'll get back to you when this is fixed

thomasp85:
Thank god for code review

rplot

hadley:
I think the final example could be expanded a bit:

```R
#' # Use `strip.position` to control where strips are display:
#' ggplot(economics_long, aes(date, value)) +
#'   geom_line() +
#'   facet_wrap(~variable, scales = "free_y", nrow = 2, strip.position = "bottom")
#'   
#' # This typically requires some other tweaks to the theme:
#' ggplot(economics_long, aes(date, value)) +
#'   geom_line() +
#'   facet_wrap(~variable, scales = "free_y", nrow = 2, strip.position = "bottom") +
#'   theme(
#'     strip.background = element_blank(), 
#'     strip.placement = "outside",
#'     strip.text = element_text(size = rel(1))
#'   ) +
#'   xlab(NULL)
```

And it makes me want an option where every row gets axes, not just the bottom.

But I'd say we should merge now, and then I can start tweaking the examples and doing some more checks.

hadley:
Here's another example that looks wrong:

```R
ggplot(mpg, aes(drv, model)) +
  geom_point() +
  facet_grid(manufacturer ~ ., scales = "free", space = "free") +
  theme(strip.text.y = element_text(angle = 0))
```

hadley:
But lets merge regardless of that problem, and I'll file bugs as I find them.

thomasp85:
No, let me try to get the visual tests working for this PR and see what pops up prior to merging. There are obviously some unexpected regressions

hadley:
I really think we should merge this now as this branch is getting _huge_ and is starting to block other features.

thomasp85:
your wish is my command

PR #1637, opened 2016-05-26
New graphics: geom_sina(), position_sina()

sidiropoulos:
Hi,

as discussed in the  [Google group](https://groups.google.com/forum/#!topic/ggplot2/FJA8EMs1XeE) I have implemented a new geometry based on [sinaplot](https://cran.r-project.org/web/packages/sinaplot/vignettes/SinaPlot.html). 

Sina plots combine the power of violin and jitter plots in a single chart where information for both the distribution and the number of the samples is conveyed. 

Sina plot is available in R as a CRAN [package](https://cran.r-project.org/web/packages/sinaplot/index.html) and, at the moment, it is using ggplot2 to produce the output. It would make much more sense though for it to be included as a geom in ggplot, while the CRAN package would use the default R graphics.

Looking forward to get your feedback on both the idea and the implementation and I hope you will find useful enough to be included in ggplot2!

Regards,
Nikos


has2k1:
This is an interesting type of plot and is an improvement on the common violin + jittered points visualisations. It is also kind of blurs the conceptual distinction between `position` adjustments and `statistics` as defined by the "grammar".

Given that the two (positioning and statistical calculations) are not separable and that the statistics
done on the data are significant, I think this would tilt the implementation to be `stat` based rather than `position` based.

This is a tally of what I see happening and what concept describes it best,

`stat`:  binning, normalization, density estimation, sampling
`position`: jittering

With that I would push it into the `stat_sina` category.

sidiropoulos:
Hi Hassan,

Do you think it can be as simple as "renaming" the `position-sina.r` to `stat-sina.r`? Meaning that I would keep all the transformation function in the stat file and leave `geom_sina` as is.

has2k1:
@sidiropoulos, you have done all the hardwork already. To convert what you have to a `stat` you just need to reorganise it. All the transformation code would remain in the `stat_sina` file.

Now, I read through the paper only once and do not yet have a concrete grasp of the process. With that said, I can certainly help some more. I'll do a code-review in two steps to help on the reorganisation. The first review will be to get it working as a `stat`, then the second (optional) which would be to make it even clearer.

R/geom-sina.r
```diff
@@ -0,0 +1,126 @@
...
+#'
+#' # Reorder levels
+#' df$Distribution <- factor(df$Distribution,
+#'                           levels(df$Distribution)[c(3, 4, 1, 2)])
+#'
+#' p <- ggplot(df, aes(Distribution, Value))
+#' p + geom_boxplot()
+#' p + geom_violin() + geom_sina()
+geom_sina <- function(mapping = NULL, data = NULL,
+                      stat = "identity", position = "sina",
```
has2k1: Would definitely change to `stat='sina', position='identity'`

R/geom-sina.r
```diff
@@ -0,0 +1,126 @@
...
+                      ...,
+                      binwidth = 0.02,
+                      scale = TRUE,
+                      neighbour_limit = 1,
+                      method = NULL,
+                      adjust = 1,
```
has2k1: These parameter would go into `stat_sina`

R/position-sina.r
```diff
@@ -0,0 +1,221 @@
...
+                                 scale = params$scale,
+                                 neighbour_limit = params$neighbour_limit,
+                                 method = params$method,
+                                 adjust = params$adjust)
+    trans_y <- NULL
+
+    transform_position(data, trans_x, trans_y)
+  }
+)
+
```
has2k1: There would be no need for `compute_layer` function.

R/position-sina.r
```diff
@@ -0,0 +1,221 @@
...
+.sina <- function(x, data, binwidth, scale, neighbour_limit, method, adjust) {
+
+
+  ### Initialise variables
+  #x-axis transpose vector
+  trans_x <- list()
+
+  #assing groups as factors of x
+  groups <- factor(x)
```
has2k1: The groups are available at `data$group` and it would guarantee that you don't run into edgecase bugs because a user became very creative in how they used `aes()`.

R/position-sina.r
```diff
@@ -0,0 +1,221 @@
...
+        trans_x[[j]]$x[points] <- trans_x[[j]]$x[points] +
+          (x_translation * global_scaling_factor * group_scaling_factor)
+      }
+    }
+  }
+
+  #collapse list to data frame
+  trans_x <- do.call(rbind, trans_x)
+  #return only the transposed x values in input order
+  trans_x$x[order(trans_x$idx)]
```
has2k1: For the `stat$compute_panel` method, you would return the `trans_x` dataframe with the `x` and `y` values. I would delete `idx` because it would not be needed any more. However, find a way to include the `group` in there.

Maybe this would do it

```R
"group" = data$group[keep],
```

R/position-sina.r
```diff
@@ -0,0 +1,221 @@
...
+  #window width
+  window_size <- (ymax - ymin) * (bw + 1e-8)
+
+  bins <- c()
+  for (i in 0:ceiling(1 / bw)) {
+    bins <- c(bins, ymin + i * window_size)
+  }
+
+  bins
+}
```
has2k1: I hope I have caught everything.

has2k1:
The first review is finished, I did not want to mix in more complicated (and general and vague) suggestions that would affect the implementation of the process. That would come in another review as it would require that you adjust the implementation.

sidiropoulos:
@has2k1 
Thanks so much for the time and guidance :-) 

I have re-implemented geom_sina() according to your review. This is my first pull_request, sorry in advance if I'm not doing things the proper way (pushing the right branch, etc).

R/stat-sina.r
```diff
@@ -0,0 +1,196 @@
...
+    neighbours <- list()
+    max_neighbours <- 0
+
+    #method == "density"
+    if (method == "density") {
+      max_density <- 0
+      densities <- list()
+    } else {
+      #scale adjust value
+      adjust <- adjust / 5
```
has2k1: `adjust` could be modified in `setup_params`.

R/stat-sina.r
```diff
@@ -0,0 +1,196 @@
...
+
+        #per bin sample density
+        densities[[j]] <- stats::density(data$y[keep], adjust = adjust)
+
+        #find the highest density value
+        tmp_max_density <- max(densities[[j]]$y)
+
+        if (tmp_max_density > max_density)
+          max_density <- tmp_max_density
+      }
```
has2k1: This (103-118) chunk has some dead code? `max_neighbours` and `max_density` are not used.

R/stat-sina.r
```diff
@@ -0,0 +1,196 @@
...
+    for (j in levels(factor(data$group))) {
+
+      keep <- data$group == j
+
+      if (nrow(data[keep, ]) < 2 ) next
+
+      #confine the samples in a (-0.5, 0.5) area around the class center
+      if (method == "density") {
+        if (max(densities[[j]]$y) > 0.48)
+          global_scaling_factor <- 0.48 / max(densities[[j]]$y)
```
has2k1: Does it make sense to life `0.48` into a full parameter value that the user can play with?

R/stat-sina.r
```diff
@@ -0,0 +1,196 @@
...
+      }
+    }
+
+    # ------------------------------------------------------------------------ #
+
+    for (j in levels(factor(data$group))) {
+
+      keep <- data$group == j
+
+      if (nrow(data[keep, ]) < 2 ) next
```
has2k1: Small style inconsistency. Used `if (sum(keep) < 2)` above.

R/stat-sina.r
```diff
@@ -0,0 +1,196 @@
...
+        } else
+          global_scaling_factor <- 1
+      }
+
+      #scale all neighbourhoods based on their density relative to the
+      #densiest neighbourhood
+      if (scale == TRUE)
+        group_scaling_factor <- max(neighbours[[j]]) / max_neighbours
+      else
+        group_scaling_factor <- 1
```
has2k1: I think the scaling can be refactored out. More on that in a subsequent comment about `calculate_group` and `calculate_panel`.

R/stat-sina.r
```diff
@@ -0,0 +1,196 @@
...
+  #window width
+  window_size <- (ymax - ymin) * (bw + 1e-8)
+
+  bins <- c()
+  for (i in 0:ceiling(1 / bw)) {
+    bins <- c(bins, ymin + i * window_size)
+  }
+
+  bins
+}
```
has2k1: Not really important, but `bin.R` has some binning functions and using them allows for binning according to the bin width or the number of bins.

has2k1:
@sidiropoulos, the code is shaping up well. This is the part I hinted at earlier that deals with the reorganisation and could be vague.

The goal is to split to the code into a chunk implemented in `compute_group` and a chunk for `compute_panel`? How this divide comes about is by recognising that the scaling helps make the different groups look proportionally nice on the same plot. Other than that the key feature in revealing the shape of the distribution is the `x_translation`. More important (I think) is that it would make the algorithm look less complicated by reducing the "book-keeping" code required for the scaling.

As I see it, if from `compute_group` you return the `x_translation`, `max_neighbours` (this is a maximum bin count) and `max_density` then in `compute_panel` you should be able to do the scaling -- maybe with the help of `plyr`.

`stat_sina` has a slightly more complicated scaling but to see an example of the split in code, checkout `stat_ydensity`. In there, you should not that to get the full range of the y values over which you would calculate the bins  you use `scales$y$dimension()`. Huh, I wonder what it would look like if the bins are calculated based on the per group range?

Finally in `calculate_panel` you would then consider which of the columns in the data.frame to leave in there so that the user can access them as calculated columns, e.g. `aes(x=..x_translation..)` or `aes(x=..scaled..)` or whatever. These give a window into the different progressions of creating the visualisation.

R/stat-sina.r
sidiropoulos: It is a confining factor for not letting the samples overlap with adjacent groups. I don't think it is a good idea to let the user exceed that limit.

R/stat-sina.r
sidiropoulos: I haven't overlooked this comment. I like the simplicity in bin_y, but it might not be needed at all. Will look into it in the next round of changes.

sidiropoulos:
@has2k1 the code is successfully split in `compute_layer` and `compute_group`. 

I'm a bit bugged about how the computed variables of a `stat` or `geom` are accessed. Hence, I don't exactly understand how to return `..x_translation..` properly. 

On another note, are [travis](https://travis-ci.org/hadley/ggplot2/builds/135865833) checks for R code bugged at the moment?

R/stat-sina.r
```diff
@@ -0,0 +1,186 @@
...
+
+    params
+  },
+
+
+  compute_panel = function(self, data, scales, binwidth = 0.02, scale = TRUE,
+                           neighbour_limit = 1, method = "d", adjust = 1,
+                           na.rm = FALSE) {
+
+    bins <- bin_y(data$y, binwidth)
```
has2k1: Should be

```R
bins <- bin_y(scales$y$dimension(), binwidth)
```
That makes it work better with faceting, especially with `scales='fixed'`

R/stat-sina.r
```diff
@@ -0,0 +1,186 @@
...
+
+
+#' @rdname ggplot2-ggproto
+#' @format NULL
+#' @usage NULL
+#' @export
+StatSina <- ggproto("StatSina", Stat,
+
+  required_aes = c("x", "y"),
+
```
has2k1: Add this for the computed variable
```R
default_aes = aes(x= ..scaled..),
```

For consistency (see `scale_density`) and see comment below why I used 'scaled'

R/stat-sina.r
```diff
@@ -0,0 +1,186 @@
...
+    #scale all neighbourhoods based on their density relative to the
+    #densiest neighbourhood
+    if (scale)
+      group_scaling_factor <- data$max_neighbours / max(data$max_neighbours)
+    else
+      group_scaling_factor <- 1
+
+    #translate x coordinates
+    data$x <- data$x + data$x_translation * group_scaling_factor
```
has2k1: It is better to create a new column

```R
data$scaled <- data$x + data$x_translation * group_scaling_factor

# maybe even
data$unscaled <- data$x + data$x_translation

# then delete
data$x_translation <- NULL
```
Every column left in there becomes a computed aesthetic.

has2k1:
For the computed aesthetics, leaving other columns in the dataframe can be of help in explaining the `stat`, because then you plot them.

The only other issue is probably subjective is the `neighbours` term, I find it more intuitive to call them `bin_counts`. On first reading the paper I thought k-nearest neighbour  was involved. And if you leave the
`max_bin_counts` in the dataframe, you can visualise them too!

Other than that there is not much more to nitpick except you do more nitpicking yourself. This would make a great addition if Hadley is in favour.

sidiropoulos:
@has2k1 I like the idea of renaming `neighbourhood` to `bin_counts` and all the variable names accordingly. I used most of your suggestions and there's also plenty room for nitpicking which I'm still doing. 

One thing I can't get done though is the `default_aes = aes(x = ..scaled..)` part. I used `stat_density`as a guideline (also other stats) but no matter how I set it, it will always use the `data$x` value as `aes(x)` which is basically the same as calling `geom_point`.

has2k1:
Oh, I see the conflict. There is clash between `required_aes` and `default_aes` where is`default_aes` guaranteed to lose.

You could do this

```R
# stat_sina
default_aes = aes(xend = ..scaled..)

# geom_sina
GeomSina <- ggproto("GeomSina", GeomPoint,
    transform(data, x = xend)
  }
)
```

For an example of how to change up `geom_sina` see, `geom_spoke`, `geom_tile` & `geom_bar`.

If successful, something like

 ```R
geom_segment(data, aes(x=x, y=y, yend=y), stat='sina')
```
 should give a not so interesting plot.

**Comment by sidiropoulos** [comment]
It works now! 
`scaled` and `bin_counts` are returned and the latter can be used to colour the samples, giving an informative visualisation of the sample count. 

Thank you once again. You've been of huge help in implementing this and teaching me a few things on how ggplot works! 😃

File: R/stat-sina.r [hunk 16]
```diff
@@ -98,59 +105,61 @@ StatSina <- ggproto("StatSina", Stat,
...
 
-    #scale all neighbourhoods based on their density relative to the
-    #densiest neighbourhood
-    if (scale)
-      group_scaling_factor <- data$max_neighbours / max(data$max_neighbours)
-    else
+    #scale all bins based on their density relative to the densiest bin
+    if (scale) {
+      group_scaling_factor <-
+        ddply(data, .(group), mutate,
```
has2k1: Use the namespaced path `plyr::ddply`

R/geom-sina.r
```diff
@@ -0,0 +1,133 @@
...
+#' p + geom_violin() + geom_sina(scale = FALSE)
+geom_sina <- function(mapping = NULL, data = NULL,
+                      stat = "sina", position = "identity",
+                      ...,
+                      method = NULL,
+                      na.rm = FALSE,
+                      show.legend = NA,
+                      inherit.aes = TRUE) {
+
+  method <- match.arg(method, c("density", "counts"))
```
has2k1: I think the `method` parameter can be left out of the geom.

has2k1:
Nice, the latest version runs without any issues on my end.

benmarwick:
+1 to this addition, I'm looking forward to using `geom_sina` in ggplot.

hadley:
I think this is a nice idea, but I'm only accepting very fundamental geoms into ggplot2 now that you can write them in your own package.  To me, it feels like it would be a very good fit to <https://github.com/eclarke/ggbeeswarm>

thomasp85:
I would accept this into [ggforce](https://github.com/thomasp85/ggforce) (not on CRAN yet) and I know that @hrbrmstr is also open to additions to [ggalt](https://github.com/hrbrmstr/ggalt), though ggbeeswarm is also a very good fit...

imagejan:
Just to follow up on this, in case others are looking for sinaplot here as well:
`geom_sina` is now available in ggforce with https://github.com/thomasp85/ggforce/pull/12.

PR #1644, opened 2016-06-03
Specify alpha for outlier points in geom_boxplot.

jonathan-g:
Add a parameter `outlier.alpha` to geom_boxplot to allow the outlier points to have different alpha from the boxes. This parameter works the same way as outlier.colour, outlier.shape, etc.

This is useful when plotting large data sets, so even when only a small fraction of the data is outliers, the number of outlier points can be large enough that it's hard to see them all. In such a case, setting alpha.outlier < 1 makes it easier to see the density of outliers.


R/geom-boxplot.r
```diff
@@ -204,7 +207,7 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
         size = outlier.size %||% data$size[1],
         stroke = outlier.stroke %||% data$stroke[1],
         fill = NA,
-        alpha = NA,
+        alpha = outlier.alpha %||% data$alpha[1],
```
hadley: `%||%` works with `NULL` not  `NA`. I think that just means the default function param above should be NULL.

NEWS
```diff
@@ -1,3 +1,7 @@
+* `geom_boxplot` gain new `outlier.alpha` argument for controlling the alpha of
```
hadley: Should be in NEWS.md

hadley:
Could you please include a couple of before/after example plots in the PR?

jonathan-g:
Will do.

R/geom-boxplot.r
jonathan-g: Thanks for catching that. I changed the default NULL and tested and it seems to work fine that way.

jonathan-g:
I have made the changes you requested, including adding an example of using outlier.alpha with the diamonds data.

jonathan-g:
Just for completeness, here is the output from the example plots so you can see them without having to run the example code yourself:

```
ggplot(diamonds, aes(carat, price)) +
     geom_boxplot(aes(group = cut_width(carat, 0.25))) + ggtitle('Before (no outlier.alpha)')
ggplot(diamonds, aes(carat, price)) +
     geom_boxplot(aes(group = cut_width(carat, 0.25)), outlier.alpha = 0.1) + ggtitle("After (outlier.alpha = 0.1)")
```

before
after

hadley:
@thomasp85 can you help finish this one off please?

thomasp85:
@jonathan-g Can you bring this PR up to date with the main branch and I'll have a look through it so we can get it merged

jonathan-g:
@thomasp85 I've fixed the conflicts with the main branch and confirmed that it passes R CMD check with no warnings or errors.

thomasp85:
Thanks

PR #1649, opened 2016-06-13
Use validate_params = FALSE to pass unvalidated params as a list

faizan-khan-iit:
This PR is with respect to [this comment about passing additional parameters](https://github.com/hadley/ggplot2/issues/1585#issuecomment-225583280).

The motivation of this PR is that packages like Animint depend on additional params to produce plots and animations using ggplot.

After the changes, the following code will still produce the same error as before:
```
library("ggplot2")

viz <- list(iris=ggplot()+
    geom_point(aes(Petal.Width, Sepal.Length, showSelected=Species),
               data=iris, chunk_vars=character()))
```
```
Error: Unknown parameters: chunk_vars
```

While this code will only produce a warning and pass the additional params as a list `extra_params`:
```
library("ggplot2")

viz <- list(iris=ggplot()+
    geom_point(aes(Petal.Width, Sepal.Length, showSelected=Species),
               data=iris, chunk_vars=character(),validate_params = F))
```
```
Warning message:
Unknown parameters used: chunk_vars
Set validate_params to TRUE avoid errors 
```
This way we will be able to access these params through the `extra_params` list under `plot$layers` and use them in development of other packages, and the general user should still see the same error as before, unless he/she specifies the `validate_params` option. I think this achieves both the goals.

EDIT: Please see my other comment below for changes.

### DIFF
--- a/tests/testthat/test-layer.r
+++ b/tests/testthat/test-layer.r
@@ -17,11 +17,8 @@ test_that("Unknown params create error with validate_params = TRUE", {
                "Unknown parameters")
 })
 
-test_that("Unknown params create warning with validate_params = FALSE", {
-  expect_warning(geom_point(blah = "red", validate_params = FALSE),
-                 paste("Unknown parameters used: blah",
-                       "Set validate_params to TRUE to avoid errors",
-                       sep = "\n"))
+test_that("Unknown params don't create error with validate_params = FALSE", {
+  expect_silent(geom_point(blah = "red", validate_params = FALSE))
 })
 
 test_that("Unknown params go in extra_params, not aes_params", {
--- a/tests/testthat/test-layer.r
+++ b/tests/testthat/test-layer.r
@@ -12,6 +12,26 @@ test_that("unknown params create error", {
   expect_error(geom_point(blah = "red"), "Unknown parameters")
 })
 
+test_that("Unknown params create error with validate_params = TRUE", {
+  expect_error(geom_point(blah = "red", validate_params = TRUE),
+               "Unknown parameters")
+})
+
+test_that("Unknown params create warning with validate_params = FALSE", {
+  expect_warning(geom_point(blah = "red", validate_params = FALSE),
+                 paste("Unknown parameters used: blah",
+                       "Set validate_params to TRUE to avoid errors",
+                       sep = "\n"))
+})
+
+test_that("Unknown params go in extra_params, not aes_params", {
+  l <- geom_point(some_param = "value1",
+                  size = "big",
+                  validate_params = FALSE)
+  expect_equal(l$extra_params, list(some_param = "value1"))
+  expect_equal(l$aes_params, list(size = "big"))
+})
+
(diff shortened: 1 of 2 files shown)


tdhock:
This looks like a great solution to me @faizan-khan-iit -- please add a test that captures our use case.

tests/testthat/test-layer.r
```diff
@@ -12,6 +12,26 @@ test_that("unknown params create error", {
   expect_error(geom_point(blah = "red"), "Unknown parameters")
 })
 
+test_that("Unknown params create error with validate_params = TRUE", {
+  expect_error(geom_point(blah = "red", validate_params = TRUE),
+               "Unknown parameters")
+})
+
+test_that("Unknown params create warning with validate_params = FALSE", {
+  expect_warning(geom_point(blah = "red", validate_params = FALSE),
```
tdhock: Hey @faizan-khan-iit I would say that when the user specifies `validate_params=FALSE`, there is no need to warn them. Could you instead add an expectation for no warning?

faizan-khan-iit: @tdhock Sure. That is certainly doable.
My idea was that while using unknown params:

- If a user doesn't specify `validate_params`, ggplot should throw an error.
- If a user does specify `validate_params = FALSE`, we let them continue with a warning.

faizan-khan-iit:
Now the following code will still produce the same error as before:
```
library("ggplot2")

viz <- list(iris=ggplot()+
    geom_point(aes(Petal.Width, Sepal.Length, showSelected=Species),
               data=iris, chunk_vars=character()))
```
```
Error: Unknown parameters: chunk_vars
```
As does this one when we specify `validate_params  = TRUE`:
```
library("ggplot2")

viz <- list(iris=ggplot()+
    geom_point(aes(Petal.Width, Sepal.Length, showSelected=Species),
               data=iris, chunk_vars=character(), validate_params = TRUE))
```
```
Error: Unknown parameters: chunk_vars
```

While this code will not produce an error as `validate_params = FALSE` and pass the additional params as a list `extra_params`:
```
library("ggplot2")

viz <- list(iris=ggplot()+
    geom_point(aes(Petal.Width, Sepal.Length, showSelected=Species),
               data=iris, chunk_vars=character(), validate_params = FALSE))
```

I got rid of the warning while using `validate_params  = FALSE` as I think we can safely assume that the user knows what he is doing, so there should be no need for a warning. Any thoughts?

tdhock:
Hey @hadley @wch when you get some time, can you please review this PR and consider merging with master?

Like our previous PR https://github.com/hadley/ggplot2/pull/1121 which was accepted last year, this PR is for animint https://github.com/tdhock/animint/pull/135

Thanks in advance!

hadley:
I don't like this approach because it requires the rather ugly `extra_params` argument.  Can you please open an issue (if there isn't one already) and we can discuss approaches there?

tdhock:
continuing disussion in https://github.com/hadley/ggplot2/issues/1585

PR #1650, opened 2016-06-14
Small change to help documentation

DDClark:
Added color to the `geom_boxplot` call in the last example to distinguish outlier points from data points plotted with `position_jitter`


hadley:
Thanks! I tweaked it myself since you spelled colour wrong 😛

PR #1657, opened 2016-06-26
Update plot.r

thk686:
Added method to dispatch ggplot on a time series object

(Oops: should be stats::time())


hadley:
I think this would be best as a new `fortify` method. Would you be interested in implementing that in a new PR?

thk686:
Makes sense. I will try to look at it soon.

PR #1661, opened 2016-07-06
Expose argument n from stats::density in the interface of stat_density.

hbuschme:
The argument controls the number of equally spaced points at which the density is to be estimated. As in `stats::density` the argument `n` defaults to 512. It is exposed through the function interface of `stat_density` and can thus also be used from `geom_density`.

An existent local variable `n`, which holds the number of values in the input vector, has been renamed to `nx` to avoid a name clash.

I needed this argument because plots created with `geom_density` contained a large number of points (1024?) per density curve. This resulted in large file sizes when saving the plot as a vector graphic. I think that it is generally useful to be able to control this parameter of the density function.

I rebuild the documentation, and tested the new functionality. However, I added a new argument to an existing function and – not being an R expert – I am not sure whether this might break someones code. The argument has a default value, but I did not add it at the very end of the argument list but rather at a point where it seemed fitting.

### DIFF
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -4,6 +4,8 @@
 #'   \code{\link{density}} for details
 #' @param kernel kernel used for density estimation, see
 #'   \code{\link{density}} for details
+#' @param n number of equally spaced points at which the density is to be
+#'   estimated, see \code{\link{density}} for details
 #' @param trim This parameter only matters if you are displaying multiple
 #'   densities in one plot. If \code{FALSE}, the default, each density is
 #'   computed on the full range of the data. If \code{TRUE}, each density
@@ -25,6 +27,7 @@ stat_density <- function(mapping = NULL, data = NULL,
                          bw = "nrd0",
                          adjust = 1,
                          kernel = "gaussian",
+                         n = 512,
                          trim = FALSE,
                          na.rm = FALSE,
                          show.legend = NA,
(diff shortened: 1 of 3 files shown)


R/stat-density.r
```diff
@@ -4,6 +4,8 @@
 #'   \code{\link{density}} for details
 #' @param kernel kernel used for density estimation, see
 #'   \code{\link{density}} for details
+#' @param n number of equally spaced points at which the density is to be
```
hadley: Can you please add a note about powers of two being faster? (IIRC)

hadley:
This looks reasonable. Would you mind including a couple of plots in this PR just to verify that is works? (i.e. picking a really extreme n)

hbuschme:
I have added a note about powers of two being recommended – though n is rounded to the nearest power of two automatically if n > 512.

Here are the requested plots for n = 512, 64, 16, and 4.

    data(tips, package = "reshape")
    ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=512)

ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=512)

    ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=64)
ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=64)

    ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=16)

ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=16)

    ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=4)

ggplot(tips, aes(x=tip)) + geom_density(stat=StatDensity, n=4)

hadley:
Thanks!

PR #1663, opened 2016-07-08
Use vdiffr for visual regression test

thomasp85:
@hadley this PR will begin implementing vdiffr for testing visual regression


hadley:
cc @lionel- just so he knows about it

lionel-:
Cool, I'll take some time next month to rewrite the non-licensed JS lib used in vdiffr so I can release it on CRAN.

thomasp85:
This first change is a mechanical transformation of the tests located in the visual_test folder. All visual tests have been collected in the same file, though one could argue that they should be distributed among multiples - I'll let you decide.

Further, There is obvious room for expansion of test suite, and the coverage of the currently covered features is very uneven. Have you given this any thoughts?

.travis.yml
```diff
@@ -4,6 +4,7 @@ sudo: false
 
 r_github_packages:
   - hadley/staticdocs
+  - lionel-/vdiffr
```
hadley: This would be better in `DESCRIPTION` remotes.

hadley:
Splitting into different files would be great please.

lionel-:
You need to commit the generated SVG files as well otherwise the tests will be skipped on Travis.

I just noticed with the ggstance repo that github does not understand the generated files, I'll have to look into this: https://github.com/lionel-/ggstance/tree/master/tests/figs

thomasp85:
Illustrator also complains about the files

thomasp85:
@lionel- This is not something that happens during upload to github in case you're wondering. The generated svgs contain a NUL character at the end (following the closing svg tag). Is the device not properly closed perhaps?

lionel-:
yup I'm going look into this soonish.

It may also be due to some xml manipulation that adjusts the dimensions of the container. I don't remember exactly why it was necessary but it had to do with the scaling of the htmlwidgets inside the shiny app (that was quite a hairy ball).

I'll try to solve this tomorrow or the day after. Do you have any other issue with vdiffr?

lionel-:
> I'll try to solve this tomorrow or the day after

I'll also rewrite the TwoFace JS lib so I can make a proper release. Unless @hongymagic would like to licence it as open source, that'd be great!

thomasp85:
Not yet but something might pop up once I get it to work past this :-)

lionel-:
The tests and the shiny app should work properly even if the SVGs are not 100% valid.

thomasp85:
Hmm... I'll have to look into why travis fails then

lionel-:
They probably need more conventional filenames.

thomasp85:
What's wrong with the filenames ;-) 

i'll see if that fixes it

thomasp85:
They all fail - even basic ones... This is not an issue with randomisation but somehow with drawing of text elements

lionel-:
Weird, I didn't have any issue with ggstance.

thomasp85:
Can it be a Mac issue? Not that I can see how...

lionel-:
Ok don't waste too much of your time on this, I'll work on vdiffr and try to fix this PR tonight or tomorrow.

thomasp85:
Ok - I'll run a last test to see if time has magically fixed it :-)

hongymagic:
> I'll also rewrite the TwoFace JS lib so I can make a proper release. Unless @hongymagic would like to licence it as open source, that'd be great!
> 
Hi guys, not sure what's going on here, but more than happy to add open source license if you need it.

Will get onto it tmr morning.

lionel-:
@hongymagic Great, thanks a lot!

lionel-:
Hi @thomasp85,

The issue was that two testcases were given the same filename. Also I updated vdiffr to fix the generated SVGs, you were correct that the trailing character was the culprit. I sent you a PR with those changes.

From a cosmetic point of view, I think it's better to have the test labels in `test_that()` statements rather than in the SVG filenames. So instead of:

```{r}
test_that("Polar coordinates draws correctly", {
  dat <- data.frame(x = 0:1, y = rep(1:80, each = 2))

  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y, group = factor(y))) + geom_line() + coord_polar(),
    "Concentric_circles_at_theta-1-80"
  )
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 80, group = factor(y))) + geom_line() + coord_polar(),
    "Concentric_circles_at_theta-1-80-80"
  )
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 40, group = factor(y))) + geom_line() + coord_polar(),
    "Concentric_circles_at_theta-1-80-40"
  )

  ...
}
```

It'd be better to have:

```{r}
dat <- data.frame(x = 0:1, y = rep(1:80, each = 2))

test_that("Concentric circles at theta = 1:80", {
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y, group = factor(y))) + geom_line() + coord_polar(),
    "coord-polar-concentric-80"
  )
}

test_that("Concentric circles at theta = 1:80 - 80", {
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 80, group = factor(y))) + geom_line() + coord_polar(),
    "coord-polar-concentric-80-minus-80"
  )
})

test_that("Concentric circles at theta = 1:80 - 40", {
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 40, group = factor(y))) + geom_line() + coord_polar(),
    "coord-polar-concentric-80-minus-40"
  )
})

...
```

hongymagic:
@lionel- I'm not sure what part of TwoFace you want, the repo seems to be empty (no code). If you're referring to code here: http://jsfiddle.net/davidhong/wbxVf/ just take it and use it. Stuff going on this PR seems incomprehensible to me 😝

lionel-:
@hongymagic Yes I use a slightly modified version of the jsfiddle code here: https://github.com/lionel-/vdiffr/blob/master/inst/htmlwidgets/lib/twoface.js

The thing is, we need a proper licence for inclusion in the R package manager. I guess if you just agree here that I put a MIT or GPL mention in that file, we're good :)

hongymagic:
Yes I agree that you can put a MIT mention in that file, and we are all good 👍

lionel-:
Thanks!

lionel-:
Wouldn't these filenames be informative enough? e.g. `coord-polar-concentric-80-minus-40`:

- we have the test file as prefix as `coord-polar` means the case can be found in `test-coord-polar.R`
- and `concentric-80-minus-40` as description.

I think it's possible to extract the description of the `test_that()` statements, and then in the future we could display that information in the Shiny app. Also maybe the `figs` folder should be organised in subfolders named after the test files?

By the way, I added a new argument to `manage_cases()` to allow filtering the test files in which to collect visual cases. This can speed up the launch of the shiny app quite a lot with large codebases like ggplot2.

thomasp85:
If possible I would like to avoid wrapping each visual test in their own `test_that` statement. Would it be possible to add a `title` argument to `expect_doppelganger` where there wouldn't be need to consider filename-safe strings? Don't know if this makes sense and is doable...

lionel-:
I think it's more consistent with the testthat API to give the test descriptions with `context()` and `test_that()` rather than with expectations.

thomasp85:
But you can have multiple visual tests covering the same case and you would like to keep track of which is which? I generally agree with your observation, but visual tests can tend to be more high-level and less defined making a per-test description useful (just for the add-in, mainly)

Consider these tests:

```r
expect_equal(sum(x), 10)

expect_length(x, 5)
```

It is very obvious what we're testing for and what is expected by the nature of the API

On the other hand `expect_dobbelganger(p, "filename")` does not communicate anything about what we're testing for besides visual regression in general. It is true that we can put describing text in the second argument but the constraints of only using filename-safe characters makes this tedious and limited. I'm simply advocating for something like:

```r
test_that("Polar coordinates draws correctly", {
  dat <- data.frame(x = 0:1, y = rep(1:80, each = 2))

  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y, group = factor(y))) + geom_line() + coord_polar(),
    "polar_coord_1",
    "Concentric circles at theta = 1:80"
  )
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 80, group = factor(y))) + geom_line() + coord_polar(),
    "polar_coord_2",
    "Concentric circles at theta = 1:80 - 80"
  )
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 40, group = factor(y))) + geom_line() + coord_polar(),
    "polar_coord_3",
    "Concentric circles at theta = 1:80 + 100"
  )
})
```

It's true that wrapping each visual test in its own `test_that` statement will achieve the same but tests are already a drag to write so I think that minimising boilerplate code when possible is a good thing - also you might want to run some of the tests in a shared environment...

Alternatively I'll just add expected outcome as a comment, but this want come up in the GUI

lionel-:
It's not really a drag to write if you use snippets.

What do you think @hadley ? Which form do you like?

Description in `test_that()` statements:

```r
test_that("Concentric circles at theta = 1:80", {
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y, group = factor(y))) + geom_line() + coord_polar(),
    "coord-polar-concentric-80"
  )
}

test_that("Concentric circles at theta = 1:80 - 80", {
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 80, group = factor(y))) + geom_line() + coord_polar(),
    "coord-polar-concentric-80-minus-80"
  )
})

test_that("Concentric circles at theta = 1:80 - 40", {
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 40, group = factor(y))) + geom_line() + coord_polar(),
    "coord-polar-concentric-80-minus-40"
  )
})
```

Description in expectations:

```r
test_that("Polar coordinates draws correctly", {
  dat <- data.frame(x = 0:1, y = rep(1:80, each = 2))

  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y, group = factor(y))) + geom_line() + coord_polar(),
    "polar_coord_1",
    "Concentric circles at theta = 1:80"
  )
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 80, group = factor(y))) + geom_line() + coord_polar(),
    "polar_coord_2",
    "Concentric circles at theta = 1:80 - 80"
  )
  vdiffr::expect_doppelganger(
    ggplot(dat, aes(x, y - 40, group = factor(y))) + geom_line() + coord_polar(),
    "polar_coord_3",
    "Concentric circles at theta = 1:80 + 100"
  )
})
```

thomasp85:
Still a problem of sharing state and environment between tests if we go with the current approach...

And snippets makes its user-friendliness reliant of a GUI with snippet support.

I can accommodate and in the end it is of course your call - just my personal observations...

lionel-:
In the end I think the figures should be browsable so their filenames should be informative enough. The numbered filenames are not ideal in that regard.

thomasp85:
Well, the `polar_coord_*`was just an example - they are of course free within the constraints of filename-safe characters. Better naming can be conceived :-)

If you go the current route I'll just add comments in the test files - this was just functionality that I, as a user, would appreciate, but this is subjective of course...

thomasp85:
Despite having resolved all file naming issues and being able to successfully pass all tests on my own system, Travis still errors out on all differ tests. @lionel- do you have any advice on how to troubleshoot travis-specific problems?

lionel-:
I saw that :/

I've just activated Travis on ggstance (which I should have done a long time ago, sorry about that) and the builds are also failing: https://travis-ci.org/lionel-/ggstance/builds/151740400

I'm starting to think that the SVG output is platform-dependent. If that is the case, I'll have to rethink how to test figures, probably using bitmap differentiation with some threshold.

I'm going to try to figure it out in a Linux docker.

thomasp85:
Thanks - I actually think that svg would be the most platform independent format as its a set of instructions rather than a rendering. .eps should be too so this might be an alternative to try out.

Could it be line-ending character or something that messes this up?

thomasp85:
I fear (though I haven't tested) that rendering will differ too much between Windows and Unix systems.

hadley:
@lionel- are you using svglite? It should be platform independent

lionel-:
Yes I am. The problem is that the fonts are rendered differently on Linux and OSX.

lionel-:
Maybe it's not even the same fonts because of defaults, I'll look into it in a little bit.

lionel-:
@hadley I was confused it's not the fonts (it's the png device that does not default to the same font apparently), but the spacing between elements is different. Also although the plots have the same dimensions, the clipping paths are different. Example files:


[diff-svg.zip](https://github.com/hadley/ggplot2/files/415592/diff-svg.zip)

Not sure how to deal with this :/

hadley:
@lionel- hmmmm, that's technically a bug in svglite - it might still be related to different font defaults because svglite has to compute the font metrics in order to correctly support (e.g.) plotmath.

lionel-:
I just submitted an issue at svglite: https://github.com/hadley/svglite/issues/69

hadley:
We should also set up a special theme to use for visual testing to avoid any spurious diffs caused by changes to the theme.

DESCRIPTION
```diff
@@ -218,3 +220,6 @@ Collate:
     'zzz.r'
 VignetteBuilder: knitr
 RoxygenNote: 5.0.1
+svgliteNote: 1.1.0
```
hadley: That's going to trigger `R CMD check` warnings on CRAN

lionel-: it's the wrong PR ;)

PR #1667, opened 2016-07-14
docs: reorder theme_gray in examples so it's clear what it does

tjmahr:
Small but clarifying change. I encountered a user who was confused why `p + theme_gray()` didn't update the plot when the examples were run in succession. 

I reordered examples so it does change the current plot. Also, commented that it is the default, in case the reader jumps right to the examples.


hadley:
Thanks!

PR #1671, opened 2016-07-17
Removes incomplete sentence

### LINKED ISSUE #1654: Dropped sentence in geom_text docs
The documentation for `geom_text` includes the `check_overlap` argument, which reads a bit strange. Looks like an incomplete sentence: 

check_overlap: If ‘TRUE’, text that overlaps previous text in the same
          layer will not be plotted. A quick and dirty way

jabranham:
Fixes #1654


hadley:
Thanks!

PR #1679, opened 2016-07-25
Various fixes to themes

jiho:
Make both the code and the visual aspect of all theme more homogeneous. The general idea was to take theme_grey as the reference for all changes (and since it changed slightly in the previous update of ggplot2, the other themes needed some love too!).

I tried to divide the PR in small standalone commits (too lazy for several pull requests ;-) ). I'll update them as needed so that the show PR can be merged. I also added the rationale for various choices as comments in the code.

Some fixes were clearly needed (default size, etc.). Visually, I think theme_bw in particular needed a review. For the others, the changes are quite subtle.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,8 @@
+
 # ggplot2 2.1.0.9000 
+
+* Themes are more homogeneous visually, and match `theme_grey` better.
+  (@jiho, #1679)
 
 * `position_stack()` and `position_fill()` now sorts the stacking order so it 
   matches the order of the grouping. Use level reordering to alter the stacking 
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -193,27 +193,27 @@ theme_linedraw <- function(base_size = 11, base_family = "") {
 
 #' @export
 #' @rdname ggtheme
-  half_line <- base_size / 2
 theme_light <- function(base_size = 11, base_family = "") {
   # Starts with theme_grey and then modify some parts
   theme_grey(base_size = base_size, base_family = base_family) %+replace%
     theme(
-      axis.ticks        = element_line(colour = "grey70", size = 0.25),
-      legend.key        = element_rect(fill = "white", colour = "grey50", size = 0.25),
+      # white panel with light grey border
       panel.background  = element_rect(fill = "white", colour = NA),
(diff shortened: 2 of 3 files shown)


jiho:
Some before-after shots; current stable on the left, new version after these commits on the right
compare002
compare003
compare004
compare005
compare006
compare007
compare008
compare009

hadley:
Does this connect to #1581 in any way?

I never really use themes, so I'd also be open to suggestions about how to decrease the likelihood of regressions when I merge future PRs.

jiho:
I don't think it relates to #1581 in any way. I think the point of that PR is well made. Once it is merged, it would (marginally) simplify some of the theme specifications but it should not create any regression.

As for regressions for the changes in this PR, it depends what you would call a regression. Some commits here do *change* the visual aspect. I think for the better but some might say any change is a regression at this point. You have the before-after shots to judge which are acceptable. I think the most problematic, potentially, would be `theme_bw`, which is the probably the most used and the one I changed the most. I think the previous version had serious issues (colors, minor grid larger than major grid, etc.).

PS: The merge tool says there are conflicts but I just branched from master a few days ago. I've changed the commits to solve a Travis error; do do so I've had to force push the new commits. Maybe this throws the merge tool astray. Let me know if you want to rebase.

hadley:
Ok, sounds reasonable.  If you rebase, I'll merge.

Katiedaisey:
I'll check more closely later with some edge cases but I see no obvious conflict with #1581. This seems to standardize the defaults while #1581 incorporates values from more general parent elements when intermediate elements are not specified.

jiho:
I've rebased. I'm not sure what that failing check means...

hadley:
Thanks! The failing check is just for unit tests, which don't matter here

jiho:
That was quick, thanks. Do you want me to test how #1728 behaves after this merge?

hadley:
I'll update that branch, but I think it should be largely independent.

krlmlr:
Is there a convenient way to access the old definitions? I need to retrofit a `theme_bw()` plot to the old look, my current workaround is to install ggplot2 2.1.0 :-/

jiho:
You can probably pull the file with the definitions from 2.1.0 and override the new ones by sourcing just this file... unless this was before the theming system changed.

PR #1687, opened 2016-07-29
Fixed bug in geom-violin.r when quantiles requested but data have no range.

DanRuderman:


hadley:
Could you please also add a bullet point to news, and a small unit test (using the simplest possible dataset and `layer_data()`)

DanRuderman:
Do you mean a unit test that is

- particular to this bug (ensuring that it doesn't error out anymore and that the `layer_data()` is correct), or

- one that is more general (e.g. creates a `ggplot()` including `geom_violin()` `draw_quantiles` for some relatively simple data and ensures that the `layer_data()` are correct)? 

Or perhaps something else?

hadley:
Particular to this bug.

DanRuderman:
Since my fix is in `draw_group()` there is no impact on `layer_data()`, but there is on `layer_grob()`. Test that instead?

hadley:
Yes, although testing grobs is hard.

DanRuderman:
How about evaluating the grob just to make sure it no longer errors out and testing the contents of the `layer_data()` for validity?

hadley:
Sounds good

DanRuderman:
All done.

hadley:
Thanks!

PR #1688, opened 2016-07-29
Updates to include count & density to fix the use of geom_hexbin 

### LINKED ISSUE #1608: geom_hex no longer recognizes ..density.. in 2.1.0
In 2.1.0:

```
DF <- data.frame(x=rnorm(1000), y=rnorm(1000))
ggplot(DF) + geom_hex(aes(x=x, y=y, fill=..density..))
```

causes error:

```
Error in eval(expr, envir, enclos) : object 'density' not found
```

But in 2.0.0 (or other older versions), we get (random seeds notwithstanding):

rplot01

mikebirdgeneau:
Updates to include count & density to fix the use of geom_hexbin with ..density.. or ..count.. 

Fixes #1608.


R/hexbin.R
```diff
@@ -31,6 +31,8 @@ hexBinSummarise <- function(x, y, z, binwidth, fun = mean, fun.args = list(), dr
 
   # Convert to data frame
   out <- as.data.frame(hexbin::hcell2xy(hb))
+  out$count = as.vector(hb@count)
```
hadley: Can you please use `<-` instead of `=`?

hadley:
Can you please also add:

* A bullet point to news
* A unit test. Use the simplest possible dataset and get the computed data with `layer_data()`

R/hexbin.R
hadley: Also I'm not sure that the code should go here - it's only needed for hexbin, not the other summary functions.

mikebirdgeneau:
I've made the requested changes to the best of my ability (still new to pull requests and contributions to third party repos). 

- Replaced `=` with `<-`.
- Added bullet in 'news'.
- Added simple unit test for `geom_hex()`
- Noticed that the `value` property retains the same output as `count`, so really only `density` was missing as far as I can tell.

As for your comment regarding if this is the right place for the code, that was why I didn't submit a pull request in the first place on #1608 - I thought there might have been an architectural reason for the change that I wasn't familiar enough with to make the fix.

Happy to help if this is useful, otherwise I'm equally happy if someone else has a better or more elegant way to resolve this. Cheers!

hadley:
Can you move the code to `StatBinhex$compute_group()`?  You'll need to save the results of `hexBinSummarise()` to a variable, and then modify that.

Otherwise looks great!

mikebirdgeneau:
Done. I've moved the code to `StatBinhex$compute_group()`. Thanks for pointing me in the right direction @hadley. If there's anything else on this, just let me know.

------
To confirm this did indeed resolve #1608 (in addition to the successful unit test), I ran the code in #1608 and here's the result, which looks good.
```r
DF <- data.frame(x=rnorm(1000), y=rnorm(1000))
ggplot(DF) + geom_hex(aes(x=x, y=y, fill=..density..))
```
image

NEWS
```diff
@@ -7,6 +7,9 @@ ggplot2 1.0.1
 * Improvements to order of colours and legend display for continuous colour
   scales extracted from colorbrewer palettes by `scale_*_distiller()` (@jiho, 1076)
 
+* Restore functionality for use of `..density..` in
```
hadley: This should go at the top of `NEWS.md` (just underneath the version number heading)

hadley:
Nearly there - thanks for all the hard work!

NEWS
```diff
@@ -2,6 +2,9 @@
 ggplot2 1.0.1
 ----------------------------------------------------------------
 
+* Restore functionality for use of `..density..` in
```
hadley: Still the wrong file - needs to be in `NEWS.md`

hadley:
Thanks!

mikebirdgeneau:
No problem - thanks for your patience! (Appreciate you taking the time to 'teach' when it could be faster to simply resolve yourself.)

PR #1690, opened 2016-08-01
Fix annotate without global data

### LINKED ISSUE #1655: Annotations work only if global aes() have been defined
Hi, 

adding an annotation works only if global aesthetics have been defined in `ggplot()`.
If there are only "local" aesthetics in individual geoms the annotation will not be drawn (no warning though).

```r
library(ggplot2)
library(grid)
df <- data.frame(x = 1:10, y= 1:10)	
anno <- annotation_custom(grid.rect(draw=FALSE), xmin = 7, xmax = 10,ymin = 1, ymax = 2)

## Global aesthetics: works fine
ggplot(df, aes(x = x, y = y)) + geom_point()+ 
		anno

## "Local" aesthetics: does not work
ggplot() + geom_point(data = df, aes(x = x, y = y))+
		anno 
```
selection_065

selection_066

I tried setting `scales` and/or `coord` limits explicitly, but this doesn't make it work either:
		
```r
ggplot() + geom_point(data = df, aes(x = x, y = y))+
		anno +
		scale_x_continuous(limits = c(-10,10)) +
		scale_y_continuous(limits = c(-10,10)) +
		coord_equal(xlim = c(-20,20), ylim = c(-20,20))
```

thomasp85:
This fixes #1655 by adding dummy data to all annotate functions in the way already done for annotate_logticks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,9 @@
 
 * Fix error message of Stats ggprotos when required aesthetics are
   missing.
+  
+* Fix bug that resulted in several annotation_x function not getting drawn when
+  global data was lacking (#1655)
 
 # ggplot2 2.1.0
 
--- a/R/annotation-custom.r
+++ b/R/annotation-custom.r
@@ -42,11 +42,11 @@ NULL
 #'   annotation_custom(grob = g, xmin = 1, xmax = 10, ymin = 8, ymax = 10)
 annotation_custom <- function(grob, xmin = -Inf, xmax = Inf, ymin = -Inf, ymax = Inf) {
   layer(
-    data = NULL,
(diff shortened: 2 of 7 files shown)


hadley:
Could you add a unit test for this?  I think it can probably just test that the annotation layer has an appropriate dummy table.

To make it easier to see this pattern in the future it might be worth defining and using:

```R
# Used in annotations to ensure printed even when no
# global data
dummy_data <- function() data.frame(x = NA)
```

hadley:
LGTM

(please squash the commits when you merge)

thomasp85:
A feeling of unease as a press merge for the first time :-)

Issue #1691, opened 2016-08-02
position_stack() should stack +ive & -ve values separately


hadley:
```R
library(ggplot2)
library(tibble)

df <- frame_data(
  ~x, ~g, ~y,
  1,  1,  1,
  1,  2,  -1,
  1,  3,  1,
  2,  1,  2,
  2,  2,  -3
)

ggplot(df, aes(x, y, fill= factor(g))) + 
  geom_bar(stat = "identity")
```

thomasp85:
There are a lot of case where, even if supported, negative values doesn't make sense when stacking. As I see it it only makes sense for stacked bar charts - can you come with other examples...

I'm asking because I need to know which types of checks to make and warnings to throw, and also where it should be solved. It could be handled directly by `pos_stack()` but this function is also called by `pos_fill()` where it doesn't make sense to have negative values. Alternatively (and easier) the data could be split into two calls to `collide` in `PositionStack$compute_panel()` which would make the functionality specific to `PositionStack` (though it would still be a mess if coupled with `geom_area`

hadley:
Can you come with an example where stacking makes sense apart from bars/areas? I think that's the only case we need to worry about, and it's reasonable for positive values to stack upwards from the y-axis, and negative values to stack downwards.

Your fix seems reasonable to me - I don't think there's any way to have an elegant solution without completely rewriting the stacking (which we don't want to do)

thomasp85:
I cannot - which was why I asked whether you knew of any other use cases. I'll move forward with #1693

smouksassi:
I think we can also stack dots which can be useful in some cases. http://www.lispworks.com/success-stories/inspiredata.html

http://docs.ggplot2.org/current/geom_dotplot.html

thomasp85:
@smouksassi true. Dot plots should mainly be used for count data though, which makes them unsuitable to show negative values.

smouksassi:
I was thinking about a kind of back to back dotplot like this:   https://nzmaths.co.nz/resource/fully-stretched:

hadley:
Dot plots have to use a completely different algorithm so aren't relevant to this discussion.

wmurphyrd:
Here's an example where the current behavior of position_stack for a mixture of positive and negative values produced the desired result: the total extent of the bar aligns with the sum of the values.
[...]

PR #1693, opened 2016-08-03
Sort within x prior to stacking

### LINKED ISSUE #1552: Stacked area charts not stacking
When I try to create a stacked area chart using the following code the segments don't stack, and they occlude each other instead.

```r
library("ggplot2")
library("dplyr")

## Create data
set.seed(40)
dat <- data.frame(x = rep(c(1:10), 3),
                  var = rep(c("a", "b", "c"), 10),
                  y = round(runif(30, 1, 5)))

## Plot
ggplot(dat, aes(x = x, y = y, fill = var)) +
  geom_area(position = "stack")
```

unstacked_area

However if I reorder the rows so all of the `var` values are grouped together, it does create a stacked chart:

```r
dat %>%
  arrange(var, x) %>%
  ggplot(aes(x = x, y = y, fill = var)) +
  geom_area(position = "stack")
```

stacked_area

Is this a bug, or is the ordering of the rows meant to matter like this?

thomasp85:
This fixes #1552. `collide` now sorts the data by x and then by negated group. The negation is performed to ensure that the stacking order follows the default legend order of the fill.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # ggplot2 2.1.0.9000 
 
+* `position_stack()` and `position_fill()` now sorts the stacking order so it 
+  matches the order of the grouping. Use level reordering to alter the stacking 
+  order. The default legend and stacking order is now also in line. The default 
+  look of plots might change because of this (#1552, #1593).
+  
+* `position_stack()` now accepts negative values which will create stacks 
+  extending below the x-axis (#1691)
+
 * Restore functionality for use of `..density..` in 
   `geom_hexbin()` (@mikebirdgeneau, #1688)
 
--- a/R/position-collide.r
+++ b/R/position-collide.r
@@ -26,9 +26,9 @@ collide <- function(data, width = NULL, name, strategy, check.width = TRUE) {
     width <- widths[1]
   }
 
-  # Reorder by x position, relying on stable sort to preserve existing
-  # ordering, which may be by group or order.
-  data <- data[order(data$xmin), ]
+  # Reorder by x position, then on group. Group is reversed so stacking order
+  # follows the default legend order
+  data <- data[order(data$xmin, -data$group), ]
 
   # Check for overlap
   intervals <- as.numeric(t(unique(data[c("xmin", "xmax")])))
--- a/R/position-stack.r
+++ b/R/position-stack.r
@@ -3,6 +3,19 @@
 #' \code{position_fill} additionally standardises each stack to have unit
 #' height.
 #'
+#' @details \code{position_fill} and \code{position_stack} automatically stacks
+#' values so their order follows the decreasing sort order of the fill
+#' aesthetic. This makes sure that the stack order is aligned with the order in
+#' the legend, as long as the scale order has not been changed using the
+#' \code{breaks} argument. This also means that in order to change stacking
+#' order while preserving parity with the legend order it is necessary to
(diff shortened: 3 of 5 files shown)


thomasp85:
To continue @karawoo's example, with this fix you get:

```r
library("ggplot2")

## Create data
set.seed(40)
dat <- data.frame(x = rep(c(1:10), 3),
                  var = rep(c("a", "b", "c"), 10),
                  y = round(runif(30, 1, 5)))

## Plot
ggplot(dat, aes(x = x, y = y, fill = var)) +
  geom_area(position = "stack")
```

stack

It is still not able to match stacking order if the scale has been manually modified:

```r
ggplot(dat, aes(x = x, y = y, fill = var)) +
  geom_area(position = "stack") + 
  scale_fill_manual(breaks = c('c', 'a', 'b'), values = c('forestgreen', 'steelblue', 'firebrick'))
```

  
stack2

thomasp85:
This is very related to #1691 so I'll try to solve it within the same PR with @hadley's blessing.

thomasp85:
Continuing @hadley's example 00eedd1 allows you to mix positive and negative y-values when stacking so you get:

```r
library(ggplot2)
library(tibble)

df <- frame_data(
  ~x, ~g, ~y,
  1,  1,  1,
  1,  2,  -1,
  1,  3,  1,
  2,  1,  2,
  2,  2,  -3
)

ggplot(df, aes(x, y, fill= factor(g))) + 
  geom_bar(stat = "identity")
```

neg-stack

Stacking order is as close to the default legend order as possible, but can get mixed up if one category jumps between negative and positive y-values

hadley:
This looks good. I think it needs a few unit tests (so we don't accidentally break again in the future), and a section in `position_stack()` docs that describes the algorithm. Changing the order of the bars is such a common question that it's probably also worth mentioning in the docs for `geom_bar()`.

hadley:
This looks good. I think it needs a few unit tests (so we don't accidentally break again in the future), and a section in position_stack() docs that describes the algorithm. Changing the order of the bars is such a common question that it's probably also worth mentioning in the docs for geom_bar().

hadley:
Do we need a similar tweak for `position_fill()`?

thomasp85:
I can't really wrap my head around how position fill should behave with negative values...

hadley:
@thomasp85 hmmmm, I was thinking it we'd apply the same principle and fill separately for positive and negative

thomasp85:
that would mean that positive and negative values received different scaling at the same point on the x-axis... And what if there are only negative values for some x? what to draw?

hadley:
LGTM - just needs a news bullet that makes it clear this is a big change (and we'll need to highlight in the blog post)

thomasp85:
@hadley do you want stronger wording/more emphasis in the news?

thomasp85:
I've squashed it and will merge (and close issues if you approve the news

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # ggplot2 2.1.0.9000 
 
+* position_stack and position_fill now sorts the stacking order so it matches
```
hadley: Should be `` `position_stack()` ``

hadley:
Looks good (apart from the small formatting issue).  We'll definitely highlight this in the final release notes, and in the blog post.

thomasp85:
How do you handle roxygenize? once, prior to release, or for every PR?

hadley:
Every PR.

Issue #1669, opened 2016-07-15
Add asymmetrical `expand` argument to continuous and discrete scales


huftis:
This is a feature request for an asymmetrical `expand` argument to the `scale_continuous` and `scale_discrete` scales. A typical use case would be for bar plots that touch the _x_ axis but still have some space above them. For example

```
ggplot(mtcars) + 
  geom_bar(aes(x = cyl))
```

has some space above the bars, but also an annoying spaces below it. Now,

```
ggplot(mtcars) + 
  geom_bar(aes(x = cyl)) +
  scale_y_continuous(expand = c(0, 0))
```

gets rid of the space below the bars, but also the space above the bars. To keep the space above the bars, one has to use a hack with a manually calculated value:

```
ggplot(mtcars) + 
  geom_bar(aes(x = cyl)) +
  scale_y_continuous(expand = c(0, 0)) +
  annotate("blank", x = 6, y = 14.7)
```

(And this doesn’t work well with more complicated plots, e.g. involving facets with `free_y` &c.)

Suggested feature/syntax: Change the `expand` argument from

```
expand = c(a, b)
```

to

```
expand = c(a, b, c, d)
```

where `a` is the multiplier for the lower limit, `b` is the additive term for the lower limit, and `c` and `d` are the corresponding multiplier and additive term for the upper limit. If `c` is missing, use the value of `a`, and if `d` is missing, use the value of `b`. This way, all old code would continue to work.

See http://stackoverflow.com/questions/34623780/asymmetric-expansion-of-limits-ggplot2-2-0 for more information. This post also has suggested code for an asymmetrical `expand` feature, but this does not work for complicated facet grids, nor with `coord_flip()`.

steveharoz:
Have a look at the `expand_limits` function

```
ggplot(mtcars) + 
  geom_bar(aes(x = cyl), width = 1) +
  scale_y_continuous(expand = c(0,0)) +
  expand_limits(y=25)
```

huftis:
@steveharoz `expand_limits()` is a wrapper for `geom_blank()`, but still requires one to manually calculate the limits. This feature request is for an expansion of `expand` to handle separate values for the lower and upper limits, for *automatic* determining the correct limits. For this example, the new syntax would be:

    ggplot(mtcars) +
      geom_bar(aes(x = cyl), width = 1) +
      scale_y_continuous(expand = c(0, 0, 0.05, 0))

huftis:
[...]

PR #1697, opened 2016-08-05
`expand` parameter can now be used selectively

sainathadapa:
Modified the CoordCartesian function so that `coord_cartesian` can also accept a logical vector of length 4 for `expand` parameter. This is useful, especially in facet plots of bar graphs, where one would want to not expand the plot on the bottom side, and instead keep the bar anchored to 0. Also, this modification solves [this question from stackoverflow](https://stackoverflow.com/questions/27028825/ggplot2-force-y-axis-to-start-at-origin-and-float-y-axis-upper-limit).

An example:
`p <- ggplot(mtcars, aes(disp, wt)) + geom_point() +  geom_smooth()`
`print(p)`
original

`p + coord_cartesian(expand = c(T, F, F, T))`
expand

Notice that the expansion factor was applied to the left and top edges of the plot, but not to the other edges.

Since this is an addition, but not a modification of the existing functionality, I don't expect any backward compatibility problems.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # ggplot2 2.1.0.9000 
 
+* `coord_cartesian()`, `coord_flip()`, `coord_fixed()`, and `coord_map()`
+  now also accept a logical vector of length 4 for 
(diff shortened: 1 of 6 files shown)


hadley:
Please see discussion in #1669

PR #1699, opened 2016-08-06
Fixed issue where coord_map() fails when passed 'parameters' argument

tdmcarthur:
This fails:

`
library(ggplot2)

nz <- map_data("nz")
nzmap <- ggplot(nz, aes(x = long, y = lat, group = group)) +
  geom_polygon(fill = "white", colour = "black")
nzmap + coord_map(projection = "lambert", parameters=c(12, 17.5))
`
Error in mapproj::mapproject(x, y, projection = coord$projection, parameters = coord$params,  : 
  (list) object cannot be coerced to type 'double'

I simply changed mproject so that unlist() is run on the 'parameters' argument before passing to mapproject()

### DIFF
--- a/R/coord-map.r
+++ b/R/coord-map.r
@@ -243,7 +243,7 @@ CoordMap <- ggproto("CoordMap", Coord,
 mproject <- function(coord, x, y, orientation) {
   suppressWarnings(mapproj::mapproject(x, y,
     projection = coord$projection,
-    parameters  = coord$params,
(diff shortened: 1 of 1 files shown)


hadley:
I think this should be fixed in the creation of the `coord_map()` — i.e. it should take an explicit `parameters` argument. Do you want to try to make that change?

tdmcarthur:
Fixed via PR #1729

PR #1702, opened 2016-08-09
Cache palette and n in ScaleDiscrete

### LINKED ISSUE #1674: Only generate one warning about shapes
```R
ggplot(data = mpg) + 
  geom_point(mapping = aes(x = displ, y = hwy, shape = class))
```

not two!

thomasp85:
This change saves the palette and the number of breaks in it and retrieves these values on subsequent calls, avoiding multiple warnings pertaining to the palette. Fixes #1674

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -62,6 +62,9 @@
 
 * Class of aesthetic mapping is preserved when adding `aes()` objects. (#1624)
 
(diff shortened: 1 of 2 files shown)


hadley:
LGTM - just add news bullet, and check that actually closes related issue.

Issue #540, opened 2012-05-03
Error in loadNamespace(i, c(lib.loc, .libPaths())) :    there is no package called ‘munsell’


cfranko:
I updated the ggplot2 package in RStudio from the CRAN repo and loading the library gives the following error.  This is the first time I've had an issue loading the library.

Error in loadNamespace(i, c(lib.loc, .libPaths())) : 
  there is no package called ‘munsell’
Error: package/namespace load failed for ‘ggplot2’

shwetamrit:
Hi,i am also not able to load swirl,plz  help me.

 install.packages("swirl", dependencies=TRUE)
Installing package into ‘C:/Users/Shweta/Documents/R/win-library/3.2’
(as ‘lib’ is unspecified)
trying URL 'http://ftp.iitm.ac.in/cran/bin/windows/contrib/3.2/swirl_2.2.21.zip'
Content type 'application/zip' length 132378 bytes (129 KB)
downloaded 129 KB

package ‘swirl’ successfully unpacked and MD5 sums checked

The downloaded binary packages are in
        C:\Users\Shweta\AppData\Local\Temp\Rtmpsngb6l\downloaded_packages
> library(swirl)
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called ‘R6’
Error: package or namespace load failed for ‘swirl’
> install.packages("evaluate")
Installing package into ‘C:/Users/Shweta/Documents/R/win-library/3.2’
(as ‘lib’ is unspecified)

hadley:
@shwetamrit please ask for help at the swirl site

msusplugas:
Is it possible to add the package munsell as a dependency in the DESCRIPTION file? It's a dependency because we need it to use ggplot2. Thanks

thomasp85:
Munsell is not a dependency of ggplot2 and should therefore not be placed under Depends. The package is imported by scales which is in turn imported by ggplot2. Unless something has gone wrong during installation all nested imports should be installed when installing ggplot2.

sahanb:
I am trying to load the library "DiagrammeR". But gettinng the below error.
install.packages("DiagrammeR", dependencies = TRUE)
library('DiagrammeR')
Error in loadNamespace(i, c(lib.loc, .libPaths()), versionCheck = vI[[i]]) : 
  there is no package called 噌ML・Error: package or namespace load failed for 船iagrammeR・

[...]

PR #1704, opened 2016-08-10
Add munsell and colorspace as  dependencies

msusplugas:
Add munsell and colorspace as dependency of ggplot2 in the DESCRIPTION file.
This is required to use the package ggplot2.


thomasp85:
See comment on #540

Issue #1589, opened 2016-03-22
scale_[xy]_discrete() causes removal of breaks/labels


gfiumara:
# Summary

Using `scale_[xy]_discrete()`, regardless of arguments, causes axis labels and breaks to disappear without warning.

# Example

```
library(ggplot2)
xyDF <- data.frame(x = 1:10, y = 1:10)
ggplot(data = xyDF, mapping = aes(x = x, y = y)) +
    geom_line() +
    geom_point() +
    scale_x_discrete()

```
# Expected results

Plot showing breaks and labels at 1:10 on the X axis.

![ggplot2_101](https://cloud.githubusercontent.com/assets/419642/13954722/5bc532bc-f018-11e5-8b54-7d513eacc60b.png)
_Created with ggplot2 1.0.1, as I cannot reproduce with 2.1.0._

# Actual results

Plot with no breaks or labels on the X axis (as if `scale_x_discrete(labels = NULL, breaks = NULL)` was added to the plot instead of the default `waiver()`).
![ggplot2_2](https://cloud.githubusercontent.com/assets/419642/13954713/5285d3a0-f018-11e5-9926-9c6294edbce0.png)

# Notes

* Regression since ggplot2 <s>1.0.1 (last major version I've been using)</s> fb3931c4e925272ddba607b81ada998df630bc10.
* Present in devtools and CRAN version as of 22 March 2016.

gfiumara:
Appears to have broken in commit 89893c71ea575e198240e008ba2f42c9cbb7c6b7 (#1542), confirmed by running sample with different `devtools` installs:

```
# Works
devtools::install_github("hadley/ggplot2", ref="fb3931c4e925272ddba607b81ada998df630bc10")

# Broken in subsequent commit
devtools::install_github("hadley/ggplot2", ref="89893c71ea575e198240e008ba2f42c9cbb7c6b7")
```

thomasp85:
This is an by design, as conversion from continuous to discrete data is not well-defined. It is easy to end up with thousands of breaks so it is up to the user to either provide the breaks explicitly or convert their data to a discrete format

geotheory:
Is this the same issue:

```
dat = data.frame(x = 1:4, y = ordered(c(4,'>5',1,1), levels = c(1:5, '>5')), grp = 1)
ggplot(dat, aes(x, y, group=grp)) + geom_step(size=1) + 
  scale_y_discrete(breaks = levels(dat$y), labels = levels(dat$y))
```
[...]

Issue #1542, opened 2016-02-10
Adding scale_*_discrete() to date variable can crash R


cpsievert:
With **scales** 0.3, this code will error:

```r
x <- c(Sys.Date(), Sys.Date() - 1e4)
y <- c(3, 4)
d <- data.frame(x, y)
library(ggplot2)
ggplot(d, aes(x, y)) + geom_line() + scale_x_discrete()
#> Error: Discrete value supplied to continuous scale
```

Running this code using the latest dev version of **scales** doesn't result in error, and ggplot tries to draw a plot with 1e4 ticks. I know this is a cornercase, but I would prefer an error since it can easily wreak havoc on an R session, especially with date-times.

cpsievert:
Here is the relevant change -- https://github.com/hadley/scales/commit/cd0395de74af429a7151b7e7cedfcdb72fa96215

hadley:
Are you sure that's the change? It seems weird that a check in continuous scales would break discrete scales

cpsievert:
Yes, it appears both `scale_x_discrete()`/`scale_y_discrete()` use a continuous range. I'm not sure whether or not that is intentional, but it does seem odd:

```r
> scale_x_discrete
function (..., expand = waiver()) 
{
    sc <- discrete_scale(c("x", "xmin", "xmax", "xend"), "position_d", 
        identity, ..., expand = expand, guide = "none")
    sc$super <- ScaleDiscretePosition
    class(sc) <- class(ScaleDiscretePosition)
    sc$range_c <- continuous_range()
    sc
}
```

Just for demonstrate that the change I refer to above is relevant, you can install https://github.com/hadley/scales/pull/70 to see that the example now throws an error.

hadley:
Ah, that makes sense.  But I'd rather fix it by handing the poor performance with many levels

hadley:
Hmmm, something is seriously messed up with code that handles the combination of discrete and continuous ranges. I need to think through exactly how this should behave - the reason why discrete scale also have a continuous component is so that this sort of plot can work:

```R
ggplot(mtcars, aes(factor(cyl), mpg)) + geom_jitter()
```

PR #1708, opened 2016-08-12
Better handling of numeric levels in scale_discrete

### LINKED ISSUE #1638: Discrete x/y scales reserve space for unused limits
For example, from the [docs](http://docs.ggplot2.org/current/scale_discrete.html)

```
d <- ggplot(subset(diamonds, carat > 1), aes(cut, clarity)) +
      geom_jitter()
d + scale_x_discrete(limits=c("Fair","Ideal"))
```

In the [0.9.3.1 docs](http://docs.ggplot2.org/0.9.3.1/scale_discrete.html), it looks like this:

thomasp85:
Fixes #1638 and #1589


hadley:
Can you please add two simple unit tests (one for each issue) using `layer_scales()` to get the scale object, and then testing its range and breaks.

thomasp85:
Are these unit tests fine?

tests/testthat/test-scale-discrete.R
```diff
@@ -16,3 +16,19 @@ test_that("discrete ranges also encompas continuous values", {
   expect_equal(x_range(base + geom_point(aes(x1)) + geom_point(aes(x2))), c(0, 4))
 })
 
+test_that("discrete scale shrinks to range when setting limits", {
+  p <- ggplot(subset(diamonds, carat > 1), aes(cut, clarity)) +
```
hadley: Can you use made up data?

tests/testthat/test-scale-discrete.R
```diff
@@ -16,3 +16,19 @@ test_that("discrete ranges also encompas continuous values", {
...
+    geom_jitter() + scale_x_discrete(limits=c("Fair","Ideal"))
+
+  expect_equal(layer_scales(p)$x$dimension(c(0, 1)), c(0, 3))
+})
+
+test_that("limits are returned correctly when using continuous input", {
+  xyDF <- data.frame(x = 1:10, y = 1:10)
```
hadley: Just call it `df`

tests/testthat/test-scale-discrete.R
```diff
@@ -16,3 +16,19 @@ test_that("discrete ranges also encompas continuous values", {
...
+  p <- ggplot(data = xyDF, mapping = aes(x = x, y = y)) +
+    geom_line() +
+    geom_point() +
+    scale_x_discrete()
+
+  expect_equal(layer_scales(p)$x$get_limits(), seq_len(10))
```
hadley: Also check that we don't get a huge number of breaks?

thomasp85: Calculations of breaks are not really affected by this... If you plot a huge continuous range on a discrete scale it will create a lot of breaks...

tests/testthat/test-scale-discrete.R
thomasp85: Sloppy copy-paste :-)

tests/testthat/test-scale-discrete.R
hadley: Wasn't that the motivating case for this whole chain of changes? i.e. this would re-break #1542

thomasp85: Yeah(-ish), the git history was not completely transparent but it was partly in response to #1542. Problem was that it broke everything else :-)

As far as I recall we decided to rollback this change as it was not obvious how a discrete scale should behave in the face of an enormous number of levels...

I would suggest we fix this be adding a `max-breaks` to ScaleDiscrete instead, as this would make it more transparent what goes on inside the scale...

hadley:
I think we definitely need to handle this case – i.e. when a scale has been trained with continuous data, but not discrete data, we should not generate any levels by default. I don't think that will break any existing behaviour.

thomasp85:
As in "any breaks at all"? wasn't this the behaviour that get marked as a bug in #1589?

hadley:
Ah, in that case #1589 is not a bug, but a deliberate design decision.

thomasp85:
Oh - that's kind of an important observation :-)

thomasp85:
Lets see if I can untangle the fix for #1638 from this then

hadley:
And maybe we should have a warning for #1589 — i.e. something like "Discrete scale used, but only given continuous data."?

thomasp85:
yeah, I think there is definitely a chance of confusion with this design

hadley:
Yes, but I'm not sure why you're applying a discrete scale to continuous data!

thomasp85:
No idea either, but with 27k downloads/month it is sure that it will get attempted at a regular basis

thomasp85:
I've rolled back the changes to `get_limits()` so this PR now only addresses #1638. I'll close #1589 as "not-a-bug"

@hadley can you have a quick look through and comment

NEWS.md
```diff
@@ -77,6 +77,9 @@
 
 * A warning is now issued when a scale transformation introduces infinite 
   values in a scale (#1696)
+  
+* Fixed bug where space for dropped levels in scale_discrete would be preserved 
+  (#1638)
```
hadley: And mention the `expand` change?

hadley:
Apart from that small tweak to NEWS, LGTM

thomasp85:
I've added a warning mechanism for #1589 as we discussed - please review that. It is caching whether it has already warned inside the scale object itself which is not ideal, but it was the only way I could avoid replicated warnings

R/scale-discrete-.r
```diff
@@ -132,6 +134,14 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
     }
   },
 
+  get_breaks = function(self, limits = self$get_limits()) {
+    if (self$warn && length(self$get_limits()) == 0) {
```
hadley: If you use `warn`, I think you'd need to reset it in the `clone()` method below.

What if in `Scale$train_df()` we called `train_init()` and `train_final()` methods? (or begin/end, setup/teardown, start/finish) Then you could call this in `train_finalise()`. (I'm not sure what you'd use `train_init()` for, but it makes sense for symmetry)

hadley: Having a function that is called between training and mapping seems most important — we could also use it to cache the palette for discrete scales.  Given that we need for two scales, it suggests it might be useful in other places.  Let's call it `train_complete()`.

thomasp85: Should we ignore `train_start()` then and only go for a `complete` hook?

hadley: That's fine but there's something nice about the summery of having both

thomasp85: Hmm... there are some problems with our clever plan. The obvious place to put these would be at the beginning and end of `train_df`, but this method is only called on non-positional scales. For positional scales `train` is called directly - further, it is called twice and if there are multiple scales due to facetting it is called for each of them... Not saying we shouldn't add the methods, but it doesn't solve the problem we are trying to solve...

PR #1709, opened 2016-08-12
Expand annotate aesthetics

### LINKED ISSUE #1609: annotate() won't expand when position aesthetics are length 1
Min reprex:
```
ggplot() + 
  annotate("text", x = 1, y = 1, label = c("a", "b"))
# Error: Aesthetics must be either length 1 or the same as the data (1): label
```
I am not sure whether it is intentional or not, but it seems the test for lengths only includes the position vectors, when it is reasonable that non position vectors may vary. (Also the error message can be a bit opaque.)

I encountered this issue when debugging some code where the distinction between position and non-position vectors in this test wasn't clear. But a possible genuine application could be a compass rose:

`ggplot() + annotate("text", x = 0, y = 0, label = c("N", "S", "E", "W"), hjust = c(0.5, 0.5, 0, 1), vjust = c(0, 1, 0.5, 0.5))`

 I tried to fork the repo and change the `unequal` definition in `annotation.r`, but it seems this applies to `check_aesthetics` too. I didn't want to change too much without raising the issue.

thomasp85:
Fixes #1609

Currently the annotate layer data is only based on positional ”aesthetics”. This PR includes all additional aesthetics given through … in the data.

### DIFF
--- a/R/annotation.r
+++ b/R/annotation.r
@@ -54,7 +54,7 @@ annotate <- function(geom, x = NULL, y = NULL, xmin = NULL, xmax = NULL,
     stop("Unequal parameter lengths: ", details, call. = FALSE)
   }
 
-  data <- data.frame(position)
+  data <- data.frame(aesthetics)
   layer(
     geom = geom,
     params = list(


hadley:
I think we should test that the length of each aesthetic is either one, or the same as same the `nrows(data)`

thomasp85:
I actually think that this is the behaviour in master. This PR was made in order to address #1609, but we kind of agreed that this should not be addressed, right?

Do we need better error messages or should we just close this PR and call it a day?

hadley:
Yeah, I think you're right. We can close this one.

PR #1713, opened 2016-08-16
Violin outliers

DanRuderman:
Initial version of outlier rendering for `geom_violin`.


hadley:
This feels like quite a big change to me, so I think it should be a different geom. And I'm not sure it belongs in ggplot2 (especially given how so ggplot2 changes). Maybe a better home would be https://github.com/thomasp85/ggforce or https://github.com/hrbrmstr/ggalt?

PR #1719, opened 2016-08-17
expanding overall package documentation

AmeliaMN:
Given 9d47ba7aacd5181d3ed0c7fcfd3c5adaefe9c9dd I'm guessing this is going to get rejected, but I like to be able to ? on any package. This would allow me to ?ggplot2.


hadley:
I'm not opposed to the idea in general, but I don't think that points to a particularly use set of docs.  Maybe better to link to r4ds? And I don't think the comparison to lattice helps most current R users.

AmeliaMN:
Honestly, for my use case I don't really care what it says 😛  --I use ?ggplot2 to quickly get to the link where I can see all the function names. (Is there a better way to get there?) But, I'm happy to change some of the wording. Just grabbed existing stuff out of DESCRIPTION.

PR #1724, opened 2016-08-22
Fix legend.justification

### LINKED ISSUE #1716: Control justification of legends
I'm not sure why there's currently no way to do this. Is this what `legend.box.just` is supposed to control?

I think the default would look nicer if the legends were left/top aligned to the panel area.

thomasp85:
This PR fixes #1716 by stretching the viewport reserved for legend drawing.

### DIFF
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -215,7 +215,8 @@ ggplot_gtable <- function(data) {
     } else {
       # x and y are adjusted using justification of legend box (i.e., theme$legend.justification)
       legend_box <- editGrob(legend_box,
-        vp = viewport(x = xjust, y = yjust, just = c(xjust, yjust)))
+        vp = viewport(x = xjust, y = yjust, just = c(xjust, yjust),
+          height = unit(1, "null"), width = unit(1, "null")))
     }
   }
 
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -216,7 +216,7 @@ ggplot_gtable <- function(data) {
       # x and y are adjusted using justification of legend box (i.e., theme$legend.justification)
       legend_box <- editGrob(legend_box,
         vp = viewport(x = xjust, y = yjust, just = c(xjust, yjust),
-          height = unit(1, "null"), width = unit(1, "null")))
+          height = unit(1, "null")+legend_height, width = unit(1, "null")+legend_width))
     }
   }
 


hadley:
That's all that's needed???

Can you show a couple of examples?

thomasp85:
I'm that good :-)

... was what I thought I would say until I tested it out - the alignment is related to the centre of the legend gtable (i.e. the center is flush with the top of the panel). I'll find a way to fix this and then add some examples as well

thomasp85:
Hmm... we might want a `legend.anchor` that determines where the justification is measured from...

thomasp85:
This works (though I have a hard time understanding why it works for all justification values 😖) <- thats apparently a confused smiley


Examples:

```r
p <- ggplot(mtcars) + geom_point(aes(x = cyl, y = mpg, colour = gear, size = hp))

p
```
rplot01

```r
p + theme(legend.justification = 'top')
```
rplot02

```r
p + theme(legend.justification = 'bottom')
```
rplot03

```r
p + theme(legend.justification = 'left', legend.position = 'bottom', legend.box.just = 'left')
```
rplot04

hadley:
Nice!  

What do you think about making top/left alignment the default?

R/plot-build.r
```diff
@@ -215,7 +215,8 @@ ggplot_gtable <- function(data) {
     } else {
       # x and y are adjusted using justification of legend box (i.e., theme$legend.justification)
       legend_box <- editGrob(legend_box,
-        vp = viewport(x = xjust, y = yjust, just = c(xjust, yjust)))
+        vp = viewport(x = xjust, y = yjust, just = c(xjust, yjust),
+          height = unit(1, "null")+legend_height, width = unit(1, "null")+legend_width))
```
hadley: Needs spaces around `+`

hadley:
Can you please add a bullet to NEWS, and update the theme docs to indicate that it also adjusts the justification outside of the plot?

(Oh and maybe check it doesn't break when `legend.position` is a two-element numeric vector)

thomasp85:
The current "fad" is to align everything against the borders of the plotting area so this would fall in line with that. Whether it is more visually pleasing etc. is beyond me but it would result in less "rearrangement" when additional guides are added...

What are your thoughts on this vs approach two?

hadley:
I think it's more aesthetically pleasing to align to borders (I can try and explain why when we talk), but it would also require some tweaking so that the legend padding only occurs between legends, not on the end.

I don't remember enough to know which is better, so I'd be inclined to go with @baptiste's advice.

baptiste:
BTW, could all `theme()` arguments be explicitly listed in the function's signature? It would be immensely helpful to have tab completion...

hadley:
@baptiste I'd accept a PR that implemented that (you'd still need ... as a catch all)

thomasp85:
I personally think that version 2 is cognitively easier to reason about (though the results are equivalent) at the expense of creating a slightly more complex gtable.

With that version we can definitely say that intermediary justification (e.g. 0.3) controls the relative size of the top and bottom margin...

thomasp85:
Both versions works fine with numeric justification btw

thomasp85:
As I discus in #1716 I think we should separate the margins around the complete legend area from the margins around each guide so it is possible to get the legend completely flush with the top without needing to give up spacing between guides...

It's a quick fix - should I make a separate PR or lump it in together with either this or #1725?

`legend.box.margin` would be the natural name I think, but I do believe that `legend.box.just` and `legend.justification` in a perfect world should be switched in meaning...

hadley:
Let's go with the other approach. And I think you might as well incorporate the margin stuff, which now includes #1727

PR #1725, opened 2016-08-23
Alternative implementation for justification of legends

### LINKED ISSUE #1716: Control justification of legends

thomasp85:
Fixes #1716. Alternative approach to #1724


R/plot-build.r
```diff
@@ -216,6 +216,10 @@ ggplot_gtable <- function(data) {
       # x and y are adjusted using justification of legend box (i.e., theme$legend.justification)
       legend_box <- editGrob(legend_box,
         vp = viewport(x = xjust, y = yjust, just = c(xjust, yjust)))
+      legend_box <- gtable_add_rows(legend_box, unit(yjust, 'null'))
```
hadley: Are `xjust` and `yjust` always numeric at this point?

thomasp85: Yes - `valid.just` takes care of that

thomasp85:
Now you can do:

```r
p + theme(legend.justification = 'top', legend.box.margin = margin(1,1,1,1, 'mm'),
  legend.box.background = element_rect())
```

rplot03

Possible additions at your discretion: 

Margin outside of legend box/distance to plot area should be user set-able so the background does not stack up too closely with the plot area

Margins around legends are currently hard-coded to 1.5mm (as discussed legend.margin sets spacing). This means that it is not possible to have the title completely flush with the plot top. Should this be user set-able?

hadley:
The margins around legends probably should have a theme setting for completeness.

thomasp85:
I really think we should make a legend.spacing theme option for the current legend.margin and repurpose legend.margin for this - I think this is what most user expect happens anyway...

hadley:
Ok - we could have a fallback where if `legend.margin` is set to a unit we issue a deprecation warning, and assign it to `legend.spacing`

thomasp85:
I'll cook something up - should we deprecate panel.margin in favour of panel.spacing in the same go?

hadley:
Sure, sounds good.  

How about we merge this first, and then do #1744?

thomasp85:
Good idea...

Also I would consider changing the defaults to drawing the `legend.box.background` as white, rather than `legend.background` (which would become `element_blank`). Thoughts?

thomasp85:
Motivation is that it seems cleaner to have a box around all legends rather than multiple boxes of different width/height

thomasp85:
I've just "discovered" (thought about) that the alignment will be related to the span of panels. That means that strips and axes are disregarded and that legend can never be flush with the top of potential strips - how do you feel about that? And if it needs to be changed, should strips and axes be regarded differently?

hadley:
In the default theme, the biggest impact of `legend.box.background` is when the legend is drawn in the plot. So the question is: do you want one background beneath all the legends, or a individual background?  It doesn't really matter to me — I think individual backgrounds is probably slightly more aesthetically pleasing.

I think the legend box should span the complete panel region including panels, axes, and strips.

thomasp85:
Ok, we'll keep the defaults as-is. I think it might look strange when the panel are aligned with the bottom of the axes rather than the bottom of the panels as the axes have a less defined bounding box... don't you?

hadley:
Hmmm, that's true. Maybe it should only include strips and panels then?

thomasp85:
I'll share some examples when done - then we can decide...

thomasp85:
I've added a last new theme option which is legend.box.spacing which governs the distance between the plot area and the legend box.

Also I've changed the defaults so that all margins are 0 so that things align nicely if you want it on top/bottom - this means it needs to be tweaked if you want to draw the background, but I think this is the least common scenario...

Will push once the deprecations are flagged...

thomasp85:
Examples:

```r
p <- ggplot(mtcars) + geom_point(aes(x = cyl, y = mpg, colour = gear, size = hp)) + facet_grid(.~gear)
p
```
rplot
```r
p + theme(legend.justification = "top")
```
rplot01
```r
p + theme(legend.justification = "bottom")
```
rplot02
```r
p + theme(legend.justification = "top", legend.box.spacing = unit(1, "cm"))
```
rplot03
```r
p + theme(legend.justification = "top", 
  legend.box.margin = margin(0.5, 0.5, 0.5, 0.5, "cm"), 
  legend.box.background = element_rect(fill = "grey90"), 
  legend.margin = margin(0.2, 0.2, 0.2, 0.2, "cm"), 
  legend.background = element_rect(colour = 'black'))
```
rplot04

R/theme-defaults.r
```diff
@@ -93,7 +93,10 @@ theme_grey <- function(base_size = 11, base_family = "") {
                          ),
 
     legend.background =  element_rect(colour = NA),
-    legend.margin =      unit(0.2, "cm"),
+    legend.spacing =     unit(0.4, "cm"),
+    legend.hspacing =    NULL,
```
hadley: Shouldn't that be `legend.spacing.x` and `legend.spacing.y`?

thomasp85: That is up to you - the legend.margin used this approach so I just ported it, but for consistency we should probably change it...

hadley: Yeah, better to  be consistent

hadley:
LGTM

thomasp85:
I'll regret asking but... unit tests?

hadley:
They really need to be visual. I'd recommend at least writing down the most important combinations to test so we can reintegrate when vdiffr is ready.

thomasp85:
Should I change default legend position while I'm at it, or do you want to fiddle with the defaults yourself at some point prior to next release?

hadley:
I'll do the fiddling later.

Issue #1714, opened 2016-08-16
geom_violin quantile lines respond to alpha


mnbram:
The lines drawn by draw_quantiles in geom_violin are affected by the alpha parameter, so they cannot be included along with a transparent fill. Maybe there is a reason, but it seems to me that they should behave like any other lines.

Simple example:
```
ggplot(mtcars, aes(factor(cyl), mpg)) +
  geom_violin(draw_quantiles = 0.5, alpha = 0)
```

mnbram:
I have a one line fix for this, but as it changes default behavior and I'm not exactly an expert software developer, I'm hesitant to submit it as a pull request. Should I do so anyway?

thomasp85:
No harm in trying - we'll guide you through any quirks

PR #1726, opened 2016-08-23
Don't use alpha for violin quantiles

mnbram:
See #1714

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -78,6 +78,9 @@
 * A warning is now issued when a scale transformation introduces infinite 
   values in a scale (#1696)
 
+* Quantile lines in geom_violin() are no longer affected by the alpha aesthetic
+  (@mnbram, #1714)
+
 # ggplot2 2.1.0
 
 ## New features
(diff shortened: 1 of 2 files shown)


thomasp85:
This seems reasonable to me - @hadley can you comment?

It will need a news bullet once Hadley approves. I don't think unit tests makes sense for this

hadley:
Seems reasonable to me, but I'd like to see a pair of before/after images in this PR.

mnbram:
**Before:**
before

**After:**
after

hadley:
LGTM

thomasp85:
@mnbram can you add a news bullet?

mnbram:
@thomasp85 Sorry, I'm new to this and I don't know what that means. Do you just want a brief summary of the change, like: "Quantile lines in geom_violin() are no longer affected by the alpha aesthetic"?

thomasp85:
Yep - something like that. And then mention the issue number

mnbram:
OK. I now see from looking at other pull requests that this should go in NEWS.md with a new commit, correct? At the bottom of the ggplot2 2.1.0.9000 section?

thomasp85:
Precisely

mnbram:
Sorry, I appear to have mangled things while trying to update NEWS.md. I'm working on fixing it, but obviously I am not a git expert.

hadley:
Thanks!

Issue #1727, opened 2016-08-23
Margin tweaks for legends


hadley:
Currently you can only control the overall margin of the legend, but ideally you'd be able to control each of t/r/b/l separately.  Then the defaults need to be adjusted so that you can easily align a legend to either the top-right or the bottom-left of the plotting area.

thomasp85:
My suggestion would be to go with the CSS approach and let users specify either one, two, or four values. Then we could have a `valid.margin()` function that automatically expands the unit vector.

Don't know the preferred approach to adapting the defaults to the user-set justification?

thomasp85:
This approach to margins should of course be applied to all margin specifications. That is, `panel.margin`, `plot.margin`, and the `margin` argument in `element_text()` should be changed as well.

Come to think of it - `panel.margin` should really be called `panel.spacing` - the same holds for `legend.margin`

I'll hold off on anything before you comment on these thoughts...

thomasp85:
Actually the `margin` function would just need a little tweaking to support the functionality in `element_text` - And it could then be applied for the other margins as well

hadley:
Right — the way to do this is to implement more `margin()` helpers. I think changing how `margin()` works is not great at this point (I did deliberately choose the current behaviour of defaulting to 0), but we could certainly add another function.

I agree that `panel.margin` and `legend.margin` would be better off as `panel.spacing` and `legend.spacing`, but I think changing them now would be a lot of work.  I think we should just adjust the docs to clarify that these are _internal_ margins only.

hadley:
1. Existing behaviour of `legend.margin` needs to become like `panel.margin` and only control the spacing between multiple legends

1. That makes `legend.margin` and `panel.margin` somewhat confusing because they really should be `legend.spacing` and `panel.spacing`. We could leave as is and make clear in documentation, or change names with code in `theme()` to work with either old or new specification.

1. We'll also need a new `legend.box.margin` that controls the margin between the legend box and the panel.

1. We could consider adding `legend.box.background` but that would require an additional margin theme setting so you could independently control the margin inside and outside the box.

assaron:
Hi,

Is it possible to keep `panel.margin` option too, but with throwing a deprecation warning? I have the package in Bioconductor that uses this option and I can't put dependence on the >=2.2.0 ggplot verision, as it's not yet realeased, and can't just ignore this, as it will break my package soon.

thomasp85:
[...]

PR #1728, opened 2016-08-23
Add descenders to heights

### LINKED ISSUE #1712: Margin tweaks
* Need slightly more space between axes and titles
* Need slightly less space between legend and plot area

hadley:
And increase margins slightly. Fixes #1712 

@baptiste, @thomasp85, @jiho – can you please take a look? I always forget the consequences of tweaking theme defaults.

```R
library(ggplot2)
df <- data.frame(x = 1:9, y = 1:9)

ggplot(df, aes(x, y)) + 
  geom_point() + 
  labs(
    y = "yyy",
    title = "This is title yyyyyy",
    subtitle = "This is a subtitle yyyyyy"
  )
```

Before: 
before

After: 
after

The difference is subtle, but an improvement.

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -85,13 +85,13 @@ theme_grey <- function(base_size = 11, base_family = "") {
     axis.ticks =         element_line(colour = "grey20"),
     axis.ticks.length =  unit(half_line / 2, "pt"),
     axis.title.x =       element_text(
-                           margin = margin(t = half_line, b = half_line / 2),
+                           margin = margin(t = half_line),
                            vjust = 1
                          ),
     axis.title.y =       element_text(
                            angle = 90,
-                           margin = margin(r = half_line, l = half_line / 2),
-                           vjust = 1
+                           margin = margin(r = half_line),
+                           vjust = 1,
                          ),
 
     legend.background =  element_rect(colour = NA),
@@ -181,12 +181,12 @@ theme_linedraw <- function(base_size = 11, base_family = "") {
       axis.ticks       = element_line(colour = "black", size = 0.25),
       # NB: match the *visual* thickness of axis ticks to the panel border
       #     0.5 clipped looks like 0.25
-      
+
       # pure black panel border and grid lines, but thinner
       panel.border     = element_rect(fill = NA, colour = "black", size = 0.5),
       panel.grid.major = element_line(colour = "black", size = 0.05),
       panel.grid.minor = element_line(colour = "black", size = 0.025),
-      
+
       # strips with black background and white text
       strip.background = element_rect(fill = "black"),
       strip.text       = element_text(colour = "white", size = rel(0.8))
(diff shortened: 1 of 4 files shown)


jiho:
I like the change.

With descenders taken into account, I think the extra space may not be needed (I prefer more room for the data and less for the legends).
In particular, I would put the left margin of `axis.title.y` and the bottom margin of `axis.title.x` to 0. This way, when setting `plot.margin` to 0, the titles are flush with the border of the plot, like the panel background and the main title are. I usually want this for plots inserted within a text, so that the left of the y axis title aligns exactly with the left of the paragraphs and the x axis falls on the baseline if the height of the plot is correctly computed.
Regarding the `0.8 * half_line` vs. `half_line` for the other margin I have no strong opinion.

This is the current with branch `descenders`

and this has the outer margins at 0

and this would be with outer margins at 0 and `plot.margin` at 0 (with titles with ascenders and descenders):

(open the images to better see the position with respect to the margin).

hadley:
The top of the `T` gets cut off in the last plot, but I could certainly dropping the outer padding to something nominal like 0.1.

R/theme-defaults.r
```diff
@@ -85,13 +85,13 @@ theme_grey <- function(base_size = 11, base_family = "") {
...
-                           margin = margin(t = half_line, b = half_line / 2),
+                           margin = margin(t = half_line),
                            vjust = 1
                          ),
     axis.title.y =       element_text(
                            angle = 90,
-                           margin = margin(r = half_line, l = half_line / 2),
-                           vjust = 1
+                           margin = margin(r = half_line),
+                           vjust = 1,
```
jiho: I'm guessing the added comma is problematic here. It should create an error.

jiho:
Indeed, the top of capitals or ascenders and the bottom of descenders are slightly cut off (although that depends on the rendering software; when I import the PDF into Inkscape and re-export it, the clipping of letters is not as pronounced). And indeed, I think the solution is to set `plot.margin` at 0.1 instead of 0, rather than add outer margins the to various outer plot elements (because you would need to add it to the title but also to the panel if there is no title etc.).

So basically, the changes look perfect to me now.

jiho:
There was a way at some point to change the position of the axes (put the x axis on top, the y axis on the right, etc.). See https://groups.google.com/forum/?fromgroups=#!topic/ggplot2-dev/JkJU5CLBkQw

Is that not relevant anymore? If such tweaks exist then, ideally, there should be a way in the themes to specify "outer margin" instead of left/right/top/bottom margin. The easy alternative of course would be to make margins symmetrical, but I would be less pleasing visually as explained above.

thomasp85:
This is not currently possible, but depending on the complexity it might be included in the facet rewrite (with @hadley's blessing of course)

with regards to margins, see my comment on #1727 - optimally this proposal should be implemented for all margin settings and would allow for symmetric margin setting

hadley:
I think the theming system is already sufficiently complicated without having to additionally distinguish between inner/outer. The unfortunate reality is that if you have a theme where the labels are in a different place you will have to tweak margin settings in numerous locations. I don't think there is a good solution for this apart from implementing a full constraint-based layout system.

PR #1729, opened 2016-08-24
Add explicit 'parameters' arg to coord_map() in response to PR #1699

tdmcarthur:
I replaced the ... argument in coord_map() with parameters = NULL and added a new example in its help file for a projection that needs a 'parameters' argument. This fixes the bug mentioned in PR #1699.

In the previous version of coord_map(), the ... argument only served to pass the 'parameters' argument to mapproj::mapproject(), so it can be safely removed. I also added documentation about the new 'parameters' argument in coord_map().

P.S. I've now realized the proper workflow would have been to modify the original pull request at https://github.com/hadley/ggplot2/pull/1699 , but it's too late now.

### DIFF
--- a/R/coord-map.r
+++ b/R/coord-map.r
@@ -21,9 +21,12 @@
 #' @export
 #' @param projection projection to use, see
 #'    \code{\link[mapproj]{mapproject}} for list
-#' @param ... other arguments passed on to mapproject. Ignored if the \code{parameters} argument is present.
-#' @param parameters optional numeric vector of parameters for use with the projection argument. This argument is optional only in the sense that certain projections do not require additional parameters. Passed to
-#'   \code{\link[mapproj]{mapproject}}
+#' @param ... other arguments passed on to \code{\link[mapproj]{mapproject}}.
+#' Ignored if the \code{parameters} argument is present.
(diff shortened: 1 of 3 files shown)


hadley:
Can you fix the examples and add a bullet point to NEWS please?

tdmcarthur:
Instead of modifying the original examples, I made coord_map() backwards-compatible by reinstating the ... argument. NEWS.md file is updated to reflect the coord_map() fix.

R/coord-map.r
```diff
@@ -21,7 +21,8 @@
 #' @export
 #' @param projection projection to use, see
 #'    \code{\link[mapproj]{mapproject}} for list
-#' @param ... other arguments passed on to
+#' @param ... other arguments passed on to mapproject. Ignored if the \code{parameters} argument is present.
+#' @param parameters optional numeric vector of parameters for use with the projection argument. This argument is optional only in the sense that certain projections do not require additional parameters. Passed to
```
hadley: Can you please break this line up? Otherwise looks good.

tdmcarthur:
Added line breaks to coord_map() docs. Should be ready to merge in.

hadley:
Thanks!

PR #1733, opened 2016-08-26
Show a list of linetypes in scale_linetype docs

steveharoz:
Show a simple legend for common line types. It just saves people a trip to the vignette or online docs.

Similar to PR https://github.com/hadley/ggplot2/pull/1578

### DIFF
--- a/R/scale-linetype.r
+++ b/R/scale-linetype.r
@@ -14,6 +14,15 @@
 #' base + geom_line(aes(linetype = variable))
 #'
 #' # See scale_manual for more flexibility
+#'
+#' # show common line types
+#' df_lines <- data.frame(linetype = factor(1:4, labels = c("solid", "longdash", "dashed", "dotted")))
+#' ggplot(df_lines) +
+#'   geom_hline(aes(linetype = linetype, yintercept = 0), size = 2) +
+#'   scale_linetype_identity() +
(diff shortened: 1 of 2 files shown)


R/scale-linetype.r
```diff
@@ -14,6 +14,15 @@
 #' base + geom_line(aes(linetype = variable))
 #'
 #' # See scale_manual for more flexibility
+#'
+#' # show common line types
+#' df_lines <- data.frame(linetype = factor(1:4, labels = c("solid", "longdash", "dashed", "dotted")))
```
hadley: Why not just create a character vector?

steveharoz: A character vector would order the types alphabetically, rather than by spatial freq/luminance. Here's an alternative:

``` r
linetypes <- c("solid", "longdash", "dashed", "dotted")
df_lines <- data.frame(value = seq_along(linetypes))
ggplot(df_lines) +
  geom_hline(aes(linetype = factor(value), yintercept = value), size = 2) +
  geom_text(aes(x = 0, y = value+.25, label = linetypes[value])) +
  scale_linetype_manual(values = linetypes, guide = 'none') +
  theme_void()
```
image

hadley:
Got it - thanks!

PR #1735, opened 2016-08-29
Pass dots to continuous_scale in scale_size/radius

DarwinAwardWinner:
I tried to pass `na.value` to `scale_size_continuous`, and was surprised when it threw an error instead of passing it along to `continuous_scale`.  This pull request fixes that by having it pass along `...`.

### DIFF
--- a/R/scale-size.r
+++ b/R/scale-size.r
@@ -36,20 +36,20 @@ NULL
 #' @usage NULL
 scale_size_continuous <- function(name = waiver(), breaks = waiver(), labels = waiver(),
                                   limits = NULL, range = c(1, 6),
-                                  trans = "identity", guide = "legend") {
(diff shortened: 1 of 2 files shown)


has2k1:
Some refactoring of the code base a while ago got rid of `...` from some of the user facing API. I think it was aiming for a more explicit documentation. So, if you are adding `...`, you should add an item in the documentation of where those other parameters are defined.

DarwinAwardWinner:
Would it be sufficient to add the following passage from other continuous scales (e.g. `scale_alpha_continuous`)? 

     ...: Other arguments passed on to 'continuous_scale' or
          'discrete_scale' as appropriate, to control name, limits,
          breaks, labels and so forth.

has2k1:
That should do.

hadley:
I'd much rather you passed along specific arguments so that the document for individual scales remains self contained.

PR #1749, opened 2016-09-15
Pass super class to scale constructor

### LINKED ISSUE #1732: Fix positional scale constructors
Don't assign super class after construction

thomasp85:
Fixes #1732.

`continuous_scale` and `discrete_scale` now takes a `super` argument that defaults to `ScaleContinuous` and `ScaleDiscrete` respectively. This avoids having to reassign the super class after construction in the different positional scale constructors


hadley:
LGTM

PR #1754, opened 2016-09-16
Add inherit.blank argument to element constructors

thomasp85:
This PR is an alternative fix to #1555, #1557, #1565, and #1567 compared to #1581.

It will propose the inclusion of an `inherit.blank` argument to the `element_*` constructors. The reason for this is that we might want to suppress drawing of a parent element and all its children, and then reenable some of the children again. This requires element calculation to sometimes ignore `element_blank()` and sometimes not. The default should be `inherit.blank = FALSE` while elements in the included themes should have `inherit.blank = TRUE` instead.


hadley:
This seems like a reasonable approach to me

thomasp85:
@hadley can you review?

Now you can do stuff like this (not a beautiful plot but an example):

```r
p + theme_minimal() + theme(line = element_blank(), axis.line.y = element_line())
```

rplot

hadley:
What if we automatically set `inherit.blank` if `complete = TRUE` in `theme()`?

R/theme-elements.r
```diff
@@ -13,6 +13,8 @@
 #' @param fill Fill colour.
 #' @param colour,color Line/border colour. Color is an alias for colour.
 #' @param size Line/border size in mm; text size in pts.
+#' @param inherit.blank Should this element inherit the existence of an
+#' element_blank among its parents?
```
hadley: Can you expand on this a little bit? I assume if it's `TRUE` the "blank-ness" will be inherited, otherwise it will skip the parent and look at the grand parent?

R/theme.r
```diff
@@ -633,10 +633,14 @@ calc_element <- function(element, theme, verbose = FALSE) {
 combine_elements <- function(e1, e2) {
 
   # If e2 is NULL, nothing to inherit
-  if (is.null(e2))  return(e1)
-
+  if (is.null(e2) || inherits(e1, "element_blank"))  return(e1)
+  # If e1 is NULL inherit everything from e2
+  if (is.null(e1)) return(e2)
   # If e1 is NULL, or if e2 is element_blank, inherit everything from e2
```
hadley: This comment needs updating too, right?

Katiedaisey:
I like this idea. It would prevent the recursive checking that #1581 requires, but your example doesn't seem to work for me. 

    ggplot(mtcars, aes(wt, mpg)) + geom_point() + theme_minimal() + theme(line = element_blank(), axis.line.y = element_line())

rplot

When I examine the plot theme, it seems `line = element_blank()` rewrites `element_line()` and so `axis.line.y` calls all NULLS.  Redoing this with `theme(axis.line = element_blank(), axis.line.y = element_line()` also results in all NULLS.

When I try `p + theme_minimal() + theme(axis.line = element_blank(), axis.line.y = element_line(color = "red"))`, I get:

rplot04

Examining the plot, I see `axis.line.y` has `color = "red"` but isn't printing.

```
$ axis.line.y          :List of 4
  ..$ colour  : chr "red"
  ..$ size    : NULL
  ..$ linetype: NULL
  ..$ lineend : NULL
  ..- attr(*, "class")= chr [1:2] "element_line" "element"
```

Also, currently `theme_classic()` doesn't inherit correctly.
rplot01

thomasp85:
This has been fixed I think - truly with the latest version

thomasp85:
I cannot confirm any problems and all changes has been pushed - can you try again and make sure you create the plot object from scratch

hadley:
LGTM

Katiedaisey:
@thomasp85 Sorry, I previewed the comment but apparently never posted it. Yes, I've found no problems with the last update!

Issue #1470, opened 2016-01-05
scales = "free_y" not working on faceted bar chart


gregorbj:
Following is some test data to be plotted in comparison bar charts of two measures that will each occupy a facet. The scale of one measure (Mixed Use) is 0 to 10 while the other measure (Walking) is 0 to 50. Individual x-axis scales are desired for the facets.

Test_df <- data.frame(Policy = rep(c("A", "B", "C", "D"), 4),
                      Measure = rep(c("Mixed Use", "Walking"), each = 8),
                      PctChg = c(10 * runif(8, 0, 1), 50 * runif(8, 0, 1)),
                      Model = rep(rep(c("RPAT", "RSPM"), each = 4), 2)
                      )

The scales = "free_y" (or "free_x", or "free") don't produce individualized scales

ggplot(data = Test_df, aes(x=Policy, y = PctChg, fill = Model)) + 
  geom_bar(stat = "identity", position=position_dodge()) +
  facet_grid(. ~ Measure, scales = "free") +
  coord_flip() +
  theme_bw() + 
  labs(y = "% Change from Reference") +
  labs(x = "") +
  theme(axis.title = element_text(size = 10))

hadley:
Free scales don't work with `coord_flip()`, unfortunately.

Issue #1393, opened 2015-10-28
geom_bar() with facet_grid(space="free") with coord_flip()


mrdanbrooks:
In the code below, I was expecting ``space="free"`` to produce a thinner plot for 'b'. 
```
library(ggplot2)
t <- "Group,Response,Freq,pct_freq,disp
none,1,0,0.000000000,a
low,1,5,0.018939394,a
high,1,65,0.250965251,a
none,2,0,0.000000000,a
low,2,15,0.056818182,a
high,2,89,0.343629344,a
none,3,1,0.007936508,a
low,3,83,0.314393939,a
high,3,73,0.281853282,a
none,4,14,0.111111111,b
low,4,40,0.151515152,b
high,4,14,0.054054054,b
none,5,21,0.166666667,a
low,5,32,0.121212121,a
high,5,14,0.054054054,a
none,6,55,0.436507937,a
low,6,60,0.227272727,a
high,6,4,0.015444015,a
none,7,35,0.277777778,a
low,7,29,0.109848485,a
high,7,0,0.000000000,a"
m <- read.csv(textConnection(t))
m$Response <- factor(m$Response, levels=1:7, ordered=TRUE)
ggplot() + 
    aes(x=Group, fill=Response, order=Response) + 
    geom_bar(data=m, aes(y=pct_freq), position="stack", stat="identity") + 
    facet_grid(. ~ disp, space="free") + 
    coord_flip())
```

![geombarfacet_prob](https://cloud.githubusercontent.com/assets/5226494/10800342/c9af7df4-7d87-11e5-8c87-13f4b142a339.jpg)

It has been pointed out to me that the most likely reason it behaves like this is that without the ``coord_flip()``, the plots are the same height and this property is being kept. If this is actually the "expected behavior" rather then a bug, it might be useful to add a ``geom_bar_horiz``.

In trying to help me find a work around, @mattantaliss sent me the following code based on the accepted answer to http://stackoverflow.com/questions/12560858/using-coord-flip-with-facet-wrapscales-free-y-in-ggplot2-seems-to-give-u
This makes the size of 'b' behave the way I was expecting, but has a side effect of messing up the bars in 'a', and also requires ``scales="free"`` to be set (which I don't want).
 
```
library(ggplot2)
library(proto)

geom_bar_horz <- function (mapping = NULL, data = NULL, stat = "bin", position = "stack", ...) {
  GeomBar_horz$new(mapping = mapping, data = data, stat = stat, position = position, ...)
[...]

Issue #95, opened 2010-04-29
Free scales on facets should take coord_flip into account


jiho:
Let's consider this data    

```
d <- data.frame(x=1:5, y=c(runif(5), runif(5)+5), a=rep(c("foo","bar"), each=5))
```

The scale for y should be free. Faceting along the horizontal makes it impossible since there should only be one y scale:

```
ggplot(d) + geom_point(aes(x=x, y=y)) + facet_grid(.~a, scales="free_y")
```

This makes sense but a warning would be welcome ("Warning: The y scale cannot be free in this faceting configuration")

Now faceting along the vertical or switching the coordinates allows a free scale

```
ggplot(d) + geom_point(aes(x=x, y=y)) + facet_grid(a~., scales="free_y")
ggplot(d) + geom_point(aes(x=y, y=x)) + facet_grid(.~a, scales="free_x")
```

But flipping the coordinate system does not, while there are now two different axes for y

```
ggplot(d) + geom_point(aes(x=x, y=y)) + facet_grid(.~a, scales="free_y") + coord_flip()
```

I would expect this last plot to look just like the previous one.

PR #1756, opened 2016-09-16
Approach facet_grid(scale="free") + coord_flip issue

### LINKED ISSUE #1492: Inconsistent behaviour of coord_flip() with facets
```r
d <- read.csv(text="var,group,min,center,max
foo,a,1,2,3
foo,b,1.5,2.5,3
foo,c,0.5,2,2.5
bar,a,10,15,30
bar,b,12,15,18
bar,c,15,18,20")

library("ggplot2")
library("ggplot2")
ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_wrap(~var)
ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_wrap(~var, scales="free_y")
# I want to get something like this plot but with the numerical value along the horizontal,
# because that's the convention for the type of values I want to plot

ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_wrap(~var, scales="free_y") +
  coord_flip()
# unsupported, :-(

ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_grid(.~var)
ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_grid(.~var, scales="free_y")
# no change but no warning/error message
# => should warn that "y coordinates cannot be free in this configuration of facet_grid" or
#    something more general and informative ;-)

ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_grid(var~.)
ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_grid(var~., scales="free_y")
ggplot(d) +
  geom_pointrange(aes(x=group, y=center, ymin=min, ymax=max)) +
  facet_grid(var~., scales="free_y") +
  coord_flip()
# scales the plots but the scale along the x axis is the one of the first group
# => should either behave like facet_wrap and say that coord_flip is unsupported or
#    add a different x axis per facet
```

I realise that, in that particular case, I am actually plotting different datasets and that the proper way of doing it would be subplots rather than facets. But facets are so easy ;-)

vlsi:
I've run into "well known" `facet_grid(scale="free") + coord_flip()` issue, and it took me a several hours to find a workaround, thus I decided to implement a proper fix.

The idea is to swap panel$layout$SCALE_X and panel$layout$SCALE_Y right after train_layout

The thing is facet_grid tries to keep the same ranges for all columns and rows,
however it does not know if the axes will be flipped, so when flip occurs
the previously chosen set of facets no longer applies.

The fix is to flip SCALE_X and SCALE_Y set of facets, so the ranges are properly computed

Fixes #1492, #1470, #1393, #95

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 2.1.0.9000 
 
+* Fixed facet_grid(scale="free") + coord_flip() behavior (@vlsi, #1492, #1470, #1393, #95)
+
 * The documentation for theme elements has been improved (#1743).
 
 * `geom_boxplot` gain new `outlier.alpha` argument for controlling the alpha of
--- a/R/facet-wrap.r
+++ b/R/facet-wrap.r
@@ -144,9 +144,9 @@ facet_render.wrap <- function(facet, panel, coord, theme, geom_grobs) {
 
   # If coord is (non-cartesian or flip) and (x is free or y is free)
   # then print a warning
-  if ((!inherits(coord, "CoordCartesian") || inherits(coord, "CoordFlip")) &&
+  if (!inherits(coord, "CoordCartesian") &&
     (facet$free$x || facet$free$y)) {
-    stop("ggplot2 does not currently support free scales with a non-cartesian coord or coord_flip.\n")
+    stop("ggplot2 does not currently support free scales with a non-cartesian coord.\n")
   }
 
   # If user hasn't set aspect ratio, and we have fixed scales, then
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -39,6 +39,15 @@ ggplot_build <- function(plot) {
 
   panel <- new_panel()
   panel <- train_layout(panel, plot$facet, layer_data, plot$data)
+
+  # If coordinates will be flipped, then flip panels to train coordinates from
+  if (inherits(plot$coordinates, "CoordFlip")) {
+    # This fixes facet_grid(scale="free") + coord_flip() issue
+    tmp <- panel$layout$SCALE_X
+    panel$layout$SCALE_X <- panel$layout$SCALE_Y
+    panel$layout$SCALE_Y <- tmp
+  }
+
   data <- map_layout(panel, plot$facet, layer_data)
 
   # Compute aesthetics to produce data with generalised variable names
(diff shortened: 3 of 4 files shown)


hadley:
I have to say I thought I was going to hate this, but it seems so simple that it's worth doing.  It doesn't solve the general problem but it's a pragmatic solution to the case that most people encounter. Could you have a look at doing it for `facet_grid()` too? 

And all the facetting code has been rewritten in #1633 but I think the basic idea will still work.

cc @thomasp85

vlsi:
Here's sample from #1470:
```
Test_df <- data.frame(Policy = rep(c("A", "B", "C", "D"), 4),
                      Measure = rep(c("Mixed Use", "Walking"), each = 8),
                      PctChg = c(10 * runif(8, 0, 1), 50 * runif(8, 0, 1)),
                      Model = rep(rep(c("RPAT", "RSPM"), each = 4), 2)
)

ggplot(data = Test_df, aes(x=Policy, y = PctChg, fill = Model)) + 
  geom_bar(stat = "identity", position=position_dodge()) +
  facet_grid(. ~ Measure, scales = "free") +
  coord_flip() +
  theme_bw() + 
  labs(y = "% Change from Reference") +
  labs(x = "") +
  theme(axis.title = element_text(size = 10))
```

Before:
before

After:
after

vlsi:
> Could you have a look at doing it for facet_grid() too?

It works for both facet_grid and facet_wrap

vlsi:
>I have to say I thought I was going to hate this, but it seems so simple that it's worth doing. It doesn't solve the general problem but it's a pragmatic solution to the case that most people encounter

Exactly. I feel what you mean by "facets and coordinate systems are not well designed to play together", however I think the fix is rather trivial, so it might have its purpose

thomasp85:
Thanks for this @vlsi - I'll add it to #1633 though in a slightly different way that doesn't modify the build code itself.

thomasp85:
This was fixed in #1633

PR #1760, opened 2016-09-19
show.legend can now be a logical vector

Tutuchan:
Hi,

With this PR, it is now possible to pass a named logical vector to `show.legend` in order to have better control of what appears in the legend or not. See this example: 

```r
library(ggplot2)

p <- ggplot(mtcars) +
  geom_point(aes(mpg, qsec, colour = factor(vs), shape = factor(am)))

p + geom_line(aes(mpg, qsec, colour = factor(vs), linetype = factor(am)))

# Current behaviour: remove this layer from the legend
p + geom_line(aes(mpg, qsec, colour = factor(vs), linetype = factor(am)), show.legend = FALSE)

# In the current version, this would have the same behaviour but here, 
# it removes only the colour aesthetic for the line geom from the legend
p + geom_line(aes(mpg, qsec, colour = factor(vs), linetype = factor(am)), 
              show.legend = c(colour = FALSE))
```

image

The only problem (that I see anyway) is that, for the colour scale, the name in the `show.legend` vector **must** be `colour`, `color` will not work. It might be solved by adding a check when trying to match the names but it's not very elegant. Any ideas ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,8 @@
   degrees Fahrenheit). The secondary axis will be positioned opposite of the 
   primary axis and can be controlled using the `sec.axis` argument to the scale
   constructor.
+  
+* Aesthetics can now independently hidden in the legend (@Tutuchan).
 
 * The documentation for theme elements has been improved (#1743).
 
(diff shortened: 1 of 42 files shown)


hadley:
Before proposing a PR, please file an issue to discuss the problem and possible solutions.

Tutuchan:
Alright my mistake, I'll file an issue.

PR #1762, opened 2016-09-20
Remove ellipsis from theme

### LINKED ISSUE #1744: Eliminate ... in themes()
Instead using multiple argument names, and documenting them with `@param`

thomasp85:
Fixes #1744

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,6 +53,8 @@
 * Themes are more homogeneous visually, and match `theme_grey` better.
   (@jiho, #1679)
 
+* The `theme()` constructor now has named arguments rather than ellipsis.
+
 * `position_stack()` and `position_fill()` now sorts the stacking order so it 
   matches the order of the grouping. Use level reordering to alter the stacking 
   order. The default legend and stacking order is now also in line. The default 
--- a/NEWS.md
+++ b/NEWS.md
@@ -56,7 +56,8 @@
 * Themes are more homogeneous visually, and match `theme_grey` better.
   (@jiho, #1679)
 
-* The `theme()` constructor now has named arguments rather than ellipsis.
+* The `theme()` constructor now has named arguments rather than ellipsis. This 
+  should make autocomplete substantially more useful.
 
 * `position_stack()` and `position_fill()` now sorts the stacking order so it 
   matches the order of the grouping. Use level reordering to alter the stacking 
--- a/R/theme.r
+++ b/R/theme.r
@@ -384,7 +384,7 @@ theme <- function(line, rect, text, title, aspect.ratio, axis.title,
                   plot.caption, plot.margin, strip.background, strip.text,
                   strip.text.x, strip.text.y, strip.switch.pad.grid,
                   strip.switch.pad.wrap, ..., complete = FALSE, validate = TRUE) {
-  elements <- as.list(environment())
+  elements <- c(as.list(environment()), list(...))
   elements$complete <- NULL
   elements$validate <- NULL
   elements <- elements[lapply(elements, class) != "name"]
--- a/R/theme.r
+++ b/R/theme.r
@@ -387,7 +387,7 @@ theme <- function(line, rect, text, title, aspect.ratio, axis.title,
                   plot.caption, plot.margin, strip.background, strip.text,
                   strip.text.x, strip.text.y, strip.switch.pad.grid,
                   strip.switch.pad.wrap, ..., complete = FALSE, validate = TRUE) {
-  elements <- find_args(..., .ignore = c("complete", "validate"))
+  elements <- find_args(..., complete = NULL, validate = NULL)
 
   if (!is.null(elements$axis.ticks.margin)) {
     warning("`axis.ticks.margin` is deprecated. Please set `margin` property ",
--- a/R/theme.r
+++ b/R/theme.r
@@ -385,10 +385,7 @@ theme <- function(line, rect, text, title, aspect.ratio, axis.title,
                   plot.caption, plot.margin, strip.background, strip.text,
                   strip.text.x, strip.text.y, strip.switch.pad.grid,
                   strip.switch.pad.wrap, ..., complete = FALSE, validate = TRUE) {
-  elements <- c(as.list(environment()), list(...))
-  elements$complete <- NULL
-  elements$validate <- NULL
-  elements <- elements[!vapply(elements, is.name, logical(1))]
+  elements <- find_args(..., .ignore = c("complete", "validate"))
 
   if (!is.null(elements$axis.ticks.margin)) {
     warning("`axis.ticks.margin` is deprecated. Please set `margin` property ",
--- a/R/theme.r
+++ b/R/theme.r
@@ -80,166 +80,145 @@ print.theme <- function(x, ...) utils::str(x)
 #' To see a graphical representation of the inheritance tree, see the
 #' last example below.
 #'
-#' @section Theme elements:
-#' The individual theme elements are:
+#' @param line all line elements (\code{element_line})
+#' @param rect all rectangular elements (\code{element_rect})
+#' @param text all text elements (\code{element_text})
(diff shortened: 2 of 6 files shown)


R/theme.r
```diff
@@ -385,8 +364,30 @@ print.theme <- function(x, ...) utils::str(x)
...
+                  panel.grid.major, panel.grid.minor, panel.grid.major.x,
+                  panel.grid.major.y, panel.grid.minor.x, panel.grid.minor.y,
+                  panel.ontop, plot.background, plot.title, plot.subtitle,
+                  plot.caption, plot.margin, strip.background, strip.text,
+                  strip.text.x, strip.text.y, strip.switch.pad.grid,
+                  strip.switch.pad.wrap, ..., complete = FALSE, validate = TRUE) {
+  elements <- as.list(environment())
+  elements$complete <- NULL
+  elements$validate <- NULL
+  elements <- elements[lapply(elements, class) != "name"]
```
thomasp85: I would love some feedback on whether this is the best way to capture all non-missing arguments as evaluated values...?

R/theme.r
```diff
@@ -80,166 +80,145 @@ print.theme <- function(x, ...) utils::str(x)
...
-#' @param ... a list of element name, element pairings that modify the
-#'   existing theme.
-#' @param complete set this to TRUE if this is a complete theme, such as
-#'   the one returned \code{by theme_grey()}. Complete themes behave
-#'   differently when added to a ggplot object.
+#' @param line all line elements (\code{element_line})
+#' @param rect all rectangular elements (\code{element_rect})
+#' @param text all text elements (\code{element_text})
+#' @param title all title elements: plot, axes, legends (\code{element_text};
+#' inherits from \code{text})
```
hadley: Can you please indent the second and subsequent lines?

R/theme.r
```diff
@@ -385,8 +364,30 @@ print.theme <- function(x, ...) utils::str(x)
...
+                  panel.grid.major, panel.grid.minor, panel.grid.major.x,
+                  panel.grid.major.y, panel.grid.minor.x, panel.grid.minor.y,
+                  panel.ontop, plot.background, plot.title, plot.subtitle,
+                  plot.caption, plot.margin, strip.background, strip.text,
+                  strip.text.x, strip.text.y, strip.switch.pad.grid,
+                  strip.switch.pad.wrap, ..., complete = FALSE, validate = TRUE) {
+  elements <- c(as.list(environment()), list(...))
+  elements$complete <- NULL
+  elements$validate <- NULL
+  elements <- elements[lapply(elements, class) != "name"]
```
hadley: Why is this necessary? Regardless, it would be cleaner to use `vapply()` + `is.name()`

R/theme.r
```diff
@@ -385,8 +364,30 @@ print.theme <- function(x, ...) utils::str(x)
...
+                  legend.box.margin, legend.box.background, legend.box.spacing,
+                  panel.background, panel.border, panel.spacing,
+                  panel.spacing.x, panel.spacing.y, panel.grid,
+                  panel.grid.major, panel.grid.minor, panel.grid.major.x,
+                  panel.grid.major.y, panel.grid.minor.x, panel.grid.minor.y,
+                  panel.ontop, plot.background, plot.title, plot.subtitle,
+                  plot.caption, plot.margin, strip.background, strip.text,
+                  strip.text.x, strip.text.y, strip.switch.pad.grid,
+                  strip.switch.pad.wrap, ..., complete = FALSE, validate = TRUE) {
+  elements <- c(as.list(environment()), list(...))
```
hadley: Hmmmm, I don't like this but I don't see a better way except creating the massive named list by hand (which seems likely to be error prone).

R/theme.r
thomasp85: If we don't remove `name` elements the theme validation will complain

hadley: But why? Where do those name elements come from?

thomasp85: these are missing arguments - they exist in the environment and is added to the list as name objects

hadley:
@wch I have vague memory you thought about this problem (capturing all arguments) a bit

R/theme.r
```diff
@@ -385,8 +365,30 @@ print.theme <- function(x, ...) utils::str(x)
...
+  elements$complete <- NULL
+  elements$validate <- NULL
+  elements <- elements[!vapply(elements, is.name, logical(1))]
```
hadley: The absolutely correct check is:

```R
elements <- elements[!vapply(elements, identical, quote(expr = ), FUN.VALUE = logical(1))]
```

R/theme.r
hadley: I would really prefer if this was extracted into a separate function and gained a couple of safe guards:

```R
find_args <- function(...) {
  env <- parent.frame()
  args <- names(formals(sys.function(sys.parent(1))))
  
  vals <- c(as.list(env)[args], list(...))
  vals <- vals[!vapply(vals, identical, quote(expr = ), FUN.VALUE = logical(1))]
  vals
}

f <- function(a = 1, b = 2, c = 3, d, e, abcdefasdfgasdfasdfasd) {
   find_args()
}
f(10, abc = 10)
```

hadley: This is slightly more elegant:

```R
find_args <- function(...) {
  env <- parent.frame()
  args <- names(formals(sys.function(sys.parent(1))))
  
  vals <- mget(args, envir = env)
  vals <- vals[!vapply(vals, identical, quote(expr = ), FUN.VALUE = logical(1))]
  
  c(vals, list(...))
}
```

hadley:
```R
is_missing_arg <- function(x) identical(x. quote(expr = )))
```

hadley: And even better:

```R
find_args <- function(...) {
  env <- parent.frame()
  args <- names(formals(sys.function(sys.parent(1))))

  vals <- mget(args, envir = env)
  vals <- vals[!vapply(vals, identical, quote(expr = ), FUN.VALUE = logical(1))]

  modifyList(vals, list(...))
}
```

R/utilities.r
```diff
@@ -281,6 +281,22 @@ dispatch_args <- function(f, ...) {
   f
 }
 
+is_missing_arg <- function(x) identical(x, quote(expr = ))
+# Get all arguments in a function as a list. Will fail if an ellipsis argument
+# named .ignore
+# @param ... passed on in case enclosing function uses ellipsis in argument list
+# @param .ignore remove these arguments from the list
+find_args <- function(..., .ignore = NULL) {
```
hadley: Why not drop `.ignore` and just use (e.g.) `complete = NULL, validate = NULL)`

NEWS.md
```diff
@@ -56,6 +56,8 @@
 * Themes are more homogeneous visually, and match `theme_grey` better.
   (@jiho, #1679)
 
+* The `theme()` constructor now has named arguments rather than ellipsis.
```
hadley: "This should make autocomplete substantially more useful"

R/utilities.r
thomasp85: Because `validate_theme` will throw an error if they are included in the list

hadley:
```R
find_args(..., complete = NULL, validate = NULL)
```

hadley: Then `modifyList` will remove them

thomasp85: Ah - didn't think of that

R/utilities.r
```diff
@@ -281,6 +281,21 @@ dispatch_args <- function(f, ...) {
   f
 }
 
+is_missing_arg <- function(x) identical(x, quote(expr = ))
+# Get all arguments in a function as a list. Will fail if an ellipsis argument
+# named .ignore
+# @param ... passed on in case enclosing function uses ellipsis in argument list
+# @param .ignore remove these arguments from the list
```
hadley: This needs to be removed now, other LGTM

PR #1737, opened 2016-08-30
Scale gradient doc fix

DarwinAwardWinner:
The documentation now mentions "continuous_scale" instead of "discrete_scale".

Fixes #1736.

PR #1763, opened 2016-09-20
Remove ellipsis from scale constructors

### LINKED ISSUE #1750: Eliminate use of ... in scale constructors
Encompasses #1735, #1737 etc.

thomasp85:
Fixes #1750, #1735, and #1737

### DIFF
--- a/R/scale-alpha.r
+++ b/R/scale-alpha.r
@@ -22,9 +22,9 @@ scale_alpha <- function(name = waiver(), range = c(0.1, 1), breaks = waiver(),
                         limits = NULL, oob = censor, expand = waiver(),
                         na.value = NA_real_, trans = "identity",
                         guide = "legend") {
-  args <- as.list(environment)
+  args <- as.list(environment())
   args$aesthetics <- "alpha"
-  args$name <- "alpha_c"
+  args$scale_name <- "alpha_c"
   args$palette <- rescale_pal(range)
   args$range <- NULL
   do.call(continuous_scale, args)
@@ -41,9 +41,9 @@ scale_alpha_discrete <- function(name = waiver(), range = c(0.1, 1),
                                  breaks = waiver(), labels = waiver(),
                                  limits = NULL, expand = waiver(),
                                  na.value = NA, drop = TRUE, guide = "legend") {
-  args <- as.list(environment)
+  args <- as.list(environment())
   args$aesthetics <- "alpha"
-  args$name <- "alpha_d"
+  args$scale_name <- "alpha_d"
   args$palette <- function(n) seq(range[1], range[2], length.out = n)
   args$range <- NULL
   do.call(discrete_scale, args)
--- a/R/scale-alpha.r
+++ b/R/scale-alpha.r
@@ -3,9 +3,7 @@
 #' \code{scale_alpha} is an alias for \code{scale_alpha_continuous} since
 #' that is the most common use of alpha, and it saves a bit of typing.
 #'
-#' @param ... Other arguments passed on to \code{\link{continuous_scale}}
-#'   or \code{\link{discrete_scale}} as appropriate, to control name, limits,
-#'   breaks, labels and so forth.
+#' @inheritParams continuous_scale
 #' @param range range of output alpha values.  Should lie between 0 and 1.
 #' @export
 #' @examples
@@ -19,8 +17,17 @@
 #' (p <- ggplot(mtcars, aes(mpg, cyl)) +
 #'   geom_point(aes(alpha = factor(cyl))))
 #' p + scale_alpha_discrete(range = c(0.4, 0.8))
-scale_alpha <- function(..., range = c(0.1, 1)) {
-  continuous_scale("alpha", "alpha_c", rescale_pal(range), ...)
+scale_alpha <- function(name = waiver(), range = c(0.1, 1), breaks = waiver(),
+                        minor_breaks = waiver(), labels = waiver(),
+                        limits = NULL, oob = censor, expand = waiver(),
+                        na.value = NA_real_, trans = "identity",
+                        guide = "legend") {
+  args <- as.list(environment)
+  args$aesthetics <- "alpha"
+  args$name <- "alpha_c"
+  args$palette <- rescale_pal(range)
+  args$range <- NULL
+  do.call(continuous_scale, args)
 }
 
 #' @rdname scale_alpha
@@ -28,8 +35,16 @@ scale_alpha <- function(..., range = c(0.1, 1)) {
 scale_alpha_continuous <- scale_alpha
 
 #' @rdname scale_alpha
+#' @inheritParams discrete_scale
 #' @export
-scale_alpha_discrete <- function(..., range = c(0.1, 1)) {
-  discrete_scale("alpha", "alpha_d",
-    function(n) seq(range[1], range[2], length.out = n), ...)
+scale_alpha_discrete <- function(name = waiver(), range = c(0.1, 1),
+                                 breaks = waiver(), labels = waiver(),
+                                 limits = NULL, expand = waiver(),
+                                 na.value = NA, drop = TRUE, guide = "legend") {
+  args <- as.list(environment)
+  args$aesthetics <- "alpha"
+  args$name <- "alpha_d"
+  args$palette <- function(n) seq(range[1], range[2], length.out = n)
+  args$range <- NULL
+  do.call(discrete_scale, args)
 }
--- a/man/scale_alpha.Rd
+++ b/man/scale_alpha.Rd
@@ -6,18 +6,83 @@
 \alias{scale_alpha_discrete}
 \title{Alpha scales.}
 \usage{
-scale_alpha(..., range = c(0.1, 1))
+scale_alpha(name = waiver(), range = c(0.1, 1), breaks = waiver(),
+  minor_breaks = waiver(), labels = waiver(), limits = NULL,
+  oob = censor, expand = waiver(), na.value = NA_real_,
+  trans = "identity", guide = "legend")
 
-scale_alpha_continuous(..., range = c(0.1, 1))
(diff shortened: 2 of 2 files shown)


thomasp85:
@haldey - I just changed it for scale_alpha for the time being. Can you have a look and see if this approach looks good to you...

hadley:
I really don't like this approach, but I don't think there's any way to do better, so I think I'd rather fix this in roxygen2, by providing someway to specify that the docs for ... should come from `continuous_scale` (i.e. https://github.com/klutometis/roxygen/issues/512).

That won't help with autocomplete though.  Maybe @kevinushey will have some ideas.

thomasp85:
Which part is it you don't like? The forwarding of arguments or the doccumentation?

hadley:
The forwarding of the arguments.

thomasp85:
I can do it manually as is done for `scale_*_position` - just tried to be smart

kevinushey:
RStudio does not yet have the smart control-flow analysis needed to detect how `...` arguments are delegated.

However, it's possible we could also look at the function documentation for available arguments, e.g. by parsing the `.Rd` help and reading the parameter names in there. We could then use that to provide smart autocompletion in the presence of `...` arguments.

R/scale-alpha.r
```diff
@@ -19,17 +17,34 @@
 #' (p <- ggplot(mtcars, aes(mpg, cyl)) +
 #'   geom_point(aes(alpha = factor(cyl))))
 #' p + scale_alpha_discrete(range = c(0.4, 0.8))
-scale_alpha <- function(..., range = c(0.1, 1)) {
-  continuous_scale("alpha", "alpha_c", rescale_pal(range), ...)
+scale_alpha <- function(name = waiver(), range = c(0.1, 1), breaks = waiver(),
```
hadley: What if we wrote `forward_to` so that this would work:

```R
scale_alpha <- forward_to(continuous_scale, 
  aesthetics = "alpha",
  scale_name = "alpha_c",
  palette = rescale_pal()
)
```

hadley: One downside is that this would require `continuous_scale` to be defined before `scale_alpha` in order to inspect the arguments.

hadley: Or what about:

```R
forward_to(contiunous_scale,
  aesthetics = "alpha",
  scale_name = "alpha_c",
  palette = rescale_pal(range),
  range = NULL
)
```

hadley:
```R
forward_to <- function(fun, ...) {
  call <- c(quote(fun), find_args(...))
  eval(call)
}
```

thomasp85:
I've made this piece of magical code that I would like you to consider (it depends on the `find_args()` function from #1762)

```r
forward_to <- function(.fun, ..., .mask, .prepare) {
  args <- as.list(match.call())[-1]
  args <- args[!names(args) %in% c(".fun", ".mask", ".prepare")]
  frml <- formals(.fun)
  .prepare <- substitute(.prepare)
  .fun <- substitute(.fun)
  new_frml <- modifyList(frml[!names(frml) %in% .mask], args)
  new_fun <- function() {
    args <- find_args()
    prepared <- new.env()
    eval(.prepare, envir = prepared)
    args <- modifyList(args, as.list(prepared))
    args <- args[names(args) %in% names(frml)]
    call <- as.call(c(.fun, args))
    eval(call)
  }
  formals(new_fun) <- new_frml
  new_fun
}
```

with it you can do:

```r
scale_alpha <- forward_to(continuous_scale, 
  range = c(0.1, 1), 
  .mask = c("palette", "super", "position", "rescaler", "aesthetics", "scale_name"),
  .prepare = {
    aesthetics <- "alpha"
    scale_name <- "alpha_c"
    palette <- rescale_pal(range)
  }
)
```

It's a bit verbose but it allow you to set new arguments with defaults, define which arguments in the destination function should be inaccessible and give freedom to manipulate and create variables prior to calling the destination function (any variable in `.prepare` that match the name of a formal in the destination function is passed along)

jimhester:
Wouldn't the following be a simpler implementation and produce more readable function bodies, maybe I am missing a subtlety though...

``` r
forward_to <- function(f, ..., .ignore = character(), .code = NULL) {
  fmls <- modifyList(formals(f), list(...))
  fmls <- fmls[!names(fmls) %in% .ignore]
  called_fmls <- stats::setNames(lapply(names(fmls), as.symbol), names(fmls))
  call <- as.call(c(substitute(f), called_fmls))

  formals(f) <- fmls
  body(f) <- bquote({
    .(code)
    .(call)},
    as.environment(list(call = call, code = substitute(.code))))
  f
}

forward_to(continuous_scale, range = c(0.1, 1),
  .ignore = c("palette", "super", "position", "rescaler", "aesthetics", "scale_name"),
  .code = {
    aesthetics <- "alpha"
    scale_name <- "alpha_c"
    pallette <- rescale_pal(range)
  })
#> function (name = waiver(), breaks = waiver(), minor_breaks = waiver(), 
#>     labels = waiver(), limits = NULL, oob = censor, expand = waiver(), 
#>     na.value = NA_real_, trans = "identity", guide = "legend", 
#>     range = c(0.1, 1)) 
#> {
#>     {
#>         aesthetics <- "alpha"
#>         scale_name <- "alpha_c"
#>         pallette <- rescale_pal(range)
#>     }
#>     continuous_scale(name = name, breaks = breaks, minor_breaks = minor_breaks, 
#>         labels = labels, limits = limits, oob = oob, expand = expand, 
#>         na.value = na.value, trans = trans, guide = guide, range = range)
#> }
#> <environment: namespace:ggplot2>
```

thomasp85:
I'm so new to meta-programming and was so proud until you came along @jimhester :-)

But one thing your function is missing is to take matching variables from `.code`/`.prepare` and add them to the call to `f`/`.fun`

jimhester:
Oh that is because you had them in ignore in your example, the following is what you are looking for right?

``` r
forward_to <- function(f, ..., .ignore = character(), .code = NULL) {
  fmls <- modifyList(formals(f), list(...))
  fmls <- fmls[!names(fmls) %in% .ignore]
  called_fmls <- stats::setNames(lapply(names(fmls), as.symbol), names(fmls))
  call <- as.call(c(substitute(f), called_fmls))

  formals(f) <- fmls
  body(f) <- bquote({
    .(code)
    .(call)},
    as.environment(list(call = call, code = substitute(.code))))
  f
}

forward_to(continuous_scale, range = c(0.1, 1),
  .ignore = c("super", "position", "rescaler"),
  .code = {
    aesthetics <- "alpha"
    scale_name <- "alpha_c"
    pallette <- rescale_pal(range)
  })
#> function (aesthetics, scale_name, palette, name = waiver(), breaks = waiver(), 
#>     minor_breaks = waiver(), labels = waiver(), limits = NULL, 
#>     oob = censor, expand = waiver(), na.value = NA_real_, trans = "identity", 
#>     guide = "legend", range = c(0.1, 1)) 
#> {
#>     {
#>         aesthetics <- "alpha"
#>         scale_name <- "alpha_c"
#>         pallette <- rescale_pal(range)
#>     }
#>     continuous_scale(aesthetics = aesthetics, scale_name = scale_name, 
#>         palette = palette, name = name, breaks = breaks, minor_breaks = minor_breaks, 
#>         labels = labels, limits = limits, oob = oob, expand = expand, 
#>         na.value = na.value, trans = trans, guide = guide, range = range)
#> }
#> <environment: namespace:ggplot2>
```

jimhester:
@thomasp85 I updated the functions a bit, let me know how that looks.

thomasp85:
Except `.code` could give rise to variables not matching any arguments so they need to be matched to the formals of the function and potentially override those given in `call`

I'm on a phone so I'm reluctant to try atm but I can probably modify your version to give me what I want

jimhester:
I don't think it matters what is in `.code` assuming you use the same names, an example of the case you are worried about would help.

thomasp85:
I'll try to break it tomorrow😉

thomasp85:
Ah, now I see what it's doing. Yeah it should work, at the expense of both defining `.dont_forward` and `.dont_call`

And I do agree that the resulting function is more comprehensible...

@hadley should I proceed down this road?

hadley:
Neither of those approaches seem particularly clear to me — I'd like something that does a better job of illuminating the relationship between the two functions. Let me think about it for a bit.

thomasp85:
Wouldn't some detailed documentation and comments help?

I think you'll have a hard time finding an approach that both eleminate ellipsis, gracefully handles new arguments to `continuous_scale`/`discrete_scale`, supports new arguments and calculations prior to calling the function forwarded to, and being cognitively simple...

hadley:
It's not the implementation of the `forwards_to` function that I'm particularly concerned about, but the code that you write with it.

thomasp85:
So it's the nature of the generated function? The last of Jims iterations almost mirrors the current source code except for the fenced block

thomasp85:
Oh you mean the verbosity of the call to `forward_to`?

hadley:
Yeah, I don't find the call to `forward_to` particularly illuminating.

thomasp85:
I'll try to ponder on a better syntax

thomasp85:
What about a syntax such as this:

```r
forward(
  args = list(range = c(0.1, 1)),
  code = {
    aesthetics <- "alpha"
    scale_name <- "alpha_c"
    pallette <- rescale_pal(range)
  },
  to = continuous_scale,
  redefine = c("palette", "aesthetics", "scale_name"),
  use.defaults = c("super", "position", "rescaler")
)
```

I think you can almost read that aloud and make sense of it... 

> forward these arguments and this code to this function, redefining these arguments and using defaults for these...

jimhester:
@thomasp85 I like that syntax, I would prefer it be `use_defaults`, but using a period is needed to fit with the rest of ggplot2 style I suppose :(

hadley:
I think this is better, but not good enough to be worth finishing off for this release.

PR #1765, opened 2016-09-20
Rename value to count in stat_binhex

### LINKED ISSUE #1747: geom_hex no longer recognizes ..count..
Strongly related to #1608.

This used to work:

```
DF <- data.frame(x=rnorm(1000), y=rnorm(1000))
ggplot(DF) + geom_hex(aes(x=x, y=y, alpha=..count..))
```
But now returns the error: `object 'count' not found`. #1608 reported the same behavior for `..density..` but this has been resolved since. There is a perfectly viable workaround using `..value..` instead, but `..count..` seems more consistent with other stats and geoms, and old code may no longer work.

Also see on [SO](https://stackoverflow.com/questions/39446852/geom-hexbin-map-bincount-to-alpha/).

thomasp85:
Fixes #1747

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -134,6 +134,8 @@
 
 * Fixed problem with `geom_dotplot()` when facetting and binning on the
   y-axis. (#1618, @has2k1)
+  
+* stat_binhex now again returns `count` rather than `value` (#1747)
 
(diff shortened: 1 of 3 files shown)


hadley:
Oh duh it was easy

hadley:
LGTM

PR #1769, opened 2016-09-21
Add finish_statistics and finish_layer as hook in layer and stat

### LINKED ISSUE #1493: Hook in Stat to modify data after scales has been applied
For some of the geoms/stats I'm developing I'm missing a way to intercept data after scales has been applied but before they are passed to `draw_layer`. Kind of like `Geom$setup_data` but attached to `Stat` and called after

```r
if (npscales$n() > 0) {
    lapply(data, scales_train_df, scales = npscales)
    data <- lapply(data, scales_map_df, scales = npscales)
}
```

in `ggplot_build`

If you're willing to have this hook in ggplot2 I'll se to it that it gets implemented...

thomasp85:
Fixes #1493


thomasp85:
@hadley can you look at the proposed implementation? I'll wait with documenting it until we agree on how it should be done

hadley:
LGTM

thomasp85:
Are you OK with the added docs?

hadley:
👍

PR #1777, opened 2016-09-25
Fix jitter width and jitter height

### LINKED ISSUE #1775: I'm confusing about width(height) argument in geom_jitter()
"width(height)" argument in geom_jitter() is described as [belows.](http://docs.ggplot2.org/current/geom_jitter.html)

> ### **width**
Amount of vertical and horizontal jitter. The jitter is added in both positive and negative directions, so the total spread is `twice the value specified here`. If omitted, `defaults to 40% of the resolution of the data`: this means the `jitter values will occupy 80%` of the implied bins.

So, I expected `geom_jitter()` and `geom_jitter(width = 0.4)` will produce same plot and `"width = 0.4"` spreads points across a range `"base ± resolution * 0.4"`. but the results are not.
```
> p <- ggplot(mtcars, aes(gear, mpg))
> p + geom_jitter()  ## jitter range is 3 ± 0.4
```
default

```
> p + geom_jitter(width = 0.4) ## jitter range is narrow than 3 ± 0.4
```
width 0 4

I'm confusing about width argument and how can i calculate a jitter range.

I expect like belows but the result doesn't

geom_jitter(`width = 0.2`) will spread points across a range [2.8 ~ 3.2] : 3 ± 0.2
geom_jitter(`width = 0.4`) will spread points across a range [2.6 ~ 3.4] : 3 ± 0.4
geom_jitter(`width = 0.6`) will spread points across a range [2.4 ~ 3.6] : 3 ± 0.6

but, `geom_point(aes(x = jitter(gear, amount = 0.2))` produce a plot what i expected.
geom_point jitter

Please, give me some hints...

has2k1:
The problem was the %||% operator accepts single tokens and so
care (bracketing as required) must be taken avoid miscalculations.

Some other places in the code-base had it right.

Fixes #1775

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -175,6 +175,9 @@
     * Added `legend.box.spacing` to control the distance between the plot area 
       and the legend area
 
+* Fixed the jitter width and jitter height of `position_jitter` when the
+  parameters are supplied by the user. (#1775, @has2k1)
+
 # ggplot2 2.1.0
 
 ## New features
--- a/R/position-jitter.r
+++ b/R/position-jitter.r
@@ -44,8 +44,8 @@ PositionJitter <- ggproto("PositionJitter", Position,
 
   setup_params = function(self, data) {
     list(
-      width = self$width %||% resolution(data$x, zero = FALSE) * 0.4,
-      height = self$height %||% resolution(data$y, zero = FALSE) * 0.4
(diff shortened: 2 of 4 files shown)


hadley:
Nice - thanks!  (FWIW, I'd refer to this problem as an operator precedence issue)

PR #1783, opened 2016-09-27
Expand ydensity range for nicer violin plots

### LINKED ISSUE #1700: violin plots are rendered with long tails in current stable GGPLOT2
Dear Hadley,

I and others have notice that the current version of GGPLOT2 renders violin plots with long tails at the upper end unlike previous versions. 

Is there any way to revert back to the old rendering?

Thanks,

Francesco

thomasp85:
Fixes #1700

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -70,7 +70,8 @@
   should make autocomplete substantially more useful.
 
 * geom_violin now again has a nicer looking range that allow the density to
-  reach zero (#1700)
+  reach zero. The range of each violin is now automatically extended 3 * bw for 
+  either end of the data range (#1700)
 
 * `position_stack()` and `position_fill()` now sorts the stacking order so it 
   matches the order of the grouping. Use level reordering to alter the stacking 
--- a/NEWS.md
+++ b/NEWS.md
@@ -69,6 +69,8 @@
 * The `theme()` constructor now has named arguments rather than ellipsis. This 
   should make autocomplete substantially more useful.
 
+* geom_violin now again has a nicer looking range that allow the density to
+  reach zero (#1700)
 
 * `position_stack()` and `position_fill()` now sorts the stacking order so it 
   matches the order of the grouping. Use level reordering to alter the stacking 
--- a/R/stat-ydensity.r
+++ b/R/stat-ydensity.r
@@ -69,21 +69,7 @@ StatYdensity <- ggproto("StatYdensity", Stat,
     } else {
       range <- scales$y$dimension()
     }
-    if (is.character(bw)) {
-      if (nrow(data) < 2)
-        stop("need at least 2 points to select a bandwidth automatically", call. = FALSE)
-      bw <- switch(
-        tolower(bw),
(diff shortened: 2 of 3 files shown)


thomasp85:
Continuing examples from #1700:

```r
res <- data.frame(a = c(1:100),b=c(rep(0,50),rep(1,50)))
ggplot(res,aes(y=a,x=factor(b))) + geom_violin(trim=F)
```
rplot

```r
ggplot(res,aes(y=a,x=factor(b))) + geom_violin(trim=T)
```
rplot01

One visual regression is in how zero-range data is displayed is this now remains as a "violin" rather than a single line:

```r
dff <- data.frame(x = factor(rep(1, 10)), y = rep(0, 10))
ggplot(dff, aes(x, y)) + geom_violin()
```
rplot02

We can make a simple check for zero range condition and not expand in that case, but I'm not sure it is the right thing to do...

hadley:
I think that's the right behavior for zero range data

thomasp85:
The current or the proposed change (I personally prefer the current as it does not treat zero range differently)

R/stat-ydensity.r
```diff
@@ -69,7 +69,22 @@ StatYdensity <- ggproto("StatYdensity", Stat,
     } else {
       range <- scales$y$dimension()
     }
-    dens <- compute_density(data$y, data$w, from = range[1], to = range[2],
+    if (is.character(bw)) {
```
hadley: Would you mind pulling this out into a separate function?

hadley:
The plot that you included here seems like the correct behaviour to me.

NEWS.md
```diff
@@ -69,6 +69,8 @@
 * The `theme()` constructor now has named arguments rather than ellipsis. This 
   should make autocomplete substantially more useful.
 
+* geom_violin now again has a nicer looking range that allow the density to
```
hadley: The range of each violin is now automatically extended 3 * bw for either end of the data range

hadley:
LGTM

wch:
I don't think this is the right behavior for `trim=T`. It's supposed to trim to the range of the data, not until the tails have a near-zero width.

hadley:
@wch you mean it should also set the `bw` multiplier to 0?

wch:
Yes, I think that's right. Here's what they looked like way back when they were first implemented:
https://github.com/wch/ggplot2/wiki/geom_violin

thomasp85:
Then there's really no need for a fix?

wch:
With `trim=T`, the violin for b=0 should go from 1 to 50, but right now it goes from about -15 to 70.

wch:
With 1.0.1, this is what it looked like.

```R
res <- data.frame(a = c(1:100),b=c(rep(0,50),rep(1,50)))
ggplot(res,aes(y=a,x=factor(b))) + geom_violin(trim=T)
```

image

wch:
And with `trim=F`, it looked like this:

image


I don't see why it would be useful to have the tails running all the way to the top or bottom.

PR #1789, opened 2016-09-28
Fix stack and clipping in facets

### LINKED ISSUE #1786: Line weights/colors don't look right with theme_bw() and facets
The panels have a bit of a raised, 3-D look to them. Also, the thinner/lighter lines on the left side look a bit off when the facet label box has a darker/heavier line.

```R
  geom_point() +
  facet_grid(am ~ cyl) +
  theme_bw()
```

image

thomasp85:
Fixes #1786

### DIFF
--- a/R/facet-grid-.r
+++ b/R/facet-grid-.r
@@ -311,7 +311,7 @@ FacetGrid <- ggproto("FacetGrid", Facet,
     }
 
     panel_table <- gtable_matrix("layout", panel_table,
-      panel_widths, panel_heights, respect = respect)
+      panel_widths, panel_heights, respect = respect, clip = "on")
     panel_table$layout$name <- paste0('panel-', rep(seq_len(ncol), nrow), '-', rep(seq_len(nrow), each = ncol))
 
     panel_table <- gtable_add_col_space(panel_table,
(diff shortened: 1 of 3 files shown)


wch:
It looks better but there's still a bit of a 3D look because the line weights aren't the same all the way around the facets.

Using your branch, I ran this code in RStudio on Linux:

```R
p <- ggplot(mtcars, aes(wt, mpg)) +
  geom_point() +
  facet_grid(. ~ am) +
  theme_bw()

# RStudio
p

png('test-png.png', type = 'cairo', width = 300, height = 200)
print(p)
dev.off()

Cairo::CairoPNG('test-CairoPNG.png', width = 300, height = 200)
print(p)
dev.off()
```

RStudio:

image

`png()`:

test-png

`CairoPNG()`:

test-cairopng

thomasp85:
Hmm... I'm not sure there is anything more I can do from the facet side of things? All panels are clipped and positioned with the lowest z-value (just like before the rewrite)...

If you have time can you try to apply the old theme settings to the new version and see if it all stems from that?

wch:
Don't let my previous comment stop you from merging if this is the best that can be done - it's still an improvement from the current situation. I don't know if it every looked exactly right.

thomasp85:
For some reason both cairo calls crashes my machine, but in RStudio the plot is tack-sharp

thomasp85:
Ok I got it to work. 'cairo' still looks strange but regular png, RStudio itself, pdf etc looks as it should. I'm almost certain that it has something to do with the antialiasing done by cairo - this could be very susceptible to theme tweaks related to line width

PR #1805, opened 2016-10-01
Allow separate expansion values for lower and upper range limits (v2)

huftis:
Fixes issue #1669.

The `expand` argument for `scale_*_continuous()` and `scale_*_discrete()` now accepts separate expansion constants for the lower and upper range limits.

This is useful for creating bar charts where the bottom of the bars are flush with the _x_ axis but the bars still have some (automatically calculated amount of) space above them:

``` R
ggplot(mtcars) +
  geom_bar(aes(x = factor(cyl))) +
  scale_y_continuous(expand = c(0, 0, 0.1, 0))
```

It can also be useful for line charts, e.g. for counts over time, where one wants to have a ’hard’ lower limit of y = 0, but leave the upper limit unspecified (and perhaps differing between panels), but with some extra space above the highest point on the line. (With symmetrical limits, the extra space above the highest point could cause the lower limit to be negative.)

The syntax for the multiplicative and additive expansion constants has been changed from `c(m, a)` to `c(m_lower, a_lower, m_uppper, a_upper)`. The old syntax will still work, as length 2 vectors `c(m, a)` are expanded to `c(m, a, m, a)` and length 3 vectors are expanded from `c(m1, a1, m2)` to `c(m1, a2, m2, a1)`.

Note that **by default**, all graphs should exactly look like they do with the current version of `ggplot2`.

Some examples (only meant to illustrate the functionality – several don’t make much sense from a graphical point of view):

``` R
# Continuous scales
p = ggplot(mtcars) +
      geom_bar(aes(x = factor(cyl)))

# No space below but 10% space above bars
p + scale_y_continuous(expand = c(0, 0, 0.10, 0))

# It also works with facets
p + facet_wrap(~cyl) +
    scale_y_continuous(expand = c(0, 0, 0.05))
p + facet_wrap(~cyl, scales="free_y") +
    scale_y_continuous(expand = c(0, 0, 0.05))
# Note that all the bars should
# look identical in this last example

# Coordinate flipping also works
p + scale_y_continuous(expand = c(.5, 0, 0.05, 0)) +
    coord_flip()
p + scale_y_continuous(expand = c(0, 2, 0, 5)) +
    coord_flip()


# Similar examples for discrete scales
d <- ggplot(subset(diamonds, carat > 1), aes(cut, clarity)) +
  geom_jitter()

d + scale_x_discrete(expand = c(.25, 0))
d + scale_x_discrete(expand = c(0, 1))
d + scale_x_discrete(expand = c(0, 1, 0, 5))
d + scale_x_discrete(expand = c(.25, 0, .5, 0))
d + scale_x_discrete(expand = c(.25, 1, .25, 5))
d + scale_x_discrete(expand = c(.25, 1, .25, 5)) + coord_flip()
```


hadley:
I think `expand = c(0, 0, 0.1, 0)` is hard to understand. Can you please add a helper function with named arguments and documentation? I think it might be best to have `mult` and `add` arguments that can take a vector of length 1 or 2.

huftis:
@hadley The reason I used `expand = c(0, 0, 0.1, 0)` is for backwards compatibility, so that all old code would continue to work. But I guess converting the `expand` argument to a list could also work, `expand = list(mult=c(0, .1), add=c(0,0))`. Or are you thinking that I should add a helper function that takes a `mult` and `add` argument and converts it into the new `c(m1, a1, m2, a2)` syntax? (BTW, any suggestion for the name of this function, given that `expand_range()` and `expand_limits()` are both taken?)

hadley:
Yes, a helper function (possibly returning an S3 object or just a vector). I don't know what to call it either 😞

hadley:
Maybe `expand_scale()`?  Or `expansion()`?

dzion:
Has this in some version been implemented yet? I generally agree with @huftis on the usefulness in bar plots, but also with @hadley on the inelegance of the solution provided here.... 
May I suggest to solve this analogous to the `limits()` solution, where `NA` can be specified as one of the parameters? 
Providing `NA` would allow for R to calculate one of the `expand()` parameters, while the other one could be provided. Unfortunately I am pretty busy right now otherwise I would take a stab at this myself.

hadley:
Are you still interested in this PR? I'm going to be working on ggplot2 off and on for a bit.

huftis:
@hadley Yes, I’m still interested; I’ve just been too busy to take a look at it (or, really, it’s been too low on my priority list for too long). I hope to have a new go at it sometime in the next few weeks. When are you planning a test/beta release of the next `ggplot2` version? I’ll try to finish it before that, so that it can be properly tested.

huftis:
@hadley, I have now added an `expand_scale()` function for specifying the expansion values, along with documentation and and couple of examples.

One thing I’m not entirely happy with regarding this patch, is that the defaults for `expand_scale()` don’t duplicate the defaults of `mult = .05, add = 0’ for continuous scales and `mult = 0, add = 0.6` for discrete scales. It’s not a problem functionality-wise; the defaults do continue to work as before. But it’s not possible change the ‘Usage’ section of the documentation for the scales from

    scale_x_discrete(..., expand = waiver(), position = "bottom")

to the more user-friendly

    scale_x_discrete(..., expand = expand_scale(), position = "bottom")

since `expand_scale()` can’t know if it’s a continuous or discrete scale that is being used (and adjust its defaults based on this). I’m not sure if it is possible to fix this.

hannes101:
Could this also be implemented for area plots, since especially when using a color scheme with a light color at the bottom it gives the impression of a much bigger area and additionally it kind of wastes space in a publications.
Thanks a lot in advance.
 Just a quick'n'dirty example showing the "problem".
```
 ggplot(melt(dt id.vars = c("Year.and.Quarter"), measure.vars = colnames(dt)[-1])[like(variable,"Gas"),], aes( x = Year.and.Quarter, y = value)) +
        geom_area(aes(fill= variable), position = 'stack') +
        scale_x_yearqtr(format = "%Y-Q%q",n = 8, expand = c(0,0)) +
        theme_bw() +
        scale_fill_brewer(name = "", labels =c(  "Directional"
                                                 , "Horizontal "
                                                 , "Unknown"
                                                 , "Vertical"),direction = -1, palette = my.color.scheme) +
        labs(title = "", x = "Time", y = "Quarterly Gas Production in mmcf ") +
        theme(text=element_text(family="CMU Serif", size=24)
              , strip.text.y = element_text(family="CMU Serif", size=14, colour = "black")
              , legend.position="bottom"
              , plot.margin = unit(c(1,1.5,1,1), "cm")) 
```
grafik

R/utilities.r
```diff
@@ -156,6 +156,77 @@ rescale01 <- function(x) {
...
+#' # when the ‘expand’ argument is not specified
+#' ggplot(subset(diamonds, carat > 2), aes(cut, price)) +
+#'   geom_jitter() +
+#'   scale_x_discrete(expand = expand_scale(add = .6)) +
+#'   scale_y_continuous(expand = expand_scale(mult = .05))
+expand_scale = function(mult = 0, add = 0) {
+  stopifnot(is.numeric(mult) && is.numeric(add))
+  stopifnot((length(mult) %in% 1:2) && (length(add) %in% 1:2))
+
+  mult = rep(mult, length.out = 2)
```
hadley: Can you please use `<-` here

R/utilities.r
```diff
@@ -156,6 +156,77 @@ rescale01 <- function(x) {
...
+  add = rep(add, length.out = 2)
+  expand = c(mult[1], add[1], mult[2], add[2])
```
hadley: No need for `expand` variable

huftis:
@hannes101 This feature works with scales in general, so yes, it should work with area plots too.

huftis:
@hadley I’ve now fixed the issues you mentioned and updated the PR to cleanly apply to `master`.

R/utilities.r
```diff
@@ -172,6 +172,76 @@ rescale01 <- function(x) {
   (x - rng[1]) / (rng[2] - rng[1])
 }
 
+#' Similar to expand_range(), but taking a vector ‘expand’
```
hadley: Let's `@noRd` this

hadley:
Thanks for all your hard work on this!

PR #1831, opened 2016-10-05
link to geom_count from geom_point help

pierucci:
It could be useful to link `geom_count()` from the `geom_point()` help, since this geom provides a nice solution to overplotting but is not easy to find.


hadley:
Thanks!

PR #1845, opened 2016-10-10
geom_text() "draw rectangle around"

lindbrook:


hadley:
Technically the underneath bit is most important as it blocks out data underneath the text

lindbrook:
Ahh. Underneath in terms of layers. At a glance it reads like "underline".
How about "behind"?

hadley:
Oh yeah, that's better

hadley:
Thanks!

PR #1857, opened 2016-10-17
Coord flip free x

### LINKED ISSUE #1848: issue with coord_flip() + facet_wrap() + free scales in one direction only
Thank you so much to all the contributors for all the work on this package!

I have run into an issue when using `scales = "free_x"` or similar in just one direction with `coord_flip()` and `facet_wrap()`.

If I do something like this:
```
diamonds %>% 
    group_by(color, cut, clarity) %>% 
    summarise(median_carat = median(carat)) %>%
    ggplot(aes(color, median_carat, fill = cut)) +
    geom_bar(stat = "identity", position = "dodge") +
    coord_flip() +
    facet_wrap(~clarity, scales = "free")
```
everything looks great and I get the result I would expect.
unnamed-chunk-1-1

If I just want to have the scales free on one axis (because all those things are the same on the y-axis):
```
diamonds %>% 
    group_by(color, cut, clarity) %>% 
    summarise(median_carat = median(carat)) %>%
    ggplot(aes(color, median_carat, fill = cut)) +
    geom_bar(stat = "identity", position = "dodge") +
    coord_flip() +
    facet_wrap(~clarity, scales = "free_x")
```
then there is a problem. The data is getting plotted correctly, but the labels on the x-axis at the bottom are wrong.
unnamed-chunk-2-1

It's all the same labels.

thomasp85:
Fixes #1848


thomasp85:
Just to reiterate - `CoordFlip` is a point-of-failure for a lot of code since it cannot take care of itself (there need to be created special cases for it in numerous places). It would be very, very good to find an alternate solution going forward...

hadley:
If it's causing a lot of hassle, I think it's ok to back out the original change

thomasp85:
You misunderstand me - this is not necessarily related to the facet changes, but the fact that CoordFlip is breaking a lot of expectations about how objects behaves and interacts due to how it is implemented and therefore it is necessary to have multiple tests for whether the plot coord is CoordFlip scattered around the code. I can't see how we can easily remedy this, as it is not obvious how CoordFlip can be reimplemented, but it should be considered going forward...

thomasp85:
One possibility is to give the Coord object more power in modifying the other ggproto objects in the plot at specific hooks, somehow making CoordFlip be responsible for it's own mess rather than expecting all surrounding code to treat it specially... This will require some work though...

hadley:
Oh ok, got it - you're thinking broadly about how coords _should_ interface with facetting and scales. It probably means moving more logic out of coords and in to the layout code, since that better captures the intersection of coord, scale, and facet.

thomasp85:
Exactly - we'll probably catch most related bugs  before release but I think it's a  design failure we should fix down the line

thomasp85:
There is a danger in moving too much code into `Layout` as this class/object is not extensible. Removing too much from `Coord` will render it unextensible, so I think that additional hooks orchestrated by `Layout` would be the preferable route to take...

hadley:
Thanks!

PR #1868, opened 2016-10-23
Let scales modify title

thomasp85:
This PR is mainly justified by the [units](https://github.com/edzer/units) developers wish to have build in support for the units class in ggplot2. While the scale implementation will be part of [ggforce](https://github.com/thomasp85/ggforce) (see the discussion in thomasp85/ggforce#8) some small changes are needed in ggplot2 itself in order to intercept the default scale titles before they are rendered…


hadley:
LGTM

I think this would also make it possible to have a scale that automatically divided by 1000, 100000 etc to make the values more readable.

That makes me wonder if the transformer should some how be involved.

hadley:
Can you add a bullet to NEWS and add to the Scale description?

thomasp85:
Which scale description? The man page?

I also think this opens up for some niceties in the scales such as what you describe but this is probably best left for the next release/intern

hadley:
Yeah, the man for the base class.  

Agreed that the rest can wait until the next release.

thomasp85:
If you look at the documentation level for `Scale` in ggplot2-ggproto.Rd I think we can agree that a mention there would seem out of place unless the whole section gets an update...

hadley:
Oh yeah, let's leave it for now. LGTM

PR #1872, opened 2016-10-24
Dynamically access ggproto super class

### LINKED ISSUE #1826: New error triggered in compute_group
In both 'ggpmisc' and 'ggspectra' I was getting a rather cryptic error triggered in my compute_group functions.

**Error in as.vector(y) : attempt to apply non-function**

With the following traceback

```
4. as.vector(y) 
3. setdiff(names(mapping), c(geom$aesthetics(), stat$aesthetics())) at layer.r#114
2. ggplot2::layer(stat = StatDebugGroup, data = data, mapping = mapping, 
    geom = geom, position = position, show.legend = show.legend, 
    inherit.aes = inherit.aes, params = list(na.rm = na.rm, summary.fun = summary.fun, 
        summary.fun.args = summary.fun.args, ...)) at stat-debug-group.R#54
1.
stat_debug_group() 
```
It can be easily fixed by adding `force(data)` as the first statement in my compute_group function. I haven't seen this error the previous time I tested with 'ggplot2' from Github (some weeks ago).

This does not necessarily need a fix, but thought good to document it here in case someone else faces some problem.

hadley:
Fixes #1826 

@wch could you please review?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Major new features
 
-## Subtitle and caption
+### Subtitle and caption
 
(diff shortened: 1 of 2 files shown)


wch:
LGTM

PR #1882, opened 2016-10-29
add room around guide title

### LINKED ISSUE #1881: legend titles need more padding
Not sure if it's a recent change (I'm using the dev version fresh from github), but with a large enough font the legend title collides with the legend and I  can't find any setting to move it.

MRE

``` r
set.seed(123)
dummy <- data.frame(x=runif(10), y=runif(10), f = gl(2,5))
ggplot(dummy, aes(x,y, shape=f)) + 
  geom_point() +
  labs(shape="guide") +
  theme_grey(24)
```

With the default font size things are OK, albeit snug

screen shot 2016-10-29 at 4 58 42 pm

With a larger font size (graphic aimed for slides), however, 

screen shot 2016-10-29 at 4 58 34 pm

I tried,

``` r
legend.margin = margin(100,100,100,100)
legend.spacing.y = unit(1,"cm")
legend.title = element_text(vjust = 2)
```

also played with guide_legend etc. to no avail.

baptiste:
Fixes #1881 but other theme parameters seem to be currently ignored


hadley:
Does this fix the problem or do you still need to manually add the height of the descenders?

baptiste:
The descenders issue is minor I think, and pretty hard to work around. I don't think it's worth it, it should be fixed at the grid level (it's listed on Paul Murrell's todo list).

What this PR does is replace the hard-coded "vgap" between the guide title and the guide by the relevant setting in theme. That way the user regains control when things look bad. The default value is also changed to better scale with the font size; I used half the grob height. That's perhaps not ideal as titles with a descending letter will have a slightly offset baseline compared to those that don't, but I don't think it's ever a noticeable problem. Better than the current situation where the spacing is absolute and hard-coded.

What I did notice, and didn't change, is a lack of consistency with `guide-colourbar.r`, a few inconsistencies in size conversions, and other theme parameters apparently not linked to the guide.

baptiste:
(obviously the latest commits are not relevant to this PR -- not sure why they got included here)

hadley:
Do we still need this? I can't remember if this is before or after I added the descenders into the height calculation.

baptiste:
Currently it looks like this (cf original issue)

screen shot 2017-01-25 at 12 23 36 pm

hadley:
The key change seems to be `R/utilities-grid.r` to correctly used `convertHeight` instead of `convertWidth`. @karawoo can you please pull out that one change?

@karawoo can you please also make a note to review these margins once we're done with the `titleGrob()` changes? We should definitely test the theme with a much larger than default font size.

hadley:
Yes please!

PR #1890, opened 2016-11-03
stat_bin accepts functions for binwidth

botanize:
Allowing binwidth to accept functions greatly improves histograms of faceted variables. Furthermore, the current default (30 bins) is somewhat arbitrary, and could be replaced with one of the three built-ins included here.

The built-ins include 'Sturgis', 'Scott', and 'FD'. These match the options included in [d3](https://github.com/d3/d3-array/blob/master/README.md#histogram-thresholds). More information is available on [wikipedia](https://en.wikipedia.org/wiki/Histogram#Number_of_bins_and_width).

Examples are included in the pull-request, here's a motivating example:
```{r}
mtlong <- reshape2::melt(mtcars)
ggplot(mtlong, aes(value)) + geom_histogram() + facet_wrap(~variable, scales = 'free_x')
ggplot(mtlong, aes(value)) + geom_histogram(binwidth = 'Sturges') + facet_wrap(~variable, scales = 'free_x')
````

### CODE REVIEW

**Comment by hadley** [comment]
I wouldn't mind extending bin to accept functions but I really dislike Sturgis/Scott/FD — I think they give the illusion that you're picking a good binwidth, but they're just as arbitrary as picking 30 bins.

**Comment by botanize** [comment]
None of the options are a default, they have to be chosen manually, which is still a level of intention beyond 30 bins. 

Would you prefer to drop the built-in options entirely?

**Comment by hadley** [comment]
Yes please

**Comment by botanize** [comment]
Is there anything else preventing you from accepting this pull request?

File: R/stat-bin.r [hunk 1]
```diff
@@ -111,9 +113,10 @@ StatBin <- ggproto("StatBin", Stat,
       stop("Only one of `boundary` and `center` may be specified.", call. = FALSE)
     }
 
-    if (is.null(params$breaks) && is.null(params$binwidth) && is.null(params$bins)) {
+    if (is.null(params$breaks) && (is.null(params$binwidth) || !(is.numeric(params$binwidth) || is.function(params$binwidth))) && is.null(params$bins)) {
```
hadley: Why does this need to be changed? If a function is supplied `binwidth` will no longer be `NULL`

R/stat-bin.r
```diff
@@ -130,8 +133,11 @@ StatBin <- ggproto("StatBin", Stat,
     if (!is.null(breaks)) {
       bins <- bin_breaks(breaks, closed)
     } else if (!is.null(binwidth)) {
-      bins <- bin_breaks_width(scales$x$dimension(), binwidth, center = center,
-        boundary = boundary, closed = closed)
+      if (!is.numeric(binwidth)) {
```
hadley: It would be better to check for `is.function()` here

R/stat-bin.r
```diff
@@ -130,8 +133,11 @@ StatBin <- ggproto("StatBin", Stat,
...
+        binwidth <- do.call(binwidth, list(data$x))
```
hadley: I don't think you need `do.call` here

tests/testthat/test-stat-bin.R
```diff
@@ -22,6 +22,14 @@ test_that("bins specifies the number of bins", {
   expect_equal(nrow(out(bins = 100)), 100)
 })
 
+test_that("binwidth computes widths for function input", {
+  df <- data.frame(x = 1:100)
+  out <- function(x, ...) {
```
hadley: I think you can remove the indirection of this function now. 

And a simpler binwidth function would just return a constant.

R/stat-bin.r
botanize: Thanks, that was leftover from checking for a string input as well.

R/stat-bin.r
botanize: Got it.

tests/testthat/test-stat-bin.R
botanize: Ok, I get the simpler function that just returns a constant, but I don't know what "indirection of this function" means.

NEWS.md
```diff
@@ -16,6 +16,10 @@
 
 ## Major new features
 
+### Binning
```
hadley: Can you please move this up a section?

tests/testthat/test-stat-bin.R
hadley: Sorry, that was a bit unclear. I meant eliminate the `out` function:

```R
out <- layer_data(ggplot(...))
```

hadley:
Thanks!

PR #1895, opened 2016-11-06
Change wrong package name in README file

delferts:
When showing how to install ggplot2 package there is line 

> # Alternatively, install just readr

 but should be 

> # Alternatively, install just ggplot2


hadley:
Thanks! For future reference, you don't need to describe simple changes in such detail since I can see exactly what's changed in the diff.

PR #1912, opened 2016-11-16
Suspected typo / thinko: should use `axis.text` instead of `axis.ticks`.

mvkorpel:
This pull request changes two document files to match what is said in [`theme.r`](https://github.com/tidyverse/ggplot2/blob/master/R/theme.r#L330). I believe the warning message in the code is correct and the documents have a typo / thinko.


hadley:
Thanks!

Issue #1902, opened 2016-11-13
Ordering within facets


drsimonj:
[This post](https://drsimonj.svbtle.com/ordering-categories-within-ggplot2-facets) discusses a method for differently ordering categories that appear in multiple facets with different values on a continuous variable. As far as I'm aware, there isn't a simple method for achieving this, and @dgrtwo suggested it could be useful for others to have a neat wrapper function or new stat option to handle it.

Here's a small reproducible example to explain the problem:

``` r
library(ggplot2)
library(dplyr)

d <- data.frame(
  f = rep(c("f1", "f2"), each = 3),
  x = rep(c("a", "b", "c"), 2),
  y = c(1, 2, 3, 3, 2, 1)
)

# No easy way to order categories within facets
ggplot(d, aes(x, y)) +
  geom_col() +
  facet_wrap(~ f)
```

![unnamed-chunk-1-1](https://cloud.githubusercontent.com/assets/11235783/20249159/91f185a4-aa47-11e6-9b94-9085cf1082a2.png)

``` r
# This will handle the ordering
d <- d %>% 
  ungroup() %>%   # As a precaution / handle in a separate .grouped_df method
  arrange(f, y) %>%   # arrange by facet variables and continuous values
  mutate(.r = row_number()) # Add a row number variable

ggplot(d, aes(.r, y)) +  # Use .r instead of x
  geom_col() +
  facet_wrap(~ f, scales = "free") +  # Should free scales (though could be left to user)
  scale_x_continuous(  # This handles replacement of .r for x
    breaks = d$.r,     # notice need to reuse data frame
    labels = d$x
  )
```

![unnamed-chunk-1-2](https://cloud.githubusercontent.com/assets/11235783/20249161/9b346884-aa47-11e6-89ce-8b89e507697e.png)

Although this example handles bars, it seems generally applicable to other geoms (e.g., points). Some options for making this readily accessible for users:

- Include a wrapper function for the data ordering, and leave the plotting up to them.
- Add an option in the `facet` functions.
- Add something like an `order_by` argument to `scale_x/y_discrete`

I've created [new fork](https://github.com/drsimonj/ggplot2) to see what works.

has2k1:
@drsimonj, I am surprised that your solution works. The scales do not know about the partitioning of the data done by the facets. So it is by happenstance that when you pass full size vectors as `breaks` and `labels` that they align with the parts created by the facets.

drsimonj:
[...]

PR #1917, opened 2016-11-18
Implementing an order aes

drsimonj:
As discussed in #1902 , this PR reintroduces the `order` aes, that will order discrete variables along the x and/or y axis based on a numeric variable.

Ordering data for visualization is a widely used/desired feature that must be handled prior to plotting. This PR is intended to provide the functionality for users to implement it directly while plotting.

Unlike the previous implementation of `order` (which was dropped), the solution proposed here is to convert discrete variables to factors with levels ordered by a group-wise summary statistic of the numeric order variable. The statistic will differ depending on the geom. For example, `sum` is used for `geom_col`, whereas `max` is used for `geom_point` or `geom_tile`.

Some examples:

```r
    # No ordering
    ggplot(mpg, aes(manufacturer, hwy)) + geom_col()
```

fig-unnamed-chunk-2-1

```r
    # Ordering by same variable as y-axis
    ggplot(mpg, aes(manufacturer, hwy, order = hwy)) + geom_col()
```

fig-unnamed-chunk-2-2

```r
    # Ordering reversed
    ggplot(mpg, aes(manufacturer, hwy, order = -hwy)) + geom_col()
```

fig-unnamed-chunk-2-3

```r
    # With points
    ggplot(mpg, aes(manufacturer, hwy, order = hwy)) + geom_point()
```

fig-unnamed-chunk-2-4

```r
    # Ordered by a different continuous variable and axis
    ggplot(mpg, aes(hwy, manufacturer, order = displ, color = displ)) + geom_point()
```

fig-unnamed-chunk-2-5


has2k1:
Nice. Here is a thought. How about instead of `order` as an aesthetic, the same effect is controlled by a `layer` parameter? That would introduce a new layer parameter, for example `order_by`. The default value would be `NULL` or `waiver()` and it could take any of the aesthetic values.

The ordering can be inserted into the building pipeline, right after [here](https://github.com/tidyverse/ggplot2/blob/e36e71a3bee6a34e29b74231a766d4e67a0d2942/R/plot-build.r#L60). The advantage of this over the current implementation is that the final (more or less) values of all the aesthetics would be in place and there would be no need of such special casing.

```diff
+  if (inherits(geom, "GeomCol")) {
+    data <- order_by(data, can_order, sum, na.rm = na.rm)
+  }
+  else if (inherits(geom, "GeomPoint")) {
+    data <- order_by(data, can_order, max, na.rm = na.rm)
+  }
+  else if (inherits(geom, "GeomTile")) {
+    data <- order_by(data, can_order, max, na.rm = na.rm)
+  }
+  else {
+    warning("`order` aesthetic not supported for this Geom")
+  }
```
The ordering would always be ascending/descending and if the user wants the opposite they just do
`order_by=-y`.

drsimonj:
Nice idea. I can see how the data can be ordered this way. But not sure how to order scales too. Any thoughts?

has2k1:
That is the difficult part and I haven't yet gotten that far in thinking about it. The motivation behind the previous suggestion was purely to avoid the `geom` specific special cases, while still being able to achieve the same results as they stand.

The scales really become a menace when facets are added. My hunch is that a solution exists somewhere in this [region](https://github.com/tidyverse/ggplot2/blob/e36e71a3bee6a34e29b74231a766d4e67a0d2942/R/plot-build.r#L75-L86). May be we could differ the ordering of the data up to that point, so that a single call to a layer function would deal with the data ordering and the scale ordering.

hadley:
It's not ok for an aesthetic to affect the scales in this way - it's a violation of the separation between the components of the grammar.

thomasp85:
I would tentatively suggest writing a new facet class to deal with this. Facets have acces to the data of each panel and can modify the scales for each panel...

Doesn't fit within base ggplot2 IMO though

drsimonj:
@hadley are you also referring to layer parameter suggestion by @has2k1 as not being appropriate?

And thanks @thomasp85. I'll see if something like this is possible too.

hadley:
Yes

drsimonj:
Seems a new facet will work, but agree with @thomasp85 that this doesn't fit. E.g., what if we want to order a single panel like in examples above? Seems odd to handle this in `facet_*`.

**NEW IDEA - ORDER GGPROTO**

Thoughts on a new plot parameter, `order`, added [here](https://github.com/tidyverse/ggplot2/blob/master/R/plot.r#L94-L103), that accepts a new ggproto? This would involve a base `Order` class that is similar to `Facet`, but specially tasked to manipulate data and scales around [here](https://github.com/tidyverse/ggplot2/blob/e36e71a3bee6a34e29b74231a766d4e67a0d2942/R/plot-build.r#L61). It might look something like:

```r
Order <- ggproto("Order", NULL,

    # Data frame containing information needed to map old values to new order.
    # Will be set/used by subsequent methods.
    order_map = NULL,  

    train_map = function(order_aes, data) {
        # Train order_map
    },

    order_data = function(self, data) {
        # Order data layers
    },

    order_scale = function(self, scales/layout) {
        # Order scales
    }

)
```

Implement functions like:

```r
order_by <- function(order_aes, facets = c("all", "within", "between"), ...) {

    # Do relevant checks etc
    # E.g., that order_aes refers to a valid aesthetic param (x,y,...)
    # Or to match.arg for facets etc

    ggproto(NULL, Order,
        params = list(order_aes = order_aes, facets = facets)
    )
}
```

Use like:

```r
ggplot() + geom_*() + order_by(y)
```

has2k1:
The recent `facet` refactoring was supposed to satiate all our needs when it came to extending ggplot, but as this case shows that there are still gaps. A separate `order` object does not solve this general problem.

I think a more general base `ggproto` object that users can extend would be better. If present, then the solution to ordering can be an example of how to extend that base object and can be included in either `ggplot` or another package like [ggforce](https://github.com/thomasp85/ggforce). @thomasp85, any input?

hadley:
I'm sorry you've spent so much time on this, but IMO this is a clear violation of the grammar. The display (including ordering) of the axes is a property of the scale and should not be handled elsewhere.   ggplot2 is a complex system and an adding entirely new component to the grammar is not worth the additional complexity for this problem which can already be solved in at least two other ways.

drsimonj:
@hadley no worries - was interesting to think about. Thanks for considering.

PR #1924, opened 2016-11-22
alpha affects text and border in geom_label

jiho:
Alpha is useful to hide plot elements from view. So all pieces of the label should be hidden. I don't see a use case for having only the background of the label transparent.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 2.2.0.9000
 
+* `alpha` affects text and border in `geom_label()` (@jiho, #1924)
+
 # ggplot2 2.2.0
 
 ## Major new features
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -83,14 +83,14 @@ GeomLabel <- ggproto("GeomLabel", Geom,
         padding = label.padding,
         r = label.r,
         text.gp = gpar(
-          col = row$colour,
+          col = alpha(row$colour, row$alpha),
           fontsize = row$size * .pt,
           fontfamily = row$family,
           fontface = row$fontface,
           lineheight = row$lineheight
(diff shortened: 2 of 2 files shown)


hadley:
The convention in ggplot2 is that for geoms that have both fill and colour, alpha only affects fill.  I looked into changing this convention a while back, but it ended up causing more problems than it solved.

jiho:
Couldn't an argument be made that, for `geom_label` the text at least is conceptually part of the "inside" (the outline being the "outside")? If so, I'll modify the code to affect only the text.

My use case for this: I want to use `geom_label` on a PCA biplot, where it sometimes helps readability compared to `geom_text`; but even then those plots are often crowded and something that ggplot makes easy is to make objects not well represented by the PCA semi-transparent. That works with `geom_text` but not `geom_label` where only the background is transparent.

hadley:
As long as you're not mapping the colour, you can always set it to a semi-transparent colour with `alpha()`.

jiho:
I am mapping the colour ;-) to the cos2 in that particular case, which quantifies the representativity of the objects in the projected PCA plane.

But, beyond that particular use case, I think that, in any situation where one would map the `alpha` of the text in `geom_text` and finally chose `geom_label` for aesthetical/legibility reasons, one would want the `alpha` to apply to both the rectangle and the text.

jiho:
BTW, in `geom_point`, `alpha` affects both fill and stroke ;-)

This is noticeable now, with the shift from shape=16 (no stroke) to shape=19 (stroke) as the default.

```r
library("ggplot2")
qplot(runif(10), runif(10), size=I(5), alpha=I(0.3), stroke=I(3))
```

screen shot 2017-01-25 at 15 43 23

PR #1925, opened 2016-11-23
Update README.Rmd

arneschillert:
Fix typo


hadley:
Thanks - but looks like I fixed it in the mean time.

PR #1934, opened 2016-11-29
Remove `x` as a required aesthetic for `geom_errorbarh`

### LINKED ISSUE #1933: Inconsistency in required aesthetics between errorbar and errorbarh
In [the geom_errorbar code](https://github.com/tidyverse/ggplot2/blob/master/R/geom-errorbar.r),
the required aesthetics are
`required_aes = c("x", "ymin", "ymax"),`

However, in [the geom_errorbarh code](https://github.com/tidyverse/ggplot2/blob/master/R/geom-errorbarh.r),
the required aesthetics are
`required_aes = c("x", "xmin", "xmax", "y"),`

In other words, the `x` aesthetic is required for both plots, but the `y` aesthetic is required only for `errorbarh`.

I believe the correct answer would be to drop the `x` requirement from `errorbarh`

metanoid:
Closes #1933 

`errorbar` is able to plot without putting a point at the midpoint of the error bar, but `errorbarh` was not able to.

An outstanding issue - It seems the `roxygen2`-generated documentation now does not pick up that `x` is an aesthetic that `geom_errorbarh` understands.

The result of this is that the user will need to use `geom_errorbarh` as well as `geom_point`.  I'm not sure how to fix this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -351,6 +351,8 @@ There were a number of tweaks to the theme elements that control legends:
 
 * Fixed issue where `coord_map()` fails when given an explicit `parameters`
   argument (@tdmcarthur, #1729)
+  
+* Fixed issue where `geom_errorbarh()` had a required `x` aesthetic (#1933)  
 
 # ggplot2 2.0.0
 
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #1951, opened 2016-12-08
clarify geom_smooth auto message

davharris:
Currently, `stat_smooth`'s "auto" method reports the function that it chooses, but not the formula. Since the formula depends on the function it chooses, reporting it seems like a good idea.

This is already in the docs, but putting it in the message will make it less likely for users to forget that they need to rewrite the formula (and tell their boss that the splines are just very close to being linear 😳).


hadley:
Can you please add a bullet to NEWS.md too?

davharris:
Done. Let me know if you need anything else :-)

hadley:
Thanks!

Issue #1928, opened 2016-11-27
facet_grid(a ~ a) produces an error


vnijs:
FYI The example from #1153 now produces the error below and does not create a plot

```r
ggplot(data.frame(a = letters[1:3], y = rnorm(3))) + geom_point(aes(y, y)) + facet_grid(a~a)

> Error in scale_apply(layer_data, x_vars, "train", SCALE_X, x_scales) : 
```

hadley:
@thomasp85 can you please take a look?  I don't think we need to make this work, but it should give a more informative error message.

PR #1973, opened 2016-12-30
Fix bug: facet_grid(a ~ a) produces an error

EmanuelContreras:
Hi.  I read this bug: #1928
[facet_grid(a ~ a) produces an error #1928](https://github.com/tidyverse/ggplot2/issues/1928)
And i want to contribute for its solution.
The results:
`ggplot(data.frame(a = letters[1:3], y = rnorm(3))) + geom_point(aes(y, y)) + facet_grid(a~a)`

rplot01

Thanks


hadley:
This doesn't feel like the right fix to me, but we'll take a look at the underlying problem.

PR #1976, opened 2017-01-01
simplify affine transformation code

edzer:
only simplifies the affine transformation expression; the aspect ratio still doesn't work out.


hadley:
Ah, perfect, thanks!  I'll merge manually since I've introduced some conflicts

PR #1990, opened 2017-01-09
Add support for sf [WIP]

hadley:


R/utilities.r
```diff
@@ -104,24 +104,40 @@ remove_missing <- function(df, na.rm = FALSE, vars = names(df), name = "",
...
     # Find all the rows where all are TRUE
-    rowSums(as.matrix(finite_cases)) == ncol(x)
+    rowSums(as.matrix(ok)) == ncol(x)
+  }
+}
+
+# Wrapper around is.finite to handle list cols
+is_finite <- function(x) {
+  if (typeof(x) == "list") {
+    !vapply(x, is.null, logical(1))
```
cpsievert: should this be `!vapply(x, is.finite, logical(1))`?

hadley: Anything could be inside that list (like another list) - being finite doesn't really have a meaning for arbitrary elements so this just drops rows where the list-col is empty.

cpsievert: Ahhh, yea, makes sense

PR #1996, opened 2017-01-13
Reproducible jitter

krlmlr:
by adding a `seed` argument to `position_jitter()`. Useful to apply the same jitter to both points and labels.

We should probably move `with_seed()` to `withr` before merging.

Reprex:

``` r
library(ggplot2)
ggplot(data.frame(x = 1, y = 1:10, text = letters[1:10]), aes(x = x, y = y)) +
  geom_point(position = position_jitter()) +
  geom_label(aes(label = text, y = y - 0.5), position = position_jitter())
```

``` r

ggplot(data.frame(x = 1, y = 1:10, text = letters[1:10]), aes(x = x, y = y)) +
  geom_point(position = position_jitter(seed = 1L)) +
  geom_label(aes(label = text, y = y - 0.5), position = position_jitter(seed = 1L))
```


digital-dharma:
@krlmlr - Great enhancement.  As soon as this gets accepted, we'll plan to use it.

R/position-jitter.r
```diff
@@ -48,13 +52,14 @@ PositionJitter <- ggproto("PositionJitter", Position,
...
     )
   },
 
   compute_layer = function(data, params, panel) {
-    trans_x <- if (params$width > 0) function(x) jitter(x, amount = params$width)
-    trans_y <- if (params$height > 0) function(x) jitter(x, amount = params$height)
+    trans_x <- if (params$width > 0) function(x) with_seed(params$seed, jitter(x, amount = params$width))
+    trans_y <- if (params$height > 0) function(x) with_seed(params$seed, jitter(x, amount = params$height))
 
     transform_position(data, trans_x, trans_y)
```
hadley: I think the `with_seed()` should probably go here, otherwise the x and y jitter will use the same random values

R/utilities.r
```diff
@@ -299,6 +299,16 @@ find_args <- function(...) {
 # global data
 dummy_data <- function() data.frame(x = NA)
 
+with_seed <- function(seed, code) {
+  if (!is.null(seed)) {
+    old_seed <- get0(".Random.seed", globalenv())
```
hadley: Is this definitely ok to do?

krlmlr: @yihui: knitr seems to cache/restore .Random.seed, too. The caching in knitr is explicit, but it looks like the restore happens implicitly. Could you please comment if my way of restoring the random seed is safe?

```r
with_seed <- function(seed, code) {
  if (!is.null(seed)) {
    old_seed <- get0(".Random.seed", globalenv())
    if (!is.null(old_seed)) {
      on.exit(assign(".Random.seed", old_seed, globalenv()), add = TRUE)
    }
  }
  code
}
```

yihui: I think it is safe (I'd add `mode = 'integer'` to `get0()`, though), but it seems that you forgot to `set.seed(seed)` before `code` is evaluated?

krlmlr: Good catch. The code worked by accident because *resetting* the random seed after applying the jitter also made it reproducible.

hadley:
Could you please also add that example to the examples in the doc?

I also wonder if `position_jitter()` should be reproducible "by default", i.e. we should always default to `seed = sample(1e6)` (or whatever the correct value is)

digital-dharma:
@hadley @krlmlr - I agree with Hadley's suggestion.  The ideal behavior would be a default seed and consistent jittering.  Thanks guys!

jimhester:
👍  to this PR, also the correct value is [42](https://simple.wikipedia.org/wiki/42_(answer)).

krlmlr:
@jimhester: Some think it's [4](https://xkcd.com/221/).

hadley:
I still think that the seed should be random each time you call geom_jitter - you should have to explicitly opt in to reusing the seed by constructing your own position_jitter (or setting the seed). 

Otherwise you'll get exactly the same offsets every single time. That's not terrible but it feels like a different geom to jitter (and you could just use an existing precomputed pseudo random sequence)

krlmlr:
It's difficult to achieve both: consistent jitter by default, and jitter dependent on random seed. I have added a mode (now default) that reuses the current random seed and resets it afterwards.

hadley:
I think my original proposal to just pick a random seed would be simpler.

krlmlr:
@hadley: Do you mean

```r
position_jitter <- function(..., seed = sample.int(2147483647L, 1L))
```

? I suspect this isn't going to work as expected, two consecutive `position_jitter()` calls still will use a different initial random seed.

hadley:
Right, that's what I mean. If you want to reuse the same jitter, you'll need to share a `position_jitter()` object.

krlmlr:
Fixed and clarified example, the default now picks a random seed and advances the RNG.

hadley:
Bump

PR #2012, opened 2017-01-24
Improve extensibility of coords

hadley:
Primarily by adding `setup_params()` and `setup_data()` methods that are called from `Layout$setup()`. This allows new coords (like `CoordSf`) to make decisions after looking globally at all the layers (`CoordSf` needs this in order to determine a common CRS across all layers)

This also includes some refactoring of the `Layout`.


hadley:
@thomasp85 would be great if you could have a look.

R/coord-.r
```diff
@@ -71,13 +81,25 @@ Coord <- ggproto("Coord",
...
+  train = function(scale_details, params = list()) NULL,
 
   transform = function(data, range) NULL,
 
   distance = function(x, y, scale_details) NULL,
 
-  is_linear = function() FALSE
+  is_linear = function() FALSE,
+
+  setup_params = function(data) {
```
thomasp85: I think this method should be passed in the params given during construction?

hadley: I don't think this - this works like the `setup_params` for position, not like the other classes. It's all a bit inconsistent, but I can't see a way to do better without a huge amount of work.

thomasp85: fair enough

R/layout.R
```diff
@@ -123,13 +184,14 @@ Layout <- ggproto("Layout", NULL,
...
+  train_ranges = function(self) {
     compute_range <- function(ix, iy) {
       # TODO: change coord_train method to take individual x and y scales
-      coord$train(list(x = self$panel_scales$x[[ix]], y = self$panel_scales$y[[iy]]))
+      scales <- list(x = self$panel_scales$x[[ix]], y = self$panel_scales$y[[iy]])
+      self$coord$train(scales, self$coord_params)
     }
     # Switch position of all scales if CoordFlip
-    if (inherits(coord, "CoordFlip") || (inherits(coord, "CoordPolar") && coord$theta == "y")) {
+    if (inherits(self$coord, "CoordFlip") || (inherits(self$coord, "CoordPolar") && self$coord$theta == "y")) {
```
thomasp85: It could be fantastic if this change/refactoring could remove the need to treat CoordFlip etc specially - that is - can we somehow create the hooks that will allow each Coord to take care of itself...

thomasp85:
I generally like the direction of this - I'll try to poke at it a bit more when time permits. My current perception is only based on the diff-view

R/layout.R
hadley: What if we introduced `Coord$setup_scales(x_scales, y_scales, params)` ?  I can also make `coord$train` take individual x and y scale objects like the TODO suggests. Those two changes would make this method much cleaner.

thomasp85: I think that's a good start, and maybe all it takes - The litmus test would be if we can remove all Coord inheritance tests from Layout and Facet

hadley: I ended up calling the method `modify_scales()` since it does in-place modification of the scales. I'm not 100% sure this is correct so I'm not planning on documenting as an extension point.

hadley:
I suspect some consistent naming conventions would make Coords much easier to understand.  Currently we're using `scale_details`, `panel_ranges` and `panel_scales` to refer to the same thing in different places. I a better name would be `panel_params` as these are a list of parameters produced by the Coord for each panel. That also suggests that the `train()` method should be called `setup_panel_params()` or similar.

hadley:
And now done - I think this makes it easier to see what the layout is doing.

Another cleanup would be for the panel_params to consistently be a list with contents for `x` and `y`.

Issue #1959, opened 2016-12-16
layer quietly renames certain parameters


nteetor:
Motivated by exercise 2 under section "Creating a new stat" of the "Extending ggplot2" vignette.
> 2. Modify the final version of StatDensityCommon to allow the user to specify the min and max parameters. You’ll need to modify both the layer function and the compute_group() method.

### case 1 (no warning generated)

Parameters named col, color, pch, cex, lty, lwd, srt, adj, bg, fg, min, or max (see `ggplot2:::.base_to_ggplot`) are renamed without warning when passed to layer(). Values do get passed to methods such as `setup_params()`, but as the renamed names. See the following example,

```R
library(ggplot2)

# create an example Stat which passes along data without
# modification, but outputs information to the console
StatExample <- ggproto(
  'StatExample', Stat,

  setup_params = function(data, params) {
    cat('\n`params` includes:\n')
    print(params)
    params
  },

  # here is where we use ymin, ymax, and fill instead of min, max, and bg
  # to work around the internal renaming
  compute_group = function(data, scales, ymin, ymax, fill, extra) {
    data
  }
)

# create a new stat function
stat_example <- function(mapping = NULL, data = NULL, geom = 'point',
                      position = 'identity', na.rm = FALSE, show.legend = NA,
                      inherit.aes = TRUE, min = 'min arg', max = 'max arg',
                      bg = 'dodgerblue', extra = 'still extra', ...) {
  layer(
    stat = StatExample, mapping = mapping, data = data, geom = geom,
    position = position, show.legend = show.legend, inherit.aes = inherit.aes,
    params = list(na.rm = na.rm, min = min, max = max, bg = bg, extra = extra,
                  ...)
  )
}

d <- data.frame(
  x = 1:5,
  y = 1:5
)

# runs without error, but we knew to name the arguments
# ymin, ymax, and fill, see console output below
ggplot(d, aes(x = x, y = y)) +
  stat_example()
```

console output,
```
> d <- data.frame(
+     x = 1:5,
+     y = 1:5
+   )
> ggplot(d, aes(x = x, y = y)) +
+     stat_example()

`params` includes:
$na.rm
[1] FALSE

$ymin
[1] "min arg"

$ymax
[1] "max arg"

$fill
[1] "dodgerblue"

$extra
[1] "still extra"
```

### proposal 1

[...]

PR #2015, opened 2017-01-29
ggproto vignette note about renamed layer parameters

nteetor:
In response to #1959.

I have added a note about renamed layer parameters as part of the question text in question 2 of section 1. This seemed logical as I encountered the issue when trying to solve this question. If there is a better place for the note, I can make the adjustment.


hadley:
Perfect, thanks!

(In the future please include `Fixes #1959` in the text of one of the commit messages so the associated issue gets automatically closed when the PR is merged)

nteetor:
You're welcome, I'll remember about a fixes tag in the future.

PR #2023, opened 2017-02-01
use correct justification for multi-line strip labels

mkuhn:
I ran into the alignment problem reported previously on [StackOverflow](https://stackoverflow.com/questions/34749636/ggplot2-how-to-left-justify-text-from-multi-line-facet-labels), where only a workaround was posted. 

Here is the problem: A left-aligned label with more than one line is still centered within the box.

before

The fix just passes on the justification parameters from `stripGrob` to `textGrob`, resulting in this plot:

after

Note that the standard case still looks fine:

after2

Code for testing:

```
library(ggplot2)

d <- data.frame(x = 1, y = 1, label = c("========\n===", "--"))

ggplot(d, aes(x=x, y=y, label=label)) + geom_text(hjust="left") + facet_grid(~label) +
  theme(strip.text = element_text(hjust=0, vjust = 0.5, debug = T))

```


hadley:
I'm nervous that this will break something else, so let's hold off on merging until the visual tests are up and running again.

hadley:
We're going to tackle this as part of our systematic review of text placement.

@karawoo can you please add to your todo?

karawoo:
This was addressed in #2232. The example above now looks like:

PR #2025, opened 2017-02-01
Fix bug: Margins don't affect strip.text

EmanuelContreras:
With this request you can solve the issue of @Eluvias. For example:
``` python
p <- ggplot(mpg, aes(displ, cty)) + geom_point() + facet_grid(. ~ cyl)
p + theme(strip.text = element_text(margin = margin(l=-40, b = 15,t=15)))
``` 

rplot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,6 @@ r_github_packages:
 after_success:
   - Rscript -e 'covr::codecov()'
 
-before_install:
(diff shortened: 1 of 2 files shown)


hadley:
@EmanuelContreras as with the other PR, we'll tackle this as part of our review of text placement.

@karawoo and another one

Issue #2028, opened 2017-02-02
geom_smooth behaves like geom_path


izahn:
I noticed that `geom_smooth` behaves like `geom_path` rather than `geom_line`. For example,

```
library(ggplot2)

ggplot(transform(mtcars, 
                 pred.mpg = predict(lm(mpg ~ poly(hp, 2), 
                                       data = mtcars), 
                                    interval = "confidence"))) + 
  geom_smooth(aes(x = hp,
                  y = pred.mpg.fit, 
                  ymin = pred.mpg.lwr, 
                  ymax = pred.mpg.upr), 
              stat = "identity")
```

produces
![smooth_example](https://cloud.githubusercontent.com/assets/321235/22558177/cab9539e-e93a-11e6-8738-3364e1467fd4.png)

Whereas ordering by the x-axis variable produces the desired result:

```
ggplot(transform(mtcars, 
                 pred.mpg = predict(lm(mpg ~ poly(hp, 2), 
                                       data = mtcars), 
                                    interval = "confidence"))[order(mtcars$hp), ]) + 
  geom_smooth(aes(x = hp,
                  y = pred.mpg.fit, 
                  ymin = pred.mpg.lwr, 
                  ymax = pred.mpg.upr), 
              stat = "identity")
```
![smooth_example2](https://cloud.githubusercontent.com/assets/321235/22558263/12bcc69e-e93b-11e6-9c4a-382a8b90298e.png)

I looked at the `geom_smooth` code, but I don't understand why it behaves this way. It looks like `GeomSmooth` calls `GeomLine`, and `GeomLine` does 

```{r}
  setup_data = function(data, params) {
    data[order(data$PANEL, data$group, data$x), ]
  }
```
so I would have thought ordering the data myself would make no difference.

hadley:
It probably just calls the draw method, not the `setup_data()` method.

izahn:
Thanks, I've submitted https://github.com/tidyverse/ggplot2/pull/2029 to address this.

PR #2029, opened 2017-02-02
Order path data in GeomSmooth

izahn:
Fixes https://github.com/tidyverse/ggplot2/issues/2028

### DIFF
--- a/R/geom-smooth.r
+++ b/R/geom-smooth.r
@@ -114,6 +114,9 @@ geom_smooth <- function(mapping = NULL, data = NULL,
 #' @usage NULL
 #' @export
 GeomSmooth <- ggproto("GeomSmooth", Geom,
+  setup_data = function(data, params) {
+    GeomLine$setup_data(data, params)
+  },
   draw_group = function(data, panel_params, coord) {
     ribbon <- transform(data, colour = NA)
     path <- transform(data, alpha = NA)
--- a/tests/testthat/test-geom-smooth.R
+++ b/tests/testthat/test-geom-smooth.R
@@ -1,0 +1,10 @@
+context("geom_smooth")
+
+test_that("Data is ordered by x", {
+  df <- data.frame(x = c(1, 5, 2, 3, 4), y = 1:5)
+
+  ps <- ggplot(df, aes(x, y))+
+    geom_smooth(stat = "identity", se = FALSE)
+
+  expect_equal(layer_data(ps)[c("x", "y")], df[order(df$x), ])
+})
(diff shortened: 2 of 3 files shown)


hadley:
I think it would be better to give `GeomSmooth` it's own `setup_data` method. That way you can test it using `layer_data()`

izahn:
That makes sense, but given that the line part of `geom_smooth` should work like `geom_line` I'd like to avoid repeating the same `setup_data` code in both places. Do you have any thoughts about that? Perhaps

```
GeomSmooth <- ggproto("GeomSmooth", Geom,
  setup_data = GeomLine$setup_data,
  draw_group = function(data, panel_params, coord) {
```

?

hadley:
I think it's slightly safer to manually forward it rather than literally copying the function (i.e. create a setup data method that calls the geom line setup data method). That avoids build time dependencies.

izahn:
Thanks for your help @hadley, I've updated the `setup_data` method as you suggested and added a test.

tests/testthat/test-geom-smooth.R
```diff
@@ -0,0 +1,13 @@
...
+test_that("Data is setup as in geom_line", {
+  df <- data.frame(x = c(1, 5, 2, 3, 4), y = 1:5)
+
+  ps <- ggplot(df, aes(x, y))+
+    geom_smooth(stat = "identity", se = FALSE)
+
+  pl <- ggplot(df, aes(x, y))+
+    geom_line()
+
+  expect_equal(layer_data(ps)[c("x", "y")], layer_data(pl)[c("x", "y")])
```
hadley: Rather than comparing to the output of another geom, the test would be simpler if you just compared to the correct output: 1:5.

izahn:
Sorry for taking so long to get back to this -- I've simplified the test as you suggested, though not all the way down to `1:5`.

izahn:
I see some checks didn't pass, but I'm not familiar enough with Travis to figure out what went wrong.

R/geom-smooth.r
```diff
@@ -114,6 +114,9 @@ geom_smooth <- function(mapping = NULL, data = NULL,
 #' @usage NULL
 #' @export
 GeomSmooth <- ggproto("GeomSmooth", Geom,
+    setup_data = function(data, params) {
```
hadley: Can you please fix the indenting here?

izahn: Fixed, thanks.

izahn:
I added bullet in NEWS.md. 

Thanks for your patience, believe it or not this is my first contribution to `ggplot2`.

hadley:
First of many hopefully!

Issue #2026, opened 2017-02-02
outliers supression in box_plot


botaspablo:
Hi!
I has been discussed in [a SO question](http://stackoverflow.com/questions/25124895/no-outliers-in-ggplot-boxplot-with-facet-wrap) that it is currently not possible to *supress* outliers in box plots . *Hiding* them is possible assigning no size, but they are just plotted with such empty symbol and so they still affect the plotting range. This is particularly painful to solve if using facets with free scales because the user should compute the stat_boxplot before hand and then plot the results manually. I think a better solution is desired.

```R
library(ggplot2)

# Import mtcars data
mtcars <- dput(mtcars)
mtcars$cyl <- as.factor(mtcars$cyl)

# Add random noise with manually created outliers
mtcars$noise <- c(92.59056, 106.38548, 183.49916, 90.43072, 127.17485, 102.60349,
                  115.57644, 1001.37786, 89.25277, 92.86013, 84.34921, 93.26888,
                  112.08607, 101.87805, 106.02954, 188.52964, 84.37176, 117.71647,
                  95.48017, 94.82705, 102.50222, 3.33536, 94.75700, 95.27059, 107.33767,
                  113.46822, 95.17370, 108.65510, 111.39168, 8.16712, 92.43324, 114.73957)

# Plot
ggplot(data = mtcars, aes(x = cyl, y = noise, fill = cyl)) +
       geom_boxplot() +
       facet_wrap( ~ gear, scales = 'free') +
       labs(x = "cyl", y = "noise")
```
Resulting in...

![issue](https://cloud.githubusercontent.com/assets/16158157/22551981/03963594-e925-11e6-8c51-9ea2a0060251.png)

Would it be possible to add an option to **supress outliers completely** from the plot?

hadley:
I think that's out of scope for ggplot2. If you want to create something like the boxplot that doesn't show outliers, it could go in a separate package.

botaspablo:
Ok, I see.
In my opinion a plotting package should provide flexibility above all. I understand ggplot2 is a bit more as it does statistics, however, there is the long lasting discussion of what is/is-not meaninful and that ultimately depends on the user. To my understanding boxplots generally provide the so-called *five number summary*. This summary is supposed to describe the shape of a distribution giving the values of 5 statistically significant points. In some cases outliers are not very significant and yet they prevent the user from seeing anything in the actually meaningful values. In these cases ggplot2 fails to show the data in a significant manner while not providing any further option.

[...]

PR #2030, opened 2017-02-02
Supress outliers: add and document parameter

botaspablo:
A new parameter was added to stat-boxplot.r to supress storing the outliers in the dataframe passed to geom-boxplot.r (or any other function that may call it). The outliers are taken into acocunt at every step of the stat array calculation, but not returned.

This feature was requested in #2026. It was motivated by some cases in which the outliers were not significant for the a given study and remove valuable information from a particular plot. A nice and compact solution is desired to deal with this unusual situation. The two following plots show the issue and the solution as implemented.

Issue:
issue

Solution:
solved

Added parameter ```report.outliers```:
```R
ggplot(data = dt, aes(x = strategy, y = y_, fill = strategy)) +
    geom_boxplot(report.outliers = FALSE) +
    facet_wrap( ~ struct, nrow = 2, scales = 'free')
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -217,5 +217,5 @@ Collate:
     'zxx.r'
     'zzz.r'
 VignetteBuilder: knitr
-RoxygenNote: 6.0.0
+RoxygenNote: 6.0.0.9000
 Remotes: hadley/svglite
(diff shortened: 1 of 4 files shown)


hadley:
Per our discussion in the issue, I think removing outliers from the boxplot makes it no longer a boxplot.

PR #2034, opened 2017-02-07
Add plot layout summary API

wch:
Examples:

```R
ggplot_build(p)$layout$get_layout_summary()
# # A tibble: 1 × 10
# panel   row   col       vars  xmin  xmax  ymin  ymax                        xscale
# <fctr> <dbl> <dbl>     <list> <dbl> <dbl> <dbl> <dbl>                        <list>
#   1      1     1     1 <list [0]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   # ... with 1 more variables: yscale <list>


pw <- p + facet_wrap(~ drv)
ggplot_build(pw)$layout$get_layout_summary()
# # A tibble: 3 × 10
# panel   row   col       vars  xmin  xmax  ymin  ymax                        xscale
# <fctr> <int> <int>     <list> <dbl> <dbl> <dbl> <dbl>                        <list>
#   1      1     1     1 <list [1]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   2      2     1     2 <list [1]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   3      3     1     3 <list [1]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   # ... with 1 more variables: yscale <list>
  

pg <- p + facet_grid(drv ~ cyl)
ggplot_build(pg)$layout$get_layout_summary()
# # A tibble: 12 × 10
# panel   row   col       vars  xmin  xmax  ymin  ymax                        xscale
# <fctr> <int> <int>     <list> <dbl> <dbl> <dbl> <dbl>                        <list>
#   1       1     1     1 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   2       2     1     2 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   3       3     1     3 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   4       4     1     4 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   5       5     2     1 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   6       6     2     2 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   7       7     2     3 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   8       8     2     4 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   9       9     3     1 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   10     10     3     2 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   11     11     3     3 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   12     12     3     4 <list [2]>  1.33  7.27  10.4  45.6 <S3: ScaleContinuousPosition>
#   # ... with 1 more variables: yscale <list>
```


R/facet-null.r
```diff
@@ -30,8 +30,10 @@ FacetNull <- ggproto("FacetNull", Facet,
     # Need the is.waive check for special case where no data, but aesthetics
     # are mapped to vectors
     if (is.waive(data) || empty(data))
-      return(cbind(data, PANEL = integer(0)))
-    data$PANEL <- 1L
+      return(cbind(data, PANEL = factor()))
```
hadley: Looking at it afresh, this code doesn't seem right. Why are we `cbind`ing to something that's absent?

R/layout.R
```diff
@@ -250,7 +250,75 @@ Layout <- ggproto("Layout", NULL,
...
+  },
+
+  find_layer_mappings = function(self, plot) {
+    # Default mappings. Make sure it's a regular list instead of an uneval
+    # object.
+    default <- unclass(plot$mapping)
+
+    self$layer_mappings <- lapply(plot$layers, function(layer) {
+      mapping <- default
+      default[names(layer$mapping)] <- layer$mapping
```
hadley: Can you use `modifyList()` here?

hadley: Also check that there isn't an existing function that does this  (by grepping for `modifyList()`)

R/layout.R
```diff
@@ -250,7 +250,75 @@ Layout <- ggproto("Layout", NULL,
...
+      mapping
+    })
+
+    self$layer_mappings
+  },
+
+  get_layout_summary = function(self) {
+    layout <- self$layout
```
hadley: Rather than aggressively modifying this data frame, might be simpler to start from scratch with a new one?

R/layout.R
```diff
@@ -250,7 +250,75 @@ Layout <- ggproto("Layout", NULL,
...
+    layout$SCALE_Y <- NULL
+
+    tibble::as_tibble(layout)
+  },
+
+  get_coord_summary = function(self) {
+    # Given a transform object, find the log base; if the transform object is
+    # NULL, or if it's not a log transform, return NA.
+    trans_get_log_base <- function(trans) {
+      if (!is.null(trans) && grepl("^log-", trans$name))
```
hadley: Can you add `{}` please? Or do an early return.

R/layout.R
```diff
@@ -250,7 +250,75 @@ Layout <- ggproto("Layout", NULL,
...
+        as.numeric(sub("^log-", "", trans$name))
```
hadley: It would be slightly cleaner to do `environment(trans$transform)$base`

R/plot-build.r
```diff
@@ -90,7 +93,10 @@ ggplot_build <- function(plot) {
   # Let Layout modify data before rendering
   data <- layout$finish_data(data)
 
-  list(data = data, layout = layout, plot = plot)
+  structure(
+    list(data = data, layout = layout, plot = plot),
+    class = "ggplot_built"
```
hadley: In `ggplot_gtable()`, do you mind adding a check for this class?

R/plot.r
```diff
@@ -175,7 +175,7 @@ print.ggplot <- function(x, newpage = is.null(vp), vp = NULL, ...) {
     upViewport()
   }
 
-  invisible(data)
+  invisible(x)
```
hadley: This needs a NEWS bullet

tests/testthat/test-layout-api.R
```diff
@@ -0,0 +1,110 @@
...
+  # Basic plot, and with facet_wrap and facet_grid
+  p <- ggplot(mpg, aes(displ, hwy)) + geom_point()
+  pw <- p + facet_wrap(~ drv)
+  pg <- p + facet_grid(drv ~ cyl)
+
+  l <- ggplot_build(p)$layout$get_layout_summary()
+  lw <- ggplot_build(pw)$layout$get_layout_summary()
+  lg <- ggplot_build(pg)$layout$get_layout_summary()
+
+  # Basic tests
```
hadley: Should be one test for each block

R/facet-null.r
wch: I'm not 100% sure.. Maybe it makes sense for zero-row data?

hadley: Maybe `is.waive()` should get its own branch?

R/layout.R
wch: I can use `modifyList`, but since `layer$mapping` can be NULL, it also needs a `is.null` check.

Grepping for `modifyList` doesn't turn up anything that already does this.

R/layout.R
wch: Sure.

R/layout.R
hadley: Oh I used to use `plyr::defaults()`

R/facet-null.r
wch: If I change it to this:

```R
    if (is.waive(data))
      return(data)
    if (empty(data))
      return(cbind(data, PANEL = factor()))
```

It errors on this plot:

```R
ggplot() + geom_point(aes(1:5, 1:5))
# Error in if (n == 0) { : argument is of length zero 
```

It appears that when it's a waiver object, still requires the `PANEL` to be cbinded to it.


```
> cbind(waiver(), PANEL=factor())
      PANEL
> str(cbind(waiver(), PANEL=factor()))
 list()
 - attr(*, "dim")= int [1:2] 0 2
 - attr(*, "dimnames")=List of 2
  ..$ : NULL
  ..$ : chr [1:2] "" "PANEL"
```

Should I just leave as is?

R/plot-build.r
wch: Done.

R/layout.R
```diff
@@ -258,33 +258,30 @@ Layout <- ggproto("Layout", NULL,
     default <- unclass(plot$mapping)
 
     self$layer_mappings <- lapply(plot$layers, function(layer) {
-      mapping <- default
-      default[names(layer$mapping)] <- layer$mapping
-      mapping
+      defaults(layer$mapping, default)
```
hadley: I think you can now eliminate the anonymous function too?

wch: I don't think so, because the `defaults` function gets passed `layer$mapping`, not `layer`.

R/facet-null.r
hadley: Sorry, missed this.   I think you should just return `tibble(PANEL = factor())` for the waiver case

R/layout.R
hadley: Oh yeah

PR #2055, opened 2017-02-25
fix discrete_scale documentation

### LINKED ISSUE #2052: Docs for discrete_scale inaccurate
In the documentation for `?discrete_scale`, the `labels` parameter is documented as
	
> `NULL` for no labels, `waiver()` for default labels (labels the same as breaks), a character vector the same length as breaks, or a named character vector whose names are used to match replacement the labels for matching breaks.

However, in `?scale_x_discrete`, the final example shows a very important alternative: passing a function as a formatter:

```r
# Use abbreviate as a formatter to reduce long names
ggplot(mpg, aes(reorder(manufacturer, displ), cty)) +
  geom_point() +
  scale_x_discrete(labels = abbreviate)
```

and thus should probably be documented more like `continuous_scale`'s `labels` parameter:

> One of:
> - NULL for no labels
> - waiver() for the default labels computed by the transformation object
> - A character vector giving labels (must be same length as breaks)
> - A function that takes the breaks as input and returns labels as output

I'm happy to assemble a quick PR if you like, but will everything get redone in the course of dealing with #1750 anyway?

alistaire47:
Updates `discrete_scale` documentation to match functionality and `continuous_scale` to make it clear the `labels` parameter can accept a formatter function as shown in the last example of `scale_x_discrete` (fixes #2052).

No code changes; just literally coped and pasted the documentation of the `labels` parameter from `continuous_scale`, called `devtools::document()`, and updated `NEWS.md` per `CONTRIBUTING.md` (delete that if it's too minor for news!).

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -631,10 +631,14 @@ continuous_scale <- function(aesthetics, scale_name, palette, name = waiver(),
 #' @param limits A character vector specifying the data range for the scale.
 #   The limits control what levels are displayed in the plot, their order,
 #'  and the default order of their display in guides.
-#' @param labels \code{NULL} for no labels, \code{waiver()} for default
-#'   labels (labels the same as breaks), a character vector the same length
-#'   as breaks, or a named character vector whose names are used to match
-#'   replacement the labels for matching breaks.
+#' @param labels One of: \itemize{
+#'   \item \code{NULL} for no labels
+#'   \item \code{waiver()} for the default labels computed by the
+#'     transformation object
+#'   \item A character vector giving labels (must be same length as \code{breaks})
+#'   \item A function that takes the breaks as input and returns labels
+#'     as output
+#' }
 #' @param expand a numeric vector of length two, giving a multiplicative and
 #'   additive constant used to expand the range of the scales so that there
 #'   is a small gap between the data and the axes. The defaults are (0,0.6)
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -504,7 +504,11 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
 #' Continuous scale constructor.
 #'
 #' @export
-#' @inheritParams discrete_scale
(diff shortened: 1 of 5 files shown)


R/scale-.r
```diff
@@ -631,10 +631,14 @@ continuous_scale <- function(aesthetics, scale_name, palette, name = waiver(),
 #' @param limits A character vector specifying the data range for the scale.
 #   The limits control what levels are displayed in the plot, their order,
 #'  and the default order of their display in guides.
-#' @param labels \code{NULL} for no labels, \code{waiver()} for default
-#'   labels (labels the same as breaks), a character vector the same length
-#'   as breaks, or a named character vector whose names are used to match
-#'   replacement the labels for matching breaks.
+#' @param labels One of: \itemize{
```
hadley: Can you please use `@inheritParams` instead?

alistaire47: I was putting this together but realized `continuous_scale` currently actually inherits parameters from `discrete_scale`, even though it has all its parameters defined anyway. They each have a couple unique parameters, so do you have a preference for which inherits from which?

hadley: Not really. My general convention has been when there's a class of similar function, inherit documentation from the first in the alphabet, but it's not a big deal.

alistaire47:
The newest commit merges shared parameters into `continuous_scale`, using whichever was more comprehensive, and now `discrete_scale` inherits from that. I left distinct parameters and the separate explanation of `na.value` that relates to `na.translate`.

I also realized the explanations of the `scale_name` and `name` parameters is repetitive and unclear, but I don't know enough about the difference and how they behave to alter anything.

hadley:
Thanks!

Do you want to have a stab at clearing up `name` vs `scale_name`? `scale_name` is the name of the scale function like `gradient2`, `brewer` etc. `name` is the name of the scale applied to a specific plot. The user never sees `scale_name` so it's less important to document well (but it would be nice if it was less confusing)

PR #2056, opened 2017-02-25
Add informative error message if user forgets data arg in ggplot()

jrnold:
The error message `"gglot2 doesn't know how to deal with data of class uneval"` is one of the more common ones that I see, and one that perplexes new users of ggplot2. I added a note about the most common reason for and fix of this error message.

If a `data` argument was omitted from `ggplot()`, the old behavior was

```
> ggplot(aes(x = x))
Error: ggplot2 doesn't know how to deal with data of class uneval
```
The new behavior is
```
> ggplot(aes(x = x))
Error: ggplot2 doesn't know how to deal with data of class uneval. Did you forget the data argument  in ggplot()?
```

### DIFF
--- a/R/fortify.r
+++ b/R/fortify.r
@@ -23,8 +23,8 @@ fortify.default <- function(model, data, ...) {
     "ggplot2 doesn't know how to deal with data of class ",
     paste(class(model), collapse = "/"), "."
   )
-  if (class(model) == "uneval") {
-    msg <- paste0(msg, " Did you forget the data argument in ggplot()?")
+  if (inherits(model, "uneval")) {
+    msg <- paste0(msg, " Did you forget the `data` argument in ggplot()?")
   }
   stop(msg, call. = FALSE)
 }
--- a/R/fortify.r
+++ b/R/fortify.r
@@ -24,7 +24,7 @@ fortify.default <- function(model, data, ...) {
     paste(class(model), collapse = "/"), "."
   )
   if (inherits(model, "uneval")) {
-    msg <- paste0(msg, " Did you forget the `data` argument in ggplot()?")
+    msg <- paste0(msg, " Did you accidentally provide the results of `aes()` to the `data` argument?")
   }
   stop(msg, call. = FALSE)
 }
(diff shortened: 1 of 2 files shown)


R/fortify.r
```diff
@@ -19,9 +19,12 @@ fortify.NULL <- function(model, data, ...) waiver()
...
 #' @export
 fortify.default <- function(model, data, ...) {
-  stop(
+  msg <- paste0(
     "ggplot2 doesn't know how to deal with data of class ",
-    paste(class(model), collapse = "/"),
-    call. = FALSE
+    paste(class(model), collapse = "/"), "."
   )
+  if (class(model) == "uneval") {
```
hadley: This should use `inherits()`

R/fortify.r
```diff
@@ -19,9 +19,12 @@ fortify.NULL <- function(model, data, ...) waiver()
...
+    msg <- paste0(msg, " Did you forget the data argument in ggplot()?")
```
hadley: Can you please surround `data` in backticks?

Is this function also called by geoms? If so, it might be confusing.

jrnold: I think that the error message will get called only if `ggplot()` has an `aes()` and no `data` argument:
```r
> ggplot() + geom_point(aes(x = x, y = y))
Error in FUN(X[[i]], ...) : object 'x' not found 
```
But 
```
> ggplot(aes(x = x, y = y)) + geom_point()
 Error: ggplot2 doesn't know how to deal with data of class uneval. Did you forget the `data` argument in ggplot()?
```

hadley: What about `geom_point(data = aes())` ?

jrnold: Good point. It does appear in that case. how about? 
```
Did you accidentally provide the results of `aes()` to the `data` argument?"
```

hadley:
Looks like something has gone wrong with your merge/rebase

jrnold:
I rewrote history so nothing bad ever happend

hadley:
Thanks!

jrnold:
👍

PR #2063, opened 2017-02-28
Keep element subclasses when adding

### LINKED ISSUE #1981: Change way themes are added
I've created a `textboxGrob` that allows for automatic wrapping of text etc. I'm trying to create an `element_textbox` as a subclass of `element_text` to allow people to use it for titles etc, but there are some problems in how themes are added that makes it impossible to subclass elements.

Specifically the class of the added elements are dropped if the element is already present, since `add_theme` is merely modifying the slots of the original element rather than using the added element as the base.

Will you be willing to consider a PR that flipped this behaviour so that `add_theme` would use the added element as base and modify the `NULL` slots of this?

thomasp85:
Fixes #1981 

This change simply ensures that the child element is used as the base when merging parameters with the parent, thus ensuring that any potential subclass of the child element is kept.

This makes it possible to subclass theme elements, though it is necessary to keep the same list elements as the parent class. Additional information can be stored as attributes.


hadley:
This seems to break a bunch of tests?

thomasp85:
Hmm - all visual tests breaks on my own machine for some reason. It hid one actual failing test, which is fixed now...

I know this is off topic, but shouldn't all the vdiffr stuff be working now?

hadley:
Yes, vdiffr should be working. You might need the dev version of some hadley/svglite though. 

I wonder if it might be worth making a `merge_element()` generic so that theme elements could override the default behaviour if they wanted to.

R/theme.r
```diff
@@ -410,19 +410,19 @@ add_theme <- function(t1, t2, t2name) {
       x <- y
     } else {
       # If x is not NULL, then copy over the non-NULL properties from y
```
hadley: i.e. in this block, just call `merge_element(y, x)`, and then add a generic:

```R
merge_element <- function(new, old) UseMethod("merge_element")

merge_element.theme <- function(new, old) {
   # Get logical vector of non-NULL properties in y
  ... 
}
```

thomasp85:
I like that idea — the reason I kept it this simple was that I think the theming is ripe for a rewrite judging by the issues that keps cropping up. It will prob be some time until that happens and a `merge_element` generic is not too much work

thomasp85:
So I guess the dispatch is on the child class. Is there a clever way of handling potentially multiple possible parent classes?

hadley:
No, it would be the job of the new class to handle the options for old classes.

R/theme.r
```diff
@@ -538,6 +532,61 @@ calc_element <- function(element, theme, verbose = FALSE) {
   Reduce(combine_elements, parents, theme[[element]])
 }
 
+#' Merge a parent element into a child element
+#'
+#' This is a generic and element classes must provide an implementation of this
+#' method
+#'
+#' @param new The child element in the theme hierarchy
+#'
```
hadley: Can you eliminate these blank lines?

R/theme.r
```diff
@@ -538,6 +532,61 @@ calc_element <- function(element, theme, verbose = FALSE) {
...
+#' # Adopt size but ignore colour
+#' merge_element(new, old)
+#'
+merge_element <- function(new, old) {
+  UseMethod("merge_element")
+}
+#' @rdname merge_element
+#' @export
+merge_element.default <- function(new, old) {
+  stop("No method for merging ", class(new), " into ", class(old), call. = FALSE)
```
hadley: `class()` is a character vector, so you should either extract the first element or paste them all together.

R/theme.r
```diff
@@ -538,6 +532,61 @@ calc_element <- function(element, theme, verbose = FALSE) {
...
+}
+merge_element_default <- function(new, old) {
```
hadley: I think this should just me `merge_element.element()` and then you can drop the aliases below. When you subclass `element`, you should get default merging behaviour unless you manually specify otherwise (by providing your own `merge_element` method)

thomasp85: My reason for this setup is to avoid ppl overriding the behaviour of the base elements if just `merge_element.element` is provided there is nothing stopping users from providing `neege_element.text` etc

thomasp85:
How do you feel about the whole `merge_element` setup? Should we just allow people to break stuff if they override the method for the base elements?

R/theme.r
```diff
@@ -538,6 +532,56 @@ calc_element <- function(element, theme, verbose = FALSE) {
...
+#'
+merge_element <- function(new, old) {
+  UseMethod("merge_element")
+}
+#' @rdname merge_element
+#' @export
+merge_element.default <- function(new, old) {
+  stop("No method for merging ", class(new)[1], " into ", class(old)[1], call. = FALSE)
+}
+merge_element_default <- function(new, old) {
```
hadley: I really think this should just be `merge_element.element`. No need to copy the method for each subclass.

thomasp85:
Been a bit quiet on this but ready to get it done now

hadley:
BTW in the future, request a review, and I'll usually get to it in <24 hours

thomasp85:
Sure. Still haven't quite figured out the new GitHub features...

tests/testthat/test-theme.r
```diff
@@ -202,6 +202,19 @@ test_that("All elements in complete themes have inherit.blank=TRUE", {
   expect_true(inherit_blanks(theme_void()))
 })
 
+test_that("Elements can be merged", {
+  text_base <- element_text(colour = "red", size = 10)
+  expect_equal(merge_element(element_text(colour = "blue"), text_base),
```
hadley: Can you please use tidyverse style indenting, e.g. http://style.tidyverse.org/syntax.html#long-lines

thomasp85: Are you referring to the long line 215 or the indentation?

hadley: It should look like 

```R
expect_equal(
  merge_element(element_text(colour = "blue"), text_base),
  element_text(colour = "blue", size = 10)
)
```

thomasp85: Ah, I see...

hadley:
@karawoo can you please merge this by hand?

PR #2070, opened 2017-03-10
changed https to http for ggplot2-exts.org

Emaasit:
HTTPS is unavailable for GitHub pages using custom domains


hadley:
Thanks, but this is generated from an `Rmd`, so I fixed the link there.

PR #2080, opened 2017-03-20
Fix calculation of theme defaults

### LINKED ISSUE #2058: theme_set() breaks theme_void() if axis.text.x or axis.text.y is modified
Hi, I'm having trouble with one of my custom themes since a recent update (I have tried with version 2.2.1 and the current snapshot from this repository). When I change the settings of `axis.text.x` or 
`axis.text.y` in a theme that is set as default using `theme_set()`, `theme_void()` throws an error. This causes trouble when I try to use `cowplot::plot_grid()` (see my [previous bug report](https://github.com/wilkelab/cowplot/issues/60)).

```r
library("ggplot2")

df <- data.frame(
    gp = factor(rep(letters[1:3], each = 10))
    , y = rnorm(30)
)
theme_set(theme_classic() + theme(axis.text.x = element_text()))

test <- ggplot(df, aes(gp, y)) + geom_point()

test # works
test + theme_void() # doesn't work
```

```
Error in element_grob.element_text(el, ...) : 
  Text element requires non-NULL value for 'angle'.
```

If I set `angle` and `debug` in `element_text()` it works.

```r
test + theme_void() + theme(axis.text.x = element_text(angle = 0, debug = FALSE)) # works
```

Is this intended behavior?

### LINKED ISSUE #2079: theme_void() does not remove grids
When using a custom grid, `theme_void` does not remove it. Here is a simple example illustrating it:

```r
library(ggplot2)
qplot(1:10, (1:10)^2)                   # Basic plot
theme_update(panel.grid.major = element_line(colour = "red",
    size = 0.5))                        # I want major grid in red
qplot(1:10, (1:10)^2)                   # Works OK
qplot(1:10, (1:10)^2) + theme_void()    # Grid still here
```

Given that `theme_void` is supposed to be a completely empty theme, I think `theme_void` should redefine all elements to `element_blank()`, including `panel.grid.major` and `panel.grid.minor`, and probably others (I believe it is somehow related to #2058, in the sense that `theme_void` should redefine more elements than it currently does).

has2k1:
**Issue**
At plot time, the calculation of theme elements always fell back
to the default theme. As a consequence, the default theme would
leak into complete themes where some of the elements were
unspecified.

**Solution**
Be more specific when calculating the theme defaults.

**Clean Up**
`theme_void` was insufficiently specified. The solution exposed
missing elements that had to be specified.

Fixes #2058
Fixes #2079

### DIFF
--- a/R/theme.r
+++ b/R/theme.r
@@ -416,7 +416,14 @@ theme <- function(line,
 
 # Combine plot defaults with current theme to get complete theme for a plot
 plot_theme <- function(x) {
-  defaults(x$theme, theme_get())
+  complete <- attr(x$theme, "complete")
+  if (is.null(complete)) {
+    theme_get()
+  } else if (complete) {
+    x$theme
+  } else {
+    defaults(x$theme, theme_get())
+  }
 }
 
 #' Modify properties of an element in a theme object
--- a/R/theme.r
+++ b/R/theme.r
@@ -415,14 +415,14 @@ theme <- function(line,
 
 
 # Combine plot defaults with current theme to get complete theme for a plot
-plot_theme <- function(x) {
+plot_theme <- function(x, default_theme=theme_get()) {
   complete <- attr(x$theme, "complete")
   if (is.null(complete)) {
-    theme_get()
+    default_theme
   } else if (complete) {
     x$theme
   } else {
-    defaults(x$theme, theme_get())
+    defaults(x$theme, default_theme)
   }
 }
 
--- a/tests/testthat/test-theme.r
+++ b/tests/testthat/test-theme.r
@@ -164,19 +164,6 @@ test_that("Complete and non-complete themes interact correctly with ggplot objec
   expect_equal(p$plot$theme$text$colour, "red")
   expect_equal(p$plot$theme$text$face, "italic")
 
-  # The final calculated plot theme should not blend a complete theme
-  # with the default theme
-  default_theme <- theme_gray() + theme(axis.text.x = element_text(colour = "red"))
-
-  ptheme <- plot_theme(qplot(1:3, 1:3) + theme_void(), default_theme)
-  expect_null(ptheme$axis.text.x)
-
-  ptheme <- plot_theme(qplot(1:3, 1:3) + theme_gray(), default_theme)
-  expect_true(is.null(ptheme$axis.text.x$colour) || ptheme$axis.text.x$colour != "red")
-
-  ptheme <- plot_theme(qplot(1:3, 1:3) + theme(axis.text.y = element_text(colour = "blue")), default_theme)
-  expect_equal(ptheme$axis.text.x$colour, "red")
-  expect_equal(ptheme$axis.text.y$colour, "blue")
 })
 
 test_that("theme(validate=FALSE) means do not validate_element", {
@@ -238,6 +225,19 @@ test_that("Elements can be merged", {
   )
 })
 
+test_that("Final calculated plot theme should not blend a complete theme with the default theme", {
+  default_theme <- theme_gray() + theme(axis.text.x = element_text(colour = "red"))
+
+  ptheme <- plot_theme(qplot(1:3, 1:3) + theme_void(), default_theme)
+  expect_null(ptheme$axis.text.x)
+
+  ptheme <- plot_theme(qplot(1:3, 1:3) + theme_gray(), default_theme)
+  expect_true(is.null(ptheme$axis.text.x$colour) || ptheme$axis.text.x$colour != "red")
+
+  ptheme <- plot_theme(qplot(1:3, 1:3) + theme(axis.text.y = element_text(colour = "blue")), default_theme)
+  expect_equal(ptheme$axis.text.x$colour, "red")
+  expect_equal(ptheme$axis.text.y$colour, "blue")
+})
 
 # Visual tests ------------------------------------------------------------
 
--- a/tests/testthat/test-theme.r
+++ b/tests/testthat/test-theme.r
@@ -163,6 +163,20 @@ test_that("Complete and non-complete themes interact correctly with ggplot objec
   expect_false(attr(p$plot$theme, "complete"))
   expect_equal(p$plot$theme$text$colour, "red")
   expect_equal(p$plot$theme$text$face, "italic")
+
+  # The final calculated plot theme should not blend a complete theme
+  # with the default theme
+  default_theme <- theme_gray() + theme(axis.text.x = element_text(colour = "red"))
+
+  ptheme <- plot_theme(qplot(1:3, 1:3) + theme_void(), default_theme)
+  expect_null(ptheme$axis.text.x)
+
+  ptheme <- plot_theme(qplot(1:3, 1:3) + theme_gray(), default_theme)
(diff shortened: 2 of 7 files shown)


has2k1:
Tests fail due to assignments that relied on the problematic default theme inheritance that this PR got rid of.
The candidates should be (atleast) some of the results of 

```sh
$ ag 'theme\$.+?<-'
```

thomasp85:
I don't think changing inheritance is necessary. All of these issues seems to come from the fact that `theme_void()` was incompletely specified, which is the root bug...

has2k1:
I noticed that it was incompletely specified, but I think if you look closer the complete specification that would make the bug go away would imply that the elements themselves have no inheritance hierarchy. That is, even if you have `axis.text = element_blank()` you would also need `axis.text.x = element_blank()`.

To be clear, these are the options for what could be meant by *complete*:

1. Every possible element is specified.
2. Every possible element is specified or have a parent that is specified.

This patch recognises the second definition, whereas your (@thomasp85) suggestion seems to imply the first. Also, if it is the first then the [line](https://github.com/tidyverse/ggplot2/pull/2080/files#diff-1430e0a997db5a2b2177830bea0a7392L383) that I removed/relocated would always be redundant as there wouldn't be any unspecified elements.

crsh:
Sorry to bother, but are there any news on this pull request?

has2k1:
@crsh, I'll look into fixing the failing test case. The first time around, I expected to ask for help with a clearer question. I did not get to it.

has2k1:
Fails on [appveyor](https://ci.appveyor.com/project/tidyverse/ggplot2/build/1.0.198#L443) but not on travis.
@hadley , any reason for this? Do the visual tests only run appveyor? Otherwise, it seems like a legitimate failure.

hadley:
@has2k1 Do you see the failure if you run locally?

has2k1:
Locally, I get many failures (with/without this PR maybe something to do with fonts) but not the particular one that fails on appveyor.

hadley:
@lionel- do you have any tools for debugging this sort of failure?

lionel-:
Nothing user-oriented yet, we definitely should add some UI for this.

IIRC, I generated a SVG from the faulty plot manually with `vdiffr:::write_svg()`, then I displayed the SVG contents with `cat()`, and compared the outputs on Travis and Appveyor.

We need a verbose `expect_doppelganger()` variant or option.

R/theme.r
```diff
@@ -380,7 +380,13 @@ theme <- function(line,
 
 # Combine plot defaults with current theme to get complete theme for a plot
 plot_theme <- function(x) {
-  defaults(x$theme, theme_get())
+  if (is.null(attr(x$theme, "complete"))) {
```
hadley: When does this branch get triggered?

hadley:
@karawoo can you please run the visual tests locally?

@has2k1 could you please add a bullet to NEWS? Are there some unit tests we could add to check that we don't fall back to the old (suboptimal) behaviour?

R/theme.r
has2k1: When using the default theme. The cases for the three branches respectively are.
```R
ggplot(...) + geom_point()

ggplot(...) + geom_point() + theme_gray()

ggplot(...) + geom_point() + theme(...)
```

karawoo:
I get the same failing test that AppVeyor complains about. The title for `theme-void.svg` is being placed in the center rather than at the left, and it looks like the size of the title text and the amount of space between the plot and the title have changed too.

R/theme.r
hadley: Ah, got it.  It might be slightly easier to read if you did:

```R
complete <- attr(x$theme, "complete")
```

hadley:
@has2k1 can you reproduce those issues locally yourself?

has2k1:
I cannot reproduce the failing test.

 Also, if I approximate the failing test with

```R
df <- data.frame(x = 1:3, y = 1:3, z = c("a", "b", "a"), a = 1)
plot <- ggplot(df, aes(x, y, colour = z)) +
  geom_point() +
  facet_wrap(~ a)

plot + ggtitle("theme_void") + theme_void()
```

The title is still in the center.

For the speculation (my system is linux), given that on `travis` (linux) the test passes, and on `appveyor` (windows) it fails, could be an issue with the svg driver defaults between different systems. @karawoo, I think you can dismiss this as an explanation if the `png` and `svg` results of the code above are different.

has2k1:
For creating a test, I do not know how wise it would be to mess with the default theme (a global variable) in the test suit.

hadley:
@has2k1 it's fine. Just use `on.exit()` to restore. Alternatively, you could add a second argument to `plot_theme` that defaults to `theme_get()` but you could override in the tests (that's probably a better solution)

tests/testthat/test-theme.r
```diff
@@ -163,6 +163,20 @@ test_that("Complete and non-complete themes interact correctly with ggplot objec
   expect_false(attr(p$plot$theme, "complete"))
   expect_equal(p$plot$theme$text$colour, "red")
   expect_equal(p$plot$theme$text$face, "italic")
+
+  # The final calculated plot theme should not blend a complete theme
```
hadley: Could you please make this a separate test?

karawoo:
Yeah, that's what I was seeing with the title.

has2k1:
I've specified the title, subtiltle and caption text elements of `theme_void` so that they take on the same values as in the other themes. This should have `theme_void` maintain a similar overall arrangement. I am not sure if any more theme elements should be specified to get the tests to pass or to make the current appearance the new normal.

hadley:
@has2k1 looks like there are still changes to the tests.

has2k1:
Since the tests where generated with `theme_void` inheriting all unspecified elements from `theme_grey`, eliminating all the changes would require that `theme_void` be specified by duplicating the elements in the set `them_gray_elements - theme_void_elements`. I don't think that is a good idea.

I think there are two alternatives;

1. The cop-out, decide that `theme_void` is should not be elaborately specified, find a sufficient level of specification and declare that the new normal.

2. Elaborately specify `theme_void` then have `theme_grey` override (replace) elements in `theme_void`.

The first is what the last commit was aiming for, the 2nd though a good idea it changes the long standing status of `theme_grey` as the reference (fully specified) theme.

hadley:
If you believe the behaviour is correct, you need to check in updated visual tests.

has2k1:
I need some help, I cannot generate images that can pass visual tests on my system.

lionel-:
What systems are you using? vdiffr is currently broken on some Linux systems, I'm investigating.

has2k1:
Indeed, I am on a linux system.

lionel-:
Which one?

lionel-:
Also could you provide the result of running these two commands please:

```r
gdtools::version_fontconfig()
gdtools::version_freetype()
```

has2k1:
Gentoo Linux

```
> gdtools::version_fontconfig()
[1] ‘2.12.1’

> gdtools::version_freetype()
[1] ‘2.8.0’
```

hadley:
@has2k1 would you be willing to update this? I can try and merge in the next week or so.

hadley:
Actually, don't worry, I can do it myself

PR #2089, opened 2017-03-27
Enclose issue instructions in comment delimiters

has2k1:
Password: 

If it works you should not see my password.


hadley:
I think it's important to leave them so there's some indication whether or not people have read them (my conclusion so far is that no one actually reads them)

PR #2095, opened 2017-04-04
Sample points uniformly from transformed space

### LINKED ISSUE #1992: sec.axis with scale_*_log10
Hi,

I have been trying to plot continuous data with a log-scaled x-axis and a secondary x-axis in order to display ticks values in another unit.
It seems to me that the ticks positions of this secondary axis are not correctly computed. 

Here is a reproducible example:
```R
foo <- tibble(x = exp(seq(log(0.001), log(1000), length.out = 100)), y = x/(1+x))

ggplot(foo, aes(x, y))+
  geom_point()+
  scale_x_log10(name = "Unit A",
                breaks = c(0.001, 0.01, 0.1, 1, 10, 100, 1000),
                sec.axis = sec_axis(~. * 100, name = "Unit B", 
                                    labels = derive(),
                                    breaks = derive()))
```
sec_axis1
Considering the transformation (`x * 100`),  ticks positions do not correspond.

Also, `dup_axis()` does not duplicate the original axis:
```R
ggplot(foo, aes(x, y))+
  geom_point()+
  scale_x_log10(name = "Unit A",
                breaks = c(0.001, 0.01, 0.1, 1, 10, 100, 1000),
                sec.axis = dup_axis())
```
sec_axis2

Am I doing something wrong?

thomasp85:
Fixes #1992


tests/testthat/test-sec-axis.R
```diff
@@ -0,0 +1,45 @@
...
+    "dup_axis",
+    ggplot(foo, aes(x, y)) +
+      geom_point() +
+      scale_x_continuous(name = "Unit A",
+                         sec.axis = dup_axis())
+  )
+})
+
+test_that("custom breaks works", {
+  vdiffr::expect_doppelganger(
```
hadley: This would be better as a regular test

tests/testthat/test-sec-axis.R
```diff
@@ -0,0 +1,45 @@
...
+  x = x,
+  y = x/(1+x)
+)
+
+test_that("dup_axis() works", {
+  vdiffr::expect_doppelganger(
+    "dup_axis",
+    ggplot(foo, aes(x, y)) +
+      geom_point() +
+      scale_x_continuous(name = "Unit A",
```
hadley: Better to set name to `NULL` here, since it's not germane to the test.  

I'd also consider using `geom_blank()` to help focus on the important part of the visual test

thomasp85: Sure, that's sloppy copy-pasting from my part

tests/testthat/test-sec-axis.R
thomasp85: All of them or just this one? Don't see how this one differs from the rest...

hadley: I think it's good to have one visual test to ensure the whole process basically works correctly. And then regular tests for everything else.

thomasp85: gotcha

hadley:
@thomasp85 are you interested in finishing this off? Otherwise @karawoo can take a look.

thomasp85:
Completely forgot about this one. Sure I'll finish it off👍

hadley:
@thomasp85 still interested?

thomasp85:
Yep. Sorry, it seems this is constantly slipping out of focus... within the week

thomasp85:
Sorry for being slow on this - how is the new tests looking?

hadley:
Looks good - but I think you need to delete a couple of the generated svgs?

thomasp85:
The fail on travis devel is related to stringr and not this PR

PR #2097, opened 2017-04-05
Add stat for easily finding slope and intercept of Q-Q line

nicksolomon:
`stat_qq_line()` computes the slope and intercept of the line connecting the points at the first and third quartiles of the theoretical and sample distributions, then plots this line using `geom_abline()`.

This is useful for judging the fit of the theoretical distribution to the observed data. Adding lines manually was especially difficult in plots that included facets where each facet required a substantially different line.


has2k1:
`geom_abline` may not be the best to use. If the computed statistics are `slope` and `intercept`, it is hard to use them with other `geoms`. It unnecessarily limits "composability". I think `geom_segment` or `geom_path` would be better.

You could also have the quantiles as parameters.

nicksolomon:
Thanks for the suggestions! I also added a parameter to control how far past the plotted points the line extends.

NEWS.md
```diff
@@ -43,6 +43,8 @@
 
 * `ggproto()` produces objects with class `c("ggproto", "gg")`. This was added so that when layers, scales, or other ggproto objects are added together, an informative error message is raised (@jrnold, #2056).
 
+* Added `stat_qq_line()` to make it easy to add a simple line to a Q-Q plot. This line makes it easier to judge the fit of the theoretical distribution.
```
hadley: Can you please wrap the line? And add `(@nicksolomon)`

R/stat-qq-line.R
```diff
@@ -0,0 +1,128 @@
...
+#'   stat_qq() +
+#'   stat_qq_line()
+#' }
+geom_qq_line <- function(mapping = NULL, data = NULL,
+                         geom = "path", position = "identity",
+                         ...,
+                         distribution = stats::qnorm,
+                         dparams = list(),
+                         line.p = c(.25, .75),
+                         line.expand = c(-.1, .1),
```
hadley: Rather than `line.expand` maybe it would be better to mimic `stat_smooth()` and use `fullrange = FALSE`?

R/stat-qq-line.R
```diff
@@ -0,0 +1,128 @@
...
+   n <- length(sample)
+
+   # Compute theoretical quantiles
+   if (is.null(quantiles)) {
+     quantiles <- stats::ppoints(n)
+   } else {
+     stopifnot(length(quantiles) == n)
+   }
+
+   theoretical <- do.call(distribution,
```
hadley: Can you please use tidyverse style here: http://style.tidyverse.org/syntax.html#long-lines (and elsewhere)

hadley:
@nicksolomon are you interested in finishing this PR off?

nicksolomon:
@hadley Yes! I was planning to get to the changes you requested this weekend.

R/stat-qq-line.R
```diff
@@ -114,15 +119,13 @@ StatQqLine <- ggproto("StatQqLine", Stat,
    slope = diff(y_coords)/diff(x_coords)
    intercept = y_coords[1L] - slope * x_coords[1L]
 
-   if (length(line.expand) != 2) {
-     stop("Parameter line.expand must have length 2.", call = FALSE)
+   if(fullrange & !is.null(scales$x$dimension)){
```
hadley: Needs space before `(`.

I think you should move the duplicated data frame creation code out of the `if` statement.

NEWS.md
```diff
@@ -43,6 +43,9 @@
 
 * `ggproto()` produces objects with class `c("ggproto", "gg")`. This was added so that when layers, scales, or other ggproto objects are added together, an informative error message is raised (@jrnold, #2056).
 
+* Added `stat_qq_line()` to make it easy to add a simple line to a Q-Q plot. This
+line makes it easier to judge the fit of the theoretical distribution (@nicksolomon).
```
hadley: Second line should be indented

R/stat-qq-line.R
```diff
@@ -0,0 +1,131 @@
+#' A quantile-quantile line
```
hadley: I wonder if this should be documented with `stat_qq`

nicksolomon: I took a crack at combining them, but I'm not sure if the results are satisfactory. I don't have much experience with roxygen.

hadley: I'd remove most of the stuff here (except for the new parameters), and instead write more in the `geom_qq` docs. I think it makes sense to document them together since they're typically used together

PR #2099, opened 2017-04-07
Adding flexible methods for drawing axis lines.

stefanedwards:
The first motivation for this addition was to draw axis lines that did not meet at the corners. Personally, I just prefer that they do not.

Secondly, I wanted a manner to indicate the range of valid values for the axis. E.g. a correlation does not exceed the range [-1, 1]. In the below example, it is the count from stat_bin2d:

image

Thirdly, for categorical variables where dots could be jittered, I wanted to display a bracket to group them. See below:

image

I have achieved this by sub-classing `CoordCart`, and the related Cartesian coordinate objects, with `CoordFlexCart`, with defining method `coord_flex_cart`, which is powered by the following:

```
coord_flex_cart <- function(xlim = NULL,
                            ylim = NULL,
                            expand = TRUE,
                            top = waiver(),
                            left = waiver(),
                            bottom = waiver(),
                            right = waiver()) {
  ggproto(NULL, CoordFlexCartesian,
          limits = list(x = xlim, y = ylim),
          expand = expand,
          top = top,
          left = left,
          bottom = bottom,
          right = right
  )
}
CoordFlexCartesian <- ggproto('CoordFlexCartesian', `_inherit` = CoordCartesian,
  render_axis_h = flex_render_axis_h,
  render_axis_v = flex_render_axis_v
)
flex_render_axis_v <- function(self, scale_details, theme) {
  arrange <- scale_details$y.arrange %||% c("primary","secondary")
  left <- self$left %|W|% render_axis
  right <- self$right %|W|% render_axis
  list(
    left = left(scale_details, arrange[1], 'y', 'left', theme),
    right = right(scale_details, arrange[2], 'y', 'right', theme)
  )
}
(flex_render_axis_h is as defined in similar way)
```

The functions `flex_render_axis_h` and `flex_render_axis_v` have been defined outside of the class, as they are reused for flipped and fixed Cartesian objects.

The two examples can be called as (with `p` defined from relevant examples):
```
p + coord_flex_cart(left=capped_vertical(capped='both'))
p + coord_flex_cart(bottom=brackets_horisontal())
```

For the capped lines, short hand methods have been defined which allows a quicker specification:
```
p + coord_capped_cart(left='both')
```


stefanedwards:
Functionality implemented in [R package 'lemon](https://cran.r-project.org/web/packages/lemon/index.html)'.

PR #2113, opened 2017-04-27
Add note recommending annotate

andrewdolman:
Add note to documentation explaining differing behaviour of geom_text, and annotate(geom = "text").


hadley:
@karawoo can you please merge this and update docs?

PR #2115, opened 2017-04-27
Minor typo: monotonous -> monotonic

jonocarroll:
Unless I'm missing something about boring graphs, that is.

Refer: https://math.stackexchange.com/a/365787/90266

> Q: Is “monotonous” ever used as a synonym for “monotonic” in math?
> A: It is always wrong.

I tripped this error and had a chuckle, nonetheless.


hadley:
Thanks!

PR #2132, opened 2017-05-15
Added linejoin parameter to geom_segment.

Ax3man:
This PR simply adds support for the `linejoin` parameter of `grid::segmentsGrob` to `geom_segment`, just like the existing `lineend` parameter. This is especially useful when drawing large arrows that need to look sharp. The default of `"round"` can look poor with filled arrows as all corners are rounded off.

temp

The default is the current fixed value.

I'm not sure about the appropriate sizes etc. for the added example as it depends a bit on the size that is drawn for the online docs.


R/geom-segment.r
```diff
@@ -42,6 +43,19 @@
 #'     arrow = arrow(length = unit(0.1,"cm"))) +
 #'   borders("state")
 #'
+#' # Use lineend and linejoin to change the style of the segments
+#' df2 <- expand.grid(lineend = c('round', 'butt', 'square'),
+#'                    linejoin = c('round', 'mitre', 'bevel'), stringsAsFactors = FALSE)
+#' segments <- lapply(seq_len(nrow(df2)), function(i) {
```
hadley: This use of `lapply()` will confuse most readers. Can you please generate with a simple helper function and copy and paste?

Ax3man: I've simplified this is to a single call to geom_segment, hope that helps.

R/geom-segment.r
```diff
@@ -42,6 +43,18 @@
 #'     arrow = arrow(length = unit(0.1,"cm"))) +
 #'   borders("state")
 #'
+#' # Use lineend and linejoin to change the style of the segments
+#' df2 <- expand.grid(lineend = c('round', 'butt', 'square'),
+#'                    linejoin = c('round', 'mitre', 'bevel'), stringsAsFactors = FALSE)
+#' df2 <- data.frame(df2, y = 1:9)
+#' ggplot() +
```
hadley: Can you please use house indenting rules? (http://style.tidyverse.org/syntax.html#long-lines)

R/geom-segment.r
```diff
@@ -42,6 +43,18 @@
...
+#'   geom_segment(aes(x = 1, y = y, xend = 2, yend = y), df2, size = 4,
+#'                lineend = df2$lineend, linejoin = df2$linejoin,
+#'                arrow = arrow(length = unit(0.5, "inches"))) +
+#'   geom_text(aes(x = 2, y = 1:9, label = paste(df2$lineend, df2$linejoin)),
```
hadley: I think you can simplify this plot by setting `data` and `aes(y = y)` in the ggplot2 call.  The `geom_text()` layer shouldn't use `df$`

Ax3man:
I've simplified to a single `aes` call. I think it's useful to show all options available (as many users won't have much of an idea about what a bevel should look like), so that does make the plot a little more complicated than most. Alternatively I could just show e.g. two examples if you'd like to further simplify.

hadley:
Looks good to me - thanks!

hadley:
@karawoo can you please merge by hand?

olli0601:
Hello, any chance the linejoin option could also be added to geom_curve? Thank you, Oliver

PR #2140, opened 2017-05-23
fix typo

Thieffen:
typo


hadley:
The source of this document is `README.Rmd` so I fixed the typo there. Thanks for spotting it!

PR #2142, opened 2017-05-25
Fix density calculations for groups with one or two elements

### LINKED ISSUE #2127: geom_density with groups of 1 or 2 elements
Hello,

I am struggling with `geom_density` when groups consist on fewer than 3 elements. 
For a group of 1 element, the density plot consists on a single line of length 1. For a group of 2 elements, this is a rectangle of height 0.5.
Since this does not depends on the _y_ scale, the interpretation of the area under the curves breaks down. This problem becomes obvious when changing coordinates on the _x_ axis (_eg_ unit conversion).

This reproducible example illustrates my point.
This is the density of the measurements of `Speed`, for the 5 experiments. Each consists on 20 runs.

```r
library(tidyverse)
library(MASS)

michelson %>%
  ggplot(aes(x = Speed, fill = Expt)) +
  geom_density(alpha = 0.5)
```
image

The density of the measurements, for the 5 experiments when we re-normalize the speed. As expected, the only changes are in the axis texts.

```r
michelson %>%
  mutate(Speed = Speed / mean(Speed)) %>%
  ggplot(aes(x = Speed, fill = Expt)) +
  geom_density(alpha = 0.5)
```
image

The problem arises when the groups are small,
so we modify the data for the last group to consist on only two runs.
```r
michelson_modif <- michelson
michelson_modif$Expt[83:100] <- 2

michelson_modif %>%
  ggplot(aes(x = Speed, fill = Expt)) +
  geom_density(alpha = 0.5)
```

image

```r
michelson_modif %>%
  mutate(Speed = Speed / mean(Speed)) %>%
  ggplot(aes(x = Speed, fill = Expt)) +
  geom_density(alpha = 0.5)

```
image

The two graphs are no longer similar.

Do I miss something here? Is it possible to force the behaviour of `geom_density` to be similar (_ie_ smooth curves) for all the groups, regardless of the number of elements in them?

Thanks a lot!

karawoo:
Fixes #2127.

`compute_density()` now uses `stats::density()` to compute the density for two-element groups, resulting in plots that look like this (where group 5 has only two data points):

example_scaled

In the case where there is only a single point in the group (where it doesn't really make sense to calculate density), the group gets dropped with a warning.

```R
## Create some data with a couple very small groups
library("tidyverse")
set.seed(1958)
dat <- tibble(x = c(rnorm(n = 10, mean = 7, sd = 2),
                    rpois(n = 10, lambda = 2),
                    1, 9, 4),
              group = c(rep(c("a", "b"), each = 10), "c", "c", "d"))

## Plot
ggplot(dat, aes(x = x, fill = group)) +
  geom_density(alpha = 0.5)
```

```
Warning message:
In compute_density(data$x, data$weight, from = range[1], to = range[2],  :
  Groups with fewer than two data points have been dropped.
```

dropped_group

There is still a legend element for the dropped group; not sure if that should be kept or not.

### DIFF
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -85,14 +85,15 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
     w <- rep(1 / nx, nx)
   }
 
-  # if less than 3 points, spread density evenly over points
-  if (nx < 3) {
+  # if less than 2 points return data frame of NAs and a warning
+  if (nx < 2) {
+    warning("Groups with fewer than two data points have been dropped.")
     return(data.frame(
-      x = x,
-      density = w / sum(w),
-      scaled = w / max(w),
-      count = 1,
-      n = nx
+      x = NA,
+      density = NA,
+      scaled = NA,
+      count = NA,
+      n = NA
     ))
   }
 
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -87,7 +87,7 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
 
   # if less than 2 points return data frame of NAs and a warning
   if (nx < 2) {
-    warning("Groups with fewer than two data points have been dropped.")
+    warning("Groups with fewer than two data points have been dropped.", call. = FALSE)
     return(data.frame(
       x = NA,
       density = NA,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # ggplot2 2.2.1.9000
 
+* `geom_density` drops groups with fewer than two data points and throws a
+  warning. For groups with two data points, the density values are now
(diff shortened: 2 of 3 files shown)


R/stat-density.r
```diff
@@ -85,14 +85,15 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
     w <- rep(1 / nx, nx)
   }
 
-  # if less than 3 points, spread density evenly over points
-  if (nx < 3) {
+  # if less than 2 points return data frame of NAs and a warning
+  if (nx < 2) {
+    warning("Groups with fewer than two data points have been dropped.")
```
hadley: Can you please add `call. = FALSE` here?

karawoo: Done!

hadley:
Thanks!

PR #2145, opened 2017-05-26
Fix theme_gray's legend/panels for large base_size

raubreywhite:
Currently theme_gray's legend and background panels don't work well at base_size=24 or base_size=30 because their sizes are hardcoded in. If we make them relative, then theme_gray looks nice at larger base_size's.


raubreywhite:
Hi, any input on this? The conflicts appear to be due to the tests not allowing changes to the default themes, which is the point of this PR...

hadley:
@karawoo I've already forgotten what other changes we made. Can you please comment as to whether this is still necessary?

karawoo:
This is still necessary but I think it should wait until I add the `base_line_size` parameter (#2176; I can get a PR in for that today). Then we can use `rel()` to set these relative to the base line size, which I think makes more sense than setting it relative to the text size.

hadley:
👍

PR #2154, opened 2017-06-09
Fix position_dodge

### LINKED ISSUE #2153: Facetting leads to faulty x values for position_dodge(preserve="single")
``` r
library(ggplot2)

ggplot(mtcars, aes(factor(cyl), fill=factor(vs))) +
  geom_bar(position=position_dodge(preserve = "single")) +
  facet_wrap(~ am)
```

The width of the bars is wrong. The problem is [here](https://github.com/tidyverse/ggplot2/blob/eedaa81b72d1ae56b86c04397455c8f0c6c84bd2/R/position-dodge.r#L85)..

has2k1:
When calculating the groups in a bin, do it per panel then get the
maximum across all panel.

fixes #2153

### DIFF
--- a/R/position-dodge.r
+++ b/R/position-dodge.r
@@ -82,7 +82,7 @@ PositionDodge <- ggproto("PositionDodge", Position,
     if (identical(self$preserve, "total")) {
       n <- NULL
     } else {
-      n <- max(table(data$xmin))
+      n <- max(plyr::ddply(data, "PANEL", plyr::summarise, n=max(table(xmin)))$n)
     }
 
     list(
--- a/tests/testthat/test-position_dodge.R
+++ b/tests/testthat/test-position_dodge.R
@@ -1,13 +1,15 @@
 context("position_dodge")
 
 test_that("can control whether to preserve total or individual width", {
-  df <- data.frame(x = c("a", "b", "b"), y = c("a", "a", "b"))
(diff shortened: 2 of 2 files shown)


hadley:
Would you mind including a before and after plot in a comment here?

has2k1:
``` r
library(ggplot2)

ggplot(mtcars, aes(factor(cyl), fill=factor(vs))) +
  geom_bar(position=position_dodge(preserve = "single")) +
  facet_wrap(~ am)
```

Before


After


Note the width of the bars.

hadley:
Is the width wrong though? It seems like you could argue either way, especially since the bars aren't correctly positioned in "after" (I'm aware that's another issue)

has2k1:
The bars should be centred at the x value of the group. In the before you have two groups but the width of the bars is `0.9 * 1/3` instead of `0.9 * 1/2`, there is an implied but missing third group.

Without facetting there is no problem.

``` r
library(ggplot2)

ggplot(mtcars, aes(factor(cyl), fill=factor(vs))) +
  geom_bar(position=position_dodge(preserve = "single"))
```

has2k1:
IIRC, I may have looked into that other issue and found that there is not enough information at the point calculation (dodging) to fix it.

hadley:
I'd rather elucidate the desired principles before we write any code. What _should_ happen if there are no overlaps in a given facet? Should bars match the size (but not position?) of the facets where dodging occurs? Or should those facets appear identical to the non-dodge equivalent?

Do we need a third value for `preserve`, maybe `preserve = "facet"` ?

PR #2161, opened 2017-06-15
Update documentation of discrete_scale params that differ from continuous_scale

### LINKED ISSUE #2068: discrete_scale documentation
This looks related to the change in https://github.com/tidyverse/ggplot2/issues/2052, the break functions should be different for `discrete_scale` and `continuous_scale` in the documentation.  Continuous takes the limits and returns the breaks, but discrete takes the full character vector for default breaks (since there's no limits).

The input to both functions for labels looks the same, and takes the breaks as an input .

karawoo:
Fixes #2068. `discrete_scale` inherits params from `continuous_scale`, but the `breaks` and `limits` arguments behave differently for discrete vs. continuous scales. This adds new `breaks` and `limits` parameter documentation for `discrete_scale` that overrides the inherited parameter documentation.

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -611,6 +611,16 @@ continuous_scale <- function(aesthetics, scale_name, palette, name = waiver(),
 #'
 #' @export
 #' @inheritParams continuous_scale
+#' @param breaks One of: \itemize{
+#'   \item \code{NULL} for no breaks
+#'   \item \code{waiver()} for the default breaks computed by the
+#'     transformation object
+#'   \item A character vector of breaks
+#'   \item A function that takes the limits as input and returns breaks
+#'     as output
+#' }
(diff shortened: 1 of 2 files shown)


R/scale-.r
```diff
@@ -611,6 +611,16 @@ continuous_scale <- function(aesthetics, scale_name, palette, name = waiver(),
...
 #' @inheritParams continuous_scale
+#' @param breaks One of: \itemize{
+#'   \item \code{NULL} for no breaks
+#'   \item \code{waiver()} for the default breaks computed by the
+#'     transformation object
+#'   \item A character vector of breaks
+#'   \item A function that takes the limits as input and returns breaks
+#'     as output
+#' }
+#' @param limits A character vector of all possible values of the scale and
```
hadley: "of all" -> "that defines" ?

PR #2166, opened 2017-06-16
Use markdown for documentation

### LINKED ISSUE #2165: Convert documentation to markdown
Use `roxygen2md::roxygen2md()` to convert Rd syntax to markdown in all of the documentation. I'm working on this but it's not ready for a PR just yet because output from the `aesthetics` macro isn't rendered properly. It should be:
screen shot 2017-06-15 at 4 01 35 pm

But I'm getting:

screen shot 2017-06-15 at 4 02 36 pm

karawoo:
Closes #2165. Documentation now uses markdown rather than Rd syntax.


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # ggplot2 2.2.1.9000
 
+* Documentation now uses markdown thanks to roxygen2md::roxygen2md() (@karawoo,
```
hadley: Generally, I don't bother listing changes that don't have some user facing impact

karawoo:
It looks like the build fails because roxygen2 tries to convert some bracketed text to links that shouldn't be. For example

```
#' @param hjust Horizontal justification (in [0, 1])
```

becomes

```
\item{hjust}{Horizontal justification (in \link{0, 1})}
```

karawoo:
Should I just use `@noMd` in the sections where this is a problem (it's only a few) and convert other markdown in those sections back to Rd syntax?

hadley:
@gaborcsardi is there an easy way to work around this buglet?

gaborcsardi:
How about putting it in `\eqn`? I guess we need a way to temporarily turn off markdown parsing....

gaborcsardi:
I don't think we can "fix" this easily, because we would need to start guessing what is a link and what is not a link. OK, this case is kind of obvious, it as a space, but there can be less obvious cases.

karawoo:
`\eqn` works though the output appears italicized like _[0, 1]_. `\verb` also works and puts the text in monospaced font. @hadley do you have a preference?

gaborcsardi:
`\eqn` is for inline equations, so it is imo appropriate here. 

Regardless, we need a way to escape from markdown...

hadley:
`\eqn` is probably the most technically correct, so let's go with that.

Issue #2164, opened 2017-06-15
Use @inheritDotParams 


karawoo:
It would be useful to use `@inheritDotParams` for documenting the `...` parameter in a lot of places. See the [roxygen2 NEWS](https://github.com/klutometis/roxygen/blob/master/NEWS.md#improved-inheritance) for more info on `@inheritDotParams`.

PR #2170, opened 2017-06-17
Start using @inheritDotParams in documentation

karawoo:
See #2164

This uses `@inheritDotParams` for some `...` parameters. There are still some other places where we might be able to switch to `@inheritDotParams`. 

There's an issue when the dot parameters come from another package and include links. For example the `map_data` function passes `...` to `maps::map()`, but some of those params have links to other stuff within the maps package, and with `@inheritDotParams` the links get written to Rd as e.g. `\link{county}` when they need to be `\link[maps:county]{county}`


karawoo:
Per discussion this morning we won't use `@inheritDotParams` for functions from other packages for now. I'll go through again and see if there are more places we can use it, though.

PR #1883, opened 2016-10-31
use rel() more consistently for linewidths

baptiste:
Small plots (e.g. in the margins of tufte-style books) tend to look heavy with the current linewidth settings, and after investigating I noticed some inconsistencies in the use of `rel()`.

PR #2173, opened 2017-06-19
Use rel() more consistently in theme elements, take two

karawoo:
This is the continuation of #1883. Includes @baptiste's changes plus a line in `NEWS.md` about the change.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 2.2.1.9000
 
+* Uses `rel()` more consistently in theme defaults (@baptiste).
+
 * `geom_density` drops groups with fewer than two data points and throws a
   warning. For groups with two data points, the density values are now
   calculated with `stats::density` (@karawoo, #2127).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # ggplot2 2.2.1.9000
 
-* Uses `rel()` more consistently in theme defaults (@baptiste).
+* Use `rel()` to set line widths in theme defaults (@baptiste).
 
 * `geom_density` drops groups with fewer than two data points and throws a
   warning. For groups with two data points, the density values are now
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -134,7 +134,7 @@ theme_grey <- function(base_size = 11, base_family = "") {
     panel.background =   element_rect(fill = "grey92", colour = NA),
     panel.border =       element_blank(),
     panel.grid.major =   element_line(colour = "white"),
-    panel.grid.minor =   element_line(colour = "white", size = rel(0.25)),
+    panel.grid.minor =   element_line(colour = "white", size = rel(0.5)),
     panel.spacing =      unit(half_line, "pt"),
     panel.spacing.x =    NULL,
     panel.spacing.y =    NULL,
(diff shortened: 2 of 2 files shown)


karawoo:
Hmm, I'm surprised the Travis build passed. It looks like this change breaks the visual tests in `test-theme.r`. Some of the panel grid lines are a little lighter now. Here are some screenshots from `vdiffr::manage_cases()`

Before:

screen shot 2017-06-19 at 11 23 54 am

After:

screen shot 2017-06-19 at 11 24 05 am

R/theme-defaults.r
```diff
@@ -134,7 +134,7 @@ theme_grey <- function(base_size = 11, base_family = "") {
     panel.background =   element_rect(fill = "grey92", colour = NA),
     panel.border =       element_blank(),
     panel.grid.major =   element_line(colour = "white"),
-    panel.grid.minor =   element_line(colour = "white", size = 0.25),
+    panel.grid.minor =   element_line(colour = "white", size = rel(0.25)),
```
hadley: I noticed that these actually change the values of the width as the `base_size` (as far as I can tell) is set to 0.5 by `line`.  I think that means all the values inside `rel()` should be doubled.

It also suggests that before merging this PR we should make some basic visual tests for the themes.

hadley: Ooops, should've refreshed before commenting. Glad the tests picked this up! (Although I'm not sure what went wrong on travis)

karawoo:
I doubled the values inside `rel()` and that indeed fixed the failing visual tests on my machine. `theme_linedraw()` doesn't have an associated visual test, so making this change for that theme didn't affect anything testing-wise. Should I add a test for this one?

has2k1:
Tests failed on travis but not on appveyor, this is the opposite of #2080.

R/theme-defaults.r
```diff
@@ -204,15 +204,15 @@ theme_linedraw <- function(base_size = 11, base_family = "") {
   theme_bw(base_size = base_size, base_family = base_family) %+replace%
     theme(
       # black text and ticks on the axes
-      axis.text        = element_text(colour = "black", size = rel(0.8)),
-      axis.ticks       = element_line(colour = "black", size = 0.25),
+      axis.text        = element_text(colour = "black", size = rel(1.6)),
```
hadley: I think this should stay as is since it was already in a `rel()`?

karawoo: Whoops yes, will change it back.

hadley:
@karawoo do you mind doing a separate PR that ensures that there's one visual test for each theme that's bundled with ggplot2? Then we can re-run this on top of that, and be confident that we haven't changed anything accidentally.

baptiste:
My original reason to look at this piece of source code and notice inconsistencies was triggered by a small plot in the margin of a Tufte-style document that looked quite bad/disproportionate. I think it may be worth setting up visual tests for a couple of different plot sizes, e.g. i) a 4x3 inches figure with 11pt font etc. for a typical publication, ii) a 24pt font, bold lines, 8x4in plot for a powerpoint, iii) a small inset-style / margin figure with 9pt font and thin lines.

karawoo:
I rebased these changes now that we've got the new test from #2174.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # ggplot2 2.2.1.9000
 
+* Uses `rel()` more consistently in theme defaults (@baptiste).
```
hadley: I think you can be a bit more explicit: use `rel()` to set line widths in theme defaults.

PR #2174, opened 2017-06-19
Make sure all themes have visual tests

karawoo:
Adds visual tests for `theme_linedraw()` and `theme_test()`.


hadley:
You don't need to test `theme_test()` (and in fact you definitely shouldn't test it), because it's the theme used for most tests and should never change (if it does will have much breakage in the tests).

Per @baptiste's suggestion, can you please also test all with `base_size = 33` (under the assumption if it looks good at 100% and 300% it's probably good in between)

karawoo:
Ah ok, I reverted the `theme_test()` test.

I'm a little confused about what the new tests should be testing. As far as I can tell the `base_size` argument only affects text elements (and a few margins). The changes in [2173](https://github.com/tidyverse/ggplot2/pull/2173) don't alter that, so I'm not sure if this will address @baptiste's concerns about the line widths. Should line/rect elements scale with `base_size` as well?

hadley:
Oh good point. Maybe we should have a base line size too?

karawoo:
I can add that if you want. Would this be additional argument to the theme functions, or would it be better to set the base line size within the theme according to the base font size, and then let people override that with `+ theme(line = element_line(size = 1))` or whatever? Either way this seems like it deserves its own PR. I think the AppVeyor build failed on this one for reasons unrelated to the changes I made, but I don't see a way to force start a new build the way I can with Travis...

hadley:
I think it would work in the same way as `base_size` - i.e. `line = element_line(size = base_line_size)`. I guess it would default to `base_size / 22`?  I agree that this should be a different PR, and you can go ahead and merge this one.

Issue #1526, opened 2016-01-28
Scale types overhaul


hadley:
- Make sure all scales work with dates/datetimes
- Better colour/fill scales for ordered factors
- Warn when mapping factor to size
- Consider different viridis scales (using viridisLite) for fill/colour
- Better default breaks for log axes
- Document how to extend with your own S3 classes
- Consider support for percentage, proportion, and likert s3 types
- Could we have a `jitter_na()` type for position scales?

zeehio:
Hi,

Sorry to bother you, I believe I found an issue related to this commit:

It seems that functions defined in [`R/zxx.R`](https://github.com/hadley/ggplot2/blob/master/R/zxx.r#L16) such as `scale_colour_datetime` do not accept common scale arguments such as `name=`.

```r
ggplot(data.frame(x=c(1,2,3),
                  y=c(2,5,6),
                  color = as.Date(c(1,2,3), origin="2015-01-01"))) + 
   geom_point(aes(x=x, y=y, color=color))  +
   scale_colour_datetime(name="a")
```

    Error in scale_colour_datetime(name = "a") : unused argument (name = "a")

Found as of:

     ggplot2               * 2.1.0   2016-03-01 CRAN (R 3.2.5) 

It seems this is work in progress, so I don't want to step in too much... I hope that reporting this small issue is useful.

Best and thanks for your awesome work!

hadley:
@zeehio please file a separate issue instead of adding a comment to an unrelated one

huftis:
@hadley Regarding your last point, ‘different colour scales for fill/colour’: If you’re thinking about the *continuous* colour scales, I recommend watching this (fascinating!) [video about the new default colour scale in matplotlib](http://matplotlib.org/style_changes.html). (There’s also [an R implementation](https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html).)

hadley:
@huftis yup I'm aware of viridis, and I'm considering making it the default.

hadley:
One hacky way of better display for missing values:

```R
add_na <- function(x, prop = 0.1) {
  x[sample(seq_along(x) <= prop * length(x))] <- NA
  x
}

df <- tibble(
  x = letters[1:3] %>% sample(100, replace = TRUE) %>% add_na(),
  y = rnorm(100) %>% add_na()
)

ggplot(df, aes(x, y)) +
  geom_point()

miss <- df %>% filter(is.na(y))

[...]

PR #2178, opened 2017-06-22
Add built-in support for viridis color scales

karawoo:
Related to [1526](https://github.com/tidyverse/ggplot2/issues/1526). This is a first stab at adding built-in support for `viridis` color scales using the `viridisLite` package. The function names are kind of long and awkward right now, but I'm not sure what better names would be. Also, should the `viridis_pal` function go as a PR to scales instead of ggplot2?

PR #2179, opened 2017-06-23
Improve scales behavior for ordered factors

karawoo:
Related to [1526](https://github.com/tidyverse/ggplot2/issues/1526). Ordered factors and regular factors currently behave pretty much the same, but this is not always appropriate. I've added new ordinal scales for `size`, `alpha`, and `shape` so that

- [x] `scale_shape_ordinal` throws a warning (but `scale_shape_discrete` still does not)
- [x] `scale_size_ordinal` does not throw a warning (but `scale_size_discrete` still does)
- [x] `scale_alpha_ordinal` still does not throw a warning (but `scale_alpha_discrete` does)

Still to do:

- [x] new colour/fill scales for ordinal variables (possibly with viridis)
- [x] test the new behavior


karawoo:
Some of this could be DRY-er, e.g.

```r
scale_size_ordinal <- function(...) {
  suppressWarnings(scale_size_discrete(...))
}
```

might be better.

karawoo:
If we're going to go with viridis for ordinal scales then we may not need any additional functions, but if not then perhaps something like this should go in the scales package:

```r
seq_ordinal_pal <- function(low, high) {
  function(n) {
    v <- seq(0, 1, length.out = n)
    scales::colour_ramp(c(low, high))(v)  
  }
}
```

Then `scale_colour_ordinal` would be:

```r
scale_colour_ordinal <- function(..., low = "#132B43", high = "#56B1F7",
                                 na.value = "grey50") {
  discrete_scale(
    "colour",
    "sequence",
    seq_ordinal_pal(low, high),
    na.value = na.value,
    ...
    )
}
```

which produces output like this:

```r
d <- diamonds[sample(nrow(diamonds), 1000), ]

ggplot(d, aes(carat, price)) +
  geom_point(aes(colour = cut))
```

diamonds

has2k1:
You need to be careful (or document the aberrations), depending on how you interpolate between the high and low colours of a colourmap, not all generated colours may be part of the continuous version of the colourmap. 

This will definitely affect colourmaps that aim to perceptually-uniform, like viridis.

karawoo:
@hadley GitHub isn't letting me request a review but can you tell me what you think of this? I can't remember if we decided on Monday what to do about the ordinal color scales.

hadley:
Looks good so far. We should make a decision about ordinal colour scales too. Using viridis seems like a reasonable choice to me.

karawoo:
Trying @jennybc's edits from #2189 in this PR to see if it makes codecov chill out.

karawoo:
Using viridis sounds good to me, so once #2178 gets merged I can add that here.

karawoo:
I removed Jenny's codecov-related commits and added `scale_fill_ordinal` and `scale_colour_ordinal`, which both default to viridis.

karawoo:
@hadley one quick question before I merge -- since we're making viridis the default for ordered factors, should `viridisLite` go in Imports rather than Suggests?

hadley:
Yes!

PR #2181, opened 2017-06-25
allow controlling ndiscr in call to sf::st_graticule

edzer:
see https://github.com/edzer/sfr/issues/396 for the motivating issue. This needs also https://github.com/edzer/sfr/commit/c9e56df2feebda344f9dcf07429a9711be026b66


hadley:
Thanks!

@karawoo can you please ask @jennybc what we need to do to make the annoying code coverage failures go away?

PR #2184, opened 2017-06-27
Clarify stat/geom_smooth docs to show method takes a fn, not a character

noamross:
stat/geom_smooth lists 'rlm' as a possible method
for calulating a smooth, but method='rlm' fails
becuase rlm() is in MASS and is not imported. This
fixes this by importing MASS::rlm.

### DIFF
--- a/R/geom-smooth.r
+++ b/R/geom-smooth.r
@@ -39,18 +39,18 @@
 #' # Instead of a loess smooth, you can use any other modelling function:
 #' ggplot(mpg, aes(displ, hwy)) +
 #'   geom_point() +
-#'   geom_smooth(method = "lm", se = FALSE)
+#'   geom_smooth(method = lm, se = FALSE)
 #'
 #' ggplot(mpg, aes(displ, hwy)) +
 #'   geom_point() +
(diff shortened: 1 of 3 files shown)


hadley:
Doesn't `method = MASS::rlm` work?

noamross:
Ah, yes.  `method = MASS::rlm` works, `method = 'MASS::rlm'` does not.  Maybe I was just confused, but perhaps the docs could be clarified by using non-quoted functions. In the help file there is:
```
`method`	smoothing method (function) to use, eg. "lm", "glm", "gam", "loess", "rlm".
```

So I thought "rlm" would work.  Maybe this would be better?:

```
`method`	smoothing method (function) to use, eg. `lm`, `glm`, `gam`, `loess`, `MASS::rlm`.
```

hadley:
Yeah, that'd be better. Want to do a PR for the docs?

noamross:
Will do.  I will update this one.

noamross:
Reverted the code and updated the docs, squashed to one commit

noamross:
Re-committed with trim trailing whitespace option turned off in RStudio.

hadley:
Thanks!

PR #2186, opened 2017-06-29
Improve datetime scales

karawoo:
Related to #1526. This PR: 

- Renames `scale_datetime()` to `datetime_scale()` and generalizes it for use with other scales besides x/y position.
- Uses the new `datetime_scale()` for size, colour, fill, and a newly created alpha scale. These scales can now all take `date_breaks` and `date_labels` parameters.

I also tidyverse style-ified `scale_(x|y)_date` and `scale_(x|y)_datetime`.


R/scale-date.r
```diff
@@ -197,20 +256,35 @@ scale_datetime <- function(aesthetics, trans,
...
     time = "datetime"
   )
-  scale_class <- switch(trans,
-    date = ScaleContinuousDate,
-    time = ScaleContinuousDatetime
-  )
+
+  # x/y position aesthetics should use ScaleContinuousDate or
+  # ScaleContinuousDatetime; others use ScaleContinuous
+  if (all(aesthetics %in% c("x", "xmin", "xmax", "xend", "y", "ymin", "ymax", "yend"))) {
```
karawoo: Not sure if this is the best way to choose the appropriate super class...

hadley: Can you remind me why we use `ScaleContinuous` for the non-position scales?

karawoo: It's so the data will get rescaled correctly, I think. `ScaleContinuous` uses `scales::rescale()` to map the data values to the right place on the scale. For position scales this isn't needed because the coordinate system handles it.

karawoo:
There are a few existing tests for datetime scales, mostly they check that timezones are correct. I can add some more.

karawoo:
I just realized that with this will make colour/fill scales behave differently for datetimes than for other continuous variables because they won't be controlled by the global options `ggplot2.continuous.colour` and `ggplot2.continuous.fill` from #2178.

hadley:
Oh hmmmm, how hard would it be to respect that option?

karawoo:
It's a little tricky because `scale_colour_gradient()` and `scale_colour_viridis_c()` don't take the `date_breaks` and `date_labels` parameters. I guess we'd need `scale_colour_gradient_date()` and `scale_colour_viridis_date()` (+ comparable fill scales, and datetime scales), and then `scale_colour_date()` can choose the right one among those. Does that sound too unwieldy?

hadley:
That seems like a lot of work for little gain. Let's leave off for now and wait to see if anyone notices/complains.

PR #2189, opened 2017-07-03
Don't let codecov fail build with "No report found to compare against"

jennybc:
From here: https://github.com/codecov/support/issues/173#issuecomment-216067683

This might help with codecov failing to find a report for the base commit. Let's find out ...

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,7 @@
 comment: false
+coverage:
+  status:
+    project:
+      default:
+        target: auto           # this is default
+        if_not_found: success  # no commit found? still set a success


jennybc:
It looks like this has been an ongoing problem that is, in theory, fixed, but in reality maybe not, based on all the +1's here: https://github.com/codecov/support/issues/363

So it might be reasonable to simply not allow codecov to fail a build for this reason. Which is the intent of this PR.

karawoo:
@hadley shall we give this a try? It seems to have (retroactively??) fixed the codecov issue in #2179. I guess the other option is "just pray codecov sorts itself out" -- my last PR (#2190) didn't have any problems.

jennybc:
It seems like you got your codecov failures for inexplicable and self-resolving reasons. Which I have also experienced. I'd err on the side of not doing anything rash unless the problem persists.

It would be different if codecov was failing builds due to picky reasons about coverage percentage. I would take control of that in the yaml. But this "no report" error is a different problem.

hadley:
Ok, let's leave for now, and if we don't see any more problems we can close this PR.

PR #2195, opened 2017-07-06
Add room around guide title 

karawoo:
This is @baptiste's change from #1882, but it breaks the visual tests because the spacing between the title and legend elements is slightly changed. @hadley should I validate those with `vdiffr` and commit the new SVGs?


baptiste:
you may want to review [guide-colorbar.R](https://github.com/tidyverse/ggplot2/blob/84eef2bf75e2fc16cbfaa38c835ca0fc9a752996/R/guide-colorbar.r#L259) for consistency, and IIRC there were other hard-coded values that should probably be set via the theme.

hadley:
@karawoo yes please!

Issue #2106, opened 2017-04-13
position dodge fails to dodge all overlapping values


hadley:
``` r
library(tidyverse)
#> Loading tidyverse: ggplot2
#> Loading tidyverse: tibble
#> Loading tidyverse: tidyr
#> Loading tidyverse: readr
#> Loading tidyverse: purrr
#> Loading tidyverse: dplyr
#> Conflicts with tidy packages ----------------------------------------------
#> filter(): dplyr, stats
#> lag():    dplyr, stats

df <- tibble(
  x = c("a", "a", "b"),
  n = c(1, 5, 10)
)

ggplot(df, aes(1, n, fill = x)) + 
  geom_col(position = "dodge", alpha = 0.5)
```

![](http://i.imgur.com/YJG5Mzx.png)

karawoo:
Fixed in #2196

Issue #2076, opened 2017-03-20
position_dodge(perserve="single") shifts everything left


alistaire47:
The new `preserve` parameter of `position_dodge` is awesome (#1935, I think), but when applied to `stat_count` or `stat_boxplot` (maybe more), with missing levels that aren't the final one, it shifts groups to the left to fill the empty level:

``` r
library(ggplot2)

df <- data.frame(x = rep(letters[1:3], each = 3), 
                 fill = factor(c(2,2,3, 1:3, 3,3,3)))

ggplot(df, aes(x, fill = fill)) + 
    geom_bar(position = position_dodge(preserve = 'single'))
```
![plot with left-shifted bars](http://i.imgur.com/w948wmq.png)

``` r
ggplot(mtcars, aes(factor(vs), mpg, fill = factor(cyl, c(8,6,4)))) + 
    geom_boxplot(position = position_dodge(preserve = 'single'))
```
![boxplot example](http://i.imgur.com/hbPrBqw.png)

While that behavior makes sense given that these stats previously didn't care about empty groups didn't need drawing or spacing, I suspect most users would expect bars to keep their group position à la `geom_col`/`stat_identity` (which has 0s to keep things in place):

``` r
ggplot(data.frame(table(df)), aes(x, Freq, fill = fill)) + 
    geom_col(position = 'dodge'))
```
![plot with normally-arranged bars](http://i.imgur.com/mUylAaz.png)

Is this desired behavior or a bug?

has2k1:
I encountered this "feature" and I think it is hard to fix. At the instant when the dodging is done, the `group` information is insufficient. I think you would have to change/add another a group/sub-group that can be used by the position aesthetics. That would be a significant change to the underlying data model that has been constant and sufficient for a very long time.

May be there is another way, I may look at it again.

karawoo:
The new `position_dodge2()` function improves this by centering each group of bars (or boxes, etc.) at their x position.

PR #2196, opened 2017-07-06
New dodging algorithm for box plots

### LINKED ISSUE #2143: overlapping boxes with geom_boxplot(varwidth=TRUE)
When I try to produce boxplots with colours depending on a categorical variable, these appear overlapping if varwidth is set to TRUE (which is what I'd like to use). 

Here's an example:
```r
ggplot(data=iris, aes(Species, Sepal.Length)) + 
   geom_boxplot(aes(colour=Sepal.Width < 3.2), varwidth=TRUE)
```
This raises the following warning: "position_dodge requires non-overlapping x intervals"

However, the boxes don't overlap (and no warning is raised) if varwidth is not specified.
```r
ggplot(data=iris, aes(Species, Sepal.Length)) + 
   geom_boxplot(aes(colour=Sepal.Width < 3.2))
```

karawoo:
Fixes #2143. This creates a new ggproto object for box plot positions, `PositionBoxdodge`, and a function `pos_boxdodge()` that can handle regular and variable-width boxes. `collide()` gets split into several functions. `collide_setup()` does the initial setup of the data and gets called by both `collide()` (which behaves the same as it has been) or `collide_box()`.

@hadley this is slightly different than the algorithm we talked about because it doesn't use a for loop to find the non-overlapping groups. The `x` value in the data already seems to capture that information, so I think we can avoid the loop, but maybe I'm missing something...

With these changes, box plots look like this:

``` r
library("ggplot2")

ggplot(data = iris, aes(Species, Sepal.Length)) +
  geom_boxplot(aes(colour = Sepal.Width < 3.2))
```

``` r

ggplot(data = iris, aes(Species, Sepal.Length)) +
  geom_boxplot(aes(colour = Sepal.Width < 3.2), varwidth = TRUE)
```


hadley:
Can you include an example with continuous x? I think that's what I was thinking of for the for loop.

hadley:
It also seems like we need a little padding between the individual elements. Could we make that a parameter?

karawoo:
Ohhh continuous x. Yeah you're right, this doesn't work well for that.

``` r
library("ggplot2")

## Without varwidth = TRUE it's ok aside from the padding issue
ggplot(diamonds, aes(y = price)) +
  geom_boxplot(aes(x = cut_width(carat, 0.5), fill = depth < 65))
```



``` r

## With the changes from this PR and varwidth = TRUE
ggplot(diamonds, aes(y = price)) +
  geom_boxplot(aes(x = cut_width(carat, 0.5), fill = depth < 65), varwidth = TRUE)
```



Here's how the second plot looks with the current dev version of ggplot.

``` r
ggplot(diamonds, aes(y = price)) +
  geom_boxplot(aes(x = cut_width(carat, 0.5), fill = depth < 65), varwidth = TRUE)
#> Warning: position_dodge requires non-overlapping x intervals
```



for loop it is, then!

karawoo:
Oh, and there is a larger problem with this: each pair of boxes is getting scaled separately to the group width, which makes the varying widths from `varwidth` meaningless.

mcol:
Note that the order of the bars in the second example of the OP is different from the rest (I've confirmed it with a checkout of your current position-dodge branch).

karawoo:
Thanks for pointing that out, @mcol. Right now the boxes get placed in the order in which they appear in the data passed to `pos_boxdodge()` and I need to change that to enforce consistency across groups of boxes.

karawoo:
Also, my example above doesn't really have a continuous x variable since `cut_width()` converts it to a factor. Here is a better example for testing with (that is still broken):

``` r
set.seed(582)
dat <- data.frame(x = sample(4:5, size = 20, replace = TRUE),
                  y = rnorm(20),
                  class = sample(c("a", "b"), size = 20, replace = TRUE))

ggplot(dat, aes(x = x, y = y)) +
  geom_boxplot(aes(group = interaction(x, class), fill = class), varwidth = TRUE)
```

karawoo:
Oh dear, I rebased some things and created a little git mess, I'll fix this.

karawoo:
Alright, here is the new box plot algorithm. Below are a bunch of examples of how it looks in practice with continuous and discrete x variables and various combinations of `varwidth = TRUE`, `varwidth = FALSE`, `preserve = "single"`, and `preserve = "total"`. Right now `preserve = "single"` is the default for all box plots, but I could change that so that it's only the default when `varwidth = TRUE`. This definitely needs some more tests, which I will add soon.

``` r
library("ggplot2")

## From the original example:
ggplot(data = iris, aes(Species, Sepal.Length)) +
  geom_boxplot(aes(colour = Sepal.Width < 3.2))
```



``` r

ggplot(data = iris, aes(Species, Sepal.Length)) +
  geom_boxplot(aes(colour = Sepal.Width < 3.2), varwidth = TRUE)
```



``` r

## Diamonds data -- padding between boxes is less noticeable
ggplot(diamonds, aes(y = price)) +
  geom_boxplot(aes(x = cut_width(carat, 0.5), fill = depth < 65), varwidth = TRUE)
```



``` r
 
## A plot with truly continuous x
set.seed(582)
dat <- data.frame(x = sample(4:5, size = 20, replace = TRUE),
                  y = rnorm(20),
                  class = sample(c("a", "b"), size = 20, replace = TRUE))

ggplot(dat, aes(x = x, y = y)) +
  geom_boxplot(aes(group = interaction(x, class), fill = class))
```



``` r

## Truly continuous x and varwidth = TRUE
ggplot(dat, aes(x = x, y = y)) +
  geom_boxplot(aes(group = interaction(x, class), fill = class), varwidth = TRUE)
```



``` r

## Preserve total width
ggplot(mtcars, aes(factor(cyl), y = mpg, fill = factor(vs))) +
  geom_boxplot(position = position_boxdodge(preserve = "total"))
```



``` r

## Preserve individual width with varwidth = TRUE
ggplot(mtcars, aes(factor(cyl), y = mpg, fill = factor(vs))) +
  geom_boxplot(position = position_boxdodge(preserve = "single"), varwidth = TRUE)
```



``` r

## preserve = "total" is incompatible with varwidth = TRUE
ggplot(mtcars, aes(factor(cyl), y = mpg, fill = factor(vs))) +
  geom_boxplot(position = position_boxdodge(preserve = "total"), varwidth = TRUE)
#> Warning: Can't preserve total widths when varwidth = TRUE.
```

R/position-boxdodge.r
```diff
@@ -0,0 +1,116 @@
+#' Position dodge for box plots
```
hadley: I'd like to give this a name that reflects that it works for any geom with variable widths - i.e. it would also work for `geom_rect()`.

R/position-boxdodge.r
```diff
@@ -0,0 +1,116 @@
...
+#' 
+#' @include position-dodge.r
+#' @inheritParams position_dodge
+#' @param padding Padding between boxes at the same position. Boxes are shrunk
+#'   by this proportion to make room for space between them.
+#' @family position adjustments
+#' @export
+#' @examples
+#' ggplot(data = iris, aes(Species, Sepal.Length)) +
+#' geom_boxplot(aes(colour = Sepal.Width < 3.2))
```
hadley: Missing indent

R/position-boxdodge.r
```diff
@@ -0,0 +1,116 @@
...
+
+pos_boxdodge <- function(df, width, n = NULL, padding = 0.1) {
+   
+  if (!all(c("xmin", "xmax") %in% names(df))) {
+    df$xmin <- df$x
+    df$xmax <- df$x
+  }
+  
+  # xid represents groups of boxes that share the same position
+  df$xid <- match(df$x, sort(unique(df$x)))
```
hadley: Oh this is why it won't work with `geom_rect()`. It's only a few lines of code, so I think it's worth using the for-loop that I originally proposed

R/position-boxdodge.r
```diff
@@ -0,0 +1,116 @@
...
+    df$new_width <- (df$xmax - df$xmin) / n[df$xid]
+  } else {
+    df$new_width <- (df$xmax - df$xmin) / n
+  }
+  
+  df$xmin <- df$x - (df$new_width / 2)
+  df$xmax <- df$x + (df$new_width / 2)
+
+  # Find the total width of each group of boxes
+  group_sizes <- plyr::ddply(df, "xid", plyr::summarize, size = sum(new_width))
```
hadley: Could this be replaced by a `tapply()`?  I'd prefer to not use plyr in new code, since one day I'd like to eliminate the dependency.

R/position-boxdodge.r
```diff
@@ -0,0 +1,116 @@
...
+
+  # Set the boxes in place
+  for (i in seq_along(starts)) {
+    divisions <- cumsum(c(starts[i], df[df$xid == i, "new_width"]))
+    df[df$xid == i, "xmin"] <- divisions[-length(divisions)]
+    df[df$xid == i, "xmax"] <- divisions[-1]
+  }
+
+  # x values get moved to between xmin and xmax
+  df$x <- rowMeans(df[, c("xmin", "xmax")])
```
hadley: I think `df$x <- (df$xmin + df$xmax) / 2` would be clearer

tests/testthat/test-geom-boxplot.R
```diff
@@ -38,6 +38,22 @@ test_that("can use US spelling of colour", {
...
 
+test_that("boxes with variable widths do not overlap", {
+  p <- ggplot(data = iris, aes(Species, Sepal.Length)) +
+    geom_boxplot(aes(colour = Sepal.Width < 3.2), varwidth = TRUE)
+  d <- layer_data(p)[c("xmin", "xmax")]
+  
+  overlaps <- vector(length = nrow(d) - 1)  
+  for (i in 2:nrow(d)) {
+    if (d$xmin[i] < d$xmax[i - 1]) {
+      print(i)
```
hadley: Need to remove this `print()`

tests/testthat/test-geom-boxplot.R
```diff
@@ -38,6 +38,22 @@ test_that("can use US spelling of colour", {
...
 })
 
+test_that("boxes with variable widths do not overlap", {
+  p <- ggplot(data = iris, aes(Species, Sepal.Length)) +
+    geom_boxplot(aes(colour = Sepal.Width < 3.2), varwidth = TRUE)
+  d <- layer_data(p)[c("xmin", "xmax")]
+  
+  overlaps <- vector(length = nrow(d) - 1)  
+  for (i in 2:nrow(d)) {
+    if (d$xmin[i] < d$xmax[i - 1]) {
```
hadley: This for loop could become (with a small modification) a separate `group_overlapping` function, and you could use above.

hadley:
The default padding looks slightly too big to my eyes — maybe try 0.05.

It would also be handy to show a couple of examples using with other geoms (i.e. `geom_bar()`)

R/position-boxdodge.r
karawoo: Oh yeah sure. `aggregate()` might be even better since it'll return a data frame.

karawoo:
I've made some tweaks that allow this to work with `geom_rect`, but this requires some more thought about how to handle y position. Prior to this PR, overlapping rects look like this:

```r
df <- data.frame(
  xmin = c(1, 3, 6),
  xmax = c(5, 7, 9),
  ymin = c(1, 1, 5),
  ymax = c(3, 4, 8),
  fill = c("a", "b", "c")
)

ggplot(df, aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax, fill = fill)) +
  geom_rect(alpha = 0.8)
```

geom_rect_current

The existing `position_dodge()` does not work with rects because it requires an `x` aesthetic.

With the changes in this PR `position_dodge()` and `position_boxdodge()` work, but the results are a little strange:

```r
ggplot(df, aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax, fill = fill)) +
  geom_rect(position = "boxdodge", alpha = 0.8)
```

geom_rect_boxdodge

First, the x values are meaningless (this is not related to the rects, it's a different bug that also appears in the continuous x plots before -- I'll fix that). But also, dodging doesn't take y position into account at all, so it doesn't know that the green and blue boxes don't overlap. Should those boxes be dodged, or should I factor in y position as well?

R/position-boxdodge.r
karawoo: `position_vardodge()`?

karawoo:
I fixed the issue with the x scale; this is now working well for box plots, bars, and arbitrary rects. I'm starting to wonder if this should just be the new algorithm for `position_dodge()`. It would address a lot of the other `position_dodge` issues. #2153 would be improved because all of the groups of bars are centered at their x value, regardless of how many bars are there (see the third plot below). This might be better than the current behavior for #2076 as well. And it would resolve #2106. We'd just need to make sure that the padding argument got set to zero for bars, and that the default values for `preserve` remained consistent.

``` r
library("ggplot2")

## A plot with truly continuous x
set.seed(582)
dat <- data.frame(x = sample(4:5, size = 20, replace = TRUE),
                  y = rnorm(20),
                  class = sample(c("a", "b"), size = 20, replace = TRUE))

ggplot(dat, aes(x = x, y = y)) +
  geom_boxplot(aes(group = interaction(x, class), fill = class), varwidth = TRUE)
```



``` r
## geom_bar
ggplot(mtcars, aes(factor(cyl), fill = factor(vs))) +
  geom_bar(position = position_boxdodge(preserve = "single"))
```



``` r
## Facetting
ggplot(mtcars, aes(factor(cyl), fill = factor(vs))) +
  geom_bar(position = position_boxdodge(preserve = "single")) +
  facet_wrap(~ am)
```



``` r
## Issue #2106
library("tibble")
df <- tibble(
  x = c("a", "a", "b"),
  n = c(1, 5, 10)
)

ggplot(df, aes(1, n, fill = x)) + 
  geom_col(position = "boxdodge", alpha = 0.5)
```



A `geom_rect` example:

``` r
## geom_rect
df <- data.frame(
  xmin = c(1, 3, 6, 11, 13),
  xmax = c(5, 7, 9, 15, 16),
  ymin = c(1, 1, 5, 2, 2),
  ymax = c(3, 4, 8, 6, 7),
  fill = c("a", "b", "c", "a", "b")
)

## no dodging
ggplot(df, aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax, fill = fill)) +
  geom_rect(alpha = 0.8)
```



``` r
## With boxdodge
ggplot(df, aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax, fill = fill)) +
  geom_rect(position = "boxdodge", alpha = 0.8)
```

R/position-boxdodge.r
karawoo: Or maybe `position_flexdodge()`, since it's a more flexible dodge that can handle variable width boxes and arbitrary rectangles?

hadley:
I like the idea of making it the default. Then we could simply call it `position_dodge2()`, since it's basically version 2 of the dodging algorithm.

karawoo:
I did some more research and it'll take a bit more work before `position_dodge2()` can be used in all the places `dodge` is currently used. Right now it doesn't work well with dot plots, violins, or error bars.

hadley:
Ok. Let's make it the default for boxplots and bars and then take a look at the others later.

karawoo:
I've made some slight changes to hopefully make this as easy to use for both boxes and bars as possible. `position_dodge2()` now defaults to having 0 padding, making it work nicely with bars -- you just have to add `position = "dodge2"` to override the default stacking position. `geom_boxplot()` uses a default position of `position_dodge2(padding = 0.05)`, so boxes still get padding by default.

Something has gone wrong again with the ordering of the boxes/bars compared to the legend, I'll track that down and fix it.

``` r
ggplot(data = iris, aes(Species, Sepal.Length)) +
  geom_boxplot(aes(colour = Sepal.Width < 3.2), varwidth = TRUE)
```



``` r
ggplot(mtcars, aes(factor(cyl), fill = factor(vs))) +
  geom_bar(position = "dodge2")
```

hadley:
This is looking great! (Although now padding of 0.05 looks too small. Maybe switch back to 0.10?)

karawoo:
Soooo close. I am rethinking the defaults that I changed above. Right now this adds padding between the boxes because `geom_boxplot`'s default position is `position_dodge2(padding = 0.1)`:

```r
ggplot(mtcars, aes(factor(cyl), y = mpg, fill = factor(vs))) +
  geom_boxplot()
```

But if someone wants to change the `preserve` argument, they're likely going to do:

```r
ggplot(mtcars, aes(factor(cyl), y = mpg, fill = factor(vs))) +
  geom_boxplot(position = position_dodge2(preserve = "total"))
```

which results in no padding (they'd need to also add `padding = 0.1`). So maybe it's better to have `position_dodge2()` default to adding padding, which will be consistent for box plots even if other position parameters get changed. And for bars, people will have to override that default if they don't want space between them.

hadley:
Yeah, I think that's a good idea. I rather like the padding between bars too.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # ggplot2 2.2.1.9000
 
+* Box plot position is now controlled by `position_dodge2()` (@karawoo,
+  #2143).
```
hadley: And `bar`. And you should add a brief descrption.

R/position-dodge2.r
```diff
@@ -0,0 +1,154 @@
+#' Alternate method for dodging overlapping objects
```
hadley: Maybe it's worth documenting together with `position_dodge()`?

Issue #2147, opened 2017-05-29
Log transform on geom_bar() silently omits layer


krlmlr:
After some thinking, `geom_bar()` cannot work with `coord_trans(y = "log10")` (https://github.com/tidyverse/ggplot2/issues/420#issuecomment-4247826). An error message (or a warning) would be great, though.

``` r
library(tidyverse, quietly = TRUE)
diamonds %>% ggplot() + geom_bar(aes(x = cut)) + coord_trans(y = "sqrt")
```

![](http://i.imgur.com/Kp9MOSa.png)

``` r
diamonds %>% ggplot() + geom_bar(aes(x = cut)) + coord_trans(y = "log10")
```

![](http://i.imgur.com/LEMvdRn.png)

has2k1:
I think there is a divide-by-zero in there somewhere that leads to infinite coordinate axes.

hadley:
`log(0)` becomes -Inf, but this should generate a warning or error somewhere.

karawoo:
Closed in #2210

PR #2210, opened 2017-07-16
Add warning for coord_trans() when resulting x/y value becomes non-finite (#2147)

foo-bar-baz-qux:
`coord_trans()` now generates a warning when a transformation results in x or y values being non-finite (#2147).

I've used `#` comments for the internal helper function so it doesn't generate MD documentation files, but not sure what the ggplot2 style is for those functions.

The logic and the message is based on the [transform in the Scale object](https://github.com/tidyverse/ggplot2/blob/4255cd8341ad1072036186e5d34ad2c1b12221b3/R/scale-.r#L207) in order to be consistent.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 2.2.1.9000
 
+* `coord_trans()` now generates a warning when a transformation results in x or y 
+  values being non-finite (@foo-bar-baz-qux, #2147).
+  
 * Legend titles and labels get a little extra space around them. Legend titles
   will no longer overlap the legend at large font sizes (@karawoo, #1881).
 
--- a/R/coord-transform.r
+++ b/R/coord-transform.r
@@ -179,11 +179,12 @@ train_trans <- function(scale, limits, trans, name) {
   out
 }
 
-# Generate warning when finite values are transformed into infinite values
-#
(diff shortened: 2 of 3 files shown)


R/coord-transform.r
```diff
@@ -174,3 +178,14 @@ train_trans <- function(scale, limits, trans, name) {
   names(out) <- paste(name, names(out), sep = ".")
   out
 }
+
+# Generate warning when finite values are transformed into infinite values
+#
+# @param old_values A vector of pre-transformation values.
+# @param new_values A vector of post-transformation values.
+# @param axis Which axis the values originate from (e.g. x, y).
```
hadley: I think it's slightly better to use `#'` and add `@noRd`. 

Would you mind filing an issue or a PR over at https://github.com/tidyverse/style to note that?

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # ggplot2 2.2.1.9000
 
+* coord_trans()` now generates a warning when a transformation results in x or y
```
hadley: Missing a `` ` ``

R/coord-transform.r
foo-bar-baz-qux: Yep, I've filed that issue and updated with those suggestions, cheers.

PR #2212, opened 2017-07-19
Make `debug = TRUE` work when `expand_x` and `expand_y` are FALSE

karawoo:
Currently `titleGrob()`'s `debug = TRUE` argument has no effect when `expand_x` and `expand_y` are both `FALSE` because the function returns output before the debugging conditional is reached. This fixes the issue by moving the debugging conditional up and ensuring that all of the `children` are returned.

Previous behavior:

```r

df <- data.frame(x = 1:10, y = 1:10, color = LETTERS[1:2])

p <- ggplot(df, aes(x, y, color = color)) +
  geom_point() +
  theme(
    legend.title = element_text(debug = TRUE),
    legend.text = element_text(debug = TRUE)
  )

p # note no debugging output
```

debug_pre

After this PR:

```r
p
```

debug-post

We could also address #1881 here by using `text_height` and `text_width` instead of `unit(1, "grobheight", text_grob)` and `unit(1, "grobwidth", text_grob)` for the widths and heights. This puts more spacing between the legend title and the legend elements, so the visual tests break.

### DIFF
--- a/R/margins.R
+++ b/R/margins.R
@@ -61,7 +61,9 @@ titleGrob <- function(label, x, y, hjust, vjust, angle = 0, gp = gpar(),
   text_height <- unit(1, "grobheight", text_grob) + cos(angle / 180 * pi) * descent
   text_width <- unit(1, "grobwidth", text_grob) + sin(angle / 180 * pi) * descent
 
-  if (debug) {
+  if (is.null(debug)) {
+    children <- gList(text_grob)
+  } else if (debug) {
     children <- gList(
       rectGrob(gp = gpar(fill = "cornsilk", col = NA)),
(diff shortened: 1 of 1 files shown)


R/margins.R
```diff
@@ -61,6 +61,18 @@ titleGrob <- function(label, x, y, hjust, vjust, angle = 0, gp = gpar(),
   text_height <- unit(1, "grobheight", text_grob) + cos(angle / 180 * pi) * descent
   text_width <- unit(1, "grobwidth", text_grob) + sin(angle / 180 * pi) * descent
 
+  if (is.null(debug)) {
+    children <- gList(text_grob)
+  } else if (debug) {
```
hadley: Maybe make this `isTRUE(debug)` so you only need the else branch?

PR #2216, opened 2017-07-20
fix various geom_sf related issues

edzer:
see description of individual commits


R/sf.R
```diff
@@ -61,6 +61,10 @@
 #' @name ggsf
 NULL
 
+geom_column = function(data) {
+  data[[ which(sapply(data, function(x) inherits(x, "sfc")))[1] ]]
```
hadley: Can you please use `vapply()` here instead of `sapply()`? You should be able to drop the `function()` too.

(And `<-` instead of `=`)

R/sf.R
```diff
@@ -189,8 +193,9 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     }
 
     for (layer_data in data) {
-      geometry <- layer_data$geometry
-      if (is.null(geometry))
+      if (inherits(layer_data, "sf"))
```
hadley: Can you please use `{}` here?

hadley: Is it worth defining `is.sf()`? We use `inherits(x, "sf")` in a few places

R/sf.R
```diff
@@ -209,18 +214,17 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
...
-      layer_data$geometry <- sf::st_transform(layer_data$geometry, params$crs)
-      layer_data
+      sf::st_transform(layer_data, params$crs)
     })
   },
 
   transform = function(self, data, panel_params) {
-    data$geometry <- sf_rescale01(
-      data$geometry,
+    data[[ which(sapply(data, function(x) inherits(x, "sfc")))[1] ]] <- sf_rescale01(
```
hadley: Maybe `geom_column()` should return an index so you can use it on the LHS and RHS here?

karawoo:
I think adding `"geom_column"` to [here](https://github.com/tidyverse/ggplot2/blob/b2b58f19f86f92967ecc69f9882a47f2842aa2cb/tests/testthat/test-function-args.r#L16-L17) will fix the failing tests.

edzer:
If you want me to add tests to address the last check, please let me know where.

tests/testthat/test-geom-sf.R
```diff
@@ -0,0 +1,14 @@
+context("geom-sf")
+
+# Visual tests ------------------------------------------------------------
+
+test_that("geom_sf draws correctly", {
+  library(sf)
```
hadley: Since sf is only suggests, this should start with `skip_if_not_installed("sf")`

I think it's slightly nicer to not attach the package, but just `sf::read_sf()`

R/sf.R
```diff
@@ -61,6 +61,20 @@
 #' @name ggsf
 NULL
 
+geom_column <- function(data) {
+  w <- which(vapply(data, inherits, TRUE, what = "sfc"))
+  if (length(w) == 0) {
+    "geometry" # avoids breaks when objects without geometry list-column are examined
+  } else {
+    # this may not be best in case more than one geometry list-column is present:
```
hadley: Maybe if `length(w) > 1`, pick the first, but issue a warning?

tests/testthat/test-geom-sf.R
```diff
@@ -0,0 +1,14 @@
...
+  f = system.file("gpkg/nc.gpkg", package="sf")
```
hadley: `<-` 😄

Issue #2156, opened 2017-06-13
Retina option for ggsave


hadley:
It would be useful to have an option to `ggsave()` that made it easy to produce retina quality graphics. This might be as simple as supporting `dpi = "retina"`.

hadley:
Perhaps it would be best to generalise to support "print", "screen", and "retina" ?

foo-bar-baz-qux:
Sounds good to me, I had to experiment with this figure when I first started with ggplot to find a DPI I was satisfied with. This option would help save some time. I could have a go at implementing this.

I'm thinking `print` is 300 DPI/PPI and `screen` is 72 DPI/PPI per standard? 

Retina is a little trickier though since it's supposed to be a resolution at which the eye cannot distinguish the individual pixels, which is dependent on viewing distance and physical screen size. The [wiki article](https://en.wikipedia.org/wiki/Retina_Display) suggests 300 PPI was the initial figure for a handheld device, with the latest iPhones having 401 PPI (the highest on the list) and larger displays ~220 PPI. Would it make sense to go with the highest PPI (401)?

hadley:
I think just doubling the dpi is standard.

dylan-stark:
I agree with Hadley: the practical answer for retina is to double the available PPI (`dpi` for `ggsave()`). This is what's done in knitr and in [this gist](https://gist.github.com/jcheng5/452a25c32d67d52d68c0529e1c08f59f). It's also the advice in the graphic design community (ex. [here](https://www.smashingmagazine.com/2015/05/retina-design-in-photoshop/)).

One thing to consider is that some phones are at triple the pixel density (e.g., iPhone 6 Plus). Not sure if that warrants having "retina-2" and "retina-3". Maybe an extra "retina-plus" option? (Though the 8 is rumored to be at 3x, too.)

foo-bar-baz-qux:
Thanks for the links, @dylan-stark they were helpful. If I'm understanding it correctly, it seems there should be 2 changes to implement this feature then:
1. Support special dpi arguments, like `screen (= 72 or 96)`, `print (= 300)`, maybe `mdpi (= 160)` for mobile developers.
2. A `retina` argument that is either FALSE (no scaling), TRUE (2x DPI). And perhaps `retina-2`, `retina-3` for 3x and 4x DPI scaling respectively.

[...]

PR #2217, opened 2017-07-23
Add string options for `ggsave()` (#2156)

foo-bar-baz-qux:
Add feature request #2156. 

3 new string options for `ggsave()`: "screen" (72 DPI), "print" (300 DPI), and "retina" (320 DPI).

One question was what DPI the "retina" option should be. As mentioned in the issue, it's not so straightforward to pick a single DPI number. The standard for retina is to double the DPI, but from what baseline DPI?

There is no such thing as a baseline DPI for Apple because they define their `@1x` baseline, non-retina template using pixel dimensions (which all designs should be based on and then scaled for output). Retina output is `@2x` and sometimes `@3x` for the newer phones. For their phones, [this roughly corresponds to ~160 DPI](http://sebastien-gabriel.com/designers-guide-to-dpi/#ios), and ~130 DPI for their tablets. For Android, the baseline MDPI (equivalent to `@1x` for Apple) is similarly around a DPI of 160, and their [baseline calculations for density independence assume 160 DPI](https://developer.android.com/guide/practices/screens_support.html).

Accordingly, I think it's best to go with 2x this baseline of ~160 DPI, making the retina option 320 DPI.

### DIFF
--- a/R/save.r
+++ b/R/save.r
@@ -68,15 +68,18 @@ ggsave <- function(filename, plot = last_plot(),
 #' @return Parsed DPI input value
 #' @noRd
 parse_dpi <- function(dpi) {
-  if (is.character(dpi)) {
-    dpi <- switch(dpi, screen = 72, print = 300, retina = 320, NULL)
-
-    if (is.null(dpi)) {
-      stop("Invalid string DPI value", call. = FALSE)
-    }
+  if (is.character(dpi) && length(dpi) == 1) {
+    switch(dpi,
+      screen = 72,
+      print = 300,
+      retina = 320,
+      stop("Unknown DPI string", call. = FALSE)
+    )
+  } else if (is.numeric(dpi) && length(dpi) == 1) {
+    dpi
+  } else {
+    stop("DPI must be a single number or string", call. = FALSE)
   }
-
-  dpi
 }
 
 plot_dim <- function(dim = c(NA, NA), scale = 1, units = c("in", "cm", "mm"),
--- a/R/save.r
+++ b/R/save.r
@@ -14,7 +14,8 @@
 #' @param scale Multiplicative scaling factor.
 #' @param width,height,units Plot size in `units` ("in", "cm", or "mm").
 #'   If not supplied, uses the size of current graphics device.
-#' @param dpi Plot resolution. Applies only to raster output types.
+#' @param dpi Plot resolution. Also accepts a string input: "retina" (320),
+#'   "print" (300), or "screen" (72). Applies only to raster output types.
 #' @param limitsize When `TRUE` (the default), `ggsave` will not
 #'   save images larger than 50x50 inches, to prevent the common error of
 #'   specifying dimensions in pixels.
@@ -44,6 +45,7 @@ ggsave <- function(filename, plot = last_plot(),
                    width = NA, height = NA, units = c("in", "cm", "mm"),
                    dpi = 300, limitsize = TRUE, ...) {
 
+  dpi <- parse_dpi(dpi)
   dev <- plot_dev(device, filename, dpi = dpi)
   dim <- plot_dim(c(width, height), scale = scale, units = units,
     limitsize = limitsize)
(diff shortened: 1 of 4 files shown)


R/save.r
```diff
@@ -58,6 +60,25 @@ ggsave <- function(filename, plot = last_plot(),
...
+#' Parse a DPI input from the user
+#'
+#' Allows handling of special strings when user specifies a DPI like "print".
+#'
+#' @param dpi Input value from user
+#' @return Parsed DPI input value
+#' @noRd
+parse_dpi <- function(dpi) {
+  if (is.character(dpi)) {
+    dpi <- switch(dpi, screen = 72, print = 300, retina = 320, NULL)
```
hadley: Instead of NULL, you should insert the error message. i.e.

```R
dpi <- switch(dpi, 
  screen = 72, 
  print = 300, 
  retina = 320,
  stop("Unknown DPI string", call. = FALSE)
)

File: R/save.r [hunk 2]
```diff
@@ -58,6 +60,25 @@ ggsave <- function(filename, plot = last_plot(),
...
+
+    if (is.null(dpi)) {
+      stop("Invalid string DPI value", call. = FALSE)
+    }
+  }
```
hadley: You should have a branch here for `is.numeric()` that returns `dpi`, and then an `else` branch that throws an error saying DPI must be character or number.  You should also check that it's length 1.

tests/testthat/test-ggsave.R
```diff
@@ -57,3 +57,17 @@ test_that("text converted to function", {
 test_that("if device is NULL, guess from extension", {
   expect_identical(body(plot_dev(NULL, "test.png"))[[1]], quote(grDevices::png))
 })
+
+
+# parse_dpi ---------------------------------------------------------------
+
+test_that("DPI string values are parsed correctly", {
+  expect_is(parse_dpi("print"), "numeric")
```
hadley: Better to use `expect_type()` here

foo-bar-baz-qux: Makes sense, updated with those changes. Looks like `expect_type` only allows a single type so had to specify `double` (as it is on my system) rather than something like `c("integer", "double")`. I'm assuming this is fine and it won't auto-cast to integer or something like that on some systems.

R/save.r
```diff
@@ -58,6 +60,29 @@ ggsave <- function(filename, plot = last_plot(),
...
+  if (is.character(dpi) && length(dpi) == 1) {
+    dpi <- switch(dpi,
+      screen = 72,
+      print = 300,
+      retina = 320,
+      stop("Unknown DPI string", call. = FALSE)
+    )
+    return(dpi)
+  } else if (is.numeric(dpi) && length(dpi) == 1) {
+    return(dpi)
```
hadley: Can you remove the explicit `return()` please?

foo-bar-baz-qux: Done. I know assignment still silently returns a value, but left `dpi` as the last expression in the first `if` block to avoid confusion. Just looking at some neighbouring code this appears to be the style. Could be something to [add to the style guide](http://style.tidyverse.org/functions.html#return)?

R/save.r
```diff
@@ -58,6 +60,29 @@ ggsave <- function(filename, plot = last_plot(),
...
+#' Parse a DPI input from the user
+#'
+#' Allows handling of special strings when user specifies a DPI like "print".
+#'
+#' @param dpi Input value from user
+#' @return Parsed DPI input value
+#' @noRd
+parse_dpi <- function(dpi) {
+  if (is.character(dpi) && length(dpi) == 1) {
+    dpi <- switch(dpi,
```
hadley: In this case you should remove the assignment and the explicit return

foo-bar-baz-qux: Updated

hadley:
Thanks!

For future reference, don't include the issue in the title of the PR, but instead include `Closes #xyz` in the body.

PR #2235, opened 2017-08-08
Add generics for widthDetails and heightDetails

karawoo:
S3 dispatch wasn't working because there were no generics defined for `widthDetails()` and `heightDetails()`. This seems subtle enough that I didn't add it to NEWS.md, but let me know if you want me to add a note there.


karawoo:
Oh whoops, that's because they're from grid.

PR #2240, opened 2017-08-16
Additional examples aes-colour-fill-alpha

hardeepsjohar:
Adding in some examples which really helped me when I was working on filing bar graphs.


karawoo:
These examples don't quite work as written, did you mean something like `c + geom_bar(aes(fill = factor(cyl)))`?

Note also that the next set of examples in this file covers mapping factor variables to fill color:

```r
# The aesthetic fill also takes different colouring scales
# setting fill equal to a factor variable uses a discrete colour scale
k <- ggplot(mtcars, aes(factor(cyl), fill = factor(vs)))
k + geom_bar()
```

PR #2247, opened 2017-08-24
Further refactor titleGrob

karawoo:
Here are some improvements to `titleGrob()` and related functions that we talked about:

- [X] In `add_margins()` eliminate `text_` from argument names
- [X] In `ggstrip()`, remove `absoluteGrob()`
- [X] Document these functions
- [X] Fixed `xp` and `yp`. Now `title_spec()` sets them based on the range in which `angle` falls.

I still need to figure out how to prevent rotated text from extending outside of the strip in facet labels.

``` r
df <- data.frame(
  x = 1:2,
  y = 1:2,
  z = c("a", "aaaaaaabc")
)

ggplot(df, aes(x, y)) +
  geom_point() +
  facet_wrap(~ z) +
  theme(strip.text.x = element_text(hjust = 0, vjust = 1, angle = 45, debug = TRUE))
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 2.2.1.9000
 
+* Rotated strip labels now correctly understand `hjust` and `vjust` parameters
+  at all angles (@karawoo).
+
 * Strip labels now understand justification relative to the direction of the
   text, meaning that in y facets the strip text can be placed at either end of
   the strip using `hjust` (@karawoo).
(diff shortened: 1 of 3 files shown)


karawoo:
Just about the rotation fix? I don't think anything else in here is user-facing at all.

hadley:
Yeah, the rotation fix.

karawoo:
A comment in case we look at this in the future -- we decided not to prevent the text from running outside the box because the placement of the anchor point is currently correct.

PR #2248, opened 2017-08-24
Add margins on all sides of strips by default

### LINKED ISSUE #1887: Margins don't affect strip.text
ggplot version: `2.1.0.9001`
```{r }
p <- ggplot(mpg, aes(displ, cty)) + geom_point() + facet_grid(. ~ cyl)
p + theme(strip.text = element_text(margin = margin(l=5, b = -5)))
  ```
gg1

The `hjust`  hack still works:
```{r g1}
p + theme(strip.text = element_text(hjust=0.055))
```
gg2
__ref:__ #1502

karawoo:
Recent changes to text placement make it possible for strips to have margins around all sides. This adds margins around all sides of strips. Default plots where the strip labels are centered will look the same, but this fully fixes #1887 because the margins are defined in `strip.text` instead of `strip.text.x` and `strip.text.y`.

In some cases where text is rotated, this also prevents the edges of the text from being cut off the strip.

``` r
df <- data.frame(
  x = 1,
  y = 1,
  z = c("aaaaaaabc")
)

ggplot(df, aes(x, y)) +
  facet_wrap( ~ z) +
  theme(strip.text = element_text(angle = 45, hjust = 0, vjust = 0, debug = TRUE))
```

Margins will need to manually be set to 0 if users want to fully justify text with the edges of the strips.

This change added a little extra space above and below the strip label in the `theme_void` visual test plot with large text size. I've included the updated visual test case.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # ggplot2 2.2.1.9000
 
+* Strips gain margins on all sides by default. This means that to fully justify
+  text to the edge of a strip, you will need to also set the margins to 0
+  (@karawoo).
+
 * Strip labels now understand justification relative to the direction of the
   text, meaning that in y facets the strip text can be placed at either end of
   the strip using `hjust` (@karawoo).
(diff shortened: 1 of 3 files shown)


smouksassi:
While talking about strips a comment about strip.background. With the switch functionality and strip.placement options one might want to have a different fill for x versus y strips. Currently there is no strip.background.x or strip.background.y. Sorry if it the wrong place to bring this up let me want if you want a separate issue for it.

karawoo:
Yes @smouksassi, if you could open a separate issue that would be great.

PR #2261, opened 2017-09-01
Added `exclude` argument to `stat_contour()`

eliocamp:
I added an argument to `stat_contour()` to remove specific levels. The motivation behind this comes from my field (atmospheric science) in which we often plot, for example, pressure anomalies and are more interested in areas of high or low pressure and less interested on where exactly the zero line is. 
 
Example:
``` r
library(ggplot2)
field <- expand.grid(x = 0:360, y = -90:90)
field$z <- with(field, cos(x*pi/180*3) + sin(y*pi/180*4))

ggplot(field, aes(x, y, z = z)) +
  geom_contour(aes(color = ..level..))
```

``` r
ggplot(field, aes(x, y, z = z)) +
  geom_contour(aes(color = ..level..), exclude = 0) 
```

Right now this can be achieved by setting the breaks parameter explicitly, but it requires to know the range of the data and it's a lot more typing.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -407,6 +407,9 @@ There were a number of tweaks to the theme elements that control legends:
 
 * `Scale` extensions can now override the `make_title` and `make_sec_title` 
   methods to let the scale modify the axis/legend titles.
+  
+* `stat_contour` gains new argument `exclude` to remove specific levels. 
+  (@eliocamp)
 
 # ggplot2 2.1.0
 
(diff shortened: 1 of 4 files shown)


hadley:
This feels a bit too specific too me - I think a better approach would be to allow `breaks()` to be a function, so that you could then easily generalise in other ways.

eliocamp:
@hadley Yes, it would definitely be better, but it wasn't such an easy fix. Would you be open to a PR adding that functionality?

hadley:
Yes, definitely! It's only a couple of lines of code (I started doing it but realised it would be better to take a bit more time and actually document those arguments too. If you have time, some improved docs would be great, but otherwise I can add them.)

Issue #2065, opened 2017-03-04
Improve error reporting for 'Found object is not a stat'


kevinushey:
Somewhat bogus example but illustrates the issue:

```r
library(ggplot2)
ggplot(mtcars, aes(x = mpg, y = cyl)) +
  geom_bar(stat = "identify")
```

Running this gives:

```
> library(ggplot2)
> ggplot(mtcars, aes(x = mpg, y = cyl)) +
+   geom_bar(stat = "identify")
Error: Found object is not a stat.
```

Note that the issue is I've accidentally typed `"identify"` instead of `"identity"`. Any chance `ggplot2` could better report 'typo' errors like this? A suggestion:

```
Error in geom_bar(stat = "identify"):
  object "identify" (discovered in "base" namespace) is not a stat
```

"did you mean..." reporting here would be awesome, but likely expensive to implement.

PR #2262, opened 2017-09-04
Improve error message from find_subclass() (#2065).

mcol:
The check for null was occurring on the wrong variable: 'name' cannot be
null, but 'obj' can. Moreover the message should report the variable as
set by the end user (such as 'identity'), rather than the internal name
constructed by the function (such as 'StatIdentity').


hadley:
Thanks!

Issue #2264, opened 2017-09-13
Adding message for overwriting a coordinate system


Ax3man:
When one supplies multiple scales for the same aesthetic, a message is printed that notifies the user. When multiple coordinate systems are supplied, however, the last one is used without notification. This is sometimes confusion for new users, that e.g. expect `coord_trans` to work with other `coord_` functions. Could a message be added?

This gives a nice message:
```r
qplot(1:2, 1:2) + ylim(0, 2) + ylim(0, 3)
# Scale for 'y' is already present. Adding another scale for 'y', which will replace the existing scale.
```
But this doesn't:
```r
qplot(1:2) + coord_trans(y = 'log10') + coord_polar()
```
Unfortunately, it's not obvious to me where a replacement could be detected, as there is no `coord_list` like there is `scales_list`. I could prepare a PR, but I'd need some pointers.

PR #2265, opened 2017-09-14
Simplistic fix for #2264

Ax3man:
This will now give a message whenever a second (or more) coordinate system is added. To avoid a message when adding one coordinate system that effectively overrides the default `coord_cartesian`, I gave the default one a `default` attribute. `add_ggplot` checks for that attribute.

If there is a preferred method of checking for the default coordinate system, I can change this.

Running e.g.

```r
qplot(1:2, 1:2) + coord_trans(y = 'log10') + coord_polar()
```

now prints: 

```r
Coordinate system already present. Adding new coordinate system, which will replace the existing one.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # ggplot2 2.2.1.9000
 
-<<<<<<< HEAD
+* Replacing a coordinate system (other than the default) will now give a message
+  (@Axeman, #2264).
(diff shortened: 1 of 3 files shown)


hadley:
Thanks for the code, but I'd like to think this through a bit more before considering specific implementations.

PR #2286, opened 2017-10-09
Fix bug when setting strips to element_blank

thomasp85:
As described here: https://stackoverflow.com/questions/46347656/ggplot2-error-when-setting-facet-strip-to-element-blank


hadley:
Thanks!

PR #2292, opened 2017-10-13
Added support for univariate box plots with no `x` aesthetic.

### LINKED ISSUE #2110: Automatically fill in x for univariate boxplot
i.e. make this work

```R
ggplot(mtcars, aes(y = disp)) + geom_boxplot()
```

foo-bar-baz-qux:
Fixes #2110.

This now works:
```r
ggplot(mtcars, aes(y = disp)) + geom_boxplot()
```
image

Which would essentially be the equivalent of this, without the mapped x-aesthetic:
```r
library(ggplot2)
ggplot(mtcars, aes(x=0, y = disp)) + geom_boxplot()
```

Tried a few approaches, but this is the cleanest way I've found so far. I think it looks OK to have the `x` location be at 0 without an axis label; user can always map the x aesthetic if more fine-grained adjustments are required. Ideally, I think the univariate boxplot would look something like this, where the x-axis is essentially removed/treated like a discrete scale. However, I couldn't find a clean way of changing it from a continuous scale from within the geom/stat ggprotos without hard coding the exceptions into the layer building process.
``` r
library(ggplot2)
ggplot(mtcars, aes(y = disp, x = "all")) + geom_boxplot()
```


hadley:
Could you please add a test for this too? (Maybe a visual test)

hadley:
Are you interested in finishing this off? No problems if you don't have time; just let me know I'll get it over the finish line.

foo-bar-baz-qux:
I'm haven't looked at the visual testing suite so might take some extra time to look through; perhaps you could add that test and finish this off? Thanks.

hadley:
Looking at the code again, it feels like a visual test would be overkill.

Issue #2231, opened 2017-08-03
Documentation for `geom_label`


jdagilliland:
The documentation for `geom_label` on [tidyverse.org](http://ggplot2.tidyverse.org/reference/geom_text.html#-code-geom-label-code-) has untranslated formatting in a section heading that looks odd. It seems to come from [here](https://github.com/tidyverse/ggplot2/blob/53a22cd04e03b6249f3809c5732d9510d16036ca/R/geom-text.r#L22), where backticks for code formatting are used within a section heading.
It has also created a weird look [here](https://www.rdocumentation.org/packages/ggplot2/versions/2.2.1/topics/geom_label)

I would have fixed it myself and made a PR, but I wasn't sure what would be the best way to handle it.

hadley:
Moved to pkgdown, which is the source of the problem.

PR #2297, opened 2017-10-15
Update geom-text.r

landesbergn:
Removing code formatting which is causing grossness here: http://ggplot2.tidyverse.org/reference/geom_text.html#-code-geom-label-code-

Addressing issue: https://github.com/tidyverse/ggplot2/issues/2231


hadley:
The correct fix is to make sure the html conversion is correct in pkgdown. I've filed an issue at https://github.com/hadley/pkgdown/issues/437

Issue #2296, opened 2017-10-15
Error in documentation for scale_*_*. Default name is not NULL.


SebastianHenz:
The documentation of for example `scale_x_continuous `or `scale_x_date `state that the default value for the name parameter is NULL, when it is in fact `waiver()`. This is also the case in most other scale functions but I didn't check all of them.

The text in the arguments section says:
name: The name of the scale. Used as axis or legend title. If NULL, the default, the name of the scale is taken from the first mapping used for that aesthetic.

But in the usage section it says:
`scale_x_continuous(name = waiver(), ...`

PR #2304, opened 2017-10-20
Correct documentation for default name parameter in scales.

gfiumara:
This corrects the documentation issue raised in #2296.


hadley:
Would you mind also including: "If `NULL`, the legend title will be omitted`"?

hadley:
Thanks!

PR #2305, opened 2017-10-22
Rework of '.element_tree' to an R6 Class, permitting user-defined new theme elements.

nhamilton1980:
Dear Hadley / Winston & Other Devs,

This is not a huge update, but it is meaningful in my opinion.

I have completely re-worked the element tree, now based of R6 Class, which permits user-defined theme elements, for use in package extentions.

Checks have been implemented so that these user-defined theme elements DO NOT CLASH with the set of protected theme elements, that is to say, those that are otherwise required for the base ggplot2 functionality. What this boils down to, is that the base ggplot2 element tree structure is protected, whilst still permitting users to extend this structure.

With this update, people will be able to create their own extensions to ggplot2, and design their own theme elements for whatever reason. I have added this functionality, primarily so I can re-work some of the painful aspects of the ggtern package, which depends on dozens and dozens of new theme elements to function correctly.

Previously, the .element_tree object was hidden and not-exported. There was no way to extend this tree, which was presumably to stop people meddling around with something that is so fundamental, with potentially catastrophic consequences to the plot routines. This is a totally valid, and understanable concern, however, by integrating a 'fixed' and 'variable' tree in the proposed R6 class that has been included within this update the following outcomes have resulted:

    1. The now publicly-visible tree (accessed via get_elements() function) is the combination of these fixed and variable components. If the both the fixed and variable trees are 'sane', then the combination will also be sane. Sanity checks are conducted when the user attempts to add a new theme element.
    2. Given the above, the base tree can be protected and preserved, innoculated from the variable tree. Any proposed new elements (ie those that the user wants to add to the 'variable' tree) can be checked against the 'fixed' tree first, to make sure such elements are not reserved.
    3. New elements can be added easily via the exported add_element(k,v) function, where k is the 'key' (non-clashing name of the desired element to be added to the tree) and 'v' is the value (instance of the el_def(...)) theme element definition.
    4. Ultimately, from the perspective of the end user nothing has changed, essentially only one tree exists, none of the previous functionality has been altered, the same methods of indexing (ie .element_tree[[elname]]) behave as previously used within ggplot2, however. This change will only be appreciated by package developers like myself.
    5. The guts of the validate_element routine has been incorporated within the R6 class.

Anyway, this update will enable ggtern to make a number of changes, reducing package fragility and complexity. I am sure other package developers will benefit too.

Any questions, please email me.

Regards,

Nicholas E. Hamilton
UNSW Sydney.
n.hamilton@unsw.edu.au


R/theme-elements.r
```diff
@@ -244,144 +244,277 @@ element_grob.element_line <- function(element, x = 0:1, y = 0:1,
...
+#'  or the reserved "character", which means a character vector (not
+#'  "character" class)
+#' @param inherit A vector of strings, naming the elements that this
+#'  element inherits from.
+#' @export
 el_def <- function(class = NULL, inherit = NULL, description = NULL) {
-  list(class = class, inherit = inherit, description = description)
+  structure(
+    list(class = class, inherit = inherit, description = description),
+    class=c('el_def','list')
```
nhamilton1980: This is makes it possible to enforce the expected arguments when using the add_element(...) function in the R6 class below.

R/theme-elements.r
```diff
@@ -244,144 +244,277 @@ element_grob.element_line <- function(element, x = 0:1, y = 0:1,
...
+      strip.switch.pad.wrap = el_def("unit"),
+
+      plot.background       = el_def("element_rect", "rect"),
+      plot.title            = el_def("element_text", "title"),
+      plot.subtitle         = el_def("element_text", "title"),
+      plot.caption          = el_def("element_text", "title"),
+      plot.margin           = el_def("margin"),
+
+      aspect.ratio          = el_def("character")
+    ),
```
nhamilton1980: varElementsFixedDefault: 
This is the default tree structure under ggplot2. Previously this was .element_tree

nhamilton1980:
This is a little disappointing, however, as suggested I have started an issue here: https://github.com/tidyverse/ggplot2/issues/2540

PR #2309, opened 2017-10-24
Allow addition of custom objects

thomasp85:
As discussed. This PR adds the possibility of developers writing custom behaviour for adding other objects than those defined by ggplot2 to a gg object.

It works by falling back to the `add_to_ggplot()` generic in `add_ggplot()`. A potential change to the PR would be to write `add_to_ggplot()` methods for all the build in classes as well thus completely removing the if else check from `add_ggplot()`


R/plot-construction.r
```diff
@@ -60,46 +60,88 @@ add_ggplot <- function(p, object, objectname) {
   if (is.null(object)) return(p)
```
hadley: This can now be a method of the generic

R/plot-construction.r
```diff
@@ -60,46 +60,88 @@ add_ggplot <- function(p, object, objectname) {
...
+  set_last_plot(p)
+  p
+}
+#' Add custom objects to ggplot
+#'
+#' This generic allows you to add your own methods for adding custom objects to
+#' a ggplot with [+.gg].
+#'
+#' @param object An object to add to the plot
+#' @param p The ggplot object to add `object` to
```
hadley: I think `plot` and `object_name` would be better argument names

tests/testthat/test-plot-construction.R
```diff
@@ -0,0 +1,16 @@
...
+test_that("Custom object cannot be added without generic", {
+  p <- ggplot()
+  custom_object <- structure(list(), class = 'test_object')
+  expect_error(p + custom_object, "Don't know how to add custom_object to a plot")
+})
+
+ggplot_add.test_object <- function(object, p, objectname) {
+  10
+}
+test_that("Methods can be defined for adding custom objects", {
```
hadley: You can define the method inside the test, just to make things slightly cleaner.

R/plot-construction.r
thomasp85: Oh - Didn’t knew you could dispatch on `NULL`

tests/testthat/test-plot-construction.R
thomasp85: For some reason I’m having problems getting dispatch to work correctly in the test. Moving it out fixed it locally but it still throws an error on travis..?

thomasp85: Do you have any idea why methods defined inside test_that expressions are not being dispatched to?

hadley: Yes, it's something to do with the weird way in which S3 dispatch happens. I forget all the details, but just move it outside for now.

hadley:
Does `R CMD check` pass locally?

thomasp85:
I have some strange (`libc++abi.dylib: __cxa_guard_acquire detected deadlock`) errors in unrelated tests that I'm currently troubleshooting

thomasp85:
OK - I'm up to speed again.

No It doesn't pass locally. The test method won't get dispatched to when called as part of Check, but works fine when pasted into the console. I guess the problem is the same - that S3 methods are not properly registered when defined within tests...

No idea how to properly test this then...

hadley:
I think you can just drop the test - if the S3 dispatch was failing, every plot would break.

thomasp85:
I never thought I should live to hear you ask me to remove a test - this will haunt you 😛

hadley:
Thanks!

Issue #1954, opened 2016-12-13
ggplot2 pipe support proof of concept (with full backward-compatibility)


DarwinAwardWinner:
**Edit: Look at the [comment below](https://github.com/tidyverse/ggplot2/issues/1954#issuecomment-266836115) for a better implementation.**

After reading somewhere that you (Hadley) wished you could have designed ggplot2 using the magrittr pipe instead of the plus sign, I wondered whether it was possible to add support for the pipe without breaking old code. After a good bit of tinkering, I came up with a solution that seems to work pretty well.

```R
library(ggplot2)
library(dplyr)
library(magrittr)
library(lazyeval)
library(assertthat)

pipify_gg <- function(ggfun) {
    function(gg, ...) {
        dots <- list(...)
        if (missing(gg)) {
            return(do.call(ggfun, dots))
        } else if (is(gg, "gg")) {
            return(gg + do.call(ggfun, dots))
        } else {
            dots <- c(list(gg), dots)
            return(do.call(ggfun, dots))
        }
    }
}

## Standard evaluation functions just need to be wrapped in this
## function.
p_aes_string <- pipify_gg(aes_string)
p_geom_point <- pipify_gg(geom_point)
p_geom_smooth <- pipify_gg(geom_smooth)

## NSE functions need something fancier. (Are there any other NSE
## functions in ggplot2 besides aes?)
p_aes <- function(...) {
    dots <- lazy_dots(...)
    first_unnamed <- match("", names(dots))
    gg_evaled <- NULL
    if (!is.na(first_unnamed)) {
        gg_evaled <- try(lazy_eval(dots[[first_unnamed]]), silent=TRUE)
    }
    if (is(gg_evaled, "gg")) {
        dots[[first_unnamed]] <- NULL
    }
    aes <- dots %>% lapply(. %$% expr %>% deparse) %>% do.call(what=aes_string)
    aes <- ggplot2:::rename_aes(aes)
    if (is(gg_evaled, "gg")) {
        return(gg_evaled + aes)
    } else {
        return(aes)
    }
}

## Make this a function to get a new random set of points for each
## plot. This makes it easier to see that each invocation of ggplot is
## actually creating and printing a new plot.
mydata <- function () data_frame(x=rnorm(50), y=x+rnorm(50, sd=0.2))

## Now let's test it out

## Regular ggplot2
ggplot(mydata()) +
    aes(x=x, y=y) +
    geom_point() +
    geom_smooth()

[...]

Issue #1382, opened 2015-10-21
Simplifying of %>% chaining and qplot()


elmar-hinz:
Hello,

I am new with R programming, having no overview what has been discussed, is in the pipe and the spirit of the community. The combination of **dplyr**, **chaining** and **qplot** is perfect for quickly evaluating data like this:

`iris %>% filter(Species == "versicolor") %>% qplot(Petal.Length, Sepal.Length, data = .)`

The user experience would almost be perfect, if I could write it as sparse as this:

`iris > filter(Species == "versicolor") > plot(Petal.Length, Sepal.Length)`

This ticket is about the `qplot()` part only. I like to suggest a function `dplot()`, a modification of `qplot()` to spare the data parameter, by putting it in first position like in `ggplot()`.

`iris %>% filter(Species == "versicolor") %>% dplot(Petal.Length, Sepal.Length)`

The **d** would serve as memo, that it is an adapter for the *dplyr* chain as well as data is first.

If this is welcome, it would be a chance for me, to learn the contributing cycle of R and ggplot in special.

Regards

Elmar

hadley:
Hi Elmar,

I think the request is reasonable, but I'm unlikely to accept a patch because I'm continuing to downplay the importance of `qplot()`, and I don't plan to make any further changes to it.

Hadley

elmar-hinz:
Thank you Hadley for your immediate response. 

I understand your reasoning. The importance of having a clean grammar is of more importance, than having the shortest language. You did so much good with this, that I don't want to question it.

Elmar

Issue #1370, opened 2015-10-14
Pipe operator into ggsave and argument order


andygarcia:
`ggsave` currently takes the `plot` argument as the second argument of the function.

I would like to be able to use the `%>%` forward pipe operator to move from a dplyr+ggplot chain into ggsave as opposed to creating a temporary named object or just let the plot render in the background to be accessed via lastplot() (note this creates a Rplots.pdf) if run via RScript which is an annoyance)

Ideally the plot argument would be first, the `file` second to make use of the pipe

hadley:
I considered doing this, but ultimately decided it was likely to break too much existing code. I wish we could though :/

andygarcia:
Makes sense, perhaps a simple wrapper function will be my solution.

BrianDiggs:
If you specify the file name in the `ggsave` call as a named argument, you don't even need a wrapper function. 

```r
library("ggplot2")
library("dplyr")

(ggplot(mtcars, aes(wt, mpg)) + geom_point()) %>%
    ggsave(filename="tmp.pdf")
```
This takes advantage of how R assigns passed parameters to formal parameters. Named parameters are assigned first, and then unnamed ones are assigned in order to what is left. Since you have already specified `filename`, the next available parameter is `plot`.

This won't work if you want to use the default naming algorithm, though.

PR #2326, opened 2017-11-05
ggplot with the pipe

zeehio:
### Motivation

This PR proposes a backwards-compatible change to ggplot2 in order to provide compatibility with the pipe. ggplot2 is both a very popular package and part of the tidyverse. Having a different "pipe symbol" in ggplot2 is surprising for newcomers, especially students.

#### Previous attempts

- #1954 was rejected, but it is not clear to me if the rejection was because it required of a custom pipe operator, or because of other reasons. 
- #1382 proposed a change for better `qplot` integration with the pipe. It was rejected because of an interest on downplaying the `qplot` function.
- #1370 aimed to change the `ggsave` arguments to make it more pipeable, but that change was rejected because it had chances of breaking existing code.

#### Why is this PR different?

It uses the conventional `magrittr` pipe. It aims to fix the plot constructor functions (`geom_`, `scale_`, `theme_`, ...) and it is fully backwards compatible.

#### What does this PR affect?

As a proof of concept I have just changed `geom_point` and `annotate`.

### How is it implemented?

I have used this approach to successfully convert my rather humble [`condformat`](https://github.com/zeehio/condformat) package, that had a similar `ggplot`-like syntax, to a tidy evaluation syntax.

For each function that needs to be adapted, follow the `geom_point` example:
1. The original `geom_point` function is renamed to `old_geom_point` (not exported).
2. A `new_geom_point` function is created, that uses an initial `gplt` argument (function not exported)
3. A `geom_point` function is created (exported and documented), using the arguments of the `new_geom_point` function. It just calls a shared helper function that dispatches the call to the `new_` or the `old_` versions, depending on whether or not the `gplt` argument (or the first argument) are `ggplot`  objects.
4. Adapt examples and tests.

Once the helper dispatcher function is written (and it already is in this PR), the rest is easy to address.

### Reasons for rejection

- Fear of breaking existing code. Mitigation: By only using the new syntax when a `ggplot` object is passed we can ensure that backwards compatibility is preserved. I can run revdep checks if required.
- More code paths to maintain. If both the "pipe" and "+" APIs are kept, there are more code paths to test. Mitigation: The use of helpers keeps boilerplate code to a minimum. A soft-deprecation path can be set, first by showing the pipe syntax in the documentation and cheatsheets, later by using warnings, then forcing the use of a "options" to keep it working, and finally removing it.
- The Pull Request is incomplete. Mitigation: If there is a chance for the PR to be merged I will work on a complete PR.

### Reasons for acceptance

- Just look at this beauty: `mtcars %>% ggplot(aes(mpg, disp)) %>% geom_point()`
- It increases the tidyverse consistency
- It makes pipe explanations straightforward to newcomers

### Final comment

While there are pros and cons to this approach and many users would love to have the tidyverse consistency, we can all understand the trade-off between stability and code evolution. Base R is very conservative while the tidyverse has already done some breaking changes in the past. Either this is done now or it will never be done in the `ggplot2` package. I am perfectly fine with your decision anyway.

Thanks for your time!


has2k1:
This is a big change. It changes how to think about operations on ggplot objects.

For example, this

```R
p <- ggplot(mtcars, aes(x = wt, y = mpg)) %>% geom_point()
```
is not descriptive of what is happening, i.e adding a layer of points.

`+` in ggplot is not a pipe symbol, it is an addition. It is used to add layers and to add attributes to a ggplot object.

> **Reasons for acceptance**
> - It increases the tidyverse consistency

I am not sure about this. It does not fall under any of the categories under magrittr [usage](http://magrittr.tidyverse.org/#usage). I think this creates a new category for the `%>%` symbol as a general purpose *chaining* operator.

zeehio:
I understand it can be a controversial change. The original [ggplot](https://github.com/hadley/ggplot1) package did use functions in a pipe-compatible way (even though the pipe did not exist back then). Because there was no pipe, using ggplot was very uncomfortable, and Hadley decided to reimplement it using the addition operator. If the pipe had existed when `ggplot2` was created we would all be using the pipe now in ggplot2 ([see here](https://www.reddit.com/r/dataisbeautiful/comments/3mp9r7/im_hadley_wickham_chief_scientist_at_rstudio_and/cvi19ly/))

This line:

    p <- ggplot(mtcars, aes(x = wt, y = mpg)) + geom_point()

can be read as *"`geom_point` adds a layer to a plot object"*.

This line:

    p <- mtcars %>% ggplot() %>% geom_point(aes(x = wt, y = mpg))

can be read as *"`mtcars` is given to the `ggplot()` canvas and the canvas is modified through `geom_point()`"*. We modify a plot by passing it through functions (that add geometries, configure scales, change aspect...).

I don't see a significant difference between changing a `tibble` object with `dplyr` or changing a `ggplot` object with `ggplot2`, but I guess this can be a matter of opinion.

Just for the sake of trying (and in case this PR was rejected) I wrote a pipeable ggplot2 package named [`ggpipe`](https://github.com/zeehio/ggpipe). If you did not like the use of the pipe in the previous example, maybe you like the next example even less :sweat_smile::

```
# devtools::install_github("zeehio/ggpipe")
library("ggpipe")
message("Files iris_petal.png and iris_sepal.png will be created")
iris %>%
  ggplot() %>% # create a canvas on top of iris
  geom_point(aes(x = Sepal.Length, y = Sepal.Width, color = Species)) %>%
  ggsave("iris_petal.png", height = 3, width = 5, dpi = 72) %>%
  unggplot() %>% # gets the data back
  ggplot() %>% # create another canvas
  geom_point(aes(x = Petal.Length, y = Petal.Width, color = Species)) %>%
  ggsave("iris_petal.png", height = 3, width = 5, dpi = 72) %>%
  unggplot() %>%
  head()
```

zeehio:
The `ggpipe` package I mentioned above is passing all the tests I could adapt from the `ggplot2` package.

By doing the implementation in a different package I get all the "pros" and none of the "cons" written in this pull request, so if there is no interest in adopting the pipe here then feel then free to close the pull request, I promise I won't push for it :smiley: .

hadley:
There are two questions to consider:
  
* Is it worth it?
* How can we cleanly implement it?

These are somewhat linked because if the implementation isn't clean and maintainable then there's no point in considering the first point. 

Unfortunately your implementation is problematic because:

* It makes extensive use of non-standard evaluation, but it does not use tidy
  evaluation, which means it's likely to be buggy.
  
* It uses `tryCatch()` and double-evaluates the first argument which means
  it's likely to have have some performance impact.
  
I think the implementation could be improved but I don't see any fundamental way to work around that the number of arguments in plus style is one less than the number of arguments in pipe style; i.e. all the names will be one off in one of the implementation. Given the number of people that use ggplot2, there are likely to be many unusual combinations of name and position, and this seems likely to break existing code.

However, even if you could come up with a fix for this issue, I'm still not sure it would be a good idea because now there would be two ways to do everything in ggplot2, which I think would create a lot of confusion. There would also be a lot of work to rewrite all the examples, and I guess the ggplot2 book would need another update to?

Issue #2307, opened 2017-10-22
Feature request: allow function as argument value in `scale_*(limits=)`


econandrew:
Request is to allow `limits` arg in e.g. `scale_x_continuous` to accept a function which takes the default limits and returns new limits, in parallel with the same functionality in `breaks` and `labels`.

At present it seems like the two options are (1) specify literal limits, which is inflexible (2) calculate the preferred limits outside of the ggplot statement, but that is messy and can result in redoing calculations already happening inside ggplot (e.g. for position="stack").

It looks like this may be a one line change in scale-.r's `get_limits`, but I don't know my way around well enough to be sure.

Motivating example follows.

``` r
library(ggplot2)


p <- ggplot(mtcars) + geom_point(aes(cyl,mpg))

p # But suppose we don't like the default y-axis breaks...
```

![](https://i.imgur.com/ADmOIcM.png)

``` r

# ...we have a general mechanism in breaks= to modify them with a function...
breaker <- function(by) {
  function(limits) {
    low <- floor(limits[1]/by)*by
    high <- ceiling(limits[2]/by)*by
    seq(low, high, by)
  }
}

# ...but it doesn't override the limits so we get an (IMHO) ugly y axis where
# the data exceeds the range of the tick marks...
p + scale_y_continuous(breaks = breaker(10))
```

![](https://i.imgur.com/RYICyku.png)

``` r

# In principle we can function-ise the limits in the same way
limiter <- function(by) {
  function(limits) {
    low <- floor(limits[1]/by)*by
    high <- ceiling(limits[2]/by)*by
    c(low, high)
  }
}

# But for now the best we can do is manually override the limits
p + scale_y_continuous(breaks = breaker(10), limits = limiter(10)(range(mtcars$mpg)))
```

![](https://i.imgur.com/EpPQvLF.png)

``` r

# Feature request: if limits arg is a function, pass it the existing limits automatically
p + scale_y_continuous(breaks = breaker(10), limits = limiter(10))
#> Warning in is.na(self$limits): is.na() applied to non-(list or vector) of
#> type 'closure'
#> Error in rep(yes, length.out = length(ans)): attempt to replicate an object of type 'closure'
```

has2k1:
Is using the `expand` parameter not sufficient?

econandrew:
I don't think so. Following on the example above, you might want a y-scale that ends exactly at multiples of 10, e.g.

[...]

PR #2334, opened 2017-11-11
Allow functional limits in continuous scales

econandrew:
An implementation of the idea in #2307, for consideration.

Motivation is to allow programmatic, data-agnostic control of scale limits, taking advantage of rather than having to replicate ggplot's aesthetic calculations. For example, suppose you want limits to always be a multiple of 50:

``` r
library(ggplot2)
library(dplyr)

# Example 1: stacked position adjustment

p1 <- ggplot(mpg, aes(class)) + geom_bar(aes(fill = drv))

# define a function to determine limits
limiter <- function(by) {
  function(limits) {
    low <- floor(limits[1]/by)*by
    high <- ceiling(limits[2]/by)*by
    c(low, high)
  }
}

# with PR, you can write something like...
p1 + scale_y_continuous(limits = limiter(50))

# ...current alternative would have to be
themax <- mpg %>% group_by(class) %>% summarise(count = n()) %>% pull(count) %>% max
p1 + scale_y_continuous(limits = limiter(50)(c(0, themax)))
```


R/scale-.r
```diff
@@ -107,13 +107,17 @@ Scale <- ggproto("Scale", NULL,
...
 
+  #  if scale contains a function, apply it to the default (inverted) scale range
   #  if scale contains a NULL, use the default scale range
   #  if scale contains a NA, use the default range for that axis, otherwise
   #  use the user defined limit for that axis
   get_limits = function(self) {
     if (self$is_empty()) return(c(0, 1))
 
-    if (!is.null(self$limits)) {
+    if (is.function(self$limits)) {
```
hadley: Since we're adding another condition here, can you please reframe the if statement along these lines:

```R
if (is.null()) {

} else if (is.function()) {

} else if (is.numeric()) {
  
} else {
  stop("Informative error message", call. = FALSE) 
}
```

R/scale-.r
```diff
@@ -526,8 +530,11 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
 #'   - A character vector giving labels (must be same length as `breaks`)
 #'   - A function that takes the breaks as input and returns labels
 #'     as output
-#' @param limits A numeric vector of length two providing limits of the scale.
-#'   Use `NA` to refer to the existing minimum or maximum.
+#' @param limits One of:
+#'   - A numeric vector of length two providing limits of the scale.
```
hadley: Or `NULL`?

R/scale-.r
```diff
@@ -107,13 +107,17 @@ Scale <- ggproto("Scale", NULL,
     stop("Not implemented", call. = FALSE)
   },
 
+  #  if scale contains a function, apply it to the default (inverted) scale range
   #  if scale contains a NULL, use the default scale range
```
hadley: "contains" -> "is"

dpseidel:
Hey @econandrew  if you're still interested in this PR, I'm around (:wave: summer intern!) and can help you get this over the finish line if you'd like. Let me know if I can help in anyway.

thomasp85:
@dpseidel would you consider finishing this off?

dpseidel:
Yep happy to.

R/scale-.r
dpseidel: I did rework this conditional to be more intuitive but since `get_limits()` is called by both discrete and continuous scales and can be passed character, numeric, POSIXt and now function classes, I ultimately did not add an explicit `is.numeric` conditional nor an error message. Happy to update further if necessary.

hadley:
The diff seems to include spurious changes which makes it hard to review.

dpseidel:
Sorry, yes, I got a little nit picky on the NEWS doc, clearly that should be in a different PR. All cleaned up now.

PR #2346, opened 2017-11-21
Fix colorbar to show all colors of the scale

### LINKED ISSUE #2343: Colourbar does not have all colors of the scale.
For some data it is possible for the colourbar to cover a fraction of the colourspace smaller than that covered by the data. It can be confusing when the guide for a continuous scale does span the full *mapped* space of an aesthetic.

Actually, this bug is probably triggered every time there is a colourbar, but in most cases it is small and imperceptible. With colourmaps like `viridis`, the cases where it is obvious should only go up.

In the reprex below, I had to do some contriving (`nbin=7`) to reveal it. 

``` r
library(ggplot2)

ggplot(faithful, aes(x = eruptions, y = waiting)) +
  geom_point()  +
  stat_density_2d(aes(fill=..level..), geom='polygon') +
  xlim(0.5, 6) +
  ylim(40, 110) +
  scale_fill_viridis_c() +
  guides(fill=guide_colorbar(nbin=7))
```

The root of the issue is `pretty` does not guarantee that the breaks it computes are close enough to the limits that the difference is not perceptible when mapped to a colour. 

A simple but partial solution is to increase the default `guide$nbin` parameter. This would give `pretty` enough wiggle room to work with and reduce incidences of the bug.

A better solution is recognising that the "raster" colourbar does not need to be generated using 'pretty' breaks, a `seq(.limits[1], .limits[2], length=guide$nbin)` can do. For this solution the current value of `nbin` need not change except for the craziest of colourmaps.

has2k1:
Previously it showed an approximation where the colours
at the scale limits were not guaranteed to be included.

``` r
library(ggplot2)

ggplot(faithful, aes(x = eruptions, y = waiting)) +
  geom_point()  +
  stat_density_2d(aes(fill=..level..), geom='polygon') +
  xlim(0.5, 6) +
  ylim(40, 110) +
  scale_fill_viridis_c() +
  guides(fill=guide_colorbar(nbin=7))
```

*Before*

fixes #2343

### DIFF
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -190,7 +190,7 @@ guide_train.colorbar <- function(guide, scale) {
 
   # bar specification (number of divs etc)
   .limits <- scale$get_limits()
-  .bar <- discard(pretty(.limits, n = guide$nbin), scale$get_limits())
+  .bar <- seq(.limits[1], .limits[2], length = guide$nbin)
   if (length(.bar) == 0) {
     .bar = unique(.limits)
   }
--- a/NEWS.md
+++ b/NEWS.md
@@ -71,6 +71,8 @@
 
 * Fix bug in `scale_*_gradient2()` where points outside limits can sometimes reappear due to rescaling. Now, any rescaling is performed after the limits are enforced (@foo-bar-baz-qux, #2230).
 
+* The colorbar shows all colours of the scale (@has2k1, 2343).
+
 ### Margins
 
 * Strips gain margins on all sides by default. This means that to fully justify


R/guide-colorbar.r
```diff
@@ -190,7 +190,7 @@ guide_train.colorbar <- function(guide, scale) {
 
   # bar specification (number of divs etc)
   .limits <- scale$get_limits()
-  .bar <- discard(pretty(.limits, n = guide$nbin), scale$get_limits())
+  .bar <- seq(.limits[1], .limits[2], length=guide$nbin)
```
hadley: Could you please put spaces around `=`?

has2k1: Ahh. python instincts.

hadley:
Thanks!

PR #2365, opened 2017-12-09
Creates an "hvh" direction in geom_step

rmatkins:
Changes the stairstep function to create the option for an "hvh" direction. In this case, the "step" (or vertical deflection) occurs halfway between point A and B.

`# Example with random distance between each x 
`
`
x <- runif(50)
`
`
y <- runif(50)
`
`
ggplot(NULL,aes(x,y)) + geom_point()+ geom_step(direction = "hvh")
`
randomx

`# Example with uniform distance between each x`
`x <- 1:50`
`y <- runif(50)`
`ggplot(NULL,aes(x,y)) + geom_point()+ geom_step(direction = "hvh")`
`
uniformx


R/geom-path.r
```diff
@@ -302,7 +302,7 @@ GeomStep <- ggproto("GeomStep", GeomPath,
 #
 # @keyword internal
 stairstep <- function(data, direction="hv") {
-  direction <- match.arg(direction, c("hv", "vh"))
+  direction <- match.arg(direction, c("hv", "vh", "hvh"))
```
hadley: Maybe "mid" ? I think "hvh" implies that we might also add "vhv"

R/geom-path.r
```diff
@@ -319,9 +319,19 @@ stairstep <- function(data, direction="hv") {
     xs <- c(1, rep(2:n, each = 2))
   }
 
-  data.frame(
-    x = data$x[xs],
-    y = data$y[ys],
-    data[xs, setdiff(names(data), c("x", "y"))]
-  )
+  if (direction == "hvh") {
+    gaps <- data$x[2:n] - data$x[1:(n-1)]
```
hadley: I think this would be simpler as

```R
 gaps <- data$x[-1] - data$x[-n]
```

And might even better to divide by two here.

R/geom-path.r
```diff
@@ -319,9 +319,19 @@ stairstep <- function(data, direction="hv") {
...
+
+    data.frame(
+      x = c(data$x[1],data$x[xs[-1]] - (gaps[rep(1:(n-1), each = 2)]/2),data$x[n]),
```
hadley: Can you pull `rep(1:(n-1), each = 2)` out into a variable?

R/geom-path.r
```diff
@@ -319,9 +319,19 @@ stairstep <- function(data, direction="hv") {
...
+      y = c(data$y[ys],data$y[n]),
+      data[c(xs,n), setdiff(names(data), c("x", "y"))]
+    )
+  } else {
+    data.frame(
```
hadley: I think it would be better to create `x` and `y` inside each branch of the if statement, and then create the `data.frame()` afterwards

R/geom-path.r
rmatkins: "mid" makes sense, "vhv" wouldn't. I have made the change.

R/geom-path.r
rmatkins: Good point. Instead of dividing by 2 here, I created a new variable that creates a vector of the midpoints by adding gaps/2 to the prior data (excepting the final datapoint).
```
gaps <- data$x[-1] - data$x[-n] 
mid_x <- data$x[-n] + gaps/2
```

R/geom-path.r
rmatkins: I reformatted this to be more consistent with the current "xs", "ys" variables and added another else if statement to the loop that creates the "xs"/"ys"

R/geom-path.r
rmatkins: Done

R/geom-path.r
```diff
@@ -314,14 +315,27 @@ stairstep <- function(data, direction="hv") {
...
 
-  data.frame(
-    x = data$x[xs],
-    y = data$y[ys],
-    data[xs, setdiff(names(data), c("x", "y"))]
-  )
+  if (direction == "mid") {
+    gaps <- data$x[-1] - data$x[-n]
+    mid_x <- data$x[-n] + gaps/2 # map the mid-point between adjacent x-values
+    x <- c(data$x[1],mid_x[xs],data$x[n])
```
thomasp85: can you add spaces around arguments in line with the tidyverse coding style

thomasp85:
```suggestion
    x <- c(data$x[1], mid_x[xs], data$x[n])
```

R/geom-path.r
```diff
@@ -314,14 +315,27 @@ stairstep <- function(data, direction="hv") {
...
+    y <- c(data$y[ys])
+    data_attr <- data[c(1,xs,n), setdiff(names(data), c("x", "y"))]
+  } else {
+    x <- data$x[xs]
+    y <- data$y[ys]
+    data_attr <- data[xs, setdiff(names(data), c("x", "y"))]
+  }
+
+  data.frame(x,y,data_attr)
```
thomasp85:
```suggestion
  data_frame(x, y, data_attr)
```

mastropi:
I solved the problem generating the error in the `geom_step()` example above and created PR#3435:
https://github.com/tidyverse/ggplot2/pull/3435

rmatkins:
Going to close this PR (resolution created in PR #3435 ). Thanks @mastropi!

PR #2367, opened 2017-12-11
Fix breaks for bin & bin2d stats for trans scales

### LINKED ISSUE #2366: stat_bin2d error with text geom when using transformed scales
I am encountering an issue with label placement with the stat_bin2d function when using a transformed axis such as `log` or `sqrt`. Essentially the labels are misplaced or missing in the latter case, as shown below

This gives a correct placement of labels showing the number of data points in each quadrant:

```
df <- data.frame(rnorm(1000) + 5, rnorm(1000) + 5)
colnames(df) <- c('x', 'y')

ggplot(df) +
  aes(x = x, y = y) +
  geom_point() +
  geom_vline(xintercept = 5) + geom_hline(yintercept = 5) +
  stat_bin2d(geom = 'label', 
             aes(label = ..count..), 
             breaks = list(x = c(1, 5, 20), y = c(1, 5, 20)), 
             fill = 'white')
```
image

However the addition of `scale_x_log10()` causes all of the labels to be lost. Depending on the exact axes, sometimes only some labels are lost. The numbers may also change indicating that the bins are incorrect.
image

has2k1:
The issue only was only about `stat_bin_2d`, but `stat_bin`
suffered the same slight.

Fixes #2366

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -207,6 +207,9 @@ up correct aspect ratio, and draws a graticule.
 
 * Restored correct scaling for computed variable `ndensity` when binning (@timgoodman, #2324)
 
+* `stat_bin()` and `stat_bin_2d()` now properly handle the `breaks` parameter when
+  the scales are transformed (@has2k1, #2366).
+
 ### Coords
 
 * Like scales, coordinate systems now give you a message when you're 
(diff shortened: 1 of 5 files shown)


hadley:
Thanks!

Issue #2345, opened 2017-11-18
Better support for custom color and fill aesthetics?


clauswilke:
Hello,

I'm wondering whether it would be possible to better support custom color and fill aesthetics. What I mean are cases where maybe we want two separate color mappings in the same geom, e.g. `line_color` and `point_color`. With the current ggplot2, if I create such aesthetics, I can't take advantage of any of the existing color scales. But with a few small changes, I think this would be possible.

First, to demonstrate that this is a real situation, some example code from ggridges:
```r
cols = c("#E69F00", "#56B4E9", "#009E73")
cols_dark = c('#C18511', '#289ACF', '#0D8561')
ggplot(iris, aes(x=Sepal.Length, y=Species, fill = Species)) +
  geom_density_ridges(aes(point_color = Species), alpha = .2, jittered_points = TRUE) +
  scale_fill_manual(values = cols, guide = "none") + scale_y_discrete(expand = c(0.01, 0)) +
  ggplot2:::manual_scale("point_color", values = cols_dark, guide = "none") +
  theme_ridges(grid = FALSE, center = TRUE)
```
<img width="699" alt="screen shot 2017-11-18 at 3 20 39 pm" src="https://user-images.githubusercontent.com/4210929/32985579-1f43796c-cc84-11e7-8432-f40c59bbf9a6.png">

In this example, everything needs to be drawn at once in one geom, because the filled areas and solid lines can overlap the points. But, I want to be able to specify line colors and point colors separately, and the point colors should also be different from the fill colors. The only way I see to solve the problem is to add a `point_color` aesthetic.

My code example brings up one immediate and easy to fix issue: ggplot2 doesn't export `manual_scale`, but it's what is needed here to make this example work without too much effort. Second, it would be helpful if the various existing color scales took an optional `aesthetics` argument that can be used to override the aesthetics in the function name. While this may sound a bit strange, the resulting code would be quite legible and look reasonable. For example, using the default hue color scale for fill and also applying it to the points might look like this:

```r
ggplot(iris, aes(x=Sepal.Length, y=Species, fill = Species)) +
  geom_density_ridges(aes(point_color = Species), alpha = .2, jittered_points = TRUE) +
  scale_color_hue(aesthetics = "point_color") + scale_y_discrete(expand = c(0.01, 0)) +
  theme_ridges(grid = FALSE, center = TRUE)
```

So, in summary, I'm making the following requests:
1. Export `manual_scale`
[...]

PR #2375, opened 2017-12-19
added fill to gpar in geom_curve()

hrbrmstr:
Based on https://stackoverflow.com/questions/47878787/fill-arrow-on-geom-curve-ggplot2/47879776#47879776

this adds a missing `fill` `gpar` (i.e. back on par with `geom_segment`) and takes care of ^^ issue.

### DIFF
--- a/R/geom-curve.r
+++ b/R/geom-curve.r
@@ -53,6 +53,7 @@ GeomCurve <- ggproto("GeomCurve", GeomSegment,
       square = FALSE, squareShape = 1, inflect = FALSE, open = TRUE,
       gp = gpar(
         col = alpha(trans$colour, trans$alpha),
+        fill = alpha(trans$colour, trans$alpha),
         lwd = trans$size * .pt,
         lty = trans$linetype,
         lineend = lineend),


R/geom-curve.r
```diff
@@ -38,6 +38,7 @@ geom_curve <- function(mapping = NULL, data = NULL,
 #' @usage NULL
 #' @export
 GeomCurve <- ggproto("GeomCurve", GeomSegment,
+  default_aes = aes(colour = "black", fill = "black", size = 0.5, linetype = 1, alpha = NA),
```
hadley: I think you can omit this since it will be inherited from `geom_segment()`

hadley:
Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`.

I think this is a slight change to behaviour? i.e. if you had arrows and mapped the colour to a variable, the arrows will now be black instead of coloured?

clauswilke:
It is a change in behavior, and on further reflection making the arrow fill an aesthetic may cause problems for some.

First, under the old behavior `geom_curve()` would always fill the arrow head in white, regardless of any color or fill settings. That's clearly wrong.

Second, the question is whether the arrow head should take the color of `colour` or the color of `fill`. If people currently draw filled arrows specified by `colour` and use `fill` for some other geom then they may get unexpected results. It might be better to use a non-standard aesthetic, such as `arrow_fill`, which by default mirrors `colour`. This then links back to the needs explained in issue #2345, but the change wouldn't cause any noticeable regressions.

Unless @hrbrmstr beats me to it, I can address this towards the end of the week. Not today, unfortunately.

hadley:
For now, I'd prefer to make this a parameter of the geom, and then we can figure out how to make it an aesthetic post-2345.

clauswilke:
So, it'd be a parameter that is set to `NA` or `NULL` by default, in which case the geom uses the `colour` aesthetic to fill the arrow, but if it's set to a color then that's used to override the fill color with that specific value? This sounds reasonable to me.

hadley:
Exactly - you can just copy what `geom_boxplot()` does with `outlier.colour` etc

hrbrmstr:
Aye. I'll make all those changes later today and also mod the docs + NWS

hrbrmstr:
```
set.seed(123)
data <- data_frame(x = rnorm(10), y = rnorm(10))

gg <- ggplot(data, aes(x, y)) + geom_point()
```

```
gg + geom_curve(aes(x = 0, y = 0, xend = 1, yend = 1),
                color = "black",
                arrow = arrow(type = "closed"))
```
image
```
gg + geom_curve(aes(x = 0, y = 0, xend = 1, yend = 1),
                color = "black", arrow.fill = "red",
                arrow = arrow(type = "closed"))
```
image
```
gg + geom_segment(aes(x = 0, y = 0, xend = 1, yend = 1),
                  color = "black",
                  arrow = arrow(type = "closed"))
```
image
```
gg + geom_segment(aes(x = 0, y = 0, xend = 1, yend = 1),
                  color = "black", arrow.fill = "red",
                  arrow = arrow(type = "closed"))
```
image

and, for good measure

```
gg + geom_curve(aes(x = 0, y = 0, xend = 1, yend = 1),
                color = "#00000000", arrow.fill = "black",
                arrow = arrow(type = "closed"))
```
image
```
gg + geom_segment(aes(x = 0, y = 0, xend = 1, yend = 1),
                  color = "#00000000", arrow.fill = "red",
                  arrow = arrow(type = "closed"))
```
image

hrbrmstr:
Roxygen docs were also updated as well as the `NEWS.md`

NEWS.md
```diff
@@ -48,6 +48,10 @@
 * Added `stat_qq_line()` to make it easy to add a simple line to a Q-Q plot. This
   line makes it easier to judge the fit of the theoretical distribution 
   (@nicksolomon).
+  
+* Added `arrow.fill` parameter to `geom_segment()` and `geom_curve()` to
```
hadley: Can you please rewrite so the function names come first? (And then use that to alphabetise within this list of changes?)

(Normally I do that just before release but this release has been dragging on because I haven't found the time to finish it off)

hrbrmstr: Done.

PR #2232, opened 2017-08-07
Refactor titleGrob() to eliminate the need for stripGrob()

karawoo:
As part of our overhaul of text placement, this PR:

* Splits `titleGrob()` into  `title_spec()` (which returns the text grob (and debugging grobs if requested), text height, and text width) and `add_margins()`, which should add the margins around the text grob but is clearly not working in strips, but does work for axis titles. `titleGrob()` now calls both `title_spec()` and `add_margins()`.
* Renames the `expand_x`/`expand_y` arguments to `margin_x`/`margin_y`.
* Removes `stripGrob()`. `build_strip()` and `ggstrip()` now call `title_spec()` and `add_margins()`.

``` r
df <- data.frame(
  x = 1:3,
  y = 1:3,
  z = c("a", "aaab", "aaaaaaabc"),
  g = c("b", "bbbb", "bbbbbbbcd")
)

base <- ggplot(df, aes(x, y)) + 
  geom_point() + 
  theme(strip.text.y = element_text(angle = 0))

base + facet_grid(g ~ z) +
  theme(
    strip.text.x = element_text(hjust = 0, debug = TRUE),
    strip.text.y = element_text(hjust = 0, debug = TRUE),
    axis.title.x = element_text(margin = margin(t = 30), debug = TRUE)
  )
```

PR #2379, opened 2017-12-22
Fix angle of switched y strips

### LINKED ISSUE #2356: Inverted y-axis strip text angle with facet_grid(...switch = "both")
I'm running the latest dev version of ggplot2 (2.2.1.9000). Going straight from the [docs](http://ggplot2.tidyverse.org/reference/facet_grid.html):

```r
data <- transform(mtcars,
  am = factor(am, levels = 0:1, c("Automatic", "Manual")),
  gear = factor(gear, levels = 3:5, labels = c("Three", "Four", "Five"))
)
p <- ggplot(data, aes(mpg, disp)) + geom_point()
p + facet_grid(am ~ gear, switch = "both")
```
This gives me the following, where you can see that the y-axis facet strip text ("Automatic" and "Manual") is inverted compared to the documented example.

inverted_strip_text

Of course, an easy manual fix  is to simply add ` + theme(strip.text.y = element_text(angle = 90))` to the above. I imagine that the previous behaviour is still the desired default, though.

karawoo:
Fixes #2356. The [titleGrob() refactor](https://github.com/tidyverse/ggplot2/pull/2232/) caused a bug where strip labels on the left side weren't rotated properly. This fixes it by extracting the creation of the text grobs into a separate function (`create_strip_labels()`) that can be called after the angle has been adjusted.

### DIFF
--- a/R/labeller.r
+++ b/R/labeller.r
@@ -552,8 +552,9 @@ build_strip <- function(label_df, labeller, theme, horizontal) {
       clip = "on"
     )
 
-    if (inherits(theme$strip.text.y, "element_text")) {
-      theme$strip.text.y$angle <- adjust_angle(theme$strip.text.y$angle)
+    # Change angle of strip labels for y strips that are placed on the left side
+    if (inherits(element, "element_text")) {
+      element$angle <- adjust_angle(element$angle)
     }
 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks @karawoo !

PR #2382, opened 2017-12-28
fix #2260 for continuous color scales

### LINKED ISSUE #2260: legend.spacing.y error "only one operand may be a unit"
This problem was first reported to us at stan-dev/bayesplot#106 for a user using a dev version of ggplot2 (tidyverse/ggplot2@c592e32). I troubleshooted the problem and came up with the following minimal example.

```r
library(ggplot2)
ggplot(iris) + 
  aes(Sepal.Length, Sepal.Width, color = Species) + 
  geom_point() +
  theme(legend.spacing.y = unit(1, "cm"))
#> Error in Ops.unit(theme$legend.spacing.y %||% 0.5, unit(title_height,  : 
#>  only one operand may be a unit
```

Adding ` theme(legend.spacing.x = unit(1, "cm"))` to the plot instead does not yield an error though.

clauswilke:
This pull request fixes #2260 for continuous color scales. #2260 has been previously fixed for discrete color scales, but it still exists for continuous ones:

```
library(ggplot2)
ggplot(iris, aes(x=Sepal.Length, y=Sepal.Width, color=Petal.Width)) + 
  geom_point() +
  theme(legend.spacing.y = grid::unit(0.5, "cm"))

#  Error in Ops.unit(theme$legend.spacing.y %||% 0.5, unit(title_height,  : 
#   only one operand may be a unit 
```

### DIFF
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -299,8 +299,8 @@ guide_gengrob.colorbar <- function(guide, theme) {
 
   # gap between keys etc
   hgap <- width_cm(theme$legend.spacing.x  %||% unit(0.3, "line"))
-  vgap <- height_cm(theme$legend.spacing.y %||% 0.5 * unit(title_height, "cm"))
-  
+  vgap <- height_cm(theme$legend.spacing.y %||% (0.5 * unit(title_height, "cm")))
+
   # label
   label.theme <- guide$label.theme %||% calc_element("legend.text", theme)
   grob.label <- {


statwonk:
Just bumped into this in the wild, thank you for the fix @clauswilke!

clauswilke:
@statwonk I'm glad you found this helpful. What I haven't addressed yet, however, is that this fix reveals two downstream bugs that will likely prevent you from taking full advantage of this fix by itself:

- The code incorrectly uses the vertical gap where it should use the horizontal gap, e.g. [here.](https://github.com/tidyverse/ggplot2/blob/3fb3fb00555bab5b270b3467a1ddd90341a62886/R/guide-colorbar.r#L378) This means that if you're changing the vertical legend spacing you're also moving the legend labels further away from the color bar.

- The code converts units into cm and then uses the values as if they were mm, see e.g. [here](https://github.com/tidyverse/ggplot2/blob/3fb3fb00555bab5b270b3467a1ddd90341a62886/R/guide-colorbar.r#L302) and then [here](https://github.com/tidyverse/ggplot2/blob/3fb3fb00555bab5b270b3467a1ddd90341a62886/R/guide-colorbar.r#L434). If you want a properly working colorbar guide, you'll have to fix those issues as well.

The required changes for these two issues are more involved and will change the `vdiffr` test cases, so I wanted to get the obvious fix in first. (And with the current version of `ggplot2`, I cannot even demonstrate these two issues, because ggplot crashes before we even get there.)

clauswilke:
I've filed #2397 and #2398. I realized that I could demonstrate both issues with the current ggplot2 code base.

hadley:
Thanks!

PR #2386, opened 2018-01-06
Make position_dodge2 preserve total width by default

### LINKED ISSUE #2290: Boxplot is too narrow when used with facetting
Boxplot is too narrow when used with faceting, and this issue never appears in 2.2.1 release.
I encountered this issue with a dataset similar to below:
```
va <- c(rnorm(10, 1, 1))
vb <- c(rnorm(10, 3, 1))
vc <- c(rnorm(10, 2, 1))
vd <- c(rnorm(10, 3, 1))

tf <- data.frame(
                 factor1 = rep(c('A', 'B'), each = 20,times=4),
                 factor2 = rep(c('A', 'B', 'A', 'B'), each = 10, times = 4),
                 factor3 = rep(c(LETTERS[1:4]), each = 40),
                 value = rep(c(va, vb, vc, vd), times = 4)
                 )

```
The plot below is with a discrete x-axis and a continuous y-axis:
```
ggplot(tf, aes(factor1, value)) +
    geom_boxplot(aes(fill = factor2)) +
    facet_wrap(~factor3, ncol = 2) +
    theme(aspect.ratio = 2 / 3)
```
And the plot produced:
image
The width of each boxplot is too narrow. This also happens when I use `facet_grid()`:
```
ggplot(tf, aes(factor1, value)) +
    geom_boxplot(aes(fill = factor2)) +
    facet_grid(.~factor3) +
    theme(aspect.ratio = 2 / 3)
```

image

If I remove `facet_wrap()`, the plot looks good:
```
ggplot(tf, aes(factor1, value)) +
    geom_boxplot(aes(fill = factor2)) +
    theme(aspect.ratio = 2 / 3)
```
image
If I increase `width`, the boxes are off the ticks:
```
ggplot(tf, aes(factor1, value)) +
    geom_boxplot(aes(fill = factor2), width = 2) +
    facet_wrap(~factor3, ncol = 2) +
    theme(aspect.ratio = 2 / 3)
```
image

I don't know if it is a bug, or something internal changed from the 2.2.1 release. My original code works fine on release version, but has this issue on the dev version.

karawoo:
Closes #2290 by setting the default `preserve` argument in `position_dodge2()` to `"total"`. 

Before: 

``` r
ggplot(mtcars, aes(x = factor(vs), y = mpg)) +
  facet_wrap( ~ factor(cyl)) +
  geom_boxplot()
```

After:

``` r
ggplot(mtcars, aes(x = factor(vs), y = mpg)) +
  facet_wrap( ~ factor(cyl)) +
  geom_boxplot()
```

### DIFF
--- a/R/geom-boxplot.r
+++ b/R/geom-boxplot.r
@@ -111,7 +111,9 @@ geom_boxplot <- function(mapping = NULL, data = NULL,
                          inherit.aes = TRUE) {
 
   # varwidth = TRUE is not compatible with preserve = "total"
-  if (!is.character(position)) {
+  if (is.character(position)) {
+    if (varwidth == TRUE) position <- position_dodge2(preserve = "single")
+  } else {
     if (identical(position$preserve, "total") & varwidth == TRUE) {
       warning("Can't preserve total widths when varwidth = TRUE.", call. = FALSE)
       position$preserve <- "single"
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -19,7 +19,7 @@ position_dodge2 <- function(width = NULL, preserve = c("single", "total"),
 #' @usage NULL
 #' @export
 PositionDodge2 <- ggproto("PositionDodge2", PositionDodge,
-  preserve = "single",
+  preserve = "total",
   padding = 0.1,
   reverse = FALSE,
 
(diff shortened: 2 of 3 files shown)


hadley:
Do you remember why we set it to single?

karawoo:
It was because `position_dodge2()` is primarily used used for box plots, which can have variable widths but that only works when `preserve = "single"`. So the reasoning was to make the default behavior of `position_dodge2()` compatible with `geom_boxplot(varwidth = TRUE)`. But this causes the boxes to be too narrow in facetted plots.

I just pushed an update to make sure that `preserve` gets set back to `"single"` whenever `varwidth = TRUE`. We warn the user about the change when they explicitly asked for `preserve = "total"`.

hadley:
Thanks!

PR #2387, opened 2018-01-07
Move mgcv from Suggests to Imports

HughParsonage:
Since `stat_smooth` calls `mgcv` unconditionally.  (https://github.com/tidyverse/ggplot2/blob/6968e36b8e2df37db6c6d71e110b88a4cbe82962/R/stat-smooth.r)

This change will help packages which use `stat_smooth` in vignettes avoid CRAN warnings.

Since `mgcv` is 'recommended', requiring it rather than suggesting it will have practically no adverse impact in terms of additional dependencies. (I wasn't sure if it was worth an entry on NEWS but have included one for completeness.)

Thank you!


hadley:
Thanks - I removed the news bullet because it caused a conflict and it wasn't really needed since it's not very user facing.

PR #2405, opened 2018-01-17
Add tag label for adding identifier to plot

thomasp85:
As per discussions on twitter, this PR adds a `tag` label alongside `title`, `subtitle`, and `caption`. It is placed in the upper left corner in its own row and column between the margin and the title/x-label.

The tag name is chosen over `label` as the `labs()` function indicate that “label” is a catchall term for all types of titles.

The default theming is like this:

```r
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test')
```

tag

The size is slightly larger than the title and the margin is chosen so that it is well separated from the main parts of the plot. In general this is based on my own taste and is up for discussion.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -80,7 +80,8 @@
 * Added `tag` label for adding identification tags to the plot. A tag is added 
   with the `labs()` function and styling is handled through the `plot.tag` theme 
   element. Position is specified with the `plot.tag.position` theme setting and
-  defauls to `"topleft"` (@thomasp85).
+  defauls to `"topleft"`. Tags are useful for identifying subplots in a 
+  multiplot figure and often used in the scientific literature (@thomasp85).
 
 ### Scales
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -77,9 +77,10 @@
   line makes it easier to judge the fit of the theoretical distribution 
   (@nicksolomon).
   
-* Added `tag` label for adding identification tags to the upper left corner of 
-  the plot. A tag is added with the `labs()` function and styling is handled 
-  through the `plot.tag` theme element (@thomasp85).
+* Added `tag` label for adding identification tags to the plot. A tag is added 
+  with the `labs()` function and styling is handled through the `plot.tag` theme 
+  element. Position is specified with the `plot.tag.position` theme setting and
+  defauls to `"topleft"` (@thomasp85).
 
 ### Scales
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -76,6 +76,10 @@
 * Added `stat_qq_line()` to make it easy to add a simple line to a Q-Q plot. This
   line makes it easier to judge the fit of the theoretical distribution 
   (@nicksolomon).
+  
+* Added `tag` label for adding identification tags to the upper left corner of 
+  the plot. A tag is added with the `labs()` function and styling is handled 
+  through the `plot.tag` theme element (@thomasp85).
 
 ### Scales
 
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -279,9 +279,12 @@ ggplot_gtable.ggplot_built <- function(data) {
   if (tag_pos == "manual") {
     xpos <- theme$plot.tag.position[1]
     ypos <- theme$plot.tag.position[2]
-
-    tag <- editGrob(tag, vp = viewport(x = xpos, y = ypos))
-    plot_table <- gtable_add_grob(plot_table, tag, name = "tag", t = 1,
+    tag_parent  <- gTree(
+      children = gList(tag),
+      vp = viewport(x = xpos, y = ypos, width = tag_width, height = tag_height,
+                    just = c(theme$plot.tag$hjust, theme$plot.tag$vjust))
+    )
+    plot_table <- gtable_add_grob(plot_table, tag_parent, name = "tag", t = 1,
                                   b = nrow(plot_table), l = 1,
                                   r = ncol(plot_table), clip = "off")
   } else {
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -262,14 +262,70 @@ ggplot_gtable.ggplot_built <- function(data) {
   plot_table <- gtable_add_grob(plot_table, title, name = "title",
     t = 1, b = 1, l = min(pans$l), r = max(pans$r), clip = "off")
 
-  plot_table <- gtable_add_rows(plot_table, tag_height, pos = 0)
-  plot_table <- gtable_add_cols(plot_table, tag_width, pos = 0)
-  plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
-    t = 1, b = 1, l = 1, r = 1, clip = "off")
-
   plot_table <- gtable_add_rows(plot_table, caption_height, pos = -1)
   plot_table <- gtable_add_grob(plot_table, caption, name = "caption",
     t = -1, b = -1, l = min(pans$l), r = max(pans$r), clip = "off")
+
+  plot_table <- gtable_add_rows(plot_table, unit(0, 'pt'), pos = 0)
+  plot_table <- gtable_add_cols(plot_table, unit(0, 'pt'), pos = 0)
+  plot_table <- gtable_add_rows(plot_table, unit(0, 'pt'), pos = -1)
+  plot_table <- gtable_add_cols(plot_table, unit(0, 'pt'), pos = -1)
+
+  tag_pos <- theme$plot.tag.position
+  if (length(tag_pos) == 2) tag_pos <- "manual"
+  stopifnot(tag_pos %in% c("topleft", "top", "topright", "left", "right",
+                           "bottomleft", "bottom", "bottomright", "manual"))
+
+  if (tag_pos == "manual") {
+    xpos <- theme$plot.tag.position[1]
+    ypos <- theme$plot.tag.position[2]
+
+    tag <- editGrob(tag, vp = viewport(x = xpos, y = ypos))
+    plot_table <- gtable_add_grob(plot_table, tag, name = "tag", t = 1,
+                                  b = nrow(plot_table), l = 1,
+                                  r = ncol(plot_table), clip = "off")
+  } else {
+    if (tag_pos == "topleft") {
+      plot_table$widths[1] <- tag_width
+      plot_table$heights[1] <- tag_height
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = 1, l = 1, clip = "off")
+    } else if (tag_pos == "top") {
+      plot_table$heights[1] <- tag_height
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = 1, l = 1, r = ncol(plot_table),
+                                    clip = "off")
+    } else if (tag_pos == "topright") {
+      plot_table$widths[ncol(plot_table)] <- tag_width
+      plot_table$heights[1] <- tag_height
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = 1, l = ncol(plot_table), clip = "off")
+    } else if (tag_pos == "left") {
+      plot_table$widths[1] <- tag_width
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = 1, b = nrow(plot_table), l = 1,
+                                    clip = "off")
+    } else if (tag_pos == "right") {
+      plot_table$widths[ncol(plot_table)] <- tag_width
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = 1, b = nrow(plot_table), l = ncol(plot_table),
+                                    clip = "off")
+    } else if (tag_pos == "bottomleft") {
+      plot_table$widths[1] <- tag_width
+      plot_table$heights[nrow(plot_table)] <- tag_height
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = nrow(plot_table), l = 1, clip = "off")
+    } else if (tag_pos == "bottom") {
+      plot_table$heights[nrow(plot_table)] <- tag_height
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = nrow(plot_table), l = 1, r = ncol(plot_table), clip = "off")
+    } else if (tag_pos == "bottomright") {
+      plot_table$widths[ncol(plot_table)] <- tag_width
+      plot_table$heights[nrow(plot_table)] <- tag_height
+      plot_table <- gtable_add_grob(plot_table, tag, name = "tag",
+                                    t = nrow(plot_table), l = ncol(plot_table), clip = "off")
+    }
+  }
 
   # Margins
   plot_table <- gtable_add_rows(plot_table, theme$plot.margin[1], pos = 0)
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -291,6 +291,8 @@ ggplot_gtable.ggplot_built <- function(data) {
                                   b = nrow(plot_table), l = 1,
                                   r = ncol(plot_table), clip = "off")
   } else {
+    # Widths and heights are reassembled below instead of assigning into them
+    # in order to avoid bug in grid 3.2 and below.
     if (tag_pos == "topleft") {
       plot_table$widths <- unit.c(tag_width, plot_table$widths[-1])
       plot_table$heights <- unit.c(tag_height, plot_table$heights[-1])
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -273,8 +273,12 @@ ggplot_gtable.ggplot_built <- function(data) {
 
   tag_pos <- theme$plot.tag.position
   if (length(tag_pos) == 2) tag_pos <- "manual"
-  stopifnot(tag_pos %in% c("topleft", "top", "topright", "left", "right",
-                           "bottomleft", "bottom", "bottomright", "manual"))
+  valid_pos <- c("topleft", "top", "topright", "left", "right", "bottomleft",
+                 "bottom", "bottomright")
+  if (!(tag_pos == "manual" || tag_pos %in% valid_pos)) {
+    stop("plot.tag.position should be a coordinate or one of ",
+         paste(valid_pos, collapse = ', '), call. = FALSE)
+  }
 
   if (tag_pos == "manual") {
     xpos <- theme$plot.tag.position[1]
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -292,42 +292,42 @@ ggplot_gtable.ggplot_built <- function(data) {
                                   r = ncol(plot_table), clip = "off")
   } else {
     if (tag_pos == "topleft") {
-      plot_table$widths[1] <- tag_width
-      plot_table$heights[1] <- tag_height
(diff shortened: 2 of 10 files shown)


clauswilke:
Thomas,

thanks for spearheading this.  I see label positioning similar to legend positioning: There could be a position option in the theme that can either be a string such as `"top-left"`, `"bottom-left"`, etc, or a vector that takes coordinates. In the first case (string), you'd make a separate cell in the gtable. In the second case (coordinates), you'd just place the label wherever the coordinates request, allowing for placement anywhere in the entire plot area.

hadley:
What if the label is longer? Does it overlap with the plot area, or does the margin expand?

clauswilke:
Hadley,

My personal opinion is that labels should generally be pasted onto the plot without expanding any margins. That almost always produces the best visual results. However, the API I proposed, which follows the `legend.position` example, allows for both approaches. Expand margins if a named position is provided and don't expand margins if coordinates are provided.

Best,
  Claus

thomasp85:
I just had a similar discussion with @baptiste on twitter where I ended up proposing something similar to @clauswilke. My personal feeling is that the default should never allow overlapping of plot elements

thomasp85:
I’ll update this pr to include a more flexible positioning and we can take it from there

baptiste:
oops sorry was on my phone and limited to twitter. I agree with the general consensus; my personal view is that

- 'tags' are 1 or 2-letter max (otherwise it's really a title in disguise), possibly with an extra character or two: little risk of terrible collisions
-  tags within the plot panel can be useful but are already straight-forward with annotate('label', x=-Inf) and co.
- whenever I use tags in practice, I really want a tight fit because of space constraints; the unused top-left cell is typically a good spot already. If they were to add an extra cell unconditionally I'd probably never use them. 
- aligning tags across multiple plots on a device shouldn't be too much of an issue regardless of the extra cell, provided the justification is smart. If the tag is placed to span the whole gtable, it would need [top-left justification to stick there when panels are aligned](https://gist.github.com/baptiste/f06578674f46d98ee35bb529dd22d954), whereas with the extra cell strategy the justification would happen within that cell and could be more flexibly tuned.

thomasp85:
I've added a `plot.tag.position` that behaves quite like `legend.position` (except it can target corners as well)

If you want a tight fit then provide coordinates, if you want automatic expansion to make room, use a string.

```r
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test') +
  theme(plot.tag.position = "topleft") # Default
```

tag3

compared to using coordinates

```r
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test') +
  theme(plot.tag.position = c(0, 1))
```

tag2

thomasp85:
Does the current implementation/features satisfy all parties?

baptiste:
Looks good to me but I don't have a computer with me. I'd suggest
eyeballing many visual test cases, with different themes and font sizes
(24pt theme) as it's usually where things go wrong with overlap etc.

clauswilke:
I think this is all fine.

Do I understand correctly that this code should produce a tag that is left-aligned with the y axis title?

```{r}
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A') +
  theme(plot.tag.position = c(0, 1),
        plot.tag = element_text(margin = margin(l = 5.5), hjust = 0))
```

bhive01:
The only thing I'm wondering about is tag length. As was mentioned on Twitter, these things should be short. Are you going to restrict their length at all? I didn't see anything in the pull (may have missed it?).

clauswilke:
I've played around with the code a bit, and I can't say that I find the behavior of `hjust`, `vjust`, and `margin` intuitive.

1. Setting `hjust=0` makes the tag disappear:
```
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test') +
  theme(plot.tag.position = c(0, 1),
        plot.tag = element_text(hjust = 0))
```
screen shot 2018-01-19 at 10 22 17 pm

2. Setting `hjust = 1` places the tag in the middle of the plot:

```
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test') +
  theme(plot.tag.position = c(0, 1),
        plot.tag = element_text(hjust = 1))
```
screen shot 2018-01-19 at 10 22 26 pm

(The same is seen for `vjust = 0` and `vjust = 1` in the vertical direction.)

3. Setting the top or left margin doesn't seem to have any effect:

```
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test') +
  theme(plot.tag.position = c(0, 1),
        plot.tag = element_text(margin = margin(t = 10, l = 10)))
```

screen shot 2018-01-19 at 10 22 32 pm

My expectation would be that `hjust` and `vjust` move the label relative to the tag position, as in (for example) `geom_text()` or `geom_label()`, and that margin would place a margin around the text (similar to `label.padding` in `geom_label()`).

thomasp85:
Yeah, I’ve stumbled onto that as well - I think the grob needs to be wrapped inside another gtree but I won’t get to it before monday

thomasp85:
@bhive01 I have no intention of limiting the number of characters in the tag - I cannot see any compelling reason to put an arbitrary limit on it...

@clauswilke I've played around with making justification work along with margins to no avail - I'll try some more but there're some grid interplay that complicates it...

thomasp85:
@clauswilke I have fixed the margin/justification interplay so that it now works as I would expect - can you test and see if it is sensible to you..?

thomasp85:
@Hadley do you think this PR will get merged into the next release (if up to standard)? It's a rather fundamental addition in the 11th hour, but `patchwork` would like it :-)

hadley:
Yes - I now thing implementing tidy evaluation isn't going to be that hard, so I'm going to hold off the release until that's in.

clauswilke:
I took a look at it. I think it's mostly working as expected but there's one remaining issue. The coordinate system doesn't seem to cover the entire plot area but instead the area minus the margin. I think this results in somewhat unintuitive behavior. First, with a tag position of `(0, 1)`, the tag doesn't end up in the top left corner if we set `hjust = 0` and `vjust = 1` as one might expect:

```r
p <- ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'A', title = 'test') +
  theme(plot.tag.position = c(0, 1))

p + theme(plot.tag = element_text(hjust = 0, vjust = 1))
p + theme(plot.tag = element_text(hjust = 0.5, vjust = 0.5))
p + theme(plot.tag = element_text(hjust = 1, vjust = 0))
```
screen shot 2018-01-22 at 4 16 26 pm
screen shot 2018-01-22 at 4 16 43 pm
screen shot 2018-01-22 at 4 16 59 pm

Second, while the middle result (with `hjust = 0.5, vjust = 0.5`) seems to look fine in this case, it actually behaves strangely when we use a longer tag:
```r
ggplot(mtcars) +
  geom_point(aes(disp, mpg)) + 
  labs(tag = 'ABCDE', title = 'test') +
  theme(plot.tag.position = c(0, 1),
              plot.tag = element_text(hjust = 0.5, vjust = 0.5)         
```
screen shot 2018-01-22 at 4 26 52 pm

I'd rather be able to actually use the true top left corner of the plot as reference point and use the margins of the `plot.tag` element to match any plot margins if necessary. But this might be a matter of opinion. It's probably workable either way.

thomasp85:
Omitting the margin was intentional, but an easy change. The philosophy being that the margin is an actual empty space around the plot and if you want plot elements to extend to the edge you set the margin to zero.

I’m not going to be a user of freely positioned tags so I can be convinced to change it by a majority vote - what are the thoughts of @hadley, @baptiste, and @bhive01?

clauswilke:
I thought about it some more. I think it's fine as is. Placing the tag with `plot.tag.position = c(0, 1), plot.tag = element_text(hjust = 0, vjust = 1)` looks exactly right with my themes. I think the idea of having a true margin makes sense. And if somebody truly wants the label further to the plot boundary they can specify a position outside the 0--1 range.

bhive01:
I'm with @clauswilke on this. I finally got a chance to play with it (thanks dev_mode()!) and I understand what the position parts mean. All of the pieces including position and hjust/njust  work how I expect so I'm 👍. Good inclusion for the needed multi-plot labeling problem.

baptiste:
Returning to a computer and playing with this last night I realised I didn't understand the design of titleGrob (namely the interaction of margins and just – I would have shifted the whole label+ margins block). With these existing conventions the tag implementation seems good to me (though there is I believe some slight inconsistency in handling justification between manual position vs added cells that may cause some surprises).

I can see why the plot margin might be seen as an extra margin around everything, but on the other hand it could be convenient to have a way to add something (not necessarily a tag) over the entire plot in npc, for instance a watermark/copyright/draft notice, or banners à la fivethirtyeight, etc. Perhaps it'd be best implemented as a separate item such as  `labs(annotation = element_custom(grob = textGrob('DRAFT', rot=45, cex=10)))`. 

I'm also still wondering if it could make sense at some stage to let `element_zzz()` (and, ideally, other components) override their default position; the tag implementation would then default to t=0, l=0, etc.

Anyway, this is just a digression. I haven't had a chance to try it but I would still strongly recommend a comprehensive visual test in many configurations (font size, layout) and various themes. I remember being quite surprised at [how bad some things looked](https://gist.github.com/baptiste/8db48ede1eacb22cdd78093f715cb945) with a large font size (as used in presentations/posters).

karawoo:
(ignore that milestone, a keyboard cat was involved 😹)

thomasp85:
@hadley are you needing anything more for this PR (except time to review it)

clauswilke:
Since this hasn't been merged yet, I wanted to propose one very minor modification. I think the default size of the tags should be `rel(1.2)` rather than `rel(1.3)` as it currently is (see [here](https://github.com/tidyverse/ggplot2/pull/2405/files#diff-9029508150cd2b9e8ac2a30f5e5b799a)). I'm proposing `rel(1.2)` because that is the size of the font used for the plot title. I think it's good typographic practice to try to limit the number of different font sizes used, and the current themes already have the problem (in my opinion) that they use too many different sizes (`rel(0.8)`, `rel(0.9)`, `rel(1.0)`, `rel(1.2)`). Let's not make this problem worse. In fact, I have a pull request pending that cleans this up a bit and uses only three default font sizes, see [here.](https://github.com/tidyverse/ggplot2/pull/2471/files#diff-9029508150cd2b9e8ac2a30f5e5b799a)

baptiste:
I agree, the font sizes are already confusing; I never realised a `theme_bw(base_size=12)` would lead to some of the text having larger font size such as `rel(1.2)`. I'm currently answering a very unhappy referee who rated our manuscript *of the lowest scholarly presentation*; one complaint is

> the panel names (a), (b) are extremely large and it is unusual to have titles above panels

(said names are set in 13pt, though the one-column draft preview blew up the whole figure).

thomasp85:
I'm pretty neutral to this, so if that is the consensus I'll gladly change it

hadley:
Can you please merge/rebase and I'll give it a reivew

thomasp85:
It seems the CI errors are not related to any changes in this PR - do you need anything more @hadley? I’ll prioritise any issues on this

NEWS.md
```diff
@@ -76,6 +76,11 @@
 * Added `stat_qq_line()` to make it easy to add a simple line to a Q-Q plot. This
   line makes it easier to judge the fit of the theoretical distribution 
   (@nicksolomon).
+  
+* Added `tag` label for adding identification tags to the plot. A tag is added
```
hadley: I think you need a brief explanation of what you might use a tag for here

R/plot-build.r
```diff
@@ -261,6 +266,70 @@ ggplot_gtable.ggplot_built <- function(data) {
   plot_table <- gtable_add_grob(plot_table, caption, name = "caption",
     t = -1, b = -1, l = min(pans$l), r = max(pans$r), clip = "off")
 
+  plot_table <- gtable_add_rows(plot_table, unit(0, 'pt'), pos = 0)
```
hadley: Why do we need all these new rows and cols?

R/plot-build.r
```diff
@@ -261,6 +266,70 @@ ggplot_gtable.ggplot_built <- function(data) {
...
+  plot_table <- gtable_add_cols(plot_table, unit(0, 'pt'), pos = 0)
+  plot_table <- gtable_add_rows(plot_table, unit(0, 'pt'), pos = -1)
+  plot_table <- gtable_add_cols(plot_table, unit(0, 'pt'), pos = -1)
+
+  tag_pos <- theme$plot.tag.position
+  if (length(tag_pos) == 2) tag_pos <- "manual"
+  stopifnot(tag_pos %in% c("topleft", "top", "topright", "left", "right",
```
hadley: Can you please make sure you use modern tidyverse style everywhere?

R/theme.r
```diff
@@ -157,6 +157,12 @@
 #'   inherits from `title`) left-aligned by default
 #' @param plot.caption caption below the plot (text appearance)
 #'   (`element_text`; inherits from `title`) right-aligned by default
+#' @param plot.tag upper-left label to identify a plot (text appearance)
+#'   (`element_text`; inherits from `title`) left-aligned by default
+#' @param plot.tag.position The position of the tag as a string ("topleft",
+#'   "top", "topright", "left", "right", "bottomleft", "bottom", "bottomright)
+#'   or a coordinate. In the former case extra space will be added to accomodate
```
hadley: (as length-2 numeric vector)

"In former case" -> "If a string,"

R/plot-build.r
hadley: Should they always be added even when there isn't a tag? I don't have sense of if adding new rows/cols will disrupt existing code

thomasp85: The idea here, which is formed both by my work with facets and my extensions, is that it is much nicer to have a predictable gtable layout to work with. This is also why rows and columns are added for axes and strips on all sides even if they aren’t drawn. If any external code depend on the gtable layout it will break either way when users start to add tags, but doing it this way makes it much easier to fix as you don’t need any gtable introspection to see if a tag is added, and where.

There is nothing inside ggplot2 that breaks so this is more a breaking change for extensions

NEWS.md
thomasp85: Fixed in 5f70fd8

R/plot-build.r
thomasp85: The error has been improved in e439a97 - are there other issues with styling? (couldn't find any)

R/theme.r
thomasp85: Fixed in 80d3543

thomasp85:
I've added a few visual tests, but unrelated errors in some of the other tests prevents me from generating the figures. If the visual tests works on your end can you run it there?

thomasp85:
@clauswilke I generally have no problem with using your function, but always feel uneasy to copy code between branches that should get merged... What say you @hadley ?

baptiste:
@clauswilke [grob justification](https://stackoverflow.com/a/25456672/471093) is definitely a missing feature in gtable, so ideally it would go there rather than in ggplot2?

hadley:
@thomasp85 I just merged the other PR so you can use if you want

hadley:
@thomasp85 appveyor is borked right now, but travis is fine on master, so you need to figure out why it's failing.

thomasp85:
I'll look into it

thomasp85:
It's due to the `theme_test()`used in `vdiffr` not having defined the `plot.tag` and `plot.tag.position` elements. Do you want me to catch `NULL` values for these elements or make a change to vdiffr (cc @lionel- )?

lionel-:
A change to vdiffr would be great!

thomasp85:
A PR is on its way

thomasp85:
My local branch had been messed up somehow, so I've made a new rebase and force pushed it. Everything should check out...

lionel-:
I wonder where this comes from in the 3.2 builds...

```
Quitting from lines 87-90 (extending-ggplot2.Rmd) 
Error: processing vignette 'extending-ggplot2.Rmd' failed with diagnostics:
INTEGER() can only be applied to a 'integer', not a 'NULL'
```

lionel-:
@thomasp85 Could you RSwitch into 3.2 and try to get more info please?

lionel-:
FYI it's practical to have this in `Renviron`:

```
R_LIBS_USER= ~/R/library/%v/
```

This way you'll get one library per major R version, though you have to create the folder manually for it take effect. So here you'd create `~/R/library/3.2` before installing ggplot2 deps in there.

This setup makes it easy to test on multiple versions.

If it's too much bother I can try to find some time to look into it.

thomasp85:
I'll have a go with a docker container, but otherwise you're very welcome - I'll get back to you

thomasp85:
I have an appointment now so can only look further tonight - if you have the time you're very welcome to investigate

lionel-:
On Travis the issue only appears in the vignette but when I run the tests locally I see many of these failures. One example:

```r
p <- ggplot(data.frame(x = c(-Inf, 2, 0), y = c(Inf, 6, 4)), aes(x, y)) +
  geom_point() +
  coord_trans(x = 'identity')
benchplot(p)
#> Error in grid.Call.graphics(L_setviewport, vp, TRUE) (from bench.r#17) :
#>   INTEGER() can only be applied to a 'integer', not a 'NULL'

traceback()
#> 16: grid.Call.graphics(L_setviewport, vp, TRUE)
#> 15: push.vp.viewport(X[[i]], ...)
#> 14: FUN(X[[i]], ...)
#> 13: lapply(vps, push.vp, recording)
#> 12: pushViewport(vp, recording = FALSE)
#> 11: pushgrobvp.viewport(x$vp)
#> 10: pushgrobvp(x$vp)
#> 9: pushvpgp(x)
#> 8: preDraw.gTree(x)
#> 7: preDraw(x)
#> 6: drawGTree(x)
#> 5: recordGraphics(drawGTree(x), list(x = x), getNamespace("grid"))
#> 4: grid.draw.gTree(grob)
#> 3: grid.draw(grob)
#> 2: system.time(grid.draw(grob)) at bench.r#17
#> 1: benchplot(p)
```

thomasp85:
Thanks Lionel. I’ll look into where I’ve inadvertently passed a null to a grid call

thomasp85:
That was one weird bug grid had back then... 😱

should be fixed now for all versions

clauswilke:
Oh, that was the issue. It's the reason why I increased the minimum R version for cowplot to 3.3. Maybe you could add a small comment explaining why the code is written this way? E.g., "avoid assigning into unit vectors for compatibility with R <= 3.2."

thomasp85:
Sure - that's a good idea

lionel-:
@thomasp85 if you rebase on master the appveyor build should now be fixed, just in case you'd like to have that sweet green checkmark ;)

hadley:
@thomasp85 Do you want to use `justify_grob()` now that the PR has been merged?

thomasp85:
Already implemented in 12646f8

thomasp85:
@lionel- it seems AppVeyor and vdiffr are still acting up despite a rebase...

lionel-:
@thomasp85 you should now use `expect_doppelganger()` without a namespace prefix, there is an indirection defined in `helper-vdiffr.R`. This way the visual test devolves to `ggplot_build()` on Appveyor (until I have a chance to finish adapting the new vdiffr version for Windows).

thomasp85:
gotcha

PR #2408, opened 2018-01-19
Clarifying description of facet_wrap.RD

ZoidbergIII:
The previous version contained the phrase: "Most displays are roughly rectangular".
The meaning is understandable but this is confusing in a strictly mathematical sense: most display are indeed rectangular but this is not the point.
The point is that the aspect ratio (width/height) is not too high, i.e. we don't have extremely wide displays (which would be also rectangular)

### DIFF
--- a/man/facet_wrap.Rd
+++ b/man/facet_wrap.Rd
@@ -57,7 +57,7 @@ either of the four sides by setting \code{strip.position = c("top",
 \description{
 \code{facet_wrap} wraps a 1d sequence of panels into 2d. This is generally
 a better use of screen space than \code{\link[=facet_grid]{facet_grid()}} because most
-displays are roughly rectangular.
+displays have an aspect ratio (width\\height) lower than 2.
 }
 \examples{
 ggplot(mpg, aes(displ, hwy)) +


batpigandme:
Hi @ZoidbergIII,

Since ggplot2 (and most R packages) use [roxygen2](https://github.com/klutometis/roxygen), edits to documentation need to be made in their source files, or else they will be overwritten when documentation is next generated (.Rd files should [not be edited by hand](https://github.com/ZoidbergIII/ggplot2/blob/5bccd3dd506192b38c9cd6f41bc5a3bbc8dd95c2/man/facet_wrap.Rd#L1-L2)). You can learn more about this in [The tidyverse style guide](http://style.tidyverse.org/code-documentation.html), CONTRIBUTING.md, Hadley's [roxygen vignette](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html), etc.

If you don't want to deal with roxygen, you can just edit the `.R` file, that's fine. The `.Rd` will be updated when we next run `devtools::document()`.

zoidberg_ggplot2

I'd also suggest leaving some reference to "roughly rectangular" in there, since different things "click" with different people, and I don't think it would detract from what you've added.

Thanks!

PR #2418, opened 2018-01-22
Add error when passing in a function as data to ggplot()

### LINKED ISSUE #2417: Passing a function instead of a data frame as data creates a confusing error message
Reproducible example:
```r
ggplot(function() {}, aes(x, y)) + geom_point()
#  Error in if (empty(data)) return(cbind(data, PANEL = factor())) : 
#  missing value where TRUE/FALSE needed 
```

This problem can arise, for example, if somebody assigns data to `df` and then mistypes it as `pf`, or if they use `df` without assigning any data to it in the first place.

thomasp85:
Fix #2417


thomasp85:
I'm unsure whether to deparse the `data` object and add the name to the error message. In interactive use it could be a boon, but will probably only confuse if the plot is created inside another function...

PR #2430, opened 2018-01-30
Fix check as a result of f61bfd62

ruaridhw:
New `fortify.default` error message is misaligned with the `expect_error` msg

See changes in f61bfd620037b7e1c816129469b106443651a4dc


ruaridhw:
Closing as #2434 is a duplicate

PR #2434, opened 2018-02-01
Fix expected error message for fortify.default with class uneval

karawoo:
f61bfd620037b7e1c816129469b106443651a4dc updated the error message but not the test that expects the old message.


hadley:
Can you make the regexp a bit smaller please?

karawoo:
I chopped off enough to fit within 80 characters and replaced it with `(.+)`, not sure if that's what you meant.

hadley:
Hmmm I actually realise I dislike the whole test - testing if something is “helpful” seems to vague. But it’s there already and it’s causing a problem so this fix looks good. Thanks!

PR #2449, opened 2018-02-12
fix rd_aesthetics for clarity

nzxwang:
document that bolded aesthetics can be either required *or* computed. I was confused when reading the documentation for geom_boxplot, it seemed to claim that the computed attributes were required arguments, so I added a note indicating the bolded attributes may be computed as well.


hadley:
I agree that the current wording is suboptimal, but I think your change goes too far in the other direction, and will add a bunch of confusion in geoms that don't have any computed variables.

PR #2450, opened 2018-02-13
add an example of using "breaks" to omit legend items

luisDVA:
added a quick example at the relevant section of the legend guide, since the documentation is not explicit on setting breaks on discrete scales to exclude/omit/drop legend items


hadley:
I think it would be better to describe this ability in the documentation for `breaks` in `discrete_scale()` (which gets pulled into `scale_discrete()`). Do you want to have a go at doing that?

hadley:
`limits` adjusts the full set of values used by the scale; `breaks` affects what appears on the legend.

luisDVA:
good point. Where should I add the example? I could only find similar examples in the 
guide-colorbar.R file

PR #2458, opened 2018-02-20
Pass `label` to `draw_key_text`

brianwdavis:
To override the default "a" in the label of a text legend: first check if `data$label` has anything in it; if not, use "a" as before. However, if the user sets a label within `guide_legend(override.aes = list(label = "foo"))`, then use that label or vector of labels as the first arg of `textGrob`.

Motivation
----
This is a relatively common request on SO, to wit: https://stackoverflow.com/questions/48892982/change-key-letter-in-legend-of-ggplot

The problem
----
The current version of `draw_key_text` has a hardcoded "a" as the first arg of `textGrob`.

Example using the fix
----
ggplot(mtcars, aes(mpg, wt, color = factor(am), label = disp)) +
  geom_text() +
  guides(color = guide_legend(override.aes = list(label = c("a", "m"))))

image

### DIFF
--- a/R/legend-draw.r
+++ b/R/legend-draw.r
@@ -169,7 +169,9 @@ draw_key_smooth <- function(data, params, size) {
 #' @export
 #' @rdname draw_key
 draw_key_text <- function(data, params, size) {
-  textGrob("a", 0.5, 0.5,
+  if(is.null(data$label)) data$label <- "a"
+  
+  textGrob(data$label, 0.5, 0.5,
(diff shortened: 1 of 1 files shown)


hadley:
Nice!

Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`.

brianwdavis:
Added under the Scales section.

PR #2467, opened 2018-03-01
Port mappings to tidy eval

lionel-:


lionel-:
@thomasp85 is there any ggplot extension out there that makes assumptions about the structure of `params$facets`, i.e. that it's a `quoted` object from the plyr package?

thomasp85:
Maybe...

Good news is that if there are, they are all in one of my packages with a very high probability - I'll make sure to take care of these...

DESCRIPTION
```diff
@@ -52,7 +52,7 @@ Suggests:
 Remotes: hadley/scales,
     hadley/svglite,
     jimhester/withr,
-    tidyverse/rlang
+    lionel-/rlang@1def2e2cffaaa
```
hadley: I assume there will be an rlang PR soon?

lionel-: There is already one: https://github.com/r-lib/rlang/pull/471

lionel-:
> I didn't read the code too carefully as I assume it works.

I couldn't run the doppelganger tests because I don't have the right deps on my machine though :/. The default legends are not deparsed correctly (you see the `~` of quosures). I'm going to fix this before merging, but it'd be nice to run the visual tests properly before release. Maybe it's time to finally fix vdiffr.

Issue #2480, opened 2018-03-14
position_dodge2() should handle both point and interval geoms


frostell:
Displaying raw data together with box plots can be useful.

If using both 'x' and 'fill' aesthetic geom_boxplot() handles NA's elegantly through the new position_dodge2(), but unfortunately geom_point() with a matching 'colour' aesthetic is misaligned with the boxes both when using position_dodge() and position_dodge2().

This problem only occurs when handling NA's, and position_dodge() and position_dodge2() gets it wrong in different ways:

```r
library('tidyverse')

dat <- data.frame("value" = rnorm(n=30, mean=2, sd=0.5),
                  "group" = LETTERS[1:3],
                  "x" = factor(1:2))

dat$value[dat$group=="A"&dat$x=="1"] <- NA

ggplot(dat, aes(x=x, y=value)) +
  geom_boxplot(aes(fill=group), alpha=0.3) +
  geom_point(aes(colour=group), position=position_dodge(width=0.75), size=3, alpha=0.5)

ggplot(dat, aes(x=x, y=value)) +
  geom_boxplot(aes(fill=group), alpha=0.3) +
  geom_point(aes(colour=group), position=position_dodge2(width=0.75), size=3, alpha=0.5)

```

batpigandme:
reprex
``` r
library(tidyverse)

dat <- data.frame("value" = rnorm(n=30, mean=2, sd=0.5),
                  "group" = LETTERS[1:3],
                  "x" = factor(1:2))

dat$value[dat$group=="A"&dat$x=="1"] <- NA

ggplot(dat, aes(x=x, y=value)) +
  geom_boxplot(aes(fill=group), alpha=0.3) +
  geom_point(aes(colour=group), position=position_dodge(width=0.75), size=3, alpha=0.5)
#> Warning: Removed 5 rows containing non-finite values (stat_boxplot).
#> Warning: Removed 5 rows containing missing values (geom_point).
```

![](https://i.imgur.com/1zM9pLb.png)

``` r

ggplot(dat, aes(x=x, y=value)) +
  geom_boxplot(aes(fill=group), alpha=0.3) +
  geom_point(aes(colour=group), position=position_dodge2(width=0.75), size=3, alpha=0.5)
#> Warning: Removed 5 rows containing non-finite values (stat_boxplot).

#> Warning: Removed 5 rows containing missing values (geom_point).
```

![](https://i.imgur.com/HaQ4ypJ.png)

[...]

PR #2481, opened 2018-03-14
small update to position-dodge2.r

frostell:
Small update to position-dodge2.r to enable alignment of geom_boxplot() and geom_point() with NA's through position_dodge2(), possible patch for #2480.

### DIFF
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -69,10 +69,10 @@ pos_dodge2 <- function(df, width, n = NULL, padding = 0.1) {
     df$xmin <- df$x
     df$xmax <- df$x
   }
-  
+
   # to enable functionality when input is not a stat, save copy and reduce df
   df_full <- df
-  if (!all(c("middle") %in% names(df))) {  
+  if (!all(c("ymin") %in% names(df))) {  
     df <- df[!is.na(df$y), ]
     df <- df[!duplicated(df$group), ]
   }
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -69,7 +69,14 @@ pos_dodge2 <- function(df, width, n = NULL, padding = 0.1) {
     df$xmin <- df$x
     df$xmax <- df$x
   }
-
+  
+  # to enable functionality when input is not a stat, save copy and reduce df
+  df_full <- df
+  if (!all(c("middle") %in% names(df))) {  
+    df <- df[!is.na(df$y), ]
+    df <- df[!duplicated(df$group), ]
+  }
+  
   # xid represents groups of boxes that share the same position
   df$xid <- find_x_overlaps(df)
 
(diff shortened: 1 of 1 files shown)


frostell:
Hmm, I realise reading the logs for the failed checks that my suggested update breaks functionality as defined in test-position-dodge2.r (it gives a different result for geom_col() and geom_rect() when there is x-overlap, but no grouping).

- changed one word in a new commit and I hope the tests will pass now!

frostell:
I don't think the travis-ci check broke because of my change, it passes all your tests scripts @karawoo, what do you think?

karawoo:
Thanks for looking into this @frostell. When I try your update I still get misaligned points and boxes on the left, does this match what you see?

``` r
# using example from #2480
set.seed(5820)
dat <- data.frame("value" = rnorm(n=30, mean=2, sd=0.5),
                  "group" = LETTERS[1:3],
                  "x" = factor(1:2))

dat$value[dat$group=="A"&dat$x=="1"] <- NA

ggplot(dat, aes(x=x, y=value)) +
  geom_boxplot(aes(fill=group), alpha=0.3) +
  geom_point(aes(colour=group), position=position_dodge2(width=0.75), size=3, alpha=0.5)
#> Warning: Removed 5 rows containing non-finite values (stat_boxplot).
#> Warning: Removed 5 rows containing missing values (geom_point).
```

karawoo:
Ah I see what's happened. I applied your changes onto the current master branch. `position_dodge2()` was updated in #2386 to make `preserve = "total"` the default. I realized in looking into this that I neglected to make one change in that PR that caused it to revert back to `"single"` for the points but not the boxes, hence the mismatch. This is addressed in #2545.

frostell:
With this small update pos.dodge2() works with geom_point too and it's possible to align boxplots and points even in trickier cases when some unique combinations are all NA's.

Background:
Displaying raw data points together with box plots can be useful. When using a fill aesthetic for the box plots, position.dodge2() gives great positioning options. Raw data can be plotted with geom_point() and position.dodge() for alignment.

Problem:
When one of the unique combinations of a category and fill/color is all NA's, position.dodge() and position.dodge2() gives different x-positions and the boxes and points end up misaligned. For geom_point, the current position.dodge2() does not work because it treats every unique point as a category that should be dodged from all the other points. See #2480 for reprex.

Solution:
Change pos.dodge2() so that:
    - It evaluates what kind of data is fed into it by checking for existing column names
    - If data is not a stat: save the full df and the reduce it to just the unique combinations of fill/color and x-category (the rest of pos.dodge2() runs like before and calculates appropriate x-positions)
    - at the end, repopulate the full df with the x-positions calculated

hadley:
Your logic doesn't seem quite general enough to me. I'll take a stab at it myself.

frostell:
ok, I admit it was patchy even though it worked for the specific problem :)

I guess ideally there would be just one position_dodge function to handle all geoms and stats with appropriate options to choose styling of the "dodging"... Looking forward to see your solution!

PR #2490, opened 2018-03-22
Port facets to tidy eval

lionel-:


NEWS.md
```diff
@@ -2,11 +2,26 @@
 
 ## New features
 
-* `aes()` now supports quasiquotation so that you can use `!!`, `!!!`, and
-  `:=`. (ggplot2 does not currently support full tidy evaluation because
-  when I wrote ggplot2 my understanding of NSE was quite flawed, and hence
-  ggplot2 only captures one environment per plot, not one environment
-  per aesthetic. We will fix this in a future release.)
+* `aes()` now supports quasiquotation so that you can use `!!`, `!!!`,
```
hadley: Oh here it is :)

You should also mention that this replaces `aes_string()` and `aes_()`

NEWS.md
```diff
@@ -2,11 +2,26 @@
...
+  and `:=`.
+
+* `facet_wrap()` and `facet_grid()` now support `vars()` inputs. Like
```
hadley: I think this needs something to emphasise that `~` is not going away (i.e. it's not even soft-deprecated) but the new interface makes ggplot2 easier to program with.

R/facet-.r
```diff
@@ -203,38 +264,152 @@ df.grid <- function(a, b) {
...
+}
+f_as_facets <- function(f) {
+  if (is.null(f)) {
+    return(rlang::as_quosures(list()))
+  }
+
+  env <- rlang::f_env(f) %||% globalenv()
+
+  # as.quoted() handles `+` specifications
+  vars <- plyr::as.quoted(f)
```
hadley: How much work would it be to eliminate `as.quoted()` altogether?

R/facet-grid-.r
```diff
@@ -50,10 +50,25 @@ NULL
...
 #' @export
 #' @examples
 #' p <- ggplot(mpg, aes(displ, cty)) + geom_point()
 #'
+#' # Use vars() to supply variables from the dataset:
+#' p + facet_grid(vars(drv))
+#' p + facet_grid(cols = vars(cyl))
+#' p + facet_grid(vars(drv), vars(cyl))
+#'
+#' # The historical formula interface is also available:
```
hadley: Maybe classic instead of historical? I think that has a more positive connotation

R/facet-.r
lionel-: If we handled the formula decomposition (turning an addition of expressions to a list of quosures) that would take care of the majority of the cases, but `as.quoted()` is a S3 generic so for backward compat I thought it was best to keep it. Do you want to remove it?

hadley: Remove it - it's esoteric.

PR #2510, opened 2018-04-01
edit geom_tile aesthetics and fix docs

malcolmbarrett:
The `geom_rect` and `geom_tile` doc examples include an aesthetic parameter for width, but neither geom has that, so the examples produce this warning:

``` r
ggplot(df, aes(x, y)) +
  geom_tile(aes(fill = z, width = w), colour = "grey50")

# > Warning: Ignoring unknown parameters: width
```
For `geom_tile`, I moved height and width over to `default_aes`. For `geom_rect`, though, I just removed `width = w` from the example in the docs because it seemed like a mistake (since `geom_rect` calculates width from `xmin` and `xmax`).

Thanks!
Malcolm


hadley:
Thanks!  I ended up merging by hand so I could add a unit test too.

PR #2521, opened 2018-04-12
Add `z` as required aesthetic in documentation of `geom_contour()`

eliocamp:
Currently the only required aesthetics in the documentation are `x` and `y`. Adding `z` to `GeomContour`s `required_aes` does not work because as a geom, it actually never uses that variable. 
I guess changing the `rd_aesthetics()` call is the easiest way to fix it.


hadley:
That fixes one problem but creates a new problem 😞

PR #2542, opened 2018-04-28
Fix vdiffr

lionel-:
There is still some work to do but this could already be merged.

TODO:

- Build static libraries of the freetypeharfbuzz package (will probably rename it to fontshaper) for Windows. Until then the Appveyor builds will fail.

- Double check freetypeharfbuzz builds correctly on all platforms. It builds on all r-hub unix platforms but I got an error on Travis macOS.

- Possibly: Create a package that links statically against Cairo in order to produce raster images deterministically.


hadley:
Thanks!

PR #2545, opened 2018-04-29
Swap order of preserve arguments in position_dodge2()

karawoo:
#2386 made `preserve = "total"` the default for `PositionDodge2`, but didn't swap the order of the defaults in `position_dodge2()`. This fixes it.


hadley:
Thanks!

Issue #2497, opened 2018-03-23
facet_wrap() giving error in development version


tbradley1013:
Hi, 

I updated to the development versions of `rlang` and `ggplot2` and ran into an issue with `facet_wrap`. I re-installed the CRAN versions of both and `facet_wrap` works under the CRAN version. I have attached a `reprex` of the problem below:

```
library(ggplot2)

# getting error using either the formula method or vars()
# when trying to use facet_wrap
ggplot(iris, aes(Sepal.Length, Sepal.Width)) + 
  facet_wrap(~Species) + 
  geom_point()
#> Error: `predicate` must be a closure or function pointer

# same thing but with vars()
ggplot(iris, aes(Sepal.Length, Sepal.Width)) + 
  facet_wrap(vars(Species)) + 
  geom_point()
#> Error: `predicate` must be a closure or function pointer

# facet_grid works
ggplot(iris, aes(Sepal.Length, Sepal.Width)) + 
  facet_grid(vars(Species)) + 
  geom_point()
```

![image](https://user-images.githubusercontent.com/23705496/37834933-5f8521ca-2e85-11e8-9a05-3bc39f87c607.png)

```r
ggplot(iris, aes(Sepal.Length, Sepal.Width)) + 
  facet_grid(~ Species) + 
  geom_point()
```
![image](https://user-images.githubusercontent.com/23705496/37834963-6c5f4d62-2e85-11e8-98a2-dbd40bb1fedb.png)

#' Created on 2018-03-23 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

tbradley1013:
I just restarted my computer and reinstalled `rlang` and `ggplot2` from github and now the facet_wrap works. I am not sure what changed. It seemed like the issue may have something to do with the `flatten_if()` call [here](https://github.com/tidyverse/ggplot2/blob/master/R/facet-wrap.r#L117), but since it is no longer happening I will close the issue.

JoGall:
I'm experiencing the same error (R 3.4.1, Windows 10).

I've reinstalled `ggplot2` from GitHub as well as reinstalling dependencies `rlang`, `withr`, `svglite`, and `scales`, all without error.

Reprex:
```
ggplot(iris, aes(Sepal.Length, Sepal.Width)) +
  geom_point() +
  facet_wrap(~Species)
```

batpigandme:
[...]

PR #2546, opened 2018-04-30
fix flatten_if bug in facet_wrap

AdeelH:
Implement the fix described here: https://github.com/tidyverse/ggplot2/issues/2497#issuecomment-385333245


hadley:
Replaced by #2572

PR #2549, opened 2018-04-30
remove unwanted annotation border from saved PDFs; closes #2407

### LINKED ISSUE #2407: Inconsistency between generated pdf and png files
I have observed an inconsistency between the generated pdf and png files in the plot below.
In the plot I insert an annotate object with the type 'label'; where the width of the black line around the label is set to zero.
This works fine when exporting to a png file.
But when I export the plot to a pdf file, the black line is still present.

```
library(tidyverse)
#> ── Attaching packages ──────────────────────────────────────────────────── tidyverse 1.2.1 ──
#> ✔ ggplot2 2.2.1     ✔ purrr   0.2.4
#> ✔ tibble  1.4.1     ✔ dplyr   0.7.4
#> ✔ tidyr   0.7.2     ✔ stringr 1.2.0
#> ✔ readr   1.1.1     ✔ forcats 0.2.0
#> ── Conflicts ─────────────────────────────────────────────────────── tidyverse_conflicts() ──
#> ✖ dplyr::filter() masks stats::filter()
#> ✖ dplyr::lag()    masks stats::lag()
p <- tibble() %>%
  ggplot() + 
  annotate('label', x = 1, y = 0, 
           label='is there a border around this label?', 
           label.size=0)

# In the generated pdf file, there is a black border around the inserted label.
# I don't think that this is the intended behaviour.
ggsave('failure.pdf', p, 
       height = 3, width=7)

# In the generated png file, there is no border around the inserted label.
# I think that this is the intended behaviour.
ggsave('failure.png', p, 
       height = 3, width=7)
```

bfgray3:
I have tried to implement the fix suggested by Winston in the comments on this issue.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -202,6 +202,9 @@ up correct aspect ratio, and draws a graticule.
 
 ### Layers
 
+* `annotate` no longer produces an undesired border around annotations when
+  `label.size` is 0, even when saving to PDF (@bfgray3, #2407).
+
 * In most cases, using `%>%` instead of `+` should generate an informative
   error (#2400).
   
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -90,7 +90,7 @@ GeomLabel <- ggproto("GeomLabel", Geom,
           lineheight = row$lineheight
         ),
         rect.gp = gpar(
-          col = row$colour,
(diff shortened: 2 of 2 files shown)


hadley:
Thanks!

Issue #2465, opened 2018-02-27
Legend title is incorrectly aligned when horizontal


hadley:
``` r
library(ggplot2)

df <- data.frame(z = 1:5)
p <- ggplot(df, aes(1, 1, colour = z)) + geom_point()

# Title should be centered with legend
p + theme(legend.position = "top")
```

![](https://i.imgur.com/a4niJZg.png)

hadley:
@clauswilke have you seen this too? I quickly skimmed through your PRs but I didn't see anything that addressed this explicitly

clauswilke:
I haven't addressed or modified this in any way. What is happening is that the title is centered relative to the entire legend:

```{r}
p + theme(legend.position = "top") + 
  theme(legend.spacing.y = grid::unit(50, "cm")) # 50 cm equals to 5 cm because of issue #2398
```
<img width="580" alt="screen shot 2018-05-01 at 5 49 25 pm" src="https://user-images.githubusercontent.com/4210929/39497334-7267c3cc-4d68-11e8-9761-5ce18a673fc1.png">

clauswilke:
Note that `guide_legend()` behaves exactly the same:
```{r}
df <- data.frame(z = 1:5)
p <- ggplot(df, aes(1, 1, colour = factor(z))) + 
  geom_point() + 
  scale_color_discrete(guide = guide_legend(label.position = "bottom"))

# Title should be centered with legend
p + theme(legend.position = "top") + theme(legend.spacing.y = grid::unit(5, "cm"))
```
<img width="580" alt="screen shot 2018-05-01 at 5 56 08 pm" src="https://user-images.githubusercontent.com/4210929/39497446-ff881e5a-4d68-11e8-88bc-567959ffabc2.png">

hadley:
Oh so this probably isn't a regression

clauswilke:
The bigger issue, in my mind, is that there is no simple workaround, because the `vjust` setting is ignored. I'm showing this here with both a discrete and a continuous scale, to show that they behave the same:

```{r}
library(ggplot2)
p <- ggplot(mtcars, aes(hp, disp, color = mpg, shape = factor(cyl))) +
  geom_point(size = 2) +
  scale_shape_discrete(guide = guide_legend(label.position = "bottom")) + 
  theme(legend.position = "top")
p
```
[...]

PR #2556, opened 2018-05-07
Enable margins settings for guide titles

clauswilke:
This pull request, while not completely solving https://github.com/tidyverse/ggplot2/issues/2465, makes it much easier to work around it, by enabling `margin`, `hjust`, and `vjust` for the theme element `legend.title`. This enables fine-grained justification control over the location of the legend title, which is needed for plots with more than one legend.

Examples:

```{r}
library(ggplot2)
p <- ggplot(mtcars, aes(hp, disp, color = mpg, shape = factor(cyl))) +
  geom_point(size = 2) +
  theme(legend.position = "top",
        legend.title = element_text(vjust = 1, margin = margin(t = 4.5)),
        legend.spacing.y = grid::unit(3, "pt"),
        legend.text.align = 0.5)
p
```
screen shot 2018-05-06 at 9 39 21 pm

```{r}
p + guides(shape = guide_legend(label.position = "bottom"))
```
screen shot 2018-05-06 at 9 39 35 pm

For comparison, with the default settings but legend on top, this plot looks as follows:
```{r}
ggplot(mtcars, aes(hp, disp, color = mpg, shape = factor(cyl))) +
  geom_point(size = 2) +
```
screen shot 2018-05-06 at 9 49 27 pm
(Unfortunately, there is no good way at this time to adjust the default settings depending on whether the legend sits on top or at the side. This will have to wait for later.)

Note that none of the visual tests except one had to be changed. Things look exactly as they always did if the default settings are not changed. The visual test that had to be changed had the problem that the two legends in that plot switched order. I have now fixed the order, so this won't happen again.

### DIFF
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -324,12 +324,17 @@ guide_gengrob.colorbar <- function(guide, theme) {
   # and to obtain the title fontsize.
   title.theme <- guide$title.theme %||% calc_element("legend.title", theme)
 
+  title.hjust <- guide$title.hjust %||% theme$legend.title.align %||% title.theme$hjust %||% 0
+  title.vjust <- guide$title.vjust %||% title.theme$vjust %||% 0.5
+
   grob.title <- ggname("guide.title",
     element_grob(
       title.theme,
       label = guide$title,
-      hjust = guide$title.hjust %||% theme$legend.title.align %||% title.theme$hjust %||% 0,
-      vjust = guide$title.vjust %||% title.theme$vjust %||% 0.5
+      hjust = title.hjust,
+      vjust = title.vjust,
+      margin_x = TRUE,
+      margin_y = TRUE
     )
   )
 
@@ -487,7 +492,11 @@ guide_gengrob.colorbar <- function(guide, theme) {
   gt <- gtable_add_grob(gt, grob.label, name = "label", clip = "off",
     t = 1 + min(vps$label.row), r = 1 + max(vps$label.col),
     b = 1 + max(vps$label.row), l = 1 + min(vps$label.col))
-  gt <- gtable_add_grob(gt, grob.title, name = "title", clip = "off",
+  gt <- gtable_add_grob(
+    gt,
+    justify_grob(grob.title, hjust = title.hjust, vjust = title.vjust),
+    name = "title",
+    clip = "off",
     t = 1 + min(vps$title.row), r = 1 + max(vps$title.col),
     b = 1 + max(vps$title.row), l = 1 + min(vps$title.col))
   gt <- gtable_add_grob(gt, grob.ticks, name = "ticks", clip = "off",
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -256,21 +256,19 @@ guide_geom.colorbar <- function(guide, layers, default_mapping) {
 guide_gengrob.colorbar <- function(guide, theme) {
 
   # settings of location and size
-  switch(guide$direction,
-    "horizontal" = {
-      label.position <- guide$label.position %||% "bottom"
-      if (!label.position %in% c("top", "bottom")) stop("label position \"", label.position, "\" is invalid")
+  if (guide$direction == "horizontal") {
+    label.position <- guide$label.position %||% "bottom"
+    if (!label.position %in% c("top", "bottom")) stop("label position \"", label.position, "\" is invalid")
 
-      barwidth <- width_cm(guide$barwidth %||% (theme$legend.key.width * 5))
-      barheight <- height_cm(guide$barheight %||% theme$legend.key.height)
-    },
-    "vertical" = {
-      label.position <- guide$label.position %||% "right"
-      if (!label.position %in% c("left", "right")) stop("label position \"", label.position, "\" is invalid")
+    barwidth <- width_cm(guide$barwidth %||% (theme$legend.key.width * 5))
+    barheight <- height_cm(guide$barheight %||% theme$legend.key.height)
+  } else { # guide$direction == "vertical"
+    label.position <- guide$label.position %||% "right"
+    if (!label.position %in% c("left", "right")) stop("label position \"", label.position, "\" is invalid")
 
-      barwidth <- width_cm(guide$barwidth %||% theme$legend.key.width)
-      barheight <- height_cm(guide$barheight %||% (theme$legend.key.height * 5))
-    })
+    barwidth <- width_cm(guide$barwidth %||% theme$legend.key.width)
+    barheight <- height_cm(guide$barheight %||% (theme$legend.key.height * 5))
+  }
 
   barlength <- switch(guide$direction, "horizontal" = barwidth, "vertical" = barheight)
   nbreak <- nrow(guide$key)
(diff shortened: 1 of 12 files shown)


clauswilke:
I have updated the pull request to cover legend text margins as well. I realize some visual tests still fail. Those are related to the default hjust setting of legend text and I will fix the issue.

clauswilke:
I think I've got this working now (https://github.com/tidyverse/ggplot2/pull/2556), except for a minor problem of the default `hjust` setting that I need to look into.

```{r}
library(ggplot2)
df <- data.frame(x = c("a", "bbbbbbbbbbbbb"))

ggplot(df, aes(x = x, fill = x)) +
  geom_bar() +
  theme(legend.text = element_text(margin = margin(l = 50), hjust = 0))
```
screen shot 2018-05-07 at 12 32 17 am

clauswilke:
Examples with both types of legends:
```{r}
ggplot(mtcars, aes(hp, disp, color = mpg, shape = factor(cyl))) +
  geom_point(size = 2) +
  theme(legend.text = element_text(margin = margin(l = 20)))
```
screen shot 2018-05-07 at 12 37 52 am

```{r}
ggplot(mtcars, aes(hp, disp, color = mpg, shape = factor(cyl))) +
  geom_point(size = 2) +
  guides(shape = guide_legend(label.position = "bottom")) +
  theme(legend.position = "top",
        legend.title = element_text(vjust = 1, margin = margin(t = 4.5)),
        legend.spacing.y = grid::unit(3, "pt"),
        legend.text = element_text(margin = margin(t = 10)))
```
screen shot 2018-05-07 at 12 38 09 am

hadley:
My main concern with enabling this is that if we do later fix the alignment of titles to align "correctly" by default, then all these smaller fixes will break. OTOH, it's not totally clear what "correct" means, because when I think about aligning the title only with the keys, not the text, that's clearly not correct for a vertical legend.

clauswilke:
The way I see it, `legend.title = element_text(vjust = 1, margin = margin(t = 4.5))` simply states that the legend title should be aligned to the top of the legend box, with a 4.5pt margin. This should continue working in the future, as long as the key height doesn't change. In fact, one way to get the alignment right automatically in the future would be to use exactly this approach but auto-calculate the margin based on the key height.

clauswilke:
Margins and alignment work now for labels, as far as I can tell. I'm quite happy with the results. 

First, manual specification of justification and margins:

```{r}
df <- data.frame(x = c(1, 10, 100))
p <- ggplot(df, aes(x, x, color = x, size = x)) + geom_point() + theme_gray(7)

p + theme(legend.text = element_text(hjust = 1, margin = margin(l = 5, t = 10, b = 10)))
```
screen shot 2018-05-07 at 11 40 41 am

Second, defaults (for labels) make sense in all arrangements:

```{r}
cowplot::plot_grid(
  p,
  p + guides(size = guide_legend(label.position = "bottom")), 
  p + guides(size = guide_legend(label.position = "left"), color = guide_colorbar(label.position = "left")), 
  p + theme(legend.position = "top"),
  p + theme(legend.position = "top") + 
    guides(size = guide_legend(label.position = "bottom")),
  p + theme(legend.position = "top") + 
    guides(size = guide_legend(label.position = "top"), color = guide_colorbar(label.position = "top")),
  nrow = 2
)
```
screen shot 2018-05-07 at 11 40 03 am

Two issues remain. None are regressions:

1. I could set more sophisticated defaults for the legend title depending on the legend layout. E.g., set `vjust = 1` for horizontal legends.

2. The code contains (from a long time ago) a check for whether a label is an expression. This check is not reliable and doesn't work for lists of expressions, e.g. as returned by `scales::math_format()`. The original line in question is this one:
https://github.com/tidyverse/ggplot2/blob/9034d535f06f0437bb1ad8fdda2c47a8a743bcf7/R/guide-legend.r#L363

karawoo:
The results of this look really good @clauswilke! This issue with legend title/text margins has plagued me for some time 😅 

The desired behavior we're aiming for is so similar to facet strip labels that it _seems_ like there should be a way to fold the placement calculations into existing functions rather than having the legend titles go through `title_spec()` -> `add_margins()` -> `titleGrob()` and then overriding the placement with `justify_grobs()`. But I haven't figured out where/how to accomplish that.

clauswilke:
@karawoo I think the problem arises because you don't know ahead of time whether the legend title and text will be integrated in a horizontal or vertical setting. In cases where you do know it, `justify_grobs()` is not needed, such as for the labels in `guide_colorbar()`.

Also, I'd like to point out that `justify_grobs()` was de facto invented by @thomasp85 in the context of plot tags here: 
https://github.com/thomasp85/ggplot2/blob/1e61f9dea5a5205cc865d1c43690240d4159bebb/R/plot-build.r#L282
I just copied this code and put it into a function.

R/guide-legend.r
```diff
@@ -677,7 +693,7 @@ guide_gengrob.legend <- function(guide, theme) {
   )
   gt <- gtable_add_grob(
     gt,
-    grob.title,
+    justify_grobs(grob.title, hjust = title.hjust, vjust = title.vjust),
```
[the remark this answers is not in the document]
clauswilke: I think passing it in makes sense. I'll update my patch. I ran into a problem with switching debugging on for labels and I wasn't sure whether it was my fault or a preexisting problem. I'll see if I can resolve that issue.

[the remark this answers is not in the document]
clauswilke: It was a preexisting problem but it's fixed now.

R/guide-colorbar.r
```diff
@@ -344,17 +349,48 @@ guide_gengrob.colorbar <- function(guide, theme) {
...
   label.theme <- guide$label.theme %||% calc_element("legend.text", theme)
+
+  # We break inheritance for hjust and vjust, because that's more intuitive here; it still allows manual
+  # setting of hjust and vjust if desired. The alternative is to ignore hjust and vjust altogether, which
+  # seems worse
+  if (is.null(guide$label.theme$hjust) && is.null(theme$legend.text$hjust)) label.theme$hjust <- NULL
+  if (is.null(guide$label.theme$vjust) && is.null(theme$legend.text$vjust)) label.theme$vjust <- NULL
+
+  # label.theme in param of guide_legend() > theme$legend.text.align > default
+  hjust <- guide$label.hjust %||% theme$legend.text.align %||% label.theme$hjust %||%
```
hadley: Is it necessary to have both of `theme$legend.text.align` and `label.theme$hjust`? Shouldn't `label.theme` inherit  from `legend.text`?

R/guide-colorbar.r
```diff
@@ -344,17 +349,48 @@ guide_gengrob.colorbar <- function(guide, theme) {
...
   grob.label <- {
     if (!guide$label)
       zeroGrob()
     else {
-      hjust <- x <- guide$label.hjust %||% theme$legend.text.align %||%
-        if (any(is.expression(guide$key$.label))) 1 else switch(guide$direction, horizontal = 0.5, vertical = 0)
-      vjust <- y <- guide$label.vjust %||% 0.5
-      switch(guide$direction, horizontal = {x <- label_pos; y <- vjust}, "vertical" = {x <- hjust; y <- label_pos})
-
+      switch(
```
hadley: Since you're rewriting this, do you mind moving the assignment into the individual blocks of the if statement? I think that will make it easier to read.

R/guide-colorbar.r
```diff
@@ -344,17 +349,48 @@ guide_gengrob.colorbar <- function(guide, theme) {
...
+        guide$direction,
+        "horizontal" = {
+          x <- label_pos
```
hadley: Similarly, it would be great to move these assignments out.

R/guide-colorbar.r
clauswilke: In the current (released) ggplot2, there are two special theme elements for legend alignment, `legend.text.align` and `legend.title.align`. I think both are obsolete and have issues, for example that they don't apply in a vertical context. I only left them in for backwards compatibility, but I'm happy to take them out. `label.theme` does inherit from `legend.text`.

Did you also have a question about me overriding inheritance of `hjust` and `vjust` for `label.theme`? I'm happy to explain more. It's needed to make guides behave intuitively under default settings, so that, e.g., a horizontal guide with labels underneath has the correct label alignments (`hjust = 0.5` instead of `hjust = 0` and `vjust = 1` instead of `vjust = 0.5`).

R/guide-colorbar.r
clauswilke: Just so I understand what you're requesting. Patterns such as:
```{r}
switch(
  guide$direction,
  "horizontal" = {
    ...
  },
  "vertical" = {
  }
)
```
are used all over the place in this code. Do you want me to replace those with the following?
```{r}
if (guide$direction == "horizontal") {
  ...
}
else {
  ...
}
```

hadley: Yes please!

clauswilke: I had originally not understood the "move assignment" part of the request, but I think I got it now. The code is easier to read now.

hadley:
Thanks!

PR #2559, opened 2018-05-07
Change all default_aes alpha from NA to 1

### LINKED ISSUE #2558: Default values for alpha
Working on gganimate I came upon a little problem. For GeomPolygon and related geoms the alpha value is set to `NA` rather than 1, which is in practice the same. This results in bad transitions as transitioning into NA is ill-defined.

I could hack around this in tweenr/gganimate, but wondered whether you would be interested in a fix here instead.

thomasp85:
Fix #2558

This PR changes all the default values for the alpha aesthetic to 1 in the cases where it is currently set to `NA`. Any colour are currently passed through `scales::alpha()` if it should be modified with transparency and this function treats `NA` as 1, so there should be no practical effect of this change.

### DIFF
--- a/R/geom-abline.r
+++ b/R/geom-abline.r
@@ -121,7 +121,7 @@ GeomAbline <- ggproto("GeomAbline", Geom,
     GeomSegment$draw_panel(unique(data), panel_params, coord)
   },
 
-  default_aes = aes(colour = "black", size = 0.5, linetype = 1, alpha = NA),
+  default_aes = aes(colour = "black", size = 0.5, linetype = 1, alpha = 1),
   required_aes = c("slope", "intercept"),
 
   draw_key = draw_key_abline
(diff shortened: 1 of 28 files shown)


thomasp85:
I've gotten the visual tests up and running on my system now, and it seems that the issue is not as clear-cut. The prior `NA` default allowed colours to be specified with an alpha inside the colour string (eg #cc000044), while setting it to 1 will override any hard-coded transparency.

As this PR was brought in as a nicety for gganimate I suggest we roll it back. gganimate users will then have to specify alpha inside the layers if they want fading effects...

What do you think?

clauswilke:
Yes, please. I'd be very unhappy if I couldn't use RGBA colors anymore.

As an alternative to rolling this back, though, you could also modify the alpha function to apply its alpha value on top of any alpha already in the color.

thomasp85:
This was a sad oversight on my part, and I'm very much against breaking support for rgba.

I thought about modifying alpha in that way, but I think it would become too unpredictable - I'm doing a revert PR now, but open to making a change in `scales::alpha()` instead

clauswilke:
Yes, I guess the question is whether `scales::alpha()` sets or modifies the alpha value. What would be the expected outcome of `scales::alpha(scales::alpha("red", 0.5), 0.5)`?

thomasp85:
I think it needs to set it if an alpha scale should ever make sense...

Anyway, my vote is firmly on *revert* - I think I can solve my own needs with some changes to tweenr and gganimate

PR #2561, opened 2018-05-09
Fix CI builds

lionel-:
- On Travis add R 3.1 now that tidyr/tidyselect are fixed on 3.1
- On Appveyor use `ggplot_build()` instead of vdiffr until the latter is fixed on Windows platforms


hadley:
Thanks!

PR #2562, opened 2018-05-09
Compute overlapping groups by panel

### LINKED ISSUE #2493: Facetting and preserve = 'single' produces ever narrower bars.
This may be similar to #2290

When using `position_dodge(preserve='single')` to maintain the width of bars in a `geom_bar`, applying facetting results in the bars getting their 'single' width from the number of bars for each x value as a whole, rather than within the facet. 

Applying `preserve='total'` simply results in bars being wider than desired.

Hopefully the example below makes things clear.

``` r
library(ggplot2)
library(reshape2)
player <- c('Player1', 'Player1', 'Player2', 'Player2', 'Player3', 'Player3', 'Player4')
bool_val <- c('TRUE', 'FALSE', 'TRUE', 'FALSE', 'TRUE', 'FALSE', 'TRUE')
var1 <- c(1.5, 1.9, 2.7, 2.8, 3.0, 1.4, 1.7)
var2 <- c(1, 2, 3, 1.5, 1.2, 2.4, 2.2)
var3 <- c(1.7, 0.2, 1.3, 2.5, 2.2, 2.2, 3)

df <- data.frame(player, bool_val, var1, var2, var3)

long_df <- melt(df, id.vars=c('player', 'bool_val'))

ggplot(long_df, aes(x=variable, y=value, fill=bool_val)) + geom_bar(stat='identity', position=position_dodge2(preserve='total')) + facet_grid(player~., scales='free_x')
```

``` r

ggplot(long_df, aes(x=variable, y=value, fill=bool_val)) + geom_bar(stat='identity', position=position_dodge2(preserve='single')) + facet_grid(player~., scales='free_x')
```

hadley:
Fixes #2493 

@karawoo can you please double check my logic? I think this makes sense as we should only be considering the maximum number of overlapping groups within a panel.

### DIFF
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -35,12 +35,13 @@ PositionDodge2 <- ggproto("PositionDodge2", PositionDodge,
       panels <- unname(split(data, data$PANEL))
       if ("x" %in% names(data)) {
         # Point geom
-        ns <- vapply(panels, function(panel) max(table(panel$x)), double(1))
+        groups <- lapply(panels, function(panel) table(panel$x))
       } else {
         # Interval geom
-        ns <- vapply(panels, find_x_overlaps, double(1))
+        groups <- lapply(panels, find_x_overlaps)
       }
-      n <- max(ns)
+      n_groups <- vapply(groups, max, double(1))
+      n <- max(n_groups)
     }
 
     list(
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -31,10 +31,16 @@ PositionDodge2 <- ggproto("PositionDodge2", PositionDodge,
 
     if (identical(self$preserve, "total")) {
       n <- NULL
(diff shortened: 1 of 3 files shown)


R/position-dodge2.r
```diff
@@ -31,10 +31,16 @@ PositionDodge2 <- ggproto("PositionDodge2", PositionDodge,
 
     if (identical(self$preserve, "total")) {
       n <- NULL
-    } else if ("x" %in% names(data)){
-      n <- max(table(data$x))
     } else {
-      n <- max(table(find_x_overlaps(data)))
+      panels <- unname(split(data, data$PANEL))
+      if ("x" %in% names(data)) {
+        # Point geom
```
karawoo: This comment confused me for a minute as I thought it was only running for `geom_point()`

hadley:
Thanks for catching that - should be resolved now.

PR #2567, opened 2018-05-10
Fix issue 1546, fill aesthetic does not show in the legend guide for stat_summary

clauswilke:
I think the way to fix this issue is to give `geom_smooth()` a first-class `se` parameter that is considered when drawing the panel. This allows the geom and the legend drawing code to always be in synch. See resulting behavior in the examples below.

One note: Currently one regression test fails because the `geom_smooth()` and `draw_group()` functions have different default arguments and the test checks for that. I think the test is wrong in this particular case, but I wanted to bring this up for discussion. Relevant lines in the code, with explanation for why it should be this way:
https://github.com/clauswilke/ggplot2/blob/ed016144c97cf67424498453a2f4e06925fdba92/R/geom-smooth.r#L119-L126

Examples of `geom_smooth()` with `stat_summary()`:
``` r
library(ggplot2)
ggplot(mtcars, aes(cyl, mpg, fill = factor(am))) +
  geom_smooth(stat = "summary")
#> No summary function supplied, defaulting to `mean_se()
```

``` r

ggplot(mtcars, aes(cyl, mpg, fill = factor(am))) +
  geom_smooth(stat = "summary", se = FALSE)
#> No summary function supplied, defaulting to `mean_se()
```

``` r

# stat_summary doesn't have an `se` parameter, hence
# default is `se = FALSE`
ggplot(mtcars, aes(cyl, mpg, fill = factor(am))) +
  stat_summary(geom = "smooth")
#> No summary function supplied, defaulting to `mean_se()
```

``` r

ggplot(mtcars, aes(cyl, mpg, fill = factor(am))) +
  stat_summary(geom = "smooth", se = TRUE)
#> No summary function supplied, defaulting to `mean_se()
```

And, for completeness, all combinations of `geom_smooth()` / `stat_smooth()` with `se = TRUE` / `se = FALSE` work as expected:

``` r
library(ggplot2)
ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species, fill = Species)) +
  geom_smooth()
```

``` r

ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species, fill = Species)) +
  geom_smooth(se = FALSE)
```

``` r

ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species, fill = Species)) +
  stat_smooth()
```

``` r

ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species, fill = Species)) +
  stat_smooth(se = FALSE)
```

Created on 2018-05-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -265,6 +265,9 @@ up correct aspect ratio, and draws a graticule.
   (@davharris #1951). It also orders by the `x` aesthetic, making it easier 
   to pass pre-computed values without manual ordering (@izahn, #2028).
   It also now knows it has `ymin` and `ymax` aesthetics (#1939).
+  
+* Fixed legend-drawing bug when using `geom_smooth()` with stats other than
+  `stat_smooth()` (@clauswilke, #1546).
 
 * `geom_tile()` now once again interprets `width` and `height` correctly 
   (@malcolmbarrett, #2510)
--- a/R/geom-smooth.r
+++ b/R/geom-smooth.r
@@ -87,12 +87,12 @@ geom_smooth <- function(mapping = NULL, data = NULL,
 
   params <- list(
     na.rm = na.rm,
+    se = se,
     ...
   )
   if (identical(stat, "smooth")) {
     params$method <- method
     params$formula <- formula
-    params$se <- se
   }
 
   layer(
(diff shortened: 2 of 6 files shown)


hadley:
I think this approach is reasonable; the way parameters works is all a little hacky, so you should feel free to disable the test for this case (IIRC there are already a few other geoms where it's disabled)

clauswilke:
I've disabled the consistency check for `geom_smooth()`. (You were correct, there was already a list of such geoms.) I have added two simple visual tests, since there weren't any for `geom_smooth()` or `stat_smooth()` yet.

hadley:
Thanks!

PR #2569, opened 2018-05-11
Update authors

hadley:


lionel-:
Thanks! Are we removing non-maintainer email adresses in all packages?

hadley:
It's more that we're not adding them - if anything we should probably make a more systematic attempt to add ORCIDs

PR #2571, opened 2018-05-11
Copy edit NEWS.md

batpigandme:
* sf section now has a bullet, so that it's inline with everything else
* minor typo fixes
* consistent period at the end of each item AFTER `(@author)`


hadley:
Thanks!

At some point we should have a chat and attempt to clarify exactly what the ideal grammatical form of a news bullet is

PR #2572, opened 2018-05-11
Eliminate remotes

hadley:
* scales was historical; needed version already on CRAN
* dev rlang needed for `as_quosures()` and vector `parse_exprs()`

(Plus I accidentally included some smaller spurious changes 😬 )


lionel-:
If I'm not mistaken registration clashes are silent so the duplicated `quosures` method (when the next rlang version is out) shouldn't have adverse effects.

PR #2580, opened 2018-05-12
revdeps for 2.2.1 vs 2.2.1.9000

topepo:
The revdep process had to be restarted twice and I don't know if that impacted any package checks. I slacked @gaborcsardi about this.


gaborcsardi:
Restarting should not be a problem for revdepcheck.

hadley:
Thanks! Can you please also paste in the output from `revdepcheck::revdep_report_cran()`.

How long did this take you to run?

hadley:
Thanks @topepo

PR #2593, opened 2018-05-15
Overwrite vars with dplyr::vars, if available

### LINKED ISSUE #2585: Resolve dplyr::vars + ggplot2::vars clash
@lionel- can you think of some hack to avoid the dplyr::vars() + ggplot2::vars() namespace clash? (In an ideal world people wouldn't @import dplyr and @import ggplot2 but clearly there are ~100 packages that do this, and getting them all to change is going to be too much work)

hadley:
Fixes #2585


jimhester:
Oh, it is because this .onLoad overwrites the one you define conditionally in 
https://github.com/tidyverse/ggplot2/blob/f16978ed3f0603360ec0e9667116fd5d3046913b/R/backports.R#L9-L13

hadley:
Yes, duh, thanks!

PR #2600, opened 2018-05-15
Ensure that we don't try and assign scale to NULL

### LINKED ISSUE #2599: Don't know how to automatically pick scale for object of type NULL
cutpointr, drugCombo, ggeffects, HMP16SData, jtools, mousetrap, OutbreakTools, stormwindmodel, useful

``` r
library(ggplot2)

ggplot(mtcars, aes(mpg, wt)) + 
  geom_point(aes(shape = NULL))
#> Don't know how to automatically pick scale for object of type NULL. Defaulting to continuous.
#> Error: A continuous variable can not be mapped to shape
```

hadley:
Fixes #2599

@karawoo any ideas why this behaviour changed?

### DIFF
--- a/R/scale-type.R
+++ b/R/scale-type.R
@@ -1,4 +1,8 @@
-find_scale <- function(aes, x, env) {
+find_scale <- function(aes, x, env = parent.frame()) {
+  if (is.null(x)) {
+    return(NULL)
+  }
+
   type <- scale_type(x)
   candidates <- paste("scale", aes, type, sep = "_")
 
--- a/R/scales-.r
+++ b/R/scales-.r
@@ -97,6 +97,7 @@ scales_add_defaults <- function(scales, data, aesthetics, env) {
   if (is.null(new_aesthetics)) return()
 
   datacols <- lapply(aesthetics[new_aesthetics], rlang::eval_tidy, data = data)
+  datacols <- compact(datacols)
 
   for (aes in names(datacols)) {
     scales$add(find_scale(aes, datacols[[aes]], env))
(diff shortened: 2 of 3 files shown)


karawoo:
It looks like the change happened with the port to tidyeval, so maybe @lionel- can weigh in too. This specific error first occurs with 1f39e138391259ebef95e35e39fa1d3579f1984c. From the [first pass at tidyeval commit](https://github.com/tidyverse/ggplot2/commit/0da5bf3497f00f826dc78c4cabf1ba41fbe8f60f) until then there's a different error:

```r
ggplot(mtcars, aes(mpg, wt)) + 
  geom_point(aes(shape = NULL))
#> Error in if (length(ans) == 0L || as.character(ans[[1L]])[1L] == "~") { : 
#>   missing value where TRUE/FALSE needed
```

Before all the tidyeval changes, `find_scale()` was not being called for the shape aesthetic at all.

hadley:
I wonder if it's https://github.com/tidyverse/ggplot2/commit/1f39e138391259ebef95e35e39fa1d3579f1984c#diff-e01fbacefeedb8a626f9a0ec103d0e12L277

lionel-:
Also this gives the same results:

```r
x <- aes(shape = NULL)
x[["shape"]] <- NULL
str(x)
#> Named list()

x <- aes(shape = NULL)
x["shape"] <- NULL
str(x)
#> Named list()

x <- aes(shape = NULL)
x$shape <- NULL
str(x)
#> Named list()
```

lionel-:
The issue was that `plyr::tryapply()` strips NULL elements automatically. Compacting aesthetics after evaluation should fix the issue, cf last commit.

hadley:
Ah I knew `tryapply()` was there for some reason!!  (I'll try and tweak the unit tests if I have time)

PR #2601, opened 2018-05-15
Document new error on duplicate column names

thomasp85:


hadley:
I'm going to manually merge into breaking-changes branch so we can do all in one place. Sorry for the hassle!

PR #2606, opened 2018-05-15
Stricter test for theme$panel.ontop

hadley:
Fixes revdep failure in ggthemes


hadley:
And hurricaneexposure

batpigandme:
Should be good for jcolors, too.

Issue #2615, opened 2018-05-16
Setting axis limits now errors when they are a subset of the full set


cpsievert:
```r
library(ggplot2)
boxes <- ggplot(PlantGrowth, aes(x = group, y = weight)) + 
  geom_boxplot()

boxes
```

<img width="607" alt="screen shot 2018-05-16 at 3 10 07 pm" src="https://user-images.githubusercontent.com/1365941/40138634-3f440d70-591b-11e8-9ef8-7168b6068608.png">

```r
boxes + scale_x_discrete(limits = c("trt1", "ctrl"))
```

```r
Error in if (df$xmin[i] >= df$xmax[i - 1]) { : 
  missing value where TRUE/FALSE needed
```

Note that this example works (with a warning) on the current CRAN version:

<img width="604" alt="screen shot 2018-05-16 at 3 09 42 pm" src="https://user-images.githubusercontent.com/1365941/40138611-30f47e1c-591b-11e8-8b45-b893d3e6f463.png">

```r
Warning message:
Removed 10 rows containing non-finite values (stat_boxplot). 
```

PR #2616, opened 2018-05-16
stricter test in find_x_overlaps() (#2615)

cpsievert:


karawoo:
Thanks for looking into this @cpsievert! With this fix the box plot succeeds silently, whereas in the CRAN version it succeeds with a warning. @hadley do you think we should try to keep the warning behavior?

``` r
library("ggplot2") # CRAN
ggplot(PlantGrowth, aes(x = group, y = weight)) + 
  geom_boxplot() +
  scale_x_discrete(limits = c("trt1", "ctrl"))
#> Warning: Removed 10 rows containing non-finite values (stat_boxplot).
```



Created on 2018-05-16 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

hadley:
I think we should try and fix the warning - I think that's probably an alternative to @cpsievert's solution - i.e. the missing values should be removed before we get to dodge2.

karawoo:
Actually I was wrong, with this PR there's a warning about _1_ missing row, and that's in line with the behavior of `position = "dodge"` on master:

``` r
devtools::install_github("tidyverse/ggplot2", quiet = TRUE)
library("ggplot2")
ggplot(PlantGrowth, aes(x = group, y = weight)) + 
  geom_boxplot(position = "dodge") +
  scale_x_discrete(limits = c("trt1", "ctrl"))
#> Warning: Removed 1 rows containing missing values (geom_boxplot).
```



Created on 2018-05-16 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

This is still a bit odd because there are 10 rows of data that are removed, but this change in behavior isn't related to `find_x_overlaps()` so I think we should merge this fix.

However it turns up another issue we need to address separately, which is that the widths of the boxes is messed up 😭:

``` r
devtools::install_github("cpsievert/ggplot2", ref = "fix_x_overlaps", quiet = TRUE)
library("ggplot2")
ggplot(PlantGrowth, aes(x = group, y = weight)) + 
  geom_boxplot() +
  scale_x_discrete(limits = c("trt1", "ctrl"))
#> Warning: Removed 1 rows containing missing values (geom_boxplot).
```



Created on 2018-05-16 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

hadley:
I'm not sure that the fix is correct - what should `counter` be if `xmin` or `xmax` is NA?

karawoo:
Looked into this some more and I propose we add 
```r
data <- remove_missing(data, na.rm = FALSE, vars = "x", name = "geom_boxplot")
``` 

here:
https://github.com/tidyverse/ggplot2/blob/4b609fb22215ce497d22df9d4fd9c70b47d01bdc/R/stat-boxplot.r#L47-L50

Which results in:
1) plot succeeds with a warning about *10* removed rows (not 1), for both versions of position dodge
2) box widths are then correct (phew)

@hadley if you agree I can submit a PR and test tomorrow, unless @cpsievert wants to add it here instead.

hadley:
Yes, that sounds perfect!

PR #2623, opened 2018-05-17
use fortify.grouped_df() instead of ggplot.grouped_df()

### LINKED ISSUE #2578: Simplify the usage of fortify() in ggplot()
`ggplot.grouped_df()` was implemented in response to #2351. I feet, this is a kind of "fortifying" and at first thought implementing this was enough

```r
fortify.grouped_df <- function(data, ...) {
  data$.group <- dplyr::group_indices(data)
  data
}
```

so that the `ggplot.default()` calls this:

https://github.com/tidyverse/ggplot2/blob/4463da6abc9b8ff7a2451686b9685de36933b358/R/plot.r#L80

But, my guess was wrong. Since a `grouped_df` is also a `data.frame`, `ggplot.data.frame()` is actually dispatched, where `fortify()` won't happen.

I wonder if it is possible to remove `ggplot.data.frame()` and change `ggplot.default()` to bellow to let `fortify()` always happen:

``` r
ggplot.default <- function(...) {
  data <- fortify(data)

  # (the current implementation of ggplot.data.frame comes here)
}
```

IMHO, it is a bit beneficial to implement the "fortifying" process as a method of `fortify()` since then the process is more transparent to the users. If this looks ok, I will try to make a PR.

yutannihilation:
fixes #2578 

This PR removes `ggplot.data.frame()` and `ggplot.grouped_df()` so that all `data` passed to `ggplot()` get `fortify()`ed. Then, `fortify.grouped_df()` can take care of fortifying `grouped_df`s.

------

Note that one possible downside of this PR is this change of behaviour of `ggplot()`:

|     | before | after|
|:---|:-------|:-----|
| `data.frame` |  not fortified  |  **fortified**  |
| others         |   fortified  |  fortified   |

`fortify.data.frame()` returns the original `data.frame` as is, but some objects inheriting from `data.frame` might be fortified if it has its own method of `fortify()`.

But, I bet this will be rarely a problem; I cannot imagine the design where both the fortified version and unfortified version of the class are needed. If the method of `fortify()` is defined, it is probably supposed to be used.


hadley:
Thanks!

Issue #2535, opened 2018-04-27
Name conflict causes  stat_smooth to fail when method = "auto"


kylebmetrum:
The function `s` in the formula `y ~ s(x, bs = "cs")` is (I believe) `mgcv::s`, but that function isn't imported (mgcv is a Suggests).  The way this is set up, when there is a namespace conflict, no warning is issued (as expected) and `stat_smooth` fails. 

``` r

s <- function(...) "A"

library(ggplot2)

data(diamonds)

myplot <- ggplot(diamonds, aes(carat,price)) + geom_smooth()

myplot
#> `geom_smooth()` using method = 'gam'
#> Warning: Computation failed in `stat_smooth()`:
#> $ operator is invalid for atomic vectors
```

![](https://i.imgur.com/egLomak.png)

``` r

rm(s)

myplot
#> `geom_smooth()` using method = 'gam'
```

![](https://i.imgur.com/bNFBXmM.png)

``` r

s <- mgcv::s

myplot
#> `geom_smooth()` using method = 'gam'
```

![](https://i.imgur.com/b1bX6mi.png)

Created on 2018-04-18 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

PR #2626, opened 2018-05-17
fix geom_smooth() and add regression tests. Closes #2621.

### LINKED ISSUE #2621: geom_smooth throws error with diamonds dataset
This is a weird error I ran into:

```r
ggplot(diamonds, aes(x = carat, y = price)) + geom_smooth()
#> `geom_smooth()` using method = 'gam' and formula 'y ~ mgcv::s(x, bs = "cs")'
#> Warning message:
#> Computation failed in `stat_smooth()`:
#> invalid type (list) for variable 'mgcv::s(x, bs = "cs")' 
```

At first I thought it was because `diamonds` were a nibble but converting to a data.frame preserves the error...

This works on the CRAN version so it is a regression introduced somewhere. Strangely enough, it is tied to something in the diamonds dataset as geom_smooth works fine with e.g. mtcars

clauswilke:
This closes #2621 and reopens #2535. I don't know how to close both.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -338,10 +338,8 @@
   plotting thick arrows (@Ax3man, #774).
 
 * `geom_smooth()` now reports the formula used when `method = "auto"` 
-  (@davharris #1951). When mgcv is used, the `s` in `y ~ s(x, bs = "cs")` 
-  is now fully qualified as `mgcv::s()` to avoid name conflicts (@bfgray3, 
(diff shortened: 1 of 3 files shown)


hadley:
Can you update the NEWS too please?

hadley:
Thanks @clauswilke. We should definitely look at the test coverage and make sure more important parts of ggplot2 have unit tests.

PR #2631, opened 2018-05-18
fixed typo in README.md

dvcv:


batpigandme:
Hi David,

Thanks for the fix. 

Couple things:
1. The `README.md` is actually generated by the `README.Rmd`, so if you could move the edit there, it'll ensure that the change doesn't get overwritten!
2. I think the `you` might be appropriate _after_ the `get`, so if you wouldn't mind putting that in as well, I can merge, and we're good to go!

Mara

README.md
```diff
@@ -81,7 +81,7 @@ documentation pages. Currently, there are three good places to start:
     communication](http://r4ds.had.co.nz/graphics-for-communication.html)
     chapters in [R for data science](http://r4ds.had.co.nz). R for data
     science is designed to give you a comprehensive introduction to the
-    [tidyverse](http://tidyverse.org), and these two chapters will you
+    [tidyverse](http://tidyverse.org), and these two chapters will
     get up to speed with the essentials of ggplot2 as quickly as
```
batpigandme: Add "you" after get ⇨ "will get you up to speed"!

PR #2634, opened 2018-05-18
Copyedit testthat tests

pkq:
I'm not sure how useful or important this is, but I went ahead and edited the testthat tests to consistently use lower case to name tests, make test names grammatically correct, and apply consistent whitespace rules. Happy to make further revisions as needed.


pkq:
Glad to hear it!

PR #2638, opened 2018-05-19
Better error message when using +.gg() as unary operator

krlmlr:
The behavior in the dev version is much better than in the CRAN version, but I think we can be even clearer here.

## With this PR

``` r
library(ggplot2)
+ geom_point()
#> Error: There's nothing on the left-hand side of `+`. Did you forget to add this object to a ggplot object?
```

> Created on 2018-05-19 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

## With 582acfec

``` r
library(ggplot2)
+ geom_point()
#> Error: Cannot add ggproto objects together. Did you forget to add this object to a ggplot object?
```

> Created on 2018-05-19 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/R/plot-construction.r
+++ b/R/plot-construction.r
@@ -39,8 +39,8 @@
 #' base + list(subset(mpg, fl == "p"), geom_smooth())
 "+.gg" <- function(e1, e2) {
   if (missing(e2)) {
-    stop("There's nothing on the left-hand side of `+`. ",
-         "Did you forget to add this object to a ggplot object?",
+    stop("Cannot use `+.gg()` with a single argument. ",
+         "Did you accidentally put + on a new line?",
          call. = FALSE)
   }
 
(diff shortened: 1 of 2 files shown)


R/plot-construction.r
```diff
@@ -38,6 +38,12 @@
 #' # This can be useful to return from a function.
 #' base + list(subset(mpg, fl == "p"), geom_smooth())
 "+.gg" <- function(e1, e2) {
+  if (missing(e2)) {
+    stop("There's nothing on the left-hand side of `+`. ",
```
hadley: We need to tweak this slightly to follow the tidyverse error style guide (and to make the hint slightly more helpful). Maybe:

Can not use `+.gg` with a single argument. Did you accidentally put `+` on a new line?

krlmlr:
Done. Do we need to adapt the wording for the related error message?

hadley:
I don't think so.

Issue #2653, opened 2018-05-23
Unexpected behavior in element_text() horizontal justification


zlskidmore:
In the upcoming release of ggplot2 (2.3.0) the behavior of element_text() has changed. Specifically it looks like the justification anchors have changed in relation to the text. For Example:

In the current release:
![ggplot_2 2 1](https://user-images.githubusercontent.com/6145692/40440459-bdd8d95a-5e83-11e8-8ea7-0c9cb3b00cfb.png)

In the development version:
![ggplot2_2 3 0](https://user-images.githubusercontent.com/6145692/40440467-c8ecf876-5e83-11e8-9b1b-91ccbeabdb53.png)

```r
mtcars$name <- rownames(mtcars)

ggplot(mtcars, aes(x=name, y=carb)) + geom_bar(stat="identity") + theme(axis.text.x=element_text(angle=60, hjust=1))
```

karawoo:
Thanks for reporting this. I believe this bug is due to the change in how `xp` and `yp` are calculated in #2577.

karawoo:
@hadley @clauswilke I think we should revert [this change](https://github.com/tidyverse/ggplot2/pull/2577/commits/963b551a3d23198163a910ca8a9b169dbf0a2a36), it's going to break a lot of users' plots.

clauswilke:
@karawoo I'll look into it. However, I'm not yet convinced about the cause of this problem, because I checked for this particular issue when I made the patch, if I remember correctly. I think this has actually to do with changed grob heights due to fixed descenders.

A workaround exists, but I agree it's not intuitive (why `vjust = 0.65`?)
``` r
library(ggplot2)
mtcars$name <- rownames(mtcars)

ggplot(mtcars, aes(x=name, y=carb)) + 
  geom_bar(stat="identity") + theme(axis.text.x=element_text(angle=60, hjust=1, vjust = 0.65))
```

![](https://i.imgur.com/cp3dRcC.png)

Created on 2018-05-23 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

clauswilke:
When looking at this in debug mode, we see that for axis tick labels along the x axis the reference point travels vertically for different `vjust` values rather than orthogonally to the direction of the text. As far as I recall, this behavior is identical to the currently released ggplot2 version, but it's different from the behavior for `hjust` (which is along the text direction) and it's also different from `hjust` and `vjust` for axis titles, legend text and titles, and so on.

``` r
[...]

PR #2654, opened 2018-05-23
revert hjust/vjust rotation for cases with mixed justification models

clauswilke:
See #2653 for details.

### DIFF
--- a/R/margins.R
+++ b/R/margins.R
@@ -297,8 +297,14 @@ justify_grobs <- function(grobs, x = NULL, y = NULL, hjust = 0.5, vjust = 0.5,
 #'
 #' @noRd
 rotate_just <- function(angle, hjust, vjust) {
-  ## ideally we would like to do something like the following commented-out lines,
-  ## but it currently breaks things in certain cases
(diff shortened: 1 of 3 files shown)


clauswilke:
On further reflection, maybe I should just move the old piecewise rotation code into the `rotate_just()` function and be done with it for this release. We can try to tackle the problem more generally in a future update. Also, I think we need a visual regression test for x axis labels at an angle. That's the most common scenario where we need angles that deviate from 0, 90, 180, 270.

hadley:
Yeah, I think it makes sense to roll back for now. We can think it through more for the next release.

clauswilke:
Rollback done, and added visual test for rotated axis tick labels.

karawoo:
```
## but it currently breaks things in certain cases
```

Knowing that we and/or others will inevitably revisit this again in the future, could we add a little information (like a link to #2653) to remind ourselves of what the problem was?

clauswilke:
Done.

hadley:
@clauswilke you can feel free to squash-merge once I've reviewed

Issue #2505, opened 2018-03-29
geom_boxplot(outlier.size = NA) doesn't remove outliers after non-ggplot2 updates


mfoos:
Hi there, I have a Shiny server on an Amazon Linux EC2 (but this is also happening with slightly different specs on a Mac) and a coworker updated a smattering of R packages, and now ggplots that have worked (in Shiny) for years are no longer giving the expected output.

I usually overlay geom_point() with a jitter over geom_boxplot() and then hide the outliers so those points do not appear twice (the jitter means you can see both). Setting outlier.size = NA used to make them invisible, but since the update of doom, they still appear (and, oddly, larger than the points from geom_point). If I switch to outlier.size = NULL, they become very small but remain.

Unfortunately, and I apologize for nonetheless posting it here, it does not seem linked to an update of ggplot2 - I can run v2.2.1 under R 3.3.0 and not have this problem. I am posting it under ggplot2 because I can't locate the source.

``` r
library(tidyr)                                                                  
library(ggplot2)                                                                
                                                                                
set.seed(5)                                                                     
sample <- data.frame("id" = 1:100,                                              
"x" = runif(100, 5, 10),                                                        
"y" = rnorm(100, mean = 7),                                                     
"group" = sample(c("A", "B"), 100, replace = TRUE)) %>% gather(key, value, x, y)
                                                                                
ggplot(sample, aes(x = key, y = value)) +                                       
geom_boxplot(outlier.size = NA) +                                               
geom_point()                                                                    
```

![](https://i.imgur.com/2nUo4F7.png)

``` r
                                                                                
sessionInfo()                                                                   
#> R version 3.4.1 (2017-06-30)
#> Platform: x86_64-redhat-linux-gnu (64-bit)
#> Running under: Amazon Linux AMI 2017.09
#> 
#> Matrix products: default
#> BLAS/LAPACK: /usr/lib64/R/lib/libRblas.so
#> 
#> locale:
#>  [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
#>  [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
#>  [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
#>  [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
#>  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[...]

PR #2658, opened 2018-05-25
issue #2505, hiding boxplot outliers

clauswilke:
Since there are already at least 4-5 different ways of hiding  outliers, I don't see the need to add a new parameter to the API. Instead, I think updating the documentation is sufficient. I didn't list `outlier.size = -1`, because that option is only guaranteed to work if the stroke value hasn't been modified.

I also fixed the problem where `geom_point()` was trying to match an empty string against the shape names.

### DIFF
--- a/R/geom-boxplot.r
+++ b/R/geom-boxplot.r
@@ -39,10 +39,10 @@
 #'   US spelling will take precedence.
 #'
 #'   Sometimes it can be useful to hide the outliers, for example when overlaying
-#'   the raw data points on top of the boxplot. Hiding the outliers can be achieved in numerous ways,
-#'   including by setting `outlier.colour = NA`, `outlier.colour = "transparent"`,
-#'   `outlier.shape = NA`, or `outlier.shape = ""`. All these options are equivalent and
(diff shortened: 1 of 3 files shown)


hadley:
I'm with @karawoo; I think we should just pick one "official" way and describe that.

smouksassi:
the effects of the various ways  will be different if default shape were  not used consider
```r
  geom_boxplot(
    outlier.color = "red",
    outlier.shape = 22,
    outlier.colour = "transparent",
    outlier.fill = "green"
  )
```
(note that ggplot will only compain if there is two outlier.color but no warning if we have outlier.color and outlier.colour)

outlier.shape = NA  seems the safest option and would prevent having "ghost transparent points" when exporting into other formats

clauswilke:
@smouksassi I agree. The same reasoning was why I didn't want to include the `outlier.size` option. I have updated the PR so it now only mentions `outlier.shape = NA`. I also now explicitly state that the range of the scale is not modified by hiding outliers.

clauswilke:
@hadley Any remaining concerns with the PR in it current form?

PR #2663, opened 2018-05-27
Add documentation about `labs(tag = )`

yutannihilation:
`tag` argument seems not documented in `?labs` yet. Since `tag` is intended to be specified by `labs()`, I think this should be described here.

### DIFF
--- a/R/labels.r
+++ b/R/labels.r
@@ -22,7 +22,7 @@ update_labels <- function(p, labels) {
 #' audience. Ensure the axis and legend labels display the full variable name.
 #' Use the plot `title` and `subtitle` to explain the main findings.
 #' It's common to use the `caption` to provide information about the
-#' data source.
+#' data source. `tag` can be used for adding identification tags.
 #'
 #' You can also set axis and legend labels in the individual scales (using
 #' the first argument, the `name`. I recommend doing that if you're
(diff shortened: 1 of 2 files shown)


yutannihilation:
Thanks! Of course the position is changeable, but "by default" is not added to other labels like `title`, whose positions are also changeable. Does adding "by default" only to `tag` make sense?

clauswilke:
Rereading it all in context, I'm fine with it as is.

One other small issue I noticed, however: In the "Details" section, there's a parenthesis that isn't closed: "You can also set axis and legend labels in the individual scales (using the first argument, the `name`." Would you mind closing it?

yutannihilation:
Sure, will do later.

yutannihilation:
Done!

hadley:
@yutannihilation if you wanted to make it even better, I think `labs()` could use named arguments for `title`, `caption` etc. That'll make the function easier to use, and the documentation more clear.

yutannihilation:
@hadley Thanks, agreed. I've wondered why they are not named args...🙄 Do you mean I should create another PR for this? If so, I'll try tomorrow. But, if you will do, I'm ok.

hadley:
A PR would be awesome 😄 (and no rush, since this shouldn't affect any code; although it might be useful to have some tests to verify that)

yutannihilation:
Sure!

PR #2664, opened 2018-05-27
Do not use ordered factor for viridis example

yutannihilation:
Since viridis is already the default colour scale for ordered factors, these first two examples in http://ggplot2.tidyverse.org/reference/scale_viridis.html#examples make no differences.

``` r
dsamp <- diamonds[sample(nrow(diamonds), 1000), ]
(d <- ggplot(dsamp, aes(carat, price)) +
  geom_point(aes(colour = clarity)))
```

``` r
d + scale_colour_viridis_d()
```

Maybe it is proper to use some other data that has factor or character columns, but using `diamond` seems the widespread manner among examples of `scale_colour_*()`. So, I choose converting `clarity` column.

### DIFF
--- a/R/scale-viridis.r
+++ b/R/scale-viridis.r
@@ -19,6 +19,13 @@
 #' @export
 #' @examples
 #' dsamp <- diamonds[sample(nrow(diamonds), 1000), ]
+#'
+#' # viridis is the default colour/fill scale for ordered factors
+#' ggplot(dsamp, aes(carat, price)) +
+#'   geom_point(aes(colour = clarity))
+#'
+#' # Usse viridis_d with discrete data
+#' dsamp$clarity <- factor(dsamp$clarity, ordered = FALSE)
 #' (d <- ggplot(dsamp, aes(carat, price)) +
 #'   geom_point(aes(colour = clarity)))
 #' d + scale_colour_viridis_d()
--- a/R/scale-viridis.r
+++ b/R/scale-viridis.r
@@ -18,39 +18,36 @@
 #' @rdname scale_viridis
 #' @export
 #' @examples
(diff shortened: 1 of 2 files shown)


hadley:
Maybe it would be better to convert the whole example to use a different datasets?

yutannihilation:
By the way, is this behaviour documented anywhere else than NEWS? I didn't know even the fact that the scale for ordered factors and that for unordered are different...

hadley:
It's not, but I'm not sure where it should be. How do you learn that numeric and factors get different scales?

yutannihilation:
Hmm, in my case, I think I've learned it when I saw the error message `Error: Continuous value supplied to discrete scale`. So, if I didn't read the news, I bet I wouldn't learn about `scale_*_ordinal()` until I see this warning 🙄 

https://github.com/tidyverse/ggplot2/blob/00ecd3670ef0f1c195bf4c6b5ada3b1895712f1c/R/scale-shape.r#L50

dpseidel:
New discrete examples use a subset of `txhousing` throughout.  Continuous examples still use `faithfuld`. I left a single `diamonds` plot to document the default behavior for ordered factors but that could be dropped.

yutannihilation:
@dpsiedel Thanks! I think I'm no more the author of this pull request. Should I close this so that you can create a new one...?

dpseidel:
@yutannihilation, I have no preference. I am currently working my way through a variety of documentation issues for `ggplot2` and figured it was easier and cleaner to update your existing PR. 😄  

Perhaps @hadley has an opinion?

yutannihilation:
OK, if it is convenient for you to update this, I'm fine! Again, thanks for taking this over, which I couldn't come up with the nice fix. 👍

hadley:
Thanks!

Issue #2446, opened 2018-02-11
ggtitle overwrites subtitle


ptoche:
Not a big deal: ``ggtitle()`` overwrites ``subtitle``

    ggplot(data=mtcars, aes(x=mpg, y=cyl)) + geom_point() + labs(title="title", subtitle="subtitle")
    p = last_plot()
    p + ggtitle("title2")  # overwrites subtitle
    p + labs(title="title2")  # does not overwrite subtitle

Below I tried to use ``reprex(clipr::read_clip())``, but it doesn't seem to work too well...

#> [1] "library(ggplot2)"                                                                                    
#> [2] "ggplot(data=mtcars, aes(x=mpg, y=cyl)) + geom_point() + labs(title=\"title\", subtitle=\"subtitle\")"
#> [3] "p = last_plot()"                                                                                     
#> [4] "p + ggtitle(\"title2\")  # overwrites subtitle"                                                      
#> [5] "p + labs(title=\"title2\")  # does not overwrite subtitle"

DarioBoh:
`ggtitle` default for the argument `subtitle` is `NULL`

batpigandme:
As @DarioBoh mentioned, the default arg in ggtitle for subtitle is NULL (it's even in bold [in the docs](http://ggplot2.tidyverse.org/reference/labs.html) for it 😉). So, you're basically overwriting the whole "title" entity (subtitle absence and all) in the second plot.

``` r
library(ggplot2)
ggplot(data=mtcars, aes(x=mpg, y=cyl)) + geom_point() + labs(title="title", subtitle="subtitle")
```

![](https://i.imgur.com/gVlvL7k.png)

``` r
p <- last_plot()
p + ggtitle("title2")  # overwrites subtitle
```

![](https://i.imgur.com/60eXzhS.png)

``` r
p + ggtitle("title2", subtitle = "subtitle")  # overwrites subtitle
```

![](https://i.imgur.com/RBGbJYT.png)

``` r
p + labs(title="title2")  # does not overwrite subtitle
```
![](https://i.imgur.com/BvdWCe5.png)

Created on 2018-02-12 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

ptoche:
[...]

PR #2669, opened 2018-05-29
Make title, subtitle, caption and tag as labs()'s named arguments

yutannihilation:
As suggested in https://github.com/tidyverse/ggplot2/pull/2663#issuecomment-392542114, making them as the named arguments let us organize the document easier.

Note that one difficulty I found is that `NULL` should be distinguished from missing; `NULL` indicates the user's intent to remove the label. For example, the code below doesn't show the title:

``` r
ggplot() +
  labs(title = "foo") +
  labs(title = NULL)
```

So, I chose to use `missing()`. But, if there is some better way to do this with rlang, please point it out.

Besides, I slightly suspect `ggtitle()` may not do the right thing; the code below doesn't show the subtitle since `ggtitle()` sets `subtitle` to `NULL`. Is this valid for the user's intent?

``` r
ggplot() +
  labs(subtitle = "bar") +
  ggtitle(title = "foo")
```

If this should also be changed, I'll incorporate in this PR as well.


ptoche:
See also: [](https://github.com/tidyverse/ggplot2/issues/2446)

yutannihilation:
Oh, thanks. If that is intended, I'm ok with it.

ptoche:
I don't like it either.

hadley:
I think we should make `labs()` and `ggtitle()` work in the same way, i.e. using missing arguments and don't override without explicit `NULL`.

R/labels.r
```diff
@@ -51,10 +55,20 @@ update_labels <- function(p, labels) {
...
+#' # If you want to remove a label, set it to NULL.
+#' p + labs(title = "title") + labs(title = NULL)
+labs <- function(..., title, subtitle, caption, tag) {
   args <- list(...)
-  if (is.list(args[[1]])) args <- args[[1]]
+  if (length(args) > 0 && is.list(args[[1]])) args <- args[[1]]
   args <- rename_aes(args)
+
+  # Since NULL should be preserved, we should wrap args with list()
+  if (!missing(title)) args["title"] <- list(title)
```
hadley: @lionel- can you think of a more elegant way to write this?

lionel-: Perhaps remove the named arguments and let users pass them through `...`?

lionel-: Hmm I can't think of anything better.

yutannihilation: Thanks, then I leave this as is :)

yutannihilation:
> I think we should make labs() and ggtitle() work in the same way

I'm OK with either. But, as @batpigandme pointed on #2446, this behavior of `ggtitle()` is documented. So, I rather hesitate to make a tiny breaking change on this, though it seems negligible.

Maybe is it better to leave this as is and encourage the users to use `labs()` (and possibly to soft-deprecate `ggtitle()`?).

clauswilke:
I disagree with @batpigandme's reading of the documentation. The current documentation says "Leave `NULL` for no subtitle." That's a far cry from "If you leave the subtitle as `NULL` then your pre-existing subtitle will be deleted."

I'm with @hadley. Don't override without explicit `NULL`.  To be absolutely clear about how things work, we could add the following to the description:

> If a plot already has a title, subtitle, caption, etc., and you want to remove it, you can do so by setting the respective argument to `NULL`. For example, if plot `p` has a subtitle, then `p +  labs(subtitle = NULL)` will remove the subtitle from that plot.

I also think the description for `label` is confusing. It currently says "The text for the axis, plot title or caption below the plot", but there is no function where `label` sets the caption. How about:

> The title of the respective axis (for `xlab()` or `ylab()`) or of the plot (for `ggtitle()`).

yutannihilation:
Thanks! Sounds reasonable to me.

yutannihilation:
I see. There's a trade off. For here, I want to choose missingness for better documentation.

yutannihilation:
Sorry, I'm not sure what is the good way to pass missingness to other function. I'll try, but any advice will be appreciated!

yutannihilation:
Done. Thanks @clauswilke for the sentences.

> I'm not sure what is the good way to pass missingness to other function

In this case, I could take the advantage of the fact that `labs()` can take a list as its first argument.

lionel-:
To pass missingness programmatically you need some meta-programming:

```r
with_qq <- function(expr) {
  expr <- enexpr(expr)
  eval_bare(expr, caller_env())
}

f <- function(x) {
  missing(x)
}
wrapper <- function(x) {
  with_qq(f(!!maybe_missing(x)))
}

wrapper()
#> [1] TRUE

wrapper(1)
#> [1] FALSE
```

yutannihilation:
Thanks! `maybe_missing()` seems interesting.

batpigandme:
> I disagree with @batpigandme's reading of the documentation. The current documentation says "Leave NULL for no subtitle." That's a far cry from "If you leave the subtitle as NULL then your pre-existing subtitle will be deleted."

^^ @clauswilke this makes sense to me. 👍

hadley:
There's also `waiver()` which is used elsewhere in ggplot2 for a similar task.

thomasp85:
`waiver` as default is def the way to go IMO

yutannihilation:
Ah, I forgot about `waiver()`. Thanks!

R/labels.r
```diff
@@ -51,10 +61,17 @@ update_labels <- function(p, labels) {
...
 #' p + labs(title = "title", tag = "A")
-labs <- function(...) {
-  args <- list(...)
-  if (is.list(args[[1]])) args <- args[[1]]
+#'
+#' # If you want to remove a label, set it to NULL.
+#' p + labs(title = "title") + labs(title = NULL)
+labs <- function(..., title = waiver(), subtitle = waiver(), caption = waiver(), tag = waiver()) {
+  args <- list(..., title = title, subtitle = subtitle, caption = caption, tag = tag)
+  if (length(args) > 0 && is.list(args[[1]]) && !is.waive(args[[1]])) args <- args[[1]]
```
yutannihilation: By the way, why should `labs()` accept a list as its first argument?
I couldn't find this is documented or used elsewhere.

thomasp85:
My guess is to avoid the `do.call` need if you programmatically collect the labels in a list prior to plot construction

yutannihilation:
Thanks, if this is useful for some people, I leave this as is.

hadley:
I’d be happy to remove that feature if it doesn’t break existing code (but it probably will)

yutannihilation:
OK. I just hoped no one knows this feature because it doesn't used internally in ggplot2 and no documents seems to advertise it. But, I agree with your concern.

Then, I think this is ready now.

R/labels.r
hadley: Now that ggplot2 uses tidy eval, I think we should remove this, and in the previous line use `rlang::list2()` instead of `list()`. That way you can use `!!!` if you do already have a list of labels.

tests/testthat/test-labels.r
```diff
@@ -25,6 +25,24 @@ test_that("setting guide labels works", {
...
+
+    # labs() can take a list as the first argument
+    expect_identical(labs(list(title = "my title", tag = "A)"))$tag, "A)")
+
+    # NULL is preserved
+    expect_equivalent(labs(title = NULL), list(title = NULL))
+
+    # ggtitle works in the same way as labs()
+    expect_identical(ggtitle("my title")$title, "my title")
+    expect_identical(ggtitle("my title",
```
hadley: Can you please use tidyverse style here? http://style.tidyverse.org/syntax.html#long-lines

yutannihilation:
Sorry, I'm yet to figure out how to squash the duplicated arguments properly... Will try tomorrow or after.

yutannihilation:
@hadley 
I think I've implemented what you suggested. But..., I'm wondering if this is OK in terms of the compatibility with the existing codes, as you said:

> I’d be happy to remove that feature if it doesn’t break existing code (but it probably will)

Do I get something wrong?

``` r
library(ggplot2)

labs(title = "test", tag = "A")
#> $title
#> [1] "test"
#> 
#> $tag
#> [1] "A"
#> 
#> attr(,"class")
#> [1] "labels"

# NULL is preserved
labs(title = NULL)
#> $title
#> NULL
#> 
#> attr(,"class")
#> [1] "labels"

# !!! can be used in labs()
labs(!!!list(title = "test"))
#> $title
#> [1] "test"
#> 
#> attr(,"class")
#> [1] "labels"

# If the argument is duplicated, the first one will be used
labs(!!!list(title = "test"), title = "TEST")
#> $title
#> [1] "test"
#> 
#> attr(,"class")
#> [1] "labels"
```

Created on 2018-07-15 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

hadley:
I think this looks good. Can you please now prepare a news bullet?

yutannihilation:
I resolved the conflict. Could someone review the news bullet?

clauswilke:
@batpigandme Do we have a tidyverse rule for the Oxford comma? I would probably write
```
`labs()` now has named arguments `title`, `subtitle`, `caption`, and `tag`. 
```
and delete "explicitly".

@yutannihilation Note that any further commits will cause a broken build on Travis until #2838 is resolved. You may want to wait until that time and then rebase.

smouksassi:
Sent from my BlackBerry 10 smartphone.,'                                                                                                                                                                                                                From: Mara AverickSent: Sunday, August 19, 2018 5:20 PMTo: tidyverse/ggplot2Reply To: tidyverse/ggplot2Cc: SubscribedSubject: Re: [tidyverse/ggplot2] Make title, subtitle, caption and tag a'''s labs()'s named arguments (#2669)@clauswilke we are pro-Oxford-comma!

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/tidyverse/ggplot2","title":"tidyverse/ggplot2","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/email/message_cards/header.png","avatar_image_url":"https://assets-cdn.github.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/tidyverse/ggplot2"}},"updates":{"snippets":[{"icon":"PERSON","message":"@batpigandme in #2669: @clauswilke we are pro-Oxford-comma!"}],"action":{"name":"View Pull Request","url":"https://github.com/tidyverse/ggplot2/pull/2669#issuecomment-414156779"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/tidyverse/ggplot2/pull/2669#issuecomment-414156779",
"url": "https://github.com/tidyverse/ggplot2/pull/2669#issuecomment-414156779",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
},
{
"@type": "MessageCard",
"@context": "http://schema.org/extensions",
"hideOriginalBody": "false",
"originator": "AF6C5A86-E920-430C-9C59-A73278B5EFEB",
"title": "Re: [tidyverse/ggplot2] Make title, subtitle, caption and tag as labs()'s named arguments (#2669)",
"sections": [
{
"text": "",
"activityTitle": "**Mara Averick**",
"activityImage": "https://assets-cdn.github.com/images/email/message_cards/avatar.png",
"activitySubtitle": "@batpigandme",
"facts": [

]
}
],
"potentialAction": [
{
"name": "Add a comment",
"@type": "ActionCard",
"inputs": [
{
"isMultiLine": true,
"@type": "TextInput",
"id": "IssueComment",
"isRequired": false
}
],
"actions": [
{
"name": "Comment",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"IssueComment\",\n\"repositoryFullName\": \"tidyverse/ggplot2\",\n\"issueId\": 2669,\n\"IssueComment\": \"{{IssueComment.value}}\"\n}"
}
]
},
{
"name": "Close pull request",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"PullRequestClose\",\n\"repositoryFullName\": \"tidyverse/ggplot2\",\n\"pullRequestId\": 2669\n}"
},
{
"targets": [
{
"os": "default",
"uri": "https://github.com/tidyverse/ggplot2/pull/2669#issuecomment-414156779"
}
],
"@type": "OpenUri",
"name": "View on GitHub"
},
{
"name": "Unsubscribe",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"MuteNotification\",\n\"threadId\": 339917471\n}"
}
],
"themeColor": "26292E"
}
]

yutannihilation:
Resolved a conflict.

yutannihilation:
@hadley @clauswilke 
If the news bullet seems good to you, could you approve and merge this? I'm sorry to have make this discussion unnecessarily complicated and long...

yutannihilation:
Thanks!

PR #2670, opened 2018-05-29
Support size and linetype in geom_hex (#2488)

### LINKED ISSUE #2488: geom_hex ignores size aesthetic/parameter
geom_hex defines a size default of 0.5, which by analogy with other filled geoms I would expect to control line weight (e.g. geom_bin2d, see bottom).

But it never gets passed to `hexGrob` so there is no change in the output size.

There's also an interesting inconsistency which may or may not be issue-worthy: geom_hex claims size as an aesthetic (in docs), whereas geom_bin2d does not, and for geom_bin2d size does indeed not work as an a aesthetic, only as a parameter, it seems.

``` r
library(ggplot2)
p <- ggplot(diamonds, aes(carat, price))

# This doesn't work as expected
p + geom_hex(size = 0.1, color = "red")
```

``` r
p + geom_hex(size = 2, color = "red")
```

``` r
p + geom_hex(aes(size = cut), color = "red")
#> Warning: Using size for a discrete variable is not advised.
```

``` r

# This does work as expected
p + geom_bin2d(size = 0.1, color = "red")
```

``` r
p + geom_bin2d(size = 2, color = "red")
```

mikmart:
Fixes #2488: `size` (and additionally `linetype`) can now be used in `geom_hex()` to control the bordering lines of the hexagons.

```
ggplot(diamonds, aes(carat, price)) +
  geom_hex(col = "red", size = 2)
```


hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Also can you please use `Fixes #2488` — that specific wording will cause the related issue to be automatically closed when this PR is merged.

mikmart:
Sure! I've updated the PR comment to use `Fixes`, and also added the NEWS bullet (hopefully correctly).

This is my first PR, and I'm a bit new to Git, so hopefully the NEWS bullet addition is OK; I had some trouble with incorporating all the other commits that happened since I opened the PR, (tried to rebase-pull-push, but that ended up with 30 commits in this branch somehow -- forced a reset on that), so I ended up just doing a deliberate merge conflict to be able to add the bullet without too much of a mess.

hadley:
Looks good, thanks!

PR #2672, opened 2018-05-30
In qplot, don't require I() for parameters. Closes #2671

### LINKED ISSUE #2671: qplot ignores binwidth parameter
With ggplot2 2.2.1.9000, the `binwidth` parameter to `qplot` has no effect:

``` r
library(ggplot2)
qplot(iris$Sepal.Length, binwidth=1)
```

clauswilke:
This closes issue #2671.

I think it's sufficient to treat everything as a parameter that is not in the complete list of aesthetics. This is a heuristic that is not guaranteed to be always right, but it is going to be mostly right. And in any case, it's the heuristic that was used prior to 2.3.0.


clauswilke:
@hadley, I assume that you clicking "approve" means I can merge. Let me know if I misunderstood.

PR #2677, opened 2018-05-31
Improve aes() news and documentation

### LINKED ISSUE #2676: Not all items in the list aes() produces are quosures
https://github.com/tidyverse/ggplot2/blob/master/NEWS.md#ggplot2-2219000 claims 

> The primary developer facing change is that aes() is now a list of quosures (expression + environment pairs) rather than a list of symbols, and you'll need to take a different approach to extracting the information you need.

But not all elements are quosures:

```r
aes(x = foo, y = 0) %>% str()
## List of 2
##  $ x: language ~foo
##   ..- attr(*, ".Environment")=<environment: R_GlobalEnv> 
##  $ y: num 0
##  - attr(*, "class")= chr "uneval"
aes(x = foo, y = 0) %>% lapply(class)
## $x
## [1] "quosure" "formula"
## 
## $y
## [1] "numeric"
## 
aes(x = foo, y = 0) %>% lapply(function(x) inherits(x, "quosure"))
## $x
## [1] TRUE
## 
## $y
## [1] FALSE
```

hadley:
Fixes #2676

### DIFF
--- a/R/aes.r
+++ b/R/aes.r
@@ -48,6 +48,8 @@ NULL
 #'   they are so common; all other aesthetics must be named.
 #' @seealso [vars()] for another quoting function designed for
 #'   faceting specifications.
+#' @return A list with class `uneval`. Components of the list are either
+#'   quosures or constants.
 #' @export
 #' @examples
 #' aes(x = mpg, y = wt)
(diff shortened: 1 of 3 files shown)


R/aes.r
```diff
@@ -56,32 +58,36 @@ NULL
...
 #' ggplot(mpg) + geom_point(aes(displ, hwy))
 #'
-#' # Aesthetics supplied to ggplot() are used as defaults for every layer
-#' # you can override them, or supply different aesthetics for each layer
-#'
-#'
-#' # aes() is a quoting function, so you need to use tidy evaluation
-#' # techniques to create wrappers around ggplot2 pipelines. The
+#' # Tidy evaluation ----------------------------------------------------
+#' # aes() automated quotes all its arguments, so you need to use tidy
```
lionel-: automatically?

Issue #2679, opened 2018-06-01
Feature request: Scaled densities/counts in 2d density/bins plots.


bjreisman:
2d density plots are one of the most common data-visualizations used to display flow cytometry data, and the `geom_bin2d` and `geom_hex` and `geom_density_2d` geoms are excellent for making these plots. However, when facetting 2d density plots, there isn't a straightforward way to set the scale such that the highest point of each plot is the same -  the convention in my field. 

geom_density (the 1d version) has an option to set `fill = ..scaled..` which rescales all groups to a maximum of 1. Would you consider adding such an option to `geom_bin2d` and `geom_hex` (and `stat_bin2d`/`stat_binhex`)?

```{r}
library(ggplot2)
library(dplyr)
library(viridis)

#2d density plot colored by level
plot.density2d_level <- diamonds %>%
  ggplot(aes(x=x, y= depth)) + 
  stat_density_2d(aes(fill = stat(level)),
                  geom = "polygon", 
                  n = 100 ,
                  bins = 10) + 
  facet_wrap(clarity~.) + 
  scale_fill_viridis(option = "A")

plot.density2d_level
```
![plot1](https://user-images.githubusercontent.com/22528952/40845166-7a5f8116-657b-11e8-8345-cf059c4a2eae.png)

As you can see, its hard to tell where the area of highest density is in all but VVS1 and IF. 

However, if I render the first panel alone, the area of highest density is very clear.

```{r}
diamonds %>%
  filter(clarity == "I1") %>%
  ggplot(aes(x=x, y= depth)) + 
  stat_density_2d(aes(fill = stat(level)),
                  geom = "polygon", 
                  n = 100 ,
                  bins = 10) + 
  facet_wrap(clarity~.) + 
  scale_fill_viridis(option = "A")
```
![plot2](https://user-images.githubusercontent.com/22528952/40845233-b844b3b6-657b-11e8-8507-a2688653b52d.png)

If I set the fill to `stat(piece)` then I'm closer to achieving the right scaling; though there are still some panels which are not using the full color palette. (This has been my temporary solution.)

```{r}
plot.denisty2d_piece <- diamonds %>%
  ggplot(aes(x=x, y= depth)) + 
  stat_density_2d(aes(fill = stat(piece)),
                  geom = "polygon", 
                  n = 100, 
                  bins = 10, 
                  contour = T) + 
  facet_wrap(clarity~.) + 
  scale_fill_viridis(option = "A")
[...]

PR #2680, opened 2018-06-01
Scaled densities/counts in 2d density/bins plots (#2679)

bjreisman:
I added 'scaled' statistics to the `stat_bin2d`, `stat_binhex`, ` stat_density_2d`, and  `stat_contour` functions for plotting 2d distributions normalized to a common height. This would be very useful for faceted 2d plots, where the maximum density/count can vary greatly between panels. 

(This is my attempt to add the feature I had requested in issue #2679.)

---
I tried my best to mirror the corresponding statistics from the 1d functions:

stat_bin | stat_bin2d | stat_binhex |  stat_bin2d (updated) | stat_binhex (updated) 
--- | --- | --- | ---| ---
`count `| `count` | `count` | `count` | `count` 
`ncount` | | | `ncount` | `ncount`
`density`|  `density` | `density` | `density` | `density`
`ndensity`|   | | `ndensity` | `ndensity`

It was a little bit harder to get things to match-up one-to-one for the density based function, so some adjustment to the syntax may be needed. 

stat_density| stat__contour | stat_density2d  | stat__contour (updated) | stat_density2d (updated)
--- | --- | --- | ---| ---
`count` | - | - | - |
`density` | - | `density` | - | `density`
`scaled` | - | - | - | `scaled`
-|`pieces`| - |`pieces`| -
 -|`level`| - |`level`| -
 -| - | - | `nlevel` | -  

---

Here is an example of the revised functions in action:
```{r}
library(ggplot2)
library(dplyr)
library(viridis)

ggplot(diamonds, aes(x=x, y= depth)) +
  stat_density_2d(aes(fill = stat(level)),
                  geom = "polygon",
                  n = 100,
                  bins = 10,
    contour = T) +
  facet_wrap(clarity~.) +
  scale_fill_viridis(option = "A")
```
plot6

```{r}
ggplot(diamonds, aes(x=x, y= depth)) +
  stat_density_2d(aes(fill = stat(nlevel)),
                  geom = "polygon",
                  n = 100,
                  bins = 10,
    contour = T) +
  facet_wrap(clarity~.) +
  scale_fill_viridis(option = "A")
```
plot7

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -167,6 +167,17 @@
 
 ### Layers: geoms, stats, and position adjustments
 
+* `stat_contour()`, `stat_density2d()`, `stat_bin2d()`,  `stat_binhex()` now
+  include  a normalized version the `level`, `density`, and `count` parameters, 
+  named `nlevel`, `ndensity`, and `ncount`. This mirrors the use of `ndensity`
+  in `stat_density()` and `ncount`/`ndensity` in `stat_bin()`. These stats
+  should be useful for faceted 2D density plots and bin plots in cases where
+  the distribution of density within a facet is more relevant than the absolute
+  density or count across facets. (@bjreisman, #2680)
+
+* `stat_density()` now includes the calculated statistic `nlevel`, an alias 
+  for `scaled`, to better match the syntax of `stat_bin` (@bjreisman, #2680)
+
 * `geom_segment()` and `geom_curve()` have a new `arrow.fill` parameter which 
   allows you to specify a separate fill colour for closed arrowheads 
   (@hrbrmstr and @clauswilke, #2375).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # ggplot2 3.0.0.9000
 
+*   `stat_contour()`, `stat_density2d()`, `stat_bin2d()`,  `stat_binhex()`
+    now calculate normalized statistics including `nlevel`, `ndensity`, and
+    `ncount`. Also, `stat_density()` now includes the calculated statistic 
+    `nlevel`, an aliasfor `scaled`, to better match the syntax of `stat_bin()`
+    (@bjreisman, #2679).
+
 *  `geom_hex()` now understands the `size` and `linetype` aesthetics
    (@mikmart, #2488).
   
@@ -166,17 +172,6 @@
   `plot.tag.position` theme setting (@thomasp85).
 
 ### Layers: geoms, stats, and position adjustments
-
-* `stat_contour()`, `stat_density2d()`, `stat_bin2d()`,  `stat_binhex()` now
-  include  a normalized version the `level`, `density`, and `count` parameters, 
-  named `nlevel`, `ndensity`, and `ncount`. This mirrors the use of `ndensity`
-  in `stat_density()` and `ncount`/`ndensity` in `stat_bin()`. These stats
-  should be useful for faceted 2D density plots and bin plots in cases where
-  the distribution of density within a facet is more relevant than the absolute
-  density or count across facets. (@bjreisman, #2680)
-
-* `stat_density()` now includes the calculated statistic `nlevel`, an alias 
-  for `scaled`, to better match the syntax of `stat_bin` (@bjreisman, #2680)
 
 * `geom_segment()` and `geom_curve()` have a new `arrow.fill` parameter which 
   allows you to specify a separate fill colour for closed arrowheads 
--- a/R/stat-density-2d.r
+++ b/R/stat-density-2d.r
@@ -60,12 +60,13 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
       lims = c(scales$x$dimension(), scales$y$dimension())
     )
     df <- data.frame(expand.grid(x = dens$x, y = dens$y), z = as.vector(dens$z))
+    df$nz <- df$z/max(df$z)
     df$group <- data$group[1]
 
     if (contour) {
       StatContour$compute_panel(df, scales, bins, binwidth)
     } else {
-      names(df) <- c("x", "y", "density", "group")
+      names(df) <- c("x", "y", "density", "scaled", "group")
       df$level <- 1
       df$piece <- 1
       df
--- a/R/stat-density-2d.r
+++ b/R/stat-density-2d.r
@@ -7,6 +7,11 @@
 #'   using [MASS::bandwidth.nrd()].
 #' @section Computed variables:
 #' Same as [stat_contour()]
+#'
+#' With the addition of:
+#'  \item{density}{the density estimate}
+#'  \item{scaled}{density estimate, scaled to maximum of 1}
+
 stat_density_2d <- function(mapping = NULL, data = NULL,
                             geom = "density_2d", position = "identity",
                             ...,
@@ -60,7 +65,7 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
       lims = c(scales$x$dimension(), scales$y$dimension())
     )
     df <- data.frame(expand.grid(x = dens$x, y = dens$y), z = as.vector(dens$z))
-    df$nz <- df$z/max(df$z)
+    df$nz <- df$z / max(df$z)
     df$group <- data$group[1]
 
     if (contour) {
--- a/R/stat-density-2d.r
+++ b/R/stat-density-2d.r
@@ -9,9 +9,10 @@
 #' Same as [stat_contour()]
 #'
 #' With the addition of:
-#'  \item{density}{the density estimate}
-#'  \item{scaled}{density estimate, scaled to maximum of 1}
-
+#' \describe{
+#'   \item{density}{the density estimate}
+#'   \item{ndensity}{density estimate, scaled to maximum of 1}
+#' }
 stat_density_2d <- function(mapping = NULL, data = NULL,
                             geom = "density_2d", position = "identity",
                             ...,
@@ -68,10 +69,10 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
     df$group <- data$group[1]
 
     if (contour) {
-      StatContour$compute_panel(df, scales, bins, binwidth)
+        StatContour$compute_panel(df, scales, bins, binwidth)
     } else {
       names(df) <- c("x", "y", "density", "group")
-      df$scaled <- df$density / max(df$density)
+      df$ndensity <- df$density / max(df$density, na.rm = TRUE)
       df$level <- 1
       df$piece <- 1
       df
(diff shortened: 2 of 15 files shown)


clauswilke:
This seems like a good idea to me. Unfortunately it's too late for 2.3.0, we're in a soft freeze.

A couple of comments:
- Please make sure your coding style matches the ggplot2 style (e.g., spaces around `/`).
- The documentation should be updated. Already, the "computed variables" sections of `geom_density_2d()` and `geom_contour()` seem incomplete, and they will be even more incomplete after this change.

R/stat-density-2d.r
```diff
@@ -60,12 +60,13 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
       lims = c(scales$x$dimension(), scales$y$dimension())
     )
     df <- data.frame(expand.grid(x = dens$x, y = dens$y), z = as.vector(dens$z))
+    df$nz <- df$z/max(df$z)
```
clauswilke: Why `nz`? This naming seems strange.

bjreisman: I wasn't sure what to call this one. I went with `nz` for "normalized z" as in `ndensity` and `ncount`. In my own code, I normally name variables which are derived from other variables using "x.difference" so `z` would become, `z.scaled` or similar, but I wasn't sure what the convention was for ggplot2.  

Alternatively, we could just call it `scaled`, but then the renaming that happens a few lines down would be somewhat redundant.

bjreisman:
Thanks for the tips! I've updated the styling and documentation as suggested. I wasn't sure how to best describe the `pieces` computed variable in `stat_contour`, so that may need more revision than the others.

R/stat-density-2d.r
clauswilke: I find it strange to introduce a variable name and then change it 5 lines down. Also, what happens if the `if (contour) {` branch is executed? What does `StatContour` do to the `nz` column?

bjreisman: I'm inclined to agree; however, I tried to preserve as much of the approach of the original `stat_density` function as I could and that was the approach taken there:

```{r}
#exerpt from stat_denisty, line 62
    df <- data.frame(expand.grid(x = dens$x, y = dens$y), z = as.vector(dens$z))
    df$group <- data$group[1]

    if (contour) {
      StatContour$compute_panel(df, scales, bins, binwidth)
    } else {
      names(df) <- c("x", "y", "density", "group")
      df$level <- 1
      df$piece <- 1
      df
    }
```
`StatContour` doesn't do anything with the `nz` column, so to avoid having to rename things we could just name it when we create it.
```{r}
    df <- data.frame(expand.grid(x = dens$x, y = dens$y), z = as.vector(dens$z))
    df$scaled <- df$z / max(df$z)
    df$group <- data$group[1]

    if (contour) {
      StatContour$compute_panel(df, scales, bins, binwidth)
    } else {
      names(df) <- c("x", "y", "density", "scaled", "group")
      df$level <- 1
      df$piece <- 1
      df
    }
```
Or we could just create it later on, which avoids passing a column that won't be used to `StatContour`
```{r}
    df <- data.frame(expand.grid(x = dens$x, y = dens$y), z = as.vector(dens$z))
    df$group <- data$group[1]

    if (contour) {
      StatContour$compute_panel(df, scales, bins, binwidth)
    } else {
      names(df) <- c("x", "y", "density", "scaled", "group")
      df$scaled <- df$density / max(df$density)
      df$level <- 1
      df$piece <- 1
      df
    }
```
Or perhaps there's a third even cleaner method to achieve the same goal. Which approach do you think would be best?

clauswilke: I think the later creation is better, though you shouldn't name it before you add it to the data frame. :-)

bjreisman: Fixed! ✨ 

---
This may be a separate branch/pull request, but it might be worthwhile to change the behavior of `stat_density2d` when `aes(fill = stat(level))` is called when `contour = FALSE` and the behavior of `aes(fill = stat(density))` when `contour = TRUE` to return an error or a warning. Neither of those statistics are meaningful when called in the wrong context and it can return some cryptic errors as seen below. 😬 

```{r}
library(ggplot2)
library(dplyr)
library(viridis)
#2d densiy plot colored by level
diamonds %>%
ggplot(aes(x=x, y= depth)) +
  stat_density_2d(aes(fill = stat(density)),
                  geom = "raster",
                  n = 100,
                  bins = 10,
    contour = T) +
  facet_wrap(clarity~.) +
  scale_fill_viridis(option = "A")

#>  Error in is.finite(x) : default method not implemented for type 'closure' 
```
```{r}
diamonds %>%
ggplot(aes(x=x, y= depth)) +
  stat_density_2d(aes(fill = stat(level)),
                  geom = "polygon",
                  n = 100,
                  bins = 10,
    contour = F) +
  facet_wrap(clarity~.) +
  scale_fill_viridis(option = "A")
```
plot7

hadley:
@clauswilke can you finish this off by approving the PR (if you think it's ready to be merged)

clauswilke:
@hadley Yes. One thing that I think needs to be fixed in the ggplot2 codebase first is a new section in NEWS.md for post 3.0.0 development.

@bjreisman At a minimum, I think roxygen needs to be rerun so all the documentation changes are processed correctly. I would also suggest adding a line to `NEWS.md` (not `NEWS`!). It would have to go above the current ggplot2 3.0.0 section. I'll read the code over carefully tomorrow and let you know if I see anything else.

R/stat-density-2d.r
```diff
@@ -66,6 +71,7 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
       StatContour$compute_panel(df, scales, bins, binwidth)
     } else {
       names(df) <- c("x", "y", "density", "group")
+      df$scaled <- df$density / max(df$density)
```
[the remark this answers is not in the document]
bjreisman: Thanks for taking the time to consider this addition!

I wasn't entirely happy with the switch between `ndensity` to `scaled` either and would actually prefer to use `ndensity` here. However, in `stat_density` [the 1d version], the rescaled statistic is called `scaled` while in `stat_bin` it's called `ndensity` or `ncount`. 

Since `stat_density_2d` is the 2d version of `stat_density` I tried to mirror the syntax by naming it `scaled`.  I think it might make the most sense to use one syntax throughout all the stat functions (`ndensity`), but I will leave that to your discretion. To summarize, I think there are a three options. 

1. Mirror the syntax from the 1d versions of the stat layers such that `stat_density_2d` and `stat_density` use `scaled` while `stat_bin2d`, `stat_hexbin`, and `stat_bin` use `ndensity` and `ncount`. (implemented this PR)

2. (2a) Switch over to a single syntax for the 2d stats with this version - `ndensity` replaces `scaled` in `stat_density_2d`. 

3. (2b) Switch over to a single syntax for all density based stats -  `ndensity` replaces `scaled` in both `stat_density` and `stat_denisty_2d`. I suppose you would want to just duplicate `scaled` and name the new variable `ndensity` to avoid breaking existing code. 

--
RE: na.rm = TRUE; you're correct, I've added that to my local version for my next PR. I can submit that change once `scaled` vs. `ndensity` is settled and I'll update the documentation and news as well.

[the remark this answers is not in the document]
clauswilke: I think option 2b (uniform naming with duplicated `scaled` column for `stat_density`) is best. But I'd like to hear @hadley's input as well.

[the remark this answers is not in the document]
hadley: Yeah, 2b sounds good (assume it doesn't affect existing code, in which case scaled should be added, instead of replacing the existing variable name)

bjreisman:
Okedoke, I've implemented option 2b with the latest commit, as updating NEWS.md and adding an example to geom_density2d() for the new functionality. 

I tried running roxygen to update the documentation, but for some reason it didn't recognize the `\section{Aesthetics}{` tag, so those were dropped from every geom on my local version. It may have something to do with the following warning: "Version of roxygen2 last used with this package is 6.0.1.9000.  You only have version 6.0.1." When I tried to install the dev version from github it wouldn't run. I think I'm running into this error[Having problems with slash direction in Roxygen2 package](https://stackoverflow.com/questions/37336673/having-problems-with-slash-direction-in-roxygen2-package). Any tips on how to overcome this?

NEWS.md
```diff
@@ -157,6 +157,17 @@ To be released as 2.3.0
...
+  include  a normalized version the `level`, `density`, and `count` parameters, 
+  named `nlevel`, `ndensity`, and `ncount`. This mirrors the use of `ndensity`
+  in `stat_density()` and `ncount`/`ndensity` in `stat_bin()`. These stats
+  should be useful for faceted 2D density plots and bin plots in cases where
+  the distribution of density within a facet is more relevant than the absolute
+  density or count across facets. (@bjreisman, #2680)
+
+* `stat_density()` now includes the calculated statistic `nlevel`, an alias 
+  for `scaled`, to better match the syntax of `stat_bin` (@bjreisman, #2680)
+
```
clauswilke: Unfortunately this whole part seems to have ended in the wrong section of the file. It should be at the very top, under ggplot2 3.0.0.9000. You may have to rebase. Also, the whole section should only be 1-2 sentences, such as "`stat_contour()`, `stat_density2d()`, `stat_bin2d()`,  `stat_binhex()` all now also calculate normalized densities, levels, and counts (@bjreisman, #2680)".

R/stat-density.r
```diff
@@ -21,6 +21,8 @@
 #'   \item{count}{density * number of points - useful for stacked density
 #'      plots}
 #'   \item{scaled}{density estimate, scaled to maximum of 1}
+#'   \item{ndensity}{alias for stat(scaled), to mirror the syntax of
+#'    geom_bin}
```
clauswilke: Can you put `stat(scaled)` and `geom_bin` in backticks so they show up as code? And also please write `geom_bin()` (with parentheses).

clauswilke:
I suspect I can fix the roxygen issue if you give me access to your repo fork. Won't get to it until later in the week though.

NEWS.md
clauswilke: Apologies, one more comment: The issue this is closing is #2679, not #2680.  #2680 is the pull request. And one of the git commit messages should contain the phrase "Closes #2679" or "Fixes #2679" so that github automatically closes the issue once the pull request is integrated. You can add this to the next commit you make.

bjreisman:
Done! Thanks for taking the time to review these changes.

clauswilke:
@bjreisman I have rebased your PR and made some documentation cleanups. Let's see if all the checks pass.

Most importantly, you were using `scale_fill_viridis()` in two examples, but that scale does not exist in ggplot2 out of the box. I changed those to use `scale_fill_viridis_c()`, which does exist. It's important that all examples can be run as written without any unspecified dependencies. Otherwise CRAN check will fail.

clauswilke:
I just merged this, thanks!

PR #2704, opened 2018-06-16
revdeps for 2.2.1 vs 2.2.1.9000

topepo:
`data.sqlite` is at https://www.dropbox.com/s/xv1hxct2446a9vs/ggplot2_data.sqlite.zip?dl=0

We checked 2105 reverse dependencies (1715 from CRAN + 390 from BioConductor), comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 101 new problems
 * We failed to check 63 packages

### DIFF
--- a/revdep/README.md
+++ b/revdep/README.md
@@ -3,285 +3,216 @@
 |field    |value                        |
 |:--------|:----------------------------|
 |version  |R version 3.5.0 (2018-04-23) |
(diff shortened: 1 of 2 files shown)


hadley:
20 fewer broken packages, so progress, but not a lot. I'll send out another email tomorrow.

PR #2721, opened 2018-06-27
New examples for theme_*()

jrnold:
Update the example used in the "complete themes" help page (`ggtheme`). I extended the current `mtcars` example to use more (all?) of the features which themes affect: title, subtitle, tag, caption, and both x- and y-facets.

There are now two plot examples. One with no facets, and one with 2x2 facets. The latter is enclosed in `\donttest{}` to keep R CMD check from complaining about the runtime.

The `mtcars` dataset may not be the best for this, but I wanted to keep it close to the orginal example, and the data isn't as important in this example.

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -49,16 +49,45 @@
 #' }
 #'
 #' @examples
-#' p <- ggplot(mtcars) + geom_point(aes(x = wt, y = mpg,
-#'      colour = factor(gear))) + facet_wrap(~am)
-#' p + theme_gray() # the default
-#' p + theme_bw()
-#' p + theme_linedraw()
-#' p + theme_light()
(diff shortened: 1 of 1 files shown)


jrnold:
So I think that the purpose of these examples is to different than that of the examples in `theme.r`. 
The docs for `theme()` covers changing various features of the theme. But I think that the examples of the included themes should be rich enough to show any differences (or lack thereof) between themes.

> as none of the complete themes differ in font family or colour or size, the title, tag, subtitle, and caption elements don’t exhibit any differences across plots, except minor positional adjustments when using theme_void()

I would have expected that to be the case, but without an example that uses the full capabilities of the themes, this wouldn't have been obvious.

jrnold:
Awesome! Thanks!

Issue #2723, opened 2018-06-28
Cannot set legend.box.background with theme_void()


clauswilke:
It doesn't seem possible to set the `legend.box.background` theme setting under `theme_void()`. Other legend settings work.

Reprex:
``` r
library(ggplot2)
df <- data.frame(x = 1:3)
ggplot(df, aes(x, x, color = factor(x))) +
  geom_point() +
  theme_void() +
  theme(legend.background = element_rect(fill = "gray80"))
```

![](https://i.imgur.com/OqMv7T4.png)

``` r

df <- data.frame(x = 1:3)
ggplot(df, aes(x, x, color = factor(x))) +
  geom_point() +
  theme_void() +
  theme(legend.box.background = element_rect(fill = "gray80"))
#> Warning in if (theme$legend.box == "horizontal") {: the condition has
#> length > 1 and only the first element will be used
#> Warning in if (theme$legend.box == "vertical") {: the condition has length
#> > 1 and only the first element will be used
#> Error in gtable_add_cols(guides, theme$legend.box.margin[4], pos = 0): is.gtable(x) is not TRUE
```

Created on 2018-06-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

clauswilke:
Also, this works without `theme_void()`, of course:

``` r
library(ggplot2)
df <- data.frame(x = 1:3)
ggplot(df, aes(x, x, color = factor(x))) +
  geom_point() +
  #theme_void() +
  theme(legend.box.background = element_rect(fill = "gray80"))
```

![](https://i.imgur.com/DVGFKr0.png)

Created on 2018-06-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

ptoche:
works for me:
sessionInfo()
R version 3.5.0 (2018-04-23)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.4

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils    
[...]

PR #2725, opened 2018-06-29
Make legend.box.background work for theme_void()

clauswilke:
Closes issue #2723.

One thing I wasn't sure about was whether I should add all the missing legend theme parameters in `theme_void()`. There are more than just `legend.box`. I tried copying them over from `theme_grey()` but that caused changed spacing in two visual tests so I reverted.


R/guides-.r
```diff
@@ -263,7 +263,7 @@ guides_build <- function(ggrobs, theme) {
     # add space between the guide-boxes
     guides <- gtable_add_col_space(guides, theme$legend.spacing.x)
 
-  } else if (theme$legend.box == "vertical") {
+  } else { # theme$legend.box == "vertical"
```
hadley: Can you change the test in the first block to use `identical()`? That's safer if `theme$legend.box` is somehow not a length-1 vector. (I'm surprised the user of `NULL` doesn't cause a warning, TBH)

clauswilke: Done.

Issue #2553, opened 2018-05-01
Review documentation of most visited webpages


hadley:
* [x] Find out top 20 pages from google analytics

dpseidel:
Top 20 most visited ggplot2 pages as of 6/11/18.

- [ ] [Create Elegant Data Visualisations Using the Grammar of Graphics • ggplot2](ggplot2.tidyverse.org)
- [ ] [Function reference • ggplot2](http://ggplot2.tidyverse.org/reference/index.html)
- [ ] [Bar charts — geom_bar • ggplot](http://ggplot2.tidyverse.org/reference/geom_bar.html)
- [ ] [Modify axis, legend, and plot labels — labs • ggplot2](http://ggplot2.tidyverse.org/reference/labs.html)
- [ ] [Points — geom_point • ggplot2](http://ggplot2.tidyverse.org/reference/geom_point.html)
- [ ] [Text — geom_label • ggplot2](http://ggplot2.tidyverse.org/reference/geom_text.html)
- [ ] [Modify components of a theme — theme • ggplot2](http://ggplot2.tidyverse.org/reference/theme.html)
- [ ] [Position scales for continuous data (x & y) — scale_continuous • ggplot2](http://ggplot2.tidyverse.org/reference/scale_continuous.html)
- [ ] [Smoothed conditional means — geom_smooth • ggplot2](http://ggplot2.tidyverse.org/reference/geom_smooth.html)
- [ ] [Lay out panels in a grid — facet_grid • ggplot2](http://ggplot2.tidyverse.org/reference/facet_grid.html)
- [ ] [Histograms and frequency polygons — geom_freqpoly • ggplot2](http://ggplot2.tidyverse.org/reference/geom_histogram.html)
- [ ] [Gradient colour scales — scale_colour_gradient • ggplot2](http://ggplot2.tidyverse.org/reference/scale_gradient.html)
- [ ] [Legend guide — guide_legend • ggplot2](http://ggplot2.tidyverse.org/reference/guide_legend.html)
- [ ] [A box and whiskers plot (in the style of Tukey) — geom_boxplot • ggplot2](http://ggplot2.tidyverse.org/reference/geom_boxplot.html)
- [ ] [Reference lines: horizontal, vertical, and diagonal — geom_abline • ggplot2](http://ggplot2.tidyverse.org/reference/geom_abline.html)
[...]

PR #2727, opened 2018-06-29
Updating most visited docs (#2553)

dpseidel:
A PR for various minor edits to documentation as I review and update the top 20 most visited webpages (#2553). 

As this grows we may need to break it up into separate PRs. For now, I'm purposely holding off on committing updated .Rd files because edits like my removing a single unnecessary period in `layer.r` propagate into 20+ pages and will clutter the PR.


R/geom-point.r
```diff
@@ -18,7 +18,7 @@
 #' There is no one solution to this problem, but there are some techniques
```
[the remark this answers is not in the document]
dpseidel: I'm going to fill this PR just with minor edits and go back through for bigger changes to text in separate PRs. For instance here with "bubblechart", and with theme, etc.

dpseidel:
@hadley This is ready to review & merge. Larger changes to documentation will be made in separate PRs.

R/fortify-map.r
```diff
@@ -85,7 +85,7 @@ map_data <- function(map, region = ".", exact = FALSE, ...) {
 #' @param regions map region
 #' @param fill fill colour
 #' @param colour border colour
-#' @param xlim,ylim latitudinal and logitudinal range for extracting map
+#' @param xlim,ylim latitudinal and longitudinal range for extracting map
```
batpigandme: range ⇨ ranges (since there are now two of them to which we are referring)

R/geom-bar.r
```diff
@@ -1,12 +1,12 @@
 #' Bar charts
 #'
-#' There are two types of bar charts: `geom_bar` makes the height of the
+#' There are two types of bar charts: `geom_bar()` makes the height of the
```
batpigandme: Stylistic, but since we open with two types...: I would follow the colon with "`geom_bar()` and `geom_col()`. `geom_bar()` makes..."

R/geom-boxplot.r
```diff
@@ -49,7 +49,7 @@
 #'   if the notches of two boxes do not overlap, this suggests that the medians
 #'   are significantly different.
 #' @param notchwidth For a notched box plot, width of the notch relative to
-#'   the body (default 0.5)
+#'   the body (default = `0.5`)
```
batpigandme: ~~Are we surrounding just the value of an argument with backticks? (I'm genuinely asking—if so, I think we're inconsistent)~~
Move backticks to surround argument and value.

R/geom-smooth.r
```diff
@@ -7,7 +7,7 @@
 #'
 #' Calculation is performed by the (currently undocumented)
 #' `predictdf` generic and its methods.  For most methods the standard
-#' error bounds are computed using the [predict()] method - the
+#' error bounds are computed using the [predict()] method -- the
 #' exceptions are `loess` which uses a t-based approximation, and
```
batpigandme: commas between the geom and the descriptor (i.e. "`loess`, which..." and "`glm`, where...") I'm less wed to the one with glm, though.

R/geom-smooth.r
```diff
@@ -20,14 +20,14 @@
 #' @seealso See individual modelling functions for more details:
 #'   [lm()] for linear smooths,
 #'   [glm()] for generalised linear smooths,
```
batpigandme: If we're giving this one sentence structure (which we are, since there's a period after the last one) there should be an "and" added to the second to last list item. (Basically, "bulleted lists within sentences" structure: http://blog.apastyle.org/apastyle/2010/03/lists-part-5-bulleted-lists.html)

R/guide-colorbar.r
```diff
@@ -1,38 +1,38 @@
...
 #'   drawn around the bar. If `NULL` (the default), no frame is drawn.
 #' @param frame.linewidth A numeric specifying the width of the frame
 #'   drawn around the bar.
 #' @param frame.linetype A numeric specifying the linetype of the frame
 #'   drawn around the bar.
-#' @param nbin A numeric specifying the number of bins for drawing colorbar. A
-#'   smoother colorbar for a larger value.
-#' @param raster A logical. If `TRUE` then the colorbar is rendered as a
-#'   raster object. If `FALSE` then the colorbar is rendered as a set of
+#' @param nbin A numeric specifying the number of bins for drawing colourbar. A
```
batpigandme: Add an article before "colourbar" (If you can have more than one, then "a," if not, then "the.")

R/guide-legend.r
```diff
@@ -4,8 +4,8 @@
 #' Legend guides for various scales are integrated if possible.
 #'
 #' Guides can be specified in each `scale_*` or in [guides()].
-#' `guide="legend"` in `scale_*` is syntactic sugar for
-#' `guide=guide_legend()` (e.g. `scale_color_manual(guide = "legend")`).
+#' `guide = "legend"` in `scale_*` is syntactic sugar for
```
batpigandme: re. my comment about backticks around arguments and their value, here there are backticks around both argument and value. I think we're wrapping both (see http://style.tidyverse.org/news.html#general-style), so I'd just move the backticks where mentioned earlier to encompass the argument name.

R/theme.r
```diff
@@ -161,16 +161,16 @@
 #'   (`element_text`; inherits from `title`) left-aligned by default
 #' @param plot.tag.position The position of the tag as a string ("topleft",
 #'   "top", "topright", "left", "right", "bottomleft", "bottom", "bottomright)
-#'   or a coordinate. If a string, extra space will be added to accomodate the
+#'   or a coordinate. If a string, extra space will be added to accommodate the
```
batpigandme: 👏 I _never_ spell the word accommodate correctly on the first try.

R/guide-legend.r
dpseidel: I agree in principle however "default" (from above) is not an actual argument name where `guide` is. Thoughts?

batpigandme: Oh, oh! I missed that. That makes sense. I almost think that it'd be better not to use `=`, in that case (e.g. is `0.5` by default, or default is `0.5` or whatever).

R/geom-boxplot.r
batpigandme: I like defaults to `notchwidth = 0.5` — though, honestly, both make sense. Go with where the spirit moves you!

dpseidel: Done!

PR #2738, opened 2018-07-07
Theme override

malcolmbarrett:
It's common for packages to include wrapper functions for ggplots that include themes. In those cases, `theme_set()` and friends don't work, and you need to manually add a theme. This PR adds a simple, hard override option to `theme_set/replace/update()` to allow it to apply under all circumstances.

``` r
qp <- function(data = mtcars) {
  ggplot(data, aes(x = hp, y = mpg)) +
    geom_point() +
    theme_bw()
}

qp()
```

``` r

# nothing happens because of internal function theme
theme_set(theme_dark())
qp()
```

``` r

# override the theme
theme_set(theme_dark(), override = TRUE)
qp()
```

``` r

# reset override to FALSE
theme_set(theme_gray()) 
# returns to normal behavior 
qp() 
```

``` r

# need to explicitly override to work
theme_set(theme_gray(), override = TRUE) 
qp()
```

``` r

# also ignores non-complete theme elements
# it's a complete override of the theme
qp() + theme(text = element_text(size = 20))
```

I did think to add a softer version of the override that doesn't ignore non-complete but it seemed better to not do that, because if you're adding theme elements to the call, you might as well just change the theme manually.

### DIFF
--- a/R/theme-current.R
+++ b/R/theme-current.R
@@ -64,7 +64,7 @@ theme_env$override <- FALSE
 #'
 #' theme_set(theme_dark(), override = TRUE)
 #'
-#' # Uses theme_dark()
+#' # Uses theme_dark() instead
 #' custom_plot()
 #'
 #' # Modifying theme objects -----------------------------------------
--- a/R/theme-current.R
+++ b/R/theme-current.R
@@ -2,6 +2,7 @@
 #' @include theme-elements.r
 theme_env <- new.env(parent = emptyenv())
 theme_env$current <- theme_gray()
+theme_env$override <- FALSE
 
 #' Get, set, and modify the active theme
 #'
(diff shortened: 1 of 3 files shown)


clauswilke:
It seems to me all of this is very unidiomatic for R. A more conventional implementation would look as follows:
``` r
library(ggplot2)

qp <- function(data = mtcars, theme = theme_bw()) {
  ggplot(data, aes(x = hp, y = mpg)) +
    geom_point() +
    theme
}

theme_set(theme_dark())
qp()
```



``` r
qp(theme = theme_classic())
```



``` r
qp(theme = NULL)
```



Created on 2018-07-07 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

malcolmbarrett:
That's certainly true, but in that sense, `theme_set()` and friends are also unidiomatic for R (and, in fact, I use them very rarely because I prefer it to be explicit in every call). I do like that your approach can set it to `NULL` and thus respect the set theme, although I think a driving factor of people wanting to do this is not to edit every call. 

This doesn't seem likely to progress, but if it's worth discussing, I can repost as an issue.

Issue #2540, opened 2018-04-27
Enable User Defined Theme Elements


nhamilton1980:
This issue is at the suggestion of Hadley, based on the following: 

https://github.com/tidyverse/ggplot2/pull/2305

I am the author of the [ggtern](http://ggtern.com) package, which requires a whole bunch (50+) of new theme elements outside of the standard ggplot2 themes defined in the non-exported `.element_tree` object.

In this issue I will speak of my experience with ggtern, however, it is possible that any extension author may face the same problem.

The complexity of ggtern (and its sensitivity to changes within ggplot2) can be HUGELY reduced if there was a mechanism for extension authors such as myself to be able to create their own theme elements, and add them to the `.element_tree`. The solution I had proposed within the above link (via use of R6 class) had resolved this, with no impinging effect or risk to existing ggplot2 functionality.

Because I cannot add new theme elements, this means that I need to duplicate the `.element_tree` within the ggtern package, and add my new theme elements / hierarchies. This in turn means that many of the plot construction routines also needs to be held as a duplicate inside the ggtern library, leading to fragilities and vulnerabilities to changes within base ggplot2, which happens almost every year. I am trying to remove these fragilities.

To me it seems very logical for extension authors to be able to create their own theme elements. At the moment there is no way to do this to my knowledge other than the method used in ggtern, which is far from ideal. The theme element tree offered by ggplot2 is quite substantial, but by no means is it exhaustive, as has been demonstrated in ggtern, requiring dozens more for it to function properly.

clauswilke:
I have reviewed your pull request. If I understand correctly what you need for ggtern, your proposed solution seems way overengineered. In particular, adding a whole new class model (R6) to ggplot2 doesn't seem warranted. I'd also like to point out that your code doesn't follow the tidyverse style guide (http://style.tidyverse.org/), in particular with respect to spaces around `if`, `else`, `{`, and after commas.

If all you need is the ability to add new theme elements, that functionality can be achieved with one new function, `add_element()`. The only issue is that direct assignment to `.element_tree` is not possible, so it'd have to be stuck into an environment, like so:

```{r}
element_env <- new.env(parent = emptyenv())
element_env$element_tree <- list()

add_element <- function(key, element_def) {
  if (is.null(element_env$element_tree[[key]])) {
    element_env$element_tree <- c(
      element_env$element_tree,
      setNames(list(element_def), key)
    )
  } else {
    warning("trying to redefine existing element.")
  }
}
```

[...]

PR #2741, opened 2018-07-08
User-defined theme elements

clauswilke:
This is the first step towards user-defined theme elements as discussed in #2540. This particular PR doesn't add any features yet, it just reorganizes the code so that global variables live in an environment. It could be merged as is, but mostly I'd want quick feedback to make sure I'm on the right track.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,6 +61,7 @@ BugReports: https://github.com/tidyverse/ggplot2/issues
 LazyData: true
 Collate: 
     'ggproto.r'
+    'ggplot-global.R'
     'aaa-.r'
     'aes-calculated.r'
     'aes-colour-fill-alpha.r'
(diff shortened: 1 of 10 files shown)


clauswilke:
Let me ponder this. Making the valid theme elements a theme setting might require solving some technical challenges but I think it makes sense conceptually.

For the valid aesthetics, I'm a bit concerned about orthogonality. It might create a zoo of different themes that only differ in the aesthetics they support, e.g. `theme_gray()` with this extra aesthetic and with that extra aesthetic. And package authors might be tempted to republish those themes under the original ggplot2 names, if they look exactly the same.

hadley:
Yeah, and providing valid aesthetics as a theme component makes them a plot level setting, which will make mingling geoms from different packages potentially complex.

Maybe layers should have a property that allows you to define extra aesthetics?

clauswilke:
I'll go ahead and merge this because I think this reorganization makes sense regardless of how we proceed. I'll move the discussion of the various possibilities to the respective issues.

Issue #2239, opened 2017-08-12
Geoms and scales: default settings in theme


baptiste:
This idea has been floated around multiple times but I'll raise it again. Currently, themes do not affect the geom properties, such as point and line colour, etc. Scales, in particular colour and fill, suffer the same issue. 

Users who wish to use a specific theme – theme_dark() is a striking example – end up having to change the geom and/or scale defaults. A consistent yet original look, such as a company's brand, may require to set default colour schemes for discrete and continuous scales. 

These two steps currently require adding multiple manual parameters to each plot, or fiddling around with the awkward combination of `set_geom_defaults()` and `scale_colour_discrete <- scale_colour_brewer`.

The suggestion made here is to consider default aesthetics an integral part of the grammar – they're not mapped to data (unless they are), but should be _consistent_ with the overall look of the theme. Themes have gained in consistency with `rel()` and inheritance, but at the moment setting a dark background will make points invisible: not a great default behaviour.

Lattice, as well as Julia's Gadfly, seem to have followed this line of thought to include geom defaults in the theme description. Beamer themes define default colour schemes to be inherited by slide elements.

A proof-of principle, setting point colours to default to the plot title's colour (an existing theme element, for simplicity), was [implemented here](https://github.com/baptiste/ggplot2/commit/03ead2bc00739560719b62822b8fc6a6ac22aa8a). (Note: the use of complex numbers is just for curiosity – it would correspond to something like `waiver()` in the standard ggplot2 way of dealing with values that are to be resolved just before rendering.

hadley:
I think we can do this for the next major release of ggplot2 (i.e. this can be a project for my next summer intern). The basic implementation is much simpler than I'd feared; the work will mostly be carefully considering which theme properties we need and then performing a routine transformation for each geom. (We'll probably also need to use a SE variant of `aes()` since we I think don't want any delayed evaluation in the defaults)

baptiste:
Sounds great!

As themes grow in scope, I wonder if one could also consider developing an alternative syntax for them, based on css. (I'm thinking of the way d3.js often uses css to set default properties, although that's a bit inconsistent because of the css vs svg attributes mess). 
[...]

Issue #2691, opened 2018-06-09
Register default scales via themes


clauswilke:
As previously mentioned in the discussion of #2239: It would be nice if default scales could be set via themes.

Currently, default scales are defined as functions that are named `scale_`*`aesthetic`*`_`*`data-type`*`()`, where *`aesthetic`* is the name of the aesthetic and *`data-type`* represents the data type, i.e., continuous, discrete, date, etc. For example, `scale_colour_discrete()` is the default scale function for discrete colors.

The definition of these defaults is distributed all over the scales code, because for some scales the default scale is the only scale that exists (example: https://github.com/tidyverse/ggplot2/blob/77fc5c705efd19c96f6b69f078a0aeacc87c933d/R/scale-alpha.r#L31) whereas for others there are multiple possible ones and hence one is set as the default (example: https://github.com/tidyverse/ggplot2/blob/77fc5c705efd19c96f6b69f078a0aeacc87c933d/R/zxx.r#L6).

It would be nice if there were a way to change the defaults other than to change the global function definitions, and in particular if the defaults could be linked to the theme. For example, the theme could have an entry `default_scales = list("scale_color_discrete" = scale_color_hue, "scale_color_continuous" = scale_color_gradient, ...)` that lists the default scales to be used with this theme.

In terms of implementation, I see two problems, and it's not clear to me how much of a hurdle they might be. First, the code that searches for the default scales currently doesn't have access to the theme: https://github.com/tidyverse/ggplot2/blob/bcc3ae59f7f5257fdfce5314a71ae255e4a3f12a/R/scale-type.R#L1-L20
It is called from this function:
https://github.com/tidyverse/ggplot2/blob/bcc3ae59f7f5257fdfce5314a71ae255e4a3f12a/R/scales-.r#L91
which is called from here:
https://github.com/tidyverse/ggplot2/blob/a7b31356ef1de985fc23e3fee14ef5d2afd20c85/R/layer.r#L214
and here:
https://github.com/tidyverse/ggplot2/blob/a7b31356ef1de985fc23e3fee14ef5d2afd20c85/R/layer.r#L274

Second, it's not immediately clear to me whether we even have the complete theme by the time these functions are called. They are ultimately called in `plot_build()` here: 
https://github.com/tidyverse/ggplot2/blob/3f63d6235d96da47c2c801d3c72afa2f2d5a0e85/R/plot-build.r#L48
[...]

PR #2749, opened 2018-07-11
Allow default geom aesthetics to be set from theme

dpseidel:
## Motivation
The goal of this PR is to allow certain default aesthetics to be set by the plot theme. This can be used to achieve better default behaviours (e.g. plotting white points by default, instead of black, when using `theme_dark()`) and to allow user created themes to more easily control default aesthetics. This PR addresses #2239. 

## Basic Approach
The basic approach is to alter each Geom's `default_aes` from something like:
``` r
# alpha must default to NA to respect RGBA color specification
default_aes = aes(colour = "black", size = 0.5, linetype = 1, alpha = NA)
```
to something more like:
``` r
default_aes = aes(
  colour = theme$geom$col,
  fill = theme$geom$fill,
  linetype = 1,
  alpha = NA
)
```
and then evaluate the `default_aes` mappings in an environment conscious of `plot$theme`. Since `aes()` quotes all arguments, I currently I do this by wrapping `default_aes` in `expr()` and evaluating it inside `geom$use_defaults`.  This in turn gets called to build each layer and the guides. Some minor adjustments were also made to handle incomplete themes (I'm confident there is a better way, consider this a temporary patch) and to maintain `update_geom_defaults()` functionality. 

@hadley, I think this is *still* slightly different from the algorithm we originally discussed. You originally indicated that we should evaluate the mappings within the plot environment and then call something like `on.exit(set_geom_theme(NULL))`, to reset to the default expression. I so far have been unable to make this work, and I think it may require our evaluation to happen earlier in the code. If this is still what you have in mind, I can reconsider and refactor the code. I agree it could be more streamline to evaluate these defaults upfront rather than at multiple points in the build, I simply have yet to make that work. 

## Next Steps
This PR is a work in progress. Currently I have only added theme elements `colour` and `fill` and only updated the `default_aes` for `GeomPoint` and `GeomRect`, eventually all geoms will need to be updated.  This will mean making some decisions about how to handle variability across geom defaults.  For example, looking just at `colour`: the default for `geom_smooth()`, `geom_contour()`, `geom_density2d()`, and `geom_quantile()` is all `#3366FF` and the rest of the geoms default to `colour = "black"` (or `grey20` for boxplots and violins but this can and probably should be `black` also). It would make sense to set the default to `black` in `theme_grey()`, but that either means that `geom_smooth()` et al. should not be coded to inherit from `colour` from `theme()` ever, or these geoms will also be forced to default to black. I lean towards leaving these special defaults hardcoded in these cases as to not change expected behaviour.  This may also influence exactly which elements should be set in themes (eg. `size` has more meaningful variability across geoms that we may want to preserve). A full list of geom defaults is [here](https://gist.github.com/dpseidel/57f28ccc034ccfdd0cf25b6c6c82a635). From my analysis, I suspect we will want to allow `theme()` to set at least `colour`, `fill`,  and `alpha`, potentially also `shape`, `size` and `linetype`. Font `family` and `size` should also be considered but likely should inherit from other text elements set by `theme()`. 

I have not yet updated documentation, added a NEWS bullet, or additional tests as all of this will be done further down the line. This branch does successfully build and pass tests. Running through existing package examples was very useful for finding side effects. 

## Visualizing the changes:
``` r
# theme_grey() is still our default and delivers expected defaults
ggplot(mpg, aes(displ, hwy)) + geom_point()
```

``` r
# custom themes can be added to change default colour and fill (other aesthetics may be added in the future)
my_theme <- theme(geom = element_geom(colour = "purple", fill = "darkblue"))
ggplot(mpg, aes(displ, hwy)) + geom_point() + my_theme
```

``` r
ggplot(mpg, aes(displ, hwy)) + geom_col() + my_theme
```

``` r
# theme dark now stands out automatically with lighter fill and colour defaults
ggplot(mtcars, aes(cyl, mpg)) + geom_col() + theme_dark() 
```

``` r
# theme had to be fed to guides to be able to handle discrete legends properly
ggplot(mpg, aes(displ, hwy, shape = drv)) + geom_point() + theme_dark()
```

``` r
# note that the mapping still overrides theme defaults
ggplot(mpg, aes(displ, hwy, colour = drv)) + geom_point() + my_theme
```

``` r
# and params still override all mapped and theme aesthetics
ggplot(mpg, aes(displ, hwy, colour = drv)) + geom_point(colour = "red") + my_theme
```

Created on 2018-07-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


clauswilke:
I think this is going to be a great addition to ggplot2. One comment that dovetails with something @baptiste mentioned at some point in a related post: It's unlikely that there's going to be one choice of color that is going to work for all geoms. Usually, the way this is solved in design is via a small palette of choices. E.g., there could be a foreground color, a background/fill color, and a few accent colors. The geom defaults would then pick the appropriate choices from this palette.

As an example, themes in PowerPoint have 10 colors, background 1, background 2, text 1, text 2, and accent 1 through 6. For ggplot, where we often need to fill things, I'd propose to use something like background 1 & 2, text 1 & 2, accent 1 - 3, and fill 1 - 3.

dpseidel:
Thanks @clauswilke - the palette idea would solve the variability issue I mentioned above. I just need to think critically about how to choose and document these palettes.

baptiste:
Nice, thanks for doing this.

For reference, there could be some [interesting ideas to pick from Gadfly](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/theme.jl#L319)

(notably: a number of foreground + background tints, but also the sort of [automatic colour pairing](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/theme.jl#L336) that I mentioned in the issue (e.g deriving a nice palette by muting/complementing/... a base colour).

hadley:
I think we'll need at least colour.accent and fill.accent. I'm not sure how we'd define multiple accents (e.g. 1-3) to make it clear where they are used. The googlesheet should be helpful for pulling these together.

We might want to consider naming them `colour`, `colour1` etc to avoid providing too many semantics.

baptiste:
Alternatively, maybe vectors? `accent[1:3]` 
It might require something like a length check to minimise confusion, but it would seem more extensible than a series of hard-coded variable names which don't add semantic value.

hadley:
Making it a vector suggests that there could be arbitrarily many of them. I don't know if that's any better.

hadley:
I don't think that's an axis of generalisation that will be helpful. This is about enabling built-in geoms to be themeable, so there is a fixed and known set of possible values.

baptiste:
Not sure what's limiting this to built-in geoms – couldn't the ggunicorn package want to write

```
GeomUnicorn$default_aes = aes(horn.colour = theme$geom$colour[13])
```

?

baptiste:
Hmm, I guess we're not seeing it the same way – I'll leave it at what I wrote above, should others want to weigh in. 
Personally I see no additional complexity in having the theme$geom element specify a vector of colours[1:n] rather than a fixed set of variables such as colour1...colour2. Built-in geoms as implemented in this PR would simply use theme$geom$colours[1], colours[2]... instead of theme$geom$colour1.
Unless I'm missing something this doesn't break the separation of geoms/themes any more than this whole proposal to inherit default values from the theme: it simply offers a more extensible way to do so beyond the built-in geoms, should the need arise in an extension package (and packages like ggthemes, etc. often do pair a theme, a scale, and a geom, in practice).

R/geom-.r
```diff
@@ -107,11 +107,18 @@ Geom <- ggproto("Geom",
   setup_data = function(data, params) data,
 
   # Combine data with defaults and set aesthetics from parameters
-  use_defaults = function(self, data, params = list()) {
+  use_defaults = function(self, data, params = list(), theme) {
```
hadley: I think the code related to evaluating aesthetics in the special theme environment should be pulled out into its own function so that we can think about it in isolation, and then test it.

R/annotation-logticks.r
```diff
@@ -200,7 +200,10 @@ GeomLogticks <- ggproto("GeomLogticks", Geom,
     gTree(children = do.call("gList", ticks))
   },
 
-  default_aes = aes(colour = "black", size = 0.5, linetype = 1, alpha = 1)
+  default_aes = expr(aes(
```
hadley: I don't think you need the `expr()` wrapper here. `aes()` already captures the unevaluated aesthetics?

R/annotation-logticks.r
```diff
@@ -200,7 +200,10 @@ GeomLogticks <- ggproto("GeomLogticks", Geom,
...
+    colour = theme$geom$colour, size = 0.5, linetype = 1,
```
hadley: I think `theme$geom$colour` is a bit long and a bit magical. I think it would be better to have a function like `foo("colour")`, but I don't know what we'd call it.

R/geom-abline.r
```diff
@@ -121,7 +121,12 @@ GeomAbline <- ggproto("GeomAbline", Geom,
     GeomSegment$draw_panel(unique(data), panel_params, coord)
   },
 
-  default_aes = aes(colour = "black", size = 0.5, linetype = 1, alpha = NA),
+  default_aes = expr(aes(
+    colour = theme$geom$colour,
+    size = 0.5, linetype = 1,
```
hadley: As you re-write these, can you put each aesthetic on its own line?

R/geom-defaults.r
```diff
@@ -13,7 +13,11 @@
 #' @rdname update_defaults
 update_geom_defaults <- function(geom, new) {
   g <- check_subclass(geom, "Geom", env = parent.frame())
-  old <- g$default_aes
+
+  env <- new.env()
```
hadley: If geoms are themeable, I think we can deprecate this function, so we should just leave as is and in the documentation mark it as internal, and point people towards the new theme system.

R/layer.r
```diff
@@ -303,11 +303,11 @@ Layer <- ggproto("Layer", NULL,
...
   },
 
-  compute_geom_2 = function(self, data) {
-    # Combine aesthetics, defaults, & params
+  compute_geom_2 = function(self, data, plot) {
     if (empty(data)) return(data)
 
-    self$geom$use_defaults(data, self$aes_params)
+    # Combine aesthetics, defaults, & params
+    self$geom$use_defaults(data, self$aes_params, plot$theme)
```
hadley: When you rewrite these calls with many unnamed arguments, can you please rewrite to add names? (i.e. here use `theme = plot$theme`)

R/sf.R
```diff
@@ -178,11 +178,11 @@ GeomSf <- ggproto("GeomSf", Geom,
 
 default_aesthetics <- function(type) {
   if (type == "point") {
-    GeomPoint$default_aes
+    aes(shape = 19, colour = "black", size = 1.5, fill = NA, alpha = NA, stroke = 0.5)
```
hadley: Why can we no longer inherit from point?

R/theme-elements.r
```diff
@@ -120,6 +121,37 @@ element_text <- function(family = NULL, face = NULL, colour = NULL,
   )
 }
 
+#' @param colour.accent1,color.accent1 accent colour 1,
+#'   typically a lighter version of colour
+#' @param colour.accent2,color.accent2 accent colour 2,
+#'   typically a bright colour used for geom_smooth et al.
+#' @param fill.accent accent fill colour, typically a darker version of fill
+#' @param alpha colour/fill transparency, between 0 & 1.
```
hadley: Do we need alpha as a default? People could always set individually with `alpha()`

R/theme-elements.r
```diff
@@ -120,6 +121,37 @@ element_text <- function(family = NULL, face = NULL, colour = NULL,
...
+#' @export
+#' @rdname element
+element_geom <- function(fill = NULL, fill.accent = NULL,
+                         colour = NULL, color = NULL,
+                         colour.accent1 = NULL, color.accent1 = NULL,
```
hadley: What if we used `fill` and `fill_1`?

And what if we ditched `colour` + `color` in favo(u)r of `col`?

R/annotation-logticks.r
dpseidel: It does, as quosures, so iirc, without `expr()` I ended up needing to do some quosure hacking -- to set the environment to one that understood `theme`. As I refactor the theme evaluation into it's own function as you suggest above, I'll aim to eliminate this as well.

R/geom-defaults.r
dpseidel: If we deprecate this, we will presumably need to make **all** aesthetics themeable so as to not lose the functionality?

R/sf.R
dpseidel: Because with the `expr()` wrapper the `default_aes` no longer played nice with `defaults()` function or `utils::modifyList()` which is used by `sf`. I expect this can/will be reverted with this round of changes.

R/theme-elements.r
dpseidel: Presumably not. Currently `alpha` has to be NA by default to respect RGBA colour specification

R/theme-elements.r
clauswilke: I.e., take the argument names of `element_geom()` and run through this function:
https://github.com/tidyverse/ggplot2/blob/4d2ca992dc37fee598d83d3d3f524dbba38f4187/R/aes.r#L155

clauswilke: However, this would require a `...` argument instead of listing everything explicitly, so maybe not the right approach. In any case, I’d argue against `col` just because in the rest of ggplot2 `col` is not used. And I think we also should talk about whether it’s `colour.accent1` or `colour_accent1`. We’re not entirely consistent with points vs underscores, and it would be good to pick a scheme and at least be consistent going forward.

R/sf.R
dpseidel: Actually now that I'm looking at this again, geom_sf can't inherit any "themeable" aesthetics because this function (and thus these aesthetics) get evaluated in `draw_geom` step of `layer.r` (rather than earlier in the `get_defaults` step as do all other geoms), so it currently passes an unevaluated expression to `grDevices::col2rgb()` which predictably gives an error . So this needs some refactoring to be "themeable" - either by evaluating aesthetics at the proper step (rather than passing NULL values as geom_sf does now) or by pulling theme into `draw_geom` so that the sf aesthetics can be evaluated in the context of the plots theme. It would seem to me that the former is a preferable solution but I expect that it was written this way for a reason that will become obvious as I dig in. I'm fixing your other concerns now and will work on sf in a future commit.

dpseidel:
So I've implemented most of your suggested changes at least enough so that we can iterate on them. A couple notes:

1. I've removed the `expr()` wrapper which makes a lot of sense but has some downstream effects, notably, since  `aes()` quotes its arguments  and given environments are ignored when evaluating quosures (as they have their own)-- I'm now evaluating `default_aes()` by passing the plot theme as *data* to `eval_tidy()`. This works fine with the slightly magical and long `theme$geom$col` expression or with the less magical (?) but necessarily (AFAICT) 2 argument  function `theme_aes("col", theme)` that I have implemented for GeomPoint as an example.
https://github.com/tidyverse/ggplot2/blob/23fdf39de17ae2ac32e976e581927eac523637d3/R/geom-point.r#L108-L118
 If I want to be able to pass an environment I have to hack the quosures before evaluation which seems really unconventional or consider adapting the `aes()` or `new_aesthetics()` functions to specially handle themeable elements which seems like a potential dangerzone for breaking things I don't want to break.  Is there a different approach I am missing? Do you still prefer the function or some variant of it? 
2. So far I have left the `update_geom_defaults()` function and `geom_sf()` as is (see response to your earlier comment), but those will be fixed in a future commit. 
3. I updated to `col`, `col_1`, etc. but that discussion is more than warranted and I'm happy to change things again. 
4. I'll begin adding tests in the next iteration.

clauswilke:
How about implementing `element_geom()` along the following lines?

``` r
element_geom <- function(fill = NULL,
                         fill_1 = NULL,
                         colour = NULL,
                         colour_1 = NULL,
                         colour_2 = NULL,
                         ...,
                         inherit.blank = FALSE) {
  extra_aes <- ggplot2:::rename_aes(list(...))
  
  aes_list <- modifyList(
    list(
      fill = fill, fill_1 = fill_1, colour = colour,
      colour_1 = colour_1, colour_2 = colour_2,
      inherit.blank = inherit.blank
    ),
    extra_aes,
    keep.null = TRUE
  )
  
  structure(
    aes_list,
    class = c("element_geom", "element")
  )
}

element_geom()
#> List of 6
#>  $ fill         : NULL
#>  $ fill_1       : NULL
#>  $ colour       : NULL
#>  $ colour_1     : NULL
#>  $ colour_2     : NULL
#>  $ inherit.blank: logi FALSE
#>  - attr(*, "class")= chr [1:2] "element_geom" "element"

element_geom(fill = "red", colour = "blue")
#> List of 6
#>  $ fill         : chr "red"
#>  $ fill_1       : NULL
#>  $ colour       : chr "blue"
#>  $ colour_1     : NULL
#>  $ colour_2     : NULL
#>  $ inherit.blank: logi FALSE
#>  - attr(*, "class")= chr [1:2] "element_geom" "element"

element_geom(fill = "red", color = "blue", color_1 = "black")
#> List of 6
#>  $ fill         : chr "red"
#>  $ fill_1       : NULL
#>  $ colour       : chr "blue"
#>  $ colour_1     : chr "black"
#>  $ colour_2     : NULL
#>  $ inherit.blank: logi FALSE
#>  - attr(*, "class")= chr [1:2] "element_geom" "element"
```

Created on 2018-08-09 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

R/geom-.r
```diff
@@ -106,12 +106,19 @@ Geom <- ggproto("Geom",
 
   setup_data = function(data, params) data,
 
+  # evaluate defaults according to theme
+  eval_defaults = function(self, theme) {
+    if (length(theme) == 0) theme <- theme_grey()
+
+    lapply(self$default_aes, rlang::eval_tidy, data = list(theme = theme))
```
hadley: This seems like the right approach to me, but I think I'd embed I do:

```R
from_theme <- function(name) {
  # plus error handling
  theme$geom[[name]]
}
```

Then `data = list(from_theme = from_theme)`

R/geom-defaults.r
hadley: Oh yeah. In that case we should pull this out into a function that can be shared with `eval_defaults()`

dpseidel:
Sorry for the long delay on my end, getting back to this now. I have just embedded and implemented `from_theme()` as suggested by @hadley and refactored `element_geom()` as suggested by @clauswilke. Unfortunately, with vdiffr now running properly, this PR will fail on multiple unrelated visual tests because of the way vdiffr adds it's own `theme_test()` to all plots without a specified theme. `vdiffr::theme_test()` does not recognize/specify the new `element_geom()` and thus serves `colour = NULL` to `scales::alpha()` which returns a `subscript out of bounds` error.  All non-visual tests pass. I need to add new tests, visual especially, but won't be able to do this without a workaround or solution to the `theme_test()` issue.  I am still working to refactor sf to handle themeable aesthetics and update documentation, examples, and tests.

clauswilke:
Dana, I think the solution to your vdiffr problem is in https://github.com/lionel-/vdiffr/issues/34, "Move theme_test to ggplot2".

Also, I assume that the theme changes you've been working on are out of scope for the upcoming 3.0.1 release, because they may break existing 3rd-party code. Therefore, the right order of things is probably:
- Move `theme_test()` to ggplot2 for 3.0.1
- Change vdiffr once ggplot2 3.0.1 is released (or make it depend on the development version)
- After 3.0.1, merge this PR into master

I have some theme changes pending that also don't fit into 3.0.1 (https://github.com/tidyverse/ggplot2/pull/2784), so maybe the release thereafter could do a major theme overhaul. I also had some thoughts about handling missing default elements better, but I don't remember whether I had integrated that already in my PR or not (I think not, though). There's also the idea of adding scales to the theme (#2691), and that could also be tackled for a major theme overhaul.

dpseidel:
Oh, interesting. Yes I suspected we would need to make a change in vdiffr but ggplot2 actually already has it's own `theme_test()` 
https://github.com/tidyverse/ggplot2/blob/5f49fb4a7b5624f9cc5b8011fc98909d7dbb23fe/R/theme-defaults.r#L500 
It's just not yet used by vdiffr, so presumably, I can open a PR there without waiting for the bug release.  

Ultimately though, I think unless we make vdiffr depend on the development version, there will be a stage before release where the visual tests break given the changes to `theme_test()` in this PR, even once vdiffr is using `ggplot2::theme_test()` just because this PR is building a whole new element tree. 

Regardless, as you say, this is most definitely outside the intended scope of 3.0.1, so there is no big rush I am just trying to keep chipping away at this.  I agree between open PRs and related issues, there is plenty to work on for a general theme overhaul and I would be happy to help.

dpseidel:
@thomasp85 has suggested I wait on refactoring `GeomSf` to handle themeable aesthetics given his intention to overhaul GeomSf performance in the near future. If there is interest in moving this forward before then, let me know. Below are some notes to jog my memory when we look at this again; they may also be helpful/interesting when approaching the overhaul.
__________
I've made some progress on this for tidy-dev-day but the sticking point continues to be the unique way `GeomSf` defaults are set.  Briefly, all other geoms evaluate their defaults during `Layer$compute_geom2()`. `GeomSf` uses separate internal functions `sf_grob()` and `default_aesthetics()` to pull aesthetics from either `GeomPolygon`,`GeomPoint`, or `GeomLine` according to the inherited geometry of each row in the given dataset, during the `Layer$build_geom` function. This causes a problem with themable aesthetics, since `Geom*$default_aes` now contains unevaluated expressions and `Layer$draw_geom()` is currently ignorant of `plot$theme` with which to evaluate them. 

For the record, @clauswilke helpfully suggested some passing sf unique defaults to `GeomSf$default_aes` that would then be evaluated appropriately in `compute_geom2` and could be specifically handled in the conditionals within `sf_grob()`, e.g.:
```r
sf_grob <- function(row, lineend = "butt", linejoin = "round", linemitre = 10) {
  # Need to extract geometry out of corresponding list column
  geometry <- row$geometry[[1]]

  if (inherits(geometry, c("POINT", "MULTIPOINT"))) {
    row <- modify_list(default_aesthetics("point"), row)
    col = alpha(
      row$colour %|% row$sf_point_colour,    # this does work exactly yet, since GeomPoint$default_aes$colour == NA
      row$alpha
    )
    gp <- gpar(
      col = col,
      fill = alpha(row$fill, row$alpha),
      # Stroke is added around the outside of the point
      fontsize = row$size * .pt + row$stroke * .stroke / 2,
      lwd = row$stroke * .stroke / 2
    )
    sf::st_as_grob(geometry, gp = gp, pch = row$shape)
  } else ...
```
I've done some local work with this suggestion but after speaking with @thomasp85, I intend to wait until he has completed his refactoring of GeomSf.

jcheng5:
Oh man, just stumbling onto this in PR now, it's exactly what we need for Shiny--we're working on a feature to automatically color ggplot2 plots to match with the web page's color scheme, and I was thinking I'd have to submit a PR to allow reverting changes to `update_geom_defaults`...

thomasp85:
I think this PR is a good candidate for the release after the next (which we are putting final touches on now)

clauswilke:
@thomasp85 Maybe create a milestone for the release after next?

dpseidel:
Hey team, I'm so glad there is still interest in this feature! I continue to think this would be really cool and am sorry I let it lie fallow for so long. 

I'm working on reviving this zombie PR now. Currently, it's breaking a lot of seemingly unrelated tests which I will work through shortly, but I would appreciate a fresh review of the general approach and scope since so much has changed internally since @hadley and I first began implementing this. CC: @clauswilke  @thomasp85

R/geom-.r
```diff
@@ -108,12 +108,24 @@ Geom <- ggproto("Geom",
 
   setup_data = function(data, params) data,
 
+  # evaluate defaults according to theme
+  eval_defaults = function(self, theme) {
+
+    if (length(theme) == 0) theme <- theme_grey()
```
clauswilke: Should this retrieve the currently set default theme instead of `theme_grey()`?

R/theme-elements.r
```diff
@@ -130,6 +131,38 @@ element_text <- function(family = NULL, face = NULL, colour = NULL,
...
+#'   typically a bright colour used for geom_smooth et al.
+#' @param fill_1 geom accent fill colour, typically a darker version of fill
+#' @param ... other accepted spellings of themable aesthetics, e.g. "color"
+#' @export
+#' @rdname element
+element_geom <- function(fill = NULL,
+                         fill_1 = NULL,
+                         colour = NULL,
+                         colour_1 = NULL,
+                         colour_2 = NULL,
```
clauswilke: I would add arguments `color`, `color_1`, `color_2` and use those if the `colour` versions are `NULL` (or the other way round, doesn't really matter).

R/theme-elements.r
```diff
@@ -491,7 +523,8 @@ el_def <- function(class = NULL, inherit = NULL, description = NULL) {
   plot.tag.position   = el_def("character"),  # Need to also accept numbers
   plot.margin         = el_def("margin"),
 
-  aspect.ratio        = el_def("character")
+  aspect.ratio        = el_def("character"),
+  geom                = el_def("element_geom", "geom")
```
clauswilke: Is the `"geom"` argument in `el_def()` correct? Doesn't this imply that `geom` inherits from `geom`?

R/layer.r
```diff
@@ -345,14 +345,15 @@ Layer <- ggproto("Layer", NULL,
     self$position$compute_layer(data, params, layout)
   },
 
-  compute_geom_2 = function(self, data) {
-    # Combine aesthetics, defaults, & params
+  compute_geom_2 = function(self, data, plot) {
     if (empty(data)) return(data)
 
+    # evaluate defaults for theme
+    defaults <- self$geom$eval_defaults(theme = plot$theme)
```
clauswilke: I'm not entirely sure that `plot$theme` is the complete theme when this is called. This needs to be verified.

dpseidel: Can confirm it's not! Which explains a lot of the test failures. Will fix.

R/geom-.r
dpseidel: yes. good catch. will fix.

dpseidel:
Hi all,  

As of these most recent commits, all geoms now respect default aesthetics set by themes :tada:. Currently I’ve only done this for color, fill, and some text aesthetics however, after reviewing and agreeing on the implementation, we should discuss the scope of aesthetics we want to make “themeable”.  

#### Some notes on current implementation: 
In order to get this to work for geom_sf (which dynamically chooses defaults not during `compute_geom2` like all other geoms but later during `draw_geom`), I needed to pull them through from plot build into draw_geom and LayerSf so defaults could be evaluated in their current context.

I originally attempted (viewable in a separate [`theme_geom-sfopt2`](https://github.com/dpseidel/ggplot2/tree/theme_geom-sfopt2) branch) to bring sf default evaluation more in line with the behavior of other geoms by providing all possible default configurations (those for point, polygon, line, & collection) as a list to default_aes so they could all be evaluated in the `eval_defaults` function and then writing some logic to choose the correct defaults dynamically in the `use_defaults` function. I made this work preliminarily but such a drastic change to structure of the default_aes object broke a number of things (e.g. parameter evaluation) and I ultimately decided that this approach required touching too many disparate parts of the API. I opted for what seemed to me to be the slightly more conservative approach but I would appreciate any and all thoughtful critiques on this implementation (ping @thomasp85, @clauswilke, @hadley) 

#### Some work still to be done
I still need to review/repair GeomSf `draw_key` function and currently defunct `default_aesthetics` function and decide where/how/if it’s necessary to reevaluate defaults in this context. 

More granular tests should absolutely be written and as I mentioned above, before this PR is merged there needs to be a larger discussion about scope and I suspect some work to implement further themeable defaults.

clauswilke:
I'm not sure I like the special casing of `draw_layer()` for `GeomSf`. We had at some point discussed handing more info about the plot into the `draw_layer()` function, in the context of making the scales available. I wonder whether both issues could be solved at once. @paleolimbot What was your final thought on getting the scales into `draw_layer()`?

clauswilke:
My thinking was simply: We now have two very different cases that suggest `draw_layer()` needs more info that in currently receives. Instead of developing weird workarounds, why not break the signature once and hand it the objects that seem relevant. (And maybe add `...` to future-proof the function.)

However, I don't remember the discussion about layer-specific scales. Would handing `draw_layer()` the layout object interfere with that?

thomasp85:
I think there are two things being conflated with "layer specific scales" (at least in my head). The discussion I remember entered around allowing the geom access to the scales used for the different aesthetics...

The other thing, which I have thought about a bit recently, is an API for setting scales on a per-layer base, such that two layer could use e.g. two different colour scales

clauswilke:
Let's move the discussion about the function signature of `draw_layer()` into a separate issue. I'm going to open one. If we decide to move forward with it, we can quickly merge a PR that makes this change, and then @dpseidel can use it in this PR here.

clauswilke:
Please use #3854 for the discussion of `draw_layer()`.

R/theme-elements.r
```diff
@@ -130,6 +131,38 @@ element_text <- function(family = NULL, face = NULL, colour = NULL,
   )
 }
 
+#' @param colour_1 geom accent colour 1,
+#'   typically a lighter version of colour
+#' @param colour_2 geom accent colour 2,
+#'   typically a bright colour used for geom_smooth et al.
+#' @param fill_1 geom accent fill colour, typically a darker version of fill
```
cpsievert: Have y'all considered that there are a handful of different levels of (grayscale) fill in ggplot2 alone (likely for a reason)?

```r
> geoms <- mget(ls(asNamespace("ggplot2"), pattern = "^Geom[A-Z]"), asNamespace("ggplot2"))
> unique(unlist(lapply(geoms, function(g) g$default_aes$fill)))
[1] "grey20" "grey35" "white"  NA       "black"  "grey50" "grey60"
```

With that in mind, a more robust interface would be something like `theme(geom = element_geom(bg = "white", fg = "black", accent = "#3366FF", ...))` and hold the geom responsible for mixing the background and foreground colors. For example, `GeomRect$default_aes$fill`  currently defaults to `"gray35"` (i.e., interpolate 35% of the way from `fg` to `bg`), so the default could be:

```r
GeomRect$default_aes$fill = scales::colour_ramp(c(from_theme('bg'), from_theme('fg')), alpha = TRUE)(0.35)
``` 

By the way, I've been thinking a lot about this sort of stuff recently for the new [wip auto-theming feature coming to shiny](https://github.com/rstudio/shiny/pull/2740), and I'm thinking of abstracting out some of that functionality  into a self-contained package so it could be used anywhere. Towards that end, having this PR, as well as #3828 and #3833, would be hugely helpful in avoiding terrible hacks to reach that goal

baptiste: In one of my earlier attempts at suggesting this geom-defaults-in-themes feature I floated the idea of extending `rel()` beyond sizes; one could think of defining a few main colours  — [a bit like a css-variables model](https://razorltd.github.io/sasscolourfunctioncalculator/) — and have functions to calculate derivative shades (e.g desaturated, complement, lighter/darker, etc.)

(The css analogy isn't entirely unmotivated as it seems likely more and more use-cases will want to style ggplots consistently with a surrounding webpage (cf Joe's comment on Shiny, for instance))

paleolimbot: (I implemented [something like this](https://github.com/paleolimbot/ggr6/blob/master/R/cascading-theme.R#L2) in my just-for-fun ggr6 package...you can pass a function as a theme element that transforms the parent!)

Issue #2674, opened 2018-05-30
scale_colour/fill aesthetic only works for British spelling


jvcasillas:
The latest addition of `aesthetics` to `scale_fill_*` and `scale_colour_*` are fantastic, but the colour geom only works with the British spelling. 

```
ggplot(iris, aes(x = Sepal.Length, y = Petal.Length, color = Sepal.Length, fill = Petal.Length)) +
  geom_point(shape = 21, size = 3, stroke = 2) +
  scale_color_viridis_c(name = "Length",
                         option = "B", aesthetics = c("color", "fill"))
```

Produces...

<img width="856" alt="screen shot 2018-05-30 at 09 26 06" src="https://user-images.githubusercontent.com/1747068/40722887-8589fbf2-63eb-11e8-924e-f2cafa737997.png">

When the desired output is...

<img width="860" alt="screen shot 2018-05-30 at 09 26 41" src="https://user-images.githubusercontent.com/1747068/40722912-98e490ea-63eb-11e8-8701-03667e265b8b.png">

clauswilke:
Yes, that's a limitation of the current code. The required changes are a bit more involved and won't make it into 2.3.0. I already opened an issue about this: #2649, so I'll close this one.

PR #2750, opened 2018-07-11
Rename aesthetics in scales and consistently convert US to British spelling

### LINKED ISSUE #2649: Improved support for aesthetic aliases
The current ggplot2 has support for aesthetic aliases, such as `color` instead of `colour`. However, there are two places where this support could be improved, and they should be tackled at the same time, I think.

## 1. Make aliases configurable

Currently the available aliases are hardcoded here: 
https://github.com/tidyverse/ggplot2/blob/eecc450f7f13c5144069705ef22feefe0b8f53f7/R/aes.r#L10-L23
I suspect most users are not even aware of some of those, e.g. `"srt" = "angle"`. More importantly, it is currently not possible to create new aliases. So, if a package creates a new aesthetic, e.g. `point_colour`, it cannot create an alias `point_color`. The solution would be to provide functions that can add (and subtract?) aliases to (from) the default list. I would like to point out that this is conceptually similar to #2540 (user-defined theme elements), and similar techniques and user interfaces could be used in both cases.

## 2. Apply aliases to scales as well

Aliases are not applied to `aesthetics` arguments of scales. This leads to confusing behavior such as the following:

``` r
library(ggplot2)
# works
ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species)) +
  geom_point() +
  scale_color_viridis_d(aesthetics = "colour")
```

``` r

# does not work
ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species)) +
  geom_point() +
  scale_color_viridis_d(aesthetics = "color")
```

Created on 2018-05-21 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

The solution is to rename aesthetics in the scales, just as it is done in the layers. There are only two places where this needs to happen, here:
https://github.com/tidyverse/ggplot2/blob/eecc450f7f13c5144069705ef22feefe0b8f53f7/R/scale-.r#L554-L571
and here:
https://github.com/tidyverse/ggplot2/blob/eecc450f7f13c5144069705ef22feefe0b8f53f7/R/scale-.r#L620-L632

clauswilke:
This PR closes #2649. It has two parts: (i) It reworks and simplifies renaming of aesthetic names. (ii) It uses the reworked code in the scale functions, to address issues such as #2674.

Part II is straightforward. Part I needs a few words of explanation.

Most importantly, I deleted code that supposedly does partial matching of aesthetic names:
https://github.com/tidyverse/ggplot2/blob/5f49fb4a7b5624f9cc5b8011fc98909d7dbb23fe/R/aes.r#L148-L150
As far as I can tell, this code does not do anything, and certainly does not do partial matching. The reason is that `match()` does not do partial matching:
``` r
match(c("col", "color"), "color")
#> [1] NA  1
```

Indeed, in my testing, the current ggplot2 does not do partial matching. It does convert `col` to `colour`, but only because that particular match is listed among the base-to-ggplot conversions:
``` r
library(ggplot2)
# col is renamed to colour. this is special-cased
aes(col = x)
#> Aesthetic mapping: 
#> * `colour` -> `x`

# colo is not renamed. there is no partial matching
aes(colo = x)
#> Aesthetic mapping: 
#> * `colo` -> `x`
```

Second, I've exported a new internal function, `standardise_aes_names()`. This function is needed by anybody who wants to implement their own scales and take advantage of aes standardisation.

With this patch, the following is possible:

``` r
library(ggplot2)
# Create a scale using base-R naming
ggplot(iris, aes(Sepal.Length, Sepal.Width, pch = Species)) +
  geom_point(size = 3) +
  scale_discrete_manual(aesthetics = "pch", values = c('+', "-", '*'))
```

``` r

# The previous example is equivalent to this:
ggplot(iris, aes(Sepal.Length, Sepal.Width, pch = Species)) +
  geom_point(size = 3) +
  scale_shape_manual(values = c('+', "-", '*'))
```

``` r

# The example from #2674 that doesn't work there works now
ggplot(iris, aes(x = Sepal.Length, y = Petal.Length, color = Sepal.Length, fill = Petal.Length)) +
  geom_point(shape = 21, size = 3, stroke = 2) +
  scale_color_viridis_c(name = "Length", option = "B", aesthetics = c("color", "fill"))
```

Created on 2018-07-11 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


R/aes.r
```diff
@@ -143,13 +143,37 @@ print.uneval <- function(x, ...) {
...
+  # convert US to UK spelling of colour
+  x <- sub("color", "colour", x, fixed = TRUE)
 
-  plyr::rename(x, ggplot_global$base_to_ggplot, warn_missing = FALSE)
+  # convert old-style aesthetics names to ggplot version
+  plyr::revalue(x, ggplot_global$base_to_ggplot, warn_missing = FALSE)
+}
+
+# x is a list of aesthetic mappings, as generated by aes()
+rename_aes <- function(x) {
```
hadley: Where does this get called from?

R/aes.r
```diff
@@ -143,13 +143,37 @@ print.uneval <- function(x, ...) {
...
+  names(x) <- standardise_aes_names(names(x))
+  duplicated_names <- names(x)[duplicated(names(x))]
+  if (length(duplicated_names) > 0L) {
+    duplicated_names_message <- paste0(
+      "`", duplicated_names, "`", collapse = ", "
+    )
+    warning(
```
hadley: Wasn't this previously the responsibility of `plyr::rename()`?  Shouldn't duplicated names also be checked in `standardise_aes_names()`? Should we also check for code like `aes(colour=1,colour=2,colour=3)`?

R/aes.r
clauswilke: The main user is the `aes()` function, but it's actually called from many places:
https://github.com/tidyverse/ggplot2/blob/17b45f906c3a55d187915f4f3010836a804051ae/R/aes.r#L77-L83
https://github.com/tidyverse/ggplot2/blob/528b0ce369df71dec2f7126e6a40640f025c390f/R/geom-defaults.r#L17
https://github.com/tidyverse/ggplot2/blob/429991746ef69d7d66b1b33bb4c24e89b521907f/R/labels.r#L57
https://github.com/tidyverse/ggplot2/blob/c1908f192db6f5b5937f38ae1f2fc0410e4be7d5/R/quick-plot.r#L86
https://github.com/tidyverse/ggplot2/blob/3129bf861dfbc4eeba6121eaef02dceabdb6dbda/R/guides-.r#L63
https://github.com/tidyverse/ggplot2/blob/17b45f906c3a55d187915f4f3010836a804051ae/R/layer.r#L99-L100
https://github.com/tidyverse/ggplot2/blob/17b45f906c3a55d187915f4f3010836a804051ae/R/guide-legend.r#L183

R/aes.r
clauswilke: Yes, this code was taken from `plyr::rename()`. Because I now have a function that converts names rather than a named list it didn't make sense to call `plyr::rename()` anymore. Also, this gives us the opportunity to customize the warning message to be more meaningful.

Whether duplicated names should be checked in `standardise_aes_names()` is not clear to me. I think an argument can be made either way. If the job of `standardise_aes_names()` is to simply take a vector of strings and standardize each of them then it doesn't necessarily have the job to check for duplicates. Maybe there's a downstream use case where somebody needs aesthetics names standardized without having duplicates checked. On the other hand, moving the check into that function would work just fine with the current code base.

Regarding the case of multiple identical aesthetics, it does get caught by the same check:
``` r
> aes(colour=1, colour=2, colour=3)
Aesthetic mapping: 
* `colour` -> 1
* `colour` -> 2
* `colour` -> 3
Warning message:
Standardisation of aesthetics has created duplicates for the following: `colour`, `colour` 
```
So maybe it's just a matter of making the error message a little clearer. I think printing out the complete list of final aesthetic names, rather than just the ones that are duplicated, might be helpful.

clauswilke:
I have improved the error message for duplicated aesthetics. Most importantly, now each duplicated aesthetic is at most listed once. I tried outputting the entire list of aesthetics but that was confusing.

``` r
> aes(shape = 1, pch = 2, colour = 1, colour = 2, colour = 3)
Aesthetic mapping: 
* `shape`  -> 1
* `shape`  -> 2
* `colour` -> 1
* `colour` -> 2
* `colour` -> 3
Warning message:
Duplicated aesthetics after name standardisation: shape, colour 
```

R/aes.r
```diff
@@ -143,13 +143,34 @@ print.uneval <- function(x, ...) {
...
+}
+
+# x is a list of aesthetic mappings, as generated by aes()
+rename_aes <- function(x) {
+  names(x) <- standardise_aes_names(names(x))
+  duplicated_names <- names(x)[duplicated(names(x))]
+  if (length(duplicated_names) > 0L) {
+    duplicated_message <- paste0(unique(duplicated_names), collapse = ", ")
+    warning(
+      "Duplicated aesthetics after name standardisation: ", duplicated_message, call. = FALSE
```
hadley: Should we make this an error?

clauswilke: I'll leave this to your decision. My own philosophy is it's only an error if the software cannot recover. Here, the duplicated aesthetics are simply ignored and a plot is produced:
``` r
library(ggplot2)
ggplot(df, aes(x, y, shape = "*", pch = "a")) + geom_point()
#> Warning: Duplicated aesthetics after name standardisation: shape
```



It's not that different from this case, which doesn't even create a warning:
``` r
ggplot(df, aes(x, y, shape = "*")) + geom_point(aes(pch = "a"))
```



Created on 2018-07-13 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

PR #2754, opened 2018-07-11
Fix cryptic error messages caused by missing aesthetics

### LINKED ISSUE #2637: Cryptic errors because `Geom*$setup_data` runs before `check_required_aes`
Failure to check aesthetics for geoms that use `setup_data` can cause cryptic error messages:
```
ggplot(iris) + geom_col()
## Error in pmin(y, 0) : object 'y' not found
## In addition: Warning messages:
## 1: In min(x, na.rm = na.rm) :
##  no non-missing arguments to min; returning Inf
## 2: In max(x, na.rm = na.rm) :
##   no non-missing arguments to max; returning -Inf
## 3: In min(diff(sort(x))) : no non-missing arguments to min; returning Inf
```
Compare to:
```
ggplot(iris) + geom_point()
## Error: geom_point requires the following missing aesthetics: x, y
```
This is because in [`Layer$compute_geom_1`](https://github.com/tidyverse/ggplot2/blob/a7b31356ef1de985fc23e3fee14ef5d2afd20c85/R/layer.r#L286) we `setup_data` before we `check_required_aesthetics`.

I can definitely see the rationale for this if it is the case that some required aesthetics are generated by `setup_data`, which I guess is possible but seems a bit odd.  After all, why would we require an aesthetic that would be auto generated?

The change is in theory trivial, but I can imagine it is possible there would be unintended consequences.  I can make the change, test to see if anything breaks, and so on, but probably not in the next couple of weeks.  If you're interested let me know and I'll work on a PR for the next release.

### LINKED ISSUE #2706: Feature Request:  Make Missing mapping Errors More Informative
When generating a plot recently, I accidentally left out the mapping.  When running the code, I got an uninformative error indicating that the mapping was missing.  Can the errors for missing mappings be improved to indicate something similar to "No mapping provided for required aesthetic 'x'" (and similar)?

``` r
library(ggplot2)         
ggplot(data.frame(A=1)) +
geom_line()              
#> Error in order(data$PANEL, data$group, data$x): argument 3 is not a vector
```

dpseidel:
Fixes #2637 ; Fixes #2706 

This is a minor change switching the order of calls in `layer$compute_geom_1` to call `check_required_aesthetics()` before `geom*$setup_data()` in order to catch missing aesthetics and offer informative error messages. 

This fixes cryptic error messages from `GeomArea`, `GeomCol`, `GeomCrossbar`, `GeomErrorbar`, `GeomErrorbarh`, `GeomLine`, `GeomRaster`, `GeomSpoke`, and `GeomTile`. 

From my analysis this should have no adverse side effects. 

The other Geoms'  were appropriately handled for one of two reasons:

1. their `setup_data()` functions simply return `data` unchanged and therefore did not call any potentially missing aesthetics which caused the cryptic errors before `check_required_aesthetics` was originally called (e.g. GeomPoint, GeomText, GeomRibbon, ...). 

Or,

2.  the missing aesthetics are caught by `stat_*$compute_layer()` and the appropriate warning returned earlier (e.g. GeomBoxplot, GeomSmooth, GeomHex, GeomBar...).

All tests pass and I have checked the new behaviour of all Geoms, but I am open to considering other ways to test this or potential side effects I have not foreseen. I was unsure if the NEWS bullet was warranted but as it fixes two user facing issues I figured it was potentially useful.


hadley:
Can you please turn one of the issues into a test? (i.e. test that the error is informative)

@clauswilke do you mind reviewing this? I suspect you are more familiar with this part of the ggplot2 code than me at the moment.

dpseidel:
✔️ test added!

clauswilke:
I had looked at this when #2637 was originally filed. My judgment then was that it's probably fine to switch the order but I don't understand the implications well enough to be 100% confident. I think it is reasonable to go ahead with the change and hope for the best. It might break some extension package somewhere out there, but it's probably not technically a breaking change, more of an API clarification.

As an extra check, I just looked at the code I have written in ggridges. It assumes that required aesthetics are checked before `setup_data()` is called. It's probably what most people would assume intuitively: You want to check your input, not what you do with your input.

Issue #2111, opened 2017-04-25
stat_sf(geom = "text")


cpsievert:
I'd expect something like this to work

```r
library(ggplot2)
library(sf)
library(albersusa)
usa_sf <- st_as_sf(usa_composite("laea"))
centroids <- st_centroid(usa_sf)

ggplot() + 
  geom_sf(data = usa_sf) + 
  stat_sf(data = centroids, aes(geometry = geometry, label = name), geom = "text")
```

hadley:
@edzer is there an obvious way to do this with `st_as_grob()`? I don't think there is.

edzer:
No, as `sf:::st_as_grob.POINT` simply calls `grid::pointGrob`. We could modify it such that it takes a `label` argument and then calls `grid::textGrob`; alternatively you could get the coordinates (as 2-column matrix) in ggplot by 
```r
> head(st_coordinates(centroids))
           X            Y
1 -1071416.4 -1116201.262
2   690503.2 -1091192.221
3 -1707964.5  -667016.439
4  -479407.9  -650634.740
5  2215964.8    -1222.643
6  1960178.5  -403145.999
```

hadley:
I had a go at implementing this but ultimately it seemed like it was going to overload `geom_sf()` with too many options. I think for this case it's probably better to extract x and y into their own columns and use `geom_text()` directly.

PR #2761, opened 2018-07-16
Implement geom_sf_label() and geom_sf_text()

### LINKED ISSUE #2742: Stat for ploting sf data as label and text
According to #2111, ggplot2 decided not to implement `stat_sf(geom = "text")` to avoid complexity.
But, I face the necessity to plot labels and texts on sf maps so often that I'm urged to make [a package](https://yutannihilation.github.io/ggsflabel/index.html)...

IIUC, it's not so complicated to implement a stat that retrieves the coordinates from a sf object and map the computed variables `X` and `Y`. Do you feel it's ok for this to live in ggplot2? If yes, I'm happy to make a PR. (If no, I will develop my package and submit it to CRAN.)

Here's an example implementation:

``` r
library(ggplot2)

StatSfCoordinates <- ggproto(
  "StatSfCoordinates", Stat,
    if (is.null(params$fun.geometry)) {
      params$fun.geometry <- sf::st_point_on_surface
    }
    params
  },
  compute_group = function(data, scales, fun.geometry) {
    points_sfc <- fun.geometry(data$geometry)
    coordinates <- sf::st_coordinates(points_sfc)
    data <- cbind(data, coordinates)
    
    data
  },
  
  default_aes = aes(x = stat(X), y = stat(Y)),
  required_aes = c("geometry")
)

ggplot(nc[1:3, ]) +
  geom_sf(aes(fill = AREA)) +
  geom_label(aes(geometry = geometry, label = NAME), stat = "sf_coordinates")
#> Warning in st_point_on_surface.sfc(data$geometry): st_point_on_surface may
#> not give correct results for longitude/latitude data
```

Created on 2018-07-08 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

yutannihilation:
This PR closes #2742.

## Background

It's very convenient that `geom_sf()` automatically chooses the proper geometries (point, line or polygon) based on the type of the simple feature in the data. But, we sometimes want to choose different type of geoms, for example, texts and labels.

However, adding flexibility to `geom_sf()` (or underlying `sf::st_as_grob()`) leads to unmaintainable complexity as discussed in #2111. So, we need to implement it as separate geoms/stats for code simplicity. Here, I propose `geom_sf_label()` and `geom_sf_text()`, and their corresponding stat `stat_sf_coordinates()`.

### `StatSfCoordinates`

Texts and labels need pairs of `x` and `y`. `StatSfCoordinates` is a stat to provide these via computed variable `x` and `y` (lower letters). `stat_sf_coordinates()` is the `stat_` function corresponding to this.

`X` and `Y` dimensions of  a `sf` object can be retrieved by `sf::st_coordinates()`. But, we cannot simply use `sf::st_coordinates()` texts and labels require exactly one coordinate per geometry whereas it returns multiple locations for a polygon or a line. Instead, these two steps are needed:

1. Choose one point per geometry by some function like `sf::st_centroid()` and `sf::st_point_on_surface()`.
2. Retrieve coordinates from the points by `sf::st_coordinates()`.

For the first step, users can pass arbitrary function via parameter `fun.geometry`. By default, `function(x) sf::st_point_on_surface(sf::st_zm(x))` is used; I think `sf::st_point_on_surface()` is more appropriate than `sf::st_centroid()` since labels and texts usually are intended to be put within the polygon or the line.

### `geom_sf_label()` and `geom_sf_text()`

`geom_sf_label()` and `geom_sf_text()` are basically same with `geom_label()` and `geom_text()` for each, except that:

* they uses `StatSfCoordinates` for `stat`
* they set `geometry` aesthetic automatically just as `geom_sf()` does
* they accept `fun.geometry` argument

### `Z` and `M` dimension

Simple feature objects may have `Z` and/or `M` dimension in addition to `X` and `Y`. These dimensions may be useful, but sometimes they are problematic:

* `M` dimension makes functions like `sf::st_point_on_surface()` and `sf::st_centroid()` fail.
* `Z` dimension may do no harm, but `Z` dimension is not always available in the results.

For simplicity, we simply drop `Z` and `M` and use only `X` and `Y`.

### Warning for longitude/latitude data

Functions that does calculation based on the distances may warn if the on data is in unprojected coordinate system. Though the warnings should not be suppressed, we need to be careful to use `geom_sf_label()` in examples. Otherwise, the tests fail.

## TODOs

- [x] Add tests
- [x] Write more docs about `stat_sf_coordinates()`
- [x] Write more docs about argument `fun.geometry`
    - [x] Write some notes about warnings for longitude/latitude data
- [x] Do we really support `Z` and `M` dimension? (*no*)

## Usages

``` r
library(ggplot2)

# texts and labels
p <- ggplot(nc[1:3, ]) +
  geom_sf(aes(fill = AREA))

p + geom_sf_text(aes(label = NAME), colour = "white")
#> Warning in st_point_on_surface.sfc(data$geometry): st_point_on_surface may
#> not give correct results for longitude/latitude data
```

``` r

p + geom_sf_label(aes(label = NAME))
#> Warning in st_point_on_surface.sfc(data$geometry): st_point_on_surface may
#> not give correct results for longitude/latitude data
```

``` r

# using the stat
p + stat_sf_coordinates(colour = "white", size = 10)
#> Warning in st_point_on_surface.sfc(data$geometry): st_point_on_surface may
#> not give correct results for longitude/latitude data
```

``` r

p +
  geom_errorbarh(
    aes(geometry = geometry,
        xmin = stat(x) - 0.1,
        xmax = stat(x) + 0.1,
        y = stat(y),
        height = 0.04),
    colour = "white",
    stat = "sf_coordinates"
  )
#> Warning in st_point_on_surface.sfc(data$geometry): st_point_on_surface may
#> not give correct results for longitude/latitude data
```

Created on 2018-08-19 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


yutannihilation:
This PR is ready to review now.

But, I don't see why Travis fails. especially, on R 3.1; it fails on `test-function-args`, which I think I've fixed. What's happening...?

clauswilke:
The sf package is missing for 3.1. For the more recent releases, I have noticed that Travis seems to have network issues today. If you restart a failed build it may pass the next time.

yutannihilation:
Thanks. Hmm, does absense of sf affects the function args? If so, I think I'm missing something. Anyway, will give it a try later.

yutannihilation:
OK, I think I found the cause. I've made two mistakes below, both of which are not appropriate when sf package is not installed.

- use `sf::st_point_on_surface()` as the default value of some functions' arguments
- reference sf's documents

yutannihilation:
Confirmed the build on 3.1 passed! Thank you for your advice, Claus.

(devel still fails due to the curl's "Connection timed out" error; I think I'm not lucky enough to let all 5 builds pass at the same time...)

yutannihilation:
> You can restart just one particular failed build by clicking on the reload icon to the right of it.

The icon doesn't show; it seems I don't have the permission, unfortunately.

yutannihilation:
(Disclaimer: I'm not a GIS expert but a guy who just wants to add labels on `geom_sf()`. So, I might be wrong, especially about technical phrasing...)

> 1. Have you considered using lower case for x, y, z, m? That seems more in line with typical ggplot2 naming.

In sf package (and probably in simple feature access), they are always written as capital. So, I felt it's more natural to use capital letters for computed variables so that we can distinguish them from ggplot's coordinates. That said, I'm still wondering whether they are essentially the same thing, `X` and `Y` dimensions of simple feature access, and `x` and `y` axis of ggplot2...

> 2. Have you considered returning NAs for z and m when those values are not available? Would that make things overall more predictable?

Thanks, this sounds nice.

> 3. I don't fully understand the comment about errors when z and m are not available. Does my point 2 address this or is there a different type of error that can occur?

Different. As described above, `StatSfCoordinates` takes these two steps and the error occurs in the first step:

1. Choose one point per geometry by some function like `sf::st_centroid()` and `sf::st_point_on_surface()`.
1. Retrieve coordinates from the points by `sf::st_coordinates()`.

Your suggestion is about when the first step succeeds but doesn't return `Z` and `M` value, but what I was thinking is when the calculation itself fails.

For example, consider we have a `sf` object of single `POLYGON` feature and want to determine a location to plot a label:

``` r
library(ggplot2)
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.2.3, proj.4 4.9.3

mat_poly_xy <- rbind(
  c(1, 1), c(-1, 1), c(-1, -1), c(1, -1), c(1, 1)
)
poly_xy <- st_polygon(list(mat_poly_xy))
poly_xy
#> POLYGON ((1 1, -1 1, -1 -1, 1 -1, 1 1))
```

In this case, we can choose a point on the polygon by `st_point_on_surface()` and extract the coordinates by `st_coordinates()`.

``` r
d_poly_xy <- st_sf(geometry = st_sfc(poly_xy))

d_pt_xy <- st_point_on_surface(d_poly_xy)
xy <- st_coordinates(d_pt_xy)
```

So we can combine the coordinates with the original data

``` r
cbind(d_pt_xy, xy)
#> Simple feature collection with 1 feature and 2 fields
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: 0 ymin: 0 xmax: 0 ymax: 0
#> epsg (SRID):    NA
#> proj4string:    NA
#>   X Y    geometry
#> 1 0 0 POINT (0 0)
```

But, if the polygon has `M` dimension, the function fails.

``` r
mat_poly_xym <- cbind(mat_poly_xy, 1)

poly_xym <- st_polygon(list(mat_poly_xym), dim = "XYM")

poly_xym
#> POLYGON M ((1 1 1, -1 1 1, -1 -1 1, 1 -1 1, 1 1 1))

d_poly_xym <- st_sf(geometry = st_sfc(poly_xym))

st_point_on_surface(d_poly_xym)
#> Error in CPL_geos_op("point_on_surface", x, numeric(0), integer(0), numeric(0), : GEOS does not support XYM or XYZM geometries; use st_zm() to drop M
```

clauswilke:
The Z and M issue still isn't entirely clear to me. Do I understand correctly that the default setup fails if the geometries contain Z or M coordinates? In this case, wouldn't it be better for the default to remove Z and M and then run `sf::st_point_on_surface()`? In other words, something like this:

``` r
#' @param fun.geometry
#'   A function that takes a `sfc` object and returns a `sfc_POINT` with the
#'   same length as the input. If `NULL`, `function(x) sf::st_point_on_surface(sf::st_zm(x))` will be
#'   used.
```

yutannihilation:
> Do I understand correctly that the default setup fails if the geometries contain Z or M coordinates?

Generally, yes. (It may not fail depending on what `fun.geometry` you use, though.)

Thanks, `function(x) sf::st_point_on_surface(sf::st_zm(x))` looks nice to me. For simplicity, now I'm thinking about supporting only `X` and `Y`, I mean:

* always drop `Z` and `M` by default (using `function(x) sf::st_point_on_surface(sf::st_zm(x))`)
* always return only `X` and `Y`
* use the lower case `x` and `y` for computed variables.

Does this look better to you?

clauswilke:
Sounds good to me.

yutannihilation:
Thanks, I will reflect this to the codes and the description.

yutannihilation:
I've made the changes I suggested in https://github.com/tidyverse/ggplot2/pull/2761#issuecomment-412380612. Could you review this again?

R/sf.R
```diff
@@ -249,6 +255,115 @@ geom_sf <- function(mapping = aes(), data = NULL, stat = "sf",
...
+  } else {
+    geometry_col <- "geometry"
+  }
+  if (is.null(mapping$geometry)) {
+    mapping$geometry <- as.name(geometry_col)
+  }
+
+  if (!missing(nudge_x) || !missing(nudge_y)) {
+    if (!missing(position)) {
+      stop("Specify either `position` or `nudge_x`/`nudge_y`", call. = FALSE)
```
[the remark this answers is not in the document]
yutannihilation: I just copied and pasted this line from here. Can I leave this for now so that we can change them both at the same time?

https://github.com/tidyverse/ggplot2/blob/c040a5b02ec87bb8d5b69be2e6aa1c8f56b2ebfb/R/geom-label.R#L20

R/stat-sf-coordinates.R
```diff
@@ -0,0 +1,111 @@
+#' Extract coordinates from 'sf' objects
+#'
+#' `stat_sf_coordinates()` extracts the coordinates from 'sf' objects and
+#' summarises them to one pair of coordinates (X and Y) per geometry. This is
```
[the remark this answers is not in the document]
yutannihilation: Ah, yes. Thanks!

yutannihilation:
Thanks for your comments! I've fixed them except for visual test (and the error message about nudges).

For visual test, I'm yet to figure out the workflow... Sorry that I didn't commit the reference images. I will address this later.

clauswilke:
> For visual test, I'm yet to figure out the workflow... Sorry that I didn't commit the reference images. I will address this later.

The workflow is:
- You run `vdiffr::manage_cases()`
- In the shiny app, you approve the images for your new tests
- The images are stored as svg files in a subdirectory under `tests/figs/`, and you add them to your commit

However, right now many reference images are broken, and you'll see all of those when you run `manage_cases()`. It's probably better to wait until that issue is resolved.

yutannihilation:
Thanks, I believe I know these basic steps. But, I meant, I didn't fully confident that I could generate reference images on my Windows (IIRC, the version of FreeType and Cairo matters, right?)... So, I wondered something like "do I have to use Docker? WSL?".

However, I think I can be less careful about this for now, given that the visual tests will be temporarily disabled. So, I will add reference images anyway. Thanks for your information!

clauswilke:
No, versions of freetype and Cairo don't matter anymore. However, you need the latest version of vdiffr. I recommend the github version from today.

yutannihilation:
Oh, cool! I see.

yutannihilation:
I have no idea why the Travis build still fails, but it seems unrelated to this PR.

yutannihilation:
Ah, do I have to wait for #2838 to be solved?

clauswilke:
I assume it's because of #2824. That needs to be resolved first.

clauswilke:
#2838 has been solved, #2824 not yet. #2824 is the problem.

yutannihilation:
I got it, Sorry that I don't keep up the discussion...

clauswilke:
#2824 is now resolved.

clauswilke:
Looks good to me. @hadley, any concerns from your side?

R/sf.R
```diff
@@ -6,6 +6,7 @@
 #' an unusual geom because it will draw different geometric objects depending
 #' on what simple features are present in the data: you can get points, lines,
 #' or polygons.
+#' For texts and labels, you can use `geom_sf_text` and `geom_sf_label`.
```
hadley: text -> text; functions should have `()` after them

R/sf.R
```diff
@@ -249,6 +255,115 @@ geom_sf <- function(mapping = aes(), data = NULL, stat = "sf",
   )
 }
 
+#' @export
+#' @rdname ggsf
+#' @inheritParams geom_label
+#' @inheritParams stat_sf_coordinates
+#' @seealso [stat_sf_coordinates()]
```
hadley: I think this seealso belongs in the documentation block above.

R/stat-sf-coordinates.R
```diff
@@ -0,0 +1,112 @@
+#' Extract coordinates from 'sf' objects
+#'
+#' `stat_sf_coordinates()` extracts the coordinates from 'sf' objects and
+#' summarises them to one pair of coordinates (x and y) per geometry. This is
+#' convenient when you draw an sf object as geoms like texts and labels (so
```
hadley: texts -> text (English is weird)

yutannihilation:
Thanks! I was very anxious if my English is strange, so the comments are very helpful...

For a news bullet, should I wait for the release of v3.0.1?

yutannihilation:
Sure!

yutannihilation:
Added a bullet.

yutannihilation:
@clauswilke Could you merge this if this is OK? Then I can create another PR to improve the error message you commented :)

yutannihilation:
Thanks!

adrfantini:
This is an extremely useful feature. Thanks!!!

clauswilke:
The merge broke the build. Could you look into what's going wrong? I'm in meetings all day.

yutannihilation:
Hmm, sorry if this is my fault. I will investigate.

yutannihilation:
Now it's all green. I definitely saw "error" status, but I can't find the errored build in the history. What happened...?

https://travis-ci.org/tidyverse/ggplot2/builds

yutannihilation:
Did I deleted my branch too early?

yutannihilation:
Anyway, it seems alright now.

PR #2781, opened 2018-07-27
Add new parameter "preset" for ggsave() to specify width/height at once

ilarischeinin:
This PR adds a new parameter `preset` for `ggsave()` that allows plot width and height (and units) to be specified at once. It recognizes common paper sizes (such as "a4" and "letter"), common screen resolutions ("4k"), and resolution specifications such as "1920x1080".

All presets default to landscape orientation, and portrait can be specified by appending an "r" (for rotated) to the end of the string. This is similar to the `paper` option for `grDevices::pdf()`, but with landscape being the default as it's probably a more common choice for plots.

(Naturally this new parameter could also be called `paper` instead of `preset`, but 1) that felt a bit too old-fashioned, and 2) might seem odd when making a plot that is to be placed on a sheet of paper together with other elements like text. `size` is another name I considered, but as the function already has a parameter called `limitsize`, I think they would be an odd pair. Anyways, `preset` is just one suggestion and I can rename it as wished.)

For now, some selected [paper sizes](https://en.wikipedia.org/wiki/Paper_size) and [screen resolutions](https://en.wikipedia.org/wiki/Computer_display_standard) are included, but I'm happy to make the list more comprehensive if that seems reasonable.

### Example Usage

```r
library(ggplot2)
p <- ggplot(data.frame(x = rnorm(100), y = rnorm(100)), aes(x, y)) + geom_point()

ggsave("letter.pdf", p, preset = "letter")
ggsave("a4r.pdf", p, preset = "a4r")
ggsave("hd.png", p, preset = "HD")
ggsave("1920x1080.png", p, preset = "1920x1080")
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,15 @@
     is now always internally converted to "colour", even when part of a longer
     aesthetic name (e.g., `point_color`) (@clauswilke, #2649).
 
+*   `ggsave()` has a new parameter `preset`, which allows plot `width` and
+    `height` (and `units`) to be specified at once. It recognizes common paper
+    sizes (such as "a4" and "letter"), common screen resolutions ("4k"), or a
+    resolution specification such as "1920x1080". All presets default to
+    landscape orientation, and portrait can be specified by appending an "r"
+    (for rotated) to the end of the string. This is similar to the `paper`
+    option for `grDevices::pdf()`, but with landscape being the default as it's
+    probably a more common choice for plots (@ilarischeinin).
+
 # ggplot2 3.0.0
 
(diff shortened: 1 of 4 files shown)


thomasp85:
@Hadley, can you comment on whether you'd want this in ggplot2? If you do, I'll give it a proper review

hadley:
This doesn't feel quite right for ggplot2 to me. It's a bit too big of a change to a very old function, and updating the preset list over time seems likely to add to the maintenance burden.

thomasp85:
Agreed.

PR #2782, opened 2018-07-27
Clarify geom_label documentation

gregrs-uk:
Specify non-support of '`angle` aesthetic' rather than '`rot` parameter', as this is clearer.

It is indeed the case that the `angle` aesthetic isn't supported. You can see below that `geom_text` is rotated but `geom_label` isn't, despite the same aesthetics. Perhaps it's worth adding a warning if you try to use the `angle` aesthetic with `geom_label`, although I'm less sure how you'd go about this.

Do I need to render the file for the `man/` directory?

``` r
library(tidyverse)
#> Warning: package 'dplyr' was built under R version 3.5.1

data <- tibble(
  text = c('one', 'two', 'three'),
  angle = c(0, 45, 90),
  x = c(1, 2, 3)
)

ggplot(data, aes(x)) +
  geom_text(aes(y = 2, label = text, angle = angle)) +
  geom_label(aes(y = 1, label = text, angle = angle))
```

Created on 2018-07-27 by the [reprex

### DIFF
--- a/R/geom-text.r
+++ b/R/geom-text.r
@@ -18,7 +18,7 @@
 #'
 #' @eval rd_aesthetics("geom", "text")
 #' @section `geom_label`:
-#' Currently `geom_label` does not support the `rot` parameter and
+#' Currently `geom_label` does not support the `angle` aesthetic and
 #' is considerably slower than `geom_text`. The `fill` aesthetic
 #' controls the background colour of the label.
 #'
--- a/man/geom_text.Rd
+++ b/man/geom_text.Rd
@@ -117,7 +117,7 @@ Learn more about setting these aesthetics in \code{vignette("ggplot2-specs")}.
 
 \section{\code{geom_label}}{
 
-Currently \code{geom_label} does not support the \code{rot} parameter and
+Currently \code{geom_label} does not support the \code{angle} aesthetic and
 is considerably slower than \code{geom_text}. The \code{fill} aesthetic
 controls the background colour of the label.
 }


dpseidel:
Can you rebuild the documentation and commit the new .Rd file please ? Then we are good to merge.

clauswilke:
I'm all for fixing the documentation so it agrees with what the code does and I support this PR. However, maybe we should also think about adding the ability to rotate labels? Shouldn't be that hard.

gregrs-uk:
Thanks @clauswilke. Shall I create a separate issue for this?

gregrs-uk:
When I run `devtools::document()`, I get a lot of errors like `Warning: @eval [geom-polygon.r#9]: unknown tag` and git shows that the 'Aesthetics' sections have been removed from the `.Rd` files. Am I doing something wrong?

I've only committed the relevant lines of `geom_text.Rd`, however, so this shouldn't pose a problem in this case.

gregrs-uk:
Thanks @batpigandme. I've installed the latest version via `devtools::install_github('klutometis/roxygen')` which has sorted the `eval` errors. There are still loads of changes shown by `git diff` but I think they're mostly line-length related whitespace changes.

Still my little change should be fine as I've only committed the relevant lines.

clauswilke:
I'll make an issue for improvements in `geom_label()`. I think there's some other things that need improvement also, but I need to investigate first.

clauswilke:
Issue filed (#2785).

PR #2786, opened 2018-07-29
Add facet variants called subplots

jiho:
Facets are meant to display different subsets of the same data.

However, with data in the "long" format (i.e. several variables stacked
in the same column and an additional column indicating which is which),
they can also be used to create the same plot for several variables
(e.g. a histogram). In that case, the scales are often free (the
variables are different and hence often do not have the same unit).

The problem with that approach is that the axis title is often a generic,
uninformative word (e.g. "values") while the real labels for the axis
are the facet labels, which are on the wrong side of the plot.

These functions use `facet_wrap` and `facet_grid` to create subplots and
adjust the position and aspect of the facet labels to make them look like
axis titles.


jiho:
Please see the examples in the documentation for use cases, but here is a quick demo:

```r
cars <- stack(mtcars, select=c(disp, drat, qsec))
# NB: did not want to depend on tidyr::gather for the examples
p <- ggplot(cars) + geom_histogram(aes(values), bins=10)
p + facet_wrap(~ind, scales="free_x")
```
old

```r
p + subplot_wrap(~ind)
```
new

clauswilke:
It seems to me this would fit better into an extension package (e.g., GGally or ggforce).

Regardless of where this could live, I'd like to mention that your TODO in the theme code could be a use case for this PR: https://github.com/tidyverse/ggplot2/pull/2784

jiho:
I thought this was a common enough use case and a simple enough change to make it into the main repo. I can submit this to ggforce though, if you think it would be more appropriate.

Actually, the best course of action would probably be an overhaul of facets because I suppose that all the various hacks that accumulated over the years (`switch`, `placement="outside"`, etc.) were in fact meant to allow this. In the usual case (facet=breakup of the same data according to a factor), the facet placement is just fine and placing them on the side of the axis makes little sense. So maybe having an argument such as `facet_is_variable`, equal to "x" or "y", that would trigger such a behaviour, and removing the previous variations around that theme would be best in the long term. However, I suppose that, as usual now, removing stuff is difficult ;-)

This would at least get the ball rolling showing use cases for such a change.

smouksassi:
Recently facets formula interface was augmented to accept rows and columns similarly we might think over the other direction by allowing a formula interface to the x / y interface?
The lattice package supported syntax like this:

```r
xyplot(uempmed+ psavert+unemploy~ date, data=economics) # you can make nicer

```
in ggplot2 you need the data to be in the long format first the question is whether we want to allow this based on the initial non stacked dataset ?
```r
ggplot(economics_long, aes(date, value)) +
 geom_line()+
facet_grid( variable~. ,scales="free_y") 

ggplot(economics, aes(formula= uempmed+ psavert+unemploy~ date)) +
 geom_line()+
facet_grid_stack( ) 
```
the yvalue and yvars would be then an internal computer variable that we can access later on
note that lattice also support y1+ y2 ~ x1+ x2 | facet1 ...

thomasp85:
FWIW I'll accept this functionality in ggforce

thomasp85:
I'm going to close this as this should be pursued in an extension package

PR #2791, opened 2018-07-31
Update theme documentation

dpseidel:
This is a PR to update the `theme()` documentation with feedback from #2727 and as a part of #2553 

includes:
- linking `element_*()` throughout doc
- documenting parameters together according to inheritance to reduce cognitive load
- updating the intro paragraph 
- linking to aesthetics vignette.


hadley:
Can you check out what the tooltip in RStudio looks like for autocomplete on the newly combined arguments? (and put a screenshot here)

Issue #2729, opened 2018-07-03
sec.axis on a log10 scale uses non-transformed breaks


jemus42:
I want to use a log10 y scale for dollar amounts, and a second axis for € conversion, but I can't figure out how to make the second axis have the "nice" breaks automatically used on the original axis. I thought `dup_axis` and the `derive()` helper would do the trick, but apparently not.

``` r
library(ggplot2)

df <- data.frame(x = rep(c("A", "B", "C"), 100),
                 y = sample(c(20, 20e6, 1e6, 50, 100), 300, replace = TRUE))

ggplot(data = df, aes(x, y)) +
  geom_boxplot() +
  scale_y_log10(labels = scales::dollar, sec.axis = sec_axis(~.*0.85))
```

![](https://i.imgur.com/8jXUVRy.png)

``` r

ggplot(data = df, aes(x, y)) +
  geom_boxplot() +
  scale_y_log10(labels = scales::dollar, sec.axis = dup_axis(~.*0.85))
```

![](https://i.imgur.com/UWeyD0P.png)

``` r

packageVersion("ggplot2")
#> [1] '2.2.1.9000'
```

Created on 2018-07-03 by the [reprex
package](http://reprex.tidyverse.org) (v0.2.0).

Edit:  
I have since tested this with ggplot2 v3.0.0 from CRAN, same result.

Please also see the discussion on [community.rstudio.com](https://community.rstudio.com/t/sec-axis-on-a-log10-scale-uses-non-transformed-breaks/10624)

batpigandme:
No solution yet — I've been kind of hacking away on this one trying to figure it out (I'm super confused by the fact that they don't line up even without the `scales::dollars()` transformation. 
But one change it occurred to me that _might_ be making a difference is that tidy evaluation tools are needed to compute on the mapping of existing ggplot2 objects (Kara Woo described this in an earlier iteration of the [release notes here](https://github.com/tidyverse/ggplot2/pull/2619/commits/a7153689e4a748df88da231c0ce9204146fd9419)).

Could be totally unrelated, just thought I'd put it out there. I'll keep on plugging away, so please let me know if you figure it out!
``` r
library(ggplot2)

df <- data.frame(x = rep(c("A", "B", "C"), 100),
                 y = sample(c(20, 20e6, 1e6, 50, 100), 300, replace = TRUE))

ggplot(data = df, aes(x, y)) +
  geom_boxplot() +
[...]

PR #2796, opened 2018-08-02
Simplify sec.axis for proper transformed breaks

dpseidel:
This PR simplifies the creation of secondary axes to use the transformation of the primary axis when setting breaks and labels rather than relying on an identity transform in all cases. This is a fix for #2729 which was caused by the identity trans calling `extended_breaks()` by default while the breaks on the primary axis were created by `log_breaks()` as is default for log-transformed axes. 

@thomasp85 As I mentioned this morning, my original fix just used a simple conditional to force log transformed axes to use `log_breaks()` but after you mentioned the identity trans this morning I decided to take another look at setting the new scale's `trans` argument by the primary transformation.  Ultimately this solves the problem at its source, and is a much more general fix. For instance, `dup.axis()` should now respect any custom breaks function fed to a custom transformation used on the primary axis. Plus it  drastically simplifies the code for building a secondary axis and makes implementation for `scale_*_date()`/`scale_*_datetime()` fairly straightforward. In fact I have already implemented sec.axis for these scales on a separate branch and will create a separate PR when this or some iteration of this is merged. Thanks again for making the time this morning, especially given my tardiness. 

I've added some new tests, examined numerous plots, and not yet found a case where this breaks anything or significantly changes behavior, but a careful review is warranted.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 3.0.0.9000
 
+*   `sec_axis()` and `dup_axis()` now return appropriate breaks for the secondary
+    axis when applied to log transformed scales (@dpseidel, #2729).
+
 *   `stat_contour()`, `stat_density2d()`, `stat_bin2d()`,  `stat_binhex()`
     now calculate normalized statistics including `nlevel`, `ndensity`, and
     `ncount`. Also, `stat_density()` now includes the calculated statistic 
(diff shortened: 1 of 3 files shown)


dpseidel:
This change also re-fixes #1992, thus #2095 could be effectively reversed, fixing the bug that it introduced into the monotonicity test for some transformed secondary axes. For example, currently this code no longer passes the monotonicity test although it should:
```r
# taken from test-sec-axis
x <- exp(seq(log(0.001), log(1000), length.out = 100))
foo <- data.frame(
  x = x,
  y = x/(1+x)
)

ggplot(foo, aes(x, y)) +
    geom_point() +
    scale_x_sqrt(sec.axis = dup_axis())
```

I can easily commit changes here fixing this, or in a separate PR after this current PR has been discussed and merged.

PR #2805, opened 2018-08-06
Fix incorrect monotonicity test failure in sec.axis

dpseidel:
PR #2095 changed the order in which the range a secondary axis was transformed, sampled, and tested for monotonicity. Sampling over the transformed space rather than the original scale caused the monotonicity test on the secondary axis to fail for certain transformations. This PR fixes those failures by sampling on the original scale rather than the transformed scale, effectively a partial reversal of PR #2095 whose motivating problem was fixed in an alternative way by #2796.

### DIFF
--- a/R/axis-secondary.R
+++ b/R/axis-secondary.R
@@ -113,10 +113,10 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
     if (self$empty()) return()
 
     # Get original range before transformation
-    along_range <- seq(range[1], range[2], length.out = self$detail)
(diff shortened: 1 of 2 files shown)


dpseidel:
Test failure is due to an invertibility issue in scales 0.5.0 `boxcox_trans()`, I've been using scales 1.0.0 (unreleased) locally. Will remove/alter test.

hadley:
@thomasp85 can you please review?

thomasp85:
Yes - I’ll give it a look tomorrow

PR #2807, opened 2018-08-07
Force arguments that are used in closures

krlmlr:
The reprex below triggered the investigation that led to possible instances where "brittle closures" are returned. I have no idea if the other instances I found have an effect, but it feels like good practice to force arguments that are used by closures.

Happy to add tests as necessary.

``` r
library(tidyverse)

scale <- c("red", "green", "blue")

p <-
  ggplot(iris, aes(Petal.Length, Petal.Width, color = Species)) +
  geom_point() +
  scale_color_manual(values = unlist(scale), labels = levels(iris$Species))

scale <- scale[-1]

p
```

> Created on 2018-08-07 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/R/save.r
+++ b/R/save.r
@@ -118,7 +118,10 @@ plot_dim <- function(dim = c(NA, NA), scale = 1, units = c("in", "cm", "mm"),
   dim
 }
 
-plot_dev <- function(device, filename, dpi = 300) {
+plot_dev <- function(device, filename = NULL, dpi = 300) {
+  force(filename)
+  force(dpi)
+
   if (is.function(device))
     return(device)
 
--- a/R/scale-manual.r
+++ b/R/scale-manual.r
@@ -102,6 +102,8 @@ scale_discrete_manual <- function(aesthetics, ..., values) {
 
 
 manual_scale <- function(aesthetic, values, ...) {
+  force(values)
+
   pal <- function(n) {
     if (n > length(values)) {
       stop("Insufficient values in manual scale. ", n, " needed but only ",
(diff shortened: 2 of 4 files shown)


clauswilke:
@krlmlr Do you want to look whether there are any other such cases or do we leave it as it currently is for this PR?

R/save.r
```diff
@@ -117,7 +117,10 @@ plot_dim <- function(dim = c(NA, NA), scale = 1, units = c("in", "cm", "mm"),
   dim
 }
 
-plot_dev <- function(device, filename, dpi = 300) {
+plot_dev <- function(device, filename = NULL, dpi = 300) {
```
clauswilke: For my own education, could you explain why we now need `filename = NULL` here but in some other cases (e.g., `make_summary_fun()`) we don't need to add a default?

krlmlr: The default doesn't make sense there to me, the function needs arguments or fails. But e.g. `manual_scale()` could use a better error message.

krlmlr:
This PR contains all instances I've found, looking for `^ +function` in the code and ignoring false positives. The initial comment describes just one instance. This is good to go from my end.

krlmlr:
NEWS entry:

- Functions that generate other functions (closures) now force the arguments that are used from the generated functions, to avoid hard-to-catch errors (#2807, @krlmlr).

hadley:
I don't think this warrants a news bullet as the chance of it affecting user code is extremely small.

clauswilke:
Thanks!

Issue #2801, opened 2018-08-03
preserve = "single" breaks violin plots


willgearty:
I'm trying to use preserve = "single" as you would with boxplots to maintain the sizes of the boxes when there are unused factors. While I figured it would work just the same way for violin plots, it appears to completely break them.

```r
p <- ggplot(mtcars, aes(factor(cyl), mpg))
#works fine
p + geom_violin(aes(fill = factor(vs)))
#works fine
p + geom_violin(aes(fill = factor(vs)), position = "dodge")
#this doesn't work
p + geom_violin(aes(fill = factor(vs)), position = position_dodge(preserve = "single"))
```

The latter attempt above produces this:

![image](https://user-images.githubusercontent.com/7232514/43668056-cbd28d00-972f-11e8-88ef-357fa18df14e.png)

clauswilke:
The problem is that `position_dodge()` tries to find the maximum number of groups by counting the maximum number of rows in the data with the same `xmin` value:
https://github.com/tidyverse/ggplot2/blob/4d2ca992dc37fee598d83d3d3f524dbba38f4187/R/position-dodge.r#L100-L101
This works for boxplots, because each boxplot is one line in the processed data. But it doesn't work for violin plots, because each violin plot consists of many lines (512 by default). I suspect the solution would be to count the number of identical `xmin` values with distinct `group` values.

Issue #2684, opened 2018-06-05
sf visual tests are unstable


lionel-:
Probably because of the sf dependencies:

```
✖ |  0 1     | coord_sf [1.0 s]
────────────────────────────────────────────────────────────────────────────
test-coord_sf.R:10: failure: multiplication works
Figures don't match: sf-polygons.svg

────────────────────────────────────────────────────────────────────────────


✖ |  1 1     | geom-sf [1.1 s]
────────────────────────────────────────────────────────────────────────────
test-geom-sf.R:13: failure: geom_sf draws correctly
Figures don't match: north-carolina-county-boundaries.svg

────────────────────────────────────────────────────────────────────────────
```

lionel-:
I think we should just fail silently when dependencies are outdated (and we should add scales to deps of ggplot2 plots). This way we don't overwhelm CRAN with spurious failures, but we still display helpful messages to the developer. Packages should always write tests for internal routines and data structures and those tests would fail verbosely on CRAN.

I think the conclusion of all this is that vdiffr is more of a monitoring mechanism than a unit testing tool.

lionel-:
Instead of a complicated dependency mechanism maybe we should fail silently by default unless an environment variable is set? Then we can set it in Travis to get verbose failures in pull requests.

PR #2813, opened 2018-08-08
make preserve = "single" more robust in position_dodge().

clauswilke:
This is an attempt at fixing #2801. Not sure this is ready to merge yet. One problem I see is that the violins are always moved to the left of the reference point, even if they should be moved to the right. Not sure how `position_dodge()` could know how to do this right.

Also, does a similar problem occur in `position_dodge2()`?

``` r
library(ggplot2)
p <- ggplot(mtcars, aes(factor(cyl), mpg))

# effect on geom_boxplot()
p + geom_boxplot(aes(fill = factor(vs)), position = "dodge")
```

``` r
p + geom_boxplot(aes(fill = factor(vs)), position = position_dodge(preserve = "single"))
```

``` r

# effect on geom_violin()
p + geom_violin(aes(fill = factor(vs)), position = "dodge")
```

``` r
p + geom_violin(aes(fill = factor(vs)), position = position_dodge(preserve = "single"))
```

Created on 2018-08-08 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/R/position-dodge.r
+++ b/R/position-dodge.r
@@ -97,7 +97,19 @@ PositionDodge <- ggproto("PositionDodge", Position,
       n <- NULL
     } else {
       panels <- unname(split(data, data$PANEL))
-      ns <- vapply(panels, function(panel) max(table(panel$xmin)), double(1))
+      ns <- vapply(
+        panels,
+        function(panel) {
+          # we need to find the maximum number of times that the
+          # same xmin is used, but making sure we count each group
+          # only once
+          if (is.null(panel$group)) return(1)
+          groups <- unname(split(panel, panel$group))
+          xmins <- vapply(groups, function(group) group$xmin[1], double(1))
+          max(table(xmins))
+        },
+        double(1)
+      )
       n <- max(ns)
     }
 
--- a/tests/figs/deps.txt
+++ b/tests/figs/deps.txt
@@ -1,2 +1,2 @@
-ggplot2: 3.0.0.9000
+ggplot2: 3.0.0
 vdiffr-svg-engine: 0.9000
--- a/tests/figs/deps.txt
+++ b/tests/figs/deps.txt
@@ -1,2 +1,2 @@
-ggplot2: 3.0.0
+ggplot2: 3.0.0.9000
 vdiffr-svg-engine: 0.9000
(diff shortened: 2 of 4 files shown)


clauswilke:
Also, one completely unrelated comment: Are visual tests properly run on Travis? I had two unrelated visual tests fail due to outdated reference images. The failure seems to have been caused by recent work on dual axes. @hadley @lionel- @dpseidel

lionel-:
Are these tests about `sf`? If yes you can ignore them, see https://github.com/tidyverse/ggplot2/issues/2684

The tests should either be removed or refactored so they don't depend on the system versions of dependent libraries.

clauswilke:
@lionel- No, they're not. It looks to me like changes in the secondary axis code caused small changes in the positioning of axis ticks:
https://github.com/tidyverse/ggplot2/pull/2813/files#diff-a37729bdac6b2e51856ab8cfb47331c7

lionel-:
(or revert it and push)

karawoo:
> One problem I see is that the violins are always moved to the left of the reference point, even if they should be moved to the right. Not sure how position_dodge() could know how to do this right.

Yeah this has been reported for bars in the past (#2076) and I don't see a way to fix it in `position_dodge()`.

> Also, does a similar problem occur in position_dodge2()?

`position_dodge2()` doesn't work at all for violins or dot plots, at the time I wrote it we decided to punt on making it compatible with those geoms.

clauswilke:
@lionel- I reverted the patch. All checks passed.

clauswilke:
@lionel- Just to reiterate, the fact that all checks passed with the visual test case reverted implies something isn't right. Either of the two builds (with or without the revised test case) should have failed. I'll open a separate issue to discuss.

hadley:
I'd rather not make any more `position_dodge()` changes without thoroughly thinking them through. This might be a project for @thomasp85 to take on — IIRC `position_dodge()` and `position_dodge2()` are quite slow, and I think the algorithms are more elegantly expressed in C/C++.

clauswilke:
That's fine. Should I leave this PR open as a reminder or close? #2801 remains open.

hadley:
Close, I think

Issue #2820, opened 2018-08-10
geom_hline() / geom_vline() don't work with geom_sf()


clauswilke:
Reprex below. Primarily reporting this to show the PR I'm about to make didn't mess this up. It's already broken. :-)

``` r
library(ggplot2)
nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"))
#> Reading layer `nc' from data source `/Library/Frameworks/R.framework/Versions/3.5/Resources/library/sf/shape/nc.shp' using driver `ESRI Shapefile'
#> Simple feature collection with 100 features and 14 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> epsg (SRID):    4267
#> proj4string:    +proj=longlat +datum=NAD27 +no_defs

# works
ggplot(nc) +
  geom_sf() + 
  geom_line(
    data = data.frame(x = c(-80, -80), y = c(34, 36.5)),
    aes(x, y),
    color = "red"
  )
```

![](https://i.imgur.com/jeKZ46e.png)

``` r

# doesn't work
ggplot(nc) +
  geom_sf() + 
  geom_hline(yintercept = 35)
#> Error in unit(x0, default.units): 'x' and 'units' must have length > 0

# doesn't work
ggplot(nc) +
  geom_sf() + 
  geom_vline(xintercept = -80)
#> Error in unit(y0, default.units): 'x' and 'units' must have length > 0
```

Created on 2018-08-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

clauswilke:
The cause of this problem is that `coord_sf()` doesn't implement its own `range()` function but names the ranges in the panel parameters `x_range` and `y_range` instead of `x.range` and `y.range`, so the default range function fails and returns `NULL`. I think a fix should be folded into #2821.

``` r
library(ggplot2)
nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"))
#> Reading layer `nc' from data source `/Library/Frameworks/R.framework/Versions/3.5/Resources/library/sf/shape/nc.shp' using driver `ESRI Shapefile'
#> Simple feature collection with 100 features and 14 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
[...]

Issue #2812, opened 2018-08-08
coord_trans() causes geom_hline() to draw incomplete lines


mafed:
`geom_hline` draws incomplete lines when used together with `coord_trans`, regardless of their respective order of addition.

```r
mydf <- data.frame(x = 0L:20,
                   y = -10:10)
ggplot(data = mydf, aes(x, y)) + 
  geom_point() + 
  coord_trans(x = "log1p") +
  geom_hline(yintercept = c(-10, 10))
```
![image](https://user-images.githubusercontent.com/2510765/43846354-6e65e25e-9b2e-11e8-93a8-f30af8cd43bb.png)

clauswilke:
Duplicate of #2149?

PR #2821, opened 2018-08-10
Fix back-transformation of ranges in coords

### LINKED ISSUE #2149: Double coordinate transformations for geoms that draw with other geoms
When a `geom` queries the coordinate range and uses the result to create data that is then plotted by another `geom`, there is potentially a double transformation. i.e. `coord.range()` yields results in coordinate space and if followed by `coord.transform()`, the data may get transformed two times. Affects `coords` that transform the limits in `coord$setup_panel_params` and also do transforms in `coord$transform`. This includes `coord_trans` and `coord_sf`.

Affected geoms include `geom_abline`, `geom_hline` and `geom_hline`.

**reprex**
``` r
library(ggplot2)

df <- data.frame(x=1:6, y=10^(1:6))
ggplot(df, aes(x, y)) +
  geom_point() +
  geom_vline(xintercept=3, color='red') +
  coord_trans(y='log10')
```

The red line (at x=3) should be top to bottom.

clauswilke:
This is a starting point for a conversation of how to fix #2149 and #2812. I realize the regression tests don't pass yet.

The key to fixing these bugs is to recognize that the `range()` function in `Coord` is meant to back-transform transformed ranges into data coordinates. It is only ever used in this way:
```
wilke$ grep -e '$range(' R/*
R/coord-munch.r:  ranges <- coord$range(range)
R/geom-abline.r:    ranges <- coord$range(panel_params)
R/geom-hline.r:    ranges <- coord$range(panel_params)
R/geom-vline.r:    ranges <- coord$range(panel_params)
```
and the API only makes sense with this interpretation: The function takes panel ranges as argument and returns ranges. If it doesn't back-transform, what else should it do? Also, it is written in this way for `coord_flip()` and `coord_polar()` already.

To make the meaning of the function more explicit, I have renamed it from `range()` to `backtransform_range()`. However, this is a breaking change if any geoms out in the world use this function. Alternatively, we could keep the current (confusing) name and just document better.

Reprex with fixed examples follows below. These geoms don't work at all yet with `coord_sf()`, but there appears to be a separate issue that needs to be fixed first (#2820).

``` r
library(ggplot2)
library(scales)
library(tibble)

# original example from #2149
df <- data.frame(x=1:6, y=10^(1:6))
ggplot(df, aes(x, y)) +
  geom_point() +
  geom_vline(xintercept=3, color='red') +
  coord_trans(y='log10')
```

``` r

# original example from #2812
mydf <- data.frame(x = 0L:20,
                   y = -10:10)
ggplot(data = mydf, aes(x, y)) + 
  geom_point() + 
  coord_trans(x = "log1p") +
  geom_hline(yintercept = c(-10, 10))
```

``` r

# original example from #2340
log_modulus_trans <- function() 
  trans_new(name = "log_modulus", 
            transform = function(x) sign(x) * log(abs(x) + 1), 
            inverse = function(x) sign(x) * ( exp(abs(x)) - 1 ))

set.seed(1)
d <- data_frame(
  tt = rep(1:10, 3),
  cc = rep(LETTERS[1:3], each = 10),
  xx = c(rnorm(10, mean = 100, sd = 10), 
         rnorm(10, mean = 0, sd = 10),
         rnorm(10, mean = -100, sd = 10)))

ggplot(data = d, 
       mapping = aes(x = tt, y = xx, group = cc)) +
  geom_line() + 
  geom_vline(xintercept = 5) +
  coord_trans(y = "log_modulus")
```

``` r

# examples with geom_line(), from #2149
df <- data.frame(x = c(-Inf, Inf), y = c(1, 2))
p <- ggplot(df, aes(x, y)) + geom_line() + xlim(1, 2)
p + coord_polar()
```

``` r
p + coord_trans(y = "log10")
```

``` r
p + coord_trans(x = "log10")
```

Created on 2018-08-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,9 +12,10 @@
     (@clauswilke, #2846).
     
 *   `Coord` objects now have a function `backtransform_range()` that returns the
-    panel range in data coordinates. This change may affect developers of custom coords. It
-    may also affect developers of custom geoms that use the `range()` function. In
-    some applications, `backtransform_range()` may be more appropriate.
+    panel range in data coordinates. This change may affect developers of custom coords,
+    who now should implement this function. It may also affect developers of custom
+    geoms that use the `range()` function. In some applications, `backtransform_range()`
+    may be more appropriate.
     (@clauswilke, #2821).
 
 *   `geom_sf()` now respects `lineend`, `linejoin`, and `linemitre` parameters 


clauswilke:
@hadley Could you comment on the naming choice of the `range()` function (keep name or change to something like `backtransform_range()`)? I can then proceed accordingly and fix the regression tests and #2820.

R/coord-.r
```diff
@@ -73,8 +73,12 @@ Coord <- ggproto("Coord",
     )
   },
 
-  range = function(panel_params) {
-    return(list(x = panel_params$x.range, y = panel_params$y.range))
+  # transform range given in transformed coordinates
+  # back into range in given in (possibly scale-transformed)
+  # data coordinates
+  backtransform_range = function(panel_params) {
+    warning("range backtransformation not implemented in this coord; plot may be wrong.")
```
hadley: Needs `call. = FALSE`

clauswilke:
I have improved the coord documentation, implemented `backtransform_range()` for all coords, implemented a deprecated `range()` function that prevents the code of any potential downstream users from breaking, and fixed the regression tests.

clauswilke:
All fixes and documentation improvements minus the API change are now in PR #2832, which I think should be merged first. Once that is done I will rebase this PR.

hadley:
I made a new label to (hopefully) remind us not to merge until we're done with the bug fix release.

wch:
Is it possible to get this in the bug-fix release? The API change from #2832 causes a problem with plot interaction in Shiny, and this PR would fix it. We'd prefer not to add in a workaround for a single version of ggplot2 that has a different API.

(@schloerke Please feel free to correct me if I've gotten details wrong.)

clauswilke:
@wch @schloerke Could you file a separate ggplot2 issue explaining exactly what the problem is. I looked at https://github.com/rstudio/shiny/pull/2184 and it's not immediately clear to me why this PR would fix the problem. More fundamentally, I'm concerned that there's a problem with the current ggplot2 that would be resolved with this PR. It suggests to me we haven't fully thought this through.

clauswilke:
@hadley Could you review this once more? I think this is now as much as possible a non-breaking commit:

- The `range()` function behaves as before for all coords.
- The `summarise_layout()` function behaves as before for all coords.
- The new `backtransform_range()` function uses `range()` as fallback (with warning) if a custom coord does not implement it.

With these changes, I think it is better to merge this before 3.0.1 than not, since it fixes #2895.

I have also provided documentation for the `summarise_layout()` function so in the future we will know what it does.

PR #2832, opened 2018-08-17
Fix back-transformation of ranges in coords, without API cleanup

### LINKED ISSUE #2149: Double coordinate transformations for geoms that draw with other geoms

clauswilke:
This PR fixes #2149 and #2812. It is essentially the same as PR #2821, except here the `range()` function is not renamed. This makes this a non-breaking fix that should be good for 3.0.1. I'm keeping PR #2821 open for its API cleanup, which I think is still appropriate but should be done later.


clauswilke:
@hadley Regarding spurious documentation changes, see #2833. I think the simplest way forward will be to merge that PR and then rebase this one.

clauswilke:
Spurious documentation changes are removed.

PR #2847, opened 2018-08-22
allow geom_sf to accept line parameters (#2826)

### LINKED ISSUE #2826: geom_sf does not pass parameters through to layer
Due to [this SO question](https://stackoverflow.com/q/51812311/4497050), I discovered that while `geom_sf` does pass aesthetics through to the appropriate layer (here to plot linestrings as paths), e.g. `size`, it doesn't pass parameters like `lineend`, `linemitre`, `linejoin`, and `arrow`, which are parameters of `geom_path`, not aesthetics like `size` or `linetype`. In this case, this makes getting rid of the gaps in this plot harder:

``` r
# Reprex from SO question linked above
library(sf)
library(osmdata)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright
library(ggplot2)

# define bounding box for osm data
my_bbox <- matrix(
    c(-0.2605616, -0.2605616,
      -0.2004485, -0.2004485,
      -0.2605616, 51.4689943,
      51.4288980, 51.4288980,
      51.4689943, 51.4689943),
    ncol = 2
)
bbox_sf <- st_geometry(st_polygon(x = list(my_bbox)))
st_crs(bbox_sf) <- 4326

# get osm road data for bounding box
osm_roads_secondary_sf <- 
    opq(bbox = st_bbox(bbox_sf)) %>%
    add_osm_feature(key = 'highway', value = 'secondary') %>%
    osmdata_sf()

ggplot(osm_roads_secondary_sf$osm_lines) + 
    geom_sf(size = 4, lineend = "round")
#> Warning: Ignoring unknown parameters: lineend
```

According to `?geom_sf`, these parameters get passed through `...` to the paired geom/stat or `layer`, though that function doesn't have parameters for everything or dots—I'm assuming they're passed to its `params` parameter. The problem here seems to be that the paired stat here is `stat_sf`, which does not have the specified parameters and is not passing them along further.

Ultimately this behavior is a product of the fact that `geom_sf` dispatches to so many...geoms? As a sort of meta-geom it's more strongly opinionated than the rest of the API, making it hard to, say, use sf points as locations for `geom_text` without pulling out the coordinates. Guidance in the docs on what sf geometry type dispatches to what geom (?) (and whether there's a way to override that dispatching) and thus what parameters can be passed would be welcome.

alistaire47:
This closes #2826, allowing `lineend`, `linejoin`, and `linemitre` arguments to be passed through `geom_sf`. I copied the defaults directly from `geom_path`, so this shouldn't alter the results of current code.

The only change I had to make beyond what Claus mentioned on the issue was assigning the parameters to `coord` so they get passed through to `sf_grob`. There may be a simpler way to structure that. If so, let me know and I'll refactor.

For now, I decided against adding them as explicit parameters to `geom_sf`, as while it would make sense for lines, it would sound like they'd apply to polygons, too, which they wouldn't, and thus may cause more confusion than it'd remove.

This builds and performs fine. I didn't add any tests, as the `geom_sf` tests don't currently test its line behavior, but can if you like. I do get an error when checking due to the `economics_long` dataset. I didn't touch it, but it gives me a C++ error when printed, even though `str` works fine: 

```r
> economics_long
Error: Not compatible with STRSXP: [type=list].
```
I suspect this means I've got a dependency issue with tibble or something, but I can't figure out what. I presume that if it has been working in CI builds, it will continue to do so.

Apologies for the extra commits in there; I realized too late that I didn't rebase fully. They shouldn't do anything at this point, so if there's some git-fu to clean them out, I'd be happy to remove them.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 3.0.0.9000
 
+*   `geom_sf()` now respects `lineend`, `linejoin` and `linemitre` parameters 
+    for lines and polygons (@alistaire47, #2826)
+
 *   `benchplot()` now uses tidy evaluation (@dpseidel, #2699).
     
 *   `fortify()` now displays a more informative error message for
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # ggplot2 3.0.0.9000
 
-*   `geom_sf()` now respects `lineend`, `linejoin` and `linemitre` parameters 
+*   `geom_sf()` now respects `lineend`, `linejoin`, and `linemitre` parameters 
     for lines and polygons (@alistaire47, #2826)
 
 *   `benchplot()` now uses tidy evaluation (@dpseidel, #2699).
--- a/R/sf.R
+++ b/R/sf.R
@@ -159,11 +159,13 @@ GeomSf <- ggproto("GeomSf", Geom,
 
     # Need to refactor this to generate one grob per geometry type
     coord <- coord$transform(data, panel_params)
-    coord$lineend <- lineend
-    coord$linejoin <- linejoin
-    coord$linemitre <- linemitre
     grobs <- lapply(1:nrow(data), function(i) {
-      sf_grob(coord[i, , drop = FALSE])
+      sf_grob(
+        coord[i, , drop = FALSE],
+        lineend = lineend,
+        linejoin = linejoin,
+        linemitre = linemitre
+      )
     })
     do.call("gList", grobs)
   },
@@ -190,7 +192,7 @@ default_aesthetics <- function(type) {
   }
 }
 
-sf_grob <- function(row) {
+sf_grob <- function(row, lineend, linejoin, linemitre) {
   # Need to extract geometry out of corresponding list column
   geometry <- row$geometry[[1]]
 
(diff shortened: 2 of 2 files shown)


clauswilke:
Looks good to me. I have only one comment: Since `sf_grob()` is an internal function used only at that one point in the code, the `lineend`, `linejoin`, and `linemitre` could also simply be handed through as arguments to that function. Not sure whether there's a strong argument for or against that approach versus yours.

The extraneous git commits will disappear when we squash-merge, so don't worry about them.

clauswilke:
Sorry, one more comment: Could you add a news bullet to [NEWS.md](https://github.com/tidyverse/ggplot2/blob/master/NEWS.md) with a brief description and your github user name and issue number?

alistaire47:
Oh, I think that's better; those `coord$lineend <- lineend` lines feel like un-R-like boilerplate code.

The arguments will still get passed to `sf_grob` for point/multipoints, but ignored due to the logic therein, so it seems fine. I also noticed that the parameters get respected for polygons, which was unexpected but nice:

``` r
library(ggplot2)

ggplot(sf::st_polygon(list(matrix(c(0, 1, 2, 0, 0, 2, 1, 0), ncol = 2)))) + 
    geom_sf(size = 3, linejoin = 'mitre', fill = 'pink')
```



``` r
ggplot(sf::st_polygon(list(matrix(c(0, 1, 2, 0, 0, 2, 1, 0), ncol = 2)))) + 
    geom_sf(size = 3, linejoin = 'bevel', fill = 'lightgreen')
```

clauswilke:
Apologies for being picky. Would you mind adding a comma after `linejoin` in the news bullet. For context, see here: https://github.com/tidyverse/ggplot2/pull/2669#issuecomment-414156779

alistaire47:
Hah, that's now the smallest commit I've ever made.

PR #2856, opened 2018-08-25
Improved vdiffr setup

clauswilke:
As discussed under #2842. I have purposefully committed a breaking visual test, so if everything is right with this initial PR then the release build will fail and all others will succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,15 @@ cache: packages
 sudo: true
 dist: trusty
 
-# build matrix; skip vdiffr on all but r release
+# build matrix; turn on vdiffr only on r release
 matrix:
   include:
-  - r: 3.1
-    env: SKIP_VDIFFR=true
-  - r: 3.2
-    env: SKIP_VDIFFR=true
-  - r: oldrel
-    env: SKIP_VDIFFR=true
-  - r: release
   - r: devel
-    env: SKIP_VDIFFR=true
+  - r: release
+    env: VDIFFR=true
+  - r: oldrel
+  - r: 3.2
+  - r: 3.1
 
 # work around temporary travis + R 3.5 bug
 r_packages: devtools
@@ -25,7 +22,10 @@ r_github_packages: lionel-/vdiffr
 # environment variables set for all builds
 env:
   global:
+  # don't treat missing suggested packages as error
   - _R_CHECK_FORCE_SUGGESTS_=false
+  # signal that this is a continuous integration run
+  - CI_RUN=true
 
 after_success:
   - Rscript -e 'covr::codecov()'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,18 @@ cache: packages
 sudo: true
 dist: trusty
 
-# build matrix; r release has custom env settings
+# build matrix; skip vdiffr on all but r release
 matrix:
   include:
+  - r: 3.1
+    env: SKIP_VDIFFR=true
+  - r: 3.2
+    env: SKIP_VDIFFR=true
+  - r: oldrel
+    env: SKIP_VDIFFR=true
   - r: release
-    env: RUN_VDIFFR=true
-
-# all other builds are not customized
-r:
-- 3.1
-- 3.2
-- oldrel
-- devel
+  - r: devel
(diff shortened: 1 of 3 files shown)


clauswilke:
@hadley @lionel- Could you take a look at this when you have a moment? The build is breaking *only* for R release on Travis, so that's the only place now where the visual tests are actually run.

With this setup, can we re-enable tests based on sf, or are there additional concerns?

lionel-:
Should we keep it on r-devel? This way we can monitor changes to graphics devices etc.

> can we re-enable tests based on sf, or are there additional concerns?

I found these tests to be dependent on system versions of sf dependencies (gdal, geos). If possible, they should be changed (for instance to use fabricated data) so they are not affected by these system libraries.

clauswilke:
If sf tests depend on gdal and geos (makes sense, now that I think about it) then anything involving a coordinate transformation or other calculation on geospatial data (e.g., centroid calculation, graticules) is likely going to be unstable. We can probably visually test basic drawing of simple features but not much else.

I’ll let @hadley make the call on whether visual tests should run on r-devel.

hadley:
The problem with checking on R devel is that any breaking r devel change will show up on a unrelated diff. I think testing with r release is likely to cause fewer false positives

clauswilke:
I have fixed the broken reference image and otherwise left the PR as is. I'll leave the problem with sf tests for now; I don't know how to resolve it. This means the PR is ready to be merged from my side.

lionel-:
@hadley I was thinking getting the r-devel failures is desirable because it allows us to monitor the effects of upstream changes on ggplot2.

hadley:
Yes, but we get them in a very un-informative way - it will all depend on the timing of when the travis request is kicked off. If we cared about r-devel changes (which I don't know if we do), it would be better to do a daily run of CRAN ggplot2 against latest r-devel

.travis.yml
```diff
@@ -3,24 +3,26 @@ cache: packages
...
-- 3.1
-- 3.2
-- oldrel
-- release
-- devel
+# build matrix; skip vdiffr on all but r release
+matrix:
+  include:
+  - r: 3.1
+    env: SKIP_VDIFFR=true
```
hadley: Can we flip this around to make skipping the default, so we only set `VDIFFR=true` for R-release?

.travis.yml
```diff
@@ -3,24 +3,26 @@ cache: packages
...
+  - r: 3.2
+    env: SKIP_VDIFFR=true
+  - r: oldrel
+    env: SKIP_VDIFFR=true
+  - r: release
```
hadley: Since you're in here, would you mind re-organsing so the versions are in reverse chronological order? That way the most recent versions (which are mostly likely to reveal problems) will be run first

.travis.yml
hadley: Oh that makes sense. Maybe we could turn off by default when the TRAVIS (IIRC) env var is ?

clauswilke: I find having a separate variable we define slightly preferable, because that means vdiffr runs can be fully controlled in any testing environment where these tests may be run, whether on travis or not.

hadley: I was thinking we could do both - if `USE_VDIFFR` is set, use its value, otherwise look at `TRAVIS` and don't run vdiffr if set.

clauswilke: Ok, this seems to be working now. I put the broken image back in as a check to see which builds run vdiffr.

.travis.yml
clauswilke: Done.

.travis.yml
```diff
@@ -3,26 +3,26 @@ cache: packages
...
 r_packages: devtools
 
 # Install manually because dev vdiffr is not compatible with Appveyor yet
 r_github_packages: lionel-/vdiffr
 
+# environment variables set for all builds
 env:
   global:
+  # don't treat missing suggested packages as error
   - _R_CHECK_FORCE_SUGGESTS_=false
```
hadley: It just occurred to me that we could set `USE_VDIFFR` to false here, and then I think the specific matrix env var would override this one.

clauswilke: Let me try that.

PR #2858, opened 2018-08-26
Enable manual setting of axis tick labels with coord_sf. Closes #2857

### LINKED ISSUE #2857: coord_sf() ignores axis label settings
As mentioned in https://github.com/tidyverse/ggplot2/issues/2846#issuecomment-415981057, axis labels cannot be set manually when using `coord_sf()`.

``` r
library(sf)
library(ggplot2) 

# graticule labeling via cardinal coordinates (N, S, E, W)
nc <- st_read(system.file("gpkg/nc.gpkg", package="sf"))
#> Reading layer `nc.gpkg' from data source `/Library/Frameworks/R.framework/Versions/3.5/Resources/library/sf/gpkg/nc.gpkg' using driver `GPKG'
#> Simple feature collection with 100 features and 14 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> epsg (SRID):    4267
#> proj4string:    +proj=longlat +datum=NAD27 +no_defs
ggplot() + 
  geom_sf(aes(fill = AREA), data=nc) + 
  coord_sf() +
  scale_y_continuous(
    breaks = c(34, 34.5, 35.7), # setting breaks manually works
    labels = c("A", "B", "C")   # but labels are ignored
  )
```

The reason is that the `setup_panel_params()` function of `coord_sf()` ignores the break info generated by the scales and instead reads and interprets the scales' `breaks` parameters directly: 
https://github.com/tidyverse/ggplot2/blob/6500a307780888d5b49aa32d97ea792c07a733c6/R/sf.R#L450-L457
For comparison, `coord_cartesian()` does this:
https://github.com/tidyverse/ggplot2/blob/6500a307780888d5b49aa32d97ea792c07a733c6/R/coord-cartesian-.r#L102

`coord_sf()` is written the way it is because it needs to let `sf::st_graticule()` make appropriate decisions about where to put graticules and how to label them, with the knowledge of the CRS that is being used. The scales don't have that knowledge and would in general produce bad breaks and labels.

What this means is that if we want to be able to allow manual overriding of labels, we will have to add additional logic to `coord_sf()` / `setup_panel_params()` that essentially mimics the scales' `get_labels()` function:
https://github.com/tidyverse/ggplot2/blob/6500a307780888d5b49aa32d97ea792c07a733c6/R/scale-.r#L289-L309

Also potentially relevant: https://github.com/r-spatial/sf/issues/829

clauswilke:
This PR closes #2857. It enables most ways in which axis tick labels can normally be set for `coord_sf()`. The main limitation is that it can only accept character vectors. Parsed expressions will not work. (However, strings that contain the degree symbol will automatically be parsed into expressions.)

``` r
library(sf)
library(ggplot2) 

nc <- st_read(system.file("gpkg/nc.gpkg", package="sf"))
#> Reading layer `nc.gpkg' from data source `/Library/Frameworks/R.framework/Versions/3.5/Resources/library/sf/gpkg/nc.gpkg' using driver `GPKG'
#> Simple feature collection with 100 features and 14 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> epsg (SRID):    4267
#> proj4string:    +proj=longlat +datum=NAD27 +no_defs

# explicitly supply label values
ggplot() + 
  geom_sf(aes(fill = AREA), data=nc) + 
  coord_sf() +
  scale_x_continuous(labels = letters[1:5]) +
  scale_y_continuous(
    breaks = c(34, 34.5, 35.7),
    labels = c("A", "B", "C")
  )
```

``` r

# set to NULL to remove labels
ggplot() + 
  geom_sf(aes(fill = AREA), data=nc) + 
  coord_sf() +
  scale_x_continuous(labels = NULL) +
  scale_y_continuous(
    breaks = c(34, 34.5, 35.7),
    labels = NULL
  )
```

``` r

# format via function
ggplot() + 
  geom_sf(aes(fill = AREA), data=nc) + 
  coord_sf() +
  scale_x_continuous(
    breaks = c(-84, -80, -76),
    labels = function(x) paste0(x, "*degree*East")
  ) +
  scale_y_continuous(
    breaks = c(34, 34.5, 35.7),
    labels = function(x) paste0(x, "*degree*North")
  )
```

``` r

# error if length mismatch
ggplot() + 
  geom_sf(aes(fill = AREA), data=nc) + 
  coord_sf() +
  scale_x_continuous(labels = letters[1:4])
#> Error: Breaks and labels along x direction are different lengths
```
Created on 2018-08-25 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@
 *   The error message in `compute_aesthetics()` now provides the names of only
     aesthetics with mismatched lengths, rather than all aesthetics (@karawoo,
     #2853).
+    
+*   `coord_sf()` now respects manual setting of axis tick labels (@clauswilke,
+    #2857).
 
 *   `geom_sf()` now respects `lineend`, `linejoin`, and `linemitre` parameters 
     for lines and polygons (@alistaire47, #2826)
--- a/R/sf.R
+++ b/R/sf.R
@@ -437,6 +437,52 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     data
   },
 
+
+  # internal function used by setup_panel_params,
+  # overrides the graticule labels based on scale settings if necessary
+  get_graticule_labels = function(self, graticule, scale_x, scale_y, params = list()) {
+    # if sf coordinates are not available in degrees latitude and longitude, label
+    # as regular numbers
+    if (is.null(params$crs) || is.na(params$crs) || !isTRUE(sf::st_is_longlat(self$datum))) {
+      x_labeller <- base::format
+      y_labeller <- base::format
+    }
+    else {
+      x_labeller <- degree_labels_EW
+      y_labeller <- degree_labels_NS
+    }
+
+    # if scales provide labeling functions override previous function choices
+    if (is.function(scale_x$labels)) {
+      x_labeller <- scale_x$labels
+    }
+    if (is.function(scale_y$labels)) {
+      y_labeller <- scale_y$labels
+    }
+
+    x_breaks <- graticule[graticule$type == "E", ]$degree
+    if (is.null(scale_x$labels)) {
+      x_labels <- rep(NA, length(x_breaks))
+    } else if (is.character(scale_x$labels)) {
+      x_labels <- scale_x$labels
+    } else {
+      x_labels <- x_labeller(x_breaks)
+    }
+
+    y_breaks <- graticule[graticule$type == "N", ]$degree
+    if (is.null(scale_y$labels)) {
+      y_labels <- rep(NA, length(y_breaks))
+    } else if (is.character(scale_y$labels)) {
+      y_labels <- scale_y$labels
+    } else {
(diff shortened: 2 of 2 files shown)


clauswilke:
I have some thoughts on how autoparsing could be avoided / used only on labels generated by `st_graticule()`, but it is quite involved and I don't want to get into this right now. The fundamental problem is that in the most general case we can end up with character labels for one direction (e.g. N/S) and parsed expressions for the other (e.g. E/W). To handle this appropriately, we'd have to break the graticule into two components, one for each direction, and store them separately. This would have repercussions in several other places in the code.

Issue #2864, opened 2018-08-27
Unexpected order of text labels when parse=TRUE


slowkow:
## The problem

Text labels can be displayed in an unexpected order if you use `parse = TRUE`:

```r
library(ggplot2)
library(patchwork)

d <- data.frame(
  x = c(1, 2, 3),
  y = c(1, 0, 1),
  label = c("alpha", "", "gamma")
)

p <- ggplot(d, aes(x = x, y = y, label = label)) + geom_point(color = "red")

p1 <- p + geom_text(size = 5, parse = FALSE) + labs(title = "Good :)")

p2 <- p + geom_text(size = 5, parse = TRUE) + labs(title = "Bad :(")

ggsave(filename = "issue.png", plot = p1 + p2, width = 4, height = 2)
```

![issue](https://user-images.githubusercontent.com/209714/44682017-39c1c900-aa10-11e8-9756-58c1e52603ea.png)

## The culprit: parse() drops empty strings

To my surprise, the `parse()` function silently drops the empty string:

```r
x <- c("alpha", "", "gamma")
x
#> [1] "alpha" ""      "gamma"
length(x)
#> [1] 3

y <- parse(text = x)
y
#> expression(alpha, gamma)
length(y)
#> [1] 2

c(expression(alpha), expression(), expression(gamma))
#> expression(alpha, gamma)
```

Unfortunately, I don't know how to deal with this. Do you have any ideas?

## A workaround

One workaround is to create a second dataframe just for the text labels. That way, you can avoid passing an empty string to the `parse()` function.

```r
d2 <- d[d$label != "",]

p3 <- p + geom_text(size = 5, parse = TRUE, data = d2) + labs(title = "Yay :)")

ggsave(filename = "workaround.png", plot = p1 + p3, width = 4, height = 2)
```

![workaround](https://user-images.githubusercontent.com/209714/44682625-e5b7e400-aa11-11e8-9254-04ac2e665b69.png)

## Related issues

@pakidermo5000 originally reported this here: https://github.com/slowkow/ggrepel/issues/114

aphalo:
Another workaround is to use NA instead of "". One way out could be to replace "" by NA before parsing. 

However, the behaviour is the same for geom_label()! So, this is bug is also present in ggplot2.
```r
library(patchwork)

d <- data.frame(
  x = c(1, 2, 3),
  y = c(1, 0, 1),
  label = c("alpha", "", "gamma")
)

[...]

PR #2867, opened 2018-08-28
replace parse() with parse_safe() in geom_text()

slowkow:
The built in `parse()` function silently drops items:

    length(parse(text = c("alpha", "", "gamma")))

We expect the length to be 3, but instead it is 2.

So, we add a new function `parse_safe()` that keeps all items.

This addresses issue #2864


clauswilke:
Several comments:

1. This still parses twice. It must be possible to parse only once, even if the result should be an expression object and not a list.

2. Please set `keep.source = FALSE` in `parse()`: `parse(text = text, keep.source = FALSE)` Otherwise regression tests won't work.

3. Please add a few regression tests, including testing with character vectors, factors, data with `NA`s, and data containing empty strings. They will generally look like this (in the testthat suite): `expect_equal(parse_safe(c("A", "B", "C")), expression(A, B, C))`

clauswilke:
Does this version of Hadley's original function do what you need? (Though I still don't understand why a list of expressions won't work. It does in my testing with `textGrob()` (see [here.](https://github.com/tidyverse/ggplot2/issues/2864#issuecomment-416357242))

``` r
parse_hadley <- function(text) {
  text <- as.character(text)
  out <- vector("list", length(text))
  for (i in seq_along(text)) {
    expr <- parse(text = text[[i]])
    out[[i]] <- if (length(expr) == 0) "" else expr[[1]]
  }
  do.call(expression, out)
}
```

hadley:
I think `vector("expression", length(text))` is slightly preferential than the final `do.call()` (partly because the semantics are so complex when the inputs are expressions)

clauswilke:
I've figured out why list of expressions don't work here but seem to work in other parts of ggplot2. We're actually converting them back to expression objects, e.g. here:
https://github.com/tidyverse/ggplot2/blob/71cb174e871d0ef96fa48d02495e6ff92b2ebb05/R/guides-axis.r#L35-L41

It's good to keep that in mind, as I think the ggplot2 code is a bit inconsistent in this regard. For example, my original code was inspired by this line:
https://github.com/tidyverse/ggplot2/blob/34c01ccd3b22fadd9c662a5ef4b386469a3dec59/R/sf.R#L481
Also, functions in the scales package return lists of expressions rather than expression objects:
``` r
> scales::math_format()(1:5)
[[1]]
10^`1`

[[2]]
10^`2`

[[3]]
10^`3`

[[4]]
10^`4`

[[5]]
10^`5`
```
In summary: I agree with your latest approach to `parse_safe()`, which returns an expression object. I'm wondering to what extent other parts of the code should use this function also, e.g. the code in sf I linked to above.

slowkow:
I agree that there might be other places, in addition to `geom-text.r` and `geom-label.r` that could benefit from the new `parse_safe()` function.

It would be nice to demonstrate that they suffer from the same problem before I modify them. Please share if you have any examples that demonstrate the issue!

clauswilke:
Kamil, here you go. I assume it's caused by this line:
https://github.com/tidyverse/ggplot2/blob/34c01ccd3b22fadd9c662a5ef4b386469a3dec59/R/sf.R#L481

Reprex:
``` r
library(ggplot2) 

# works
ggplot(nc) +
  geom_sf(aes(fill = AREA)) +
  scale_y_continuous(
    breaks = c(34, 35, 36),
    labels = c("34*degree*N", "35*degree*N", "36*degree*N")
  )
```



``` r

# breaks
ggplot(nc) +
  geom_sf(aes(fill = AREA)) +
  scale_y_continuous(
    breaks = c(34, 35, 36),
    labels = c("34*degree*N", "", "36*degree*N")
  )
#> Error in parse(text = x)[[1]]: subscript out of bounds
```

Created on 2018-08-30 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

slowkow:
I don't know how to fix the `label_parsed()` issue. I followed `facet_grid()` to `ggproto()` and now I am lost. It's hard to figure out where the chain of function calls ends up.

Could I please ask for help? Would you use `traceback()` or some other tools to figure this out?

slowkow:
As far as I can tell, there is only one more location where we might be interested to use `parse_safe()`:

https://github.com/tidyverse/ggplot2/blob/510b4b4b3eae992255ddca86de63d1fada2221ca/R/aes.r#L317-L320

I would leave this as is, because:
- This is in the deprecated `aes_auto()` function.
- I don't understand this code.

What do you think? Is there anything missing in this pull request?

hadley:
No need to touch deprecated functions.

But I do think it's worth using the same code in `label_parsed()`. Unfortunately I can't tell exactly what the code is trying to do there 😞

R/utilities.r
```diff
@@ -430,3 +430,24 @@ is_column_vec <- function(x) {
   dims <- dim(x)
   length(dims) == 2L && dims[[2]] == 1L
 }
+
+#' Parse a vector of expressions without silently dropping any items.
```
hadley: Since this is not exported, I don't think it needs to be documented. A comment mentioning that `parse(text = text)` has subtle problems with link to issue should be adequate.

tests/testthat/test-parse.r
```diff
@@ -0,0 +1,52 @@
+context("Parsing expressions")
+
+expect_equal(
```
hadley: These need to live a couple of `test_that()` blocks in `test-utilities.R` - see detail in http://style.tidyverse.org/tests.html#organisation (ggplot2 doesn't fully follow this organisation but should move towards it)

tests/testthat/test-parse.r
```diff
@@ -0,0 +1,52 @@
...
+  parse_safe(c("alpha", "beta", "gamma")),
+  expression(alpha, beta, gamma)
+)
+
+expect_equal(
+  parse_safe(c("alpha", "", "gamma", " ")),
+  expression(alpha, NA, gamma, NA)
+)
+
+expect_equal(
```
hadley: I think you should put all "multi" expressions into one test

tests/testthat/test-parse.r
```diff
@@ -0,0 +1,52 @@
...
+  parse_safe(c("alpha ~ beta", " ", "integral(f(x) * dx, a, b)")),
+  expression(alpha ~ beta, NA, integral(f(x) * dx, a, b))
+)
+
+expect_equal(
+  parse_safe(c("alpha \n beta", " ", "integral(f(x) * dx, a, b)")),
+  expression(alpha, NA, integral(f(x) * dx, a, b))
+)
+
+expect_equal(
```
hadley: This doesn't test anything new?

tests/testthat/test-parse.r
```diff
@@ -0,0 +1,52 @@
...
+  parse_safe(factor(c("alpha", "beta", ""))),
+  expression(2, 3, 1)
+)
+
+expect_equal(
+  parse_safe(factor(c(NA, "a", NA, "alpha"))),
+  expression(NA, 1, NA, 2)
+)
+
+expect_equal(
```
hadley: I don't think this tests anything new either.

R/geom-label.R
```diff
@@ -63,7 +63,7 @@ GeomLabel <- ggproto("GeomLabel", Geom,
                         label.size = 0.25) {
     lab <- data$label
     if (parse) {
-      lab <- parse(text = as.character(lab))
+      lab <- parse_safe(text = as.character(lab))
```
hadley: Since `parse_safe()` only has one argument, there's no longer a need to name it.

clauswilke:
I don't understand the code in `label_parsed()` either, but isn't `parse_safe()` meant as drop-in replacement for `parse()` and hence should just work? Have you tried leaving everything else untouched and just writing `parse_safe()` instead of `parse()`?

R/sf.R
```diff
@@ -553,11 +549,18 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     graticule <- panel_params$graticule
     east <- graticule[graticule$type == "E" & !is.na(graticule$degree_label), ]
 
+    labs <- east$degree_label
+
+    # parse labels into expressions if required
```
clauswilke: You're actually fixing another problem here at the same time, which is good. (The problem is that we may have to parse only labels along one axis.) However, this distributes the autoparsing to two separate locations. Maybe you could document this a little better by stating that labels created by `sf::st_graticule()` contain degree symbols and need to be parsed to look right?

clauswilke: Kamil, I've been pondering this some more. It's bugging me that we're breaking the parsing into two different places because I have a PR pending where we're labeling based on cardinal directions (N, S, E, W) rather than plot side and it can shuffle which labels appear where, and that will break with your current code. Is there any reason why we can't just do something like the following (using `parse_safe()` instead of `parse()`, though):
``` r
graticule <- data.frame(
  degree_label = c("abcd", "b", "10 * degree * E"),
  stringsAsFactors = FALSE
)
parse_ids <- grepl("degree", graticule$degree_label)
if (any(parse_ids)) {
  graticule$degree_label <- ifelse(
    parse_ids,
    parse(text = graticule$degree_label),
    as.expression(graticule$degree_label)
  )
}
graticule$degree_label
#> expression("abcd", "b", 10 * degree * E)
```
This will keep strings as strings but stick everything into an expression object so we don't have to worry about which label is of which type.

slowkow: I broke the parsing into two different places to avoid having an `expression()` column in the `graticule` dataframe.

Here is an example that shows my struggle:

```r
graticule <- data.frame(
  type = c("E", "E", "E"),
  degree_label = c(NA, "10 * degree * E", "15 * degree * E"),
  stringsAsFactors = FALSE
)

# This works if degree_label is not an 'expression'
east <- graticule[graticule$type == "E" & !is.na(graticule$degree_label), ]

parse_ids <- grepl("degree", graticule$degree_label)
if (any(parse_ids)) {
  graticule$degree_label <- ifelse(
    parse_ids,
    parse(text = graticule$degree_label),
    as.expression(graticule$degree_label)
  )
}
graticule
#> Warning in format.data.frame(x, digits = digits, na.encode = FALSE):
#> corrupt data frame: columns will be truncated or padded with NAs
#>   type                                                degree_label
#> 1    E expression(NA_character_, 10 * degree * E, 15 * degree * E)
#> 2    E                                                        <NA>
#> 3    E                                                        <NA>
graticule$degree_label
#> expression(NA_character_, 10 * degree * E, 15 * degree * E)
is.na(graticule$degree_label)
#> Warning in is.na(graticule$degree_label): is.na() applied to non-(list or
#> vector) of type 'expression'
#> [1] FALSE FALSE FALSE
```

R/utilities.r
slowkow: I have a feeling future developers will appreciate the hints in the documentation. I know I was definitely surprised by the default behavior of `parse()` dropping things. Totally unexpected for me.

Regarding the link, is `#2864` sufficient or do you prefer to put a full url like `https://github.com/tidyverse/ggplot2/issues/2864`?

tests/testthat/test-parse.r
slowkow: Got it! I'll move the tests into `test-utilities.R`

tests/testthat/test-parse.r
slowkow: You're right, some of the tests are probably redundant. I'll try to reduce redundancy.

R/sf.R
clauswilke: And to show that this creates a proper data frame:
``` r
graticule <- data.frame(
  type = c("E", "E", "E", "E"),
  degree_label = c(NA, "abcd", "10 * degree * E", "15 * degree * E"),
  stringsAsFactors = FALSE
)

parse_ids <- grepl("degree", graticule$degree_label)
if (any(parse_ids)) {
  graticule$degree_label <- ifelse(
    parse_ids,
    lapply(graticule$degree_label, function(x) parse(text = x)[[1]]),
    lapply(graticule$degree_label, function(x) as.expression(x)[[1]])
  )
}
graticule
#>   type    degree_label
#> 1    E              NA
#> 2    E            abcd
#> 3    E 10 * degree * E
#> 4    E 15 * degree * E
```

Created on 2018-09-01 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

slowkow: Thanks! I'm starting to understand the motivation for wrapping expressions in lists here and also in `label_parsed()`.

R/utilities.r
hadley: Here, I'd actually prefer the full url since that is clickable in RStudio. I'd suggest something like:

```R
# Parse takes n lines and returns m _expressions_
# See https://github.com/tidyverse/ggplot2/issues/2864 for discussion
```

R/sf.R
clauswilke: Putting the conditional inside the mapping than the other way round is probably more elegant.
``` r
graticule <- data.frame(
  type = c("E", "E", "E", "E"),
  degree_label = c(NA, "abcd", "10 * degree * E", "15 * degree * E"),
  stringsAsFactors = FALSE
)

parse_ids <- grepl("degree", graticule$degree_label)
if (any(parse_ids)) {
  graticule$degree_label <- Map(
    function(b, l) {
      if (b) {
        parse(text = l)[[1]]
      } else {
        as.expression(l)[[1]]
      }
    },
    parse_ids, graticule$degree_label
  )
}
graticule
#>   type    degree_label
#> 1    E              NA
#> 2    E            abcd
#> 3    E 10 * degree * E
#> 4    E 15 * degree * E
```

Created on 2018-09-02 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

slowkow:
Thanks for your patience on this pull request! I hope we're getting closer to the end.

What are your comments?

- [x] I adjusted the comments for `parse_safe()` as you suggested.
- [x] I moved the tests to `tests/testthat/test-utilities.r`
- [x] I removed the unnecessary instances of `text = ` in `geom_text()` and `geom_label()`
- [x] I changed `geom_sf()` back to parsing only once instead of parsing once for east and once for north.
- [x] I still vote to leave `label_parsed()` as-is, with more discussion below.

---

### Regarding `label_parsed()`

I have not been able to find an example where `label_parsed()` requires the new `parse_safe()` function. That's the main reason I don't want to change it.

To give you an idea of how `label_parsed()` would change with the new `parse_safe()`, here is an example:

```r
# with parse_safe()
label_parsed(c("", "alpha", "gamma"))
#
# [[1]]
# [[1]][[1]]
# expression(NA)
#
#
# [[2]]
# [[2]][[1]]
# expression(alpha)
#
#
# [[3]]
# [[3]][[1]]
# expression(gamma)
```

```r
# with parse()
label_parsed(c("", "alpha", "gamma"))
#
# [[1]]
# [[1]][[1]]
# expression()
#
#
# [[2]]
# [[2]][[1]]
# expression(alpha)
#
#
# [[3]]
# [[3]][[1]]
# expression(gamma)
```

If you think it's better to have `expression(NA)` instead of `expression()`, then we should replace `parse()` with `parse_safe()`. (I think it doesn't matter.) Otherwise, I think it's ok to let it be.

R/sf.R
```diff
@@ -511,6 +507,21 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
...
+          if (parse_id) {
+            parse(text = label)[[1]]
+          } else {
+            as.expression(label)[[1]]
+          }
+        },
+        parse_ids, graticule$degree_label
+      )
+    }
+
```
clauswilke: Would you mind moving this code block into the `fixup_graticule_labels()` function?

clauswilke:
This all looks good to me. I have one last, minor comment. Could you please move the label parsing code into the function `fixup_graticule_labels()`? The reason is that with the changes you're making to `CoordSf()`, I'll be able to properly fix the problem that we currently have where we're guessing whether labels should be parsed or not, rather than making this configurable. For the fix I envision, I'll need this code block in that function.

R/utilities.r
```diff
@@ -430,3 +430,21 @@ is_column_vec <- function(x) {
   dims <- dim(x)
   length(dims) == 2L && dims[[2]] == 1L
 }
+
+# Parse takes takes n strings and returns n expressions
```
clauswilke: I noticed "takes takes". Since I asked for one more substantive change, maybe you could fix this also.

slowkow:
- [x] Moved parsing code into `fixup_graticule_labels()`
- [x] Fixed typo.

Hadley, do you still prefer that `label_parsed()` use `parse_safe()`? Or shall we leave it as-is?

I think this might be the last issue to agree upon in this pull request.

R/utilities.r
```diff
@@ -430,3 +430,21 @@ is_column_vec <- function(x) {
   dims <- dim(x)
   length(dims) == 2L && dims[[2]] == 1L
 }
+
+# Parse takes n strings and returns n expressions.
```
hadley: This should be: "Parse takes a vector of n lines and returns m expressions."

hadley:
I would like to understand what `label_parsed()` does and ensure that it uses the same logic as here. But I think that would be better done in another PR.

NEWS.md
```diff
@@ -59,6 +59,11 @@
 *   `position_nudge()` is now more robust and nudges only in the direction
     requested. This enables, for example, the horizontal nudging of boxplots
     (@clauswilke, #2733).
+    
+*   `geom_text(..., parse = TRUE)` now correctly renders the expected number of
+    items instead of silently dropping items that are not valid expressions.
```
hadley: Not valid isn't quite the right turn of phrase - they are valid; they're just empty. We also fixed handling of multiple expressions in a single string.

R/utilities.r
```diff
@@ -430,3 +430,21 @@ is_column_vec <- function(x) {
...
+# See https://github.com/tidyverse/ggplot2/issues/2864 for discussion.
+#
+# parse(text = c("alpha", "", "gamma"))
+# #> expression(alpha, gamma)
+#
+# parse_safe(text = c("alpha", "", "gamma"))
+# #> expression(alpha, NA, gamma)
+#
+parse_safe <- function(text) {
```
hadley: I think it's worth adding a `stopifnot(is.character(text))` here.

R/sf.R
```diff
@@ -476,9 +476,20 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
...
-      graticule$degree_label <- lapply(graticule$degree_label, function(x) parse(text = x)[[1]])
+    # Convert the string 'degree' to the degree symbol
+    parse_ids <- grepl("\\bdegree\\b", graticule$degree_label)
+    if (any(parse_ids)) {
+      graticule$degree_label <- Map(
+        function(parse_id, label) {
+          if (parse_id) {
+            parse(text = label)[[1]]
+          } else {
+            as.expression(label)[[1]]
```
hadley: Why are we turning `label` into an expression object and then immediately extracting it out of the expression object?

R/sf.R
```diff
@@ -476,9 +476,20 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     if (!is.null(graticule$plot12))
       graticule$degree_label[!graticule$plot12] <- NA
 
-    # parse labels into expressions if required
-    if (any(grepl("degree", graticule$degree_label)))
-      graticule$degree_label <- lapply(graticule$degree_label, function(x) parse(text = x)[[1]])
+    # Convert the string 'degree' to the degree symbol
+    parse_ids <- grepl("\\bdegree\\b", graticule$degree_label)
+    if (any(parse_ids)) {
+      graticule$degree_label <- Map(
```
hadley: I'd rather not use `Map()` here - I suspect you could more simply right as something like:

```R
labels <- as.list(graticule$degree_label)
labels[parse_ids] <- lapply(labels[parse_ids], parse_safe)
```

R/sf.R
```diff
@@ -476,9 +476,20 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     if (!is.null(graticule$plot12))
       graticule$degree_label[!graticule$plot12] <- NA
 
-    # parse labels into expressions if required
-    if (any(grepl("degree", graticule$degree_label)))
-      graticule$degree_label <- lapply(graticule$degree_label, function(x) parse(text = x)[[1]])
+    # Convert the string 'degree' to the degree symbol
+    parse_ids <- grepl("\\bdegree\\b", graticule$degree_label)
```
hadley: I think this would be more informative if called `needs_parsing` or similar

slowkow: I agree!

R/utilities.r
slowkow: Yes.

R/sf.R
clauswilke: Because we need the symbol itself, not `expression(symbol)`. This is directly copied from the current code.

hadley: One problem with this code chunk as a whole, is that when I read it I can't easily tell what type of object `graticule$degree_label` is supposed to be. It's obviously a list, and some of the components are expressions (not [expression objects](https://adv-r.hadley.nz/expressions.html#naming-conventions)), but what are the others supposed to be? 

You could fix that with a comment, but it would be even better to use functions where it's obvious what the output type is.

hadley: If we need a symbol, and we have a character vector of length one, then the function to use is `as.symbol()`

R/sf.R
clauswilke: Yes, this is much simpler. For some reason I always forget that we can assign to a subset of elements in a list. However, it still needs the part where only the first element of the expression object is extracted, so we get the element itself not inside an `expression()` statement:

``` r
labels <- as.list(graticule$degree_label)
labels[parse_ids] <- lapply(labels[parse_ids], function(x) parse_safe(x)[[1]])
```

R/sf.R
clauswilke: Yes, correct. See also my other comment, below.

R/sf.R
hadley: Oh but parse_safe is already vectorised so you can skip the lapply

hadley: If you apply the character vector that hasn’t been coerced to a list

clauswilke: Not sure. It may be best to work with a full reprex. The following works:
``` r
parse_safe <- function(text) {
  out <- vector("expression", length(text))
  for (i in seq_along(text)) {
    expr <- parse(text = text[[i]])
    out[[i]] <- if (length(expr) == 0) NA else expr[[1]]
  }
  out
}

graticule <- data.frame(
  type = c("E", "E", "E", "E"),
  degree_label = c(NA, "abcd", "10 * degree * E", "15 * degree * E"),
  stringsAsFactors = FALSE
)

needs_parsing <- grepl("degree", graticule$degree_label)
labels <- as.list(graticule$degree_label)
labels[needs_parsing] <- lapply(labels[needs_parsing], function(x) parse(text = x)[[1]])
graticule$degree_label <- labels
graticule
#>   type    degree_label
#> 1    E              NA
#> 2    E            abcd
#> 3    E 10 * degree * E
#> 4    E 15 * degree * E
```
Do you see a way to make it simpler?

slowkow:
Hadley, I think the labeller functions have some issues that go beyond the scope of this pull request... it's difficult to write up my findings because I am running into unexpected behavior as I try to make examples. I agree with you that we ought to discuss this issue separately.

clauswilke:
Thanks Kamil!

clauswilke:
Squash-merge is the default on github. I just need to click a button. :-) GitHub then opens a window where you can edit the commit message, and that's where I inserted "Closes issue #2864".

PR #2874, opened 2018-08-29
Make nudging more robust

### LINKED ISSUE #2733: position_nudge does not work with geom_boxplot
position_nudge seems to not not work with geom_boxplot

```r
library(ggplot2)
df <- data.frame(x = factor("x"), y = 1:10)
ggplot(df, aes(x = x, y = y)) +
  geom_boxplot(position = position_nudge(x = -0.2, y = 0))
#> Error: position_nudge requires the following missing aesthetics: y
```

What I want to do eventually, is to place a dotplot next to a  boxplot like on this image
tmp
(I manage to produce this image with ggplot2. Oddly, this issue does not occur with some data. That is why I am completely lost...)

I use R 3.5.1 with ggplot2 3.0.0 on debian stretch. The reproducible example was made with reprex.

clauswilke:
This closes #2733.

``` r
library(ggplot2)

df <- data.frame(
  x = c(1,3,2,5),
  y = c("a","c","d","c")
)

ggplot(df, aes(x, y)) +
  geom_point() +
  geom_text(aes(label = y), position = position_nudge(x = 0.2, y = -0.1))
```

``` r

df <- data.frame(x = factor("x"), y = 1:10)
ggplot(df, aes(x = x, y = y)) +
  geom_boxplot(position = position_nudge(x = -0.2, y = 0)) +
  geom_vline(xintercept = 1)
```

``` r
  
df <- data.frame(
  x = factor(c("preop", "3m", "1w", "1d", "1w", "3w", "1d",
               "1d", "1w", "3m", "1d", "1w", "6m", "preop",
               "6m", "6w", "6w", "preop", "3w", "6m"),
             levels = c("preop", "1d", "1w", "3w", "6w",
                        "3m", "6m")),
  y = c(14, 10, 13, 12, 10, 11, 7, 14, 15, 18, 15, 9, 17,
        26, 8, 8, 14, 23, 21, 13)
)
ggplot(df, aes(x = x, y = y)) +
  geom_boxplot(width = .3, position = position_nudge(x = -0.2, y = 0),
               outlier.shape=NA) +
  geom_dotplot(binaxis = "y", binwidth = 0.5, dotsize = 0.7)
```

Created on 2018-08-29 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


R/position-nudge.R
```diff
@@ -27,10 +27,23 @@
...
-    x = x,
-    y = y
-  )
+  # find the appropriate position adjustment for the given x and y settings;
+  # not nudging in dimensions that are not required reduces the risk
+  # spurious breaks, e.g. when the geom only has an x coordinate but no
+  # y coordinate
+  if (x != 0) {
+    if (y != 0) {
+      pos <- ggproto(NULL, PositionNudge, x = x, y = y)
```
hadley: Why do these need to be separate objects? Couldn't the choice (i.e. this if-else block) be placed in `PositionNudge$compute_layer()`?

clauswilke: The primary difference among these objects is the required aes. If we want a single nudge object then it can't have any required aes. I don't have a strong opinion about it one way or another.

hadley: Seems reasonable to me to just drop the required aes.

clauswilke: Done.

PR #2875, opened 2018-08-30
Allow creation of custom layers that have access to global plot data

### LINKED ISSUE #2872: Reduce "object 'geometry' not found" error with geom_sf()
Currently, `geom_sf()` works without specifying `geometry` aes only when:

1. the `sf` object has a geometry column named `geometry`
2. the `sf` object is passed directly to `geom_sf()`, not via the plot data

I'm afraid it's a bit confusing that the example of `nc` won't work if we just replace `nc` with another `sf` object.

``` r
library(ggplot2)

attr(nc, "sf_column")
#> [1] "geometry"

# this works
ggplot(nc) +
  geom_sf(aes(fill = AREA))
```

``` r

nc_gpkg <- sf::st_read(system.file("gpkg/nc.gpkg", package = "sf"), quiet = TRUE)
attr(nc_gpkg, "sf_column")
#> [1] "geom"

# this fails
ggplot(nc_gpkg) +
  geom_sf(aes(fill = AREA))
#> Error in FUN(X[[i]], ...): object 'geometry' not found
```

Created on 2018-08-29 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

I want to let `ggplot(nc_gpkg) + geom_sf(aes(fill = AREA))` simply work. For example, implementing `fortify.sf()` seems to solve our problem quickly.

```r
fortify.sf <- function(model, data, ...) {
  if (! "geometry" %in% colnames(model)) {
    model$geometry <- sf::st_geometry(model)
  }
  model
}
```

But, I'm not fully sure this does the right thing... I feel this is similar kind of problem that we wanted to map grouping variable of `grouped_df` automatically to `group` aes, but couldn't (https://github.com/tidyverse/ggplot2/issues/2378). Are there any smarter way to create mappings automatically?

clauswilke:
This PR creates a general framework for custom layers that have access to the global plot data. It provides a solution to the problem that we're combining layers with `+` instead of `%>%`, and it closes #2872.


R/sf.R
```diff
@@ -135,11 +135,38 @@ stat_sf <- function(mapping = NULL, data = NULL, geom = "rect",
...
+#' @rdname ggsf
+#' @usage NULL
+#' @format NULL
+LayerSf <- ggproto("LayerSf", Layer,
+  setup_layer = function(self, plot) {
+    # Automatically determine the name of the geometry column
+    # and add the mapping if it doesn't exist
+    data <- self$layer_data(plot$data)
+
+    if (!is.null(data) && is_sf(data)) {
```
yutannihilation: When the data is not specified either in the plot or the layer, `Layer$layer_data()` probably returns `waiver` instead of `NULL`. `ggplot()$data` is `waiver` thanks to `fortify.NULL()`. So, maybe `is.waive()` here?

But, I'm not sure if it makes sense to set `geometry` aes when we already know `data` doesn't have `geometry` column...

yutannihilation:
In general, looks good to me!

R/sf.R
clauswilke: I'll change it to `is.waive()`. I'm not sure about the addition of the `geometry` aes when there is no geometry data, I copied that part from the existing code.

yutannihilation: > I'm not sure about the addition of the geometry aes when there is no geometry data

I'm not sure, too... You know, with the current implementation, `geom_sf()` cannot know the geometry column when the `sf` object is specified in the plot, not in the layer. So, placing the last glimmer of hope on `"geometry"` may be a reasonable strategy.

But, with `setup_layer()`, maybe we can just `stop()` when the data is not `sf`?

clauswilke: I don't even think we need to stop. We can just not do anything, and the general ggplot2 machinery will complain about a missing geometry aesthetic.

yutannihilation: Yes, plotting will fail in the either case. I'm OK with the current implementation, but, FWIW, my thoughts was:

`LayerSf` ensures the mapping has `geometry` aesthetic as long as the `data` is `sf`. In other words, if there's no `geometry`, `data` is probably not `sf`. In this case, specifying `geometry` won't fix the problem. So, maybe is the error message more helpful if it complains about the fact that `data` is not `sf`?

hadley:
Doesn't this assume that the plot data is already set at the time the layer is added? That's been the stumbling block for my previous attempts to fix the problem (and is a design decision that I now regret)

clauswilke:
@hadley I'm not sure I fully understand the question. As far as I understand, the data for a layer can come either from the original `ggplot()` call or from the `data` argument in the layer. My function `setup_layer()` has access to both and uses the `layer_data()` function to figure out which is the right one. The `layer_data()` is used later again to figure out the data for each layer, so the behavior should be identical. (And there is room for optimization so it's called only once, but `layer_data()` is pretty lightweight so I don't see a big need for this optimization at this time.)

hadley:
What about the case where the data is added later?

```R
ggplot() + geom_sf() + aes(x, y) %+% data
```

clauswilke:
I wasn't aware of the `%+% data` option. Let me ponder it. I think you're right that that scenario can't be handled with my current proposal, but it may be possible to move the `setup_layer()` call to later in the toolchain where this would work also.

clauswilke:
The `setup_layer()` function was called in the wrong location. It needs to be called from within `ggplot_build()`. Then it's possible to add data at the very end and still have all layers behave correctly.

``` r
library(tidyverse)

## SF example
# Geometry column is called "geom". It is found correctly.
nc_gpkg <- sf::st_read(system.file("gpkg/nc.gpkg", package = "sf"), quiet = TRUE)
ggplot(nc_gpkg) + geom_sf() + aes(fill = AREA)
```



``` r
# even if we add the data at the very end
(ggplot() + geom_sf() + aes(fill = AREA)) %+% nc_gpkg 
```



``` r

## Grouping example
# grouping ignored
group_by(iris, Species) %>%
  ggplot(aes(Sepal.Length, Sepal.Width)) + geom_smooth(method = "lm")
```



``` r

# autogrouping function
setup_autogrp = function(self, data, plot) {
  if ((isTRUE(self$inherit.aes) && is.null(self$mapping$group) && is.null(plot$mapping$group)) ||
      (!isTRUE(self$inherit.aes) && is.null(self$mapping$group))) {
    if (!is.null(data$.group)) {
      self$mapping$group <- as.name(".group")
    }
  }
  data
}

# now grouping is used but only in that layer
smooth_grp_layer <- geom_smooth(method = "lm")
smooth_grp_layer$setup_layer <- setup_autogrp
group_by(iris, Species) %>%
  ggplot(aes(Sepal.Length, Sepal.Width)) + 
    smooth_grp_layer +
    geom_text(
      data = tibble(Sepal.Length = 5.5, Sepal.Width = 3.25, label = "A label."),
      aes(label = label)
    )
```



Created on 2018-09-01 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

clauswilke:
One more example, doing the kind of filtering operation needed for gghighlight.

``` r
library(ggplot2)
# filtering function
setup_filter = function(self, data, plot) {
  # example filter
  filter <- quo(mpg > 25)

  # filter data
  index <- rlang::eval_tidy(filter, data)
  data[index, ]
}

# points that meet the filter criterion are shown in red
point_filter_layer <- geom_point(color = "red")
point_filter_layer$setup_layer <- setup_filter
  geom_point() +
  point_filter_layer
```



Created on 2018-09-01 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

R/layer.r
```diff
@@ -199,9 +199,10 @@ Layer <- ggproto("Layer", NULL,
     }
   },
 
-  # hook to allow a layer access to global plot data
-  # as the plot is constructed.
-  setup_layer = function(self, plot) {
+  # hook to allow a layer access to the final layer data
+  # in input form and to global plot info
+  setup_layer = function(self, data, plot) {
+    data
```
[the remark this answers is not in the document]
clauswilke: So you can filter data. See other example I just posted.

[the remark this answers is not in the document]
yutannihilation: Oh, sorry, I got it.

yutannihilation:
Thanks for the gghighlight example! Seems very useful. 👍

clauswilke:
@hadley Now that 3.1.0 is out of the way, I was wondering whether we can go ahead with the PR. I think the latest version gets the implementation right and addresses your concern from Aug. 31 about data being added later. Overall, this is a very small code modification that will open up a wide range of new extension capabilities (see examples of autogrouping layers and highlighting layers provided above).

Might be good to get @thomasp85's input as well. Are there any implications for gganimate?

thomasp85:
In general the changes looks benign, and I can't see how it would affect gganimate (other than I have to update `ggplot_build.gganim()` to reflect these additions...

I'm pretty sure someone will use it for something that will break something at some point, but that's for future us to worry about 😄 

LGTM

clauswilke:
@thomasp85 We should coordinate carefully to avoid breaking gganimate with the currently released ggplot2 and/or with github master. I think this means patching gganimate first. I can prepare a PR.

I see two ways to do this. Any preference?

``` r
# version 1: explicit version check
if (packageVersion("ggplot2") <= "3.1.0") {
  # old code
} else {
  # new code
}

# version 2: check for capabilities
data <- layer_data
data <- by_layer(
  function(l, d) {
    # setup_layer() function doesn't exist in ggplot2 3.1.0 or earlier
    if (is.null(l$setup_layer)) d else l$setup_layer(d, plot)
  }
)
```
Does the same issue arise with patchwork?

thomasp85:
Down the line it would be nice if we could figure out a way that didn't require gganimate to copy large swaths of ggplot2 code in order to insert itself the right places, but that is for another day to worry about

clauswilke:
@thomasp85 I realized that gganimate will mostly continue to work without the fix, so I'm merging the change here now. I also prepared a PR for gganimate.

PR #2878, opened 2018-09-04
Improve document about handling of NA in geom_path()

### LINKED ISSUE #2877: geom_line not ignoring NA values
The documentation for `geom_line()` indicates that NA values are always ignored (and that the only difference with `na.rm` or not is if ignoring NA comes with a warning.  They don't currently appear to be ignored as the break the lines in the first two figures below.

``` r
library(ggplot2)
d_fake <- data.frame(
  x=1:5,
  y=c(0, NA, 1, NA, 0))

ggplot(d_fake, aes(x=x, y=y)) +
  geom_point() + geom_line() +
  scale_y_log10()
#> Warning: Transformation introduced infinite values in continuous y-axis

#> Warning: Transformation introduced infinite values in continuous y-axis
#> Warning: Removed 2 rows containing missing values (geom_point).
```

``` r

ggplot(d_fake, aes(x=x, y=y)) +
  geom_point(na.rm=TRUE) + geom_line(na.rm=TRUE) +
  scale_y_log10()
#> Warning: Transformation introduced infinite values in continuous y-axis
#> Warning: Transformation introduced infinite values in continuous y-axis
```

``` r

ggplot(d_fake[!is.na(d_fake$y),],
       aes(x=x, y=y)) +
  geom_point(na.rm=TRUE) + geom_line(na.rm=TRUE) +
  scale_y_log10()
#> Warning: Transformation introduced infinite values in continuous y-axis

#> Warning: Transformation introduced infinite values in continuous y-axis
```

Created on 2018-09-03 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

yutannihilation:
Minor document fix. This closes #2877.

`geom_path()` (and its family functions) handles `NA` in this manner:

* If `NA` is at the middle of the line, it breaks the line and doesn't show any warnings (so, `na.rm` doesn't matter here)
* If `NA` is at the start or the end of the line and `na.rm` is `FALSE` (default), it removes the `NA` with a warning.
* If `NA` is at the start or the end of the line and `na.rm` is `TRUE`, it silently removes the `NA`.

This PR is intended to improve the document of `geom_path()` to illustrate this behaviour well.

### DIFF
--- a/R/geom-path.r
+++ b/R/geom-path.r
@@ -20,7 +20,7 @@
 #'  [geom_polygon()]: Filled paths (polygons);
 #'  [geom_segment()]: Line segments
 #' @section Missing value handling:
-#' `geom_path()`, `geom_line()`, and `geom_step` handle `NA` in this manner:
+#' `geom_path()`, `geom_line()`, and `geom_step` handle `NA` as follows:
 #'
 #' * If an `NA` occurs in the middle of a line, it breaks the line. No warning
 #'   is shown, regardless of whether `na.rm` is `TRUE` or `FALSE`.
--- a/R/geom-path.r
+++ b/R/geom-path.r
@@ -19,6 +19,15 @@
 #' @seealso
 #'  [geom_polygon()]: Filled paths (polygons);
 #'  [geom_segment()]: Line segments
+#' @section Missing value handling:
+#' `geom_path()`, `geom_line()` and `geom_step` handle `NA` in this manner:
+#'
+#' * If `NA` is at the middle of the line, it breaks the line. It doesn't
+#'   show any warnings no matter whether `na.rm` is `TRUE` or `FALSE`.
+#' * If `NA` is at the start or the end of the line and `na.rm` is `FALSE`
+#'   (default), it removes the `NA` with a warning.
+#' * If `NA` is at the start or the end of the line and `na.rm` is `TRUE`,
+#'   it silently removes the `NA`.
 #' @export
 #' @examples
 #' # geom_line() is suitable for time series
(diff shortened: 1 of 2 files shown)


clauswilke:
Thanks, this looks good to me. Can we say it closes #2877?

yutannihilation:
Updated the comment.

But..., I thought this doesn't directly closes #2877, since it seems a bit different misunderstanding  that `na.rm` could make `geom_line()` ignore `NA`s instead of breaking the lines by `NA`s.

hadley:
I'd rather have a new `@section Missing value handling:` rather than cluttering up the examples with these very special cases.

yutannihilation:
Sure.

yutannihilation:
Added the section and removed the examples except the one that illustrates how to break lines by `NA`.

R/geom-path.r
```diff
@@ -19,6 +19,15 @@
 #' @seealso
 #'  [geom_polygon()]: Filled paths (polygons);
 #'  [geom_segment()]: Line segments
+#' @section Missing value handling:
+#' `geom_path()`, `geom_line()` and `geom_step` handle `NA` in this manner:
```
clauswilke: Would you mind adding a comma after `geom_line()`?

R/geom-path.r
```diff
@@ -19,6 +19,15 @@
...
+#'
+#' * If `NA` is at the middle of the line, it breaks the line. It doesn't
+#'   show any warnings no matter whether `na.rm` is `TRUE` or `FALSE`.
+#' * If `NA` is at the start or the end of the line and `na.rm` is `FALSE`
+#'   (default), it removes the `NA` with a warning.
+#' * If `NA` is at the start or the end of the line and `na.rm` is `TRUE`,
+#'   it silently removes the `NA`.
```
clauswilke: Maybe write something like:
``` r
#' * If an `NA` occurs in the middle of a line, it breaks the line. No warning 
#'   is shown, regardless of whether `na.rm` is `TRUE` or `FALSE`.
#' * If an `NA` occurs at the start or the end of the line and `na.rm` is `FALSE`
#'   (default), the `NA` is removed with a warning.
#' * If an `NA` occurs at the start or the end of the line and `na.rm` is `TRUE`,
#'   the `NA` is removed silently, without warning.
```

R/geom-path.r
yutannihilation: Sure.

R/geom-path.r
yutannihilation: Thanks!

R/geom-path.r
```diff
@@ -19,6 +19,15 @@
 #' @seealso
 #'  [geom_polygon()]: Filled paths (polygons);
 #'  [geom_segment()]: Line segments
+#' @section Missing value handling:
+#' `geom_path()`, `geom_line()`, and `geom_step` handle `NA` in this manner:
```
hadley: in this manner -> "as follows"

yutannihilation: Thanks!

PR #2886, opened 2018-09-08
Update visual tests with dev vdiffr

dpseidel:
This PR regenerates the visual test case files using the development version of vdiffr which now uses `ggplot2::theme_test()` rather than an internal function to allow for easy editing. The shift changes plot dimensions slightly for all plots for which `theme_test()` is applied because of minor differences between the theme specifications -- see lionel-/vdiffr#40 for more details.


dpseidel:
Thanks @lionel-. So, just to confirm, this change will require ggplot2 developers to update their version of vdiffr in order to pass tests.  Already as of #2856, visual testing is dependent on the dev version of vdiffr but if a developed downloaded 0.2.3.9000 more than 5 days ago (prior to lionel-/vdiffr#40), their tests will run and break until they update.  OK to merge @hadley ?

lionel-:
Good point. I have just bumped the development version, could you make ggplot2 depend on vdiffr >= 0.2.3.9001 please?

dpseidel:
Done. I bumped it both in the check within the `helper-vdiffr.R` file and added the version predicate to the description file.

clauswilke:
With the changes to vdiffr, are we now in a situation where all Travis runs will break until this is merged? If yes, then we should merge ASAP. And also, is there any reason why the AppVeyor build failed? That seems strange.

dpseidel:
Strange, it looks like it failed to load scales and the proper version of vdiffr:  https://ci.appveyor.com/project/tidyverse/ggplot2/build/1.0.1197#L392 maybe related to my change of the description file or possibly just a random appveyer failure? I can take a better look as soon as I am back to my desktop. 

And yes, if Travis is pulling the newest version of vdiffr dev then yeah builds are going to fail. Let's try and get this merged today.

clauswilke:
I've sometimes seen random Appveyor breaks. Unfortunately we can't restart the build except by making another commit.

dpseidel:
Hmm yeah. Appveyor does not like the predicate version number added to the DESCRIPTION file. It's not strictly required so I will remove it.

dpseidel:
I'm ready to merge this, any concerns @hadley, @lionel- ?

clauswilke:
@dpseidel I think you should go ahead and merge. Other PRs are now blocked because of failing vdiffr runs (see e.g. [here](https://travis-ci.org/tidyverse/ggplot2/builds/428004376?utm_source=github_status&utm_medium=notification)). If there's anything else that needs to be addressed in regards to vdiffr we can do that later, when we notice it.

PR #2921, opened 2018-09-30
fix up NEWS.md for release

clauswilke:
I have taken a first stab at NEWS.md.

@hadley According to the [style guide](http://style.tidyverse.org/news.html#before-release), package authors should not be listed for items they wrote. Do you want me to remove @clauswilke and @karawoo from all 3.1.0 news items? (I think no other contributors are official authors.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,11 +6,11 @@ This is a minor release and breaking changes have been kept to a minimum. End us
 
 *   In non-user-facing internal code (specifically in the `aes()` function and in
     the `aesthetics` argument of scale functions), ggplot2 now always uses the British
-    spelling of aesthetics containing the word "colour". When users specify a "color"
-    aesthetic it is automatically renamed to "colour". This principle is also applied
+    spelling for aesthetics containing the word "colour". When users specify a "color"
+    aesthetic it is automatically renamed to "colour". This renaming is also applied
     to non-standard aesthetics that contain the word "color". For example, "point_color"
     is renamed to "point_colour". This convention makes it easier to support both
-    British and American spelling for novel non-standard aesthetics, but it may require
+    British and American spelling for novel, non-standard aesthetics, but it may require
     some adjustment for packages that have previously introduced non-standard color
     aesthetics using American spelling. A new function `standardise_aes_names()` is
     provided in case extension writers need to perform this renaming in their own code
@@ -45,7 +45,7 @@ This is a minor release and breaking changes have been kept to a minimum. End us
 
 *   `benchplot()` now uses tidy evaluation (@dpseidel, #2699).
 
-*   The error message in `compute_aesthetics()` now provides the names of only
+*   The error message in `compute_aesthetics()` now only provides the names of
     aesthetics with mismatched lengths, rather than all aesthetics (@karawoo,
     #2853).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,76 +1,84 @@
-# ggplot2 3.0.0.9000
+# ggplot2 3.1.0
 
-*   `geom_boxplot()` now understands the `width` parameter even when used with
-    a non-standard stat, such as `stat_identity()` (@clauswilke, #2893).
+## Breaking changes
 
-*   `scale_*_date()`, `scale_*_time()` and `scale_*_datetime()` can now display 
-    a secondary axis that is a __one-to-one__ transformation of the primary axis,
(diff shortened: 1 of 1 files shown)


hadley:
I don't mind either way about authorship — when I wrote that advice, I was mostly thinking of my team at RStudio.

clauswilke:
@hadley Ok, then I'll leave them in.

@batpigandme Any copy edits from your side?

NEWS.md
```diff
@@ -1,89 +1,109 @@
...
-    
-*   `coord_sf()` now respects manual setting of axis tick labels (@clauswilke,
-    #2857, #2881).
-    
-*   `coord_sf()` now accepts two new parameters, `label_graticule` and `label_axes`,
-    that can be used to specify which graticules to label on which side of the plot
-    (@clauswilke, #2846).
+This is a minor release and breaking changes have been kept to a minimum. End users of ggplot2 are unlikely to encounter any issues. However, there are a few items that developers of ggplot2 extensions should be aware of. For additional details, see also the discussion accompanying issue #2890.
+
+*   In non-user-facing internal code (specifically in the `aes()` function and in
```
batpigandme: Switch "of" to "for" for "British spelling for aesthetics…"? (I had a split second of wondering how the British spell the word of aesthetics)

NEWS.md
```diff
@@ -1,89 +1,109 @@
...
+    the `aesthetics` argument of scale functions), ggplot2 now always uses the British
+    spelling of aesthetics containing the word "colour". When users specify a "color"
+    aesthetic it is automatically renamed to "colour". This principle is also applied
```
batpigandme: Simplify the "This principle is also…" to "This also applies to non-standard aesthetics…" ?
(stylistic, it's fine either way)

NEWS.md
```diff
@@ -1,89 +1,109 @@
...
+    to non-standard aesthetics that contain the word "color". For example, "point_color"
+    is renamed to "point_colour". This convention makes it easier to support both
+    British and American spelling for novel non-standard aesthetics, but it may require
```
batpigandme: Comma after novel? (though I guess non-standard aesthetics could be considered the noun phrase — so, again, tossup!)

NEWS.md
```diff
@@ -1,89 +1,109 @@
...
+    some adjustment for packages that have previously introduced non-standard color
+    aesthetics using American spelling. A new function `standardise_aes_names()` is
```
batpigandme: Style-wise, I think the goal has been to have the new function near the front of the bullet (see http://style.tidyverse.org/news.html#common-patterns), but I get why it's not in this case.

NEWS.md
```diff
@@ -1,89 +1,109 @@
...
+    on sf objects. Under the hood, a new `stat_sf_coordinates()` calculates the
+    x and y coordinates from the coordinates of the sf geometries. You can customize
+    the calculation method via `fun.geometry` argument (@yutannihilation, #2761).
+    
+
+## Minor improvements and fixes
 
 *   `benchplot()` now uses tidy evaluation (@dpseidel, #2699).
 
+*   The error message in `compute_aesthetics()` now provides the names of only
```
batpigandme: Awk word ordering here, maybe change to: "now only provides the names of aesthetics…" or "now returns the names only for aesthetics…"

NEWS.md
clauswilke: Ok.

NEWS.md
clauswilke: I normally try to avoid using "this" as a pronoun (as opposed to a modifier), as it leads to imprecise writing. But I agree that "this principle" is awkward, so maybe imprecise is better here.

NEWS.md
clauswilke: Ok.

NEWS.md
clauswilke: Yeah, I think I'll leave it as is.

NEWS.md
clauswilke: Ok, thanks!

NEWS.md
clauswilke: I think "this renaming" works better than just "this":
> When users specify a "color" aesthetic it is automatically renamed to "colour". This renaming is also applied to non-standard aesthetics that contain the word "color".
versus:
> When users specify a "color" aesthetic it is automatically renamed to "colour". This is also applied to non-standard aesthetics that contain the word "color".

batpigandme: Sure. 👍

Issue #2472, opened 2018-03-05
stat_contour's breaks parameter is undocumented


alistaire47:
When trying to use `geom_contour` to draw lines at specific levels à la `contour(..., levels = ...)`, I discovered [this SO answer](https://stackoverflow.com/a/19658236/4497050) which documents that `geom_contour` can be passed a `breaks` parameter to specify particular _z_ values where contour lines should be drawn.

Weirdly, this behavior is not documented in `?geom_contour`, `?geom_density_2d`, `?layer`, or `?ggplot2:::StatContour`, and taking a cursory look at the source, I'm not sure where the `breaks` parameter is picked up. Given the behavior is important enough to merit a parameter in `contour`, seems like it should be in the docs.

I can make a PR if you like, but the problem is that I'm not exactly sure _how_ this behavior should be documented, as I'm not sure what `breaks` is a parameter of. An example wouldn't be too hard to add, though.

Reprex:

``` r
library(ggplot2)

v <- ggplot(faithfuld, aes(waiting, eruptions, z = density)) 

v + geom_contour()
```

![](https://i.imgur.com/6O8WuRl.png)

``` r
v + geom_contour(breaks = c(0.002, 0.02))
```

![](https://i.imgur.com/yedk4HS.png)

clauswilke:
The `bins`, `binwidth`, `breaks`, and `complete` parameters are parameters of `stat_contour()`: 
https://github.com/tidyverse/ggplot2/blob/d3bd92e8002672a016038e8ca3f5588f80a796d8/R/stat-contour.r#L37-L38
(`complete` has no effect, though, as far as I can tell.)

They should be documented just like the comparable parameters in `stat_bin()`, and they should probably also be added to the `stat_contour()` constructor (again, just like in `stat_bin()`):
https://github.com/tidyverse/ggplot2/blob/d3bd92e8002672a016038e8ca3f5588f80a796d8/R/stat-bin.r#L42-L49

JHonaker:
I'll take a stab at fixing this.

paleolimbot:
This TDD issue is a relatively easy fix! The open PR on this (#2926) has been abandoned and needs a few fixes, notably that 

```r
      bins = NULL,
      binwidth = NULL,
      breaks = NULL,
```

(in geom-contour.r) needs to be

```r
      bins = bins,
      binwidth = binwidth,
      breaks = breaks,
```

[...]

PR #2926, opened 2018-10-03
Added documentation for the stat/geom_contour.

JHonaker:
Referencing #2472, I added the parameters bins, binwidth, and breaks to the stat_contour and geom_contour parameter lists. For some reason they were undocumented (except in the examples), but still available for use.

There was also an unused parameter "complete." I couldn't find any reference to it in either function, so I removed it as well. If there's a reason that it shouldn't be removed, like a future feature, I can revert it.


clauswilke:
I'm closing this PR, since the issue has been addressed since.

PR #2936, opened 2018-10-12
Use .ignore_empty="all" for enquos()

yutannihilation:
I saw `aes()` is mentioned here: https://github.com/r-lib/rlang/pull/643. It says `enquos()` will get `"all"` option for `.ignore_empty`, which ignores named missing arguments as well, in rlang v0.3.0 . This PR uses it to simplify the code.

Besides, https://github.com/r-lib/rlang/issues/430 seems fixed, so the workaround is no longer needed.


tests/testthat/test-aes.r
```diff
@@ -89,7 +89,7 @@ test_that("assignment methods pull unwrap constants from quosures", {
 })
 
 test_that("quosures are squashed when creating default label for a mapping", {
-  p <- ggplot(mtcars) + aes(!!quo(identity(!!quo(cyl))))
+  p <- ggplot(mtcars) + aes(!!rlang::quo(identity(!!rlang::quo(cyl))))
```
yutannihilation: I don't know why this matters, but if I use `.ignore_empty = "all"`, this fails...

yutannihilation:
Thanks for reviewing!

Issue #2945, opened 2018-10-17
geom_hline() syntax inconsistency - cannot specify yintercept outside of aes when using additional aesthetics


RichardJActon:
It seems that you cannot use additional aesthetics with `geom_hline` if you leave `yintercept` outside of the `aes` call, unlike say `y` in `geom_line`.

    df <- data.frame(
	    "x"=rep(c(1,2,3),2),
	    "y"=rep(c(1,2,3),2),
	    "g"=rep(c(TRUE,FALSE),each=3)
	)

---
This does not work contrary to what you would expect from things like: `geom_line(y=1,aes(colour=x))`

	ggplot(df,aes(x=x,y=y)) + 
	    geom_point() +
	    geom_hline(yintercept = 1.5,aes(colour=g)) +
	    facet_wrap(~g)

![goem_hline](https://user-images.githubusercontent.com/6893043/47107582-9fc51280-d241-11e8-9778-6dda306982f9.png)

This does work but seems counter intuitive. Especially if you wanted, for example, 2 hlines one with a fixed yintercept and one that varied when it would seem more natural to leave the fixed intercept outside of the aesthetic and the variable one in.

    ggplot(df,aes(x=x,y=y)) + 
      geom_point() +
      geom_hline(aes(yintercept = 1.5, colour=g)) +
      facet_wrap(~g)

![geom_hline_aes](https://user-images.githubusercontent.com/6893043/47107618-ace20180-d241-11e8-98fb-f7618bb2cf05.png)

yutannihilation:
When `yintercept` argument is provided, new `data` and `mapping` for the layer are created , so the plot-wide data and mapping will be ignored. This trickiness seems intentional, but I too feel that this is a bit counter intuitive...

https://github.com/tidyverse/ggplot2/blob/09ff5a4f2b608df99d722b3afbc1193a7cea0d39/R/geom-hline.r#L12-L17

`geom_vline()` and `geom_abline()` do the same thing.

https://github.com/tidyverse/ggplot2/blob/09ff5a4f2b608df99d722b3afbc1193a7cea0d39/R/geom-vline.r#L12-L17

https://github.com/tidyverse/ggplot2/blob/09ff5a4f2b608df99d722b3afbc1193a7cea0d39/R/geom-abline.r#L83-L91

RichardJActon:
[...]

PR #2950, opened 2018-10-19
Warn when a supplied mapping is going to be overwritten by geom_hline / vline / abline

RichardJActon:
This pull request addressed issue #2945

geom_hline / vline / abline will now throw a warning if the user supplies both a x|y|intercept/slope value and a mapping such as this:

    geom_hline(yintercept = 3, aes(colour = colour))

As the above will cause the user supplied mapping to be ignored/overwritten

Users might resaonably expect the above to work as other geoms do not have this behaviour. The warning messages suggest that the x|y|intercept/slope value(s) be moved inside the aes call thus:

    geom_hline(aes(yintercept = 3,colour = colour))


R/geom-hline.r
```diff
@@ -9,6 +9,15 @@ geom_hline <- function(mapping = NULL, data = NULL,
                        na.rm = FALSE,
                        show.legend = NA) {
 
+  # Warn if supplied mapping is going to be overwritten
+  if (!missing(yintercept) && !missing(mapping)) {
```
hadley: This would be better placed in the block below

yutannihilation:
Is it OK to recommend to place values directly in `aes()`? In this case, it's fine, but I'm a bit afraid this may be a bit confusing because using direct values in `aes()` is not always good.

Maybe preparing data is the right thing? Or considering the hacky nature of the "annotation mode", is this allowed specially?

clauswilke:
> Maybe preparing data is the right thing?

Actually, `geom_hline()` will overwrite the dataset if one is provided, so this won't work. I've been wondering whether the code block here:
https://github.com/tidyverse/ggplot2/blob/35507725d82d1d9d91fe5d2e4943671b568b59b2/R/geom-hline.r#L13-L17
should only be executed when data is `NULL`.

yutannihilation:
I meant not providing `*intercept` and map all the variables.

```r
data2 <- data
data2$y <- 3

geom_hline(data = data2,
           aes(yintercept = y, colour = colour))
```

But, please ignore my comment above. I think I don't get the setantics of variable mappings yet. Sorry for the noise.

R/geom-hline.r
hadley: I think you forgot to remove the code here

R/geom-hline.r
```diff
@@ -9,8 +9,26 @@ geom_hline <- function(mapping = NULL, data = NULL,
...
+                   "  e.g. `aes(yintercept=2,colour=colour)`"
+            )
+    )
+  }
+
   # Act like an annotation
   if (!missing(yintercept)) {
+    # Warn if supplied mapping is going to be overwritten
+    if (!missing(yintercept) && !missing(mapping)) {
+      warning(paste0("Using both `yintercept` and `mapping` may not have the desired result as mapping is overwritten if `yintercept` is specified\n",
```
hadley: You don't need to check `!missing(yintercept)` again (this is the reason that this code belongs in this block).

For now, I think it would be better if the warning message simply stated what was happening (i.e. setting `yintercept` overrides `mapping`) rather than recommending a solution.

tests/testthat/test-geom-hline-vline-abline.R
```diff
@@ -43,3 +43,38 @@ test_that("curved lines in map projections", {
     nzmap + coord_map(projection = 'azequalarea', orientation = c(-36.92, 174.6, 0))
   )
 })
+
+# Warning tests ------------------------------------------------------------
+
+test_that("Warning if a supplied mapping is going to be overwritten",{
+  dat <- data.frame(x = LETTERS[1:5], y = 1:5,colour = c(TRUE,TRUE,rep(FALSE,3)))
```
hadley: There's no need to construct a plot here since the warning comes from the `geom_vline()` call.

RichardJActon:
Are any further changes needed here?

- I have placed the warning block in the existing check for a missing intercept value and removed the redundant test for the intercept.
- I removed the plot from the test and just checked the geoms directly
- I altered the error message so it does suggest the solution of placing a value directly in a mapping

(I'm not sure what's up with travis - the error says something about exceeding the API rate limit?)

R/geom-abline.r
```diff
@@ -82,6 +82,16 @@ geom_abline <- function(mapping = NULL, data = NULL,
 
   # Act like an annotation
   if (!missing(slope) || !missing(intercept)) {
+
+    # Warn if supplied mapping is going to be overwritten
+    if (!missing(mapping)) {
+      warning(paste0("Using both `intercept` and/or `slope` with `mapping` may",
```
clauswilke: I think the word "both" is misplaced, as it refers to the two words "intercept" and "slope". Maybe write: "Using `intercept` and/or `slope` in combination with `mapping` may"

R/geom-abline.r
```diff
@@ -82,6 +82,16 @@ geom_abline <- function(mapping = NULL, data = NULL,
 
   # Act like an annotation
   if (!missing(slope) || !missing(intercept)) {
```
clauswilke: I still think this code path should only be executed if no dataset is provided. It's not intuitive to overwrite a dataset the user has provided (and here this also happens without any kind of warning or message.) I can't envision any scenario in which replacing an explicitly provided dataset would be the right course of action.

tests/testthat/test-geom-hline-vline-abline.R
```diff
@@ -43,3 +43,33 @@ test_that("curved lines in map projections", {
...
+    "Using both"
+  )
+
+  expect_warning(
+    geom_abline(intercept = 3, aes(colour = colour)),
+    "Using both"
+  )
+
+  expect_warning(
+    geom_abline(intercept = 3,slope=0.5, aes(colour = colour)),
```
clauswilke: Please add a space after the comma.

hadley: And around `=`

clauswilke:
This cannot currently be merged because the code in `geom_abline()`, `geom_hline()`, and `geom_vline()` has changed.

@RichardJActon Would you mind updating this? Could you please also add a bullet to NEWS.md? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

clauswilke:
We just opened a separate issue about the overwriting of `data`: #3090 I suggest you finish this PR as it currently stands.

tests/testthat/test-geom-hline-vline-abline.R
```diff
@@ -43,3 +43,33 @@ test_that("curved lines in map projections", {
     nzmap + coord_map(projection = 'azequalarea', orientation = c(-36.92, 174.6, 0))
   )
 })
+
+# Warning tests ------------------------------------------------------------
+
+test_that("Warning if a supplied mapping is going to be overwritten",{
```
clauswilke: Could you please add a space between the comma and the opening curly brace?

NEWS.md
```diff
@@ -1,4 +1,13 @@
 # ggplot2 3.1.0.9000
+
+* `geom_hline()`, `geom_vline()` & `geom_abline()` now throw a warning if the user supplies both an xintercept, yintercept or slope value & a mapping such as this:
+  
+  ```R
+  geom_hline(yintercept = 3, aes(colour = colour))
+ 
+  ```
+  
+  The above will cause the user supplied mapping to be ignored/overwritten and as users might resaonably expect the above to work given that other geoms do  not have this behaviour, they will now be warned about mapping being overwritten. (@RichardJActon, #2950)
```
clauswilke: This news item seems overly verbose. Something like the first sentence (on line 3) would be sufficient. Also, please write "and" instead of "&" throughout, and please use the Oxford comma: `` `geom_hline()`, `geom_vline()`, and `geom_abline()` ``. Finally, please place the final period after `(@RichardJActon, #2950)`, not before.

clauswilke: Two more comments on the same topic:
1. `xintercept`, `yintercept`, etc should be enclosed in backticks in the text.
2. The Oxford comma also applies to "or".
 
I.e.: `` `xintercept`, `yintercept`, or `slope` ``

clauswilke:
Thanks!

PR #2961, opened 2018-10-25
Broad review of governance structure

hadley:
I've removed some duplication, and attempted to refine the description of the different categories of community member. I've also moved away from the "benevolent dictator" language in favour of "project lead".

What other important parts of the ggplot2 governance process have I forgotten?

PR #2968, opened 2018-10-29
Fix extraneous "the" in GOVERNANCE.md

karawoo:
I think this is what @batpigandme was talking about in #2961


hadley:
Sorry, I fixed this directly locally while un-borking the website 😞

PR #2988, opened 2018-11-09
Fix geom_rug + coord_flip

### LINKED ISSUE #2987: geom_rug and coord_flip do not work together
``` r
library(ggplot2)

df <- data.frame(x=1:10, y=(1:10)^3)
g <- ggplot(df, aes(x, y)) + geom_point() + geom_rug(sides='l')

# Normal plot
g
```

``` r

# coord_flip does not flip the rugs
g + coord_flip()
```

<sup>Created on 2018-11-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

**Problem**

The location of the rugs is contained in the `sides` parameter of `geom_rug`. The parameters of a geom are not passed to the coordinate system and so it cannot manipulate them.

has2k1:
The rugs in `geom_rug` are drawn along an axis depending the
`sides` parameter. Since the coordinate system does not alter
the `geom` parameters, the `geom` has to alter any parameters
that determine where the geoms are plotted.

fixes #2987 

----

``` r
library(ggplot2)

df <- data.frame(x = 1:10, y = (1:10)^3)
g <- ggplot(df, aes(x, y)) + geom_point() + geom_rug(sides = 'l')

# Normal plot
g
```

``` r

# The rugs are flipped
g + coord_flip()
```

<sup>Created on 2018-11-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/geom-rug.r
+++ b/R/geom-rug.r
@@ -66,6 +66,10 @@ GeomRug <- ggproto("GeomRug", Geom,
     rugs <- list()
     data <- coord$transform(data, panel_params)
 
+    if (inherits(coord, 'CoordFlip')) {
+      sides <- chartr('tblr', 'rlbt', sides)
+    }
+
     gp <- gpar(col = alpha(data$colour, data$alpha), lty = data$linetype, lwd = data$size * .pt)
     if (!is.null(data$x)) {
       if (grepl("b", sides)) {
--- a/R/geom-rug.r
+++ b/R/geom-rug.r
@@ -66,6 +66,8 @@ GeomRug <- ggproto("GeomRug", Geom,
     rugs <- list()
     data <- coord$transform(data, panel_params)
 
+    # For coord_flip, coord$tranform does not flip the sides where to
+    # draw the rugs. We have to flip them.
     if (inherits(coord, 'CoordFlip')) {
       sides <- chartr('tblr', 'rlbt', sides)
     }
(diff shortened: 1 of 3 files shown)


R/geom-rug.r
```diff
@@ -66,6 +66,10 @@ GeomRug <- ggproto("GeomRug", Geom,
     rugs <- list()
     data <- coord$transform(data, panel_params)
 
+    if (inherits(coord, 'CoordFlip')) {
```
clauswilke: I wish we wouldn't test for capabilities by testing for inheritance, but this is how it's done for `CoordFlip` all over the ggplot2 code base, so I guess it's fine.

R/geom-rug.r
```diff
@@ -66,6 +66,10 @@ GeomRug <- ggproto("GeomRug", Geom,
...
+      sides <- chartr('tblr', 'rlbt', sides)
```
clauswilke: Would you mind adding a comment to this line explaining what's happening?

R/geom-rug.r
has2k1: The first instinct was to use `is()`, but it was not used anywhere in the coord base.

R/geom-rug.r
has2k1: I have added a comment to the block.

R/geom-rug.r
clauswilke: What I meant was other coord features are queried by asking the coord, e.g. `coord$is_free()`. So it would probably make sense to introduce `coord$is_flip()`. But that's for a different PR.

clauswilke:
@hadley This looks good to me. Any concerns from you?

Issue #2687, opened 2018-06-07
Descent heights still not right


clauswilke:
An issue to be considered after 2.3.0: I have noticed that descent heights are still not correct. The fault lies in this line: https://github.com/tidyverse/ggplot2/blob/4db5122f59920eae59a87abf84c6ac1bbd9f719c/R/margins.R#L67

It looks like a reasonable line, but it turns out that the function `descentDetails()` ignores the font settings of the grob it is given:
``` r
library(grid)
descentDetails(textGrob("pqgj", gp = gpar(fontsize = 5)))
#> [1] 0.0350748697916667inches
descentDetails(textGrob("pqgj", gp = gpar(fontsize = 10)))
#> [1] 0.0350748697916667inches
descentDetails(textGrob("pqgj", gp = gpar(fontsize = 20)))
#> [1] 0.0350748697916667inches
```

Instead it takes the font settings from the currently active viewport. The following works correctly:
``` r
pushViewport(viewport(gp = gpar(fontsize = 5)))
descentDetails(textGrob("pqgj"))
#> [1] 0.0146145290798611inches
popViewport()

pushViewport(viewport(gp = gpar(fontsize = 10)))
descentDetails(textGrob("pqgj"))
#> [1] 0.0292290581597222inches
popViewport()

pushViewport(viewport(gp = gpar(fontsize = 20)))
descentDetails(textGrob("pqgj"))
#> [1] 0.0584581163194444inches
popViewport()
```

The result in ggplot is that we're getting the exact same descent height every time, regardless of theme settings:
``` r
library(ggplot2)
df <- data.frame(x = 1)
ggplot(df, aes(x, x)) + geom_point() +
  ggtitle("gjpjQ") +
  theme(plot.title = element_text(size = 10))
```

![](https://i.imgur.com/yrFT0c7.png)

``` r
ggplot(df, aes(x, x)) + geom_point() +
  ggtitle("gjpjQ") +
  theme(plot.title = element_text(size = 20))
```

![](https://i.imgur.com/oweV802.png)

``` r
ggplot(df, aes(x, x)) + geom_point() +
  ggtitle("gjpjQ") +
  theme(plot.title = element_text(size = 100))
```

![](https://i.imgur.com/8O59Gdg.png)

The reason why things seem to work correctly is because the default fontsize is 12, so we're getting a descent height that is approximately right for most typical font sizes.

[...]

PR #2993, opened 2018-11-12
Memoise descent

thomasp85:
The current caching of descentDetails calls is based on the device in use, along with all relevant font specifications.

I believe @clauswilke has looked into something similar but unsure what he ended up with suggesting - this works well, though...


DESCRIPTION
```diff
@@ -32,7 +32,8 @@ Imports:
     stats,
     tibble,
     viridisLite,
-    withr (>= 2.0.0)
+    withr (>= 2.0.0),
```
hadley: Can you please keep this in alphabetical order?

thomasp85:
Agree - this started off by using memoise, but the digest call was a measurable overhead when the arguments to memoise on were simple strings and integers that could be pasted together to a key

clauswilke:
Yes, this is basically what I did here: https://github.com/clauswilke/gridtext/blob/f46c0caa018ed32fefc2ba13240e272b37ba7eed/R/grob-descent.R#L25-L55

Two comments:

1. This does not yet fix #2687, correct?

2. Should the descent cache be added to `ggplot2_global`? I had added that global environment a while back to collect all these global variables in one place.
https://github.com/tidyverse/ggplot2/blob/f5a88a71e7b60b6531ef6c4a936370b5e3c4ef92/R/ggplot-global.R#L1-L4

thomasp85:
1) It does not - I had forgot about that issue, but will incorporate it

2) I'm a bit unsure about the purpose of ggplot_global so I can't really say if it should be added

hadley:
I think it's probably better to have multiple independent environments that are declared close to the code that uses them.

thomasp85:
1 is now fixed in #2996

clauswilke:
The history of `ggplot_global` is that I started working on user-customizable theme elements and for that needed to put the element tree into an environment. While doing that, I found a few other global variables that were strewn about in the code and thought it might be good to put them all into one place. This is the PR: https://github.com/tidyverse/ggplot2/pull/2741, though it doesn't seem to have the associated discussion on this topic.

One difference between those global variables and the font cache here is that the font cache should never be accessed by anything other than the `font_descent()` function. By contrast, the element tree is accessed by numerous functions all over the codebase.

Of course, then @hadley suggested to make the element tree part of the theme itself, and if we went that route then we won't ever need to interactively change the default element tree that is stored in the global element tree variable, and thus could conceivably get rid of `ggplot_global` after all. There's a stalled PR on that topic, I'll have to revive it at some point: https://github.com/tidyverse/ggplot2/pull/2784

PR #2994, opened 2018-11-12
New data frame

thomasp85:
This is a very sweeping PR, that changes almost all `data.frame()` and `as.data.frame()` calls to use the included `new_data_frame()` constructor.

unit tests and examples does not show any problems with this, but it is possible that reverse dependency check will surface some places where this PR introduces errors, so we have to look out for that (or run a revdep check on this branch before merging)


clauswilke:
I'm all on board with avoiding recycling in principle, but it seems to me that giving up on the ability to construct data frames from constants and vectors is a big loss and puts a lot of extra mental requirements on the code writer. I think this will almost certainly result in new bugs at some point.

Would it be possible/make sense to rewrite the constructor with Rcpp and do a basic sanity check (all input vectors are the same length or constants) and recycle constants only? Would that generate a lot of overhead?

I did an experiment a while back and if I remember correctly I could construct basic R objects such as structures quite a bit faster using Rcpp than using `structure()` and `list()`.

thomasp85:
`structure()` carries its own overhead so I'm not surprised you could beat that...

I've toyed with the idea of a recycling list constructor so that you could do something like this:

```r
new_data_frame(list_recycle(
  letters = c('a', 'b', 'c'),
  one_integer = 1L
))
```

this would keep the minimal constructor but allow some ease when creating data frames from a mix of constants and vectors

R/axis-secondary.R
```diff
@@ -140,7 +140,7 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
   },
 
   transform_range = function(self, range) {
-    range <- structure(data.frame(range), names = '.')
+    range <- new_data_frame(list("." = range))
```
hadley: You don't actually need the `"` here

hadley:
I think the long term performance plan is to rely on vctrs or tibble for the low-level code.

@thomasp85 for performance PRs like this, could you please include a small summary of the change in the performance in the PR?

clauswilke:
One other issue: If we're starting to introduce performant replacements for widely used R functions then I think recommended use, best practices, and potential pitfalls need to be clearly documented. Not sure where this would go, but maybe it's time to start a vignette that describes internal coding practices for ggplot2?

hadley:
@clauswilke that what I'm thinking with the `data.frame()` shim I described above — that makes it a good place to include documentation, because you're forced to read it when you accidentally use the other function. In my experience, project documentation alone is not sufficient (even when I'm the only person contributing code 🤣)

clauswilke:
@hadley I think we need all of the above. The proposed plan for `data.frame()` is good, but I don't think it'll make sense to hide all relevant documentation in that one error message. I'd propose something like the following:

``` r
data.frame <- function(...) stop('Please use `data_frame()` instead of `data.frame()` for better performance. See the vignette "ggplot2 internal programming guidelines" for details.')
```

There are also things that I can think of adding to such a vignette where one wouldn't be able to create such a check, such as best practices on working with aesthetics.

thomasp85:
`new_data_frame` now does recycling and takes `n` as the longest of the elements in `x` if not given. `data_frame` is simply `new_data_frame(list(...))`, and `data.frame` has been masked with the error message provided by @clauswilke (we'll need to write that vignette obviously).

The overhead of recycling was minor (~1µs) so I think the safety of it makes it fair to have in both constructors.

One unexpected side effect of masking `data.frame()` was that it also affected all test code - it has been updated accordingly.

The performance benefit of this is very much related to the nature of the plot. In a standard scatter-plot it is going to be non-existing (or at least hidden by more taxing operations) as there is only 5 data frames being constructed. The moment you begin to have faceting along with stats implementing `compute_group` and geoms implementing `draw_group` it becomes significant, and I could easily get a 20% reduction in `ggplotGrob()` time on a facetted boxplots of the diamonds dataset

R/aaa-.r
```diff
@@ -12,3 +12,45 @@ NULL
 #' @keywords internal
 #' @name ggplot2-ggproto
 NULL
+
+# Fast data.frame constructor and indexing
+# No checking, recycling etc. unless asked for
+new_data_frame <- function(x = list(), n = NULL) {
+  if (is.null(n)) {
+    n <- if (length(x) == 0) 0 else max(lengths(x))
```
hadley: You'll need an alternative to `lengths()` for older versions of R

R/aaa-.r
```diff
@@ -12,3 +12,45 @@ NULL
...
+  }
+  for (i in seq_along(x)) {
+    if (length(x[[i]]) != n) x[[i]] <- rep(x[[i]], length.out = n)
```
hadley: This is now too flexible - it recycles vectors of any length

R/aaa-.r
```diff
@@ -12,3 +12,45 @@ NULL
...
+
+data_frame <- function(...) {
+  new_data_frame(list(...))
+}
+
+data.frame <- function(...) {
+  stop('Please use `data_frame()` or `new_data_frame()` instead of `data.frame()` for better performance. See the vignette "ggplot2 internal programming guidelines" for details.', call. = FALSE)
+}
+
+validate_data_frame <- function(x) {
```
hadley: Is this used anywhere? It might be better to just inline into `data_frame()`

thomasp85:
@Hadley can you look this through again and see if you are content with the implementation?

R/aaa-.r
```diff
@@ -12,3 +12,44 @@ NULL
...
+data_frame <- function(...) {
+  new_data_frame(list(...))
+}
+
+data.frame <- function(...) {
+  stop('Please use `data_frame()` or `new_data_frame()` instead of `data.frame()` for better performance. See the vignette "ggplot2 internal programming guidelines" for details.', call. = FALSE)
+}
+
+mat_2_df <- function(x, col_names = NULL, .check = FALSE) {
+  if (is.null(col_names)) col_names <- colnames(x)
```
hadley: Make this the default value?

R/aaa-.r
```diff
@@ -12,3 +12,44 @@ NULL
...
+  if (is.null(n)) {
+    n <- if (length(x) == 0) 0 else max(lengths)
+  }
+  for (i in seq_along(x)) {
+    if (lengths[i] == n) next
+    if (lengths[i] != 1) stop("Elements must equal the number of rows or 1", call. = FALSE)
+    x[[i]] <- rep(x[[i]], n)
+  }
+
+  class(x) <- "data.frame"
```
hadley: Might be a little cleaner to set with `structure` ?

R/aaa-.r
```diff
@@ -12,3 +12,44 @@ NULL
...
+  new_data_frame(list(...))
+}
+
+data.frame <- function(...) {
+  stop('Please use `data_frame()` or `new_data_frame()` instead of `data.frame()` for better performance. See the vignette "ggplot2 internal programming guidelines" for details.', call. = FALSE)
+}
+
+mat_2_df <- function(x, col_names = NULL, .check = FALSE) {
+  if (is.null(col_names)) col_names <- colnames(x)
+  x <- split(x, rep(seq_len(ncol(x)), each = nrow(x)))
```
hadley: Would `lapply(seq_len(ncol(x)), function(i) x[, i])` be faster/cleaner?

R/annotation-logticks.r
```diff
@@ -238,7 +238,7 @@ calc_logticks <- function(base = 10, ticks_per_base = base - 1,
   longtick_after_base <- floor(ticks_per_base/2)
   tickend[ cycleIdx == longtick_after_base ] <- midend
 
-  tickdf <- data.frame(value = ticks, start = start, end = tickend)
+  tickdf <- new_data_frame(list(value = ticks, start = rep(start, length(ticks)), end = tickend), n = length(ticks))
```
hadley: Probably worth going through and removing the `rep()`s, now that we no longer need them.

R/aaa-.r
thomasp85: `structure()` is measurably slower... If the purpose is performance I think we can sacrifice a bit of cleanliness

R/annotation-logticks.r
thomasp85: Oh, I thought I had removed them all... will give it a second pass

R/aaa-.r
hadley: Oh wow, so it is. Maybe this means vctrs will need it's own `structure()` alternative.

R/aaa-.r
thomasp85: Wow - that is *much* faster

R/aaa-.r
thomasp85: I was a bit surprised, honestly

thomasp85:
ok, @Hadley everything should be fixed now

clauswilke:
There's one test remaining that uses `data.frame()`. Looks like a race condition between your commit and @yutannihilation's recent commit that introduced that line.
https://github.com/tidyverse/ggplot2/blob/4380cb97d55a91b6ef430d5b23327fab5c421b27/tests/testthat/test-layer.r#L42

thomasp85:
yeah, I've fixed this in #3003

PR #2997, opened 2018-11-13
Allow using functions that return NULL in aes()

yutannihilation:
It's common practice to wrap unsafe calculations with `tryCatch()` and return `NULL`, which allows us to use the result when available and skip when unavailable. But, currently, ggplot2 doesn't accept this in `aes()`:

``` r
library(ggplot2)

wrap <- function(...) tryCatch(..., error = function(e) NULL)

ggplot(df, aes(x, y, colour = wrap(no_such_column))) +
  geom_point()
#> Error: Aesthetics must be either length 1 or the same as the data (10): colour
```

This is because `Layer$compute_aesthetics()` removes `NULL` aes here BEFORE evaluation:

https://github.com/tidyverse/ggplot2/blob/f5a88a71e7b60b6531ef6c4a936370b5e3c4ef92/R/layer.r#L221-L223

Considering the following facts, I think it's safe to move `compact()` after evaluation

- `ggplot2::compact()` removes only `NULL`
- `NULL` will be still `NULL` after evaluation

### DIFF
--- a/tests/testthat/test-layer.r
+++ b/tests/testthat/test-layer.r
@@ -40,8 +40,8 @@ test_that("missing aesthetics trigger informative error", {
 
 test_that("if an aes is mapped to a function that returns NULL, it is removed", {
   df <- data.frame(x = 1:10)
-  wrap <- function(...) tryCatch(..., error = function(e) NULL)
-  p <- cdata(ggplot(df, aes(x, wrap(no_such_column))))
+  null <- function(...) NULL
+  p <- cdata(ggplot(df, aes(x, null())))
   expect_identical(names(p[[1]]), c("x", "PANEL", "group"))
 })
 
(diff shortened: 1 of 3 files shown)


tests/testthat/test-layer.r
```diff
@@ -38,6 +38,13 @@ test_that("missing aesthetics trigger informative error", {
   )
 })
 
+test_that("if an aes is mapped to a function that returns NULL, it is removed", {
+  df <- data.frame(x = 1:10)
+  wrap <- function(...) tryCatch(..., error = function(e) NULL)
```
hadley: I think you can simplify this to just `function() NULL`

hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

tests/testthat/test-layer.r
yutannihilation: Agreed.

yutannihilation:
Thanks! I tend to forget about NEWS bullets, sorry.

PR #2998, opened 2018-11-13
Update vdiffr svgs after descent changes

dpseidel:
The changes to descent in #2996 broke numerous vdiffr tests by slightly altering plot height (see screenshots below). This PR updates the svg files. 

@thomasp85, I've gone through and inspected the mismatches, all containing the same slight shift, but perhaps you should review to confirm this shift is an expected and acceptable change in plot behavior before we update the test case files. 

See slight shift in the xaxis height:
vdiffr before/after 2996

The differences are much more pronounced in the "theme-large" tests, e.g. :
theme-large vdiffr before/after 2996


dpseidel:
Great, thanks! if your local vdiffr version is not up to date, visual tests will be skipped silently so I'd check that first.

PR #3003, opened 2018-11-15
Only use data_frame in tests

thomasp85:
Some tests were added between the new_data_frame implementation and it's merge that used the now masked data.frame constructor. This PR fixes these so tests now passes


clauswilke:
I see one more problem, here:
https://github.com/tidyverse/ggplot2/blob/4380cb97d55a91b6ef430d5b23327fab5c421b27/tests/testthat/test-geom-polygon.R#L13-L15
Not immediately sure what the cause is.

``` r
library(ggplot2)
ggplot(faithful, aes(x = eruptions, y = waiting)) +
  stat_density2d(aes(fill = stat(level)), geom = "polygon", colour = "white") +
  xlim(0.5, 6) + ylim(40, 110)
#> Warning in validDetails.polygon(x): NAs introduced by coercion
```



<sup>Created on 2018-11-15 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
One other, minor issue that could be fixed as part of this PR: The `rep()` inside `max()` seems superfluous here:
https://github.com/tidyverse/ggplot2/blob/92d27779e210ce16f441eeec26926b8b045a47a8/R/stat-contour.r#L96

yutannihilation:
`GeomPolygon`'s error seems because now `munched$group` is character here, whereas `polygonGrob()` only accepts numeric vectors. I didn't know the mercy of `stringsAsFactors` secretly saved us...

https://github.com/tidyverse/ggplot2/blob/4380cb97d55a91b6ef430d5b23327fab5c421b27/R/geom-polygon.r#L96-L105

yutannihilation:
The `group` comes from here. So you can wrap this `as.factor()` or convert it to integer before passing it to `polygonGrob()`. I feel the latter approach is right, but I'm not fully sure what it is supposed to be.

https://github.com/tidyverse/ggplot2/blob/4380cb97d55a91b6ef430d5b23327fab5c421b27/R/stat-contour.r#L92-L101

thomasp85:
Thanks - weirdly this didn't pop up locally... 🤔

PR #3011, opened 2018-11-19
Handle Turkish dotted and dotless i properly

### LINKED ISSUE #3000: Error: Can't find `stat` called “identity”
I didn't find any results on my internet search. How to fix this error is caused?

My codes;

```
> library(ggplot2)
> 
> ggplot(data = mtcars, aes(x=disp , y=wt)) + geom_point(stat = "identity")
```

result;

`Error: Can't find `stat` called "identity"`

My sessionInfo()

```
R version 3.5.1 (2018-07-02)
Running under: Ubuntu 18.04.1 LTS

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:
 [1] LC_CTYPE=tr_TR.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=tr_TR.UTF-8        LC_COLLATE=tr_TR.UTF-8    
 [5] LC_MONETARY=tr_TR.UTF-8    LC_MESSAGES=tr_TR.UTF-8   
 [7] LC_PAPER=tr_TR.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=tr_TR.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] ggplot2_3.1.0 shiny_1.2.0  

 [1] Rcpp_1.0.0         pillar_1.3.0       compiler_3.5.1    
 [4] cellranger_1.1.0   RColorBrewer_1.1-2 later_0.7.5       
 [7] plyr_1.8.4         bindr_0.1.1        forcats_0.3.0     
[10] tools_3.5.1        digest_0.6.18      viridisLite_0.3.0 
[13] jsonlite_1.5       tibble_1.4.2       gtable_0.2.0      
[16] pkgconfig_2.0.2    rlang_0.3.0.1      ggplotgui_1.0.0   
[19] rstudioapi_0.8     yaml_2.2.0         haven_1.1.2       
[22] bindrcpp_0.2.2     withr_2.1.2        stringr_1.3.1     
[25] httr_1.3.1         dplyr_0.7.8        htmlwidgets_1.3   
[28] hms_0.4.2          grid_3.5.1         tidyselect_0.2.5  
[31] data.table_1.11.8  glue_1.3.0         R6_2.3.0          
[34] plotly_4.8.0       readxl_1.1.0       readr_1.1.1       
[37] tidyr_0.8.2        purrr_0.2.5        magrittr_1.5      
[40] scales_1.0.0       promises_1.0.1     htmltools_0.3.6   
[43] assertthat_0.2.0   xtable_1.8-3       mime_0.6          
[46] colorspace_1.3-2   httpuv_1.4.5       stringi_1.2.4     
[49] lazyeval_0.2.1     munsell_0.5.0      crayon_1.3.4


```

yutannihilation:
Fix #3000 

Turkish has two versions of i; dotted and dotless. Accordingly, the capital of `i` is `İ`, not `I`, which make ggplot2 to fail to find `PositionIdentity` and `StatIdentity`. The same thing can happen on dotless `ı` and `I` (c.f. http://www.i18nguy.com/unicode/turkish-i18n.html)

This seems a well-known problem, and ICU has a flag for this, but unfortunately, there's no way to use this via R's `toupper()`.

> It is also not language-sensitive, although there is a flag for whether to apply special mappings for use with Turkic (Turkish/Azerbaijani) text data.
(http://userguide.icu-project.org/transforms/casemappings)

After googling around, I found several options to fix this:

1. Use `chartr()` instead of `toupper()` and `tolower()` (This PR)
2. Use `stringi::stri_trans_toupper()` and `stringi::stri_trans_tolower()`, where we can specify `locale`.
3. Modify the locale temporarily

For 2., stringi is cool, but adding it to the dependency is a bit heavy. For 3., it seems dangerous to mess the user's locale. So, from these, I think option 1 is rather reasonable, although not very cool. `chartr()` is around 4x slower than `toupper()`, but I hope this is acceptable considering that people in Turkish locale probably almost cannot use ggplot2... (I'm not familiar with this problem, so I might be wrong.)

```r
bench::mark(
  chartr("abcdefghijklmnopqrstuvwxyz", "ABCDEFGHIJKLMNOPQRSTUVWXYZ", "a"),
  toupper("a")
)
#> # A tibble: 2 x 10
#>   expression     min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:t> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 "chartr(\~   3.3us 7.03us  3.8us 732us   142357.        0B     0 10000
#> 2 "toupper(~   700ns 2.24us    1us 291us   446879.        0B     0 10000
#> # ... with 1 more variable: total_time <bch:tm>
```

### DIFF
--- a/R/utilities.r
+++ b/R/utilities.r
@@ -302,6 +302,12 @@ has_name <- function(x) {
   !is.na(nms) & nms != ""
 }
 
+# Use chartr() for safety since toupper() fails to convert i to I in Turkish locale
+lower_ascii <- "abcdefghijklmnopqrstuvwxyz"
+upper_ascii <- "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+tolower_ascii <- function(x) chartr(upper_ascii, lower_ascii, x)
+toupper_ascii <- function(x) chartr(lower_ascii, upper_ascii, x)
+
 # Convert a snake_case string to camelCase
 camelize <- function(x, first = FALSE) {
   x <- gsub("_(.)", "\\U\\1", x, perl = TRUE)
@@ -313,11 +319,11 @@ snakeize <- function(x) {
   x <- gsub("([A-Za-z])([A-Z])([a-z])", "\\1_\\2\\3", x)
   x <- gsub(".", "_", x, fixed = TRUE)
   x <- gsub("([a-z])([A-Z])", "\\1_\\2", x)
-  tolower(x)
+  tolower_ascii(x)
 }
 
 firstUpper <- function(s) {
-  paste(toupper(substring(s, 1,1)), substring(s, 2), sep = "")
+  paste0(toupper_ascii(substring(s, 1, 1)), substring(s, 2))
 }
 
 snake_class <- function(x) {
--- a/R/utilities.r
+++ b/R/utilities.r
@@ -308,6 +308,14 @@ upper_ascii <- "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 to_lower_ascii <- function(x) chartr(upper_ascii, lower_ascii, x)
 to_upper_ascii <- function(x) chartr(lower_ascii, upper_ascii, x)
 
+tolower <- function(x) {
+  stop('Please use `to_lower_ascii()`, which works fine in Turkish locale.', call. = FALSE)
+}
+
+toupper <- function(x) {
+  stop('Please use `to_upper_ascii()`, which works fine in Turkish locale.', call. = FALSE)
+}
+
 # Convert a snake_case string to camelCase
 camelize <- function(x, first = FALSE) {
   x <- gsub("_(.)", "\\U\\1", x, perl = TRUE)
--- a/R/utilities.r
+++ b/R/utilities.r
@@ -305,8 +305,8 @@ has_name <- function(x) {
 # Use chartr() for safety since toupper() fails to convert i to I in Turkish locale
 lower_ascii <- "abcdefghijklmnopqrstuvwxyz"
 upper_ascii <- "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-tolower_ascii <- function(x) chartr(upper_ascii, lower_ascii, x)
-toupper_ascii <- function(x) chartr(lower_ascii, upper_ascii, x)
+to_lower_ascii <- function(x) chartr(upper_ascii, lower_ascii, x)
+to_upper_ascii <- function(x) chartr(lower_ascii, upper_ascii, x)
 
 # Convert a snake_case string to camelCase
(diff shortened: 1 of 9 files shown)


R/utilities.r
```diff
@@ -302,6 +302,12 @@ has_name <- function(x) {
   !is.na(nms) & nms != ""
 }
 
+# Use chartr() for safety since toupper() fails to convert i to I in Turkish locale
+lower_ascii <- "abcdefghijklmnopqrstuvwxyz"
+upper_ascii <- "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+tolower_ascii <- function(x) chartr(upper_ascii, lower_ascii, x)
```
hadley: Can we make this `to_lower_ascii()` and `to_upper_ascii()`?

And I think we should wrap `to_lower()` and `to_upper()` to throw an error, to make sure that this error doesn't arise again in the future.

yutannihilation:
Ah, sorry, it seems I'm doing wrong about detecting the available locales... Let me fix.

> `any(x == "tr_TR") isn't true.`
https://travis-ci.org/tidyverse/ggplot2/builds/457214479#L5545

hadley:
Sorry my comment was supposed to reinforce _not_ having a unit test. I don't think it gets us much here as it's likely to be fragile, and we've reduced the chance by overwriting `toupper()`

R/utilities.r
```diff
@@ -302,6 +302,20 @@ has_name <- function(x) {
...
+upper_ascii <- "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+to_lower_ascii <- function(x) chartr(upper_ascii, lower_ascii, x)
+to_upper_ascii <- function(x) chartr(lower_ascii, upper_ascii, x)
+
+tolower <- function(x) {
+  stop('Please use `to_lower_ascii()`, which works fine in Turkish locale.', call. = FALSE)
+}
+
+toupper <- function(x) {
+  stop('Please use `to_upper_ascii()`, which works fine in Turkish locale.', call. = FALSE)
```
hadley: "in Turkish locale" -> "in all locales"

yutannihilation:
Oh, I got it wrong, sorry. Agreed.

yutannihilation:
Thanks for confirmation, @cystein!

PR #3013, opened 2018-11-23
No plyr

thomasp85:
This PR writes out all ply dependencies from ggplot2. The following strategies has been used:

#### **ply
Except for `ddply` all use of `**ply` functions has been substituted with base equivalents. `ddply` has been substituted with the new `dapply` function that mimics its API and behaviour. It is ~40% faster in normal use by avoiding the `as.quoted` call.

#### Utility functions
Use of `split_indices`, `summarise`, and `colwise` has been substituted with base equivalent approaches.

`quickdf` has been substituted with the already existing `new_data_frame`

`empty` is now exported so the vignette doesn't need to depend on the plyr version

`defaults`, `unrowname`, `rename`, `revalue`, `id`, `count`, `join.keys`, `as.quoted`, `round_any`, and `rbind.fill` has all been reimplemented. In a few circumstances they have been lifted as-is from plyr, but often they have been rewritten to reduce complexity by only targeting use in ggplot2. All reimplementations are either on par or slightly faster than the plyr counterparts

Currently, a selection of reimplementations has been exported as internal functions to make it easier for extension developers to write out plyr as well.

All rewritten functions are currently collected in `compat-plyr.R` but I think it makes sense to distribute them into files guided by their use rather than keep them there for historical reasons.

There are no unit tests yet but these will come once we have agreed on the implementations and approaches.


R/compat-plyr.R
```diff
@@ -0,0 +1,334 @@
+# Adds missing elements to a vector from a default vector
+defaults <- function(x, y) c(x, y[setdiff(names(y), names(x))])
```
clauswilke: Can we export `defaults()` also? In my experience, it's the plyr function most likely to be used by extension developers, since it's used to adjust default aesthetics in geoms.

thomasp85: Sure

yutannihilation:
r-devel complains about plyr in tests:

```
checking for unstated dependencies in ‘tests’ ... WARNING
'::' or ':::' import not declared from: ‘plyr’
```
(https://travis-ci.org/tidyverse/ggplot2/jobs/459483350#L1765)

It seems you need to replace two more plyr things:

* `ddply()` in `tests/testthat/helper-plot-data.r`
* `arrange()` in `tests/testthat/test-fortify.r`

R/bench.r
```diff
@@ -23,7 +23,7 @@ benchplot <- function(x) {
 
   times <- rbind(construct, build, render, draw)[, 1:3]
 
-  plyr::unrowname(base::data.frame(
+  unrowname(base::data.frame(
```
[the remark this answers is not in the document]
thomasp85: I think my reason for doing this was that I was unsure about the consistency of the input... I'll look into it

hadley:
Sounds good.

jcheng5:
Thank you for this @thomasp85! I was profiling one of my ggplots to try to speed it up, and noticed a bunch of time spent in plyr, but when I came to look for the source code on master I saw that you'd made this change. Just by installing from master, the time dropped from 1020ms to 720ms!

image

thomasp85:
Great to hear, though a lot of the performance improvements is probably due to text dimension caching implemented in #2993...

You should get an additional speedup by installing the development version of gtable

PR #3023, opened 2018-12-04
Sort ribbon/area in setup_data

thomasp85:
This PR streamlines the handling of data in line-like gloms where an implicit sorting of data along the x-axis is happening. Previously GeomLine had the sorting happening in `setup_data` while GeomRibbon (and by extension GeomArea) had it inside `draw_group`.

The motivation is that it solves some issues in gganimate that couldn't be fixed outside of ggplot2, but I also think that conceptually this is the right approach, and we should be consistent in how we treat data operations across kindred geoms/stats

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,9 @@
 
 * ggplot2 now works in Turkish locale (@yutannihilation, #3011).
 
+* `GeomRibbon` and `GeomArea` now sort the data along the x-axis in the 
+  `setup_data()` method rather than as part of `draw_group()` (@thomasp85, #3023)
+
 # ggplot2 3.1.0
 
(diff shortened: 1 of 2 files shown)


thomasp85:
Yeah, maybe... I think in the worst case they just do an additional sort on an already sorted data.frame, but who knows... I'll add a note

clauswilke:
Actually, if they implement their own `setup_data()` function but use the `draw_group()` function provided by `GeomRibbon` then they'll draw unsorted data. Just like you had to fix `setup_data()` for `GeomArea`.

thomasp85:
That is true. The note is there anyway, but I don’t expect any problems

PR #3024, opened 2018-12-05
fix nudging with multiple values. closes #2977

### LINKED ISSUE #2977: Multiple nudge arguments results in warning
The documentation for 'position_nudge' implies that you should only give one value for nudge, but isn't explicit. I often give different nudge values for each label so I can nudge them in different directions.

This works, but results in a warning. I would suggest fixing the if() statement to check multiple values, or make it explicit that we shouldn't be using more than one nudge value.

Thanks!

``` r
library(ggplot2)

df <- data.frame(
  x = c(1,3,2,5),
  y = c("a","c","d","c")
)

ggplot(df, aes(x, y)) +
  geom_point() +
  geom_text(aes(label = y), nudge_y = c(-0.1, 0.1, -0.1, -0.1))

#> Warning in if (params$y != 0) {: the condition has length > 1 and only the
#> first element will be used
```

<sup>Created on 2018-11-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
Fix spurious warning in `position_nudge()`, as described in #2977.

``` r
library(ggplot2)

df <- data.frame(
  x = c(1,3,2,5),
  y = c("a","c","d","c")
)

ggplot(df, aes(x, y)) +
  geom_point() +
  geom_text(aes(label = y), nudge_y = c(-0.1, 0.1, -0.1, -0.1))
```

<sup>Created on 2018-12-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/position-nudge.R
+++ b/R/position-nudge.R
@@ -47,13 +47,13 @@ PositionNudge <- ggproto("PositionNudge", Position,
 
   compute_layer = function(data, params, panel) {
     # transform only the dimensions for which non-zero nudging is requested
-    if (params$x != 0) {
-      if (params$y != 0) {
+    if (any(params$x != 0)) {
+      if (any(params$y != 0)) {
         transform_position(data, function(x) x + params$x, function(y) y + params$y)
       } else {
         transform_position(data, function(x) x + params$x, NULL)
       }
-    } else if (params$y != 0) {
+    } else if (any(params$y != 0)) {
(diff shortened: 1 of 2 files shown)


clauswilke:
Ok. I'll merge this now, since it's such a small change and really was an oversight when I wrote the code originally.

PR #3034, opened 2018-12-12
Stricter unit handling

thomasp85:
As part of my rewrite of the grid unit system I came upon these two cases where units are constructed, but then subsequently treated as numeric vectors. Apart from the unneeded overhead, this can potentially mess up unit construction downstream if the units are not in the expected structure


R/coord-.r
```diff
@@ -64,10 +64,10 @@ Coord <- ggproto("Coord",
...
 
   render_bg = function(panel_params, theme) {
-    x.major <- if (length(panel_params$x.major) > 0) unit(panel_params$x.major, "native")
-    x.minor <- if (length(panel_params$x.minor) > 0) unit(panel_params$x.minor, "native")
-    y.major <- if (length(panel_params$y.major) > 0) unit(panel_params$y.major, "native")
-    y.minor <- if (length(panel_params$y.minor) > 0) unit(panel_params$y.minor, "native")
+    x.major <- if (length(panel_params$x.major) > 0) panel_params$x.major
+    x.minor <- if (length(panel_params$x.minor) > 0) panel_params$x.minor
+    y.major <- if (length(panel_params$y.major) > 0) panel_params$y.major
+    y.minor <- if (length(panel_params$y.minor) > 0) panel_params$y.minor
```
clauswilke: It's not clear to me whether `guide_grid()` takes numerical values or units as arguments. If I understand correctly, it hands these arguments off to `element_render()` which ultimately hands them of to `polylineGrob()` which can take either numeric values or units for its x and y arguments. I think this ambiguity in what arguments grobs take leads to a lot of confusing code and potential errors.

Maybe, as a small step forward, it would help to add a comment to the definition of `guide_grid()` explaining what the variable type and format of `x.minor`, `x.major`, etc should be.

thomasp85: Yeah, some documentation is probably in order. The offending lines are the two calls to setdiff which removes any attributes on the output so that polylineGrob ends up constructing the unit with default.unit anyway

R/guide-legend.r
```diff
@@ -656,7 +656,7 @@ guide_gengrob.legend <- function(guide, theme) {
   krows <- rep(vps$key.row, each = ngeom)
 
   # padding
-  padding <- convertUnit(theme$legend.margin %||% margin(), "cm")
+  padding <- convertUnit(theme$legend.margin %||% margin(), "cm", valueOnly = TRUE)
```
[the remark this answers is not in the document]
thomasp85: Good catch

thomasp85:
@clauswilke I've updated based on your comments

Issue #2008, opened 2017-01-20
problem with test for monotonous transformation (secondary axis)?


hansvancalster:
```{r}
testdat <- data.frame(x = runif(11),
                      y = seq(0,1,0.1))

# NOT OK
ggplot(data = testdat, aes(x = x, y = y)) + 
  geom_point() + 
  scale_y_continuous(sec.axis = sec_axis(trans = ~ .^0.5))

#  OK
ggplot(data = testdat[testdat$y > 0, ], aes(x = x, y = y)) + 
  geom_point() + 
  scale_y_continuous(sec.axis = sec_axis(trans = ~ .^0.5))
```

hadley:
I've cleaned up the issue title/comment thread to focus on key components of problem. Thanks for the correspondence so far :smile:

hadley:
This is the same root cause as #980 — the problem is that the transformation has to be monotonous over the range of the scales, not the range of the data.   You can work around it by eliminating the scale expansion:

``` r
library(ggplot2)
df <- data.frame(x = runif(5), y = seq(0,1, length = 5))

ggplot(df, aes(x, y)) + 
  scale_y_continuous(
    sec.axis = sec_axis(trans = ~ .^0.5), 
    expand = c(0, 0)
  )
```

![](http://i.imgur.com/JQ7Is3V.png)

hansvancalster:
Thanks! That explains what happened and the work-around is OK.

PR #3040, opened 2018-12-16
Fix tick misalignment and revert transformation inheritance in sec.axis (#2978)

### LINKED ISSUE #2978: Ticks misaligned for sec_axis with some scale transformations and data in 3.1.0

dpseidel:
### Version 3.1.0 introduced significant regressions in regards to the secondary axis API. 
These regressions have been jointly documented in issue #2978 ad PR #3004 with reprexes but I will just briefly review them here. 

1) changes made in #2796, eliminated the rescaling of the secondary axis to that of the primary axis, 
introducing tick alignment issues when non-linear secondary transformations were called.

2) changes made in #2805, reversed the order in which range transformation and sampling were completed, introducing poorly aligned ticks in the secondary axis in the case of non-linear transformations. 

These two regressions taken together, break the expected sec.axis tick positioning for some [custom transformation functions](https://github.com/tidyverse/ggplot2/issues/2978#issuecomment-435487510). 

3) Finally, in a more fundamental change, changes made in 2796 to inherit the transformation of the primary scale when creating the secondary axis (allowing for simple implementation of secondary axes with date and datetime scales, and transformed break defaults), introduced the assumption/requirement that the two axes shared a transformation, breaking [any use cases that would treat the two axes as independent](https://github.com/tidyverse/ggplot2/issues/2978#issuecomment-442651113). 

PR #3004 provided a fix for the first two issues, but would leave plots implementing independent primary and secondary scale transformations broken. In its stead, this PR reverts both 2796 and 2805 implementing a more conservative fix for the original motivating issue (log transformed breaks #2729) in the `init()` function of `AxisSecondary`. 

https://github.com/dpseidel/ggplot2/blob/fded720827358ad0012fa635a5af5292f45f4c1c/R/axis-secondary.R#L139

I believe this is the best course of action in light of the nature of all the regressions, but this fix opens two new issues:

#### 1. Renewed spurious monotonicity test failures given power transformations with data near 0. (present in 3.0.0)

Upon digging into it, I realized the monotonicity issue is one we have chosen not to remedy in the past (#2008) simply offering a work-around: users should specify `expand = c(0,0)`. From a users perspective it is rather unexpected that a duplicate secondary axis would fail for a primary that does not, but because the range passed to the sec.axis is already expanded, there does not appear to be a simple solution. 

#### 2. lost functionality for secondary date/datetime axes (datetime scales)
To preserve the current functionality of sec.axis with date and date time scales introduced in 3.1, I have opted for a bit of a hack: a conditional that passes the datetime transformation to the sec.axis scale preserving the current (inheritance) behavior in 3.1 only for date and datetime scales. I don't think this is a permanent solution and would appreciate advice or suggestion. At the very least we need to better document the expected functionality and limitations of sec.axis with all continuous scales, perhaps  explicitly supporting more limited sec.axis functionality with datetime scales. 

It seems to me that if we are preserving independent transformations for all other secondary axes, we should do so for date and datetime scales too. It's not implausible that users of datetime axes would want independent scales either: for example, one might want to create a "Time since event" secondary axis, which is currently only doable by passing custom labels to a sec.axis with a time/date transformation. However, the current rescaling framework, as well as the way init values are set via the primary scale, throws errors when POSIX values used with an identity transformation. Short of committing to a complete retooling, I did not find a way to work with POSIX objects in the current framework that would not require some special handling, at least to reformat labels, and this seemed, at least for now, the cleanest way to approach it with the obvious downfall that it makes independent axis transformations impossible. I can't immediately think of a use case where a nonlinear transformation would be appropriate with date time scale - so I don't forsee tick misalignment issues with this current patch. Regardless,  please consider the current patch a place to start the discussion not necessarily an appropriate final solution. 

### Wrap up 
In preparing this PR, I've reviewed all open issues regarding sec.axis and added a host of visual and other tests to prevent new regressions. Note that I have commented out the tests for the monotonicity issue which at this point still cause a failing build - if we opt not to fix this issue these can be removed. Any ideas, thoughts, or advice from others, especially the core-developers, are appreciated. Please let me know if I can clarify anything.

@thomasp85, I would specifically appreciate your comments/review when you get a minute. I'm not 100% happy with where this is at but I figure getting this open gives us a place to start from. 

Fixes #2978.


thomasp85:
Can we somehow make the monotonicity check only work on the unexpanded range?

dpseidel:
So yes, we can call the unexpanded range from `scale`, which I have now done in a new function `mono_test` to reduce code redundancy in `break_info`. It seems to solve the monotonicity issue without any additional regressions in tests.

dpseidel:
So, as per discussion at tidy-dev-day @thomasp85 , I think we are done here. I've added some documentation about the limits to sec.axis with date time scales and significant tests. I considered adding a check in the `AxisSecondary$init()` function, for non-linear transformations passed to date or datetime scales but ultimately did not come up with any good options -- ideas welcome of course. The error messages are, e.g.  `Error in Ops.POSIXt(., 2) : '^' not defined for "POSIXt" objects`, at least informative and the documentation is now explicit about what transformations are allowed.

Zadchow:
Hi, I had a feature on a package where the second Y-axis would be 1 - the first Y-axis value. So if the Y axis on the left went up from 0 to 1 in increments of 0.05, the second axis did the same, but in the reverse direction, from 1 to 0 in the same incremements. Would this feature still be available after this update? It was available before 3.1, but then it seems that a bug in 3.1 seems to have prevented this transformation which I coded as `1 - .` in the sec_axis field. Thanks for all your work and for taking the time to read this! Love all of your works

dpseidel:
@Zadchow Yes, this feature is now supported again and I have added a test to ensure no future regressions. Thanks for your comment!

@thomasp85 could you review this again when you get a chance?

japhir:
This fixed [my issues with a sqrt transformation](https://stackoverflow.com/questions/55105952/sec-axis-in-ggplot2-doesnt-show-up-where-expected), thank you very much!

thomasp85:
Thanks Dana

japhir:
Wait, did this just get overwritten again?
After just updating, my problem with a secondary axis with complex transformation broke.
I deleted ggplot, re-installed with `devtools::install_github("tidyverse/ggplot2")`, and ran the code again: No luck.
I'm on ggplot2_3.1.1.9000.tar.gz
trying to add a secondary axis with the following transformation

```{r}
    ... +
    scale_x_continuous(expand = c(0, 0),
      sec.axis = sec_axis(~ sqrt((0.0449 * 1e+6)/(. - 0.167)) - 273.15,
                          "Sample temperature (°C)", temp_breaks, temp_ticks)) +
    ...
```
I thought the new tests would prevent this from being overwritten :(.

with the following `sessionInfo()`
```
R version 3.6.0 (2019-04-26)
Running under: Arch Linux

Matrix products: default
BLAS:   /usr/lib/libopenblasp-r0.3.6.so
LAPACK: /usr/lib/liblapack.so.3.8.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
 [1] stdsim_0.0.0.9000  furrr_0.1.0        future_1.12.0      testthat_2.1.1    
 [5] magrittr_1.5       purrr_0.3.2        tictoc_1.0         ggplot2_3.1.1.9000
 [9] tibble_2.1.1       tidyr_0.8.3.9000   dplyr_0.8.0.1      usethis_1.5.0     
[13] devtools_2.0.2    

 [1] nlme_3.1-139         crancache_0.0.0.9000 fs_1.3.1            
 [4] bit64_0.9-7          RColorBrewer_1.1-2   rprojroot_1.3-2     
 [7] tools_3.6.0          backports_1.1.4      R6_2.4.0            
[10] rpart_4.1-15         Hmisc_4.2-0          DBI_1.0.0           
[13] lazyeval_0.2.2       colorspace_1.4-1     nnet_7.3-12         
[16] withr_2.1.2          tidyselect_0.2.5     gridExtra_2.3       
[19] prettyunits_1.0.2    processx_3.3.1       bit_1.1-14          
[22] curl_3.3             compiler_3.6.0       cli_1.1.0           
[25] htmlTable_1.13.1     microbenchmark_1.4-6 desc_1.2.0          
[28] sandwich_2.5-1       labeling_0.3         scales_1.0.0        
[31] checkmate_1.9.3      mvtnorm_1.0-10       cranlike_1.0.2      
[34] callr_3.2.0          rappdirs_0.3.1       stringr_1.4.0       
[37] digest_0.6.18        foreign_0.8-71       base64enc_0.1-3     
[40] pkgconfig_2.0.2      htmltools_0.3.6      sessioninfo_1.1.1   
[43] htmlwidgets_1.3      rlang_0.3.4.9003     ggthemes_4.1.1      
[46] rstudioapi_0.10      RSQLite_2.1.1        zoo_1.8-5           
[49] acepack_1.4.1        Formula_1.2-3        Matrix_1.2-17       
[52] Rcpp_1.0.1           munsell_0.5.0        viridis_0.5.1       
[55] stringi_1.4.3        multcomp_1.4-10      debugme_1.1.0       
[58] MASS_7.3-51.4        pkgbuild_1.0.3       plyr_1.8.4          
[61] blob_1.1.1           grid_3.6.0           parallel_3.6.0      
[64] listenv_0.7.0        ggrepel_0.8.1        crayon_1.3.4        
[67] lattice_0.20-38      splines_3.6.0        parsedate_1.1.3     
[70] zeallot_0.1.0        knitr_1.22           ps_1.3.0            
[73] pillar_1.3.1         tcltk_3.6.0          reshape2_1.4.3      
[76] codetools_0.2-16     pkgload_1.0.2        glue_1.3.1          
[79] latticeExtra_0.6-28  data.table_1.12.2    remotes_2.0.4       
[82] vctrs_0.1.0.9003     gtable_0.3.0         assertthat_0.2.1    
[85] xfun_0.6             survival_2.44-1.1    viridisLite_0.3.0   
[88] memoise_1.1.0        cluster_2.0.9        investr_1.4.0       
[91] globals_0.12.4       TH.data_1.0-10       ellipsis_0.1.0.9000 
```

PR #3041, opened 2018-12-18
WIP: extension of stat-ellipse

ptoche:
Work In Progress.

Objective: extend ``stat_ellipse()`` to make "confidence ellipses" in addition to "data ellipses".

Followed/Borrowed code from the ``car`` library. First version. Runs, but does not appear to produce the correct values. Leaving it out there for any interested parties to motivate discussion. Hopefully someone will spot the errors.


clauswilke:
Could you please add a reprex for the new use case (confidence ellipse)? And could you please prepare it with the reprex package: https://reprex.tidyverse.org/articles/articles/magic-reprex.html

clauswilke:
A couple of comments:
1. The reprex should not refer to a specific library location on your computer. You can use `devtools::dev_mode()` to switch back and forth between your regular ggplot2 and the development version.

2. I don't see the value of providing `stat_ellipse()` with a formula. Wouldn't the only useful formula one could ever put there `y ~ x`, in which case you can just make it the default.

3. From looking at how the ellipse is drawn relative to the points, I suspect that the covariance matrix is transposed somewhere in the process, so that the variation in y is applied to x and vice versa.

ptoche:
Thanks Claus. 

  1. Do you mean this?

    library(reprex)
    devtools::install_github("ptoche/ggplot2")
    devtools::dev_mode()
    library(ggplot2)
    # code here

  2. This is so that you can do ``formula = log(y) ~ log(x)`` or ``formula = y ~ poly(x, 2)``. Following the way ``stat_smooth()`` works.  This ``formula = mpg ~ wt + gear`` wouldn't work though ("object 'wt' not found") ...

  3. Thanks for the hint. Will look into fixing that next.

ptoche:
The problem is that ``stat_ellipse()`` plots ``y` versus ``x``, while the ellipse is computed for ``x`` versus the ``intercept``. I must select the correct plane to display. 

I'm not clear yet on what ``mtcars`` does. If you supply "y ~ x", ``car::confidenceEllipse()`` plots the ``x`` versus the ``intercept``, but if you supply "y ~ x + z", it plots ``z`` versus ``x``. Not sure why. Presumably we should be able to supply a formula like ``y ~ x + z`` and select the plane like  ``c("x", "y")``. No? Does the ``car`` implementation make sense? 

``` r
library(reprex)
library(car)
#> Loading required package: carData
# y ~ x plots x versus Intercept
confidenceEllipse(model = lm(mpg ~ wt, data = mtcars))  
```



<sup>Created on 2018-12-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


``` r
library(reprex)
library(car)
#> Loading required package: carData
# y ~ x + z plots z versus x
confidenceEllipse(model = lm(mpg ~ wt + gear, data = mtcars)) 
```



<sup>Created on 2018-12-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

ptoche:
Claus, sorry to bother you, but I'd appreciate a few tips when you have a minute.

I'm very new to this. I'm trying to add arguments to ``stat_ellipse()``. I'm getting errors like ``unused argument (formula = formula)``. I wasn't getting these errors initially, so I'm not quite sure what I've done wrong. Below is a basic template of how I would add the argument ``formula`` to ``stat_ellipse``, now renamed to ``stat_ellipse_mode``.  I have commented every line I've modified. Can you tell right away what I'm doing wrong?

I'll keep trying, it's probably very stupid.

    # 1. renamed the function (next line)
    stat_ellipse_model <-  
    function(mapping = NULL, 
            data = NULL,
            geom = "path", 
            position = "identity",
            ...,
    # 2. added a formula argument with default (next line)
            formula = y ~ x,  
            na.rm = FALSE,
            show.legend = NA,
            inherit.aes = TRUE) {
    ggplot2::layer(
    # 3. renamed the stat in layer (next line)
        stat = StatEllipseModel,  
        data = data,
        mapping = mapping,
        geom = geom,
        position = position,
        show.legend = show.legend,
        inherit.aes = inherit.aes,
        params = list(
    # 4. added formula to params (next line)
        formula = formula, 
        na.rm = na.rm,
        ...
        )
    )
    }

    #' @rdname ggplot2-ggproto
    #' @format NULL
    #' @usage NULL
    #' @export
    # 5. renamed the ggproto (next two lines)
    StatEllipseModel <- 
    ggplot2::ggproto("StatEllipseModel", ggplot2::Stat,
    # 6. added formula to compute_group (next line)
        compute_group = function(data, scales, formula = y ~ x, na.rm = FALSE) {
        # calculations here 
        required_aes = c("x", "y"),
    }
    )

clauswilke:
I'm closing this PR. I think confidence ellipses would fit better into an extension package.

PR #3046, opened 2018-12-25
trivial doc fix

coolbutuseless:
Fix an unescaped % sign.


yutannihilation:
Looks good to me. Could you regenerate docs with `devtools::document()`?

PR #3047, opened 2018-12-27
Allow stat_bin() to compute over single-unique-value data

### LINKED ISSUE #3043: Error : Computation failed in `stat_bin()`: `binwidth` must be positive
This might be related to a previous issue: 
https://github.com/tidyverse/ggplot2/issues/2312

I am using the function below to create multiple histograms in one plot. It seems that features with zero variance / one value are not getting plotted.  It was working with previous version of ggplot2. Please help me in fixing this. 

plot_hist_facet <- function(data, bins = 10, ncol = 5, 
                            fct_reorder = FALSE, fct_rev = FALSE,
                            fill = palette_light()[[3]],
                            color = "white", scale = "free") {
  
  data_factored <- data %>%
    mutate_if(is.character, as.factor) %>%
    mutate_if(is.factor, as.numeric) %>%
    gather(key = key, value = value, factor_key = TRUE)
  
  if(fct_reorder) {
    data_factored <- data_factored %>%
      mutate(key = as.character(key) %>% as.factor())
  }
  
  if(fct_rev){
    data_factored <- data_factored %>%
      mutate(key = fct_rev(key))
  }
  
  g <- data_factored %>%
    ggplot(aes(x = value, group = key )) +
    geom_histogram(bins = bins, fill = fill , color = color,
                   ) +
    facet_wrap(~ key, ncol = ncol, scale = scale) +
    theme_tq()

  
  return(g)
  
}

Getting Computation failed in `stat_bin()`: `binwidth` must be positive  error 
snip20181221_20

yutannihilation:
Fixes #3043.

`stat_bin()` uses the width of the range of `x` to calculate the binwidth. So, it fails when `x` contains only one unique value, because the range is 0-width. Similaly to `resolution()`, we need to treat zero-ranges specially.

https://github.com/tidyverse/ggplot2/blob/7f13dfafccea19304ae140eb14a6cdfd5b5275ec/R/utilities-resolution.r#L20-L22

As suggested in https://github.com/tidyverse/ggplot2/pull/3047#issuecomment-450584147, the width of the range should be 0.1, for consistency with the width of the expansion `expand_default()` gives for the 0-width range.

``` r
library(ggplot2)

d <- data.frame(x = rep(1, 100))
ggplot(d, aes(x = x)) +
  geom_histogram()
```

<sup>Created on 2019-01-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


clauswilke:
The default range for scales without data range is only 0.1 in the latest version of ggplot2. Maybe use the same value for default bin width?

``` r
library(ggplot2)

ggplot(data.frame(), aes(x = 1, y = 0)) + geom_point()
```



<sup>Created on 2018-12-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
I realize that the expansion will be added once more for `stat_bin()` relative to `stat_identity()`. I think that's fine.

yutannihilation:
Now I'm convinced that the binwidth for 0-width range should be the same value as the width of the expansion for 0-width range, 0.1. I added the commits and updated the description.

yutannihilation:
@clauswilke Could you review this again?

PR #3050, opened 2018-12-27
Add lineend and linejoin params to geom_rect() and geom_tile()

### LINKED ISSUE #3037: geom_tile borders missing notch at top left corners
This is an issue that has been discussed over at Stackoverflow. The original problem can be found [here](https://stackoverflow.com/questions/53767758/geom-tile-border-missing-at-corners). A minimal reprex is included below.

I have the following data frame:

    # Dummy data frame
    df <- expand.grid(x = 1:3, y = 1:3)
    
I would like to plot it as a `geom_tile` using `ggplot2` like so:

    # Tile plot
    ggplot(df) + 
      geom_tile(aes(x = x, y = y), 
                fill = NA, colour = "red", size = 3, width = 0.7, height = 0.7)

which gives, 

rplot

Notice that in the top left corner of each tile the border has a notch missing and that the corner doesn't dovetail nicely like the other corners. The outcome I expected was a square around each tile, perhaps something like this:

rplot01

A solution offered [here](https://stackoverflow.com/a/53778706/1552004)  by Z. Lin in that Stackoverflow thread suggests that `linejoin` and `lineend` for `geom_tile` are switched to `mitre` and `square`, respectively. Or, that they are at least offered as parameters in the `geom` so that the user can choose for themselves. Hopefully, this demonstrates my issue adequately, but please let me know if I can clarify any aspect.

yutannihilation:
Fix #3037 

Just as #2132 for `geom_segment()`, this PR changes 

1. `geom_rect()`'s default of `linejoin` from `"butt"` to `"square"`, and 
2. makes it configurable via `linejoin` parameter by users.
3. `draw_key_polygon()` is changed as well.

In addition, as a workaround for Windows, this PR 

4. specifies the `lineend` parameter.

Note that I expect this breaks many visual tests as `geom_rect()` and `draw_key_polygon()` are used in many places.

``` r
library(ggplot2)


l <- function(linejoin) {
  geom_tile(aes(x = !!linejoin, y = 0), fill = NA, size = 5, width = 0.7, height = 0.7, linejoin = linejoin, colour = "black")
}

ggplot() +
  l("mitre") +
  l("round") +
  l("bevel") +
  ylim(-1, 1)
```

<sup>Created on 2019-05-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


ptoche:
I'm new to this, so forgive me if I'm talking nonsense, but do you need to update the legend key as well?

https://github.com/tidyverse/ggplot2/blob/1c09bae2aa24320bcb4891c664cccc16efc86a8a/R/legend-draw.r#L48

yutannihilation:
Thanks, you are right!

thomasp85:
@yutannihilation can you have a look at this again and see if it is ready for review? I think exposing a `lineend` argument is weird in a graphic element that is always closed. Can we handle it under the hood based on the `linejoin` argument?

yutannihilation:
> I think exposing a `lineend` argument is weird in a graphic element that is always closed.

Yeah, I fixed the implementation to use `linejoin` only (sorry for not updating the description of this PR). But, I'm stuck in confusion about these two points:

1. Why do some visual tests fail with the open form, which `polygonGrob()` expects? (c.f. https://github.com/tidyverse/ggplot2/issues/3037#issuecomment-458406857)
2. Should we specify `linend = "square"` fixedly as a workaround for Windows' bug (c.f. https://github.com/tidyverse/ggplot2/issues/3037#issuecomment-458385498)? As `geom_rect()` is one of the fundamental geoms, I expect this will break many existing visual tests, but I'm not sure if this is worth.

Let me think a bit if I can finish this...

thomasp85:
1. is probably as you mention that closed form is needed for correct munching - I didn't anticipate this when making the initial suggestion

2. I think lineend should simply follow the linejoin parameter and not be user-accessible... `linejoin = 'round'` -> `lineend = 'round'`, everything else `lineend = 'square'`... Further, this is a graphic device problem, and I think there a limits to the lengths we need to go to to fix it in ggplot2

yutannihilation:
> is probably as you mention that closed form is needed for correct munching - I didn't anticipate this when making the initial suggestion

Thanks, then I'll revert it to closed form.

For the second one, I meant, changing the default lineend (`butt`) will break visual tests, which means we might have many broken revdeps. As you say, this is a graphic device problem. Do you think it's worth? (I agree it's only `linejoin` that we should expose to the users)

thomasp85:
Are the visual tests not only run on linux systems, where the `lineend` parameter is rightfully ignored?

In any way I think it is worth breaking the tests... they are mainly there to avoid unforeseen regressions

yutannihilation:
> where the `lineend` parameter is rightfully ignored?

Oh, I didn't know this. Thanks.

Then, I'll try to finish this PR!

thomasp85:
hmm... can I get you to look into the test failures? Apparently the tests are run on AppVeyor after all. It seems you missed updating the one in test-scales-breaks-labels.R

yutannihilation:
Yes, I will. Sorry that I can't make much time for this...

thomasp85:
No worries - didn't know you were hung up... Is there anything I can help with?

yutannihilation:
> It seems you missed updating the one in test-scales-breaks-labels.R

Thanks, this comment really helped me! I didn't merge the current master so the new visual tests added in #2334 were not included here.

Currently I'm trying to figure out what the `draw_legend()` should be. I didn't notice `geom_rect()` doesn't use `draw_key_rect()`...

yutannihilation:
@thomasp85 
I'm really sorry that I couldn't make this in due. Should this be still included in v3.2.0? I think this is ready for review now.

thomasp85:
This is fine - well add it🙂. Reviewing later today

yutannihilation:
Sure. I'll do my best to be responsive this week to let this be merged early.

yutannihilation:
Thanks!

PR #3054, opened 2019-01-03
Use CRAN vdiffr

yutannihilation:
Since vdiffr 0.3.0 is released, let's

* ~~switch vdiffr from the GitHub version to the CRAN version and~~ (Travis already uses CRAN version of vdiffr since this commit: https://github.com/tidyverse/ggplot2/commit/9f1904df332096f9336c663c38980aa718c24a47)
* rely on `VDIFFR_RUN_TESTS` envvar, which vdiffr checks directly, instead of `USE_VDIFFR` envvar

~~Note that it seems still better to check the version of vdiffr since we ignore the installation failure of suggested packages by `_R_CHECK_FORCE_SUGGESTS_=false`.~~

### DIFF
--- a/tests/testthat/helper-vdiffr.R
+++ b/tests/testthat/helper-vdiffr.R
@@ -1,33 +1,11 @@
-# default is equal to whether NOT_CRAN is true or not
-enable_vdiffr <- identical(Sys.getenv("NOT_CRAN"), "true")
-
-# disable or enable vdiffr based on the state of USE_VDIFFR, if set
-if (identical(Sys.getenv("USE_VDIFFR"), "true")) {
-  enable_vdiffr <- TRUE
-} else if (identical(Sys.getenv("USE_VDIFFR"), "false")) {
-  enable_vdiffr <- FALSE
-}
+# vdiffr ignores failures when
+#   - VDIFFR_RUN_TESTS is "false" (on Travis CI with older versions and dev version of R)
+#   - CI is not set (on CRAN)
 
 # disable vdiffr if version is old
 if (!requireNamespace("vdiffr", quietly = TRUE) ||
   utils::packageVersion("vdiffr") < "0.2.3.9001") {
-  enable_vdiffr <- FALSE
+  Sys.setenv(VDIFFR_RUN_TESTS = "false")
 }
 
-expect_doppelganger <- function(title, fig,
-                                path = NULL,
-                                ...,
-                                user_fonts = NULL,
-                                verbose = FALSE) {
-  if (!enable_vdiffr) {
-    expect_error(regexp = NA, ggplot_build(fig))
-    return(invisible(NULL))
-  }
-
-  vdiffr::expect_doppelganger(title, fig,
-    path = path,
-    ...,
-    user_fonts = user_fonts,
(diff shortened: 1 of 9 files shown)


lionel-:
It is not necessary to set `NOT_CRAN` because vdiffr also checks for `CI`, which is set on both Travis and Appveyor. Also the wrapper in helper-vdiffr might be simplified by setting `VDIFFR_RUN_TESTS` to override the default behaviour on Travis.

yutannihilation:
Thanks, I need to learn about vdiffr more...

> It is not necessary to set `NOT_CRAN` because vdiffr also checks for `CI`, which is set on both Travis and Appveyor.

I set `NOT_CRAN` because the default value of `enable_vdiffr` depends on it. But, it sounds like it's better to remove this line and let vdiffr to skip tests on CRAN.

https://github.com/tidyverse/ggplot2/blob/2b00ea79843053c9984b341adfc6c218fe0445b2/tests/testthat/helper-vdiffr.R#L1-L2

yutannihilation:
Confirmed the visual changes are detected properly with the new settings.

https://travis-ci.org/tidyverse/ggplot2/builds/475155850
https://ci.appveyor.com/project/hadley/ggplot2/builds/21373584

lionel-:
LGTM! But let's wait until Claus has time to take another look, since he worked on the Travis settings last time.

tests/testthat/helper-vdiffr.R
```diff
@@ -1,33 +1,11 @@
...
-} else if (identical(Sys.getenv("USE_VDIFFR"), "false")) {
-  enable_vdiffr <- FALSE
-}
+# vdiffr ignores failures when
+#   - VDIFFR_RUN_TESTS is "false" (on Travis CI with older versions and dev version of R)
+#   - CI is not set (on CRAN)
 
 # disable vdiffr if version is old
 if (!requireNamespace("vdiffr", quietly = TRUE) ||
   utils::packageVersion("vdiffr") < "0.2.3.9001") {
```
clauswilke: Should this be `< "0.3.0"`?

clauswilke:
Looks good to me to.

@lionel- When vdiffr is disabled, does `expect_doppelganger()` still execute the plotting code, but just doesn't compare svgs, or does it not even build the plot? I think it would be good to always build the plot, even if svg comparison is switched off.

lionel-:
Yup that's how it works. The SVGs are always generated no matter what, so we still get hard failures when the plot doesn't build properly, and we can still use `expect_warning()` when we expect plot building to warn.

tests/testthat/helper-vdiffr.R
lionel-: I think this `if` can be removed entirely.

yutannihilation:
I removed the `if` statement, which is not really necessary for now; I added the `if` for the case when we would come to want to enable vdiffr for older versions of R, where the installation failures of suggested package are ignored (`_R_CHECK_FORCE_SUGGESTS_=false`) so that the older version of vdiffr might be used. But, this seems unlikely to occur.

yutannihilation:
Thanks for reviewing!

PR #3069, opened 2019-01-15
Fix coord_map to allow switching axis position

### LINKED ISSUE #3042: coord_map(): moving y axis to the right does not move all axis elements
Hi,
this is almost identical to issue #2005 (fixed by #2062) but for `coord_map()`, instead of `coord_polar()`. When the y-axis is moved to the right by `scale_y_continuous(position = "right")`, only the axis label is moved. I adapted the reprex from that previous issue:

```r
library(ggplot2)

df <- data.frame(
  heights = c(100, 110, 120, 130, 140, 150), 
  weights = c(50.1, 55.1, 60.1, 65.1, 70.1, 75.1)
)

plot <- ggplot(df, ggplot2::aes(x = heights, y = weights, group = 1)) + 
  geom_line(color = "purple") + 
  coord_map()

# OK
plot 

# NOT OK: Label moves, but rest does not
plot + scale_y_continuous(position = "right")
```

okplot
notokplot

Cheers!

karawoo:
Fixes #3042 by adding `x.arrange` and `y.arrange` as panel params that reflect the primary/secondary axes. The order of the `axes` list elements needed to be swapped because `scale_x$axis_order()` gives the top first, then the bottom.


R/coord-map.r
```diff
@@ -259,8 +260,8 @@ CoordMap <- ggproto("CoordMap", Coord,
     pos <- self$transform(x_intercept, panel_params)
 
     axes <- list(
-      bottom = guide_axis(pos$x, panel_params$x.labels, "bottom", theme),
-      top = guide_axis(pos$x, panel_params$x.labels, "top", theme)
+      top = guide_axis(pos$x, panel_params$x.labels, "top", theme),
+      bottom = guide_axis(pos$x, panel_params$x.labels, "bottom", theme)
```
clauswilke: Just wondering: Is there anything happening here other than changing the order of the two lines? Is there a reason to switch the order?

karawoo: Yes, `scale_x$axis_order()` returns which axis is primary and which is secondary in this order (e.g. `"secondary" "primary"` meaning the top axis is secondary and the bottom is primary), so for the subsetting in the next line to work, this list needs to follow the same order (which is also why I had to reorder the values in `arrange <- panel_params$x.arrange %||% c("secondary", "primary")`)

PR #3081, opened 2019-01-19
error on more than 1e6 bins

thomasp85:
As discussed with @hadley - `stat_bin()` can end up creating a huge number of bins depending on data and settings and end up freezing the session. This PR errors when the number of bins exceed 1e6 which should be more than enough for any reasonable use

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,9 @@
 * `coord_sf()`, `coord_map()`, and `coord_polar()` now squash `-Inf` and `Inf`
   into the min and max of the plot (@yutannihilation, #2972).
 
+* `stat_bin()` will now error when the number of bins exceed 1e6 to avoid 
+  accidentally freezing the user session
+
 # ggplot2 3.1.0
 
 ## Breaking changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,7 +28,7 @@
   into the min and max of the plot (@yutannihilation, #2972).
 
 * `stat_bin()` will now error when the number of bins exceed 1e6 to avoid 
-  accidentally freezing the user session
+  accidentally freezing the user session (@thomasp85).
 
 # ggplot2 3.1.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,7 +27,7 @@
 * `coord_sf()`, `coord_map()`, and `coord_polar()` now squash `-Inf` and `Inf`
   into the min and max of the plot (@yutannihilation, #2972).
 
-* `stat_bin()` will now error when the number of bins exceed 1e6 to avoid 
+* `stat_bin()` will now error when the number of bins exceeds 1e6 to avoid 
   accidentally freezing the user session (@thomasp85).
 
 # ggplot2 3.1.0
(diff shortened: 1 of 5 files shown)


R/bin.R
```diff
@@ -87,6 +87,10 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
   max_x <- x_range[2] + (1 - 1e-08) * width
   breaks <- seq(origin, max_x, width)
 
+  if (length(breaks > 1e6)) {
```
yutannihilation: should be:

``` r
   if (length(breaks) > 1e6) { 
```

NEWS.md
```diff
@@ -27,6 +27,9 @@
 * `coord_sf()`, `coord_map()`, and `coord_polar()` now squash `-Inf` and `Inf`
   into the min and max of the plot (@yutannihilation, #2972).
 
+* `stat_bin()` will now error when the number of bins exceed 1e6 to avoid
```
clauswilke: "number of bins *exceeds*"

R/bin.R
```diff
@@ -87,6 +87,10 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
   max_x <- x_range[2] + (1 - 1e-08) * width
   breaks <- seq(origin, max_x, width)
 
+  if (length(breaks) > 1e6) {
+    stop("Number of bins exceed the maximum of 1e6", call. = FALSE)
```
clauswilke: As in the news item: "Number of bins *exceeds*"

R/bin.R
```diff
@@ -87,6 +87,10 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
   max_x <- x_range[2] + (1 - 1e-08) * width
   breaks <- seq(origin, max_x, width)
 
+  if (length(breaks) > 1e6) {
+    stop("Number of bins exceeds the maximum of 1e6", call. = FALSE)
```
hadley: Can you write in the style of http://style.tidyverse.org/error-messages.html? I think a hint of the form "Did you make `binwidth` too small?` would be useful.

thomasp85: Ah - sure

R/bin.R
```diff
@@ -88,7 +88,7 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
   breaks <- seq(origin, max_x, width)
 
   if (length(breaks) > 1e6) {
-    stop("Number of bins exceeds the maximum of 1e6", call. = FALSE)
+    stop("The number of bins exceeds the maximum of 1e6. Did you make binwidth too small?", call. = FALSE)
```
hadley: I think this should be:

```
The number of histogram bins must be less than 1,000,000.
Did you make `binwidth` too small?
```

tests/testthat/test-stat-bin.R
```diff
@@ -124,6 +124,17 @@ test_that("weights are added", {
...
 
+test_that("bin errors at high bin counts", {
+  df <- data_frame(
+    x = runif(1e5, max = 2e6)
+  )
+  p <- ggplot(df) + geom_histogram(aes(x = x), binwidth = 1)
+
+  expect_warning(
+    ggplot_build(p),
+    "The number of bins exceeds the maximum of 1e6. Did you make binwidth too small?")
```
hadley: I think it would be better to test this by calling `bin_breaks_width` directly, and you should only include enough text to ensure that you're catching the right warning.

thomasp85: gotcha

Issue #2673, opened 2018-05-30
Provide a guide that has discrete colors, but labels shifted between them to highlight intervals


adrfantini:
This is a feature request.
`ggplot2` does not provide a way to use discrete color scales in which the labels are staggered compared to the colors. In my opinion, this kind of color bars is more intuitive than color bars where each color is assigned an explicit interval (usually in a form like `[2,5)`).

Using only base `ggplot`, this is the best that I could come up with ([SO question link](https://stackoverflow.com/questions/26802272/generating-a-color-legend-with-shifted-labels-using-ggplot2)), but it's very hackish, involving shifting the labels and using a dummy empty label at the end:

![](https://i.stack.imgur.com/7g4mf.png)

While this works, it is a bit hackish, the labels often are a bit shifted depending on plot size, and the white spaces between the color boxes do not look very good.

Most other plotting systems allow for this kind of color bar: here are some examples of the kind of color bar that I'd like in `ggplot`, implemented in other plotting systems:

- Lattice ([source](https://stackoverflow.com/questions/21641186/r-levelplot-colorkey-label-and-color-range)): 
![](https://i.stack.imgur.com/3j2lZ.png)

- Base graphics ([source](https://stackoverflow.com/questions/49490246/plot-continuous-data-with-discrete-colors)):
![](https://i.stack.imgur.com/oEteR.png)

- Matplotlib ([source](https://stackoverflow.com/questions/14777066/matplotlib-discrete-colorbar)):
![](https://i.stack.imgur.com/6OvdM.png)

- MATLAB ([source](https://www.mathworks.com/matlabcentral/answers/203750-define-a-sequential-discrete-color-map-with-first-color-white)):
![](https://www.mathworks.com/matlabcentral/answers/uploaded_files/29144/fig.bmp)

I have written a small function ([SO link with examples](https://stackoverflow.com/a/50540633/4124334)) to be able to plot this kind of color bar as a `ggplot` object, to be added to other objects using `patchwork` or similar. It's not a great solution however: a built-in guide would be much much better.
Example result:
![](https://i.stack.imgur.com/RXBhr.png)

Note that in this case the last "bin" in the color bar indicates that the last color represents values going up to infinity; this is a useful feature in many cases.

Could something like this be implemented as a built-in guide?

thomasp85:
Can you clarify whether merely the guide should be discrete while the scale itself is continuous such as in your lattice example, or if it is intended for discrete scales?

[...]

PR #3096, opened 2019-01-22
scale_binned

thomasp85:
This PR is created in advance to allow for discussion of a potential new fundamental scale (that is, something at the same level as continuous and discrete). The purpose of scale_binned is to discretise continuous data for use with discrete palettes (e.g. shape or brewer colours), or with geoms expecting discrete positions (e.g. `geom_bar()`)...

The main advantage of this setup is that it allows the guides to automatically place ticks at the border between bins, rather than at the middle of each bin - something that has been requested before (e.g. #2673)

Currently the scale work for positions, but the main holdback for the other aesthetics is the creation of a new type of legend that supports the desired output.

```r
ggplot(mtcars) + 
  geom_bar(aes(mpg)) + 
  scale_x_binned()
```
image

as can be seen, the ticks are between the bars which provide a much more intuitive understanding of the range of each bin compared to using `geom_histogram()` where the scale remain continuous and thus has tick marks at "nice" locations from a numerical point of view... For colour scales as desired in #2673 this effect is also of great use.

### Open questions
- what should the naming be for the shorthand scales for colour? `scale_colour_brewer()` etc are already taken (for good reasons) — is there a pattern that can be applied to all? The most graceful thing would be to have discrete scales that are used on continuous data be binned automatically, but I don't think this is doable (and the magic might be limiting in other ways)
- should the outer ticks be shown (e.g. the limits of the data)?
- can we support non-linear binning gracefully (e.g. have a bin twice as wide as the others and have the bar expand to that size automatically)
- Currently the behaviour is to default to 10 bins - is this a reasonable choice or is there anything better?
- whatever questions and suggestions you have @Hadley, @clauswilke, @karawoo, @yutannihilation


yutannihilation:
> what should the naming be for the shorthand scales for colour?

Do you think we can implement this as such options of `scale_colour_continuous()` as `bins` and `binwidth`, so that the developers of color scale packages don't need to implement their own binned version of the scale by themselves? (The shorthand names are still useful, though).

clauswilke:
@yutannihilation I think other parameters would have to be changed, such as the default guide, so having a separate scale function may be needed after all.

If we go with the naming for the viridis scale, `scale_color_viridis_c()` for continuous and `scale_color_viridis_d()` for discrete would suggest `scale_color_viridis_b()` for binned. 

Also, is it time to overhaul the guide system and port it to ggproto? It's the only part of ggplot2 that hasn't been ported over, and the code is difficult to maintain and extend. Lots of code duplication between `guide_legend()` and `guide_colorbar()`.

yutannihilation:
> I think other parameters would have to be changed, such as the default guide, so having a separate scale function may be needed after all.

Hmm..., agreed. Thanks.

thomasp85:
I've been eying the guide internals for years... I might rewrite them as part of making the new bin guide...

ptoche:
> Currently the behaviour is to default to 10 bins - is this a reasonable choice or is there anything better?

A popular rule of thumb is Sturges's rule, i.e. round base-2 log of sample size plus 1. Other rules are listed [here](https://en.wikipedia.org/wiki/Histogram#Number_of_bins_and_width). The Freedman-Diaconis rule may be worth looking into. But for simplicity, Sturges's rule or 10 bins can hardly be beat.

hadley:
@ptoche I don't think Sturge's rule is not appropriate because it is motivated by minimising some mathematical quantity that is not directly related to needs of a visualisation. I'd recommend using 30 bins to match `geom_histogram()`.

thomasp85:
Hmm... I think 30 may be too much as a default as it would clutter the axes with labels

thomasp85:
Case in point

image

hadley:
Hmmmm, in that case maybe it's better to behave like the default labels for continuous legends? i.e. pick `n` (5? 10?) nicely spaced numbers

thomasp85:
I'll look into that... that would give some sort of consistency

thomasp85:
hmm... it's a shame that the current trans implementation doesn't allow for modifying the number of breaks after creation

a `n_breaks` argument would make sense for both binned and continuous scales...

thomasp85:
We could set `n` in the `breaks()` method environment from ggplot2 in order to allow this, but it feels kinda hacky...

clauswilke:
Just a minor comment: I think it makes sense to have a default bin number of 30 for position scales, but you'd want a default of on the order of five to seven for color scales. And odd numbers have the advantage that they contain the midpoint for diverging scales. Ideally we'd have different defaults for position and color.

clauswilke:
Also, the scale could simply produce empty labels by default if it judges that there are too many bin boundaries. E.g., if there are 30 bin boundaries, leave 20 of the label strings empty.

thomasp85:
I though about this as well... essentially making some of the breaks into minor breaks...

problem is if the user chose to have breaks of uneven length and the scale then choose to remove some of the labels - there is no way of deducing where the cut is made in that case...

thomasp85:
Using the same break algorithm as continuous scales:

```r
ggplot(df) + geom_bar(aes(x)) + scale_x_binned()
```
image

Support for an `n_breaks` argument is build in:

```r
ggplot(df) + geom_bar(aes(x)) + scale_x_binned(n_breaks = 10)
```
image

thomasp85:
I think the colour scales should have a default n_breaks matching the number of discrete colours in the palette or something

thomasp85:
Does it make sense to have a binned guide for anything but colour/fill (same constraints as colourbar)?

On one hand binning allows continuous data to be used with aesthetics that only have discrete representation (e.g. shape), but on the other hand I'm not sure when it will ever be a good idea to use something without an inherent order to symbolise different bins.

For something like size I cannot figure out why you would want it binned in the first place instead of a direct mapping...

hadley:
I think it makes sense for size for the same reason it does for colour: it’s almost certainly easier to perceive differences with a small number of categories.

thomasp85:
that is a fair point...

thomasp85:
Here's the Vega implementation of such a scale for reference: 

https://vega.github.io/vega/docs/scales/#quantize

BangerNPoo:
Fuck u

thomasp85:
It just occurred to me that `guide_colourbar()` can be retrofitted for use as a binned guide (still need a proper one as well though):

```r
ggplot(mtcars) + 
  geom_point(aes(disp, mpg, colour = hp)) + 
  scale_colour_blender(guide = guide_colorbar(
    raster = FALSE, 
    nbin = 500, 
    barheight = 25, 
    barwidth = .5, 
    ticks = FALSE
  ))
```
image

thomasp85:
I think I'm pretty much done with the main implementation of the new scale by now. For the time being there are only position and colour/fill versions of the scale as I still haven't dealt with a generalised binned guide (colour/fill use a slightly modified version of `guide_colorbar()` called `guide_colorsteps()`)

Some things you may ponder:
- If `breaks = waiver()` and `limits = NULL` the limits will be recalculated so the outer bins have the same width as the other bins. If you want the limits to reflect the data range you can set it to `c(NA, NA)`. This behaviour is only if `breaks = waiver()` as this is the only case where we are certain the bins are of equal width
- `guide_colorsteps()` will default to show bins of equal height, even if the breakpoints are not evenly spaced — is that a good default
- The default is to not show the limits of the scale thus implying that the outer bins goes to +/- infinity (which is true as the oob behaviour is to squish values). This is in line with how Vega construct their quantize scales. Is this the default behaviour we want (can be controlled with a `show_limits` arg in the scale/guide)
- As mentioned above, the default oob behaviour is `squish()`. This is in line with the outer bins going to infinity and makes sense IMO - do you agree?

Can I get you all to play with the current implementation a bit to the extend your time permits and report back any surprises you might find. Other things would be whether the naming I've settled on could be improved...

clauswilke:
Should there be a binned version of the viridis scales? People would probably want to use those.

thomasp85:
Sure, but I think that’s up to the different palette packages to provide..?

clauswilke:
We currently provide `scale_*_viridis_c()` for continuous values and `scale_*_virids_d()` for discrete values:

``` r
library(ggplot2)

ggplot(mtcars, aes(disp, mpg, color = hp)) +
  geom_point() + scale_color_viridis_c()
```



``` r

ggplot(mtcars, aes(disp, mpg, color = factor(cyl))) +
  geom_point() + scale_color_viridis_d()
```



<sup>Created on 2019-02-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

There's also some option one can set to make `scale_*_viridis_c()` the default continuous color scale of ggplot2, but I don't quite remember right now how.

It seems to me that for consistency we should add `scale_*_viridis_b()` for binned data.

thomasp85:
Right - I remembered wrongly that we just imported those functions from the viridis package

adrfantini:
This scale bins the (continuous) input data, right? What if the input data is already binned? There are some usecases when this might be handy: for example, as far as I can tell, `ggplot` is much faster at plotting a large `sf` MULTILINESTRINGS object compared to a large collection of small `LINESTRINGS`, so I usually first `cut %>% group_by %>% summarise` them. This means the data is already in binned form.

thomasp85:
In next ggplot2 version there will be no performance difference (or only slight) between plotting a single multilinestring or several separate

thomasp85:
If the data is already binned there is no sense in using a binning scale on them, right? I'm not sure I understand (in general) what you'd expect to happen..?

adrfantini:
> In next ggplot2 version there will be no performance difference (or only slight) between plotting a single multilinestring or several separate

Awesome, is this already in the current github master branch?

> If the data is already binned there is no sense in using a binning scale on them, right?

You would not need binning _per se_, but you would still need a proper discrete scales with labels that are placed between bins, and not at their center. I'm not sure if I am clear enough, maybe looking at the [SO question linked](https://stackoverflow.com/questions/26802272/generating-a-color-legend-with-shifted-labels-using-ggplot2) in #2673 would clarify things.

>  I'm not sure I understand (in general) what you'd expect to happen..?

Well, I'd like to have the benefit of the binned scale (proper alignment, looks nice, etc.) even with previously binned data. There are some usecases where you already have binned data, and un-binning it just to be able to use the function (which bins it again) seems silly.

Let me know if I should maybe explain it a bit better.

thomasp85:
The improved sf support is already in master...

What you want is a binned guide, not a binned scale... I'll try to make sure the new guides will work with the output from discrete scales as well (maybe they already do, I can't remember), which should solve your request...

adrfantini:
Thanks, much appreciated!

thomasp85:
Picking this up again. I've now successfully made a `guide_bins()` which is a binned version of `guide_legend()` in the same way as `guide_colorsteps()` is a binned version of `guide_colorbar()`. This means that binned scales are ready for all the different scale types. Example of `guide_bins()`

```r
ggplot(mtcars) +
  geom_point(aes(disp, mpg, colour = hp)) +
  scale_colour_blender(guide = guide_bins()) + 
  theme_minimal()
```
image

Currently the tick style is equivalent to that of `guide_colorbar()` and I'm frankly conflicted about the look. It shouldn't be a full line IMO as that would be too heavy. It should also not be absent as that would leave the labels dangling without any fixture. But something about this is not right either...

adrfantini:
In my original issue (#2673) you can see style examples different languages. The vast majority of the times I've seen this kind of scales is with filled squares/rectangles, with ticks in between and no spacing.
Like this:


In the field I know best (climate) this is the usual look. Maybe I am biased, but I like it more than other styles.

thomasp85:
But this is more or less what you get with `guide_colorsteps()`... I'm thinking about styling where the legend needs to be represented by a shape or size

```r
ggplot(mtcars) +
  geom_point(aes(disp, mpg, colour = hp)) +
  scale_colour_blender(guide = guide_colorsteps()) + 
  theme_minimal()
```
image

thomasp85:
yes... but we need to decide on a look for non-colour binned scales... something Vega has ignored

adrfantini:
> But this is more or less what you get with guide_colorsteps()... I'm thinking about styling where the legend needs to be represented by a shape or size

Whoopsie, missed that, sorry. Ok then, I understand now. Your approach looks good to me for this usecase (which in al fairness seems rare to me). Maybe the styling could have a filled horizontal tick line extending a bit more to the right and linking with the number, whereas you now have two separate ticks?

Maybe it looks a bit weird only due to the sizes of the dots, with larger dots it would probably look fine.

smouksassi:
can we imagine like a mini axis showing the bin then we put whatever we want on the side of it ?
scale_bin_legend

thomasp85:
I could maybe get behind something like this

image

I don't think it is well-balanced if the ticks extend towards the labels

adrfantini:
It looks good. Is there any possibility to add an arrow at the extremes of the scale, if the bins are open in that direction?

thomasp85:
```r
guide_bins(axis.arrow = arrow(length = unit(1.5, 'mm'), ends = 'both'))
```
image
```r
guide_bins(show.limits = TRUE)
```
image

adrfantini:
Looks very good. Ideally if the user specifies manual breakpoints the arrows could be automatic. For the lower bound for example:

- no arrow if the break is <= than the smallest value
- arrow otherwise

thomasp85:
I don't think I want automatic arrows TBH

thomasp85:
`arrow()` is a grid function so I can’t really change that...

clauswilke:
You could also draw a column/row of rectangles, as in @adrfantini's example, and then instead of filling the rectangles with color you place the discrete symbols inside each rectangle. 

Something like this.
``` r
library(ggplot2)

df <-data.frame(x = 1:5)

ggplot(df, aes(x = x, y = 1)) +
  geom_tile(color = "black", fill = NA, size = 1.5) +
  geom_point(aes(color = x), size = 5) +
  guides(color = "none") +
  ylim(-2, 3) +
  theme_void()
```



<sup>Created on 2019-09-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
I don’t really like that look though — too heavy and a strong depart from the look of `guide_legend()`

R/guide-colorbar.r
```diff
@@ -567,6 +567,57 @@ guide_gengrob.colorbar <- function(guide, theme) {
 #' @rdname guide_colourbar
 guide_colorbar <- guide_colourbar
 
+#' @export
+#' @rdname guide_colourbar
+#'
+#' @param even.steps Should the bin sizes reflect their size or be even across all
```
hadley: I don't quite follow what this means, or why you'd want it to be false.

R/guide-colorbar.r
```diff
@@ -567,6 +567,57 @@ guide_gengrob.colorbar <- function(guide, theme) {
...
+#'   bins? Defaults to `TRUE`
+#' @param show.limits Should labels for the outer limits of the bins be printed?
+#'   Default is `NULL` which makes the guide use the setting from the scale
+guide_coloursteps <- function(even.steps = TRUE, show.limits = NULL, ticks = FALSE, nbin = 100, ...) {
+  guide <- guide_colourbar(raster = FALSE, ticks = ticks, nbin = nbin, ...)
```
hadley: We really should make this a proper constructor that permits subclassing at some point.

R/scale-.r
```diff
@@ -473,7 +473,13 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
   transform = function(self, x) {
      new_x <- self$trans$transform(x)
      if (any(is.finite(x) != is.finite(new_x))) {
-       type <- if (self$scale_name == "position_c") "continuous" else "discrete"
+       type <- if (self$scale_name == "position_c") {
+         "continuous"
+       } else if (self$scale_name == "position_d") {
+         "discrete"
+       } else {
+         "binned"
```
hadley: It feels a little weird for "binned" to be the default. Would it be better to throw an error if the `scale_name` was an unknown value?

R/scale-.r
```diff
@@ -810,6 +816,399 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+#' @format NULL
+#' @usage NULL
+#' @export
+ScaleBinned <- ggproto("ScaleBinned", Scale,
+  range = continuous_range(),
+  na.value = NA_real_,
+  oob = squish,
+  n.breaks = NULL,
+  right = TRUE,
+  after_stat = FALSE,
```
hadley: `after.stat`?

R/scale-.r
```diff
@@ -810,6 +816,399 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+  show.limits = FALSE,
+
+  is_discrete = function() FALSE,
+
+  train = function(self, x) {
+    if (!is.numeric(x)) {
+      stop("Binned scales only support continuous data", call. = FALSE)
+    }
+
+    if (length(x) == 0) return()
```
hadley:
```suggestion
    if (length(x) == 0) {
      return()
    }
```

R/scale-.r
```diff
@@ -810,6 +816,399 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+    self$range$train(x)
+  },
+
+  transform = function(self, x) {
+    new_x <- self$trans$transform(x)
+    if (any(is.finite(x) != is.finite(new_x))) {
+      type <- if (self$scale_name == "position_c") {
```
hadley: This probably should be pulled out into a function

R/scale-.r
```diff
@@ -810,6 +816,399 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+    new_x
+  },
+
+  map = function(self, x, limits = self$get_limits()) {
+    if (self$after_stat) {
+      x
+    } else {
+      breaks <- self$get_breaks(limits)
+
+      x_binned <- cut(x, c(limits[1], breaks, limits[2]), labels = FALSE,
```
hadley: Style

R/scale-.r
```diff
@@ -810,6 +816,399 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+
+    limits <- self$trans$inverse(limits)
+
+    if (is.null(self$breaks)) {
+      return(NULL)
+    } else if (identical(self$breaks, NA)) {
+      stop("Invalid breaks specification. Use NULL, not NA", call. = FALSE)
+    } else if (is.waive(self$breaks)) {
+      if (!is.null(self$n.breaks)) {
+        old_n <- get("n", environment(self$trans$breaks))
```
hadley: This is moderately horrendous. Why do we need to write into this environment?

R/scale-.r
```diff
@@ -810,6 +816,399 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+    # labels
+    labels <- self$get_labels(major)
+
+    list(range = range, labels = labels,
+         major = pal, minor = NULL,
+         major_source = major, minor_source = NULL)
+  }
+)
+
+#' Continuous scale constructor.
```
hadley: Why is this included in the diff?

R/scale-binned.R
```diff
@@ -0,0 +1,115 @@
...
+    self$range$train(x)
+  },
+
+  map = function(self, x, limits = self$get_limits()) {
+    breaks <- self$get_breaks(limits)
+    all_breaks <- unique(sort(c(limits[1], breaks, limits[2])))
+
+    if (self$after_stat) {
+      # Backtransform to original scale
+      x_binned <- cut(x, seq_len(length(all_breaks) + 1) - 0.5, labels = FALSE,
```
hadley: Style

hadley: And a few other places below

R/scale-brewer.r
```diff
@@ -104,6 +104,27 @@ scale_fill_distiller <- function(..., type = "seq", palette = 1, direction = -1,
     gradient_n_pal(brewer_pal(type, palette, direction)(7), values, space), na.value = na.value, guide = guide, ...)
 }
 
+#' @export
+#' @rdname scale_brewer
+scale_colour_blender <- function(..., type = "seq", palette = 1, direction = -1, na.value = "grey50", guide = "coloursteps", aesthetics = "colour") {
```
hadley: Why blender?

hadley: If we have a brewer (discrete, beer) and a distiller (continuous, wine), it feels like the obvious name would be vintner.

R/scale-gradient.r
```diff
@@ -134,3 +134,24 @@ scale_fill_gradientn <- function(..., colours, values = NULL, space = "Lab", na.
   continuous_scale(aesthetics, "gradientn",
     gradient_n_pal(colours, values, space), na.value = na.value, guide = guide, ...)
 }
+#' @rdname scale_gradient
+#' @export
+scale_colour_steps <- function(..., low = "#132B43", high = "#56B1F7", space = "Lab",
```
hadley: Why isn't this scale colour binned?

R/guide-colorbar.r
thomasp85: If you set the breaks to be uneven intervals, e.g. [5, 10, 20, 40] should this be reflected in the relative size of each bin in the colour ramp

R/guide-colorbar.r
thomasp85: agreed. overhaul is for v3.5

R/scale-.r
thomasp85: thanks

R/scale-.r
thomasp85: That is putting it mildly 🙂 It is atrocious and it won't be part of the merge. I put it in as a hack to allow setting number of breaks in the transformation object, but this has been fixed in scales (also why we need a scale release soon)

My plan is to inspect the formals of the `breaks` method and, if `n` is present, use it. In this way we don't need to rely on updated transformation objects

R/scale-.r
thomasp85: 🤷‍♂️

R/scale-brewer.r
thomasp85: blender for mixed drinks 🙂 I'm open to another naming honestly. I just wanted to continue this weird tradition

R/scale-gradient.r
thomasp85: It follows the style of `scale_colour_gradient` that gets defined and then is assigned to `scale_colour_continuous`

R/scale-brewer.r
clauswilke: But "brewer" is the last name of Cynthia Brewer who developed the color palettes. So maybe `scale_color_pedersen()`? On a more serious note, I prefer more informative naming schemes, for example as implemented in the [colorspace ggplot2 scales.](http://colorspace.r-forge.r-project.org/articles/ggplot2_color_scales.html) Would `scale_color_brewer_binned()` or  `scale_color_brewer_b()` be so terrible? Similarly, we already have `scale_color_viridis_c()` and `scale_color_viridis_d()`, so adding `scale_color_viridis_b()` for binned would be consistent with that scheme.

R/scale-.r
thomasp85: I don't think we can throw an error since this only looks at position scale names... I've changed the ordering so that it defaults to discrete as before

R/scale-brewer.r
thomasp85: I only went this way because of `distiller` and thought it would be a nice joke to continue. I must say that really do not like the look of the `_d|c|b` suffix, so if anything better is available I will always choose that

thomasp85: for the scales that already have a `_d` version then a `_b` is an obvious choice but otherwise I'm forcefully against :-)

R/guide-colorbar.r
hadley: Do you mean: Should the rendered size of the bins be proportional to their length in the data space?

R/scale-.r
hadley: Why can't it be an error? It would only be seen by developers.

thomasp85: The check is in ScaleContinuous, not ScalePositionContinuous... just seems weird to enforce a position specific name at this place

R/guide-colorbar.r
thomasp85: yeah - is that a clearer phrasing?

hadley: Yeah, I think so

thomasp85:
@hadley I consider this done now. Open questions are the two unresolved comments. Can I get you to chime in if you consider them answered and then finish off the review?

R/scale-brewer.r
hadley: I think we're slightly stuck with the overly clever names for now. I'd prefer vintner over mixer, to be consistent with "people" rather than objects.

thomasp85: I can see the argument for it being a person but vintner is just super obscure (or maybe I’m just not well educated). Would “bartender” be too on-the-nose?

clauswilke: I feel the same. If I saw `scale_color_vintner()` I'd have no idea what it does and I probably wouldn't associate it with `scale_color_brewer()` either. I was just thinking along the same lines as Thomas. What about `scale_color_tender()`?

thomasp85: Actually, I’m warming up to it

thomasp85: +1 for tender

clauswilke: One other thought: I can envision somebody brewing colors, and I can also envision somebody distilling discrete colors into a continuum. If we stay with the metaphor of a person working the colors, something like `scale_color_smith()` or `scale_color_plumber()` might not be bad either.

hadley: I think tender is equally too far from a person who makes drinks — it's not a standard shortening of bar tender. How about `scale_brewer_fermenter()`?

thomasp85: We have now settled on `fermenter` — I expect the perfect name will show itself to us after release 😛

R/scale-.r
thomasp85: I'm going to refuse this as it would require a huge rewrite unrelated to this PR. We could track this in another issue, but honestly it doesn't feel important enough for its trouble

PR #3098, opened 2019-01-23
Fix unit-to-numeric coercion assumptions

thomasp85:
This PR addresses several small problems with how ggplot2 treats units that have come up during my rewrite of the unit object. They all centre around the assumptions that units are a numeric vector with attributes, which are only true for freshly created unit objects.

These changes makes ggplot2 work with the new unit spec without breaking backward compatibility with old grid versions.

### DIFF
--- a/R/facet-wrap.r
+++ b/R/facet-wrap.r
@@ -281,10 +281,10 @@ FacetWrap <- ggproto("FacetWrap", Facet,
       axis_mat_y_left[, -1] <- list(zeroGrob())
       axis_mat_y_right[, -ncol] <- list(zeroGrob())
     }
-    axis_height_top <- unit(apply(axis_mat_x_top, 1, max_height), "cm")
-    axis_height_bottom <- unit(apply(axis_mat_x_bottom, 1, max_height), "cm")
-    axis_width_left <- unit(apply(axis_mat_y_left, 2, max_width), "cm")
-    axis_width_right <- unit(apply(axis_mat_y_right, 2, max_width), "cm")
+    axis_height_top <- unit(apply(axis_mat_x_top, 1, max_height, value_only = TRUE), "cm")
+    axis_height_bottom <- unit(apply(axis_mat_x_bottom, 1, max_height, value_only = TRUE), "cm")
+    axis_width_left <- unit(apply(axis_mat_y_left, 2, max_width, value_only = TRUE), "cm")
+    axis_width_right <- unit(apply(axis_mat_y_right, 2, max_width, value_only = TRUE), "cm")
     # Add back missing axes
     if (any(empties)) {
       first_row <- which(apply(empties, 1, any))[1] - 1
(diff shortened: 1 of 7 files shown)


R/grob-dotstack.r
```diff
@@ -13,7 +13,7 @@ dotstackGrob <- function(
         y <- unit(y, default.units)
     if (!is.unit(dotdia))
         dotdia <- unit(dotdia, default.units)
-    if (attr(dotdia,"unit") != "npc")
+    if (!grepl('npc', as.character(dotdia))) # Only cross-version reliable way to check the unit of a unit object
```
clauswilke: Maybe wrap this in a function and make the version dependency explicit by executing different code paths for different version? This would enable us to update to a more elegant check in the future once we stop supporting the old R versions that require the workaround.

thomasp85: The old one was also a hack that could fail (but didn’t for the specific use case). This check is cross-version so I can’t see the benefit of having separate approaches..?

R/facet-wrap.r
```diff
@@ -281,10 +281,10 @@ FacetWrap <- ggproto("FacetWrap", Facet,
       axis_mat_y_left[, -1] <- list(zeroGrob())
       axis_mat_y_right[, -ncol] <- list(zeroGrob())
     }
-    axis_height_top <- unit(apply(axis_mat_x_top, 1, max_height), "cm")
-    axis_height_bottom <- unit(apply(axis_mat_x_bottom, 1, max_height), "cm")
-    axis_width_left <- unit(apply(axis_mat_y_left, 2, max_width), "cm")
-    axis_width_right <- unit(apply(axis_mat_y_right, 2, max_width), "cm")
+    axis_height_top <- unit(apply(axis_mat_x_top, 1, max_height, value_only = TRUE), "cm")
```
karawoo: These lines are now longer than 80 characters, I think tidyverse style would be:

```r
axis_height_top <- unit(
  apply(axis_mat_x_top, 1, max_height, value_only = TRUE),
  "cm"
)
```
Is `value_only = TRUE` needed here since `apply()` coerces the result to numeric anyway? Or is there some inconsistency there that this is addressing?

R/grob-dotstack.r
clauswilke: I see. In any case, just a suggestion: I usually prefer to have such tests encapsulated in a function. Makes the code more readable and easier to maintain.

thomasp85: I can see the point of encapsulation - will do

R/facet-wrap.r
thomasp85: it does for current unit implementation provided that the units are freshly made and haven't been modified in advance (in which case the coercion will fail)

R/grob-dotstack.r
```diff
@@ -13,14 +13,16 @@ dotstackGrob <- function(
         y <- unit(y, default.units)
     if (!is.unit(dotdia))
         dotdia <- unit(dotdia, default.units)
-    if (attr(dotdia,"unit") != "npc")
+    if (!is_npc(dotdia)) # Only cross-version reliable way to check the unit of a unit object
```
clauswilke: The comment should probably be moved to where the function `is_npc()` is defined.

clauswilke: Otherwise looks all good.

PR #3102, opened 2019-01-25
Implement n_breaks in continuous_scale

thomasp85:
This is following thoughts and discussion in #3096 and implement an `n_breaks` argument in `continuous_scale()` to allow guiding the number of major breaks that gets created. Before you'd had to make the breaks manually if you wanted to have a different number of breaks (or construct your own trans object by hand)...

The implementation can feel a bit hacky as it modifies the parent environment of the breaks method of the trans object. I make sure to reset the parent environment on exit so there should be no side effects, but a more elegant approach is if scales allowed dynamic setting of `n` in `trans$breaks()`

Anyway, consider this PR as discussion ground for the feature and we can figure out the best implementation later if we want it...

example use:

``` r
library(ggplot2)

ggplot(mtcars) + 
  geom_point(aes(mpg, disp))
```

``` r

ggplot(mtcars) + 
  geom_point(aes(mpg, disp)) + 
  scale_x_continuous(n_breaks = 10)
```

<sup>Created on 2019-01-25 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


hadley:
Can't we implement this by setting `breaks` to a function?

thomasp85:
Indeed, but I think having a direct argument accepting an integer will make users actually use it... having to construct a function that wraps something from the labelling package seems a high bar for such a simple need (IMO)

thomasp85:
Or do you mean that we should do this under the hood during construction of the scale?

thomasp85:
I don’t think that it is super important, but it needs to be figured out for scale_binned anyway where it’ll be an important argument. I’d much rather fix this in scales, so if you’d accept a fix there I’d happily provide one

hadley:
Yeah, I'd happily accept a fix in scales

R/scale-.r
```diff
@@ -888,68 +890,16 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
-#'   [guides()] for more info.
-#' @param position The position of the axis. "left" or "right" for vertical
-#' scales, "top" or "bottom" for horizontal scales
-#' @param super The super class to use for the constructed scale
-#' @keywords internal
-continuous_scale <- function(aesthetics, scale_name, palette, name = waiver(),
-  breaks = waiver(), minor_breaks = waiver(), n_breaks = NULL, labels = waiver(),
-  limits = NULL, rescaler = rescale, oob = censor, expand = waiver(), na.value = NA_real_,
-  trans = "identity", guide = "legend", position = "left", super = ScaleContinuous) {
+ScaleBinned <- ggproto("ScaleBinned", Scale,
```
hadley: What's up with the big diff here?

thomasp85: You must be looking at the code from before I merged master in? `scale-.r` was reorganised while this PR was developed

PR #3106, opened 2019-01-26
Add default add to functions method

JohnMount:
Add ggplot_add.function() to plot-construction.r. This allows examples such as "ggplot(mtcars, aes(x = disp, y = mpg)) + geom_point".  This follows the pattern in magrittr of allowing both "5 %>% sin" and "5 %>% sin()".  This increased regularity may be helpful for new users.  New example worked and tests passed after change.


yutannihilation:
Thanks for the idea. But, `+` and `%>%` have very different semantics.

`%>%` forwards an object to a function (e.g. `sin`) or function call (e.g. `sin()`). Since what `%>%` does is to modify an expression and construct a call, it matters little whether it's a function or a function call (until you get supprised that `5 %>% base::sin` doesn't work...)

On the other hand, `+` just adds the evaluated results of both sides. It doesn't modify the expression. So, I think a function (e.g. `geom_point`) and the result of a function call (e.g. result of `geom_point()`) are different things here.

I don't strongly disagree with this PR, but we need some discussion about the semantics, at least.

JohnMount:
Well part of what inspired it was that students generalize from one to the other (and often both pipes an ggplot2 compositing are taught in the same course) plus the error message is confusing.

```

**Comment by thomasp85** [comment]
I can’t really figure out whether I like this idea or not. I cannot come up with any strong arguments against, but on the other hand I generally am against making piping and adding more alike as I think it will lead to greater confusion in the end...

If the error is confusing we should def tackle that, but I’d like to hear from @hadley, @karawoo, and @clauswilke about this PR before we decide a course of action

**Comment by clauswilke** [comment]
Same here. I can't point to any reason why it's obviously problematic, but my gut feeling says no. I'm not sure it'll help students if they think `geom_point` and `geom_point()` are the same thing.

**Comment by clauswilke** [comment]
However, I agree that the error message is a problem. If the error message was the following, would that help?
```
#> Error: Don't know how to add geom_point to a plot. Did you mean geom_point()?
```

karawoo:
I too think that blurring the distinction between `%>%` and `+` is likely to cause confusion of its own, but agree that the error message should be improved at the very least.

hadley:
I definitely don’t think it’s a good idea. A better error might be nice, but it’s not something I’ve seen widespread confusion with.

yutannihilation:
I feel Claus's suggestion is compact enough, if we are to implement a better error.

> Did you mean geom_point()?

Let this issue closed for now (until the need arises), in the hope that someday people will get unfamiliar with using `%>%` without `()`, which [the tidyverse style doesn't recommend](https://style.tidyverse.org/pipes.html#no-arguments).

PR #3108, opened 2019-01-29
Fix typo

dvcv:


batpigandme:
Thanks

yutannihilation:
Ah, maybe we need to update `README.md` as well?

PR #3109, opened 2019-01-29
add length argument to geom_rug to allow different lengths

daniel-wells:
``` r
library(ggplot2)

# current behaviour with default rug length
ggplot(airquality,
       aes(x = Ozone,
           y = Solar.R)) + 
  geom_point() +
  geom_rug()
#> Warning: Removed 42 rows containing missing values (geom_point).
```

``` r

# with smaller rug lengths
ggplot(airquality,
       aes(x = Ozone,
           y = Solar.R)) + 
  geom_point() +
  geom_rug(length=0.015)
#> Warning: Removed 42 rows containing missing values (geom_point).
```

``` r

# with larger rug lengths
ggplot(airquality,
       aes(x = Ozone,
           y = Solar.R)) + 
  geom_point() +
  geom_rug(length=0.05) +
  scale_x_continuous(expand=c(0.1,0.1)) +
  scale_y_continuous(expand=c(0.1,0.1))
#> Warning: Removed 42 rows containing missing values (geom_point).
```

<sup>Created on 2019-01-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


yutannihilation:
This was once addressed in #662, which was not merged (due to the development bandwidth shortage for reviewing?). Comparing with that, I have two questions:

* Is `length` the best name for the parameter? (`width`, or `size`?)
* Should the value be specified with or without the unit? (`3` or `unit(0.3, "npc")`?) In other places, the unitless numbers are interpreted as `.pt`.

clauswilke:
I think rug length should be specified in units, just like axis tick length. Otherwise, rug length will depend on the chosen coordinate system, and that is awkward. In fact, we see in the example figures that horizontal and vertical rug lines have different lengths. I realize this is how the current code behaves, but I think it can be done better, or at least we can provide the option to do better, even if we leave the current default in npc units.

yutannihilation:
> I think rug length should be specified in units, just like axis tick length. Otherwise, rug length will depend on the chosen coordinate system, and that is awkward.

I agree. Specifying in units gives us more flexibility.

daniel-wells:
Thanks for the feedback!

I have changed the length specification into units and added a test based on the one in #662

I chose length for the name as size is currently used to specify the thickness, and width sounds like it should control the thickness for the vertical bottom or top rugs. Looking at #662 we could use rugsize or ruglength instead?

``` r
library(ggplot2)

ggplot(airquality,
       aes(x = Ozone,
           y = Solar.R)) + 
  geom_point() +
  geom_rug(length = unit(10,"pt"))
#> Warning: Removed 42 rows containing missing values (geom_point).
```



<sup>Created on 2019-02-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

yutannihilation:
> I chose length for the name as size is currently used to specify the thickness, and width sounds like it should control the thickness for the vertical bottom or top rugs.

Thanks, then it sounds reasonable to choose `length`. (Sorry, I forgot we already use `length` as in `axis.ticks.length` argument of `theme()`).

R/geom-rug.r
```diff
@@ -43,11 +40,17 @@
 #'    coord_cartesian(clip = "off") +
 #'    theme(plot.margin = margin(1, 1, 1, 1, "cm"))
 #'
+#' # increase the line length and
+#' # expand axis to avoid overplotting
+#' p + geom_rug(length = unit(0.05,"npc")) +
+#'    scale_y_continuous(expand=c(0.1,0.1))
```
yutannihilation: Could you follow [the tidyverse style guide](https://style.tidyverse.org/syntax.html#spacing)? You need more spaces here.

tests/testthat/test-geom-rug.R
```diff
@@ -20,3 +20,10 @@ test_that("coord_flip flips the rugs", {
   expect_equal(length(b[[1]]$children[[1]]$y0), 1)
   expect_equal(length(b[[1]]$children[[1]]$y1), 1)
 })
+
+
+test_that("Rug length needs unit object", {
+  p <- ggplot(df, aes(x,y))
+  expect_is(p + geom_rug(length=grid::unit(0.01, "npc")), "ggplot")
```
yutannihilation: Could you add some tests about the actual lengths of the grob? As the tests above does, you can check these values:

``` r
library(ggplot2)

n = 10
df <- data.frame(x = 1:n, y = (1:n)^3)

g <- layer_grob(ggplot(df, aes(x, y)) + geom_point() + geom_rug(sides = 'l'), 2)
g[[1]]$children[[1]]$x0
#> [1] 0npc
g[[1]]$children[[1]]$x1
#> [1] 0.03npc
```

<sup>Created on 2019-02-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # ggplot2 3.1.0.9000
 
+* `geom_rug()` gains a "length" option to allow for changing the length of the rug lines. (@daniel-wells, #3109)
```
yutannihilation: "length" should be `length`.

R/geom-rug.r
```diff
@@ -76,7 +80,10 @@ geom_rug <- function(mapping = NULL, data = NULL,
 GeomRug <- ggproto("GeomRug", Geom,
   optional_aes = c("x", "y"),
 
-  draw_panel = function(data, panel_params, coord, sides = "bl", outside) {
+  draw_panel = function(data, panel_params, coord, sides = "bl", length, outside) {
+    if (!is(length, "unit")) {
```
yutannihilation: You can use `inherits()` for S3 classes.

R/geom-rug.r
```diff
@@ -76,7 +80,10 @@ geom_rug <- function(mapping = NULL, data = NULL,
...
+      stop("'length' must be a 'unit' object.")
```
yutannihilation: Please add `call. = FALSE` to suppress verbose tracebacks. c.f https://style.tidyverse.org/error-messages.html

daniel-wells:
Thanks for the review! I have incorporated the suggested improvements :)

yutannihilation:
@clauswilke Do you have any more concerns or comments? I'll wait if you need some time to look the code more detailedly (I know you are busy).

R/geom-rug.r
```diff
@@ -76,7 +80,10 @@ geom_rug <- function(mapping = NULL, data = NULL,
 GeomRug <- ggproto("GeomRug", Geom,
   optional_aes = c("x", "y"),
 
-  draw_panel = function(data, panel_params, coord, sides = "bl", outside) {
+  draw_panel = function(data, panel_params, coord, sides = "bl", length, outside) {
```
clauswilke: It's generally better to add new arguments (here, `length`) after previously existing arguments, in case some downstream code depends on the order of the arguments. Also, better to provide a default length here (and maybe also add a default `outside` value).

R/geom-rug.r
```diff
@@ -4,17 +4,14 @@
 #' with the two 1d marginal distributions. Rug plots display individual
 #' cases so are best used with smaller datasets.
 #'
-#' The rug lines are drawn with a fixed size (3\% of the total plot size) so
-#' are dependent on the overall scale expansion in order not to overplot
-#' existing data.
```
clauswilke: I think the deleted lines 7-9 contain important information and should be kept. But edit so it's clear that the default size can be modified.

Example:
> By default, the rug lines are drawn with a length that corresponds to 3\% of the total plot size. Since the default scale expansion of for continuous variables is 5\% at both ends of the scale, the rug will not overlap with any data points under the default settings.

clauswilke:
Looks good to me after the two comments I made have been addressed. I don't entirely understand how the regression tests work but I trust you've thought them through.

yutannihilation:
Thanks!

> I don't entirely understand how the regression tests work but I trust you've thought them through.

The current master only checks the numbers of rug lines on a normal plot and on the flipped version. This PR adds tests about the validation on `length` and the actual length of rug lines. We might be able to check things more thoroughly, but I feel it's enough.

tests/testthat/test-geom-rug.R
```diff
@@ -20,3 +20,26 @@ test_that("coord_flip flips the rugs", {
   expect_equal(length(b[[1]]$children[[1]]$y0), 1)
   expect_equal(length(b[[1]]$children[[1]]$y1), 1)
 })
+
+test_that("Rug length needs unit object", {
+  p <- ggplot(df, aes(x,y))
+  expect_is(p + geom_rug(length=grid::unit(0.01, "npc")), "ggplot")
```
yutannihilation: Sorry, one more comment here. This `expect_is()` seems no more needed. If `geom_rug()` raises an error with a `unit` object, the test case below will fail so that we can notice something is wrong.

R/geom-rug.r
daniel-wells: Should I change the ordering in the `geom_rug` function as well as, or instead of, `draw_panel`? It seems changing the order or defaults for draw_panel makes no difference when adding geom_rug() to a plot.

yutannihilation: No. In `geom_rug()`, `length` is placed after `...`, which means the user need to pass the argument with its name. So, there's no chance for `length` to be mistaken. But, in `draw_panel()`, `outside` can be specified without the name. If some existing code uses `outside` positionally, it might fail due to this change.

Does this answer your question?

(Sorry that I didn't notice this on my review...)

daniel-wells: Ah I see thanks! I've implemented the three new suggestions.

R/geom-rug.r
```diff
@@ -4,9 +4,10 @@
...
+#' By default, the rug lines are drawn with a length that corresponds to 3%
```
yutannihilation: You need to escape this `%` like the current one:

https://github.com/tidyverse/ggplot2/blob/9df0a076025aa14ec8cf079ae53e0d662884625a/R/geom-rug.r#L7

R/geom-rug.r
```diff
@@ -4,9 +4,10 @@
...
+#' of the total plot size. Since the default scale expansion of for continuous
+#' variables is 5% at both ends of the scale, the rug will not overlap with
```
yutannihilation: ditto

R/geom-rug.r
daniel-wells: Oops sorry! corrected

yutannihilation:
Perfect, thanks!

daniel-wells:
Awesome! Thanks for all your feedback and help :)

Issue #2525, opened 2018-04-16
sf graticules are drawn even if relevant theme elements are set to element_blank()


karldw:
`theme_void()` is a great way to clear out all the extra lines in a chart.  My intuition was that this theme would remove to graticules from `geom_sf()` as well, but it doesn't.  Do you think it should?

```r
library(sf) 
library(ggplot2)

nc <- st_read(system.file("gpkg/nc.gpkg", package = "sf")) 

ggplot(nc) + 
  geom_sf(data = nc, aes(fill = AREA)) + 
  theme_void()

# The actual way to remove graticules:
ggplot(nc) + 
  geom_sf(data = nc, aes(fill = AREA)) + 
  coord_sf(datum = NA) + 
  theme_void()
```

Here's a `theme_void` map of North Carolina:
![North Carolina with graticule](https://user-images.githubusercontent.com/12504708/38788362-1e0d215a-40e8-11e8-8d17-3b2ed8959d00.png)

And here's what I expected, with no graticule:
![North Carolina no graticule](https://user-images.githubusercontent.com/12504708/38788354-14030562-40e8-11e8-9866-83542389080f.png)

Ref: https://github.com/tidyverse/ggplot2/issues/2071

hadley:
Duplicate of #2389

mikmart:
@hadley this issue still persists after #2389 and the underlying cause seems to be that `CoordSf$render_bg()` doesn't handle `element_blank()` as the `panel.grid` element properly.

clauswilke:
I can confirm that this problem still exists in the current github master. I'm reopening the issue.

``` r
library(sf) 
#> Linking to GEOS 3.6.1, GDAL 2.1.3, proj.4 4.9.3
library(ggplot2)

nc <- st_read(system.file("gpkg/nc.gpkg", package = "sf"), quiet = TRUE) 

ggplot(nc) + 
  geom_sf(data = nc, aes(fill = AREA)) + 
  theme_void()
```

![](https://i.imgur.com/DAjduHm.png)

Created on 2018-09-24 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

PR #3113, opened 2019-02-01
Better respect theme settings when drawing coord_sf() panel grid

### LINKED ISSUE #2991: coord_sf() doesn't adjust line size for graticule lines
Line sizes in themes are generally provide in units of mm and converted to grid units by multiplication by `.pt`. However, `coord_sf()` doesn't do that, it uses the value of `size` directly. Reprex:

``` r
library(ggplot2)

df <- data.frame(x = c(1, 2, 3), y = c(1, 2, 3))

cowplot::plot_grid(
  p + coord_fixed() + theme(panel.grid.minor = element_blank()) +
    ggtitle("coord_fixed()"),
  p + coord_sf() +
    ggtitle("coord_sf() default"),
  p + coord_sf() + theme(panel.grid = element_line(size = 0.5*.pt)) +
    ggtitle("coord_sf() theme adjusted"),
  nrow = 2
)
```

<sup>Created on 2018-11-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

The offending lines are here:
https://github.com/tidyverse/ggplot2/blob/6786969ecb2ad38876ec1df43969467339d7818e/R/sf.R#L576-L582
The fix is simple, but it would affect the visual appearance of a lot of plots out in the wild.

clauswilke:
This PR makes `coord_sf()` behave more similarly to `coord_cartesian()` when drawing the panel grid, by applying the same line-width correction and by respecting `element_blank()`. Closes #2991 and #2525.

Reprex 1:

``` r
library(ggplot2)

df <- data.frame(x = c(1, 2, 3), y = c(1, 2, 3))

# default grid
cowplot::plot_grid(
  p + coord_fixed() + theme(panel.grid.minor = element_blank()) +
    ggtitle("coord_fixed()"),
  p + coord_sf() +
    ggtitle("coord_sf()")
)
```

``` r

# different theme
cowplot::plot_grid(
  p + coord_fixed() + theme_bw() + theme(panel.grid.minor = element_blank()) +
    ggtitle("coord_fixed()"),
  p + coord_sf() + theme_bw() + 
    ggtitle("coord_sf()")
)
```

``` r

# setting grid lines to element_blank()
cowplot::plot_grid(
  p + coord_fixed() + theme(panel.grid = element_blank()) +
    ggtitle("coord_fixed()"),
  p + coord_sf() + theme(panel.grid = element_blank()) +
    ggtitle("coord_sf()")
)
```

<sup>Created on 2019-02-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Reprex 2:

``` r
library(sf) 
#> Linking to GEOS 3.6.1, GDAL 2.1.3, PROJ 4.9.3
library(ggplot2)

nc <- st_read(system.file("gpkg/nc.gpkg", package = "sf"), quiet = TRUE) 

ggplot(nc) + 
  geom_sf(data = nc, aes(fill = AREA)) + 
  theme_void()
```

<sup>Created on 2019-02-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # ggplot2 3.1.0.9000
 
+* `coord_sf()` graticule lines are now drawn in the same thickness as 
+  panel grid lines in `coord_cartesian()`, and seting panel grid
+  lines to `element_blank()` now also works in `coord_sf()` 
+  (@clauswilke, #2991, #2525).
+
 * `geom_hline()`, `geom_vline()`, and `geom_abline()` now throw a warning if the user supplies both an `xintercept`, `yintercept`, or `slope` value and a mapping (@RichardJActon, #2950).
   
 * `scale_color_continuous()` now points at `scale_colour_continuos()` so that it 
--- a/R/coord-sf.R
+++ b/R/coord-sf.R
@@ -201,7 +201,11 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     if (inherits(el, "element_blank")) {
       grobs <- list(element_render(theme, "panel.background"))
     } else {
-      line_gp <- gpar(col = el$colour, lwd = el$size*.pt, lty = el$linetype)
+      line_gp <- gpar(
+        col = el$colour,
+        lwd = len0_null(el$size*.pt),
+        lty = el$linetype
+      )
       grobs <- c(
         list(element_render(theme, "panel.background")),
         lapply(sf::st_geometry(panel_params$graticule), sf::st_as_grob, gp = line_gp)
--- a/R/coord-sf.R
+++ b/R/coord-sf.R
@@ -195,7 +195,14 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
 
   render_bg = function(self, panel_params, theme) {
     el <- calc_element("panel.grid.major", theme)
-    line_gp <- gpar(col = el$colour, lwd = el$size, lty = el$linetype)
+
+    # we don't draw the graticules if the major panel grid is
+    # turned off
+    if (inherits(el, "element_blank")) {
+      return(zeroGrob())
+    }
+
+    line_gp <- gpar(col = el$colour, lwd = el$size*.pt, lty = el$linetype)
     grobs <- c(
       list(element_render(theme, "panel.background")),
       lapply(sf::st_geometry(panel_params$graticule), sf::st_as_grob, gp = line_gp)
--- a/R/coord-sf.R
+++ b/R/coord-sf.R
@@ -199,14 +199,14 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
     # we don't draw the graticules if the major panel grid is
     # turned off
     if (inherits(el, "element_blank")) {
-      return(zeroGrob())
+      grobs <- list(element_render(theme, "panel.background"))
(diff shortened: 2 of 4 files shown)


clauswilke:
@thomasp85 You previously [commented:](https://github.com/tidyverse/ggplot2/issues/2991#issuecomment-454375481)

> I think this makes sense to change irrespectively of the changes in the wild it may affect... but better wait until the next big(ish) release

Do you feel we're at that point of the development cycle now? It seems to me we're making substantial changes throughout the codebase.

thomasp85:
Pretty sure next release will be major as I’m working on two substantial PRs (one of them sf related so we might as well combine that in the same release)

clauswilke:
To answer my own question about visual tests: I think whenever a graticule is drawn the result may depend on the specific version of sf at least.

thomasp85:
It will very much depend on sf

clauswilke:
Well, I've added one visual test, for the case where no graticule is drawn and no CRS is set (and we only use `geom_point()`). I think it should be fine, but let's see whether it passes on Travis.

clauswilke:
I just noticed that this PR currently introduces a regression (see below). I'll need to fix this before it can be merged.

``` r
library(ggplot2)

df <- data.frame(x = c(1, 2, 3), y = c(1, 2, 3))
p <- ggplot(df, aes(x, y)) + 
  geom_point() +
  theme_void() +
  theme(panel.grid.major = element_line(color = "white"))

# works fine with coord_cartesian()
p
```



``` r

# doesn't work
p + coord_sf()
#> Error in check.length(gparname): 'gpar' element 'lwd' must not be length 0
```
<sup>Created on 2019-02-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

lionel-:
> @lionel- Do you think it's possible to add some visual tests for coord_sf() if we don't actually use geom_sf() (just like the first reprex here)? Should that work, or would you also expect it to be unstable and depend on the specific versions of GDAL, PROJ, etc.?

Sorry I am not familiar with sf.

clauswilke:
@lionel- No problem. I think there's a broader issue here with geoms that may produce visually identical output via different but equivalent mathematical representations. I'm running into this with the isoband library right now. I'll open an issue on vdiffr.

lionel-:
The solution is probably to use bitmap diffing with a threshold for these cases.

clauswilke:
@thomasp85 Any concerns about the specific visual test I added? I've reproduced it here for your reference. Because it doesn't have a panel grid, I think it'll be robust to changes in sf, but I'll remove it if you think it remains a risk.

``` r
library(ggplot2)

df <- data.frame(x = c(1, 2, 3), y = c(1, 2, 3))
ggplot(df, aes(x, y)) +
  geom_point() +
  coord_sf() +
  theme_gray() + # to test for presence of background grob
  theme(panel.grid = element_blank())
```



<sup>Created on 2019-02-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
Unless anybody has any other concerns I'm going to go ahead and merge this. In the unlikely event that the sf visual test starts acting up in the future we can always disable it then.

PR #3124, opened 2019-02-11
Modified line glyph as used in the fable package

mitchelloharawild:
As requested from @earowang 's rstudioconf19 talk: https://slides.earo.me/rstudioconf19/#26

Background shade is not included because it is representative of the associated interval.

``` r
library(ggplot2)
ggplot(economics_long, aes(date, value01, colour = variable)) +
  geom_line()
```

``` r

devtools::load_all("~/github/ggplot2")
#> Loading ggplot2
ggplot(economics_long, aes(date, value01, colour = variable)) +
  geom_line()
```

<sup>Created on 2019-02-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/legend-draw.r
+++ b/R/legend-draw.r
@@ -112,7 +112,14 @@ draw_key_crossbar <- function(data, params, size) {
 draw_key_path <- function(data, params, size) {
   data$linetype[is.na(data$linetype)] <- 0
 
-  segmentsGrob(0.1, 0.5, 0.9, 0.5,
+  # Scale the arrow to fit within the draw key
+  if(!is.null(params$arrow)){
+    params$arrow$length <- unit(0.3, "npc")
+  }
+
+  linesGrob(
+    x = c(0, 0.4, 0.6, 1),
+    y = c(0.1, 0.6, 0.4, 0.9),
     gp = gpar(
       col = alpha(data$colour, data$alpha),
       lwd = data$size * .pt,


clauswilke:
Could you first open an issue explaining exactly what the desired feature is, so we can have a discussion about how to approach this? If we merged this as is, we'd change everybody's plots, and it's not clear people would be happy about this. I suspect the correct approach would make it possible to choose different legend styles for each geom.

mitchelloharawild:
This is purely at the request of @hadley during [rstudio::conf 2019](https://www.rstudio.com/conference/).
Perhaps he can better explain why he wants this.

clauswilke:
I see. Hadley, can we take this as an opportunity to make the legend glyphs configurable, maybe as part of the theme in conjunction with the other geom defaults that we want to make themeable? I often change legend glyphs by overwriting the `draw_key` slot of a geom. Having a more elegant approach to this would be appreciated.

hadley:
I think the new display looks a lot nicer, but you're probably right that this will cause anxiety among people. Maybe we could make it an option so you could restore the previous behaviour? I'm not sure we want to fully themify the key for this one use case.

clauswilke:
I think there's also the issue of combined legends, where the new glyph can look strange (see below). In any case, I'll investigate making this configurable.

``` r
library(ggplot2)

draw_key_path2 <- function(data, params, size) {
  data$linetype[is.na(data$linetype)] <- 0

  grid::linesGrob(
    x = c(0, 0.4, 0.6, 1),
    y = c(0.1, 0.6, 0.4, 0.9),
    gp = grid::gpar(
      col = scales::alpha(data$colour, data$alpha),
      lwd = data$size * .pt,
      lty = data$linetype,
      lineend = "butt"
    )
  )
}

GeomLine$draw_key <- draw_key_path2

df <- data.frame(
  x = rep(1:3, 3),
  y = c(1, 2, 3, 2, 4, 2, 5, 3, 1),
  type = rep(letters[1:3], each = 3)
)
ggplot(df, aes(x, y, color = type)) + geom_line() + geom_point()
```



<sup>Created on 2019-02-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
Closing this PR as this feature is now available in ggplot2 master.

PR #3128, opened 2019-02-11
Holed polygons

thomasp85:
This PR adds the ability to draw polygons with holes to `geom_polygon`. It does so by adding a new, optional, aesthetic called `subgroup`. If provided, and R >= 3.6 the geom will use the new improved pathGrob to draw the polygons.

Questions to mull over:
- Do you agree on the `subgroup` name. I've elected for a generic name so it can be reused elsewhere if secondary grouping is needed, rather than using something that signifies the "hole"-ness of it
- Currently the geom will draw the path using the "evenodd" rule as this is how most people will probably think it works (winding order is quite specialised knowledge). It could be made a parameter of the geom, but isn't currently as I think it would just muddy the waters

Example:
```r
ids <- factor(c("1.1", "2.1", "1.2", "2.2", "1.3", "2.3"))

values <- data.frame(
  id = ids,
  value = c(3, 3.1, 3.1, 3.2, 3.15, 3.5)
)

positions <- data.frame(
  id = rep(ids, each = 4),
  x = c(2, 1, 1.1, 2.2, 1, 0, 0.3, 1.1, 2.2, 1.1, 1.2, 2.5, 1.1, 0.3,
        0.5, 1.2, 2.5, 1.2, 1.3, 2.7, 1.2, 0.5, 0.6, 1.3),
  y = c(-0.5, 0, 1, 0.5, 0, 0.5, 1.5, 1, 0.5, 1, 2.1, 1.7, 1, 1.5,
        2.2, 2.1, 1.7, 2.1, 3.2, 2.8, 2.1, 2.2, 3.3, 3.2)
)
datapoly <- merge(values, positions, by = c("id"))

holes <- do.call(rbind, lapply(split(datapoly, datapoly$id), function(df) {
  df$x <- df$x + 0.5 * (mean(df$x) - df$x)
  df$y <- df$y + 0.5 * (mean(df$y) - df$y)
  df
}))
datapoly$subid <- 1L
holes$subid <- 2L
datapoly <- rbind(datapoly, holes)

p <- ggplot(datapoly, aes(x = x, y = y)) +
  geom_polygon(aes(fill = value, group = id, subgroup = subid))
p
```
image

### DIFF
--- a/R/geom-polygon.r
+++ b/R/geom-polygon.r
@@ -60,21 +60,24 @@
 #' # And if the positions are in longitude and latitude, you can use
 #' # coord_map to produce different map projections.
 #'
-#' # As of R version 3.6 geom_polygon() supports polygons with holes
-#' # Use the subgroup aesthetic to differentiate holes from the main polygon
+#' if (packageVersion("grid") >= "3.6") {
+#'   # As of R version 3.6 geom_polygon() supports polygons with holes
+#'   # Use the subgroup aesthetic to differentiate holes from the main polygon
 #'
-#' holes <- do.call(rbind, lapply(split(datapoly, datapoly$id), function(df) {
-#'   df$x <- df$x + 0.5 * (mean(df$x) - df$x)
-#'   df$y <- df$y + 0.5 * (mean(df$y) - df$y)
-#'   df
-#' }))
-#' datapoly$subid <- 1L
-#' holes$subid <- 2L
-#' datapoly <- rbind(datapoly, holes)
+#'   holes <- do.call(rbind, lapply(split(datapoly, datapoly$id), function(df) {
+#'     df$x <- df$x + 0.5 * (mean(df$x) - df$x)
+#'     df$y <- df$y + 0.5 * (mean(df$y) - df$y)
+#'     df
+#'   }))
+#'   datapoly$subid <- 1L
+#'   holes$subid <- 2L
+#'   datapoly <- rbind(datapoly, holes)
 #'
-#' p <- ggplot(datapoly, aes(x = x, y = y)) +
-#'   geom_polygon(aes(fill = value, group = id, subgroup = subid))
-#' p
+#'   p <- ggplot(datapoly, aes(x = x, y = y)) +
+#'     geom_polygon(aes(fill = value, group = id, subgroup = subid))
+#'   p
+#' }
+#'
 geom_polygon <- function(mapping = NULL, data = NULL,
                          stat = "identity", position = "identity",
                          rule = "evenodd",
--- a/R/geom-polygon.r
+++ b/R/geom-polygon.r
@@ -127,8 +127,8 @@ GeomPolygon <- ggproto("GeomPolygon", Geom,
         )
       )
     } else {
-      if (!grid_has_multipath) {
-        stop("Polygons with holes are only supported in R versions from 3.6 and onwards", call. = FALSE)
+      if (utils::packageVersion('grid') < "3.6") {
+        stop("Polygons with holes requires R 3.6 or above", call. = FALSE)
       }
       # Sort by group to make sure that colors, fill, etc. come in same order
       munched <- munched[order(munched$group, munched$subgroup), ]
--- a/R/geom-polygon.r
+++ b/R/geom-polygon.r
@@ -3,7 +3,10 @@
 #' Polygons are very similar to paths (as drawn by [geom_path()])
 #' except that the start and end points are connected and the inside is
 #' coloured by `fill`. The `group` aesthetic determines which cases
-#' are connected together into a polygon.
+#' are connected together into a polygon. From R 3.6 and onwards it is possible
+#' to draw polygons with holes by providing a subgroup aesthetic that
+#' differentiate the outer ring points from those describing holes in the
+#' polygon.
 #'
 #' @eval rd_aesthetics("geom", "polygon")
 #' @seealso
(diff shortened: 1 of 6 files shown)


clauswilke:
Thanks for making this happen! I can’t really test this right now, because I don’t want to touch my R install during the final stages of book production, in case I have to re-render any figures.

I like the name `subgroup`, since as you say it could be used in other contexts as well.

I think winding order should be made available as an option. I originally thought I’d need it for isobands, but thinking some more about it it may not matter. Am I correct that for perfectly nested polygon rings running in alternating directions evenodd and winding order give the same result?

It would be good to test this with the isoband output on some non-trivial datasets. Would you mind trying this out? The `isobands()` function returns a list of polygons specified just like `grid.path()` expects them, with `x`, `y`, and `id`, so should be clear how to convert this into four columns `x`, `y`, `group`, `subgroup`.

clauswilke:
One question: Do the polygon paths need to be closed (last point = first point) or not? `grid.polygon()` and `grid.path()` behave differently in this regard, I believe.

R/geom-polygon.r
```diff
@@ -3,7 +3,10 @@
 #' Polygons are very similar to paths (as drawn by [geom_path()])
 #' except that the start and end points are connected and the inside is
 #' coloured by `fill`. The `group` aesthetic determines which cases
-#' are connected together into a polygon.
+#' are connected together into a polygon. From R 3.6 and onwards it is possible
+#' to draw polygons with holes by providing a subgroup aesthetic that
+#' differentiate the outer ring points from those describing holes in the
```
clauswilke: “differentiates”

R/geom-polygon.r
```diff
@@ -52,6 +55,22 @@
 #'
 #' # And if the positions are in longitude and latitude, you can use
 #' # coord_map to produce different map projections.
+#'
+#' # As of R version 3.6 geom_polygon supports polygons with holes
```
clauswilke: Add `()` after `geom_polygon`?

R/geom-polygon.r
```diff
@@ -83,30 +102,60 @@ GeomPolygon <- ggproto("GeomPolygon", Geom,
...
-    # We'll pull the first value from each group, and assume all these values
-    # are the same within each group.
-    first_idx <- !duplicated(munched$group)
-    first_rows <- munched[first_idx, ]
+    if (is.null(munched$subgroup)) {
+      # Sort by group to make sure that colors, fill, etc. come in same order
+      munched <- munched[order(munched$group), ]
+
+      # For gpar(), there is one entry per polygon (not one entry per point).
+      # We'll pull the first value from each group, and assume all these values
```
clauswilke: Other geoms check this assumption and issue a warning if an aesthetic changes where it’s not supposed to. See e.g. `geom_area()`. Needed here?

R/geom-polygon.r
```diff
@@ -83,30 +102,60 @@ GeomPolygon <- ggproto("GeomPolygon", Geom,
...
-        )
+      ggname("geom_polygon",
+             polygonGrob(munched$x, munched$y, default.units = "native",
+                         id = munched$group,
+                         gp = gpar(
+                           col = first_rows$colour,
+                           fill = alpha(first_rows$fill, first_rows$alpha),
+                           lwd = first_rows$size * .pt,
+                           lty = first_rows$linetype
+                         )
```
clauswilke: Indentation

R/geom-polygon.r
```diff
@@ -83,30 +102,60 @@ GeomPolygon <- ggproto("GeomPolygon", Geom,
...
+                      id = id, pathId = munched$group,
+                      rule = "evenodd",
+                      gp = gpar(
+                        col = first_rows$colour,
+                        fill = alpha(first_rows$fill, first_rows$alpha),
+                        lwd = first_rows$size * .pt,
+                        lty = first_rows$linetype
+                      )
+             )
+      )
```
clauswilke: Indentation

thomasp85:
For the example you describe "evenodd" and "winding" should give the same result. I don't feel strongly about omitting it so I'll add a `rule` argument to control it (which will default to `"evenodd"`)

pathGrob uses the same open polygon spec as polygonGrob so no need to repeat the last point

R/geom-polygon.r
thomasp85: I think that may become prohibitly expensive to check for the kind of data that people will throw at this... Maybe make it more clear in the docs?

clauswilke: Do you think a single call to `unique()` is that much more expensive than, say, the ordering of the data that is also happening?
https://github.com/tidyverse/ggplot2/blob/18be30e1515b2b9d05ade4f3f2b5051aa1336fcf/R/geom-ribbon.r#L80-L83

At a minimum, this might be worth a careful benchmark on a very large dataset, such as the output from `isobands()` on a large raster image.

thomasp85: I’ll happily do the benchmark, but the big difference is that geom_ribbon uses a draw_group method whereas geom_polygon uses draw_panel. This means that geom_polygon would have to split up the data and call unique on each sub-data.frame all for the sake of a possible warning. For geom ribbon the split has already been done so it is rather cheap

clauswilke: I see. In any case, it's your call. I won't insist.

clauswilke:
I just wanted to confirm that I tried "evenodd" with the current `geom.path()` and it renders my polygons as I'd expect. So I'm good with "evenodd" as the default.

R/geom-polygon.r
hadley: We may wanted to eventually tackle this problem by providing a new geom where there's one row per polygon and the vertices are stored in a nested data frame.

R/ggplot-global.R
```diff
@@ -44,3 +44,5 @@ ggplot_global$all_aesthetics <- .all_aesthetics
 )
 
 ggplot_global$base_to_ggplot <- .base_to_ggplot
+
+grid_has_multipath <- packageVersion('grid') >= "3.6"
```
hadley: I don't think computing this value at build time is correct. Instead, I think you should just inline into `geom_polygon()`.

PR #3131, opened 2019-02-11
Remove warning when guide functions are used in identity scales

### LINKED ISSUE #3127: Warning message when adding a manual scale in ggplot
On [Stackoverflow](https://stackoverflow.com/questions/54580005/how-to-fix-an-error-when-adding-a-manual-scale-in-ggplot/54586124#54586124) they suggested to report my question as a bug, so here it is:

When I add a manual scale like in the example below. I get the following error message:
"Warning message: In if (self$guide == "none") return() : the condition has length > 1 and only the first element will be used"

```r
d <- data.frame(week = seq(1, 52, 1), 
            revenue = round(runif(52, 0, 100)), 0)

p <- ggplot(data = d, aes(x = week, y = revenue, fill = 'lightskyblue')) + 
  geom_bar(stat = 'identity', colour = 'black') 


p <- p + scale_fill_identity(name = NULL,
                             guide = guide_legend(label.position = 'top'),
                             labels = c('2019'))
```

clauswilke:
This PR closes #3127 by adding a proper check for the presence or absence of a guide in the identity scales.

``` r
devtools::load_all("~/github/ggplot2")
#> Loading ggplot2

d <- base::data.frame(
  week = seq(1, 52, 1), 
  revenue = round(runif(52, 0, 100))
)

# discrete identity scale
ggplot(data = d, aes(x = week, y = revenue, fill = 'lightskyblue')) + 
  geom_col() + 
  scale_fill_identity(guide = guide_legend())
```

``` r

# continuous identity scale
ggplot(data = d, aes(x = week, y = 1, size = revenue/20)) + 
  geom_point() + 
  scale_size_identity(guide = guide_legend())
```

<sup>Created on 2019-02-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/scale-identity.r
+++ b/R/scale-identity.r
@@ -146,7 +146,7 @@ ScaleDiscreteIdentity <- ggproto("ScaleDiscreteIdentity", ScaleDiscrete,
 
   train = function(self, x) {
     # do nothing if no guide, otherwise train so we know what breaks to use
-    if (isTRUE(self$guide == "none")) return()
+    if (identical(self$guide, "none")) return()
     ggproto_parent(ScaleDiscrete, self)$train(x)
   }
 )
@@ -167,7 +167,7 @@ ScaleContinuousIdentity <- ggproto("ScaleContinuousIdentity", ScaleContinuous,
 
   train = function(self, x) {
     # do nothing if no guide, otherwise train so we know what breaks to use
-    if (isTRUE(self$guide == "none")) return()
+    if (identical(self$guide, "none")) return()
     ggproto_parent(ScaleContinuous, self)$train(x)
   }
 )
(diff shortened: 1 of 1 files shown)


R/scale-identity.r
```diff
@@ -146,7 +146,7 @@ ScaleDiscreteIdentity <- ggproto("ScaleDiscreteIdentity", ScaleDiscrete,
 
   train = function(self, x) {
     # do nothing if no guide, otherwise train so we know what breaks to use
-    if (self$guide == "none") return()
+    if (isTRUE(self$guide == "none")) return()
```
yutannihilation: I prefer `identical(self$guide, "none")` in this case.

clauswilke: thanks, fixed.

Issue #3126, opened 2019-02-11
do we really need to attach package maps to use `map_data` ?


moodymudskipper:
The package maps is loaded silently when using `ggplot2::map_data`. 

It does message about the conflicts, but it's easy to miss it and  is especially annoying because it masks `purr::map`.

Is there a good reason why it needs to be attached ? In the current situation I think I'd rather have it fail and tell me to attach `maps` myself, so I can add the library call before attaching *purrr* and control what is happening.

*This happens with version 3.0, apologies if it has been solved since, I'm not able to test it on the last version.*

clauswilke:
The line that is responsible here is this one: 
https://github.com/tidyverse/ggplot2/blob/9108d9bb4a1ef58e813dc59d3a383e7d7edae4c2/R/fortify-map.r#L78

@hadley I've seen the `try_require()` pattern in several places in the ggplot2 code base. Is there a reason we attach the respective packages instead of simply testing whether they exist? I think it would generally be preferable to not attach packages from inside ggplot2.

clauswilke:
In other words, would anything terrible happen if we removed line 47 here:
https://github.com/tidyverse/ggplot2/blob/e9d4e5dd599b9f058cbe9230a6517f85f3587567/R/utilities.r#L45-L49

hadley:
I think you'd need to review on a case-by-case basis. I agree that it's undesirable to load packages, but I'm almost certain it was necessary for maps (but a recent release may have fixed the problem). I also have vague recollection that `stat_smooth()` may have some issues for mgcv — i.e. something like `y ~ mgcv::s(x)` doesn't work?

clauswilke:
How about adding an argument `attach = FALSE` to `try_require()`, so we need to explicitly turn it on when we're sure we need attaching. Initially, all calls to `try_require()` would be changed so they say `attach = TRUE`, and then we can one-by-one remove those statements.

I'm pretty certain mgcv is fine without attaching now. I just used it in my ungeviz package.

hadley:
There are only 5 uses of `try_require()` so I think someone could just remove the `library()` call, and then carefully check the code. I think `StatBinhex` and `StatSummaryHex` will need some `::`.  Looking at the code I now wonder if it was actually quantreg and/or MatrixModels in `stat_quantile()` caused problems. It's also quite possible the problems have been resolved since I wrote the code, or I may have been confused in the first place (I think much of this was written before I understood namespaces).

clauswilke:
[...]

PR #3134, opened 2019-02-12
Don't attach external packages

clauswilke:
This is work in progress, as discussed in #3126. There are a few cases where we attach external packages. Ideally, we would never attach external packages.

### DIFF
--- a/R/stat-quantile.r
+++ b/R/stat-quantile.r
@@ -54,10 +54,13 @@ StatQuantile <- ggproto("StatQuantile", Stat,
 
     if (is.null(formula)) {
       if (method == "rqss") {
-        # we need to attach quantreg for qss to work inside formula
-        require("quantreg")
-        formula <- eval(substitute(y ~ qss(x, lambda = lambda)),
-          list(lambda = lambda))
+        formula <- eval(
+          substitute(y ~ qss(x, lambda = lambda)),
+          list(lambda = lambda)
+        )
+        # make qss function available in case it is needed;
+        # works around limitation in quantreg
+        qss <- quantreg::qss
       } else {
         formula <- y ~ x
       }
@@ -76,10 +79,12 @@ StatQuantile <- ggproto("StatQuantile", Stat,
 
     # if method was specified as a character string, replace with
     # the corresponding function
-    if (is.character(method)) {
-      if (identical(method, "rq")) method <- quantreg::rq
-      else if (identical(method, "rqss")) method <- quantreg::rqss
-      else method <- match.fun(method) # allow users to supply their own methods
+    if (identical(method, "rq")) {
+      method <- quantreg::rq
+    } else if (identical(method, "rqss")) {
+      method <- quantreg::rqss
+    } else {
+      method <- match.fun(method) # allow users to supply their own methods
     }
 
     rbind_dfs(lapply(quantiles, quant_pred, data = data, method = method,
--- a/R/stat-quantile.r
+++ b/R/stat-quantile.r
@@ -50,11 +50,11 @@ StatQuantile <- ggproto("StatQuantile", Stat,
   compute_group = function(data, scales, quantiles = c(0.25, 0.5, 0.75),
                            formula = NULL, xseq = NULL, method = "rq",
                            method.args = list(), lambda = 1, na.rm = FALSE) {
-    try_require("quantreg", "stat_quantile")
+    try_require("quantreg", "stat_quantile", attach = TRUE)
 
     if (is.null(formula)) {
       if (method == "rqss") {
-        try_require("MatrixModels", "stat_quantile")
+        try_require("MatrixModels", "stat_quantile", attach = TRUE)
         formula <- eval(substitute(y ~ qss(x, lambda = lambda)),
           list(lambda = lambda))
       } else {
--- a/R/stat-quantile.r
+++ b/R/stat-quantile.r
@@ -54,9 +54,6 @@ StatQuantile <- ggproto("StatQuantile", Stat,
 
     if (is.null(formula)) {
       if (method == "rqss") {
-        # this should not be needed, since quantreg imports MatrixModels
-        # try_require("MatrixModels", "stat_quantile")
-
         # we need to attach quantreg for qss to work inside formula
         require("quantreg")
         formula <- eval(substitute(y ~ qss(x, lambda = lambda)),
--- a/R/stat-quantile.r
+++ b/R/stat-quantile.r
@@ -50,11 +50,15 @@ StatQuantile <- ggproto("StatQuantile", Stat,
   compute_group = function(data, scales, quantiles = c(0.25, 0.5, 0.75),
                            formula = NULL, xseq = NULL, method = "rq",
                            method.args = list(), lambda = 1, na.rm = FALSE) {
-    try_require("quantreg", "stat_quantile", attach = TRUE)
+    try_require("quantreg", "stat_quantile")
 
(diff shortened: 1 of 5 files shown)


clauswilke:
@hadley I think this is ready for review now. `try_require()` doesn't attach packages anymore. There's one special case in `stat_quantile()` where we do need to attach quantreg, however. I don't think it can be worked around, because the quantreg functions can't handle namespaces in the formula interface (just like mgcv).

moodymudskipper:
Defining in the package unexported functions `rqss <- quantreg::rqss` and `rq <- quantreg::rq` is not good enough ?

yutannihilation:
> Defining in the package unexported functions

We can do it at the moment, but there's no guarantee that package will work only with these functions in future. I think we have no choice but to attach, if the package is designed to work fine only when it is attached.

clauswilke:
Hadley, I agree, but I don't think that's currently possible. See reprex below. The `rqss()` function doesn't work unless quantreg is attached.

I would propose to merge the patch as is, since it's a major improvement over the current status, and then open an issue about that remaining `require()` and contact the quantreg maintainers to see if they can modify their package so we can use it without attaching.

Alternatively, I think we could import quantreg instead of only suggesting it, but that seems the wrong step to take for such a minor feature.

``` r
n <- 200
x <- sort(rchisq(n,4))
z <- x + rnorm(n)
y <- log(x)+ .1*(log(x))^2 + log(x)*rnorm(n)/4 + z

# doesn't work
quantreg::rqss(y ~ qss(x, constraint= "N") + z)
#> Error in qss(x, constraint = "N"): could not find function "qss"

# doesn't work
quantreg::rqss(y ~ quantreg::qss(x, constraint= "N") + z)
#> Error in model.frame.default(formula = y ~ quantreg::qss(x, constraint = "N") + : invalid type (list) for variable 'quantreg::qss(x, constraint = "N")'

# works
library(quantreg)
#> Loading required package: SparseM
#> 
#> Attaching package: 'SparseM'
#> 
#>     backsolve
rqss(y ~ qss(x, constraint= "N") + z)
#> Formula:
#> y ~ qss(x, constraint = "N") + z
#> Quantile fidelity at tau =  0.5 is 21.72566 
#> Estimated Model Dimension is 18
```

<sup>Created on 2019-02-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

hadley:
Random idea: what if we did `qss <- quantreg::qss` just before fitting the model?

R/stat-quantile.r
```diff
@@ -73,7 +74,13 @@ StatQuantile <- ggproto("StatQuantile", Stat,
     }
     grid <- new_data_frame(list(x = xseq))
 
-    method <- match.fun(method)
+    # if method was specified as a character string, replace with
+    # the corresponding function
+    if (is.character(method)) {
```
hadley: You need to protect this block with `is.character()`, and can you please use https://style.tidyverse.org/syntax.html#indenting style?

R/stat-quantile.r
```diff
@@ -54,7 +54,8 @@ StatQuantile <- ggproto("StatQuantile", Stat,
 
     if (is.null(formula)) {
       if (method == "rqss") {
-        try_require("MatrixModels", "stat_quantile")
+        # we need to attach quantreg for qss to work inside formula
+        require("quantreg")
```
hadley: This still needs `try_require()` to get the nice error message

hadley:
Ooops, reading the thread, I see that's what @moodymudskipper suggested — but I think it will work fine given that this I'd argue that the current behaviour (not supporting `quantreq::qss()`) is a buglet in their evaluation of formulas. If it works now, I don't think it is likely that changes to quantreg in the future will break it, or require importing additional functions.

R/stat-quantile.r
clauswilke: Do you mean "you *don't* need ..."? I added `is.character()` in an earlier version that needed it but now it seems superfluous.

R/stat-quantile.r
clauswilke: No, we just called it before entering the outer `if` statement:
https://github.com/tidyverse/ggplot2/blob/8d02957ffed9816046cd85ff0ea2952b4e6a7ad7/R/stat-quantile.r#L53-L61

The `try_require()` for MatrixModels is not needed at all, as far as I can tell, because we are not directly calling any functions from MatrixModels and quantreg imports MatrixModels.

clauswilke:
I want to say I tried something like the `qss <- quantreg::qss` for mgcv in `geom_smooth()` and it didn't work then. But I'll try here and see what happens.

R/stat-quantile.r
hadley: Yup sorry.

clauswilke:
I just committed a version that seems to work without attaching anything. It's not entirely clear to me why. I think the reason is `do.call()` uses `envir = parent.frame()`, and that is the environment of the `compute_group()` function in this case.

``` r
library(ggplot2)

m <- ggplot(mpg, aes(displ, 1 / hwy)) + geom_point()
m + geom_quantile()
#> Smoothing formula not specified. Using: y ~ x
```



``` r
m + geom_quantile(method = "rqss")
#> Smoothing formula not specified. Using: y ~ qss(x, lambda = 1)
```



``` r

sessionInfo()
#> R version 3.5.0 (2018-04-23)
#> Platform: x86_64-apple-darwin15.6.0 (64-bit)
#> Running under: macOS Sierra 10.12.6
#> 
#> Matrix products: default
#> BLAS: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] ggplot2_3.1.0.9000
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.0         knitr_1.20         magrittr_1.5      
#>  [4] tidyselect_0.2.5   munsell_0.5.0      lattice_0.20-35   
#>  [7] colorspace_1.4-0   R6_2.3.0           rlang_0.3.1       
#> [10] stringr_1.3.1      dplyr_0.8.0.9000   tools_3.5.0       
#> [13] grid_3.5.0         gtable_0.2.0       quantreg_5.36     
#> [16] withr_2.1.2.9000   MatrixModels_0.4-1 htmltools_0.3.6   
#> [19] assertthat_0.2.0   yaml_2.2.0         lazyeval_0.2.1    
#> [22] rprojroot_1.3-2    digest_0.6.18      tibble_2.0.1      
#> [25] crayon_1.3.4       Matrix_1.2-14      purrr_0.2.5       
#> [28] glue_1.3.0         evaluate_0.11      rmarkdown_1.10    
#> [31] labeling_0.3       stringi_1.2.4      compiler_3.5.0    
#> [34] pillar_1.3.1       scales_1.0.0       backports_1.1.2   
#> [37] SparseM_1.77       pkgconfig_2.0.2
```

<sup>Created on 2019-02-14 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

yutannihilation:
> I don't think it is likely that changes to quantreg in the future will break it, or require importing additional functions.

OK, agreed. Thanks.

PR #3140, opened 2019-02-13
Update READMEs to reflect 2nd edition cookbook

StefanBRas:
The second edition of the The R Graphics Cookbook has been released, but the READMEs links to the first edition. This pull request should fix this and remove the text stating that a new version is due in 2018.

I have no experience with the amzn.to URL shortener so the link is the full URL obtained by following the current link and then the link to the updated version.


yutannihilation:
Good catch! We can use bit.ly to get an amzn.to URL (I didn't know...). Here's the one I made just now:

https://amzn.to/2TU78ip

yutannihilation:
Thanks!

PR #3141, opened 2019-02-13
Munch polygons

thomasp85:
The recent addition of polygon holes did not consider the munching done when using a polar coordinate system... This PR factors `subgroup` in when dividing parts to be munched, thus avoiding to connect holes to the main polygon


thomasp85:
sorry for the flurry of weird commits in this one... was testing out some git stuff and things got messy

clauswilke:
18 commits for 2 simple lines of code must be some sort of record. :-)

Issue #2776, opened 2018-07-26
Documentation of stat_ellipse() is misleading


MJochim:
According to its documentation, ```stat_ellipse()``` creates “normal confidence ellipses.” These ellipses can be configured by providing a confidence level as the parameter ```level```.

This leads me to believe that ```stat_ellipse()``` creates a confidence ellipse, in the sense that it is the bivariate equivalent of a confidence interval. If this were the case, I would expect the ellipse to shrink with growing sample sizes (as do confidence intervals).

However, I do not observe this shrinking behavior (see code example below).

The documentation does contain a reference to ```?car::ellipse```, which “draw[s] ellipses, including data ellipses, and confidence ellipses.” To the best of my understanding, what ```stat_ellipse()``` produces is a data ellipse, not a confidence ellipse.

I must admit, though, that I am not a statistician and I am not sure if there is a standard way to calculate such confidence ellipses; so my expectation to have one drawn very easily might be a bit off. Plus, I hope to understand the concepts of data and confidence ellipse right – I did my best to read up on them for a couple of days. Still, I think that the documentation needs clarification about what kind of ellipse ```stat_ellipse()``` actually produces.

Thanks,

Markus

```r
## This is to demonstrate that the ellipse drawn by ggplot2::stat_ellipse() does
## not correlate with sample size.

library(tidyverse)

for (my_sample_size in c(5, 20, 400, 1000)) {
  # Construct the 2x2 covariance matrix needed to draw a bivariate sample
  my_covariance_matrix = matrix(nrow = 2,
                                ncol = 2,
                                data = c(10^2,    0, # SD of parameter 1 is 10
                                            0, 10^2) # SD of parameter 2 is 10
                                # the 0's indicate that there is no correlation between parameters 1 and 2.
  )
  
  # Draw the bivariate sample
  my_sample =
    MASS::mvrnorm(n = my_sample_size,
                  mu = c(100, 100), # mean of parameters 1 and 2, respectively
                  Sigma = my_covariance_matrix) %>%
    as_tibble()
  

  # Plot the sample in a fixed coordinate system with stat_ellipse()
  print(
    ggplot(my_sample) + 
    ggtitle(paste("Sample size", my_sample_size)) +
    aes (y = V2, x = V1) +
    geom_point() +
    stat_ellipse() +
    coord_cartesian(xlim = c(50, 150), ylim = c(50, 150))
  )
}
```

leonardblaschek:
[...]

PR #3151, opened 2019-02-19
Issue 2776 updated stat_ellipse documentation.

llendway:
Changed documentation to reflect computing data ellipses rather than confidence ellipses. Fixes documentation portion of issue #2776

### DIFF
--- a/R/stat-ellipse.R
+++ b/R/stat-ellipse.R
@@ -1,7 +1,7 @@
-#' Compute normal confidence ellipses
+#' Compute normal data ellipses
 #'
 #' The method for calculating the ellipses has been modified from
-#' `car::ellipse` (Fox and Weisberg, 2011)
+#' `car::dataEllipse` (Fox and Weisberg, 2011)
 #'
 #' @references John Fox and Sanford Weisberg (2011). An \R Companion to
 #'   Applied Regression, Second Edition. Thousand Oaks CA: Sage. URL:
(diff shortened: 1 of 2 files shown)


thomasp85:
You are correct in that it is derived from `dataEllipse()` and not `ellipse()`, but I fear the title change will confuse users... Is data ellipses a well understood description?  cc @hadley

clauswilke:
I think the documentation needs a few sentences explaining what a data ellipse is/what this function does.

hadley:
I think the proposed title is confusing too. These draw confidence ellipses computed from data.

leonardblaschek:
> These draw confidence ellipses computed from data.

The trouble is that confidence ellipses are a rather tightly defined construct, and stat_ellipse does _not_ compute those. It computes data ellipses.

I agree that data ellipses are a somewhat obscure concept, and I too had to look it up when I stumbled over it (#2776). But I think keeping a misleading documentation just to not confuse people is a much worse idea. I regularly see high profile publications whose authors plot their PCAs in ggplot and annotate the data ellipses from stat_ellipse() as confidence ellipses, which is incorrect and misleading for everybody involved.

Ideally the underlying code of stat_ellipse() would be fixed, but for now this small change in documentation could prevent a lot of serious headaches.

clauswilke:
Closing this as a similar PR (#3410) was merged recently.

PR #3160, opened 2019-02-22
rlang/purrr style anonymous function specification in stat_function

### LINKED ISSUE #3159: Feature request: rlang/purrr-style anonymous functions in stat_function
It'd be nice to be able to specify __rlang__/__purrr__ style anonymous functions in `stat_function()`:
```r
df <- data.frame(x = 1:10, y = (1:10)^2)
ggplot(df, aes(x, y)) + 
  geom_point() +
  stat_function(fun = ~ .x^2)
```
Associated pull request forthcoming.

dkahle:
Closes #3159.


R/ggplot2.r
```diff
@@ -3,5 +3,5 @@
 
 #' @import scales grid gtable
 #' @importFrom stats setNames
-#' @importFrom rlang quo quos
+#' @importFrom rlang quo quos as_function
```
yutannihilation: As you use `rlang::`, you don't need to import this.

R/stat-function.r
```diff
@@ -5,7 +5,7 @@
 #' the x axis, and the results are drawn (by default) with a line.
 #'
 #' @eval rd_aesthetics("stat", "function")
-#' @param fun function to use. Must be vectorised.
+#' @param fun function to use or formula. Must be vectorised.
```
clauswilke: Please expand this sentence a bit more. Now that we have two different possible inputs, the documentation needs a little more care. Also, please add a simple example to the examples, or maybe better rewrite one or two of the current ones (so we don't increase the total number of examples).

clauswilke: The `.x` in the formula should also be explained here.

tests/testthat/test-stats-function.r
```diff
@@ -34,3 +34,16 @@ test_that("works with discrete x", {
...
   expect_equal(ret$y, 1:2)
 })
+
+test_that("works with formula syntax", {
+  dat <- data_frame(x = 1:10)
+
+  base <- ggplot(dat, aes(x, group = 1)) +
+    stat_function(fun = ~ .x^2, geom = "point", n = 5) +
+    scale_x_continuous(limits = c(0, 10))
+  (ret <- layer_data(base))
```
clauswilke: Is there a reason this assignment is in parentheses?

R/stat-function.r
```diff
@@ -49,6 +49,7 @@
 #' # Using a custom function
 #' test <- function(x) {x ^ 2 + x + 20}
 #' f + stat_function(fun = test)
+#' f + stat_function(fun = ~ .x^2 + .x + 20)
```
yutannihilation: Probably, `.` is better for single-argument function?

R/stat-function.r
yutannihilation: And we also need a link to `as_function`'s doc.

tests/testthat/test-stats-function.r
dkahle: No, accidental leave-in from testing. :)
```suggestion
  ret <- layer_data(base)
```

R/ggplot2.r
dkahle: Won't we need it somewhere to avoid CRAN warnings about not being imported? Otherwise, I'm fine removing of course.

R/stat-function.r
dkahle: I thought about that, too: `.x` seems to be more tidyverse consistent in that it's more referentially transparent in a piping context, but piping is not native to __ggplot2__.  I don't have a strong preference either way.

R/stat-function.r
dkahle: Happy to update the docs for the argument and update the examples more fully, if you'd like, I was just trying to be as minimalist as possible. Want me to use `.` or `.x` for the formula stuff? (@yutannihilation brought the `.` vs `.x` part up in another comment, too.)

R/stat-function.r
clauswilke: I have a slight preference for `.x`, because I think it imposes slightly lower mental load, in particular for complex formulas. Compare:
``` r
fun = ~ (5*.^2 + 3*.) * exp(-2*.^2 + 3*. + 2)
fun = ~ (5*.x^2 + 3*.x) * exp(-2*.x^2 + 3*.x + 2)
```

yutannihilation: OK, I prefer `.` because it's short, but it's just my preference and I don't have strong opinion. Let's use `.x`.

(Note that in some places e.g. `sec_axis()` `.x` is not allowed, but it's just implementational details.)

dkahle:
I think I've included the updates in some new commits (dkahle/ggplot2/tree/purrr-style-stat-function@0e1a56f and dkahle/ggplot2/tree/purrr-style-stat-function@612b6d1), how can I add those to this PR? Or, do I need to submit a new PR with all of them?

R/ggplot2.r
clauswilke: I don't think so. I suggest removing. If it's a problem the Travis build will break.

R/stat-function.r
```diff
@@ -16,39 +19,44 @@
 #'   \item{x}{x's along a grid}
 #'   \item{y}{value of function evaluated at corresponding x}
 #' }
+#' @seealso \code{\link[rlang:as_function]{as_function()}}
```
clauswilke: Should this be `rlang::as_function`? (two colons instead of one)

clauswilke:
The new commits get added automatically.

Note that the travis build fails with R devel. Not exactly sure from a quick glance what the problem is. Could you check if you can identify the issue?

R/ggplot2.r
yutannihilation: What do you mean? As long as fully qualified form (`rlang::as_function()`), you don't need to import.

R/stat-function.r
dkahle: I thought in the link you put a single colon; works in the docs as I built them, at least.

clauswilke:
One more issue from my side: Maybe remove 2 or 3 of the examples. Examples are somewhat expensive, because they all get executed on every CRAN check. The current version has 7 plots, which is already a lot.

dkahle:
Cleaned the docs a bit. Looking into the R-devel issue. It's been flagging on me in some other projects, too, so I'm not sure what's going on there.

R/stat-function.r
yutannihilation: A single colon is correct.

c.f. https://cran.r-project.org/doc/manuals/R-exts.html#Cross_002dreferences

yutannihilation:
It seems R devel is currently broken; I believe #3161 is a harmless change, but it also fails only on r-devel.

https://travis-ci.org/tidyverse/ggplot2/jobs/497352631

We might need to look into it, but I think it's irrelevant to this PR.

R/stat-function.r
yutannihilation: Sorry, I was not accurate above. A single colon is correct as a raw `Rd` code, but we should use Markdown-style as we do so in other places: `[rlang::as_function]`.

yutannihilation:
The R-devel issue was solved by #3163.

R/stat-function.r
dkahle: Looks right with parentheses. (I'm seeing more like `[rlang::as_function()]`.) Got it.

yutannihilation: Yes, `[rlang::as_function()]` is correct, sorry. More details are here: https://cran.r-project.org/web/packages/roxygen2/vignettes/markdown.html#links

yutannihilation:
@dkahle Could you remove `as_function` from `@importFrom`? I feel this is ready to merge if you fix it.

yutannihilation:
@clauswilke I'm merging this. The number of examples might still needs to be reduced, but I think we can restructure the examples when ggplot2 gets ready to promote `~` notations.

yutannihilation:
Thanks!

clauswilke:
@yutannihilation Ok. Maybe open an issue about this so we don't forget?

yutannihilation:
Good idea, thanks. Here it is: https://github.com/tidyverse/ggplot2/issues/3181

PR #3161, opened 2019-02-23
Use the fully qualified forms for quos()

yutannihilation:
Though `quos()` are imported in `utilities-tidy-eval.R`, I think it's good practice to stick with the fully qualified forms in package code (I'm tolerant for test code). There are a few places that lacks `rlang::`, so this PR added them.

In addition, I want to remove this line to make it clear that we don't encourange the unqualified forms of `quo()` and `quos()`. I guess this was added by mistake.

https://github.com/tidyverse/ggplot2/blob/033fb521a084447c15178c591ab96a158f3b1e1e/R/ggplot2.r#L6

### DIFF
--- a/R/facet-grid-.r
+++ b/R/facet-grid-.r
@@ -150,7 +150,7 @@ facet_grid <- function(rows = NULL, cols = NULL, scales = "fixed",
     stop("A grid facet specification can't have more than two dimensions", call. = FALSE)
   }
   if (n == 1L) {
-    rows <- quos()
+    rows <- rlang::quos()
     cols <- facets_list[[1]]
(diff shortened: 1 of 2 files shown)


yutannihilation:
@lionel- Thanks, I'm merging this for consistency this time. Do you think ggplot2 should also import the whole rlang package? If it's a common practice among tidyverse, ggplot2 should follow, and using unqualified form seems a bit better in terms of the speed.

lionel-:
I don't have a strong opinion about this. Maybe best discussed with the regular contributors?

yutannihilation:
Sure.

Issue #3062, opened 2019-01-09
Consider adding "static" parameter to geom/stat functions


thomasp85:
The current approach to repeating a layer across panels is to not have the layer data contain the variable needed for the faceting. This is an approach also implemented in gganimate when it comes to having layers be static across the animation. While this works intuitively, I feel it often requires some additional dataprep before the plotting code, and sometimes require that layers that otherwise use the same data need to target separate almost identical datasets.

Would it make sense to add a `static` or perhaps `repeat` argument to geom and stat functions to explicitly mark them for being repeated across panels (and frames in the case of gganimate)?

Example API:

```r
library(dplyr)
library(ggplot2)

# Current approach
diamonds_static <- mutate(diamonds, cut = NULL)
ggplot(diamonds, aes(x = color)) + 
  geom_bar(data = diamonds_static, fill = 'grey70') +
  geom_bar() + 
  facet_wrap(~cut)

# New approach
ggplot(diamonds, aes(x = color)) + 
  geom_bar(fill = 'grey70', static = TRUE) +
  geom_bar() + 
  facet_wrap(~cut)
```
![image](https://user-images.githubusercontent.com/1775316/50890266-25e49580-13fa-11e9-8528-582e451a7704.png)

The "problem" with the current approach is that it requires changes to the data source if we decide to change the faceting variable — not a huge problem, but still a barrier to experimentation.

If the `static` name is too close to the idea of animation, then we can figure out another name for it...

clauswilke:
I think this is a good idea. I very much adhere to the philosophy that we should optimize the API so we don't usually have to provide the dataset more than once. This is in line with @yutannihilation's gghighlight and my ideas for in-layer sampling in the ungeviz package.

I'm just not fully convinced about `static`.  A good name would make sense in the context of facets first. I can't think of anything better, though.

yutannihilation:
I couldn't agree this more! Currently, gghighlight needs ugly colnames (you cannot always remove the column as it might be mapped to a necessary aes) to prevent the unhighlighted data from being facetted:

https://github.com/yutannihilation/gghighlight/blob/8dc648cedad2b0f6137a4440b9973e3f9d4c24a0/R/gghighlight.R#L275-L291

thomasp85:
completely open to another name... too deep in animation at the moment to be able to think of something better though (except for `repeat` which I don't particularly like)

smouksassi:
[...]

PR #3162, opened 2019-02-23
allow empty facet specs

### LINKED ISSUE #2986: Should we compact facet grid specs?
This would make it easier to program with facets.

From https://stackoverflow.com/questions/53218152/ggplot2-facet-grid-with-conditional-facets-and-tidy-evaluation

```r
plot_iris <- function(df_in, facet_var = NULL){
  ggplot(df_in) +
    geom_point(aes(Sepal.Length, Sepal.Width)) +
    facet_grid(vars(!!enquo(facet_var)), vars(idx))
}
```

### LINKED ISSUE #3070: Should facet + empty spec be a noop?
Should these be equivalent to `ggplot(mtcars)` instead of failing?

```r
ggplot(mtcars) + facet_wrap(vars())
#> Error: Can't subset with `[` using an object of class NULL.

ggplot(mtcars) + facet_wrap(NULL)
#> Error: Must specify at least one variable to facet by
```

Allowing empty specs would make sense in functions:

```r
my_plot <- function(.data, .mapping, ...) {
  ggplot(.data, .mapping) +
    geom_point() +
    facet_wrap(vars(...))
}
```

Then both of these would work, instead of just the latter:

```r
my_plot(starwars, aes(mass, height / 100))

my_plot(starwars, aes(mass, height / 100), gender)
```

yutannihilation:
Fix #3070, fix #2986.

### Summary

This PR compacts facet specs automatically (#2986) and allow empty specs (#3070).

I admit this makes the implementation of `FacetGrid` and `FacetWrap` a bit complicated in that it needs some special treatments for the case of 0-variable. But, on the other hand, accepting empty specs makes the process of building facet specs simpler and clearer. So, I believe this change is worth making.

### Major changes (visible to users)

* Compact facet specs automatically.
    * If the input is `NULL`, treat it as an empty quosures object (`quos()`).
    * If the input is a quosures object, remove `NULL`s in it. 
* Allow empty specs. When the spec is empty,
    * `facet_wrap()` generates a single plot with a strip. A string `(all)` is used for the strip label.
    * `facet_grid()` generates a single plot without a strip.   

### Internal changes

To make the major changes, I want to define the internal facet specs more strictly as following:

* `FacetWrap` requires **a quosures object** (e.g. `quos(foo, bar)`)
* `FacetGrid` requires **a list of two quosures objects, `rows` and `cols`** (e.g. `list(rows = quos(foo), cols = quos(bar))`)

Accordingly, the corresponding functions below return these specs. 

* `grid_as_facet_list()`
* `wrap_as_facet_list()` (new)

Under these functions, `as_facet_list()` is responsible for translating a spec (e.g. a quosures object, a formula, and a character) to a list of quosures. It's also responsible for the basic validation (e.g. raise an error when the spec is an `uneval` object), so the upper functions should always use this for the input even it's just a simple quosures object.

Note that the results of `as_facet_list()` are not compacted; compacting these specs is basically the role of the upper functions, `grid_as_facet_list()` and `wrap_as_facet_list()`. For convenience, a new internal function `compact_facets()` can be used as a short cut of flattening and compacting the spec.

### Minor changes

* `facet_wrap()` and `facet_grid()` now properly rejects `aes()`

### Example usages

``` r
library(ggplot2)

d <- data.frame(x = 1:10, g = rep(c("a", "b"), each = 5))
p <- ggplot(d, aes(x, x)) + geom_point()

# NULLs are compacted automatically
p + facet_wrap(vars(NULL, g, NULL))
```

``` r

# accepts empty specs
p + facet_wrap(NULL)
```

``` r

# accepts empty specs
p + facet_grid(NULL, NULL)
```

<sup>Created on 2019-03-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


clauswilke:
I think this is good in principle. Not entirely sure about the `(empty)` label, though. Would it be possible to just use an empty string?

yutannihilation:
Oh, nice idea. Thanks. But, we cannot use empty names for variable names, so it seems I need some tweaks.

``` r
list("" = "")
#> Error: attempt to use zero-length variable name
```

clauswilke:
And the label and the variable name should be the same, I assume.

So maybe `(empty)` is fine, or alternatively how about `(all data)`. After all, faceting usually indicates subsets of data, and here the subset is everything.

yutannihilation:
I forgot `"(all)"` is used when `facet_grid(margins = TRUE)`. Probably we should use this label in `facet_wrap()` too, (and in `facet_grid()` too...?).

clauswilke:
Sounds good to me.

yutannihilation:
Two small corrections:

> we cannot use an empty string as a variable name

We can, and it seems to work with the current code of ggplot2.

``` r
l <- setNames(list(""), "")
names(l)
#> [1] ""
```

<sup>Created on 2019-02-25 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

> `"(all)"` is used

`(all)` label is not coded in ggplot2, but is returned by `reshape2::add_margins()`.


Anyway I'll use `(all)` since it would be slightly look nicer when `labeller = label_both`.

yutannihilation:
Is this a typo of `"uneval"`...? 🤔 

https://github.com/tidyverse/ggplot2/blob/9df0a076025aa14ec8cf079ae53e0d662884625a/R/facet-.r#L278-L280

yutannihilation:
@lionel- @thomasp85 I think the proposal about empty facet specs is almost fixed, though I need to add more test cases before asking for revewing. I'd appreciate if you give some feedbacks and concerns :)

tests/testthat/test-facet-.r
```diff
@@ -36,17 +46,32 @@ test_that("as_facets_list() coerces lists", {
   expect_identical(out, exp)
 })
 
-test_that("as_facets_list() errors with empty specs", {
-  expect_error(as_facets_list(list()), "at least one variable to facet by")
-  expect_error(as_facets_list(. ~ .), "at least one variable to facet by")
-  expect_error(as_facets_list(list(. ~ .)), "at least one variable to facet by")
-  expect_error(as_facets_list(list(NULL)), "at least one variable to facet by")
+test_that("wrap_as_facets_list() and grid_as_facets_list() accept empty specs", {
+  expect_equal(wrap_as_facets_list(NULL), quos())
```
lionel-: I usually use `expect_identical()` for checking quosures, this way the environment is checked as well. Can be a little tricky though sometimes.

R/facet-grid-.r
```diff
@@ -145,54 +145,46 @@ facet_grid <- function(rows = NULL, cols = NULL, scales = "fixed",
...
   ggproto(NULL, FacetGrid,
     shrink = shrink,
-    params = list(rows = rows, cols = cols, margins = margins,
+    params = list(rows = facets_list$rows, cols = facets_list$cols, margins = margins,
       free = free, space_free = space_free, labeller = labeller,
       as.table = as.table, switch = switch, drop = drop)
   )
 }
+
+# returns a list containing exactly two quosures `rows` and `cols`
```
lionel-: Here it's not clear whether you're mentioning two single quosure or two quosure lists. I would use "quosure lists" or "lists of quosures" instead of just the plural "quosures".

Also we generally capitalise comments. No full stop, unless there are more than one sentence.

yutannihilation: Thanks! Actually I'm struggling about the wording here...

> Also we generally capitalise comments. No full stop, unless there are more than one sentence.

 I see.

tests/testthat/test-facet-.r
yutannihilation: I got it.

yutannihilation:
Note that, I gave up testing strip labels (`"(all)"`) because it seems there's no other way than visual tests, but I feel it's not worth for the cost.

R/facet-.r
```diff
@@ -311,13 +311,16 @@ as_facets_list <- function(x) {
...
-  }
-
   x
 }
 
+# Flatten a list of quosures objects to a quosures object, and compact it
+compact_facets <- function(x) {
+  x <- rlang::flatten_if(x, rlang::is_list)
+  null <- vapply(x, rlang::quo_is_null, logical(1))
+  rlang::as_quosures(x[!null])
```
lionel-: I think this should be `new_quosures()`. The casting form converts formulas to quosures etc, and at this point we should have standardised the inputs already.

yutannihilation: Makes sense. Thanks!

yutannihilation:
@thomasp85 Do you have any comments? Since this PR is mostly about facet specs and I don't think the change is big, I expect this doesn't get in the way of #3062. But, if you have any concerns, please tell me. If there's no major concern, I'll merge.

yutannihilation:
Will merge this after the release of 3.2.

yutannihilation:
Thanks!

Issue #579, opened 2012-06-14
Using scale_fill_continuous(breaks=NULL) results in error


wch:
Some of the visual tests in visual_test/scale-breaks.r now throw errors. This was introduced by e6594a11b5cec0941a0c8c15b5c31ff41ba22be2, where the colorbar guide started being used as the default.

Here are the two bad tests:

``` R
dat <- data.frame(x=1:3, y=1:3)

ggplot(dat, aes(x=1, y=y, fill=x)) + geom_point(shape=21) + scale_fill_continuous(breaks=NULL)

ggplot(dat, aes(x=1, y=y, colour=x)) + geom_point() + scale_colour_continuous(breaks=NULL)
```

We should also have some unit tests for this...

PR #3163, opened 2019-02-24
Fix Travis failures on r-devel

yutannihilation:
It seems the failure on r-devel is related to `R_CHECK_LENGTH_1_LOGIC2`, an option to raise error when multiple logicals are passed to `||` or `&&`.

```
 --- failure: length > 1 in coercion to logical ---
```

I'm not sure the next version of R will be released with this strict behaviour, but it's a nice chance to fix inappropriate if conditions.

### DIFF
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -244,7 +244,7 @@ guide_geom.colorbar <- function(guide, layers, default_mapping) {
   guide_layers <- lapply(layers, function(layer) {
     matched <- matched_aes(layer, guide, default_mapping)
 
-    if (length(matched) && (isTRUE(is.na(layer$show.legend)) || isTRUE(layer$show.legend))) {
+    if (length(matched) > 0 && (isTRUE(is.na(layer$show.legend)) || isTRUE(layer$show.legend))) {
       layer
     } else {
       # This layer does not use this guide
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -244,7 +244,7 @@ guide_geom.colorbar <- function(guide, layers, default_mapping) {
   guide_layers <- lapply(layers, function(layer) {
     matched <- matched_aes(layer, guide, default_mapping)
 
-    if (length(matched) && ((is.na(layer$show.legend) || layer$show.legend))) {
+    if (length(matched) && (isTRUE(is.na(layer$show.legend) || isTRUE(layer$show.legend)))) {
       layer
     } else {
       # This layer does not use this guide
--- a/R/guide-colorbar.r
+++ b/R/guide-colorbar.r
@@ -244,7 +244,7 @@ guide_geom.colorbar <- function(guide, layers, default_mapping) {
   guide_layers <- lapply(layers, function(layer) {
     matched <- matched_aes(layer, guide, default_mapping)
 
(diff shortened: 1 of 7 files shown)


clauswilke:
I believe `isFALSE()` is not the same as `!isTRUE()` and it’s best to avoid it. Just write `!isTRUE()`.

yutannihilation:
I see. One of the two `isFALSE()` was a mistake. The other one in `R/guides-.r` is the right usage, I believe.

R/guide-colorbar.r
```diff
@@ -244,7 +244,7 @@ guide_geom.colorbar <- function(guide, layers, default_mapping) {
   guide_layers <- lapply(layers, function(layer) {
     matched <- matched_aes(layer, guide, default_mapping)
 
-    if (length(matched) && ((is.na(layer$show.legend) || layer$show.legend))) {
+    if (length(matched) && (isTRUE(is.na(layer$show.legend)) || isTRUE(layer$show.legend))) {
```
clauswilke: Should `length(matched)` get an explicit comparison, like `length(matched) > 0` (I assume that's what is meant)?

R/scale-.r
```diff
@@ -563,7 +563,7 @@ continuous_scale <- function(aesthetics, scale_name, palette, name = waiver(),
 
   position <- match.arg(position, c("left", "right", "top", "bottom"))
 
-  if (is.null(breaks) && !is_position_aes(aesthetics) && guide != "none") {
+  if (is.null(breaks) && all(is_position_aes(aesthetics)) && identical(guide, "none")) {
```
clauswilke: I'm confused here. Doesn't this invert the logic for the last two checks? Wouldn't they need `!` in front?

R/scale-.r
```diff
@@ -634,7 +634,7 @@ discrete_scale <- function(aesthetics, scale_name, palette, name = waiver(),
...
```
clauswilke: Same as previous. Inverted logic?

R/utilities.r
```diff
@@ -475,3 +475,6 @@ parse_safe <- function(text) {
   }
   out
 }
+
+# For older version of R
+isFALSE <- function(x) is.logical(x) && length(x) == 1L && !is.na(x) && !x
```
clauswilke: Should this be in [backports.R](https://github.com/tidyverse/ggplot2/blob/master/R/backports.R) and have a version check?

R/scale-.r
yutannihilation: Opps..., sorry, I'm also confused because the previous logic seems unclear to me about its intension. Let me think. (I believe `all(is_position_aes(aesthetics))` doesn't need `!`; if there are only positional aesthetics, we don't need guides, right?)

R/guide-colorbar.r
yutannihilation: Agreed, thanks.

R/utilities.r
yutannihilation: Sure, it seems the right place.

R/scale-.r
yutannihilation: OK, I got it. #581 and #579 are the context; `breaks = NULL` can be used for removing the guide of a non-positional scales.

yutannihilation:
Thanks for the prompt review!

Issue #2718, opened 2018-06-26
Ploting POINT sf with geom_sf() is slow


yutannihilation:
In my environment, it takes 10 seconds for plotting 1000 `POINT`s by `geom_sf()`.
I feel this is a bit too long for exploring the data.

https://gist.github.com/yutannihilation/479cb5f254826915a7a3d36ff84b4b43

I suspect this is because `draw_panel()` applies `sf_as_grob()` on every row. Is it possible to do this more efficiently, e.g. `st_combine()` per `group` so that it plots one `MULTIPOINT` that contains 1000 points?

clauswilke:
Could you check whether there is any relation to issue #2655?

yutannihilation:
Thanks, I roughly think this is a different issue because I'm using Windows (sorry I forgot to write this). But, I will check further.

yutannihilation:
Here's the result of profvis. This shows `st_grob()` takes most of the time.

![image](https://user-images.githubusercontent.com/1978793/42005379-0251cf32-7aaf-11e8-8ec3-11b7846c4a3e.png)

yutannihilation:
Here's a reprex. This shows `st_combine()` significantly save time.

``` r
library(ggplot2)
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.2.3, proj.4 4.9.3

set.seed(1)
n <- 1000

d <- data.frame(
  x = runif(n) * 360,
  y = runif(n) * 180
)

d_sf <- st_multipoint(as.matrix(d)) %>%
  st_sfc() %>% 
  st_cast("POINT")

system.time({
  p <- ggplot(d_sf) +
    geom_sf()
  print(p)
})
```

![](https://i.imgur.com/JweUbKx.png)

    #>    user  system elapsed 
    #>    7.36    0.10    7.71
    
    
    system.time({
      d_sf_combined <- st_combine(d_sf)
      p <- ggplot(d_sf_combined) +
        geom_sf()
      print(p)
    })

![](https://i.imgur.com/grqLLie.png)

    #>    user  system elapsed 
    #>    0.27    0.05    0.31

Created on 2018-06-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).
<details><summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ──────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.5.0 (2018-04-23)
#>  os       Windows 10 x64              
[...]

PR #3164, opened 2019-02-25
Better sf support

thomasp85:
This PR adapts `geom_sf` and `coord_sf` to make use of the improvements in performance introduced in sf v0.7-3. The changes are as follows:

- Use `st_normalize` during rescaling to greatly speed up `coord_sf
- Construct a single grob from the geometry and let `sf` decide on how/if to split it up to multiple grobs

This PR introduces a new version dependency on sf as it does not make much sense to support both implementations simultaneously

### DIFF
--- a/R/geom-sf.R
+++ b/R/geom-sf.R
@@ -134,8 +134,8 @@ sf_grob <- function(x, lineend = "butt", linejoin = "round", linemitre = 10) {
   # Need to extract geometry out of corresponding list column
   geometry <- x$geometry
   type <- sf_types[sf::st_geometry_type(geometry)]
-  is_point <- type %in% 'point'
-  type_ind <- match(type, c('point', 'line', 'other'))
+  is_point <- type %in% "point"
+  type_ind <- match(type, c("point", "line", "other"))
   defaults <- list(
     GeomPoint$default_aes,
     GeomLine$default_aes,
--- a/R/geom-sf.R
+++ b/R/geom-sf.R
@@ -105,15 +105,7 @@ GeomSf <- ggproto("GeomSf", Geom,
 
     # Need to refactor this to generate one grob per geometry type
     coord <- coord$transform(data, panel_params)
-    grobs <- lapply(1:nrow(data), function(i) {
-      sf_grob(
-        coord[i, , drop = FALSE],
-        lineend = lineend,
-        linejoin = linejoin,
-        linemitre = linemitre
-      )
-    })
-    do.call("gList", grobs)
+    sf_grob(coord, lineend = lineend, linejoin = linejoin, linemitre = linemitre)
   },
 
   draw_key = function(data, params, size) {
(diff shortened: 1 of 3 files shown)


yutannihilation:
This fixes #2718 at last. Thanks Thomas!

clauswilke:
This looks good to me in principle, but I don't have time this week to go over it in detail (proofing my book this week, and likely reformatting many of the figures).

I think increasing the required sf version is good anyways, because otherwise axis ticks may be placed in the wrong location due to a pernicious bug that was only recently fixed.

R/geom-sf.R
```diff
@@ -138,33 +130,37 @@ default_aesthetics <- function(type) {
...
-      lty = row$linetype,
-      lineend = lineend,
-      linejoin = linejoin,
-      linemitre = linemitre
-    )
-    sf::st_as_grob(geometry, gp = gp)
-  }
+  geometry <- x$geometry
+  type <- sf_types[sf::st_geometry_type(geometry)]
+  is_point <- type %in% 'point'
```
yutannihilation: Use `"`, not `'`, as the style guide says: https://style.tidyverse.org/syntax.html#quotes

R/geom-sf.R
```diff
@@ -282,3 +278,11 @@ geom_sf_text <- function(mapping = aes(), data = NULL,
     layer_class = LayerSf
   )
 }
+
+sf_types <- c(GEOMETRY = "other", POINT = "point", LINESTRING = "line",
```
yutannihilation: IIUC, the types of simple feature supported by sf can be different in future. Can we just list up geometries of `point` and `line`, and treat others as `others`? (Not sure it works for unknown types, though)

R/geom-sf.R
```diff
@@ -138,33 +130,37 @@ default_aesthetics <- function(type) {
...
-      linejoin = linejoin,
-      linemitre = linemitre
-    )
-    sf::st_as_grob(geometry, gp = gp)
-  }
+  geometry <- x$geometry
+  type <- sf_types[sf::st_geometry_type(geometry)]
+  is_point <- type %in% 'point'
+  type_ind <- match(type, c('point', 'line', 'other'))
+  defaults <- list(
```
hadley: We will need to re-think this at some point in order to expose to the user in the same way as other geoms. But we don't need to do that in this PR.

R/geom-sf.R
thomasp85: Oops - nice catch. My life is a constant struggle between my own preference and Hadleys ;-)

R/geom-sf.R
thomasp85: This was made in order to get a quick look-up. Most of these are not really supported by sf but just part of the simple feature spec. I'm quite sure sf won't begin to support something other than that in the next 10 years

yutannihilation: OK, it's a good reason. Let's talk about this again in 2029 :)

PR #3168, opened 2019-02-28
Extend the range of recognised positional aesthetics

thomasp85:
This PR extends the recognised positional aesthetics to include `x0` and `y0`. This is needed for a range of geoms/stats in ggforce to make sure they play nice with scale transformations. `x0` and `y0` are traditionally used to denote center values and a natural fit for many parametrised geometries...

Such a change has happened before in c9f276b60c so it should be fairly safe

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -62,6 +62,9 @@
 * `geom_polygon()` can now draw polygons with holes using the new `subgroup` 
   aesthetic. This functionality requires R 3.6 (@thomasp85, #3128)
 
+* `x0` and `y0` is now recognized positional aesthetics so they will get scaled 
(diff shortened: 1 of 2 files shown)


yutannihilation:
By the way, these (what aesthetics are treated as positional) are the official spec of ggplot2? If so, I think it should be described in the docs (and I'll do it in a seperate PR).

thomasp85:
You can see the list of recognised position aesthetics in the construction of the continuous position scales...

yutannihilation:
Yeah, but I still think it's nice to have it documented (and probably needs tests), though I have no idea where should those developer-only documentaton should go (Extending ggplot2?)...

Issue #3116, opened 2019-02-03
Feature request: allow access to the trained scales from the `layout` object


paleolimbot:
This is a fairly easy (I think) feature to add that would help me implement lazy reading of huge raster files (a stars proxy object), such that they can be downsampled appropriately. To do this, I would need access to the trained scales (for the `fill` and `alpha` aesthetics at draw time. Right now there is no way to access the trained scales except limited information about them in the `panel_params` and `coordinates` objects. I think that a reference to the `ScalesList` object could be included as an element of the `layout` object without any consequences? The `Geom` has access to the `layout` at draw time.

It is a completely valid point that this would be pushing ggplot2 where it perhaps should not go, coercing it to do things that it was not meant to do. I would argue that it is necessary if spatial rasters are ever going to be plotted in ggplot by those of us who don't have an intimate knowledge of downsampling (and maybe they shouldn't be). There is a brief discussion on the implementation of this at paleolimbot/ggspatial#17 . A reprex is included below.

``` r
library(ggplot2)
library(grid)

# in real life, this would be a reference to a large (scary) raster file
big_scary_raster <- tibble::tibble(raster = list(matrix(1:9, nrow = 3)))
```

``` r

StatMatrixList <- ggproto(
  "StatMatrixList",
  Stat,

  required_aes = "raster",
  default_aes = ggplot2::aes(fill = stat(z)),

  compute_layer = function(self, data, params, layout) {
    data$raster <- lapply(data$raster, function(x) {
      df <- reshape2::melt(x)
      names(df) <- c("x", "y", "z")
      df
    })

    tidyr::unnest(data, .data$raster)
  }
)

ggplot(big_scary_raster, aes(raster = raster)) +
  geom_raster(stat = StatMatrixList, hjust = 0, vjust = 0)
```

![](https://i.imgur.com/DWvYltD.png)

``` r

StatLazyMatrixList <- ggproto(
  "StatLazyMatrixList",
  StatMatrixList,

  compute_layer = function(self, data, params, layout) {
    # only return limits in the stat (these are usually cached in the raster file,
    # so the raster doesn't need to be loaded). Scales get trained based on the
    # result of this function.
    data$limits <- lapply(data$raster, function(raster) {
      tibble::tibble(
        x = c(0, ncol(raster)),
        y = c(0, nrow(raster)),
        z = range(raster)
      )
    })

    tidyr::unnest(data, .data$limits, .drop = FALSE)
  }
)

[...]

PR #3170, opened 2019-03-01
Add access to the scales at draw time using a new `layer_params` object

paleolimbot:
This is a PR addressing #3116, and allows access to the trained scales at draw time through a new `layer_params` argument to `Geom$draw_layer()`. This enables geometries that may require extensive loading of data to be drawn lazily (among other things). The overall structure of using a new argument to `Geom$draw_layer()` was discussed at length in the issue, but the details of the implementation and testing could definitely use the keen eye of those more familiar with the ggplot2 internals. Cheers!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 3.1.0.9000
 
+* Scales are now accessible to the `Geom` at draw time through a new 
+  `layer_params` argument to `Geom$draw_layer()` (#3116).
+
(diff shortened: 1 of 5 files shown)


clauswilke:
@hadley Could you take a look at the way this PR uses ggproto to generate a `layer_params` object (lines 210 to 220 in layer.r) and let us know if this is how you envision ggproto to be used? The alternative would be a list that contains the `get_scale()` function.

clauswilke:
This is temporarily on hold pending further discussion on issue #3175.

Issue #2828, opened 2018-08-14
Improve binwidth document in geom_histogram()


hadley:
Should mention that `binwidth()` can be a function, how often the function is called (once for each group?), and what the input is (the unscaled x values?)

clauswilke:
This issue was addressed with #3172 but not closed. Closing now.

PR #3172, opened 2019-03-04
Attempt to improve binwidth document in geom_histogram()

ppanko:
Hello! This pull request is attempting to address a change to the documentation of `geom_histogram` as laid out in #2828.  Not entirely certain about my use of nomenclature here; please advise if there are issues.


yutannihilation:
Thanks for taking this. This seems better than the status quo, but it doesn't seem very understandable to me yet, Maybe it needs some example?

By the way, could you generate the document by running `devtools::document()`? If you don't know how to do, I'll take care of it.

ppanko:
Whoops, I forgot to re-generate the .Rd files - sorry about that. The new commit should change that.

ppanko:
There is already an example of this behavior at the bottom of the `geom_histogram` documentation. Would you like me to add another, or perhaps to change wording in the help text?

yutannihilation:
Ah, I forgot the example! Sorry for my carelessness... Then it seems fine.

It might be good to describe explicitly here that the function is applied per facet. What do you think?
https://github.com/tidyverse/ggplot2/blob/6be789309d3972c6d5bd6963c9a4bb5cbd5b75d1/R/geom-histogram.r#L81-L86

ppanko:
Great point! I added some text to reflect this.

yutannihilation:
Thanks, looks good to me.

@clauswilke 
Could you take a look? I think I'm not good at reviewing documentations.

R/stat-bin.r
```diff
@@ -1,5 +1,7 @@
 #' @param binwidth The width of the bins. Can be specified as a numeric value,
-#'   or a function that calculates width from x.
+#'   or a function that calculates width from unscaled x. When specifying a
```
clauswilke: The comma at the end of the preceding line (after "value") should be removed, and it would be slightly better to write "or *as* a function ..."

clauswilke: Also, thinking about it some more, I'm not entirely sure what "unscaled x" means. Does it mean before any scale transformations have been applied? Maybe better to clarify.

ppanko: Thank you for taking the time to review. I solved the former issue in the most recent commit, but I want to make certain that the "unscaled x" portion is unambiguous before I go further. Would the text below be more clear? 

"or as a function that calculates width from x prior to any x-axis transformations."

clauswilke: I think the text will read better if you leave "unscaled x" and then define what that means:
> ...unscaled x. Here, "unscaled x" refers to the original x values in the data, before application of any scale transformation.

ppanko: Wonderful! I will do exactly that.

yutannihilation:
Thanks!

PR #3177, opened 2019-03-06
prevent margin from changing the unit class when subclassing

thomasp85:
I've uncovered another "bad" assumption in ggplot2's handling of units. The margin constructor will drop any unit subclasses when subclassing the unit it creates. This PR fixes it and makes it work with the experimental unit implementation in https://github.com/thomasp85/grid/pull/1 in a back-compatible way

### DIFF
--- a/R/margins.R
+++ b/R/margins.R
@@ -4,7 +4,9 @@
 #' @rdname element
 #' @export
 margin <- function(t = 0, r = 0, b = 0, l = 0, unit = "pt") {
-  structure(unit(c(t, r, b, l), unit), class = c("margin", "unit"))
+  u <- unit(c(t, r, b, l), unit)
+  class(u) <- c("margin", class(u))
+  u
 }
 is.margin <- function(x) {
   inherits(x, "margin")


yutannihilation:
Travis says (https://api.travis-ci.org/v3/job/502563012/log.txt):

```
* checking for future file timestamps ... WARNING
unable to verify current time
```

This error comes from [here](https://github.com/wch/r-source/blob/9a65c8381f86db2215fb2e76af402aef25ed135b/src/library/tools/R/check.R#L439-L447):

```r
                now <- tryCatch({
                    foo <- readLines("http://worldclockapi.com/api/json/utc/now",
                                     warn = FALSE)
                    as.POSIXct(gsub(".*\"currentDateTime\":\"([^Z]*).*", "\\1", foo),
                               "UTC", "%Y-%m-%dT%H:%M")
                }, error = function(e) NA)
                if (is.na(now)) {
                    any <- TRUE
                    warningLog(Log, "unable to verify current time")
```

But this seems unlikely to success... Probably we can ignore.

``` r
readLines("http://worldclockapi.com/api/json/utc/now",
                                     warn = FALSE)
#> Warning in file(con, "r"): cannot open URL 'http://worldclockapi.com/api/
#> json/utc/now': HTTP status was '403 Forbidden'
#> Error in file(con, "r"): cannot open the connection to 'http://worldclockapi.com/api/json/utc/now'
```

<sup>Created on 2019-03-07 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

yutannihilation:
I restarted the job. Let's see what will happen. If this consistently fails, it seems better to set `_R_CHECK_SYSTEM_CLOCK_` to false. 

c.f. https://twitter.com/krlmlr/status/1084496032187199488

thomasp85:
@yutannihilation this is due to upstream problems, not something in ggplot2

yutannihilation:
@thomasp85 Yeah, could you approve #3179?

PR #3192, opened 2019-03-18
Skip picking a scale when the layer data is Inf

### LINKED ISSUE #3184: when using annotation as first layer with x discrete and annotation xmin/xmax = -Inf/Inf error
Hi developpers ! 
I posted first on Rstudio Community 
https://community.rstudio.com/t/understanding-when-and-why-discrete-value-supplied-to-continuous-scale-happens/26011/5 
but then it was thought of as a potential bug that is maybe related to 
https://github.com/tidyverse/ggplot2/issues/3120
Basically when we want to annotate a plot with discrete x scale edge to edge it errors out unless we have a fake first geom blank
when we specify the levels as xmin or xmax the order get changed  
I am on ggplot2_3.1.0.9000
``` r
require(ggplot2)
#> Loading required package: ggplot2
# geom blank make the scales wider than what we want but it works

ggplot(diamonds, aes(x = cut, y = price)) +
  geom_blank()+
  annotate("rect",
           xmin = -Inf, xmax = Inf,
           ymin = 15, ymax = 15000, fill = "lightblue",
           alpha = .2
  )+
  stat_summary(data=diamonds,fun.y="median",geom="line",mapping = aes(group=1L))
```

``` r

ggplot(diamonds, aes(x = cut, y = price)) +
  annotate("rect",
           xmin = -Inf, xmax = Inf,
           ymin = 15, ymax = 15000, fill = "lightblue",
           alpha = .2
  )+
  stat_summary(data=diamonds,fun.y="median",geom="line",mapping = aes(group=1L))
#> Error: Discrete value supplied to continuous scale

#  cut order changed annotation does not start from the edge of plot
ggplot(diamonds, aes(x = cut, y = price)) +
  annotate("rect",
           xmin = "Fair", xmax = "Ideal",
           ymin = 15, ymax = 15000, fill = "lightblue",
           alpha = .2
  )+
  stat_summary(data=diamonds,fun.y="median",geom="line",mapping = aes(group=1L))
```

<sup>Created on 2019-03-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

yutannihilation:
Fix #3184 

In ggplot2, `Inf` might mean continuous value or discrete value. But, currently, the plot fails in the following case

* the first layer's data is all `Inf`
* the second layer's data is discrete

It seems better to skip picking a scale for the layer that has only infinite values. Note that, if all layers are such ones, the plot fails no matter we ignore them or not, because we cannot get a range for plots. So, I think this change is fine.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -74,6 +74,8 @@
   columns (especially in `psavert`), but more importantly, strips the grouping
   attributes from `economics_long`.
 
+* All-`Inf` layers are now ignored for picking the scale (@yutannihilation, #3184).
+
 # ggplot2 3.1.0
 
 ## Breaking changes
--- a/R/scale-type.R
+++ b/R/scale-type.R
@@ -1,5 +1,8 @@
 find_scale <- function(aes, x, env = parent.frame()) {
-  if (is.null(x)) {
+  # Inf is ambiguous; it can be used either with continuous scales or with
+  # discrete scales, so just skip in the hope that we will have a better guess
+  # with the other layers
+  if (is.null(x) || all(is.infinite(x))) {
     return(NULL)
   }
 
--- a/R/scale-type.R
+++ b/R/scale-type.R
@@ -2,7 +2,7 @@ find_scale <- function(aes, x, env = parent.frame()) {
   # Inf is ambiguous; it can be used either with continuous scales or with
   # discrete scales, so just skip in the hope that we will have a better guess
   # with the other layers
-  if (is.null(x) || all(is.infinite(x))) {
+  if (is.null(x) || (rlang::is_atomic(x) && all(is.infinite(x)))) {
     return(NULL)
   }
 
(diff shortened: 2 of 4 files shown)


clauswilke:
What happens when all layers satisfy this condition?

yutannihilation:
I was thinking about that... I'm 90% sure it fails anyway, because it doesn't have a range. Am I missing something?

``` r
library(ggplot2)
library(patchwork)

p <- ggplot() +
  annotate("rect", xmin = -Inf, xmax = Inf, ymin = -Inf, ymax = Inf, fill = "black")

p / p + xlim(0, 1) + ylim(0, 1)
```



<sup>Created on 2019-03-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
I can confirm that it fails for the current ggplot2, so the patch is probably fine.

PR #3194, opened 2019-03-19
Make geom_col() ignore data$width

### LINKED ISSUE #3142: geom_bar()/geom_col() erroneously warn that they ignore width aesthetic
`geom_bar()` and `geom_col()` let you specify a `width` aesthetic to control the width of the bars.

The behavior is as expected, but it generates an erroneous warning "Warning: Ignoring unknown aesthetics: width".

`width` isn't listed in the aesthetics section of [`?geom_bar`](https://ggplot2.tidyverse.org/reference/geom_bar.html), so it appears that this is an unofficial behavior.

----

``` r
library(ggplot2)
suppressPackageStartupMessages(library(dplyr))
mtcars_by_cyl <- mtcars %>% 
  group_by(cyl) %>% 
  summarize(
    mean_wt = mean(wt),
    n = n()
  ) %>% 
  mutate(prop = n / sum(n))

ggplot(mtcars_by_cyl) + 
  geom_col(aes(cyl, mean_wt, width = prop))
#> Warning: Ignoring unknown aesthetics: width
```

Created on 2019-02-13 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

----

Similar closed issues.

- https://github.com/tidyverse/ggplot2/issues/1904
- https://github.com/tidyverse/ggplot2/issues/2473

yutannihilation:
Fix #3142

Currently, `geom_bar()` does ignore `width` mapping, but `geom_col()` does not. Here's my thinking:

* We want to enforce a constant bar width within a panel, so `width` cannot be an aesthetic.
* Yet, the widths can vary among panels, so we need to pass `width`s per bar via `data`, not a single value via `param`.
* `data$width` should be used only when it's a calculated value by Stat. But, as `geom_col()` is not the case, it should ignore `data$width`.

Note that, this doesn't prevent variable widths when `geom_bar()` uses such Stat as

* `StatIdentity`
* some Stat that actually returns variable widths

but I think `geom_col()` should ignore `width` for consistency at least.

### Current HEAD

``` r
library(ggplot2)
library(patchwork)

d <- data.frame(x = c(1,2,3,3,3))

p1 <- ggplot(d) + geom_bar(aes(x, width = x / 2), alpha = 0.7) + ggtitle("bar")
#> Warning: Ignoring unknown aesthetics: width
p2 <- ggplot(d) + geom_col(aes(x, x, width = x / 2), alpha = 0.7) + ggtitle("col")
#> Warning: Ignoring unknown aesthetics: width

p1 / p2
#> Warning: position_stack requires non-overlapping x intervals
```

<sup>Created on 2019-03-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

### This PR

``` r
library(ggplot2)
library(patchwork)

d <- data.frame(x = c(1,2,3,3,3))

p1 <- ggplot(d) + geom_bar(aes(x, width = x / 2), alpha = 0.7) + ggtitle("bar")
#> Warning: Ignoring unknown aesthetics: width
p2 <- ggplot(d) + geom_col(aes(x, x, width = x / 2), alpha = 0.7) + ggtitle("col")
#> Warning: Ignoring unknown aesthetics: width

p1 / p2
```

<sup>Created on 2019-03-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

### DIFF
--- a/R/geom-col.r
+++ b/R/geom-col.r
@@ -38,8 +38,11 @@ GeomCol <- ggproto("GeomCol", GeomRect,
   non_missing_aes = c("xmin", "xmax", "ymin", "ymax"),
 
   setup_data = function(data, params) {
-    data$width <- data$width %||%
-      params$width %||% (resolution(data$x, FALSE) * 0.9)
+    # To ensure bar widths are constant within a PANEL, `width` is not allowed
+    # to be mapped to data. So, even if the `data` already contains `width`, it
+    # should be ignored except when the width is calculated by Stat. But, as
+    # geom_col() uses stat_identity() fixedly, this is not the case.
+    data$width <- params$width %||% (resolution(data$x, FALSE) * 0.9)
     transform(data,
       ymin = pmin(y, 0), ymax = pmax(y, 0),
       xmin = x - width / 2, xmax = x + width / 2, width = NULL
--- a/R/geom-bar.r
+++ b/R/geom-bar.r
@@ -118,6 +118,10 @@ GeomBar <- ggproto("GeomBar", GeomRect,
   non_missing_aes = c("xmin", "xmax", "ymin", "ymax"),
 
   setup_data = function(data, params) {
+    # Contrary to the case of geom_col(), the `width` in `data` is very likely
+    # a calculated column by Stat. So, geom_bar() should use data$width if
+    # available. Note that StatIdentity is not the case, but, as we cannot know
+    # what Stat was used, we can do nothing here.
     data$width <- data$width %||%
       params$width %||% (resolution(data$x, FALSE) * 0.9)
     transform(data,
--- a/tests/testthat/test-geom-col.R
+++ b/tests/testthat/test-geom-col.R
@@ -14,3 +14,12 @@ test_that("geom_col removes columns with parts outside the plot limits", {
     "Removed 1 rows containing missing values"
   )
 })
+
+test_that("geom_col() ignores width", {
+  dat <- data_frame(x = c(1, 2, 3))
+
+
+  p <- ggplot(dat, aes(x, x, width = x)) + geom_col()
+  d <- layer_data(p)
+  expect_equal(d$xmax - d$xmin, c(0.9, 0.9, 0.9))
+})


R/geom-col.r
```diff
@@ -38,8 +38,11 @@ GeomCol <- ggproto("GeomCol", GeomRect,
   non_missing_aes = c("xmin", "xmax", "ymin", "ymax"),
 
   setup_data = function(data, params) {
-    data$width <- data$width %||%
-      params$width %||% (resolution(data$x, FALSE) * 0.9)
+    # To ensure bar widths are constant within a PANEL, `width` is not allowed
+    # to be mapped to data. So, even if the `data` already contains `width`, it
+    # should be ignored except when the width is calculated by Stat. But, as
+    # geom_col() uses stat_identity() fixedly, this is not the case.
```
clauswilke: Just wanted to point out that if somebody writes code such as `stat_bin(geom = "col")` they could arrive here with a calculated `width` column, but that would probably be considered inappropriate use then. May be worth a mention in the comment, for the future.

yutannihilation: Sure.

yutannihilation:
> Do you want to mention in the documentation that `geom_col()` enforces equal column widths?

Yes, I think it's good to have it documented clearly.

yutannihilation:
Hmmm, sorry, after thinking about `stat_bin(geom = "col")`, now I feel this PR just creates another inconsistency. Let me withdraw this. Really sorry for taking your time, Claus.

To fix this, I think we need to remove `data$width` from `geom_bar()` too, but we don't have any measure to propagate calculated widths from a Stat to  a Geom other than `data` (for now).

``` r
library(ggplot2)

ggplot(diamonds) +
  stat_bin(aes(carat, width = stat(count)), geom = "col") 
#> Warning: Ignoring unknown aesthetics: width
```



``` r

ggplot(diamonds) +
  stat_bin(aes(carat, width = stat(count)), geom = "bar") 
#> Warning: Ignoring unknown aesthetics: width
#> Warning: position_stack requires non-overlapping x intervals
```



<sup>Created on 2019-03-22 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Issue #3207, opened 2019-03-27
Release ggplot2 3.1.1


thomasp85:
Prepare for release: 
 
* [x] `devtools::check()` 
* [x] `devtools::check_win_devel()` 
* [x] `rhub::check_for_cran()` 
* [x] `revdepcheck::revdep_check(num_workers = 4)` 
* [x] [Polish NEWS](http://style.tidyverse.org/news.html#before-release) 
* [x] Polish pkgdown reference index 
* [x] Draft blog post 
 
Submit to CRAN: 
 
* [x] `usethis::use_version()` 
* [x] Update `cran-comments.md` 
* [x] `devtools::submit_cran()` 
* [x] Approve email 
 
Wait for CRAN... 
 
* [x] Accepted :tada: 
* [ ] `usethis::use_github_release()` 
* [ ] `usethis::use_dev_version()` 
* [ ] Finish blog post 
* [ ] Tweet 
* [ ] Add link to blog post in pkgdown news menu

thomasp85:
@Hadley @karawoo @clauswilke @yutannihilation we will make a minor release of the current master shortish, in order to make the integration of the new unit system in grid as smooth as possible so it can be part of the R 3.6 release... There are no breaks in reverse dependencies, so it should be smooth sailing

Please hold of on merging anything into master until the release process has ended

clauswilke:
I think we should address #3158 before a 3.2.0 release, so we don't release a new API we immediately invalidate with the next release. 

I would delete these five lines and not make any other changes:
https://github.com/tidyverse/ggplot2/blob/033fb521a084447c15178c591ab96a158f3b1e1e/R/layer-sf.R#L3-L7

@yutannihilation [suggested](https://github.com/tidyverse/ggplot2/issues/3158#issuecomment-466286581) to remove the `layer_class` argument from `layer()`, but my preference is to leave it in. Even if it isn't useful outside of ggplot2 at this point, it will be useful for internal ggplot2 code and substantially reduce code duplication. We could add "for ggplot2 internal use only" to the documentation to make this clear.

thomasp85:
I'll make sure to remove those lines

clauswilke:
@thomasp85 Thanks! I'm traveling through the end of the week and won't have much time to look into anything before the weekend.

thomasp85:
No problem... there is not much to do as there is zero revdep problems :-)

dpseidel:
[...]

PR #3208, opened 2019-03-27
Don't export LayerSf

### LINKED ISSUE #3158: Don't export LayerSf
When I implemented custom layers for sf objects, I exported the relevant layer object, `LayerSf()`:
https://github.com/tidyverse/ggplot2/blob/033fb521a084447c15178c591ab96a158f3b1e1e/R/layer-sf.R#L3-L8

This is bad, because we currently don't export its parent, `Layer()`, and we don't want to export it at this time.

The correct way to make the sf-specific layer available is to add a constructor function, `layer_sf()`, which replaces the `layer()` function for sf layers.

thomasp85:
Fix #3158 This will undo the export of `LayerSf`


thomasp85:
I'll merge it now as it may otherwise be lost in the process... the `layer_type` argument is something we can have another time I think...

clauswilke:
I’m more concerned that we have unexported `LayerSf` but haven’t added `layer_sf()` yet. This may have downstream implications for packages under development, e.g. ggrepel. @yutannihilation is the person most likely to have such code floating around somewhere.

yutannihilation:
Thanks for mentioning me! I think it's fine not to add `layer_sf()`. We can add it when we get sure it's needed.

One thing, do you think we can remove `layer_class` argument of `layer()`? Though this argument is visible to users, this is not usable because no `Layer` class is exported.

https://github.com/tidyverse/ggplot2/blob/033fb521a084447c15178c591ab96a158f3b1e1e/R/layer.r#L66

It's not necessary but nice if we can remove this to avoid confusion.

yutannihilation:
Ah, sorry I missed this comment: https://github.com/tidyverse/ggplot2/issues/3207#issuecomment-477293511

OK, let's leave `layer_class` and do this:

> We could add "for ggplot2 internal use only" to the documentation to make this clear.

PR #3221, opened 2019-04-02
ggplot2 theme background with strip color: white and grey gird background

LuffyLuffy:
Dear ggplot2 authors,
Thanks for your awesome job at data visualization!
There is only one parameters to set for panel background color, and I read the theme_grey() function, the element_rect is set only with one color setting. No idea how to set a strip color.
Although it is possible to add a geom_rect layer by hand, however, the categorical axis is hard to present a better outlook. 
Hopefully, there is an theme choice for color band background.


clauswilke:
I don't exactly understand what happened here. Somehow you created a pull request of commits you didn't author. I'm going to close this. If you meant to open an issue, please open an issue, though I have to admit that I don't fully understand what you are asking or proposing. Please try to be more precise and more specific the next time, and please provide a reprex.

karawoo:
I believe the request is to add an option for `panel.background` to be striped with two colors instead of a solid color. If that's right, it seems like it would introduce perceptual issues, and I don't see a clear benefit. I doubt we would add this to the theming system.

LuffyLuffy:
> I believe the request is to add an option for `panel.background` to be striped with two colors instead of a solid color. If that's right, it seems like it would introduce perceptual issues, and I don't see a clear benefit. I doubt we would add this to the theming system.
Exactly what I means, karawoo!
I think when there is a lot of groups in a category, a strip background would be more clear to be viewed.
Below is a graph I tried to use matplotlib.
image
Below is I tried with ggplot2
image.

clauswilke:
In any case, this discussion doesn't belong into a broken pull request. @LuffyLuffy, if you want to continue making a case for your feature request, please open a proper issue and please explain clearly what you are requesting.

PR #3223, opened 2019-04-02
Fix Position docs

flying-sheep:
The function signature is `compute_panel(self, data, params, scales) -> data`

### DIFF
--- a/R/position-.r
+++ b/R/position-.r
@@ -17,7 +17,7 @@
 #'     per layer. `panel` is currently an internal data structure, so
 #'     this method should not be overridden.
 #'
-#'   - `compute_panel(self, data, params, panel)` is called once per
+#'   - `compute_panel(self, data, params, scales)` is called once per
 #'     panel and should return a modified data frame.
 #'
 #'     `data` is a data frame containing the variables named according


thomasp85:
Thanks

yutannihilation:
FYI: you needed to run `devtools::document()` to re-generated the docs.

flying-sheep:
Sure, but for one-off fixes using the GitHub UI, that’s not feasible.

Maybe there should be a bot that recognizes PRs that change `#' doc comments` where no `.Rd` changes happened and adds them in commits to (or after) that PR.

batpigandme:
Given that there can be several functions inside a single `.R` file that export to different `.Rd`s, and roxygen's parsing of tags, this would be non-trivial. I'm not sure where a Feature Request for something like this world go. Maintainers with write access can always document for you, if it's a problem.

flying-sheep:
The idea is simpler: If any line in a `.R` file starting with `#'` is changed, but no file with an `.Rd` extension is changed in the same PR, the bot checks out the changes, runs `devtools::document()` and checks if that would produce any changes. If it does, it either commits them directly to the existing PR or creates another PR depending on it.

batpigandme:
I like the idea. It deserves to live somewhere other than ggplot2 PR comments.

flying-sheep:
What are you thinking? Sadly roxygen2 doesn’t live in an organization where we could propose that.

hadley:
It's already in the pipeline.

PR #3231, opened 2019-04-11
Add menu-bar link to the extension site

### LINKED ISSUE #3189: Better integrate extensions site with main site
At a minimum, I think we should link to from the main navbar, and we should probably mention it in the readme.

@yutannihilation do you have other ideas?

thomasp85:
Fix #3189

This PR add a menu item that links to the extension site... @Hadley is this what you had in mind?

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -243,7 +243,7 @@ navbar:
     - text: "Change log"
       href: news/index.html
   - text: Extensions
-    href: https://www.ggplot2-exts.org
+    href: http://www.ggplot2-exts.org/gallery/
   right:
   - icon: fa-github fa-lg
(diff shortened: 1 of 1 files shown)


_pkgdown.yml
```diff
@@ -242,6 +242,8 @@ navbar:
     - text: "------------------"
     - text: "Change log"
       href: news/index.html
+  - text: Extensions
+    href: https://www.ggplot2-exts.org
```
hadley: Should it link directly to http://www.ggplot2-exts.org/gallery/ ?

thomasp85: yeah, that is probably a more inviting landing page...

PR #3234, opened 2019-04-11
Don't export empty

thomasp85:
Latest CRAN release showed a couple of reverse dependency problems. One is that we have exported our own implementation of `empty()` which creates namespace conflicts if people import both plyr and ggplot2. This PR undoes that


karawoo:
Do we know that other revdeps aren't using it? The extending ggplot2 vignette demonstrates it in two places:

https://github.com/tidyverse/ggplot2/blob/47d65ddbb92066513578846512f98355d9b5e0d9/vignettes/extending-ggplot2.Rmd#L596

https://github.com/tidyverse/ggplot2/blob/47d65ddbb92066513578846512f98355d9b5e0d9/vignettes/extending-ggplot2.Rmd#L740

thomasp85:
Ah - perhaps that was why we ended up exporting it. We never reexported it before so whatever used it before had to import it from plyr so the unexport shouldn’t create a problem. I’ll fix the use in the vignettes somehow

PR #3235, opened 2019-04-11
fix problems during ribbon data sort

thomasp85:
reverse dependency check picked up an issue with the new sorting of `geom_ribbon()` in `setup_data()`. When `ymin` is set outside `aes()` it would fail as it wouldn't have access to the variable at setup time. This PR mitigates that.

### DIFF
--- a/R/geom-ribbon.r
+++ b/R/geom-ribbon.r
@@ -63,7 +63,12 @@ GeomRibbon <- ggproto("GeomRibbon", Geom,
   required_aes = c("x", "ymin", "ymax"),
 
   setup_data = function(data, params) {
-    transform(data[order(data$PANEL, data$group, data$x), ], y = ymin)
+    if (is.null(data$ymin) && is.null(data$ymax)) {
+      stop("Either ymin or ymax must be given as an aesthetic", call. = FALSE)
+    }
+    data <- data[order(data$PANEL, data$group, data$x), , drop = FALSE]
+    data$y <- data$ymin %||% data$ymax
+    data
   },
 
   draw_key = draw_key_polygon,
--- a/R/geom-ribbon.r
+++ b/R/geom-ribbon.r
@@ -64,7 +64,7 @@ GeomRibbon <- ggproto("GeomRibbon", Geom,
 
   setup_data = function(data, params) {
     if (is.null(data$ymin) && is.null(data$ymax)) {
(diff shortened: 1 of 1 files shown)


clauswilke:
I vaguely recall that sometimes there are cases where a variable doesn't make it into `data` but is available in `params`. This has caused me to write code such as this: https://github.com/clauswilke/ggridges/blob/bb2b9b4c6b1d304fad5da33492b7377c96388f5e/R/geoms.R#L126-L140

Is this applicable here?

thomasp85:
it seems to be dealing with the same problem... For this specific instance I can't see how/why one would provide both a constant `ymin` and `ymax` to `geom_ribbon()` so either should be available during `setup_data()`... andI think it is fair to error out if non are an aesthetic...

But if you feel strongly about it I'll adopt your approach

clauswilke:
I don't have a strong opinion of how it should be coded. However, I think there are 4 distinct cases that should work, and you can verify whether they all do:

1. Geom used directly (via `geom_ribbon()`), `ymin`/`ymax` is provided inside of `aes()`.
2. Geom used directly (via `geom_ribbon()`), `ymin`/`ymax` is provided outside of `aes()`.
3. Geom used via the `geom` argument of a stat, `ymin`/`ymax` is provided inside of `aes()`.
4. Geom used via the `geom` argument of a stat, `ymin`/`ymax` is provided outside of `aes()`.

The error message implies that `ymin`/`ymax` cannot be used outside of `aes()` ("must be given as an aesthetic"), and I found that concerning.

thomasp85:
The error says _either_ so at least one of them must be given in `aes()`. This is a fair requirement IMO because the existence of one of them defines the horizontal span of the geom... If both were given outside `aes()` the ribbon would have no width.

R/geom-ribbon.r
```diff
@@ -63,7 +63,12 @@ GeomRibbon <- ggproto("GeomRibbon", Geom,
   required_aes = c("x", "ymin", "ymax"),
 
   setup_data = function(data, params) {
-    transform(data[order(data$PANEL, data$group, data$x), ], y = ymin)
+    if (is.null(data$ymin) && is.null(data$ymax)) {
+      stop("Either ymin or ymax must be given as an aesthetic", call. = FALSE)
```
thomasp85:
```suggestion
      stop("Either ymin or ymax must be given as an aesthetic.", call. = FALSE)
```

Issue #3145, opened 2019-02-15
Make legend glyphs configurable


clauswilke:
Currently all geoms have a hard-coded legend glyph that can't be changed without some knowledge about the inner workings of ggplot2. It would be helpful if the glyphs could be configurable. Associated with such a change, we might want to provide a broader array of different glyphs. One example is the stylized time series glyph of PR #3124.

It seems to me the easiest way to achieve this modification is to provide an additional parameter `draw_key` to the constructor of each geom. The downside is that it requires a modification of each individual `geom_*()` and will not automatically propagate to extension packages. The upside is that the modification required is rather minor. Example implementation follows.

``` r
library(ggplot2)

# alternative key glyph
draw_key_timeseries <- function(data, params, size) {
  data$linetype[is.na(data$linetype)] <- 0

  grid::linesGrob(
    x = c(0, 0.4, 0.6, 1),
    y = c(0.1, 0.6, 0.4, 0.9),
    gp = grid::gpar(
      col = scales::alpha(data$colour, data$alpha),
      lwd = data$size * .pt,
      lty = data$linetype,
      lineend = "butt"
    )
  )
}

# helper function
set_draw_key <- function(geom, draw_key = NULL) {
  if (is.null(draw_key)) {
    return(geom)
  }
  if (is.character(draw_key)) {
    draw_key <- paste0("draw_key_", draw_key)
  }
  draw_key <- match.fun(draw_key)
  
  ggproto("", geom, draw_key = draw_key)
}

# version of `geom_line()` with modifiable draw key
geom_line2 <- function(mapping = NULL, data = NULL, stat = "identity",
                       position = "identity", na.rm = FALSE,
                       show.legend = NA, inherit.aes = TRUE, draw_key = NULL,
                       ...) {
  layer(
    data = data,
    mapping = mapping,
    stat = stat,
    geom = set_draw_key(GeomLine, draw_key),
    position = position,
    show.legend = show.legend,
    inherit.aes = inherit.aes,
    params = list(
      na.rm = na.rm,
      ...
    )
  )
}

# usage
base <- ggplot(data.frame(x = 1:3, y = 3:1), aes(x, y))
base + geom_line2(aes(color = "default"))
```

![](https://i.imgur.com/YsAdjII.png)

``` r
base + geom_line2(aes(color = "modified"), draw_key = "timeseries")
```

![](https://i.imgur.com/oEOvSpf.png)

``` r
[...]

PR #3236, opened 2019-04-11
Customizable key glyphs

clauswilke:
This PR implements customizable legend key glyphs, as discussed in #3145.

``` r
library(ggplot2)

base <- ggplot(data.frame(x = 1:3, y = 3:1), aes(x, y))

cowplot::plot_grid(
  base + geom_line(aes(color = "default")),
  base + geom_line(aes(color = "modified"), key_glyph = "timeseries"),
  base + geom_line(aes(color = "modified"), key_glyph = "vpath"),
  base + geom_line(aes(color = "modified"), key_glyph = "abline")
)
```

<sup>Created on 2019-04-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

@hadley @thomasp85 @yutannihilation I think this is the simplest and minimally invasive implementation. However, it does some shuffling around with the `params` parameter in `layer()`, so I'd like to hear your perspective. The alternative would require modifications in every geom, and would not automatically work in extension packages until they modify their geoms also.


clauswilke:
I think this is ready for review now.

yutannihilation:
> The alternative would require modifications in every geom, and would not automatically work in extension packages until they modify their geoms also.

Do you think there are actual use cases of `key_glyph` argument? If we don't modify the existing geoms, to me, it seems we'll stick with `params` only.

NEWS.md
```diff
@@ -35,6 +35,9 @@ core developer team.
 * `geom_rug()` gains an "outside" option to allow for moving the rug tassels to 
   outside the plot area (@njtierney, #3085) and a `length` option to allow for 
   changing the length of the rug lines (@daniel-wells, #3109). 
+  
+* All geoms now take a `key_glyph` paramter that allows users to customize
+  how legend keys are drawn (@clauswilke, #3145).
```
yutannihilation: Can we credit @mitchelloharawild for inventing the timeseries glyph here?

R/legend-draw.r
```diff
@@ -200,10 +219,34 @@ draw_key_label <- function(data, params, size) {
...
+      col = alpha(data$colour %||% data$fill %||% "black", data$alpha),
+      lwd = (data$size %||% 0.5) * .pt,
+      lty = data$linetype %||% 1,
       lineend = "butt"
     )
   )
 }
+
+
+
```
yutannihilation: Seems too much blank lines.

NEWS.md
clauswilke: Done.

R/legend-draw.r
clauswilke: Fixed.

clauswilke:
> Do you think there are actual use cases of `key_glyph` argument? If we don't modify the existing geoms, to me, it seems we'll stick with `params` only.

I added this to avoid a secret, undocumented parameter. I'm not sure how to handle this best. One perspective is that all geoms should define a `key_glyph` argument and then pass it on to `layer()` explicitly. However, then setting this argument to `NULL` in the geom constructors would look weird, and setting it to the actual key glyph in the geom constructor would always override the setting in the `Geom`, which is also weird. The way I did it seemed the best compromise. Sufficiently hidden that it's not in your face, but sufficiently present that it is discoverable.

yutannihilation:
As for documentation, I think we can list and describe notable parameters in the doc of `params`, but I too am not sure what's the best way for this... I want to wait for others' comments.

thomasp85:
>Do you think there are actual use cases of key_glyph argument? If we don't modify the existing geoms, to me, it seems we'll stick with params only.

I confess to not really understanding the above. @yutannihilation can you elaborate?

clauswilke:
@thomasp85 What @yutannihilation was referring to is that now `layer()` has a `key_glyph` argument but there is currently no code that actually uses it, because the geoms provide `key_glyph` via the `params` argument. He asked whether we should eliminate the `key_glyph` argument from `layer()`. My opinion is "no", and I explained it here: https://github.com/tidyverse/ggplot2/pull/3236#issuecomment-482763195

thomasp85:
Ah, I see. Yeah, I like your approach

PR #3238, opened 2019-04-11
Add ... to cut width to pass to cut.default

### LINKED ISSUE #3055: cut_width should be able to take ... for cut.default
From a question in [community.rstudio.com](https://community.rstudio.com/t/cut-width-labels-false-returns-unused-argument-error/20675) it seems like `cut_width` is the [only one](https://ggplot2.tidyverse.org/reference/cut_interval.html) of the three functions to not take `...` to pass it to `cut.default`

In particular, `labels` argument can not be used with `cut_width`. Just adding `...` would allow this
```r
> cut_width(runif(1000), 0.1, center = 0, labels= FALSE)
  [1]  2  6  9 10  8  7 11  6  6  3 11 11  5  6 10  3  4  8  8 11  2  3  5  2  3  6  8  7  4  6  7 11 10  4 10 11
 [37]  6  5  6  1  3  6  7  1  9  9  3  2  9 11  8  3  8  6 11 10  6  9  4  3  1  8  6  6  5 10  6  9  9  9  8  4
 [73] 11  2  2 10  4  8  9  3  3  3 11 11  4  2  8 11  8 10  5  7 10  2  1  2  6  3  2  2
 [ reached getOption("max.print") -- omitted 900 entries ]
```
or to change labels if the call (could always be done afterward using `forcats`)

I open the issue to reference the question and see if it worth doing. Seems quite easy to be consistent between those three `cut_*` helper in lubridate.

If you're interested I'll do a PR.

cderv:
This fixes #3055 

I added also a small example but not test as there were none for `cut_*` functions. 

Tell me if you need some.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -50,6 +50,9 @@ core developer team.
 
 ## Minor improvements and bug fixes
 
+* `cut_width()` now accepts `...` to pass further arguments to `base::cut.default()`
+   like `cut_number()` and `cut_interval()` already did (@cderv, #3055)
+
 * `coord_map()` now can have axes on the top and right (@karawoo, #3042).
 
 * `coord_sf()`, `coord_map()`, and `coord_polar()` now squash `-Inf` and `Inf`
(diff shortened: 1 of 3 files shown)


thomasp85:
Thanks!

PR #3239, opened 2019-04-11
New news section for next release

karawoo:
I think what was just released in 3.1.1 should be separate from whatever gets merged between now and the next release.


thomasp85:
The 3.1.1 release did not include any of that 🙂 it was a purely under-the-hood change and didn't really have any news bullets

karawoo:
Ohh I see, I missed that. Never mind!

PR #3240, opened 2019-04-12
adding se, level, n, fullrange parameters to stat_quantile

smouksassi:
I tried to follow the procedures to the best of my knowledge this is the first time I do a PR.
I modified the examples to show the usage of se and n. fullrange = TRUE will issue a message telling the user that fullrange = FALSE is forced ( better than erroring out in my opinion)


smouksassi:
closing  as this is more suitable to an extension package

PR #3242, opened 2019-04-13
Import rlang and use unqualified form for rlang functions

### LINKED ISSUE #3195: @import rlang?
[Lionel said](https://github.com/tidyverse/ggplot2/pull/3161#pullrequestreview-215714849):

> Our default practice is to import the whole rlang package and use the non-qualified form.

If it's a common practice among tidyverse, I think we should follow it. It is probably good for performance (though I'm not sure how much it matters).

yutannihilation:
Fixes #3195 

Since it's common practice to import whole rlang package, ggplot2 should follow it. This PR adds `import(rlang)` to `NAMESPACE` and removes all `rlang::`.


yutannihilation:
Thanks.

PR #3244, opened 2019-04-13
Fix formals of compute_layer()

### LINKED ISSUE #3202: Inconsistent formals for `StatIdentity$compute_layer`
The formals for `compute_layer` in the majority of the Stats in ggplot2 are `c(self, data, params, layout)`

Two Stats are different from the others: `StatBindot` and `StatIdentity`.

Is there a reason for the different formals for these 2 stats?  Is this just a leftover from earlier version?  

I'm certain the `StatIdentity$compute_layer` and `StatBindot$compute_layer` argument names don't matter (either passed through or ignored) but it may be worth fixing for consistency

```
# formals for `compute_layer` across all Stats
Stat                [1] "self"   "data"   "params" "layout"
StatBin             [1] "self"   "data"   "params" "layout"
StatBin2d           [1] "self"   "data"   "params" "layout"
StatBindot          [1] "self"   "data"   "params" "panels"
StatBinhex          [1] "self"   "data"   "params" "layout"
StatBoxplot         [1] "self"   "data"   "params" "layout"
StatContour         [1] "self"   "data"   "params" "layout"
StatCount           [1] "self"   "data"   "params" "layout"
StatDensity         [1] "self"   "data"   "params" "layout"
StatDensity2d       [1] "self"   "data"   "params" "layout"
StatEcdf            [1] "self"   "data"   "params" "layout"
StatEllipse         [1] "self"   "data"   "params" "layout"
StatFunction        [1] "self"   "data"   "params" "layout"
StatIdentity        [1] "data"   "scales" "params"
StatQq              [1] "self"   "data"   "params" "layout"
StatQqLine          [1] "self"   "data"   "params" "layout"
StatQuantile        [1] "self"   "data"   "params" "layout"
StatSf              [1] "self"   "data"   "params" "layout"
StatSfCoordinates   [1] "self"   "data"   "params" "layout"
StatSmooth          [1] "self"   "data"   "params" "layout"
StatSum             [1] "self"   "data"   "params" "layout"
StatSummary         [1] "self"   "data"   "params" "layout"
StatSummary2d       [1] "self"   "data"   "params" "layout"
StatSummaryBin      [1] "self"   "data"   "params" "layout"
StatSummaryHex      [1] "self"   "data"   "params" "layout"
StatUnique          [1] "self"   "data"   "params" "layout"
StatYdensity        [1] "self"   "data"   "params" "layout"
```

yutannihilation:
Fixes #3202

`Stat*$compute_layer()` and `Position*$compute_layer()` have various names of formals, which seems for no valid reason; just because the common formals were so at that time (c.f. https://github.com/tidyverse/ggplot2/issues/3202#issuecomment-478307143). 

It seems these methods are not passed parameters by name, but by positions only, so I think this is safe to fix:

``` r
$ git grep "compute_layer("
R/layer.r:    self$stat$compute_layer(data, params, layout)
R/layer.r:    self$position$compute_layer(data, params, layout)
R/position-.r:#'   - `compute_layer(self, data, params, panel)` is called once
R/stat-.r:#'     `compute_layer(self, data, scales, ...)`,
R/stat-.r:#'     `compute_layer()` is called once per layer, `compute_panel_()`
R/stat-bindot.r:    ggproto_parent(Stat, self)$compute_layer(data, params, panels)
man/ggplot2-ggproto.Rd:\code{compute_layer(self, data, scales, ...)},
man/ggplot2-ggproto.Rd:\code{compute_layer()} is called once per layer, \code{compute_panel_()}
man/ggplot2-ggproto.Rd:\item \code{compute_layer(self, data, params, panel)} is called once
```

This PR fixes the inconsistency. I expect this would break nothing, but changing the argument names is (technically) a breaking change. So, I added a NEWS item for this.


NEWS.md
```diff
@@ -121,6 +121,9 @@ core developer team.
 
 * `sec_axis()` now accepts functions as well as formulas (@yutannihilation, #3031).
 
+* `Stat$compute_layer()` and `Position$compute_layer()` now have the same names of
+  arguments (@yutannihilation, #3202).
```
clauswilke: I feel something is missing here. "have now the same names of arguments *as other `compute_layer()` functions*"?

yutannihilation: Thanks. I should probably say something like "The arguments of `Stat*$compute_layer()` and `Position*$compute_layer()` are now renamed to match the ones of `Stat$compute_layer()` and `Position$compute_layer()`." Does this make sense...?

clauswilke: Yes, maybe with this edit: "...now renamed to *always* match the ones...". After all, they mostly were matching already, but not always.

yutannihilation: Got it!

PR #2934, opened 2018-10-11
theme: Add different ticks length for different axes

pank:
Hi,

I would like to add support for different ticks length for different axes.
My main motivation is to match an Excel theme that I have been given, which features upwards ticks on the x-axis. I don't want inwards ticks on the y-axis though.

It was first requested in issue #1319.

I have not submitted patches to `ggplot2` before, neither am I too familiar with the `tidyverse` style guide normally, nor the github pull interface (I normally use `git format-patch`.  If I have broken any conventions let me know and I will try to fix the mistake.

It was not clear to me what to do where the names I added were longer than the length of other nearby names.  The problem is that the alignment goes wrong. 
[Here](https://github.com/pank/ggplot2/commit/8340ec76fdaafe153713f1a2672ac786e1fdc3ee#diff-aac553f8bf010ed911dab683e9090a18) is an example.

Should I,
- Fix the alignment issues in a subsequent commit?  (still bad for `git blame`...)?
- Or find shorter names?
- Or just leave it be?

Also, I have not used visual tests before, but it seems to work. (I just ran `devtools::test()` and saw no more errors; the documentation is a bit sparse on visual tests).

Finally, [this bit](https://github.com/pank/ggplot2/commit/8340ec76fdaafe153713f1a2672ac786e1fdc3ee#diff-348ffaa265a58dc2db9470750b00523b) is pretty crude:

```r
  theme$axis.ticks.length.x <- theme$axis.ticks.length.x %||% theme$axis.ticks.length
  theme$axis.ticks.length.x.bottom <- theme$axis.ticks.length.x.bottom %||% theme$axis.ticks.length.x
  theme$axis.ticks.length.x.top <- theme$axis.ticks.length.x.top %||% theme$axis.ticks.length.x
  theme$axis.ticks.length.y <- theme$axis.ticks.length.y %||% theme$axis.ticks.length
  theme$axis.ticks.length.y.left <- theme$axis.ticks.length.y.left %||% theme$axis.ticks.length.y
  theme$axis.ticks.length.y.right <- theme$axis.ticks.length.y.right %||% theme$axis.ticks.length.y
```
I just followed the example of `legend.spacing.x`, but if desired I can try to cut it down.

Below is a minimal example of a somewhat of useful thing that could be done after applying the patch.  Notice the ticks on the button is sort-of enclosing the bar and the ones on the right axis serve as an explanation of the graph. (I'm not saying this is the best way to explain what is going on in this graph!)

Thanks,
Rasmus

min_example
```r
[...]

PR #3245, opened 2019-04-13
Make layer() accept formula notation for data argument 

### LINKED ISSUE #3138: Support formula notation in `data` argument ?
The data argument can be used to pass a function to apply on the plot's data. It would be nice to be able to use the formula notation as well.

```
# works
ggplot(iris,aes(Sepal.Length,Sepal.Width)) + 
  geom_point(data = head)
# works as well
ggplot(iris,aes(Sepal.Length,Sepal.Width)) + 
  geom_point(data = function(x) head(x,10))
# doesn't
ggplot(iris,aes(Sepal.Length,Sepal.Width)) + 
  geom_point(data = ~head(.,10))
```

It should be a simple fix, either in `layer` with something like `if(inherits(data,"formula")) data <- rlang::as_function(data)` or in later in `ggproto` after defining `fortify.formula <- function(model, data, ...) model` to return the input as is done with `fortify.function`, with the same but also testing for missing with something like `if(!missing(data) && inherits(data,"formula")) data <- rlang::as_function(data)`

```
trace(ggplot2:::layer,at =2,quote(
  if(inherits(data,"formula")) data <- rlang::as_function(data)))   

# now works
ggplot(iris,aes(Sepal.Length,Sepal.Width)) + 
  geom_point(data = ~head(.,10))
```

yutannihilation:
Fix #3138

`data` is processed by `fortify()` once and stored in layer. Then, if it's a function, it's applied to the plot data when the plot is built,

https://github.com/tidyverse/ggplot2/blob/475042e9ce209e1279d77d46288315f801150d4d/R/layer.r#L191-L192

So, all we have to do is to define a `fortify()` method that converts a formula to a function.

## Usages

``` r
devtools::load_all("~/repo/R/ggplot2/")
#> Loading ggplot2

  geom_point(data = ~ head(., 10))
```

<sup>Created on 2019-04-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


thomasp85:
There's a dependency on #3242... If you merge that one first and then update this to remove the rlang namespacing it's all good :-)

yutannihilation:
Sure, thanks!

man/theme.Rd
```diff
@@ -251,15 +251,20 @@ p1 + theme(
 )
 
 # Axes ----------------------------------------------------------------------
+# Change styles of axes texts and lines
```
yutannihilation: It seems these diffs of `theme.Rd` come from #2934. It might be better to create a separate PR, but I let this included.

clauswilke: Actually, that PR also missed to generate a new vdiffr reference image. Do you have the time to create a separate PR that includes the updated .Rd file and the missing image? That would probably be the best approach.

yutannihilation: Oh, sure. Will do.

Issue #3250, opened 2019-04-17
Unexpected failure of density plot without any error or warning


clauswilke:
I encountered the following example when teaching density plots. If a discrete variable is encoded as numerical (e.g., 0/1), and that variable is mapped onto `fill`, the variable values get replaced with `NA` and no fill is applied/no grouping is calculated. Importantly, this happens without any kind of error or warning message, so students get no indication of what they have done wrong, or that something went wrong at all.

It's not immediately clear to me whether and how we can catch this situation, but it would be good if we could. Reprex follows.

``` r
library(ggplot2)

# desired plot
p1 <- ggplot(mtcars, aes(mpg, fill = factor(am))) + geom_density(alpha = 0.4)
p1
```

![](https://i.imgur.com/a8Uc6S1.png)

``` r

# broken plot because factor() is missing
p2 <- ggplot(mtcars, aes(mpg, fill = am)) + geom_density(alpha = 0.4)
p2
```

![](https://i.imgur.com/frE2UxC.png)

``` r

# fill has been set to NA
head(layer_data(p2))
#>            y PANEL        x    density    scaled  ndensity     count  n
#> 1 0.01922476     1 10.40000 0.01922476 0.2829075 0.2829075 0.6151923 32
#> 2 0.01951519     1 10.44599 0.01951519 0.2871815 0.2871815 0.6244861 32
#> 3 0.01980689     1 10.49198 0.01980689 0.2914740 0.2914740 0.6338204 32
#> 4 0.02010390     1 10.53796 0.02010390 0.2958448 0.2958448 0.6433248 32
#> 5 0.02040182     1 10.58395 0.02040182 0.3002289 0.3002289 0.6528583 32
#> 6 0.02070509     1 10.62994 0.02070509 0.3046918 0.3046918 0.6625630 32
#>   group ymin       ymax fill weight colour alpha size linetype
#> 1    -1    0 0.01922476   NA      1  black   0.4  0.5        1
#> 2    -1    0 0.01951519   NA      1  black   0.4  0.5        1
#> 3    -1    0 0.01980689   NA      1  black   0.4  0.5        1
#> 4    -1    0 0.02010390   NA      1  black   0.4  0.5        1
#> 5    -1    0 0.02040182   NA      1  black   0.4  0.5        1
#> 6    -1    0 0.02070509   NA      1  black   0.4  0.5        1
```

[...]

PR #3251, opened 2019-04-19
WIP: Emit warning if non-constant data columns are dropped by stat transformation.

clauswilke:
Here is a first stab at fixing #3250, which is the silent dropping of mapped aesthetics. The code works in most cases, but there are false positives when the stat purposefully drops an aesthetic. For example, `stat_boxplot()` takes `y` and turns it into `ymin`, `ymax`, etc., and the current code doesn't know that.

One way to fix this would be to add a mechanism so that stats can declare aesthetics that are expected to be dropped, something like `Stat$dropped_aesthetics`, and the warning message would ignore those.

``` r
library(ggplot2)

# fill aesthetic is dropped, emit warning
ggplot(mtcars, aes(mpg, fill = am)) + geom_density(alpha = 0.5)
#> Warning: 1 aesthetic(s) dropped during statistical transformation: fill.
#> Did you forget to define a group aesthetic or to convert a numeric variable into a factor?
```

``` r

# fill aesthetic is not dropped because color sets up
# a grouping, no warning
ggplot(mtcars, aes(mpg, fill = am, color = factor(am))) + geom_density(alpha = 0.5)
```

``` r

# false positive, y aesthetic is dropped because it is converted
# to ymin, ymax, etc.
ggplot(mtcars, aes(factor(cyl), mpg)) + geom_boxplot()
#> Warning: 1 aesthetic(s) dropped during statistical transformation: y.
#> Did you forget to define a group aesthetic or to convert a numeric variable into a factor?
```

<sup>Created on 2019-04-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/stat-.r
+++ b/R/stat-.r
@@ -61,6 +61,10 @@ Stat <- ggproto("Stat",
 
   non_missing_aes = character(),
 
+  # aesthetics that are dropped from the data frame during the
+  # process of the statistical transformation
+  dropped_aes = character(),
+
   setup_params = function(data, params) {
     params
   },
@@ -119,7 +123,7 @@ Stat <- ggproto("Stat",
     # The above code will drop columns that are not constant within groups and not
     # carried over/recreated by the stat. This can produce unexpeted results,
     # and hence we warn about it.
-    dropped <- base::setdiff(names(data), names(data_new))
+    dropped <- base::setdiff(names(data), base::union(self$dropped_aes, names(data_new)))
     if (length(dropped) > 0) {
       warn_message <- paste0(
         length(dropped), " aesthetic(s) dropped during statistical transformation: ",
--- a/R/stat-.r
+++ b/R/stat-.r
@@ -114,7 +114,22 @@ Stat <- ggproto("Stat",
       )
     }, stats, groups, SIMPLIFY = FALSE)
 
-    rbind_dfs(stats)
+    data_new <- rbind_dfs(stats)
+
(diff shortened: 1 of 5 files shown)


clauswilke:
@hadley Could you provide some high-level guidance? Should we ignore the problem or extend stats so they can suppress spurious warning messages? The latter approach would be a breaking change.

hadley:
It seems reasonable, but I have no idea what existing plots this will cause to throw an error :/

clauswilke:
I updated the PR so all the unit tests pass. It required one-line modifications to four stats. Seems to have a relatively modest impact and easy to fix, but undoubtedly many extension packages would have to make similar adjustments.

The more I think about it, the more I like it thought that stats have to declare any aesthetics that they drop. I've always felt that the way auxiliary aesthetics (e.g., `colour`) were magically carried over during the stat processing was mysterious and would break unexpectedly. This PR would create a reliable framework around this issue.

thomasp85:
@clauswilke what are your current feelings around this - is it integrated into other internal changes we have on the map or is it fine to merge this in as is (for the release after the patch release)?

clauswilke:
This is a standalone modification that doesn't overlap with any other changes as far as I can see. And I keep running into the bug when teaching, so definitely something I'd like to move forward.

clauswilke:
Closing this PR in favor of #4866.

PR #3254, opened 2019-04-23
Fix grid errors when quantiles falls outside of violin

### LINKED ISSUE #2885: geom_violin with draw_quantiles - "Error in grid.Call.graphics(C_lines, x$x, x$y, index, x$arrow) : invalid hex digit in 'color' or 'lty'"
See https://stackoverflow.com/questions/48435319/geom-violin-error-in-grid-call-graphicsc-lines-xx-xy-index-xarrow/52228544#52228544

I can't share enough data to make it reproducible, but it appears that when `draw_quantiles` has too few data to calculate the quantiles, it results in the following error:

```
Error in grid.Call.graphics(C_lines, x$x, x$y, index, x$arrow) : 
  invalid hex digit in 'color' or 'lty'
```

thomasp85:
Fixes #2885 

In certain instances quantiles will have `NA` coordinates and the computed aesthetics will not match the number of lines to draw, leading to grid errors

### DIFF
--- a/R/geom-violin.r
+++ b/R/geom-violin.r
@@ -136,12 +136,17 @@ GeomViolin <- ggproto("GeomViolin", Geom,
       quantiles <- create_quantile_segment_frame(data, draw_quantiles)
       aesthetics <- data[
         rep(1, nrow(quantiles)),
-        setdiff(names(data), c("x", "y")),
+        setdiff(names(data), c("x", "y", "group")),
         drop = FALSE
       ]
       aesthetics$alpha <- rep(1, nrow(quantiles))
       both <- cbind(quantiles, aesthetics)
-      quantile_grob <- GeomPath$draw_panel(both, ...)
+      both <- both[!is.na(both$group), , drop = FALSE]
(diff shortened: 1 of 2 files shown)


R/geom-violin.r
```diff
@@ -136,12 +136,17 @@ GeomViolin <- ggproto("GeomViolin", Geom,
...
-        setdiff(names(data), c("x", "y")),
+        setdiff(names(data), c("x", "y", "group")),
         drop = FALSE
       ]
       aesthetics$alpha <- rep(1, nrow(quantiles))
       both <- cbind(quantiles, aesthetics)
-      quantile_grob <- GeomPath$draw_panel(both, ...)
+      both <- both[!is.na(both$group), , drop = FALSE]
+      quantile_grob <- if (nrow(both) == 0) {
+        zeroGrob()
```
yutannihilation: Do we need some warning here?

thomasp85: I don't think so, personally but will add it iff you feel strongly about it

yutannihilation: After some thinking, now I feel you are right.

PR #3256, opened 2019-04-23
Add warning when raster data doesn't fall on an even grid

### LINKED ISSUE #2932: Feature Request: Warning message and support for non-uniform tile sizes in geom_raster
Feature Request:
(a) add warning in geom_raster() when tiles are NOT all the same size
(b) handle irregularly sized tiles better in geom_raster()

a) sometimes you think that your data are all regularly gridded and then you run geom_raster() and rows appear to be missing from your raster.  In this case, a warning message would be a super helpful alert to the user that they've violated one of the underlying assumptions of geom_raster.

Either in geom-raster.R -> GeomRaster() or in utilities-resolution.R -> resolution(), run a quick check:

```r
if( length( unique( diff( unique(data$x) ) ) ) != 1 ) 
  warning("tiles not all of uniform size in x dimension") 

if( length( unique( diff( unique(data$y) ) ) ) != 1 ) 
  warning("tiles not all of uniform size in y dimension") 

```

b) Special case where entire rows/cols are different height/width.  If there is a way to do this without massively negatively impacting the performance of geom_raster... handle differently sized rows/cols better (e.g., fall back on geom_tile and calculate height/width from the data).  

Pseudo code in GeomRaster(), setup_data() : 

```r

ydiffs <- diff( unique(data$y) )
if( length( unique( ydiffs ) ) != 1 ) {
  warning("tiles not all of uniform size in y dimension") 
  diffvec <- data.frame(y = unique(data$y),
                                       h = c(ydiffs[1], ydiffs))
  h <- diffvec$h[match(data$y, diffvec$y)]
} else {
  h <- resolution(data$y, FALSE)
}

```

thomasp85:
Fixes #2932

This adds a warning when `geom_raster()` can't truthfully represent the provided data. To avoid recompilation the call to `resolution()` has been changed to direct calculation.

@batpigandme can you have a look at the wording of the warning?

### DIFF
--- a/R/geom-raster.r
+++ b/R/geom-raster.r
@@ -49,18 +49,23 @@ GeomRaster <- ggproto("GeomRaster", Geom,
   required_aes = c("x", "y"),
 
   setup_data = function(data, params) {
+    precision <- sqrt(.Machine$double.eps)
     hjust <- params$hjust %||% 0.5
     vjust <- params$vjust %||% 0.5
 
-    x_diff <- diff(sort(unique(data$x)))
-    if (!length(unique(x_diff)) == 1) {
+    x_diff <- diff(sort(unique(as.numeric(data$x))))
+    if (length(x_diff) == 0) {
+      w <- 1
+    } else if (any(abs(diff(x_diff)) > precision)) {
       warning("Raster pixels are placed at uneven horizontal intervals and will be shifted. Consider using geom_tile() instead.")
       w <- min(x_diff)
     } else {
       w <- x_diff[1]
     }
-    y_diff <- diff(sort(unique(data$y)))
-    if (!length(unique(x_diff)) == 1) {
+    y_diff <- diff(sort(unique(as.numeric(data$y))))
+    if (length(y_diff) == 0) {
+      h <- 1
+    } else if (any(abs(diff(y_diff)) > precision)) {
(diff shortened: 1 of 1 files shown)


clauswilke:
On first glance the code looks good to me but since the visual tests fail maybe there are some unanticipated side effects? The one difference I see between `resolution()` and your new code is that `resolution()` uses `unique(as.numeric(x))` where you use `unique(x)`.

clauswilke:
And also please make sure things behave in a reasonable way when the data vector contains only a single unique value. Not sure if that's the case right now or not. At a minimum, the warning in that case seems confusing.

R/geom-raster.r
```diff
@@ -52,8 +52,20 @@ GeomRaster <- ggproto("GeomRaster", Geom,
     hjust <- params$hjust %||% 0.5
     vjust <- params$vjust %||% 0.5
 
-    w <- resolution(data$x, FALSE)
-    h <- resolution(data$y, FALSE)
+    x_diff <- diff(sort(unique(data$x)))
+    if (!length(unique(x_diff)) == 1) {
```
yutannihilation: I think this assumption is too strong. For example:

``` r
unique(diff(c(1:10 / 10)))
#> [1] 0.1 0.1 0.1 0.1
```

<sup>Created on 2019-04-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

We need a version of `unique()` with tolerance.

Issue #3018, opened 2018-11-28
feature request: `layer_data` returns a tibble


IndrajeetPatil:
I often use the `layer_` functions from `ggplot2` for units tests. I was wondering if the `layer_data` function can return a `tibble`, which would make it much easier to peruse the data. Currently, this just overwhelms the console.

``` r
library(ggplot2)
#> Registered S3 methods overwritten by 'ggplot2':
#>   method         from 
#>   [.quosures     rlang
#>   c.quosures     rlang
#>   print.quosures rlang

# plot
p <- ggplot(iris, aes(Species, Sepal.Length)) + geom_point()

# checking data used
ggplot2::layer_data(p)
#>     x   y PANEL group shape colour size fill alpha stroke
#> 1   1 5.1     1     1    19  black  1.5   NA    NA    0.5
#> 2   1 4.9     1     1    19  black  1.5   NA    NA    0.5
#> 3   1 4.7     1     1    19  black  1.5   NA    NA    0.5
#> 4   1 4.6     1     1    19  black  1.5   NA    NA    0.5
#> 5   1 5.0     1     1    19  black  1.5   NA    NA    0.5
#> 6   1 5.4     1     1    19  black  1.5   NA    NA    0.5
#> 7   1 4.6     1     1    19  black  1.5   NA    NA    0.5
#> 8   1 5.0     1     1    19  black  1.5   NA    NA    0.5
#> 9   1 4.4     1     1    19  black  1.5   NA    NA    0.5
#> 10  1 4.9     1     1    19  black  1.5   NA    NA    0.5
#> 11  1 5.4     1     1    19  black  1.5   NA    NA    0.5
#> 12  1 4.8     1     1    19  black  1.5   NA    NA    0.5
#> 13  1 4.8     1     1    19  black  1.5   NA    NA    0.5
#> 14  1 4.3     1     1    19  black  1.5   NA    NA    0.5
#> 15  1 5.8     1     1    19  black  1.5   NA    NA    0.5
#> 16  1 5.7     1     1    19  black  1.5   NA    NA    0.5
#> 17  1 5.4     1     1    19  black  1.5   NA    NA    0.5
#> 18  1 5.1     1     1    19  black  1.5   NA    NA    0.5
#> 19  1 5.7     1     1    19  black  1.5   NA    NA    0.5
#> 20  1 5.1     1     1    19  black  1.5   NA    NA    0.5
[...]

PR #3264, opened 2019-04-24
return a tibble with `ggplot_build()` and `layer_data()`

malcolmbarrett:
Currently, `ggplot_build()` and `layer_data()` return a `data.frame`. This seems a little inconsistent with the rest of the tidyverse. This is little PR (motivated by some recent work with `layer_data()` that filled up my console) changes `ggplot_build()` to convert the data for the layers to `tibble`s., which is also reflected in `layer_data()`. 

I also had to make a few tweaks to the tests due to some subtle differences with using a `tibble`. However, I think these are mostly technical and that the spirit of each test is unchanged.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -56,7 +56,7 @@ core developer team.
 
 ## Minor improvements and bug fixes
 
-* `ggplot_build()` and `layer_data()` now return data as `tibble`s (@malcolmbarrett, #)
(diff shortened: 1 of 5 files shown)


yutannihilation:
Thanks, but this has been discussed on #3018 and declined.

malcolmbarrett:
Oops, missed that one. I did actually benchmark it because I was aware of the performance changes, and it was only ~2 microsecs different on my machine, but that was very limited. I'll close this for now

PR #3275, opened 2019-04-26
Move build-time calculations to .onLoad()

### LINKED ISSUE #3233: Remove build time code execution
One example:

https://github.com/tidyverse/ggplot2/blob/47d65ddbb92066513578846512f98355d9b5e0d9/R/theme-current.R#L3

but we need to scour the code base for others

thomasp85:
Fixes #3233

Two instances were observed and moved too `.onLoad()` after going through the full source code

### DIFF
--- a/R/zzz.r
+++ b/R/zzz.r
@@ -19,6 +19,10 @@
 .onLoad <- function(...) {
   backport_unit_methods()
 
+  .zeroGrob <<- grob(cl = "zeroGrob", name = "NULL")
+
+  ggplot_global$theme_current <- theme_gray()
+
   # To avoid namespace clash with dplyr.
   # It seems surprising that this hack works
   if (requireNamespace("dplyr", quietly = TRUE)) {
--- a/R/grob-null.r
+++ b/R/grob-null.r
@@ -4,7 +4,9 @@
 #' @export
 zeroGrob <- function() .zeroGrob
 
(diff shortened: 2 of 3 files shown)


R/zzz.r
```diff
@@ -19,6 +19,10 @@
 .onLoad <- function(...) {
   backport_unit_methods()
 
+  .zeroGrob <<- grob(cl = "zeroGrob", name = "NULL")
```
clauswilke: If I understand the documentation to `.onLoad()` correctly (https://stat.ethz.ch/R-manual/R-devel/library/base/html/ns-hooks.html), this should be `grid::grob`.

thomasp85: I don’t think so - where do you see that? `.onLoad()` is called after the namespace is loaded

clauswilke: I was commenting based on the following sentence. It's possible I misinterpreted it.

> Note that the code in `.onLoad` and `.onUnload` should not assume any package except the base package is on the search path. Objects in the current package will be visible (unless this is circumvented), but objects from other packages should be imported or the double colon operator should be used.

thomasp85: I think this is a throwback to ancient times when packages could assume default packages were loaded (e.g. stats) and thus didn't need to be imported in the namespace file... grid is imported by ggplot2 so it's all good

Issue #3214, opened 2019-03-28
stat.smooth: Add se.type option for "prediction" interval


baderstine:
https://github.com/tidyverse/ggplot2/blob/92d27779e210ce16f441eeec26926b8b045a47a8/R/stat-smooth-methods.r#L14

Suggest something like the following, where interval could take the value "confidence" or "prediction":

    predictdf.default <- function(model, xseq, se, level, se.type="confidence") {
        pred <- stats::predict(model, newdata = new_data_frame(list(x = xseq)), se.fit = se,
    level = level, interval = if (se) se.type else "none")
   ...

hadley:
I think this needs to be handled elsewhere given the generality of `geom_smooth()`, my regrets for adding a standard error by default (since its interpretation is so fraught), and the variation across model fitting methods.

PR #3277, opened 2019-04-27
add 'interval' argument to stat_smooth. addresses #3214

baderstine:
This is an initial attempt to address #3214 
Only applies to `predict.lm` calls.

Example usage where 'prediction' interval is used:
```
ggplot(mtcars, aes(x=hp, y=disp)) + 
  geom_point() + 
  geom_smooth(formula = y ~ poly(x,2), method="lm", se=T, interval="p") 
```
image

A potentially unfortunate side effect is that the following warning is always printed:

> Warning message:
> In predict.lm(model, newdata = new_data_frame(list(x = xseq)), se.fit = se,  :
>   Assuming constant prediction variance even though model fit is weighted

From the predict.lm documentation: "If the fit was weighted and newdata is given, the default is to assume constant prediction variance, with a warning."  
This warning _always_ displays because both `weights` and `newdata` are always provided:

https://github.com/tidyverse/ggplot2/blob/cd6f7cac7b1e27341cf2d08791f9c2e8bca90437/R/stat-smooth.r#L137

 https://github.com/tidyverse/ggplot2/blob/cd6f7cac7b1e27341cf2d08791f9c2e8bca90437/R/stat-smooth-methods.r#L13 

I could remove the warning message by conditionally passing a weight argument, however, it is not entirely clear what the weight values would be except in the simple case when all the weights were set to 1 by stat-smooth.r:
https://github.com/tidyverse/ggplot2/blob/cd6f7cac7b1e27341cf2d08791f9c2e8bca90437/R/stat-smooth.r#L112
It think it makes sense to address this specific case but I haven't done so yet:

For non-predict.lm models, I added a warning message to let the user know that `interval="prediction"` isn't supported and fall back on the default confidence interval:

> interval = 'prediction' not supported, using interval = 'confidence'

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+* `stat_smooth()` and `geom_smooth()` now supports prediction intervals using 
+  the `interval` argument when method = lm (@baderstine, #3214)
+
 # ggplot2 3.1.1.9000
 
 This is a minor release with an emphasis on internal changes to make ggplot2 
--- a/R/stat-smooth-methods.r
+++ b/R/stat-smooth-methods.r
@@ -6,12 +6,12 @@
 # @alias predictdf.glm
 # @alias predictdf.loess
 # @alias predictdf.locfit
-predictdf <- function(model, xseq, se, level) UseMethod("predictdf")
+predictdf <- function(model, xseq, se, level, interval) UseMethod("predictdf")
 
 #' @export
-predictdf.default <- function(model, xseq, se, level) {
+predictdf.default <- function(model, xseq, se, level, interval) {
   pred <- stats::predict(model, newdata = new_data_frame(list(x = xseq)), se.fit = se,
-    level = level, interval = if (se) "confidence" else "none")
+    level = level, interval = interval)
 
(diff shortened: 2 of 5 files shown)


clauswilke:
If this works only for `method = "lm"` then I'm not convinced it has sufficient generality to be included in the main ggplot2 distribution. I think glm and gam should be supported as well, and I see no technical reason why they can't. Loess may be a bit more difficult.

Also, we cannot have a geom create a warning when it is used exactly as intended, so the issue of the warning message would have to be addressed somehow.

baderstine:
All fair points.  "lm" was the low hanging fruit since there is a 'interval' argument in the `predict.lm()` method that seems a shame not to expose to users.  The others (gam, glm, loess) don't appear to be as straightforward, as their predict methods don't include the 'interval' argument. There are expansive discussions about these on SO that warrant further investigation.  Perhaps this is why 'interval' was not implemented?

hadley:
Yeah, I don't think ggplot2 is the right place for this code.

Issue #3090, opened 2019-01-19
geom_hline() and geom_vline() overwrite provided data frame


dariyasydykova:
This comes from a discussion with @clauswilke. The issue here is related to #2950 but somewhat different.

When a data frame is provided to `geom_hline()`, it is overwritten. The problem seems to be in the following code chunk:
https://github.com/tidyverse/ggplot2/blob/3c3062d1a37a3ab3f26dc9d85904fc4d72acde8a/R/geom-hline.r#L12-L17

Here is a reprex to demonstrate this issue:

This works as expected
``` r
library(ggplot2)
library(tidyr)

d <- data.frame(c = c("a", "b"),
                y = c(1, 3))

d %>% ggplot() +
  geom_hline(data = d, aes(yintercept = y, color = c)) +
  facet_wrap(~c)
```

![](https://i.imgur.com/2hweySP.png)

This doesn't work as expected. The provided data frame is dropped, and colors aren't mapped.
``` r

d %>% ggplot() +
  geom_hline(data = d, aes(color = c), yintercept = 1) +
  facet_wrap(~c)
```

![](https://i.imgur.com/36WEwjT.png)

<sup>Created on 2019-01-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

No warning is given when the original data frame is being overwritten.

PR #3286, opened 2019-05-01
redo warnings for overwritten arguments in abline, hline, and vline

paleolimbot:
This is a pull request to address #3090, and in part overwrites the work from #2950. This PR ensures that:

- An informative warning is issued when `mapping` and one or more of `yintercept`/`xintercept`/`slope`/`intercept` are provided. 
- An informative warning is issued when `data` and one or more of `yintercept`/`xintercept`/`slope`/`intercept` are provided. 

It is also worth noting that I changed the test for "data is provided" from `missing(data)` to `is.null(data)`. I am not sure if there was a good reason for it to have been `missing(data)` in the first place, as something like

```r
data <- NULL
geom_hline(yintercept = 3, data = data)
```

would possibly not issue a warning.

I may have gone overboard on the gramatically-correct informative warning message function (`warn_ovewritten_args()`), but was trying make all the messages informative without duplicating code.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,12 @@
+# ggplot2 3.2.0.9000
+
+## Minor improvements and bug fixes
+
+* `geom_abline()`, `geom_hline()`, and `geom_vline()` now issue 
+  more informative warnings when supplied with set aesthetics
+  (i.e., `slope`, `intercept`, `yintercept`, and/or `xintercept`)
+  and mapped aesthetics (i.e., `data` and/or `mapping`).
+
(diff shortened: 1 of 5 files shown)


thomasp85:
Is this intended for 3.2.0?

thomasp85:
Ah. Just make a new heading and a bullet then

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# ggplot2 3.2.0.9000
```
hadley: Can you please merge to get updated NEWS.md? And remove the heading?

PR #3307, opened 2019-05-07
default formula argument to NULL in geom_smooth()

### LINKED ISSUE #3205: consider making geom_smooth() formula = "auto" the default
In observing the code of my students, I can see that many have mental models of `geom_smooth()` that lead them to use it incorrectly. Their mental model is to use the `method` argument to specify the type of smoothing, and this is reasonable. It works fine for `method = "lm"` and `method = "loess"` but not `method = "gam"`, since the default `formula` works well with `"lm"` and `"loess"` but not `"gam"`. Why not have the default `formula` be `"auto"`, which would look at the specified `method` and use `y ~ x` for `"lm"` and `"loess"` and `y ~ s(x, bs = "cs")` for `"gam"`.

I realize that the formula used by `method = "auto"` for `"gam"` is explained in the third paragraph of the help entry for the `method` argument, but in practice this is not connecting with a large proportion of the users I'm seeing.

@skaltman

bfgray3:
fixes #3205


clauswilke:
Unfortunately you just missed the cutoff for the 3.2.0 release, so this will probably have to wait a while until 3.2.0 is out the door.

Generally looks good to me, but one issue I see is that I believe things will break if a method argument is provided (e.g., `method = "lm"`) and a formula is not. The `setup_params()` method should check for a `NULL` formula and substitute `y ~ x`, somewhere around here:
https://github.com/tidyverse/ggplot2/blob/508e1bc315789a7c3ae37226830da3647647ee25/R/stat-smooth.r#L102-L103

Also, to actually close #3205, you'd have to insert the default `gam` formula if formula is `NULL` in this `if` statement: 
https://github.com/tidyverse/ggplot2/blob/508e1bc315789a7c3ae37226830da3647647ee25/R/stat-smooth.r#L99-L101

R/stat-smooth.r
```diff
@@ -86,21 +88,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
 
       if (max_group < 1000) {
         params$method <- "loess"
+        params$formula <- params$formula %||% (y ~ x)
```
hadley: You're now setting the formula in two places — I don't think you need to set when figuring out which `method` to use, you can just set it once the method is known.

bfgray3: If the formula isn't set before the [message()](https://github.com/tidyverse/ggplot2/blob/master/R/stat-smooth.r#L93) call just below, then the message could include a `NULL` instead of the relevant formula, breaking [this test](https://github.com/tidyverse/ggplot2/blob/master/tests/testthat/test-geom-smooth.R#L27) for example.  Maybe a solution could be to have a logical keep track of whether or not `params$method` is originally `"auto"` and then move the `message()` call down until after the formula has been set once and wrap it in `if (using_auto_method)`, although that doesn't seem great to me.

hadley: Oh, I see the problem — now that `method` and `formula` are being set independently, you need to break that message up into two pieces. You should only ever inform the user when the parameter was chosen automatically, not when they supplied it.

R/stat-smooth.r
```diff
@@ -38,7 +40,7 @@ stat_smooth <- function(mapping = NULL, data = NULL,
                         geom = "smooth", position = "identity",
                         ...,
                         method = "auto",
-                        formula = y ~ x,
+                        formula = NULL,
```
hadley: I also now realise that this makes `method` and `formula` inconsistent. We should probably also default `method` to `NULL` (following https://principles.tidyverse.org/def-inform.html) and also allow `"auto"` for backward compatibility.

R/stat-smooth.r
```diff
@@ -86,21 +88,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
...
+      params$formula <- params$formula %||% (y ~ s(x, bs = "cs"))
+    } else {
+      params$formula <- params$formula %||% (y ~ x)
     }
 
     params
   },
 
-  compute_group = function(data, scales, method = "auto", formula = y~x,
+  compute_group = function(data, scales, method = "auto", formula = NULL,
```
hadley: This should probably have `method = "loess"` and `formula = y ~ x`

R/geom-smooth.r
```diff
@@ -91,7 +91,7 @@ geom_smooth <- function(mapping = NULL, data = NULL,
     ...
   )
   if (identical(stat, "smooth")) {
-    params$method <- method
+    params$method <- if (!is.null(method) && identical(method, "auto")) NULL else method
```
hadley: This check should be handled lower down so you don't need to repeat it in both `geom_smooth()` and `stat_smooth()`

R/stat-smooth.r
```diff
@@ -78,7 +82,8 @@ stat_smooth <- function(mapping = NULL, data = NULL,
 StatSmooth <- ggproto("StatSmooth", Stat,
 
   setup_params = function(data, params) {
-    if (identical(params$method, "auto")) {
+    msg <- NULL
```
hadley: I think it would be easier to do `msg <- character()`

R/stat-smooth.r
```diff
@@ -88,19 +93,36 @@ StatSmooth <- ggproto("StatSmooth", Stat,
         params$method <- "loess"
       } else {
         params$method <- "gam"
-        params$formula <- y ~ s(x, bs = "cs")
       }
-      message("`geom_smooth()` using method = '", params$method,
-              "' and formula '", deparse(params$formula), "'")
+
+      msg <- paste0("`geom_smooth()` using method = '", params$method, "'")
```
hadley: Then this becomes `msg <- c(msg, paste0("method ='", ...)`

R/stat-smooth.r
```diff
@@ -88,19 +93,36 @@ StatSmooth <- ggproto("StatSmooth", Stat,
...
+        msg <- paste0(msg, " and formula '", deparse(params$formula), "'")
+      }
     }
+
     if (identical(params$method, "gam")) {
       params$method <- mgcv::gam
     }
 
+    if (!is.null(msg)) {
+      message(msg)
```
hadley: Then here you do:

```R
if (length(msg) > 0) {
  message("geom_smooth() using ", paste0(msg, collapse = " and "))
}

thomasp85:
@Hadley any reason this hasn't been merged in yet?

hadley:
@thomasp85 Can you fix the style at https://github.com/tidyverse/ggplot2/pull/3307/files#diff- 10973f7c4485dfcd2b91934488f19f2aR101? Then I think you can merge.

PR #3320, opened 2019-05-08
Allow empty annotations.

### LINKED ISSUE #3317: Annotations with zero-length aesthetics are broken
In the revdep checks (#3303), an issue was identified where zero-length aesthetics in `annotate()` caused newly broken plots (**interflex** package).

```r
library(ggplot2)
ggplot() +
  annotate("point", x = numeric(0), y = numeric(0), colour = "red")
#> Error: Elements must equal the number of rows or 1
```

In 3.1.1, this did not error:

``` r
library(ggplot2)
ggplot() +
  annotate("point", x = numeric(0), y = numeric(0), colour = "red")
```

clauswilke:
This fixes #3317, and likely some other current issues.

``` r

library(ggplot2)
ggplot() +
  annotate("point", x = numeric(0), y = numeric(0), colour = "red")
```

<sup>Created on 2019-05-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/annotation.r
+++ b/R/annotation.r
@@ -48,8 +48,8 @@ annotate <- function(geom, x = NULL, y = NULL, xmin = NULL, xmax = NULL,
   lengths <- vapply(aesthetics, length, integer(1))
   n <- unique(setdiff(lengths, 1L))
   if (length(n) == 0L) n <- 1L # if all lengths are equal to 1L then above line fails, this fixes that
-  unequal <- length(n) > 1L
-  if (unequal) {
+
+  if (length(n) > 1L) {
     bad <- lengths != 1L
     details <- paste(names(aesthetics)[bad], " (", lengths[bad], ")",
       sep = "", collapse = ", ")
--- a/R/annotation.r
+++ b/R/annotation.r
@@ -46,14 +46,14 @@ annotate <- function(geom, x = NULL, y = NULL, xmin = NULL, xmax = NULL,
 
   # Check that all aesthetic have compatible lengths
   lengths <- vapply(aesthetics, length, integer(1))
+  n <- unique(lengths)
 
-  # To determine the final number of rows `n` in the data frame,
-  # we need to find the unique lengths not equal to 1L (and there
-  # should be at most one such length). However, if all lengths
-  # are equal to 1L, then the final number of rows is also 1L.
-  n <- unique(setdiff(lengths, 1L)) # unique lengths except 1L
-  if (length(n) == 0L) n <- 1L # all lengths are equal to 1L
+  # if there is more than one unique length, ignore constants
+  if (length(n) > 1L) {
+    n <- setdiff(n, 1L)
+  }
 
+  # if there is still more than one unique length, we error out
   if (length(n) > 1L) {
     bad <- lengths != 1L
     details <- paste(names(aesthetics)[bad], " (", lengths[bad], ")",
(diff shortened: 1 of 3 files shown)


clauswilke:
@hadley @yutannihilation @thomasp85 I would need a formal approval before I can merge this. It was Ok'd informally by Thomas here: https://github.com/tidyverse/ggplot2/issues/3317#issuecomment-491869632

tests/testthat/test-performance.R
```diff
@@ -1,5 +1,8 @@
 context("Performance related alternatives")
 
+# ********************
+# modify_list()
```
hadley: Can you please use our standard sectioning convention? 

```
# modify_list() ----------------------------------------
```

R/annotation.r
```diff
@@ -46,15 +46,17 @@ annotate <- function(geom, x = NULL, y = NULL, xmin = NULL, xmax = NULL,
 
   # Check that all aesthetic have compatible lengths
   lengths <- vapply(aesthetics, length, integer(1))
-  unequal <- length(unique(setdiff(lengths, 1L))) > 1L
-  if (unequal) {
+  n <- unique(setdiff(lengths, 1L))
+  if (length(n) == 0L) n <- 1L # if all lengths are equal to 1L then above line fails, this fixes that
```
hadley: Why does it fail?

clauswilke: Maybe a poorly formulated comment. The purpose of line 49 is to find the unique lengths that are not `1L`. However, if `lengths == c(1L)`, then we get `n <- numeric(0)` in line 49, and we need `n <- 1L` in this case.

R/annotation.r
```diff
@@ -46,15 +46,22 @@ annotate <- function(geom, x = NULL, y = NULL, xmin = NULL, xmax = NULL,
 
   # Check that all aesthetic have compatible lengths
   lengths <- vapply(aesthetics, length, integer(1))
-  unequal <- length(unique(setdiff(lengths, 1L))) > 1L
-  if (unequal) {
+
+  # To determine the final number of rows `n` in the data frame,
+  # we need to find the unique lengths not equal to 1L (and there
+  # should be at most one such length). However, if all lengths
+  # are equal to 1L, then the final number of rows is also 1L.
```
hadley: Maybe this would be a clearer approach?

```R
ulength <- unique(lengths)
unequal <- length(ulength) > 1 || ulength != 1
```

clauswilke: No, that doesn't capture it. `unequal` should only be true if there is more than one length *excluding* 1. So if `lengths <- c(1, 2)`, we should have `unequal <- FALSE`, but if `lengths <- c(2, 3)`, we should have `unequal <- TRUE`. I'll see if I can find a better way to code this.

The original statement to calculate `unequal` was exactly right:
> `unequal <- length(unique(setdiff(lengths, 1L))) > 1L`
But now I'm also trying to capture the length `n` in addition, that required the additional code.

clauswilke: I've thought about it some more, and I can't come up with something that would be clearer and simpler than the code in the current PR. Again, we also need to capture `n`, since it's used in line 64:
https://github.com/tidyverse/ggplot2/blob/4ff8243cf0215e5598045811dfc2ac94cd9a39ee/R/annotation.r#L64

clauswilke: When I woke up this morning, I realized that using `setdiff()` only when needed may make the code easier to understand. I pushed a new version.

PR #3321, opened 2019-05-09
Better error message when geom_boxplot() gets group data with more than 1 row 

### LINKED ISSUE #3316: geom_boxplot() with data rows >1 and stat = "identity" fails
In the revdep checks (#3303) at least one plot in an example failed to build that used `geom_boxplot(stat = "identity")` with more than one data row.

``` r
library(ggplot2)
ggplot(
  data.frame(x = "one value", y = 3, value = 4:6),
  aes(x, ymin = 0, lower = 1, middle = y, upper = value, ymax = 10)
) + 
  geom_boxplot(stat = "identity", position = "identity")
#> Error: Elements must equal the number of rows or 1
```

In 3.1.1, this works as expected (position dodge would make more sense here, but I wanted to make sure this wasn't the issue).

``` r
library(ggplot2)
ggplot(
  data.frame(x = "one value", y = 3, value = 4:6),
  aes(x, ymin = 0, lower = 1, middle = y, upper = value, ymax = 10)
) + 
  geom_boxplot(stat = "identity", position = "identity")
```

paleolimbot:
This is a PR to fix #3316, where group data with more than 1 row in `GeomBoxplot$draw_layer()` was causing an error that was difficult to diagnose. This may occur when using `geom_boxplot(stat = "identity")`, and is likely to occur if a user wants to draw several boxplots at specific locations on a plot in a boxplot/scatter mashup situation.

```r
library(ggplot2)
ggplot(
  data.frame(x = "one value", y = 3, value = 4:6),
  aes(x, ymin = 0, lower = 1, middle = y, upper = value, ymax = 10)
) + 
  geom_boxplot(stat = "identity", position = "identity")
#> Error: Elements must equal the number of rows or 1
```

### DIFF
--- a/R/geom-boxplot.r
+++ b/R/geom-boxplot.r
@@ -203,7 +203,7 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
     # this may occur when using geom_boxplot(stat = "identity")
     if (nrow(data) != 1) {
       stop(
-        "Can't draw more than one boxplot per group. Did you forget aes(group=...)?",
+        "Can't draw more than one boxplot per group. Did you forget aes(group = ...)?",
         call. = FALSE
       )
     }
--- a/R/geom-boxplot.r
+++ b/R/geom-boxplot.r
@@ -200,6 +200,14 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
                         outlier.alpha = NULL,
                         notch = FALSE, notchwidth = 0.5, varwidth = FALSE) {
 
+    # this may occur when using geom_boxplot(stat = "identity")
+    if (nrow(data) != 1) {
+      stop(
+        "Can't draw more than one boxplot per group. Did you forget aes(group=...)?",
+        call. = FALSE
+      )
+    }
+
     common <- list(
       colour = data$colour,
       size = data$size,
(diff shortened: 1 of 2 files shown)


R/geom-boxplot.r
```diff
@@ -200,6 +200,14 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
                         outlier.alpha = NULL,
                         notch = FALSE, notchwidth = 0.5, varwidth = FALSE) {
 
+    # this may occur when using geom_boxplot(stat = "identity")
+    if (nrow(data) != 1) {
+      stop(
+        "Can't draw more than one boxplot per group. Did you forget aes(group=...)?",
```
clauswilke: I think there should be a space before and after the = sign.

clauswilke:
Sorry, one more issue: Could you amend the latest commit message so it includes "Closes #3316" or "Fixes #3316"? This will allow github to automatically link the commit to the bug and it will close the bug upon merge.

Example: https://github.com/tidyverse/ggplot2/pull/3320/commits/cfd4962fe01c5cd6546a8072dd2b70fdba863469

clauswilke:
Did the force push undo the commit that fixed the spacing?

Also, one issue that may or may not be related: I'm seeing a lot of github activity with strange time stamps (events occurring in the future, or in the wrong order). Is your machine set up to use local time? It's generally best to set up the machine time to UTC and then tell the machine what time zone you're in, rather than running it on local time.

Issue #3304, opened 2019-05-06
Release notes for 3.2.0 need to mention possible failures of downstream visual tests


clauswilke:
See: https://github.com/tidyverse/ggplot2/pull/3303#issuecomment-489648068

There have been subtle changes to text rendering, and this may cause vdiffr changes in downstream packages. The release notes should state that.

yutannihilation:
#3050 also introduces visual changes.

clauswilke:
@paleolimbot Could you contact the authors of the packages who will have to update their vdiffr reference images?

paleolimbot:
Done!

- **bayesplot**: https://github.com/stan-dev/bayesplot/issues/189
- **blorr**: https://github.com/rsquaredacademy/blorr/issues/88
- **contextual**: https://github.com/Nth-iteration-labs/contextual/issues/2
- **cowplot**: https://github.com/wilkelab/cowplot/issues/134
- **crplyr**: https://github.com/Crunch-io/crplyr/issues/15
- **customLayout**: https://github.com/zzawadz/customLayout/issues/8
- **dabestr**: (emailed maintainer)
- **descriptr**: https://github.com/rsquaredacademy/descriptr/issues/101
- **easyalluvial**: https://github.com/erblast/easyalluvial/issues/12
- **epiflows**: https://github.com/reconhub/epiflows/issues/38
- **fingertipscharts**: https://github.com/PublicHealthEngland/fingertipscharts/issues/11
- **getTBinR**: https://github.com/seabbs/getTBinR/issues/72
- **ggExtra**: https://github.com/daattali/ggExtra/issues/141
- **ggformula**: https://github.com/ProjectMOSAIC/ggformula/issues/122
- **gggenes**: https://github.com/wilkox/gggenes/issues/11
- **ggperiodic**: https://github.com/eliocamp/ggperiodic/issues/7
- **ggridges**: https://github.com/clauswilke/ggridges/issues/34
- **ggstance**: https://github.com/lionel-/ggstance/issues/32
- **ggthemes**: https://github.com/jrnold/ggthemes/issues/113
- **ggvoronoi**: https://github.com/garretrc/ggvoronoi/issues/10
- **gratia**: https://github.com/gavinsimpson/gratia/issues/40
- **hrbrthemes**: https://github.com/hrbrmstr/hrbrthemes/issues/40
- **incidence**: https://github.com/reconhub/incidence/issues/114
[...]

Issue #3318, opened 2019-05-08
geom_ribbon() now errors when ymin and ymax aesthetics are set rather than mapped


paleolimbot:
This was identified in the revdep checks (#3303), as it breaks a plot in **bayesAB**. This error was introduced on purpose (#3235). It's worth noting that the behaviour of layers when all the required aesthetics are set and not mapped is undefined and not consistent between geometries. See https://github.com/tidyverse/ggplot2/pull/3303#issuecomment-489715276 , https://github.com/tidyverse/ggplot2/pull/3235#issuecomment-482605641 , and https://github.com/tidyverse/ggplot2/pull/3303#issuecomment-489662707 .

``` r
library(ggplot2)
ggplot() + 
  geom_ribbon(x = 1:5, ymin = 0, ymax = 1:5, data = data.frame(1:5)) +
  geom_blank(aes(x = 1:5, y = 1:5))
#> Error: Either ymin or ymax must be given as an aesthetic.
```

Whereas in 3.1.1:

``` r
library(ggplot2)
ggplot() + 
  geom_ribbon(x = 1:5, ymin = 0, ymax = 1:5, data = data.frame(1:5)) +
  geom_blank(aes(x = 1:5, y = 1:5))
```

![](https://i.imgur.com/5IDYe9g.png)

clauswilke:
Per @thomasp85 (https://github.com/tidyverse/ggplot2/pull/3303#issuecomment-489715276), we should contact the bayesAB authors to fix their code.

@paleolimbot Could you do that?

paleolimbot:
Done! https://github.com/FrankPortman/bayesAB/issues/33

clauswilke:
Thanks!

Issue #3305, opened 2019-05-06
Annotation breaks with facets


clauswilke:
As discussed here: https://github.com/tidyverse/ggplot2/pull/3303#issuecomment-489675095

Annotations break when there are multiple facets and a non-position aesthetic/parameter is provided as a vector:
``` r
library(ggplot2)
data <- data.frame(x = c(1, 2), y = c(1, 2), facet = c("a", "b"))
p <- ggplot(data, aes(x, y)) + geom_point() + facet_grid(vars(facet))
p + annotate("text", x = 1.5, y = 1.5, label = c("label1", "label2"))
#> Error: Aesthetics must be either length 1 or the same as the data (4): label
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

I note that this problem does not occur when a position aesthetic is provided as a vector:

``` r
library(ggplot2)
data <- data.frame(x = c(1, 2), y = c(1, 2), facet = c("a", "b"))
p <- ggplot(data, aes(x, y)) + geom_point() + facet_grid(vars(facet))
p + annotate("text", x = c(1, 2), y = 1.5, label = "label")
```

![](https://i.imgur.com/ERNQOE7.png)

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

As an additional hint to what may be happening, I note that the data frame that is internally being generated is twice as long as the input vector, as we can tell from the error message:

> `...same as the data (4)...`

If we try to provide a longer label vector, we get an even longer data frame:

``` r
library(ggplot2)
data <- data.frame(x = c(1, 2), y = c(1, 2), facet = c("a", "b"))
p <- ggplot(data, aes(x, y)) + geom_point() + facet_grid(vars(facet))
p + annotate("text", x = 1.5, y = 1.5, label = c("label1", "label2", "label3", "label4"))
#> Error: Aesthetics must be either length 1 or the same as the data (8): label
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

I don't immediately know what the correct behavior here should be. Since we're not in any way specifying facets in the annotation, how should ggplot2 know that one label should be placed onto each facet, rather than both labels on each facet?

clauswilke:
Some additional thoughts: `annotate()` only transfers position values into the data frame it creates:
[...]

PR #3324, opened 2019-05-10
Improved release notes

clauswilke:
I added a paragraph about visual changes to the release notes, as discussed in #3304.

@hadley @batpigandme Could you review? I wasn't entirely sure where to place the paragraph, but I feel it should be up on top so people see it even if they don't read through the entire set of notes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,12 @@ extension developers if they have relied on internals that have been changed.
 This release also sees the addition of Hiroaki Yutani (@yutannihilation) to the 
 core developer team.
 
+Two patches (#2996 and #3050) fixed minor rendering problems. In most cases, the
+visual changes are so subtle that they are difficult to see with the naked eye.
+However, these changes are detected by the vdiffr package, and therefore any
+package developers who use vdiffr to test for visual correctness of ggplot2 plots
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -7,6 +7,12 @@ extension developers if they have relied on internals that have been changed.
 This release also sees the addition of Hiroaki Yutani (@yutannihilation) to the 
 core developer team.
 
+Two patches (#2996 and #3050) fixed minor rendering problems. In most cases, the
```
hadley: Maybe put under `## Breaking changes`?

clauswilke: That works for me. I remember we had some discussion in the past about what exactly a breaking change is, but it feels like a breaking change to me. I also added a bullet point where I'm trying to express that ggplot2 is now somewhat stricter and now creates errors or warnings where previously it would just create some arbitrary output. Not sure I'm totally happy with this. I'm referring to issues such as #3316, #3305, #3318.

clauswilke: I added the specific examples to the bullet point. I think that makes it better.

hadley: I'd say that breaking a revdep check is sufficient (but not necessary) for a change to be breaking.

Issue #3315, opened 2019-05-08
Differences between plyr functions and functions in compat-plyr.R are causing failures


paleolimbot:
This is a revdep issue (#3303) in which the functions in `compat-plyr.R` are returning slightly different output than their former **plyr** counterparts, which in turn cause package failures.

- Different output column ordering between `plyr::ddply` and `ggplot2:::dapply` results in a different order of evaluation, which causes random jitters to be slightly different between **ggplot2** and **ggstance** (which is still using plyr). **gstance** tests for consistency between `position_jitterdodge()` and `position_jitterdodgev()`, which is why this results in a revdep failure.

``` r
df <- data.frame(y = 1:4, by_col = c("a", "b"))
plyr::ddply(df, "by_col", identity)
#>   y by_col
#> 1 1      a
#> 2 3      a
#> 3 2      b
#> 4 4      b
ggplot2:::dapply(df, "by_col", identity)
#>   by_col y
#> 1      a 1
#> 2      a 3
#> 3      b 2
#> 4      b 4
```

- Different (reversed order of) type coercion  in `rbind_dfs()` is causing the behaviour of `join_keys()` to differ from `plyr::join.keys()`. This results in the failure of faceting by some S3 vectors (#3313).

``` r
df <- data.frame(date = as.Date("2019-01-01"))
df_fct <- data.frame(date = as.factor(df$date))
unlist(ggplot2:::join_keys(df, df_fct, "date"))
#> x y n 
#> 1 1 1
unlist(ggplot2:::join_keys(df_fct, df, "date"))
#> x y n 
#> 2 1 2

unlist(plyr::join.keys(df, df_fct, "date"))
#> x y n 
#> 2 1 2
unlist(plyr::join.keys(df_fct, df, "date"))
#> x y n 
#> 1 1 1
```

clauswilke:
@thomasp85 Is it possible to rewrite the internal ggplot2 functions so they mirror the output from the plyr functions? If it's possible without too much effort then I think we should do it.

thomasp85:
It should be (and I thought it was before the revdepchecks)

thomasp85:
closed by #3327

PR #3327, opened 2019-05-14
Fix ordering in plyr compats...

### LINKED ISSUE #3313: facet_wrap() and facet_grid() with a date variable is broken
The revdeps (#3303) identified that using `facet_grid()` or `facet_wrap()` with a date variable is broken. This probably affects other S3 vectors as well.

``` r
library(ggplot2)
df <- data.frame(date = as.Date("2019-01-01"), x = 1, y = 1)
p + facet_grid(vars(date))
#> Error in scale_apply(layer_data, x_vars, "train", SCALE_X, x_scales):
p + facet_wrap(vars(date))
#> Error in scale_apply(layer_data, x_vars, "train", SCALE_X, x_scales):
```

The error is in `Facet(Grid|Wrap)$map_data()`, where there is a join between the layout and the panels. In one of these, `date` is a factor, and in the other, it is an S3 `Date`. This changed because of different behaviour between `join_keys()` and its previous **plyr** counterpart.

thomasp85:
Fixes #3313 and #3315 

This fixes the differences in ordering between plyr and the replacement functions. Some notes

- the plyr ordering in `ddply()` is based on a pretty weird heuristic but I think I've nailed it
- the difference in `join_keys()` is the result of differences in `rbind_dfs()` and `plyr::rbind.fill()`. I honestly think the current ggplot2 approach is more correct, but I've made Date handling even with plyr to unbreak revdeps... If that fixes everything we should leave it at that instead of pursuing other parity with other weird vector formats

I probably don't have a head for more work this week, so bear with me if there are any problems with the PR

### DIFF
--- a/R/compat-plyr.R
+++ b/R/compat-plyr.R
@@ -292,12 +292,16 @@ rbind_dfs <- function(dfs) {
     allocated[new_columns] <- TRUE
     if (all(allocated)) break
   }
+  is_date <- lapply(out, inherits, 'Date')
   pos <- c(cumsum(nrows) - nrows + 1)
   for (i in seq_along(dfs)) {
     df <- dfs[[i]]
     rng <- seq(pos[i], length.out = nrows[i])
     for (col in names(df)) {
-      if (inherits(df[[col]], 'factor')) {
+      date_col <- inherits(df[[col]], 'Date')
+      if (is_date[[col]] && !date_col) {
+        out[[col]][rng] <- as.Date(unclass(df[[col]]), origin = date_origin)
+      } else if (date_col || inherits(df[[col]], 'factor')) {
         out[[col]][rng] <- as.character(df[[col]])
       } else {
         out[[col]][rng] <- df[[col]]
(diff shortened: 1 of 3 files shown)


yutannihilation:
It seems this test fails:

https://github.com/tidyverse/ggplot2/blob/291cd7e826a2581d1fa6013d8840fe2ac31ddc76/tests/testthat/test-layer.r#L59-L64

Error:

```
names(p[[1]]) not identical to c("PANEL", "x", "group").
2/3 mismatches
x[1]: "x"
y[1]: "PANEL"

x[2]: "PANEL"
y[2]: "x"
```

This ordering was tweaked in #3013, so I guess you just need to revert this.

https://github.com/tidyverse/ggplot2/pull/3013/files#diff-12165842aeed0c5cb8c55e833132b58c

R/compat-plyr.R
```diff
@@ -310,6 +314,7 @@ rbind_dfs <- function(dfs) {
   attributes(out) <- list(class = "data.frame", names = names(out), row.names = .set_row_names(total))
   out
 }
+date_origin <- Sys.Date() - unclass(Sys.Date())
```
clauswilke: Is this the kind of global assignment that should be moved to `.onLoad()`?

thomasp85: Oh, yeah. Good catch 🙂

PR #3303, opened 2019-05-04
updated revdeps

topepo:
[database file](https://www.dropbox.com/s/llilyu97detb5xv/data.sqlite?dl=0)

We checked 2622 reverse dependencies (2212 from CRAN + 410 from BioConductor), comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 30 new problems
 * We failed to check 70 packages

Issues with CRAN packages are summarised below.

### New problems
(This reports the first line of each new failure)

* bayesAB
  checking examples ... ERROR

* breathteststan
  checking examples ... ERROR

* CSTools
  checking tests ...

* esmisc
  checking examples ... ERROR

* fergm
  checking examples ... ERROR

* fingertipscharts
  checking examples ... ERROR
  checking tests ...

* ggalt
  checking whether package ‘ggalt’ can be installed ... WARNING

* ggforce
  checking whether package ‘ggforce’ can be installed ... WARNING

* ggpol
  checking whether package ‘ggpol’ can be installed ... WARNING

* ggpolypath
  checking whether package ‘ggpolypath’ can be installed ... WARNING

* ggsolvencyii
  checking whether package ‘ggsolvencyii’ can be installed ... WARNING

* ggspatial
  checking whether package ‘ggspatial’ can be installed ... WARNING

* ggstance
  checking tests ...

* ggstatsplot
  checking tests ...

* ggtern
  checking whether package ‘ggtern’ can be installed ... WARNING

* heatmaply
  checking installed package size ... NOTE

* HistDAWass
  checking examples ... ERROR

* incR
  checking examples ... ERROR

* interflex
  checking examples ... ERROR

* jcolors
  checking examples ... ERROR

* lime
  checking tests ...

* mlr
  checking examples ... ERROR

* MTLR
  checking tests ...

* SCORPIUS
  checking tests ...

* shiny
  checking tests ...

* ssMousetrack
  checking installed package size ... NOTE
  checking for GNU extensions in Makefiles ... NOTE

* stats19
  checking tests ...

* trialr
  checking installed package size ... NOTE
  checking dependencies in R code ... NOTE
  checking for GNU extensions in Makefiles ... NOTE

* vdiffr
  checking tests ...

* xpose
  checking examples ... ERROR
  checking tests ...

### Failed to check

* aslib                  (NA)
* BACA                   (NA)
* BACCT                  (NA)
* bamdit                 (NA)
* BayesRS                (NA)
* BNSP                   (NA)
* BPEC                   (NA)
[...]

PR #3331, opened 2019-05-15
Dynamically assign pathGrob in .onLoad()

### LINKED ISSUE #3313: facet_wrap() and facet_grid() with a date variable is broken

paleolimbot:
This is a PR to fix #3313, where packages that subclass `GeomPolygon` were giving install warnings with the new version of ggplot2 in the revdep checks (#3303).

I was hesitant to prepare this PR, because I don't understand why this works, and because I can't replicate the install warning that occurred in the revdep checks, I can't verify that this PR fixes the problem. I can verify that it doesn't cause any CMD check failures for either ggplot2 or ggspatial, which was one of the packages affected by the revdep check issue.

### DIFF
--- a/R/zzz.r
+++ b/R/zzz.r
@@ -16,8 +16,20 @@
   })
 }
 
+# Assigning pathGrob in .onLoad ensures that packages that subclass GeomPolygon
+# do not install with error `possible error in 'pathGrob(munched$x, munched$y, ':
+# unused argument (pathId = munched$group)` despite the fact that this is correct
+# usage
+pathGrob <- NULL
+
 .onLoad <- function(...) {
   backport_unit_methods()
+
+  if (getRversion() < as.numeric_version("3.6")) {
+    pathGrob <<- function(..., pathId.lengths) {
+      grid::pathGrob(...)
+    }
+  }
 
   .zeroGrob <<- grob(cl = "zeroGrob", name = "NULL")
 


clauswilke:
If I understand correctly what is happening, the install warnings should only occur if you're installing with R version 3.5 or earlier. Do you have an R 3.5 environment with which you could test this?

R/zzz.r
```diff
@@ -16,9 +16,21 @@
   })
 }
 
+# Assigning pathGrob in .onLoad ensures that packages that subclass GeomPolygon
+# do not install with error `possible error in 'pathGrob(munched$x, munched$y, ':
+# unused argument (pathId = munched$group)` despite the fact that this is correct
+# usage
+pathGrob <- NULL
```
clauswilke: Just for my own understanding: Why do we need this assignment outside of `.onLoad()`?

paleolimbot: I copied what was in sf:::.onLoad() that @thomasp85 suggested we use as a reference. My guess is that it prevents the superassign from accidentally assigning in the global environment (or prevents a warning to that effect)

clauswilke:
Thanks!

PR #3340, opened 2019-05-23
Fix position_dodge2() example

karawoo:
The default behavior was changed in #2386 but the example wasn't updated.

I've opened this against master, but maybe it could go in the 3.2.0 release? @thomasp85 let me know if you want this PR against the release candidate branch instead.


clauswilke:
Looks good to me, but could you regenerate the docs? And since it's just a change in the docs, maybe we can still squeeze it into 3.2.0.

PR #3346, opened 2019-05-30
Warn about discouraged aes() usage during plot build

### LINKED ISSUE #2693: Warn about use of $ in aes()
Quite a few people seem to do this, and it introduces potentially subtle bad behaviour when you start facetting. I think we now have sufficient understanding to detect this correctly, although we’ll need to carefully consider how to override `$` so ok usage is not caught.

paleolimbot:
This is a PR that implements warnings when the layer `mapping` contains syntax like `data$column` instead of `.data$column` or (even better) `column`. This PR also implements an experimental and very open to discussion warning about layers that never map any columns from the data. Collectively, this is designed to encourage the correct usage of data and mappings, particularly in packages where this is sometimes done to avoid non-standard evaluation. This PR fixes #2693.

Warnings for `$` and `[[` usage in `aes()`:

``` r
library(ggplot2)

build <- function(p) invisible(ggplot_build(p))

p <- ggplot(mpg, aes(cty, hwy)) + geom_point()

# these do not issue warnings
build(p)
build(p + aes(.data$cty))
build(p + aes(.data[["cty"]]))

# these issue warnings
build(p + aes(mpg$cty))
#> Warning: Use of `mpg$cty` is discouraged. Use `cty` instead.
build(p + aes(mpg[["cty"]]))
#> Warning: Use of `mpg[["cty"]]` is discouraged. Use `.data[["cty"]]`
#> instead.
mpg2 <- mpg
build(p + aes(mpg2$cty))
#> Warning: Use of `mpg2$cty` is discouraged. Use `cty` instead.
```

<sup>Created on 2019-05-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

As discussed in #2693, it is probably going too far to check for zero column mappings. I implemented it here just for discussion...it causes 5 warnings in the test cases (all of which fit the description of zero mapped columns from data).

``` r
library(ggplot2)

build <- function(p) invisible(ggplot_build(p))

p <- ggplot(mpg, aes(cty)) + geom_histogram(bins = 30)

# these do not issue warnings
build(p)
build(p + aes(.data$cty))
build(p + aes(.data[["cty"]]))
build(p + aes(cty, fill = 1))

# this doesn't issue a warning about the mappings
build(p + aes(mpg$cty))
#> Warning: Use of `mpg$cty` is discouraged. Use `cty` instead.

# these issue warnings
build(p + aes(1))
#> Warning: Mapping contains zero mapped columns from data
build(p + aes(runif(nrow(mpg))))
#> Warning: Mapping contains zero mapped columns from data
```

<sup>Created on 2019-05-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


clauswilke:
I don't see a good reason to issue warnings on zero mapped columns. I think it'll get in the way of meaningful use cases, in particular annotations.

``` r
library(ggplot2)
ggplot() + geom_text(aes(x = 1, y = 1, label = "annotation"))
```



<sup>Created on 2019-05-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/aes.r
```diff
@@ -334,3 +334,114 @@ mapped_aesthetics <- function(x) {
...
+quo_column_refs <- function(quosure, data) {
+  expr_column_refs(get_expr(quosure), data, get_env(quosure))
+}
+
+expr_column_refs <- function(x, data, env = emptyenv()) {
+  if (is.name(x) && (as.character(x) %in% names(data))) {
+    as.character(x)
+  } else if (is_call(x, "[[") && extract_target_is_quo_data(x, data, env)) {
+    # in extract calls from .data, the index is not overscoped with the data
+    index_value <- try(eval_tidy(x[[3]], data = NULL, env), silent = TRUE)
```
hadley: Why do you try and evaluate here?

R/aes.r
```diff
@@ -334,3 +334,114 @@ mapped_aesthetics <- function(x) {
...
+
+#' Check a mapping for discouraged usage
+#'
+#' @param mapping A mapping created with [aes()]
+#' @param data The data to be mapped from
+#'
+#' @noRd
+check_aes <- function(mapping, data) {
+  check_aes_extract_usage(mapping, data)
+  check_aes_column_refs(mapping, data)
```
hadley: Let's drop `check_aes_column_refs()` from consideration for now.

R/aes.r
paleolimbot: As per your later comment I dropped this bit from the PR, but I had been evaluating there to get the index value (rather than the expression between `[[` and `]]`. I used `try()` to minimize the chance of an error getting generated in the process.

R/aes.r
```diff
@@ -334,3 +334,51 @@ mapped_aesthetics <- function(x) {
...
+#'
+#' @param mapping A mapping created with [aes()]
+#' @param data The data to be mapped from
+#'
+#' @noRd
+check_aes_extract_usage <- function(mapping, data) {
+  lapply(mapping, check_aes_extract_usage_quo, data)
+}
+
+check_aes_extract_usage_quo <- function(quosure, data) {
```
hadley: I think you could eliminate this function now (and inline into `check_aes_extract_usage()`)

R/aes.r
```diff
@@ -334,3 +334,51 @@ mapped_aesthetics <- function(x) {
...
+        "Use `", good_usage,  "` instead.",
+        call. = FALSE
+      )
+    }
+  } else if (is.call(x)) {
+    lapply(x, check_aes_extract_usage_expr, data, env)
+  }
+}
+
+check_aes_get_alternative_usage <- function(x) {
```
hadley: `check_` implies that this might throw an error/warning; maybe call it something like `find_correct_usage()`

R/aes.r
```diff
@@ -334,3 +334,51 @@ mapped_aesthetics <- function(x) {
...
+  lapply(mapping, check_aes_extract_usage_quo, data)
+}
+
+check_aes_extract_usage_quo <- function(quosure, data) {
+  check_aes_extract_usage_expr(get_expr(quosure), data, get_env(quosure))
+}
+
+check_aes_extract_usage_expr <- function(x, data, env = emptyenv()) {
+  if (is_call(x, "[[") || is_call(x, "$")) {
+    if (extract_target_is_data(x, data, env)) {
```
hadley: Only check here if `x[[2]]` is a symbol — I don't think we need to handle cases like `foo(1,2,3)$x`

R/aes.r
```diff
@@ -334,3 +334,51 @@ mapped_aesthetics <- function(x) {
...
+    good_call <- call2("[[", quote(.data), x[[3]])
+    format(good_call)
+  } else if (is_call(x, "$")) {
+    as.character(x[[3]])
+  } else {
+    stop("Don't know how to get alternative usage for `", format(x), "`", call. = FALSE)
+  }
+}
+
+extract_target_is_data <- function(x, data, env) {
```
hadley:
```
extract_target_is_data <- function(x, data, env) {
  tryCatch({
    data_eval <- eval_tidy(x[[2]], data, env)
    identical(data_eval, data)  
  }, error = function(err) FALSE)
}
```

tests/testthat/test-aes.r
```diff
@@ -111,6 +111,41 @@ test_that("aes standardises aesthetic names", {
...
+test_that("Improper use of $ and [[ is detected by check_aes_extract_usage()", {
+
+  returns_x <- function() "x"
+  df <- data_frame(x = 1:5, nested_df = data_frame(x = 6:10))
+
+  # valid extraction in aes()
+  expect_silent(check_aes_extract_usage(aes(x), df))
+  expect_silent(check_aes_extract_usage(aes(.data$x), df))
+  expect_silent(check_aes_extract_usage(aes(.data[["x"]]), df))
+  expect_silent(check_aes_extract_usage(aes(.data[[!!quo("x")]]), df))
```
hadley: I think you can remove the examples with `!!` because `aes()` does the unquoting, so `check_aes_extract_usage()` never sees them.

tests/testthat/test-aes.r
```diff
@@ -111,6 +111,41 @@ test_that("aes standardises aesthetic names", {
...
+  expect_silent(check_aes_extract_usage(aes(.data[[returns_x()]]), df))
+  expect_silent(check_aes_extract_usage(aes(!!sym("x")), df))
+  expect_silent(check_aes_extract_usage(aes(x * 10), df))
+  expect_silent(check_aes_extract_usage(aes(nested_df$x), df))
```
hadley: I'd also introduce a few comments mentioning dividing this into sections.

tests/testthat/test-aes.r
```diff
@@ -111,6 +111,41 @@ test_that("aes standardises aesthetic names", {
   expect_warning(aes(color = x, colour = y), "Duplicated aesthetics")
 })
 
+test_that("Improper use of $ and [[ is detected by check_aes_extract_usage()", {
+
+  returns_x <- function() "x"
+  df <- data_frame(x = 1:5, nested_df = data_frame(x = 6:10))
+
+  # valid extraction in aes()
```
hadley: These should go in a separate test: titles something about false positives.

hadley:
Worth checking to see if this flags problems with any existing examples or vignettes.

And worth adding a sentence or two to the parameter documentation of `aes()`.  Something like: Aesthetics are automatically looked for in the plot data; do no specific like `aes(x = mydf$variable1)`

R/aes.r
```diff
@@ -22,9 +22,10 @@ NULL
 #' programming vignette](http://dplyr.tidyverse.org/articles/programming.html)
 #' to learn more about these techniques.
 #'
-#' @param x,y,... List of name value pairs giving aesthetics to map to
-#'   variables. The names for x and y aesthetics are typically omitted because
-#'   they are so common; all other aesthetics must be named.
+#' @param x,y,... List of name-value pairs in the form `aesthetic = column_name`
+#'   describing which variables in the layer data should be mapped to which
+#'   aesthetics used by paired geom/stat. The names for x and y aesthetics are typically
+#'   omitted because they are so common; all other aesthetics must be named.
```
hadley: I do think it's worth being explicit about the wrong way to do things here. Maybe start a new paragraph and say something like: It's not necessary to refer to the dataset in `aes()`. Instead of writing `aes(x = data$myvariable)`, all you need is `aes(myvariable)`.

Maybe @mine-cetinkaya-rundel has thoughts on this?

mine-cetinkaya-rundel: I agree with @hadley that it's good to be explicit about what not to do, especially for those coming to ggplot2 from `plot()` where `data$column_name` usage is common. I might suggest matching the wording to the warning and say 

"It's discouraged to refer to the data in `aes()`. Instead of aes(x = data$variable), use aes(variable)." 

and similarly use `aesthetic = variable` earlier in the description (or use `column` or `column_name` or `variable_name` but keep it consistent throughout). I *think* it's more correct to avoid `_name` since it's not the name (character string) but the variable itself we're referring to, but I'm not sure.

paleolimbot: I went for a mashup of both your suggestions and it feels much better now. Thank you!

Issue #3322, opened 2019-05-09
Export guide_axis() and consider improvements


hadley:
Currently we have no equivalent to `guide_legend()` for axes (there is an non-exported `guide_axis()` but it draws the axes rather than describing their parameters). We should have an exported guide object to make it possible to have new types of axes.

The build in `guide_axes()` also needs to gain better tools for handling long labels:

* Should have`angle` parameter which handles all related theme settings
* Should optionally trim overlapping labels by starting and either end and bisecting inwards (cf `check.overlap = TRUE` in `grid.text()`
* Should be able to draw multiple lines of labels, alternating the position to avoid overlaps 

And generally look at what [vega-lite](https://vega.github.io/vega-lite/docs/axis.html) does for other ideas of useful parameters.

paleolimbot:
I spent about a day and a half on this, and wanted to leave a few things here for when I pick it back up again. So far I've:

- Refactored `guide_axis()` to `draw_axis()`, since that's what the current version actually does (creates a grob)
- Created a `guide_axis()` implementation that mimics `guide_legend()` and changed `draw_axis()` to create a guide, train it, then draw it.
- Rewrote the grob creation code so that it might be easier to add features to (in a way that doesn't cause any dopplegangers to change)

https://github.com/paleolimbot/ggplot2/blob/0267a0aa48863b119f8288f95eac6fe87d89d95f/R/guides-axis.r#L125-L256

I still have to (provided all agree this is a good idea)

- Refactor scale and guide code to treat position guides like other guides so they can be specified within `scale_x_*()`, `scale_y_*()`, `sec_axis()`, and `guides()`
- Refactor Coord code to use guides the guides specified in `scale_x_*()`, `scale_y_*()`, `sec_axis()`, and `guides()` instead of using `draw_axis()`
- Collect input on what (if any) improvements should go in `guide_axis()`, or if they should go in some other guide function (`guide_axis_fancy()`, or the like)

For reference, the [vega-lite axis label documentation](https://vega.github.io/vega-lite/docs/axis.html#labels), which includes an option in inward-align labels on the end.

clauswilke:
Would this be an opportunity to move the entire guide system to ggproto? It's the one remaining major part of ggplot2 that doesn't use ggproto, and it's a bit of a mess and difficult to extend by external package developers.

hadley:
I think it's probably worth making that a separate issue.

clauswilke:
[...]

PR #3349, opened 2019-05-31
Rewrite guide_axis() and rename to draw_axis()

paleolimbot:
This is a PR to prepare for improvements in axis guides as part of #3322. This PR  renames the internal function `guide_axis()` to `draw_axis()` and includes a complete rewrite of (what is now) `draw_axis()`.

Renaming `guide_axis()` to `draw_axis()` is necessary to make way for an exported `guide_axis()` function that returns a guide (instead of a grob). I used the term "draw" because it is used in many places in ggplot2 for functions that return grobs.

The rewrite of `draw_axis()` is necessary because the current code is difficult to read and contains much duplication. This makes it difficult to change or improve, as is planned for #3322. The rewrite does not introduce any visual changes in the current tests, and I didn't add any new tests because I couldn't find any part of axis appearance that wasn't already tested between the guides and themes.

### DIFF
--- a/R/theme.r
+++ b/R/theme.r
@@ -612,21 +612,31 @@ merge_element.element <- function(new, old) {
   new
 }
 
-# Combine the properties of two elements
-#
-# @param e1 An element object
-# @param e2 An element object which e1 inherits from
+#' Combine the properties of two elements
+#'
+#' @param e1 An element object
+#' @param e2 An element object from which e1 inherits
+#'
+#' @noRd
+#'
 combine_elements <- function(e1, e2) {
 
   # If e2 is NULL, nothing to inherit
   if (is.null(e2) || inherits(e1, "element_blank"))  return(e1)
+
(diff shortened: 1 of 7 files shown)


clauswilke:
I'm Ok with the change in `combine_elements()`. The current way of doing things where code all over the place does it's own inheritance calculations is bad. All of this should be centralized and better formalized.

R/theme.r
```diff
@@ -612,21 +612,31 @@ merge_element.element <- function(new, old) {
...
-#
-# @param e1 An element object
-# @param e2 An element object which e1 inherits from
+#' Combine the properties of two elements
+#'
+#' @param e1 An element object
+#' @param e2 An element object from which e1 inherits
+#'
+#' @noRd
+#'
```
hadley: Need to bring this function up to code, since you've touched. (In this case, that's just restyling) — https://style.tidyverse.org/functions.html#return

PR #3354, opened 2019-06-05
Cleanup Travis config

jeroen:
Using declarative installation of `apt` packages will result in better caching. Also `sudo` is deprecated and `dist: trusty` is EOL soon, it's better to use the default linux distribution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: R
 cache: packages
-sudo: true
-dist: trusty
 
 # build matrix; turn on vdiffr only on r release
 matrix:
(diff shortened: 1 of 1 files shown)


yutannihilation:
It seems Travis still uses Trusty here. So, we might need some additional tweaks when actually migrate to Xenial.

https://travis-ci.org/tidyverse/ggplot2/jobs/541740990#L9

Anyway, I think we should merge this as Trusty is already EOL. Unless there's some concern, I'll merge this.

yutannihilation:
Confirmed tests run fine on Xenial with my forked repo: https://travis-ci.org/yutannihilation/ggplot2/builds/563266792

The main repo stays on Trusty even after I cleared the build cache, but I think we can just wait for their migration is done: https://travis-ci.org/yutannihilation/ggplot2/builds/563266792

PR #3356, opened 2019-06-06
WIP: scale-like objects in panel_params that can be used to train guides

paleolimbot:
In order for position guides to act like other guides (#3322), there needs to be a scale-like object for the position guides to be trained on. This PR is one experiment on how to do this.

Currently, the `panel_params` are the only valid representation of the position scales that is available at draw time, as scales are not expanded in place, but temporarily expanded by the Coord as part of the `Coord$setup_panel_params()`. The `panel_params` for most coordinate systems is based on `Scale$break_info()`, which is only ever used in coordinate systems. The `panel_params` serve as the source of information for the `Coord` transformation functions, range functions, axis guides, and panel guides (the panel grid/grill). Some layers (like `geom_dotplot()`, `annotation_logticks()`, and `ggspatial::annotation_scale()`) also use the `x.range` and `y.range` elements of the `panel_params`.

Expanded position scales used by coordinate systems serve a different purpose than scales used for training, and at first I tried [coercing the existing scales to serve as both](https://github.com/tidyverse/ggplot2/compare/master...paleolimbot:scales-refactoring) but even though it sort of works, it's messy and causes at least two tests to fail.

This PR introduces an internal class, `ViewScale`, that serves as the definitive source of truth for coordinate systems (and potentially layers) regarding the expanded position scale. To demonstrate how it would be used, I refactored `CoordCartesian` to use view scales instead of a composite of `Scale$break_info()` `list()`s. Ideally, the `break_info()` function should be moved to the `ViewScale()` constructor to separate the concerns of `Scale` objects (transforming, censoring, training, and calculating limits/breaks/labels) from the concerns of the `Coord`/`Geom` (transforming position aesthetics from transformed data space to coordinate space) from the concerns of the guide (what are the break values, break labels, and break positions in coordinate space). Only `CoordCartesian` has a 1-to-1 mapping between a `ViewScale` and a future `guide_axis()`, but all the `Coord`s need information about the expanded `x` and `y` scales and could potentially be refactored to use them.

The main questions I have are:

- Is this a reasonable way to write a scale-like object with which to train a future `guide_axis()`
- Should `CoordCartesian` use `ViewScales` for anything except guides?
- If so, should other coordinate systems use `ViewScales`?


clauswilke:
I think this is going in the right direction. It's often bugged me that there is no authoritative source of information of the exact axis range in a plot, and it seems this PR would fix that. I find the function names a bit confusing, though. E.g., what does `view_scales_from_scale()` do? It might help if you could add some comments that explain what the various pieces do. That may also provide better guidance about whether the names are appropriately chosen or not.

Would this allow different panels to have different scales? As an example, we don't currently (I believe) have a faceting system that can do something like the plot below, but after attending the grammar of graphics session at SDSS 2019 I believe this is a major limitation of ggplot2 and should be supported in the future. It would require that we can somehow access and style the scales for each panel.

``` r
library(ggplot2)
library(patchwork)

p1 <- ggplot(mtcars, aes(disp, mpg)) + geom_point()
p2 <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
p1 + p2
```



<sup>Created on 2019-06-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/coord-cartesian-.r
```diff
@@ -79,50 +79,124 @@ CoordCartesian <- ggproto("CoordCartesian", Coord,
...
-    data <- transform_position(data, rescale_x, rescale_y)
+    data <- transform_position(data, panel_params$x.scale$map, panel_params$y.scale$map)
     transform_position(data, squish_infinite, squish_infinite)
   },
 
   setup_panel_params = function(self, scale_x, scale_y, params = list()) {
-    train_cartesian <- function(scale, limits, name) {
-      range <- scale_range(scale, limits, self$expand)
+    c(
+      view_scales_from_scale(scale_x, self$limits$x, self$expand),
```
hadley: I think it would be cleaner to have name this list, and not embedded the aesthetic name in the list returned by `view_scales_from_scale()`, so the code below could be `panel_params$x$scale$break_positions_minor()`

R/coord-cartesian-.r
```diff
@@ -79,50 +79,124 @@ CoordCartesian <- ggproto("CoordCartesian", Coord,
...
+      view_scales_from_scale(scale_y, self$limits$y, self$expand)
+    )
+  },
 
-      out <- scale$break_info(range)
-      out$arrange <- scale$axis_order()
-      names(out) <- paste(name, names(out), sep = ".")
-      out
-    }
+  render_bg = function(panel_params, theme) {
```
hadley: Maybe also make `draw_view_grid()`?

smouksassi:
while with tidyr::gather the facet can be done quite easily, we cannot style (e.g. specifying limits or specific ticks for each facet). Lattice support a list of  limits
``` r

library(tidyr)
library(ggplot2)
#> Registered S3 methods overwritten by 'ggplot2':
#>   method         from 
#>   [.quosures     rlang
#>   c.quosures     rlang
#>   print.quosures rlang
library(patchwork)
library(lattice)

p1 <- ggplot(mtcars, aes(disp, mpg)) + geom_point()
p2 <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
mtcarsgathered<- gather(mtcars,key,value,disp,wt)

p3<- ggplot(mtcarsgathered, aes(value, mpg)) + geom_point()+
  facet_grid(~key,scale="free_x", switch= "x")+
  theme(strip.placement = "outside",strip.background = element_rect(fill="white"),axis.title.x = element_blank())

(p1 + p2) / p3
```



``` r


p4<- xyplot(mpg  ~ value | key, data= mtcarsgathered,
       scales=list(x=list(relation="free",
                           limits=list(c(50,500), c(1,10)) 
                          )),
       )

gridExtra::grid.arrange( p3  , p4)
```



<sup>Created on 2019-06-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/coord-cartesian-.r
```diff
@@ -79,50 +79,91 @@ CoordCartesian <- ggproto("CoordCartesian", Coord,
...
-    range <- range(scale$transform(limits))
-    expand_range(range, expansion[1], expansion[2])
+    continuous_range <- range(scale$transform(coord_limits))
+    continuous_range <- expand_range4(continuous_range, expansion)
+  }
+
+  aesthetic <- scale$aesthetics[1]
+
+  view_scales <- list(
+    view_scale(scale, limits, continuous_range),
```
hadley: Maybe call `primary` and `secondary`?

R/scale-.r
```diff
@@ -1,11 +1,275 @@
+
+#' Continuous scale constructor
```
hadley: What are all the changes here?

R/scale-.r
```diff
@@ -490,184 +792,115 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
-
-  trans <- as.trans(trans)
-  if (!is.null(limits) && !is.function(limits)) {
-    limits <- trans$transform(limits)
+#' @param scale The scale from which to construct a view scale.
+#' @param limits The final scale limits
+#' @param continuous_range The final dimensions of the scale
+#'
+#' @noRd
+view_scale <- function(scale, limits = scale$get_limits(),
```
hadley: It might be easier to put this in a separate file?

R/scale-.r
```diff
@@ -490,184 +792,115 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+    rescale = function(x) stop("Not implemented", call. = FALSE),
+    map = function(x) stop("Not implemented", call. = FALSE),
+    make_title = function(title) title,
+    break_positions = function() NULL,
+    break_positions_minor = function() NULL
+  )
+}
 
-  aesthetics <- standardise_aes_names(aesthetics)
+ViewScale <- ggproto("ViewScale", NULL,
```
hadley: I think the code is a bit easier to follow if the parent class is defined first.

R/scale-.r
```diff
@@ -490,184 +792,115 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
-
-  # If the scale is non-positional, break = NULL means removing the guide
-  if (is.null(breaks) && all(!is_position_aes(aesthetics))) {
-    guide <- "none"
-  }
-
-  trans <- as.trans(trans)
-  if (!is.null(limits) && !is.function(limits)) {
-    limits <- trans$transform(limits)
+#' @param scale The scale from which to construct a view scale.
```
hadley: Can you briefly describe in the docs why they can't be computed from the scale? (Presumably because the coord has extra info)

R/scale-.r
```diff
@@ -490,184 +792,115 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
+  # to the original scale
+  scale = ggproto(NULL, Scale),
 
-  position <- match.arg(position, c("left", "right", "top", "bottom"))
+  is_empty = function(self) {
+    is.null(self$get_breaks()) && is.null(self$get_breaks_minor())
+  },
+  is_discrete = function(self) self$scale_is_discrete,
+  dimension = function(self) self$continuous_range,
+  get_limits = function(self) self$limits,
```
hadley: Is it worth having these wrapper methods?

R/scale-continuous.r
```diff
@@ -147,6 +147,35 @@ ScaleContinuousPosition <- ggproto("ScaleContinuousPosition", ScaleContinuous,
...
+second_view_scale <- function(scale, limits = scale$get_limits(), range = scale$dimension(limits = limits)) {
+  if (is.null(scale$secondary.axis) || is.waive(scale$secondary.axis) || scale$secondary.axis$empty()) {
+    view_scale_empty()
+  } else {
+    scale$secondary.axis$init(scale)
+    break_info <- scale$secondary.axis$break_info(range, scale)
+    names(break_info) <- gsub("sec\\.", "", names(break_info))
+
+    ggproto(
+      "ViewScaleSecondary", NULL,
```
hadley: Can you explain briefly why this needs to be a separate class?

paleolimbot: The short answer is that I needed a way to keep all the tests passing while forcing CoordCartesian to use scale-like objects rather than break_info()-generated objects. Because second axes aren't scales they can't be constructed in the same way, but if this PR turns into something that is worth merging, I will need to find a better way to do this.

R/scale-.r
paleolimbot: I think so. I was trying to keep `ViewScale`s interface-compatible with `Scale`s, so that the eventual guide training code is portable regardless of the type of guide. Also, it means that there isn't a new interface for anybody to learn, just a simpler implementation for guides that don't have to calculate anything.

R/scale-.r
paleolimbot: As in, `view_scale()` should go after `ViewScale <- ggproto(...)` ?

R/scale-.r
paleolimbot: These are changes that were required to keep the view scale constructor short. There are a few that were required to get previous version of the code to work that I can either take out or put into a separate pull request. Finally, there were a lot of comments and some documentation that were straight up lying, which I redocumented since I had to figure out what all these methods did to write everything else. I think this should probably be a separate pull request.

R/scale-continuous.r
paleolimbot: I see what you mean - it should be an instance of `ViewScale` not its own class declaration.

PR #3361, opened 2019-06-12
Fix coord_sf() regression introduced in draw_axis rewrite

paleolimbot:
In the `draw_axis()` rewrite (#3349), I neglected a piece of code that allowed `list()`s of `language` objects to be passed as labels. This is only used in `coord_sf()` to draw degree signs next to labels, and because the `coord_sf()` tests are all skipped, I didn't notice that I had missed this piece of code. This PR fixes the regression so that `coord_sf()` plots have degree signs in axis labels once more.


clauswilke:
I've been thinking that the decision to skip the sf tests may not be a good one. I realize the problems created by tests whose outcome depends on the version of other packages, but it may be better to update the reference images to those tests from time to time than to risk introducing major regressions.

clauswilke:
It seems like 2 of the 3 items I listed in that issue have been resolved already. Should we turn tests for sf back on and see what happens? @hadley?

hadley:
Maybe we should just skip them on CRAN? That way we'll still see on travis and locally, but changes to sf or other stream packages won't force a ggplot2 release.

clauswilke:
All visual tests are already skipped on CRAN. That's what I meant when I said 2 of the 3 items I listed in the issue have been resolved: We skip visual tests on CRAN, and on Travis we only run visual tests for the currently released R version.

hadley:
Ok got it, then it's definitely worth re-enabling.

PR #3374, opened 2019-06-19
Kill site search

batpigandme:
Per @hadley 1-on-1…

* Wasn't actually searching the right site.
* Was harshing our vibe in the navbar.


hadley:
And everyone uses google anyway. In general, I don't think we should both with site search for tidyverse sites.

Issue #3281, opened 2019-04-30
Feature Request: Axis Text Warning or Automatic Rotation when Labels Overlap


billdenney:
I sometimes work with data where the axis text may overlap, and it would be helpful if ggplot either warned when the axis text overlapped or even better could automatically adjust rotation and justification when overlap occurs.

``` r
library(ggplot2)

data_with_long_names <-
  data.frame(
    A=c(paste0(rep(LETTERS, 3), collapse=""), paste0(rep(letters, 3), collapse="")),
    B=1
  )

# Feature request: This gives a warning
ggplot(data_with_long_names, aes(x=A, y=B)) +
  geom_point()
```

![](https://i.imgur.com/V6ipZuO.png)

``` r

# Option for consideration as an interface to automatically fix the issue
#ggplot(data_with_long_names, aes(x=A, y=B)) +
#  geom_point() +
#  theme(axis.text.x.overlap=element_text(angle=45, hjust=1))
```

Created on 2019-04-30 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

thomasp85:
This is already being considered and will hopefully appear in the future

paleolimbot:
Referencing #3322 as I'm working on this today as part of that project. The current strategies I think that I can build in to the current axis function relatively easily are (1) smart ordering of the labels and using `check.overlap = TRUE` to hide "lower priority" labels, (2) stacking alternate labels into multiple rows/columns, and (3) rotating labels (you can already rotate the labels using the theme, but the details are hard to get right). Depending on how well this works, we'll see if one of these methods can't be applied by default.

clauswilke:
@paleolimbot Have you thought about how you want the user to turn on or off/customize this? I think ideally it would be encapsulated into a new element type, so that it can be all customized by setting the appropriate element type and arguments in the theme.

The reason why rotated labels are difficult is because we don't have enough parameters to specify how rotation should be performed. Just `angle`, `vjust`, and `hjust` are not enough, because in the general case you also need to specify the reference point around which the label is rotated. This doesn't fit into the current `element_text()` setup, and a general fix would again require a different element type. So it's best if we can rewrite this section of axis drawing code with that in mind.

For an example of the need for more parameters, see the readme here: 
https://github.com/clauswilke/gridtext

paleolimbot:
[...]

PR #3375, opened 2019-06-19
Add strategies to deal with overlapping text in draw_axis()

paleolimbot:
This is a PR as part of #3322 to build in some basic strategies to the default axis renderer (`draw_axis()`). The inspiration for this PR is #3281, although the issue can't be closed until there is actually a way to change the arguments to `draw_axis()`, which is the almost next step for #3322.

The three strategies are (1) recursively prioritize the "middle" label, and pass `check.overlap = TRUE` to `textGrob()`, (2) rotate the axis text, and (3) dodge the text into multiple rows/columns. These are the strategies I chose because they can operate independent of each other, don't require overhauling the existing axis code, and don't cause any visual changes with the default parameters. There are probably better strategies that can be implemented when axis guides are extensible (I imagine that @clauswilke's gridtext would introduce a number of alternatives).

A reprex so that any/all can play around with the new options:

``` r
library(ggplot2)
library(grid)
library(gtable)

# theme with which axes should be tested
theme_test_axis <- theme_test() + theme(axis.line = element_line(size = 0.5))

# function to test several axes at once
test_draw_axis <- function(n_breaks = 10,
                           # make long labels
                           labels = function(breaks) scales::comma(breaks * 1e7),
                           positions = c("top", "bottom"),
                           theme = theme_test_axis,
                           ...) {
  
  break_positions <- seq_len(n_breaks) / (n_breaks + 1)
  break_labels <- labels(seq_len(n_breaks))
  
  # create the axes
  axes <- lapply(positions, function(position) {
    ggplot2:::draw_axis(break_positions, break_labels, axis_position = position, theme = theme, ...)
  })
  axes_grob <- gTree(children = do.call(gList, axes))
  
  # arrange them so there's some padding on each side
  gt <- gtable(
    widths = unit(c(0.05, 0.9, 0.05), "npc"),
    heights = unit(c(0.05, 0.9, 0.05), "npc")
  )
  gt <- gtable_add_grob(gt, list(axes_grob), 2, 2, clip = "off")
  plot(gt)
}

test_draw_axis(check.overlap = TRUE)
```

``` r
test_draw_axis(angle = 45)
```

``` r
test_draw_axis(n_dodge = 2)
```

<sup>Created on 2019-06-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


R/guides-axis.r
```diff
@@ -124,3 +145,109 @@ draw_axis <- function(break_positions, break_labels, axis_position, theme) {
...
+  labels_grob <- exec(
+    element_grob, label_element,
+    !!position_dim := unit(break_positions, "native"),
+    !!label_margin_name := TRUE,
+    label = break_labels,
+    check.overlap = check.overlap
+  )
+}
+
+#' Deterine the label priority for a given number of labels
```
clauswilke: "Determine"?

paleolimbot: Good catch!

clauswilke:
@paleolimbot I'm trying to figure out how exactly this code works, and I can't map it to my mental model of how I think this should be done. Could you help me figure out whether I'm thinking about this wrong?

Normally, in grid, if we want to write a grob that can make drawing decisions based on the dimensions of text, we need to do this inside a `makeContent()` function for an appropriate grob class, so that the drawing decisions adjust appropriately in interactive use. For an example, see my implementation of `isolines_grob()`: https://github.com/wilkelab/isoband/blob/master/R/isolines-grob.R 

So for an axis, I would expect that there is a grob class that draws all axis labels, and inside the `makeContent()` function for that class the code would calculate the dimensions of the labels, drop labels if needed, rotate them, etc. I can't find this structure in your current pull request, but maybe I'm overlooking something.

clauswilke:
Yeah, I don't know. Until this morning I didn't even know this was possible, though now that I think about it it can see how it can work.

I'm a bit worried that the text rendering code in ggplot2 is not well encapsulated and makes inconsistent assumptions in different places (e.g., the whole `titleGrob()` code assumes that we're drawing only a single text label [as otherwise adding descent data to grob width/height doesn't make much sense], but then it's sometimes used to draw multiple labels at once). As we rewrite drawing code, we should make sure we're moving forward and are cleaning things up, not making code even more complex and confusing.

If I understand correctly, your code draws separate text labels with separate grobs for each axis label, so that's good. It means we should be able to use a different text drawing function in the future without too much difficulty.

I'm a bit behind on where I wanted to be with gridtext, but I think I'll quickly (over the next week or so) hack together a simple replacement text grob that we can use for testing. If it can be used instead of the current `titleGrob()` for your axis drawing code, then we're on the right track. If not, then the failure mode will likely tell us where we're going wrong.

hadley:
If the end unit alignment is the only thing that needs lazy units, I think we should just remove it.

clauswilke:
Aren't lazy units required for all three features, since overlap can also change dynamically as the plot window is resized?

Dewey, what I was trying to get at is the following: If the code is modular, then it should be possible to provide a drop-in replacement for `element_text()` that provides more capabilities without requiring any changes in the underlying ggplot2 code base. This currently is possible for things like plot title, subtitle, axis title, etc., but it is not possible for axis tick labels. If there's a chance that after your rewrite, it will also be possible for axis tick labels, then we should make this happen.

I realize this will make more sense with an example, and I'll put one together over the weekend.

clauswilke:
> Axis labels are just (vectorized) rendered theme elements, so anything you do for those will work for the code I've written.

The vectorization has been the problem, actually. But never mind, your refactoring will make it possible to plop in different axis drawing code in the future, so some of the limitations I see could be addressed this way then.

clauswilke:
(Basically, it is not clearly defined whether `element_text()` creates one or more individual labels or whether it creates one container with one or more labels inside. This distinction causes inconsistencies and weird effects when we try to replace `element_text()` with something else.)

PR #3376, opened 2019-06-20
fix groupwise proportion in stat_count()

### LINKED ISSUE #3369: computing column proportions in stat_count()
When using `geom_bar()` with `position_fill()` it would be great to be able to add percentage labels, corresponding to the y axis (column percentage), i.e. to be able to do something like:

```r
ggplot(as.data.frame(Titanic)) +
   aes(x = as.integer(Class), fill = Survived, weight = Freq) +
   geom_bar(position = "fill") + 
  geom_text(aes(label = scales::percent(..cprop..), stat = "count", position = position_fill(.5)) +
  facet_wrap(~ Sex)
```

For now, it cannot be done using the `..prop..` as it is computed per group and not within panel.

larmarange:
fix #3369 

Will allow to easily display percentages as labels when using `position_fill()`.

An example:

```r
ggplot(mpg, aes(x = class, fill = drv)) + 
  geom_bar(position = "fill") + 
  geom_text(aes(label = scales::percent(..cprop..)), stat = "count", position = position_fill(.5))
```

image

### DIFF
--- a/R/geom-bar.r
+++ b/R/geom-bar.r
@@ -71,6 +71,39 @@
 #' ggplot(df, aes(x)) + geom_bar()
 #' # cf. a histogram of the same data
 #' ggplot(df, aes(x)) + geom_histogram(binwidth = 0.5)
+#'
+#' # You can use position_fill() for comparing proportions
+#' ggplot(mpg, aes(x = class, fill = drv)) +
+#'   geom_bar(position = "fill")
+#' # If you want to display proportions as labels,
+#' # you can use groupwise proportions
+#' # (the 'group' aesthetic needs to be defined)
+#' ggplot(mpg, aes(x = class, fill = drv, group = class)) +
+#'   geom_bar(position = "fill") +
+#'   geom_text(
+#'   aes(label = scales::percent(..prop..)),
+#'   stat = "count",
+#'   position = position_fill(.5)
+#' )
+#'
+#' # You can use position_dodge() to compare two distributions
+#' titanic <- as.data.frame(Titanic)
+#' ggplot(titanic, aes(x = Class, fill = Age, weight = Freq)) +
+#'   geom_bar(position = "dodge")
+#' # If you want to adjust of the size of each age group,
+#' # you can use groupwise proportions
+#' # (the 'group' aesthetic needs to be defined)
+#' ggplot(titanic, aes(x = Class, fill = Age, weight = Freq, group = Age, y = ..prop..)) +
+#'   geom_bar(position = "dodge")
+#' # If you want to display the proportions as labels
+#' ggplot(titanic, aes(x = Class, fill = Age, weight = Freq, group = Age, y = ..prop..)) +
+#'   geom_bar(position = "dodge") +
+#'   geom_text(
+#'     aes(label = scales::percent(..prop..)),
+#'     stat = "count",
+#'     position = position_dodge(width = .9),
+#'     vjust = "top"
+#'   )
 geom_bar <- function(mapping = NULL, data = NULL,
                      stat = "count", position = "stack",
                      ...,
--- a/R/stat-count.r
+++ b/R/stat-count.r
@@ -8,6 +8,9 @@
 #'   number of cases at each `x` position (without binning into ranges).
 #'   [stat_bin()] requires continuous `x` data, whereas
 #'   `stat_count` can be used for both discrete and continuous `x` data.
+#'   `..prop..` corresponds to groupwise proportions, i.e. all proportions
+#'   for a specific group sum to 1. The group could be define using the
+#'   `group` aesthetic (see examples).
 #'
 #' @export
 #' @rdname geom_bar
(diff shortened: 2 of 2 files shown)


clauswilke:
This may be a useful addition, but the code definitely needs better documentation. For example, please write a paragraph in the function documentation explaining the `prop` and `cprop` stats. I'm also not entirely sold on the name `cprop`, it's not immediately obvious what it could mean. Finally, please add comments throughout your code explaining why you're doing what you're doing. E.g., what does the `aggregate()` line do?

ptoche:
something like ``prop.col`` would be more explicit and arguably more memorable.

larmarange:
I will update the documentation and provide additional explanations. To better understand the difference, I will also provide examples for the current `..prop..`

Regarding the name of `..cprop..`, another option could be `..xprop..` as this is the proportion on `x`, i.e. all proportions for a same value on `x` sum to 1. Or `..col.prop...` for column proportion.

Do you have any preference?

larmarange:
- `..cprop..` has been renamed `..xprop..` as it corresponds to relative proportions at each `x`
- in documentation, a paragraph now explains what are `..prop..` and `..xprop..`
- new examples have been added presenting some use cases of `..prop..` and `..xprop..`

larmarange:
The main change in this pull request is that count computation is now performed through `compute_panel` instead of `compute_group`.

The starting point was that it was not possible so far to use groupwise proportions to compute proportions on x. See the following example done with the current version of ggplot2.

```r
ggplot(as.data.frame(Titanic)) + 
  aes(x = Class, fill = Survived, weight = Freq, group = Class) + 
  geom_bar(position="fill") + 
  geom_text(aes(label = ..prop..), stat = "count", position = position_fill(.5))
```

produce

image

However, with the changes proposed in this pull request, the same code produce:

image

Therefore, `..xprop..`  is not needed anymore, as it could be achieved through groupwise proportions (controlled by the `group` aesthetic), as long as count computation remains in `compute_panel`.

larmarange:
Hi, is there a possibility for that proposition to be integrated within `ggplot2`? Are there any discussion points before to consider it?

larmarange:
Any feedback about that proposition? Displaying group proportion is a very common need.

Regards

clauswilke:
I've looked over your PR. The problem I see is that there are many places where the currently proposed code doesn't match ggplot2 conventions or has other small issues. I'll provide a list (not necessarily exhaustive) below. It might be best if you tried yourself on a smaller issue (maybe fixing just documentation) first, to get a better sense of how we do things.

Issues I notice:
- You added a lot of new examples. Examples are costly when running CRAN check, and therefore we should be judicious in which ones we add.
- You use `..aesthetic..`. This style is outdated. We now use `after_stat(aesthetic)`.
- Use of `aggregate()` with a formula is not ggplot2 style, and it makes the reviewers uncomfortable because we can't quickly see that the code would do the right thing.
- Similarly, since you deleted the explicit creation of the output data frame we now can't easily see what the function returns. This may be appropriate, but should be explained carefully in code comments. What is it that `aggregate()` returns exactly?
- Would this `aggregate()` code work with categorical aesthetics, or would it throw an error?
- In general, I would recommend to make the minimal change that achieves your goal. I *think* you changed the overall behavior of the code, where previously the stat was working on only a single column in the data frame but now `aggregate()` works on all columns. If you want to introduce this change it should be in a separate PR with a separate discussion about the pros and cons.
- We need unit tests for the new functionality.

larmarange:
Dear @clauswilke ,

first of all, thank you for your feedback. 

I completely understand that the discussion of this PR is becoming confusing and that the proposed PR is out of date regarding last evolutions of `ggplot2`. I will, therefore, develop a new PR. However, before doing so, I would like some advice.

First, I will try to summarize the issue.  It comes from the way groupwise proportions are computing by `stat_count()`. Groupwise proportions should be understood as the fact that all proportions within a specific group should sum to 1.

Just to illustrate, if you no `group` aesthetic is indicated with a plot, then the computed proportions are always equal to 1.

```r
ggplot(as.data.frame(Titanic)) + 
 aes(x = Class, fill = Survived, weight = Freq) + 
 geom_bar() + 
 geom_text(aes(label = after_stat(prop)), stat = "count", position = position_stack(.5))
```

produces

image

If we precise in `group` aesthetic the same variable as for the `fill` aesthetic, then groupwise proportions are correctly computed. Therefore, it's possible to produce such graph:

```r
ggplot(as.data.frame(Titanic)) + 
 aes(x = Class, fill = Survived, weight = Freq, group = Survived, y = after_stat(prop)) + 
 geom_bar(position = "dodge") + 
 geom_text(aes(label = scales::percent(after_stat(prop))), stat = "count", position = position_dodge(.9))
```
image

However, if we try to group by the variable used for `x` aesthetic, we have an error.

```r
ggplot(as.data.frame(Titanic)) + 
  aes(x = Class, fill = Survived, weight = Freq, group = Class) + 
  geom_bar(position="fill") + 
  geom_text(aes(label = ..prop..), stat = "count", position = position_fill(.5))
```

produces

image

In fact, the expected behaviour would be to get 

image


When using `geom_bar()` with `position = "fill"`, it's very common to display proportions on `x` which are the proportions used on the `y` axis.

The problem comes to the fact that the proportions in `stat_count()` are so far computed through `compute_group`  which doesn't allow to group on `x` and could be solved by moving this computation at a higher level in `compute_panel` and this was the purpose for that PR.

**Alternatives to `aggregate()`**

An alternative to `stats::aggregate()` would be to use `dplyr` with `group_by()` and `summarise()`.

However, so far, `dplyr()` is not imported by `ggplot2()`. This is why I through it could be better to use base R `aggregate()` to avoid adding dependencies to `ggplot2`. However, it is possible to use a `dplyr` syntax instead but it will require imports from `dplyr`.

`cp <- aggregate(weight ~ ., data = data, sum, na.rm = TRUE)`  is equivalent to `cp <- data %>% group_by_at(vars(-weight)) %>% summarise(count = sum(weight, na.rm = TRUE))`

Similarly `stats:ave()` could be replaced by a dplyr syntax using `group_by()`

**Regarding documentation**

I understand the need to keep examples computation time low for CRAN check. However, the best way to explain what a function does is to provide examples. Therefore, which alternative do you think would be better?

- examples with `\dontrun{}`
- example in the `\details{}` section
- a specific vignette?


Best regards

Note: `aggregate()` works perfectly with categorical aesthetics.

clauswilke:
I've pondered this some more, and I think at this time this feature would best live in an extension package. That's exactly what the extension system was designed for. Just write your own stat, put it into a package, and submit to CRAN.

PR #3056, opened 2019-01-04
Deprecate `limx` and `limy` in `coord_trans()`

billdenney:
The `limx` and `limy` arguments to `coord_trans()` don't align with most other commands.  I've suggested deprecating them and using the more standard `xlim` and `ylim`.

Issue #3371, opened 2019-06-18
Single strategy for scale expansion/limiting that works with transformations


paleolimbot:
There are a number of issues that surround expansion and limiting, which is treated differently in various coordinate systems. It would be nice if specifying xlim and ylim were like specifying `scale_(x|y)_*(limits = ...)`, and if scale expansion were identical for `scale_(x|y)_continuous(trans = ...)` and `coord_trans((x|y) = ...)`.

Related issues:

- #3270: interaction between xlim/ylim and expand
- #2907: Allowing `NA` in xlim/ylim as one can do in the scale `limits`
- #980: breaks for expanded scales with transformations
- #2990, #3338: expansion behaviour in `coord_trans()`

I think these could almost all be handled in the relatively new `ViewScale` class, whose objects are currently being created here (this will probably change):

https://github.com/tidyverse/ggplot2/blob/6c2fe76e77a50aae5935935a8d17e404a8100871/R/coord-cartesian-.r#L139-L161

paleolimbot:
In a slightly similar way, coordinate systems should probably be able to replace scale defaults that aren't explicitly set (should they?). Off the top of my head, `coord_trans()` should probably be able to coerce the (unset) breaks and labels on the transformed axes to `trans$breaks` or `trans$labels`, and there is one open issue on something similar with `coord_fixed()` (#2888)

Issue #3338, opened 2019-05-22
Remove extra space created by `coord_trans`


itsvenu:
I'm trying to produce a two panel figure for my data. At one point I had to use `coord_trans` in my code which created an extra space between y-axis and first beeswarm dots.

Here is the figure

![fig](https://i.imgur.com/wPdZA1C.png)

I would like to remove the space between y-axis and first set of dots so that gray-white shading align in 2 panels. If I don't use `coord_trans`, shadings are aligned perfectly. Although, `coord_trans` documentation says that it will change visual appearance of geoms, I'm hoping there might be a solution to produce a figure I'm thinking of.

Following is the sample data and to reproduce the figure

data: [Link to data](https://gist.github.com/itsvenu/97605e07b626d04f290fb39af7ccbc3a)

```r
all_sb <- unique(mp_mb_sampledata$SB) 

cl_min <- seq(from = 0.5, to = max(as.numeric(as.factor(all_sb))), by = 1)[1:length(all_sb)]
cl_max <- seq(from = 1.5, to = max(as.numeric(as.factor(all_sb))) + 0.5, by = 1)[1:length(all_sb)]

shading_cols <- data.frame(min = cl_min,
                           max = cl_max,
                           col = rep(c(0, 1), length(cl_min)))


## top panel
p1 <- ggplot()+
  ggbeeswarm::geom_quasirandom(data = mp_mb_sampledata, aes(x = SB, y = mut_per_mb))+
  geom_rect(data = shading_cols,
            aes(xmin = min, xmax = max, ymin = -Inf, ymax = Inf,
                fill = factor(col), alpha = 0.01))+
  scale_fill_manual(values = c("white", "gray"), guide = FALSE)+
  ggbeeswarm::geom_quasirandom(data = mp_mb_sampledata, aes(x = SB, y = mut_per_mb))+
  coord_trans(y="log2")+
  theme_classic(base_size = 18)+
  xlab("")+ylab("")+
  theme(legend.position = "none",
        axis.text.x = element_blank(),
        axis.ticks.x = element_blank(),
        axis.line.x = element_blank(),
        axis.text = element_text(color = "black"))

## bottom panel
p2 <- ggplot()+
  geom_bar(data = req_bc, aes(x = SB, y = Freq, fill = MutationType), stat = "identity")+
  geom_rect(data = shading_cols,
            aes(xmin = min, xmax = max, ymin = -Inf, ymax = Inf,
[...]

Issue #3270, opened 2019-04-24
Scale expand value from scales ignored when coord_cartesian ylim is used


brodieG:
It would be desirable to be able to specify 'expand' values in conjunction with `coord_cartesian(ylim=...)`.  That does not appear to work, as the expand values are reverted back to defaults.  If you agree the current behavior is unexpected I can look into a PR.

``` r
library(ggplot2)
p <- ggplot(data.frame(x='a', y=100)) +
  geom_col(aes(x, y)) +
  scale_y_continuous(expand=expand_scale(mult=c(0.05, 0.50)))
p + ggtitle('Simple')
```

![](https://i.imgur.com/fq5P4GR.png)

``` r
p + ggtitle('w/ coord_cart lim') +
  coord_cartesian(ylim=c(0, 100))
```

![](https://i.imgur.com/0NStTzU.png)

``` r
p + ggtitle('w/ coord_cart lim expand false') +
  coord_cartesian(ylim=c(0, 100), expand=FALSE)
```

![](https://i.imgur.com/lvRRqbM.png)

<sup>Created on 2019-04-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> Session info -------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.5.3 (2019-03-11)
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  tz       America/New_York            
#>  date     2019-04-24
#> Packages -----------------------------------------------------------------
#>  package    * version date       source                         
#>  assertthat   0.2.1   2019-03-21 CRAN (R 3.5.2)                 
#>  backports    1.1.2   2017-12-13 CRAN (R 3.5.0)                 
#>  base       * 3.5.3   2019-03-13 local                          
#>  colorspace   1.3-2   2016-12-14 cran (@1.3-2)                  
#>  compiler     3.5.3   2019-03-13 local                          
#>  crayon       1.3.4   2017-09-16 CRAN (R 3.5.0)                 
#>  curl         3.2     2018-03-28 CRAN (R 3.5.0)                 
#>  datasets   * 3.5.3   2019-03-13 local                          
#>  devtools     1.13.6  2018-06-27 CRAN (R 3.5.0)                 
[...]

PR #3380, opened 2019-06-23
Consolidate coordinate scale expansion and limiting code

### LINKED ISSUE #2907: `NA` limits in coords
Setting `NA` in limits of `coords` in a similar fashion to `scales` ends up in an error. 

``` r
library(ggplot2)

ggplot(cars, aes(speed, dist)) +
   geom_point() +
   coord_cartesian(ylim = c(25, NA))  # emulating scale_y_continuous(limits = c(25, NA))
#> Error in if (zero_range(range)) zero_width else diff(range): missing value where TRUE/FALSE needed
```

Created on 2018-09-21 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

I've tinker a bit and found that the error is in this line:
 https://github.com/tidyverse/ggplot2/blob/01155ba2e9394e17066668eace50c13354e03b3a/R/coord-cartesian-.r#L121

Adding this:
`limits <- ifelse(is.na(limits), scale$get_limits(), limits)`
solves the issue in this case. 

``` r
library(ggplot2)

scale_range <- function(scale, limits = NULL, expand = TRUE) {
   expansion <- if (expand) ggplot2:::expand_default(scale) else c(0, 0)
   
   if (is.null(limits)) {
      scale$dimension(expansion)
   } else {
      limits <- ifelse(is.na(limits), scale$get_limits(), limits)
      range <- range(scale$transform(limits))
      scales::expand_range(range, expansion[1], expansion[2])
   } 
}

assignInNamespace("scale_range", scale_range, ns = "ggplot2")

ggplot(cars, aes(speed, dist)) +
   geom_point() +
   coord_cartesian(ylim = c(25, NA))  
```

Created on 2018-09-21 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

I'm not versed enough on how `ggplot2` handles coords to know if it's a robust fix across different coords, though.

### LINKED ISSUE #2990: Secondary axis doesn't show axis ticks or labels with coord_trans
Exactly the same issue as #2072, which I take it was fixed since it is closed.

The secondary axis ticks and labels do not show up when coord_trans is present, even if using the identity transformation. The axis label ("name" parameter) does still appear.

For example, roughly taking from the examples in sec_axis:

```r
ggplot(mtcars, aes(cyl, mpg)) + geom_point()  +
    scale_y_continuous(sec.axis = sec_axis(~.+10, name='MPG + 10'))
```

works as expected, but

```r
ggplot(mtcars, aes(cyl, mpg)) + geom_point()  +
    scale_y_continuous(sec.axis = sec_axis(~.+10, name='MPG + 10')) +
    coord_trans()
```

loses the axis ticks and labels (but keeps the title "MPG + 10").

### LINKED ISSUE #3336: Whiskers and outliers for geom_boxplot with transformed scale
Boxplot whiskers and outliers seem to be incorrect when using geom_boxplot with a transformed scale. It probably affects notches too, but I haven't checked.

The hinges are fine, which makes sense—monotonic transformations preserve quantiles.

``` r
library(ggplot2)
df <- data.frame(x = 1, y = c(1, 4, 5, 6, 9))
print(ggplot(df, aes(x, y)) + geom_boxplot())
```

``` r
print(ggplot(df, aes(x, y)) + geom_boxplot() + scale_y_sqrt())
```

<sup>Created on 2019-05-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I'm using ggplot2 3.1.1.

paleolimbot:
This is a PR to address coordinate scale expansion and limiting using a single approach, implementing several feature requests and fixing several bugs in the process (closes #2990, closes #2907, closes #3056, closes #3336). The inspiration for this PR was a collection of issues that @hadley collected when issue triaging that were related to the interface between scales and coordinate systems (#3371). I'd be happy to split this up into a few different PRs if it is easier to review (one for expansion, one for coord limiting, one for sprucing up coord_trans).

- All scale expansion code now lives in `scale-expansion.r`, and is tested independently of the scales/coords. All `Coord` objects now use these expansion functions. Previously, some form of scale expansion code lived in `utilities.r`, `scale-.r`, `coord-sf.r`, `coord-cartesian.r`, `scale-discrete.r`, and `coord-transform.r`. This led to at least one issue of inconsistent expansion between coords (#3338), and was thought to cause several other issues, although it turns out that #2990 was unrelated, #3270 had already been fixed before this PR, and nothing about this PR addressed #980.

- All `coord_*()` functions with `xlim` and `ylim` arguments now accept vectors with `NA` as a placeholder for the minimum or maximum value (e.g., `ylim = c(0, NA)` would zoom the y-axis from 0 to the maximum value observed in the data). This mimics the behaviour of the `limits` argument in continuous scale functions (#2907).

``` r
library(ggplot2)
ggplot(mpg, aes(cty, hwy)) + 
  geom_point() +
  coord_cartesian(xlim = c(20, NA), ylim = c(NA, 40))
```

- `coord_trans()` now draws second axes (#2990) and accepts `xlim`, `ylim`, and `expand` arguments to bring it up to feature parity with `coord_cartesian()`. This also closes #3056, which is an abandoned PR that deprecated `limx` and `limy` in favour of `xlim` and `ylim`. This PR does the same, and also fully removes `xtrans` and `ytrans`, which have been deprecated long enough that using them resulted in an error anyway.

- `coord_trans()` now calculates breaks using the expanded range (previously these were calculated using the unexpanded range, which resulted in differences between plots made with `coord_trans()` and those made with `coord_cartesian()`. The expansion for discrete axes in `coord_trans()` was also updated such that it behaves identically to that in `coord_cartesian()` (see #3338 - this was already closed by the OP, but it is the best discussion of this topic).

``` r
library(ggplot2)

p <- ggplot(mpg, aes(class, hwy)) + geom_boxplot() 

# these two plots didn't used to align and sometimes had
# different breaks on the y-axis
p + scale_y_log10()
p + 
  coord_trans(y = "log10") + 
  scale_y_continuous(breaks = scales::log_breaks())
```

- Added regression tests for `coord_trans()`, for which previously there were none.


R/axis-secondary.R
```diff
@@ -187,21 +187,49 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
 
     # patch for date and datetime scales just to maintain functionality
     # works only for linear secondary transforms that respect the time or date transform
-    if (scale$trans$name %in% c("date", "time")){
+    if (scale$trans$name %in% c("date", "time")) {
```
hadley: @thomasp85 can you please review this part of the PR?

R/coord-sf.R
```diff
@@ -127,8 +127,16 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
 
   setup_panel_params = function(self, scale_x, scale_y, params = list()) {
     # Bounding box of the data
-    x_range <- scale_range(scale_x, self$limits$x, self$expand)
-    y_range <- scale_range(scale_y, self$limits$y, self$expand)
+    x_range <- expand_limits_scale(
+      scale_x,
+      expand_default(scale_x, expand = self$expand),
```
hadley: Maybe save `expand_default(scale_x, expand = self$expand)` in variable so the code resembles the other coords a bit more closely?

R/coord-transform.r
```diff
@@ -78,31 +74,25 @@
 #' plot + coord_trans(x = "log10")
 #' plot + coord_trans(x = "sqrt")
 #' }
-coord_trans <- function(x = "identity", y = "identity", limx = NULL, limy = NULL, clip = "on",
-  xtrans, ytrans)
-{
-  if (!missing(xtrans)) {
-    gg_dep("1.0.1", "`xtrans` arguments is deprecated; please use `x` instead.")
-    x <- xtrans
+coord_trans <- function(x = "identity", y = "identity", xlim = NULL, ylim = NULL,
```
hadley: Can you please follow the advice in https://github.com/tidyverse/principles/issues/67 for deprecating an argument?

You also need to mention the removal of `xtrans` and `ytrans` in the news

R/scale-expansion.r
```diff
@@ -0,0 +1,223 @@
...
+#'   scale_x_discrete(expand = expand_scale(add = 2))
+#'
+#' # Reproduce the default range expansion used
+#' # when the 'expand' argument is not specified
+#' ggplot(subset(diamonds, carat > 2), aes(cut, price)) +
+#'   geom_jitter() +
+#'   scale_x_discrete(expand = expand_scale(add = .6)) +
+#'   scale_y_continuous(expand = expand_scale(mult = .05))
+#'
+expand_scale = function(mult = 0, add = 0) {
```
hadley: `=` 😱

R/scale-expansion.r
```diff
@@ -0,0 +1,223 @@
...
+  stopifnot(is.numeric(mult) && is.numeric(add))
```
hadley: I think it's better to organise `stopifnot()` so that each refers to one variable.

R/scale-expansion.r
```diff
@@ -0,0 +1,223 @@
...
+#' Expand a numeric range
+#'
+#' @param limits A numeric vector of length 2 giving the
+#'   range to expand.
+#' @param expand A numeric vector of length 2 (`c(add, mult)`)
+#'   or length 4 (`c(mult_left, add_left, mult_right, add_right)`),
+#'   as generated by [expand_scale()].
+#'
+#' @return The expanded `limits`
+#'
```
hadley: `@export` or `@noRd`?

R/scale-expansion.r
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  scale$expand %|W|% if (scale$is_discrete()) discrete else continuous
+}
+
+#' Expand limits in (possibly) transformed space
+#'
+#' These functions calculate the continuous range in coordinate space
+#' and in scale space. Usually these can be calculated from
+#' eachother using the coordinate system transformation, except
```
hadley:
```suggestion
#' each other using the coordinate system transformation, except
```

R/scale-expansion.r
```diff
@@ -0,0 +1,223 @@
...
+  } else {
+    # using the inverse transform to resolve the NA value is needed for date/datetime/time
+    # scales, which refuse to transform objects of the incorrect type
+    coord_limits <- coord_limits %||% scale$trans$inverse(c(NA_real_, NA_real_))
+    coord_limits_scale <- scale$trans$transform(coord_limits)
+    expand_limits_continuous(limits, expansion, coord_limits_scale)
+  }
+}
+
+expand_limits_continuous <- function(limits, expansion = expand_scale(0, 0), coord_limits = c(NA, NA)) {
```
hadley: I assume most of this code has been moved from elsewhere. Is there anything I should read closely?

R/scale-expansion.r
paleolimbot: ...at the name of this function? It's already exported or I would have changed it...

R/scale-expansion.r
paleolimbot: I think you might have to treat it as new code... there was never any attempt to include coordinate systems in scale transformations prior to this PR (in coord_trans(), scales were just expanded afterward, which doesn't work for discrete scales).

R/axis-secondary.R
paleolimbot: Just as a note...so @thomasp85 know what the purpose of this was - the main change was to make `sec.major_source` in the panel params be in scale-transformed space rather than user space. This results in much simpler drawing code in `coord_trans()`. I renamed the current value `sec.major_source_user` and changed the tests accordingly. This resulted in improving the consistency of the `minor*` elements of the params, which weren't transformed in the same way as the major elements. All of this is tested in the sec-axis tests. I could alternatively add `sec.major_source_scale` `sec.minor_source_scale` and leave the current values alone...

R/scale-expansion.r
clauswilke: I think the comment was about `=` instead of `<-` for assignment.

hadley: At the equals sign instead of `<-`

paleolimbot: I can't believe I didn't see that...

R/axis-secondary.R
thomasp85: LGTM...(I also think we have gotten the secondary axis functionality pretty well-covered by unit tests by now)

R/scale-expansion.r
```diff
@@ -0,0 +1,229 @@
...
+#'   scale_x_discrete(expand = expand_scale(add = 2))
+#'
+#' # Reproduce the default range expansion used
+#' # when the 'expand' argument is not specified
+#' ggplot(subset(diamonds, carat > 2), aes(cut, price)) +
+#'   geom_jitter() +
+#'   scale_x_discrete(expand = expand_scale(add = .6)) +
+#'   scale_y_continuous(expand = expand_scale(mult = .05))
+#'
+expand_scale <- function(mult = 0, add = 0) {
```
hadley: Lets call these functions `expansion_`; but continue to also export, but deprecate, `expand_scale()` for backwards compatibility.

Issue #3379, opened 2019-06-23
Make ggsave return filename


eliocamp:
`ggsave()` now returns `invisible()` (i.e. `NULL`). I think having it (invisibly) return the complete path to the saved image would be useful.

HenrikBengtsson:
Came here with the same feature request; being able to do:
```r
pathname <- ggsave(filename, path = path, ...)
```
would be handy.

yutannihilation:
Thanks for catching this. I also felt this was convenient and nice (and was almost approving the PR), but I'm not sure what (and whether or not) tidyverse has some rule on this kind of functions. For example, `readr::write_csv()` returns the data, not the filename.

``` r
x <- readr::write_csv(data.frame(x = 1:3), tempfile())
x
#>   x
#> 1 1
#> 2 2
#> 3 3
```

<sup>Created on 2020-03-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

But you example seems convincing.

yutannihilation:
Closed by #4396

hesscl:
This change breaks an unintended usage of `ggsave()` where users `+` it to the `ggplot()` they have assembled. It's discussed here [here](https://www.google.com/search?q=ggsave&oq=ggsave&aqs=chrome..69i57j69i60l2j69i61.2382j0j1&sourceid=chrome&ie=UTF-8) on StackOverflow and I thought I would mention it here since it appeared no one had yet.

The discussion here suggested some ambivalence about needing `ggsave()` to return filepaths, so to the extent that it is breaking some folks code I wondered if reverting might be worth consideration.

PR #3397, opened 2019-07-05
Makes ggsave return invisibly plot filename.

eliocamp:
I realised that implementing #3379 needed only the slightest change, so I could do it :).


clauswilke:
@thomasp85 Should we merge this? Looks reasonable to me.

thomasp85:
@eliocamp can you merge in master and add a news bullet - then we can merge this in for the next release

eliocamp:
I rebased and added the news bullet (I never know if I need to add it at the beginning or at the end).

clauswilke:
@eliocamp For future reference, it's much better if you merge in master rather than rebase. Rebasing creates this long list of commits and files changed that is extremely difficult to review. In the end, we squash-merge anyways, so the final commit to the ggplot code base will look the same regardless of whether you rebase or merge.

clauswilke:
Yes, just start over.

PR #3398, opened 2019-07-05
Make position guides customizable

### LINKED ISSUE #3281: Feature Request: Axis Text Warning or Automatic Rotation when Labels Overlap

paleolimbot:
This is a PR to address #3322, and allows the same syntax that is currently used to customize non-position guides to be used for position guides. This allows the strategies for dealing with overlapping labels that were implemented in #3375 to actually be used (closes #3281). Currently, these only work for `CoordCartesian`, but could easily be adapted to `CoordTrans` as well.

The main changes are:

- Added two methods to `Coord`s: `setup_panel_guides()` and `train_panel_guides()`. These are both called from `Layout$setup_panel_guides()` at draw time. Both `Coord` methods return modified versions of the `panel_params`, where the `guides` are kept.
- Added `guide_axis()` and `guide_none()` constructors, which return list S3s similar to the other guides.
- Added S3 methods for `guide_axis()` and `guide_none()` objects that exist for all other guides.
- Added an S3 generic `guide_transform()`, which allows position guides access to the `coord` and `panel_params`. 

Some problems include:

- A circular dependency (the guide object is kept in the panel params, but needs access to it so that it can call `Coord$transform()`). I don't know how to get around this, but am open to suggestions.
- Theoretically one could specify two guides at a single position (e.g., two position guides at the left of the panel). I can't get the `gtable` code to draw this (see [this reprex](https://gist.github.com/paleolimbot/e484639f72f775771bc0e2746e4f1efa) ). Currently the code draws the first guide and issues a warning (there is a test for this).

A reprex giving the main syntax:

<details>

``` r
library(ggplot2)

# plot with overlapping text
p <- ggplot(mpg, aes(cty * 100, hwy * 100)) +
  geom_point() +
  facet_wrap(vars(class))

# axis guides can be customized in the scale_* functions or
# using guides()
p + scale_x_continuous(guide = guide_axis(n_dodge = 2))
```

``` r
p + guides(x = guide_axis(n_dodge = 2))
```

``` r

# can also be used to add a duplicate guide
p + guides(x = guide_axis(n_dodge = 2), y.sec = guide_axis())
```

<sup>Created on 2019-07-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

</details>


R/coord-cartesian-.r
```diff
@@ -103,6 +103,73 @@ CoordCartesian <- ggproto("CoordCartesian", Coord,
...
+
+    panel_params
+  },
+
+  labels = function(self, labels, panel_params) {
+    positions_x <- c("top", "bottom")
+    positions_y <- c("left", "right")
+
+    list(
+      x = lapply(c(1, 2), function(i) panel_guide_label(
```
hadley: Something about the style here looks wonky to me

R/coord-cartesian-.r
```diff
@@ -153,10 +212,41 @@ view_scales_from_scale <- function(scale, coord_limits = NULL, expand = TRUE) {
...
 
-draw_view_scale_axis <- function(view_scale, axis_position, theme) {
-  if(is.null(view_scale) || view_scale$is_empty()) {
+panel_guide_label <- function(guides, position, default_label) {
+  guides <- guides_filter_by_position(guides, position)
+
+  if (length(guides) == 0) {
+    default_label
+  } else {
+    guides[[1]]$title %||% waiver() %|W|% default_label
```
hadley: Isn't this just `guides[[1]]$title %||% default_label` ?

R/guides-axis.r
```diff
@@ -1,4 +1,157 @@
...
+#' @examples
+#' # plot with overlapping text
+#' p <- ggplot(mpg, aes(cty * 100, hwy * 100)) +
+#'   geom_point() +
+#'   facet_wrap(vars(class))
+#'
+#' # axis guides can be customized in the scale_* functions or
+#' # using guides()
+#' p + scale_x_continuous(guide = guide_axis(n_dodge = 2))
+#' p + guides(x = guide_axis(n_dodge = 2))
```
hadley: Maybe use `angle` example for second case? (Just to show another useful argument)

R/coord-cartesian-.r
paleolimbot: I don't think so...`guides[[1]]$title` is usually `waiver()`:

``` r
withr::with_namespace("ggplot2", {
  guides <- list(list(title = waiver()))
  default_label <- "default"
  print(guides[[1]]$title %||% default_label)
  print(guides[[1]]$title %||% waiver() %|W|% default_label)
})
#> list()
#> attr(,"class")
#> [1] "waiver"
#> [1] "default"
```

<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

hadley: In that case do we need the `%||% waiver()`?

paleolimbot: If a `guide()`always has a non-null title we don't...until guides are ggproto I'm not sure there's any guarantee of that? In any case, I will make this bit of code more clear.

``` r
withr::with_namespace("ggplot2", {
  guides <- list(list())
  default_label <- "default"
  print(guides[[1]]$title %|W|% default_label)
  print(guides[[1]]$title %||% waiver() %|W|% default_label)
})
#> NULL
#> [1] "default"
```

<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/guides-none.r
```diff
@@ -0,0 +1,44 @@
...
+#' @export
+#'
+guide_none <- function(title = waiver(), position = waiver()) {
+  structure(
+    list(
+      title = title,
+      position = position,
+      available_aes = "any"
+    ),
+    class = c("guide", "guide_none")
```
thomasp85: Is there a reason why this guide doesn't follow the naming of the other guides? shouldn't the class be. `c('guide', 'none')`

R/coord-cartesian-.r
```diff
@@ -153,10 +214,41 @@ view_scales_from_scale <- function(scale, coord_limits = NULL, expand = TRUE) {
   view_scales
 }
 
-draw_view_scale_axis <- function(view_scale, axis_position, theme) {
-  if(is.null(view_scale) || view_scale$is_empty()) {
+panel_guide_label <- function(guides, position, default_label) {
```
thomasp85: Shouldn't all these helpers below be placed in `coord-.r` they seem global to all coords?

R/guides-none.r
hadley: It's not clear to me what we should do in this situation where the previous code is "wrong" in some sense. (Although here it should be `c("guide_none", "guide")`).

thomasp85: We could elect to fix it once and for all when we finally move guides from S3 to ggproto

hadley: In other words, should we stay consistent with the past (i.e. the previous S3 class naming scheme which we know now is suboptimal) and hence be inconsistent with current naming schemes, or should be consistent with current naming schemes introducing internal inconsistencies? (I think the answer is different to (e.g.) the argument naming scheme because this is internal, not user facing)

thomasp85: I think we should stay consistent for the time being and update them all in one fell swoop during a larger rewrite

R/guides-axis.r
```diff
@@ -1,4 +1,157 @@
...
+#' # axis guides can be customized in the scale_* functions or
+#' # using guides()
+#' p + scale_x_continuous(guide = guide_axis(n_dodge = 2))
+#' p + guides(x = guide_axis(angle = 90))
+#'
+#' # can also be used to add a duplicate guide
+#' p + guides(x = guide_axis(n_dodge = 2), y.sec = guide_axis())
+#'
+#'
+guide_axis <- function(title = waiver(), check.overlap = FALSE, angle = NULL, n_dodge = 1,
```
thomasp85: I think we should stick with old ggplot2 style and call it `n.dodge` instead of `n_dodge` (though I don't like it 🙂)

R/coord-cartesian-.r
```diff
@@ -153,10 +214,41 @@ view_scales_from_scale <- function(scale, coord_limits = NULL, expand = TRUE) {
...
+panel_guides_grob <- function(guides, position, theme) {
+  guides <- guides_filter_by_position(guides, position)
+  grobs <- lapply(guides, guide_gengrob, theme)
+
+  if (length(grobs) == 0) {
     return(zeroGrob())
+  } else if (length(grobs) == 1) {
+    grobs[[1]]
+  } else {
+    warning(
```
thomasp85: There is something weird going on with the logic here... If I modify your own examples I get:
```r
p <- ggplot(mpg, aes(cty * 100, hwy * 100)) +
    geom_point() +
    facet_wrap(vars(class))

# axis guides can be customized in the scale_* functions or
# using guides()
p + scale_x_continuous(guide = guide_axis(position = 'top'))
#> Warning messages:
#> 1: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
#> 2: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
#> 3: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
#> 4: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
#> 5: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
#> 6: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
#> 7: In panel_guides_grob(panel_params$guides, position = "top", theme = theme) :
#>   More than one position guide found at `position = "top". Only showing the first guide.
```

R/guides-none.r
```diff
@@ -0,0 +1,44 @@
+
+#' Empty guide
+#'
+#' This guide draws nothing.
+#'
+#' @inheritParams guide_axis
+#'
+#' @export
+#'
+guide_none <- function(title = waiver(), position = waiver()) {
```
thomasp85: I think we should make sure this can be used for non-positional guides as well to turn them off. If I do that now I get:
```r
ggplot(mpg, aes(cty * 100, hwy * 100)) +
  facet_wrap(vars(class)) +
  scale_color_discrete(guide = guide_none())
#> Error in if (g$order == 0) { : argument is of length zero 
```

R/scale-view.r
```diff
@@ -87,6 +109,8 @@ ViewScale <- ggproto("ViewScale", NULL,
   # map, rescale, and make_title need a reference
```
thomasp85: I'm somewhat conflicted on the naming and implementation of `ViewScale`. At a glance it looks like a `Scale` object with many (all?) of the same methods etc. The file naming also suggest it is a specific scale type. But obviously this is not the case. At the very least we should rename the file to put it away from the other scale implementation, but can you also share a bit of the reasoning for this design (you may have already discussed this with Hadley, but I'm curious as well)

R/guides-.r
```diff
@@ -148,6 +154,11 @@ legend_position <- function(position) {
   }
 }
 
+# resolve the guide from the scale and guides
+resolve_guide <- function(aesthetic, scale, guides, default = "none", null = "none") {
+  guides[[aesthetic]] %||% scale$guide %|W|% default %||% null
+}
+
 # validate guide object
 validate_guide <- function(guide) {
```
thomasp85: This function does not seem to use the same lookup as the rest of ggplot2 when resolving a string to a function (ie it uses `match.fun()` instead of `find_global()`). Unless there is a compelling reason for this it should use the same search mechanism (this may mean that the plot environment has to be passed down to this function)

R/scale-continuous.r
```diff
@@ -77,14 +77,14 @@ NULL
 scale_x_continuous <- function(name = waiver(), breaks = waiver(),
                                minor_breaks = waiver(), labels = waiver(),
                                limits = NULL, expand = waiver(), oob = censor,
-                               na.value = NA_real_, trans = "identity",
+                               na.value = NA_real_, trans = "identity", guide = waiver(),
```
thomasp85: Somewhere in the code we should check and fail gracefully if the wrong type of guide is passed into the different scales. E.g. trying to use a `guide_legend()` with a `scale_x_continuous()` gives ` Error in f(..., self = self) : unused argument (breaks) `, instead of `Error: guide_legend() cannot be used with a position scale` (or something similar)

R/coord-flip.r
```diff
@@ -72,14 +72,29 @@ CoordFlip <- ggproto("CoordFlip", CoordCartesian,
...
+  )
+
+  invisible(scale)
+}
 
-flip_labels <- function(x) {
+# maintaining the position of the x* and y* names is
+# important for re-using the same guide_transform()
+# as CoordCartesian
+flip_axes_labels <- function(x) {
```
thomasp85: minor, but shouldn't it be `flip_axis_labels` for correct plural?

R/guides-axis.r
```diff
@@ -1,4 +1,157 @@
...
+      # general
+      order = order,
+      position = position,
+
+      # parameter
+      available_aes = c("x", "y"),
+
+      name = "axis"
+    ),
+    class = c("guide", "axis")
```
thomasp85: I will suggest that we add an additional layer of class here (don't have a good name). There should be an abstract subclass of `guide` that we use to test for inheritance of when figuring out if the guide is appropriate for the scale. Then this will subclass that abstract class. This will make it easier to extend with new axis guides in the future while making it simple to test whether the guide works for the scale

R/guides-axis.r
```diff
@@ -1,4 +1,157 @@
...
+
+#' @export
+guide_transform.axis <- function(guide, coord, panel_params) {
+  if (is.null(guide$position) || nrow(guide$key) == 0) {
+    return(guide)
+  }
+
+  aesthetics <- names(guide$key)[!grepl("^\\.", names(guide$key))]
+
+  if (all(c("x", "y") %in% aesthetics)) {
```
thomasp85: I think this needs to be switched around, no? you want to check if there is no other aesthetics than `x` and `y`, right?

R/guides-none.r
thomasp85: let's just keep it for now... your reasons seem sensible

R/coord-cartesian-.r
thomasp85: Can you provide an example of how it doesn't play nicely with `labs()`

R/scale-view.r
thomasp85: ok

R/guides-.r
thomasp85: Can you add a comment to the code so we don't forget

thomasp85:
There will be facilities for fast text-box dimension calculation down the line without going into C++ so we should be able to get this into ggplot2 proper at some point (not needed for this PR though)

Do you have an overview of which packages this will affect? If we could reach out to them as soon as possible we can make sure they are ready for the next release

thomasp85:
I think, with the one minor remaining change request, that we should merge this in and begin bracing for impact 🙂 It is better to get a lot of hands on this before we make a release...

any objections @hadley ?

R/coord-cartesian-.r
paleolimbot: This is what I had in mind:

``` r
library(ggplot2)
p <- ggplot(mpg, aes(cty, hwy)) +
  geom_point() 

p +
  guides(x.sec = guide_axis()) +
  labs(x.sec = "fish")
```



<sup>Created on 2019-09-22 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/guides-.r
paleolimbot: Done!

thomasp85:
When you've added a NEWS item you are free to merge... then we'll begin working through any possible consequences

smouksassi:
@paleolimbot  after this merge some code that used to work  is failing removing the faceting works but why now facet with free_x is not ? the breaks and labels have the same length

``` r
library(ggplot2)
ggplot(mtcars,aes(mpg ,disp)) +
  geom_point()+
facet_grid(~cyl,scales="free_x")+
  scale_x_continuous(breaks=c(10,15,20,21,25,30),
                     labels=c("10","15","20","21","25","30") )
#> Error: Breaks and labels are different lengths
       
ggplot(mtcars,aes(mpg ,disp)) +
  geom_point()+
    scale_x_continuous(breaks=c(10,15,20,21,25,30),
                     labels=c("10","15","20","21","25","30") )
```



<sup>Created on 2019-10-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
@smouksassi Could you file a separate issue please? I can confirm this problem.

thomasp85:
FWIW, we are aware of it as it has cropped up in revdepchecks. @paleolimbot is working through these right now

smouksassi:
@clauswilke   let me know if we still need to open a new issue it seems it is already taken care of?

clauswilke:
I think it never hurts to have an explicit issue of a known problem. @paleolimbot can close it when he submits his fix.

Issue #2630, opened 2018-05-18
Change the default estimation method for gam in geom_smooth()


clauswilke:
As recommended by the maintainer of the mgcv package, Simon Wood, it might be better to set `method = "REML"` in the `gam()` call for `geom_smooth()`. I quote:

> REML smoothing parameter estimation is less prone to occasional under-smoothing than GCV (e.g. `gam(..., method="REML")`)

clauswilke:
One more comment from Simon Wood:
> Theoretically REML will undersmooth relative to GCV in the large sample limit and have slightly worse MSE performance, but for me the finite sample performance trumps this: Reiss and Ogden, 2009, JRSSB give some theory for why.

The Reiss and Ogden paper is available here: http://www.jstor.org/stable/40247585

`gam(..., method = "REML")` will work for mgcv versions >=1.5-0, March 2009.

hadley:
This should be a straightforward tweak to the existing code, and just needs a news bullet. It is likely to create small changes in some existing plots, but I don't think the differences should be too serious (although it might cause visual tests to fail)

PR #3402, opened 2019-07-08
Stat smooth+reml

ikosmidis:
Closes issue #2630


NEWS.md
```diff
@@ -1,40 +1,43 @@
 # ggplot2 (development version)
 
+* `stat_smooth()` user `REML` by default, if `method = "gam"` and
+  `gam`'s method is not specified (@ikosmidis, #2630).
+
 * `expand_scale()` was deprecated in favour of `expansion()` for setting
```
hadley: I think something has gone wrong with this edit as its made a bunch of whitespace changes to the file. 

To help me figure out the source of the problem, are you using RStudio?

R/stat-smooth.r
```diff
@@ -5,7 +5,7 @@
 #'   For `method = "auto"` the smoothing method is chosen based on the
 #'   size of the largest group (across all panels). [stats::loess()] is
 #'   used for less than 1,000 observations; otherwise [mgcv::gam()] is
-#'   used with `formula = y ~ s(x, bs = "cs")`. Somewhat anecdotally,
+#'   used with `formula = y ~ s(x, bs = "cs")` with \code{method = "REML"}. Somewhat anecdotally,
```
hadley: Can you please use markdown rather than Rd syntax?

R/stat-smooth.r
```diff
@@ -127,12 +124,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
         xseq <- seq(range[1], range[2], length.out = n)
       }
     }
-    # Special case span because it's the most commonly used model argument
+
+    ## Special case span because it's the most commonly used model argument
```
hadley: It looks like you just changed `#` to `##`. This makes me suspect that you're using ESS?

R/stat-smooth.r
```diff
@@ -127,12 +124,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
...
     if (identical(method, "loess")) {
       method.args$span <- span
     }
 
-    if (is.character(method)) method <- match.fun(method)
+    ## If gam and gam's method is not specified by the user then use REML
+    if (identical(method, "gam") & is.null(method.args$method)) {
```
hadley:
```suggestion
    if (identical(method, "gam") && is.null(method.args$method)) {
```

hadley: You should always use `&&` in if statements because it "short circuits"

R/stat-smooth.r
```diff
@@ -127,12 +124,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
...
+        method.args$method <- "REML"
+    }
+
+    if (is.character(method)) {
+        if (identical(method, "gam")) {
```
hadley: Can you please check the style here? You can use the styler package to match our usual guidelines (https://style.tidyverse.org)

R/stat-smooth.r
ikosmidis: Yes!

R/stat-smooth.r
ikosmidis: Done

R/stat-smooth.r
ikosmidis: Should be OK now; used styler on stat-smooth.R

R/stat-smooth.r
ikosmidis: Indeed. Missed that

NEWS.md
ikosmidis: I am using ESS. Any advice on how I can prevent this next time?

hadley:
Can you please find @gaborcsardi and get him to help you tweak your ESS settings so that it doesn't reformat these files?

gaborcsardi:
> Can you please find @gaborcsardi and get him to help you tweak your ESS settings so that it doesn't reformat these files?

@ikosmidis  I am at the table next to Hadley if you need help. You probably need 

```
(setq-default indent-tabs-mode nil)
```

For the spaces/indentation, and
```
(setq ess-indent-with-fancy-comments nil)
```
to avoid the fancy ess comment character.

R/stat-smooth.r
```diff
@@ -127,12 +125,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
...
+    ## If gam and gam's method is not specified by the user then use REML
+    if (identical(method, "gam") && is.null(method.args$method)) {
+      method.args$method <- "REML"
+    }
+
+    if (is.character(method)) {
+      if (identical(method, "gam")) {
+        method <- mgcv::gam
+      }
+      else {
```
hadley: Can you please move the `else` to be on the same line as `}`

R/stat-smooth.r
```diff
@@ -127,12 +125,25 @@ StatSmooth <- ggproto("StatSmooth", Stat,
...
-    # Special case span because it's the most commonly used model argument
+
+    ## Special case span because it's the most commonly used model argument
     if (identical(method, "loess")) {
       method.args$span <- span
     }
 
-    if (is.character(method)) method <- match.fun(method)
+    ## If gam and gam's method is not specified by the user then use REML
+    if (identical(method, "gam") && is.null(method.args$method)) {
```
hadley: This check should probably be `identical(method, mgcv::gam)` and happen after the conversion from character to function in the next block.

R/stat-smooth.r
ikosmidis: done

R/stat-smooth.r
ikosmidis: I changed the design slightly here. In master `params$method` was replaced by `mgcv::gam` if it was equal to `"gam"`, withiin `setup_params`. Now `params$method` keeps being `gam` and matching names to functions happens in `compute_group`. This is to preserve symmetry with how the other methods are handled (e.g. `loess`) where `method` is `match.fun`ed. 

The `is.character` check may as well be redundant but I kept it in just in case I was breaking anything I could not foresee. 

Happy to revert to the earlier design and test functions being identical instead of character strings, if this causes any issues; though `ggplot2` tests were successful my side.

ikosmidis: Or perhaps an alternative closer to what you are suggesting, that avoids checking twice is method is `"gam"` can be something like:

```r
    if (is.character(method)) {
        if (identical(method, "gam")) {
            method <- mgcv::gam
            if (is.null(method.args$method)) {
                method.args$method <- "REML"
            }
        } else {
            method <- match.fun(method)
        }
    }
```

Again the `is.character` may not be needed anymore.

ikosmidis: OK; just figured out that ggplot2 wants the user to be able to supply functions for `method` as well as character strings so checking `method` against `mgcv::gam` after matching makes more sense. Just pushed a fix.

hadley:
Thanks @ikosmidis!

Issue #3299, opened 2019-05-03
Improve documentation for scale_*_gradient


paleolimbot:
There are several opportunities to improve the documentation for [scale_*_gradient()](https://ggplot2.tidyverse.org/reference/scale_gradient.html). This is a documentation issue intended for the tidyverse dev day (#3267).

- In `scale_*_gradient()`, it may be worth mentioning that passing `na.value = NA` is useful when used with the `fill` aesthetic (`fill = NA` results in no fill), but will not work for the `colour` aesthetic.

- In `scale_*_gradient()`, the behaviour of `midpoint` when a `trans` is applied should be documented (depends on resolution of #3198).

PR #3407, opened 2019-07-08
Improve documentation for `scale_*_gradient`

MeganBeckett:
Addresses #3299 

Added examples to illustrate the different behaviour when setting `na.value = NA` when used with the fill and colour aesthetics, and data has missing values.

The second suggested bullet point of #3299 is dependent on a bug fix and hence I did not document this.

### DIFF
--- a/R/scale-gradient.r
+++ b/R/scale-gradient.r
@@ -29,6 +29,14 @@
 #'   z2 = abs(rnorm(100))
 #' )
 #'
+#' # Dataframe with missing values
+#' df_na <- data.frame(
+#'   value = seq(1,20),
+#'   x = runif(20),
+#'   y = runif(20),
+#'   z1 = c(rep(NA, 10), rnorm(10))
+#' )
+#'
 #' # Default colour scale colours from light blue to dark blue
 #' ggplot(df, aes(x, y)) +
 #'   geom_point(aes(colour = z2))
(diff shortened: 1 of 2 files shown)


R/scale-gradient.r
```diff
@@ -29,6 +29,14 @@
 #'   z2 = abs(rnorm(100))
 #' )
 #'
+#' # Dataframe with missing values
```
DavisVaughan: I don't think you need a comment here, the name is descriptive enough!

R/scale-gradient.r
```diff
@@ -29,6 +29,14 @@
...
+#' df_na <- data.frame(
+#'   value = seq(1,20),
```
DavisVaughan: Could you add a space after the `,`? So `seq(1, 20)`?

R/scale-gradient.r
```diff
@@ -54,6 +62,19 @@
 #'   scale_colour_gradient(low = "white", high = "black")
 #' # Avoid red-green colour contrasts because ~10% of men have difficulty
 #' # seeing them
+#'
+#' # Pass na.value = NA with the fill aesthetic to result in no fill for missing values
```
DavisVaughan: Suggested change which highlights the fact that the original axis range is kept:

```
# Use `na.value = NA` to hide missing values 
# but keep the original the axis range
```

R/scale-gradient.r
```diff
@@ -54,6 +62,19 @@
...
+#' ggplot(df_na, aes(x = value, y)) +
+#'   geom_bar(aes(fill = z1), stat = "identity") +
+#'   scale_fill_gradient(low = "yellow", high = "red",
+#'                       na.value = NA)
```
DavisVaughan: Maybe move `na.value = NA` to the same line as `scale_fill_gradient()`?

R/scale-gradient.r
```diff
@@ -54,6 +62,19 @@
...
+#'
+#' # Pass na.value = NA with the colour aesthetic to hide data points with missing values
```
DavisVaughan: This comment can probably be removed then

R/scale-gradient.r
```diff
@@ -54,6 +62,19 @@
...
+#'  ggplot(df_na, aes(x, y)) +
+#'    geom_point(aes(colour = z1)) +
+#'    scale_colour_gradient(low = "yellow", high = "red",
+#'                          na.value = NA)
```
DavisVaughan: Same comment about being on the same line

PR #3408, opened 2019-07-08
added documentation to geom_area

### LINKED ISSUE #2873: geom_area doesn't respond to ymax aesthetic
The documentation for `geom_area()` indicates "For each x value, geom_ribbon displays a y interval defined by ymin and ymax. geom_area is a special case of geom_ribbon, where the ymin is fixed to 0."

When I tried setting `ymax` only, I got an error, and it is actually using the `y` aesthetic instead of `ymax`.

``` r
library(ggplot2)
#> Warning: package 'ggplot2' was built under R version 3.4.4
ggplot(data.frame(x = 1:3), aes(x = x, ymin = 0, ymax = 1)) + geom_area()
#> Error in eval(substitute(list(...)), `_data`, parent.frame()): object 'y' not found

ggplot(data.frame(x = 1:3), aes(x = x, y = 1)) + geom_area()
```

That suggests that for `geom_area()`, either
* the documentation needs to be updated so that it indicates "`geom_area()` uses the aesthetic `y` instead of `ymax`" or
* the functionality needs to be updated so that `geom_area()` uses `ymax` instead of `y`.

AnneLyng:
resolves #2873 (documentation added to clarify aesthetics for geom_area compared to geom_ribbon)


NEWS
```diff
@@ -1,3 +1,11 @@
+ggplot2 3.2.0.9000
+----------------------------------------------------------------
+
+* Corrected documentation for `geom_area` aethestic (`ymax` to
```
hadley: This should go in `NEWS.md`

R/geom-ribbon.r
```diff
@@ -2,7 +2,8 @@
 #'
 #' For each x value, `geom_ribbon` displays a y interval defined
 #' by `ymin` and `ymax`. `geom_area` is a special case of
-#' `geom_ribbon`, where the `ymin` is fixed to 0.
+#' `geom_ribbon`, where the `ymin` is fixed to 0 and `y` is used instead
```
hadley: Since you're in here, can you please add `()` after function names?

NEWS
AnneLyng: moved to NEWS.md

R/geom-ribbon.r
AnneLyng: sure!  added () to the function names

hadley:
Thanks @AnneLyng! (In the future, I'd recommend creating a branch rather than committing straight from master; it'll make your life a little easier in the long run)

PR #3410, opened 2019-07-08
fixes stat_ellipse misleading documentation

### LINKED ISSUE #2776: Documentation of stat_ellipse() is misleading

violetr:
fixes #2776 : data ellipse instead of confidence ellipse

### DIFF
--- a/R/stat-ellipse.R
+++ b/R/stat-ellipse.R
@@ -1,12 +1,12 @@
-#' Compute normal confidence ellipses
+#' Compute normal data ellipses
 #'
 #' The method for calculating the ellipses has been modified from
-#' `car::ellipse` (Fox and Weisberg, 2011)
+#' `car::dataEllipse` (Fox and Weisberg, 2011)
 #'
 #' @references John Fox and Sanford Weisberg (2011). An \R Companion to
 #'   Applied Regression, Second Edition. Thousand Oaks CA: Sage. URL:
 #'   \url{http://socserv.socsci.mcmaster.ca/jfox/Books/Companion}
-#' @param level The confidence level at which to draw an ellipse (default is 0.95),
+#' @param level The level at which to draw an ellipse,
 #'   or, if `type="euclid"`, the radius of the circle to be drawn.
 #' @param type The type of ellipse.
 #'   The default `"t"` assumes a multivariate t-distribution, and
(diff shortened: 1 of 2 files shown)


hadley:
Thanks @violetr!

Issue #3298, opened 2019-05-03
Improve scale_(x|y)_continuous documentation


paleolimbot:
There are several opportunities to improve the documentation for [scale_(x|y)_continuous()](https://ggplot2.tidyverse.org/reference/scale_continuous.html). This is a documentation issue intended for the tidyverse dev day (#3267). 

- The current parameter documentation for `oob` describes `scales::censor`, which may be worth linking to. `scales::squish` and `scales::squish_infinite` can be used here as well, which is not obvious from either the ggplot2 or scales documentation.

- The parameter documentation for `sec.axis` in scale_continuous should link to `sec_axis()` (and possibly be removed from the "see also" section).

- The links to `scales::rescale()` aren't active on the site, and it isn't obvious that `rescale()` isn't a function in ggplot2 (from several scale-related documentation pages).

PR #3411, opened 2019-07-08
Improve scale_(x|y)_continuous docs

CateGitau:
issue #3298 
The third issue on scales::rescale() link not being active seems to have been solved. probably only thing left is to include somewhere in the doc that rescaler() is not a function in ggplot2.


R/scale-.r
```diff
@@ -45,7 +45,9 @@
 #'   scales, which ways use [scales::rescale()].
 #' @param oob Function that handles limits outside of the scale limits
 #'   (out of bounds). The default ([scales::censor()]) replaces out of
-#'   bounds values with `NA`.
+#'   bounds values with `NA`. ([scales::squish()]) can also be used,
```
hadley: I think this might now be better as a bulleted list. Do you want to have a go at formatting it that way?

CateGitau: Sure, on it.

Issue #3382, opened 2019-06-25
Deprecate ggdep


hadley:
Since once #3380 is merged it will no longer be used, and it doesn't align with the latest tidyverse standards for deprecated.

Part of #3057

paleolimbot:
This is brief advice, but I think following it is a reasonable way to deprecate `gg_dep()`: tidyverse/principles#72

PR #3412, opened 2019-07-08
Deprecate ggdep

perezp44:
@hadley it is good for #3382

### DIFF
--- a/R/utilities.r
+++ b/R/utilities.r
@@ -178,6 +178,7 @@ rescale01 <- function(x) {
 
 #' Give a deprecation error, warning, or message, depending on version number.
 #'
+#' This function is deprecated.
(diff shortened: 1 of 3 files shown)


R/utilities.r
```diff
@@ -178,6 +178,7 @@ rescale01 <- function(x) {
 
 #' Give a deprecation error, warning, or message, depending on version number.
 #'
+#' This function is deprecated.
 #' Version numbers have the format <major>.<minor>.<subminor>, like 0.9.2.
```
hadley: Can you please remove the old text below? Otherwise, looks good.

perezp44: I think I haver remove the old text

hadley:
Looks good - can you please also add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

PR #3415, opened 2019-07-08
Improved stat_function() documentation #3362

### LINKED ISSUE #3362: Layer data mapping are ignored in stat_function()
I have tried to use stat_function in combination with `aes(group = ...)` and `args=list(...)`, taking into account the documentation
```
stat_function() understands the following aesthetics (required aesthetics are in bold):
- group
- y
```
of what I expected different curve plots depending on arguments taken from each group, but I only obtained one curve.

I have asked in [stackoverflow](https://stackoverflow.com/questions/56560466/is-it-possible-to-use-stat-function-by-group) if there is a way to do this, but further, I was looking for examples using these aesthetics combined with stat_function, and I did not find anything.

I copy my example of the link:
```
library(ggplot2)
mtcars$vs <- as.factor(mtcars$vs)
ggplot(mtcars,aes(x=mpg, fill = vs, colour = vs)) + geom_density(alpha = 0.1) + 
    stat_function(aes(group = vs), fun = dnorm, args = list(mean = mean(mtcars$mpg), sd = sd(mtcars$mpg)))
```

I have also tried deleting `fill = vs, colour = vs`, and changing it by `group = vs`, but no result. Since I have also not found any example, not even in the vignette `("ggplot2-specs")`, I wonder if this really works.

Thank you!

thiyangt:
Closes #3362

### DIFF
--- a/R/geom-point.r
+++ b/R/geom-point.r
@@ -35,7 +35,6 @@
 #'   often aesthetics, used to set an aesthetic to a fixed value, like
 #'   `colour = "red"` or `size = 3`. They may also be parameters
 #'   to the paired geom/stat.
-#' @inheritParams layer
 #' @export
 #' @examples
 #' p <- ggplot(mtcars, aes(wt, mpg))
--- a/R/stat-function.r
+++ b/R/stat-function.r
@@ -5,8 +5,6 @@
 #' and the results are drawn (by default) with a line.
 #'
 #'
-#'
-#' @eval rd_aesthetics("stat", "function")
 #' @param fun Function to use. Either 1) an anonymous function in the base or
 #'   rlang formula syntax (see [rlang::as_function()])
 #'   or 2) a quoted or character name referencing a function; see examples. Must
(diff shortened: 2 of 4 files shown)


lionel-:
Would you like to also add a warning when `mapping` or `data` is supplied, as suggested in https://github.com/tidyverse/ggplot2/issues/3362#issuecomment-501716612?

lionel-:
Thanks! The warnings will also need unit tests that you can add in `tests/testthat/test-stats-function.r`. This should look like:

```r
expect_warning(stat_function(aes(), fun = identity), "Ignoring `mapping`")
```

However I see the warning currently is:

```r
warn_overwritten_args("stat_function()", "mapping", "group")
#> Warning message:
#> stat_function(): Ignoring `mapping` because `group` was provided.
```

Which doesn't look right. It's probably better to just call `warning()` directly with custom messages. Any suggestion as to how best formulate the two warnings @paleolimbot? I'm not sure I entirely understand its semantics.

lionel-:
The PR just needed another small change, the warnings needed to be above the `layer()` call, otherwise the `stat_function()` would not return anything. No big deal, I have pushed a quick fix: https://github.com/tidyverse/ggplot2/pull/3415/commits/6ba927541fcac743e0164c777ed60b27e380721b.

Thanks for these improvements to ggplot2!

Issue #2863, opened 2018-08-27
Add section on colour blindness to colour scales


hadley:
i.e. describe the problems with the default scales, and what you might want to use instead.

tdelpopolo:
why not change the default to be more inclusive?
https://twitter.com/SeanTrende/status/1062142416952664064

paleolimbot:
This would be an excellent addition to `?scale_(colour|fill)_discrete` and/or `?scale_(colour|fill)_continuous`.

I know little about this, but I know that http://colorbrewer2.org gives advice on which brewer scales are colour-blind friendly, and I believe the virdis scale is designed to be colour-blind friendly by default. Possibly worth linking to is the [colorblindr](https://github.com/clauswilke/colorblindr) package, which simulates common forms of color blindness.

clauswilke:
I would recommend linking to the [colorspace package](http://colorspace.r-forge.r-project.org/) rather than colorblindr. Also check out this paper: https://arxiv.org/abs/1903.06490

PR #3417, opened 2019-07-08
advice users regarding colourblindness

Ahobert:
closes https://github.com/tidyverse/ggplot2/issues/2863

### DIFF
--- a/R/scale-colour.r
+++ b/R/scale-colour.r
@@ -12,18 +12,17 @@
 #'   [scale_fill_gradient()], and [scale_fill_viridis_c()]
 #' @export
 #' @rdname scale_colour_continuous
-#' @section Color Blindness
+#' @section Color Blindness:
 #' Many color palettes derived from RGB combinations (like the "rainbow" color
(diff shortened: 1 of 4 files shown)


R/scale-colour.r
```diff
@@ -4,22 +4,36 @@
...
+#' uniform in both colour and black-and-white display is an easy option to
+#' ensure good perceptive properties of your visulizations.
+#' The colorspace package offers functionalities
+#' - to generate color palettes with good perceptive properties,
+#' - to analyse a given color palette, like emulating color blindness,
+#' - and to modify a given color palette for better perceptivity.
+#'
+#' For more information on color vision deficiencies and suitable color choices
+#' see the [paper on the colorspace package](https://arxiv.org/abs/1903.06490)
+#' and references therein.
```
Ahobert: Duplicate code from `scale-manual.r`
Considered using [roxygen templates](https://stackoverflow.com/questions/15100129/using-roxygen2-template-tags), but decided against it because of overhead.

Issue #2767, opened 2018-07-20
position_dodge2() works without a grouping variable in a layer


aosmith16:
When working with `position_dodge()`, every layer that is being dodged needs to have the variable that represents the groups present.  When working with multiple layers that need dodging this usually means mapping the variable globally or making sure each layer as an aesthetic mapping involving that grouping variable.  

This is not true for `position_dodge2()`, which seems to allow dodging for all layers as long as any layer has the variable mapped.

Here is an example to that shows `position_dodge2()` working where `position_dodge()` fails.  Is this the expected behavior?

``` r
library(ggplot2)
dat = data.frame(bin = factor(c(1,2,3,1,2,3)),
              type = factor(c(1,1,1,2,2,2)), 
              beta = c(10,20,30,40,50,60))

# Put "fill" in global aes() and dodging works in all layers
ggplot(dat, aes(x = bin, y = beta, fill = type)) + 
    geom_col(position = "dodge")  +
    geom_errorbar(aes( ymin = beta - 4, ymax = beta + 4),
                  position = position_dodge(width = .9) )
```

![](https://i.imgur.com/HKk4NzA.png)

``` r

# Move "fill" to geom_bar() and geom_errorbar() doesn't dodge
ggplot(dat, aes(x = bin, y = beta)) + 
    geom_col(aes(fill = type), position = "dodge")  +
    geom_errorbar(aes( ymin = beta - 4, ymax = beta + 4),
                  position = position_dodge(width = .9) )
```

![](https://i.imgur.com/2wy9fKN.png)

``` r

# But if use position_dodge2() all layers dodge
ggplot(dat, aes(x = bin, y = beta)) +
    geom_col(aes(fill = type), position = "dodge2")  +
    geom_errorbar(aes( ymin = beta - 4, ymax = beta + 4),
                  position = position_dodge2(width = .9) )
```

![](https://i.imgur.com/S782SG2.png)

Created on 2018-07-20 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

paleolimbot:
The `position_*()` functions dodge based on the x/xmin/xmax of items on the panel. `position_dodge()` and `position_dodge2()` have different strategies, and as I understand it `position_dodge()` incorporates the `group` whereas `position_dodge2()` does not. Therefore, I think that the behaviour you describe is entirely correct!

aosmith16:
Oh, good to know!

[...]

PR #3418, opened 2019-07-08
Issue 2767

CateGitau:
issue #2767 

Added some information about position_dodge2()


hadley:
It looks like your changes got tangled up with your other PR - this is typically because you're not creating a branch for each PR. Undoing is a little tricky so you might want to find a helper.

CateGitau:
Fixed this.

hadley:
Generally, I don't think we bother with NEWS bullets for documentation fixes (unless they're v. large)

PR #3420, opened 2019-07-08
Improving documentation for geom_histogram().

### LINKED ISSUE #3269: Improved documentation for geom_histogram()
In response to #3267: `geom_histogram()` would benefit from a paragraph in Details explaining how in practice one can achieve the bin settings one wants. Currently, the text simply says:

> By default, the underlying computation (`stat_bin()`) uses 30 bins; this is not a good default, but the idea is to get you experimenting with different bin widths. You may need to look at a few to uncover the full story behind your data.

This will prompt most people to change the `bins` parameter, but in practice that rarely works out well. A much better strategy is to use `binwidth` combined with either `center` or `boundary`.

angelica93:
- Details about modifing binwidths and center/boundary
- Removed repeated word from stat-bin documentation

Closes #3269

### DIFF
--- a/R/geom-histogram.r
+++ b/R/geom-histogram.r
@@ -12,8 +12,10 @@
 #'
 #' By default, the underlying computation (`stat_bin()`) uses 30 bins;
 #' this is not a good default, but the idea is to get you experimenting with
-#' different bin widths. You may need to look at a few to uncover the full
-#' story behind your data.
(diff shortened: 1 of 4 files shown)


angelica93:
By mistake this pull request wasn't in my name. I will delete this and resubmit in a few minutes.

PR #3421, opened 2019-07-08
Explicit warning about data removal with limits on positional scales

### LINKED ISSUE #2887: Should limits be enforced on the calculated geoms?
(This is originally reported on #2879, and I asked on [RStudio Community](https://community.rstudio.com/t/how-can-i-specify-the-limit-of-geom-bar-properly-when-the-x-axis-is-continuous/13709). But, I couldn't find any reasonable answer so far...)

When the x axis is continuous and has limits, `geom_bar()` and `geom_col()` won't automatically consider the width of bars, which means the bars at both ends disappear even if the values are within the range of the limits. I feel this behaviour is counter-intuitive, whereas this won't happen with discrete scales.

``` r
library(ggplot2)

d <- data.frame(x = 1:3)

# bars at both ends disappear
ggplot(d) +
  geom_bar(aes(x)) +
  scale_x_continuous(limit = c(1, 3))
#> Warning: Removed 2 rows containing missing values (geom_bar).
```

``` r
# no bars disappear
ggplot(d) +
  geom_bar(aes(factor(x))) +
  scale_x_discrete(limit = c("1", "2", "3"))
```

Created on 2018-09-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

Though I don't come up with how to control whether or not to ignore limits, IIUC about the semantics, `limits` should not be enforced on the computed geoms, but on the values before calculation.

After some investigation, I'm wondering if it's possible to reset `limits` as long as `range` of the scale after computing the widths of the geoms (`compute_geom_1()`) and computing positions (`compute_position()`). More preciesely, `reset_scales()` here:

https://github.com/tidyverse/ggplot2/blob/f636f34597e77e9a2a99783aff6233b838490ff3/R/plot-build.r#L68-L78

Though I don't come up with how to implement this yet, does my explanation make sense?

Elesday:
Fixes #2887 

Added an explicit warning in scales documentation about limits sometimes removing data.

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -38,6 +38,9 @@
 #'     Use `NA` to refer to the existing minimum or maximum
 #'   - A function that accepts the existing (automatic) limits and returns
 #'     new limits
+#'
+#'  Note that setting limits on positional scales can unpredictably remove data.
+#'  If the purpose is to zoom, use the limit argument in the coordinate system.
 #' @param rescaler A function used to scale the input values to the
 #'   range \[0, 1]. This is always [scales::rescale()], except for
 #'   diverging and n colour gradients (i.e., [scale_colour_gradient2()],
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -39,7 +39,7 @@
 #'   - A function that accepts the existing (automatic) limits and returns
 #'     new limits
 #'
-#'  Note that setting limits on positional scales can unpredictably remove data.
+#'  Note that setting limits on positional scales will **remove** data outside of the limits.
 #'  If the purpose is to zoom, use the limit argument in the coordinate system.
 #' @param rescaler A function used to scale the input values to the
 #'   range \[0, 1]. This is always [scales::rescale()], except for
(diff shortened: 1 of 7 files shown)


R/scale-.r
```diff
@@ -38,6 +38,9 @@
 #'     Use `NA` to refer to the existing minimum or maximum
 #'   - A function that accepts the existing (automatic) limits and returns
 #'     new limits
+#'
+#'  Note that setting limits on positional scales can unpredictably remove data.
```
hadley:
```suggestion
#'  Note that setting limits on positional scales will **remove** data outside of the limits.
```

R/scale-.r
```diff
@@ -38,6 +38,9 @@
...
+#'  If the purpose is to zoom, use the limit argument in the coordinate system.
```
hadley: Maybe add link to `coord_cartesian()` here?

hadley:
Thanks @Elesday!

PR #3423, opened 2019-07-08
Document breaks/bins/binwidth parameters in stat/geom contour

### LINKED ISSUE #2472: stat_contour's breaks parameter is undocumented

sinarueeger:
- Added explicit `breaks`, `nbin` and `binwidth` arguments to `stat_contour()` and `geom_contour()`, along with documentation for `stat_contour()`. 

- In the parameter description, I added a default value to `breaks`. Not sure this is ok, as it is the default value to all parameter (when `breaks=NULL`, `nbin=NULL`, `binwidth=NULL`).

Closes #2472.


sinarueeger:
## Questions
- Does this need a news bullet?
- Shall we abandon `complete = FALSE` as proposed in PR #2926?
- Out of curiosity, why does `geom-contour.r` have two `@export`'s?

sinarueeger:
Excellent - thanks @paleolimbot for merging + thanks for carrying my changes through the merge conflicts!

PR #3424, opened 2019-07-08
Improving documentation for geom_histogram().

### LINKED ISSUE #3269: Improved documentation for geom_histogram()

angelica93:
- Details about modifing binwidths and center/boundary
- Removed repeated word from stat-bin documentation

Closes #3269

### DIFF
--- a/R/stat-bin.r
+++ b/R/stat-bin.r
@@ -3,8 +3,8 @@
 #'   refers to the original x values in the data, before application of any
 #'   scale transformation. When specifying a function along with a grouping
 #'   structure, the function will be called once per group.
-#'   The default is to use `bins`
-#'   that cover the range of the data. You should always override
+#'   The default is to use the number of bins in `bins`,
+#'   covering the range of the data. You should always override
 #'   this value, exploring multiple widths to find the best to illustrate the
 #'   stories in your data.
 #'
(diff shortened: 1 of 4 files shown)


R/stat-bin.r
```diff
@@ -1,10 +1,10 @@
 #' @param binwidth The width of the bins. Can be specified as a numeric value
-#'   or as a function that calculates width from unscaled x. Here, "unscaled x" 
-#'   refers to the original x values in the data, before application of any 
+#'   or as a function that calculates width from unscaled x. Here, "unscaled x"
+#'   refers to the original x values in the data, before application of any
 #'   scale transformation. When specifying a function along with a grouping
-#'   structure, the function will be called once per group. 
+#'   structure, the function will be called once per group.
 #'   The default is to use `bins`
-#'   bins that cover the range of the data. You should always override
```
karawoo: I think this might have been intentional, i.e. "The default is to use `bins` [i.e. the `bins` parameter] number of bins". Maybe another way to word this would be "The default is to use the number of bins in `bins`, covering the range of the data"?

karawoo: I think this might have been intentional, i.e. "The default is to use `bins` [i.e. the `bins` parameter] number of bins". Maybe another way to word this would be "The default is to use the number of bins in `bins`, covering the range of the data"?

PR #3425, opened 2019-07-08
Ggplot_add.by added TidyDevDay

### LINKED ISSUE #2734: ggplot_add.by()
`ggplot_add()` lacks a method for "by" objects. Unfortunately, "by" objects do not have "list" in their class attribute, so `ggplot_add.default()` is hit instead of `ggplot_add.list()`. Given the prevalent use of `by()`, is it worth adding a method?

Maschette:
This should close #2734 

Ran local test and it seemed to work fine.


hadley:
Thanks @Maschette!

PR #3429, opened 2019-07-08
Fix typos that were causing the CMD check not to pass

paleolimbot:
These are a few typo fixes from the last few PRs that might cause automatic checks to fail.


hadley:
Thanks!!

Issue #3044, opened 2018-12-21
Refactor geom_contour()?


clauswilke:
As discussed on Twitter [here](https://twitter.com/thomasp85/status/1076149558281932801) and previously explained in a [blog post by @thomasp85](https://www.data-imaginist.com/2018/what-are-we-plotting-what-are-we-animating/) an argument can be made that `geom_contour()` should take a 2d density map as input and calculate its own contour lines without using a stat. I'm opening this issue so we can discuss pros and cons of various approaches.

I think there is precedent for this kind of refactoring with the refactoring of `geom_bar()`, `geom_histogram()`, and `geom_col()` some time back.

clauswilke:
One of my personal pet peeves is that `stat_density_2d()` returns a contour by default. I've always thought that is confusing.

``` r
library(ggplot2)

set.seed(4393)
dsmall <- diamonds[sample(nrow(diamonds), 1000), ]
d <- ggplot(dsmall, aes(x, y))

# works
d + geom_raster(aes(fill = stat(density)), stat = "density_2d", contour = FALSE)
```

![](https://i.imgur.com/ltiM2wI.png)

``` r

# doesn't work but should in my opinion
d + geom_raster(aes(fill = stat(density)), stat = "density_2d")
#> Don't know how to automatically pick scale for object of type function. Defaulting to continuous.
#> Error in is.finite(x): default method not implemented for type 'closure'
```

![](https://i.imgur.com/w3NoIVF.png)

<sup>Created on 2018-12-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
The argument for changing the default of the `contour` argument to `FALSE` in `stat_density_2d()` if we change the input data for `geom_contour()` is that then `stat_density_2d()` will work by default with the new `geom_contour()` and also with `geom_tile()` and `geom_raster()`. By contrast, if we leave `contour = TRUE` it won't work with either.

The argument against changing this default is that it will potentially break existing code.

One way around this problem may be to soft-deprecate `stat_density_2d()` and introduce a new `stat_2d_density()`. This will give users of the current `stat_density_2d()` time to adjust their code while introducing a new function that can be designed cleanly and without baggage.

clauswilke:
[...]

Issue #2702, opened 2018-06-15
Proposal: Warning when using density2d + coord_map


mwaldstein:
## Problem

stat_density_2d does not account for non-euclidean coordinates, such as when using Latitude / Longitude. As a result, when densities are incorrect for densities near the poles and 180 Longitude. 

*note* This may apply for any non-cartesian coordinates (e.g. polar), but I have not tested it.

## Options

 * **Swap in a spherical density function** (e.g. from the Density package) when using map coordinates. This feels like a big lift
 * **Block use of density with map coordinates** (similar to geom_raster) - There are enough cases (e.g. city-level data sets) where the grid approximation is OK that leaving the option available seems appropriate. A counter argument is that if you're using coord_map, you care about projection where densities will be off. The current behavior has been in place for long enough I'd worry about breaking workflows.
 * **Include a warning** - My proposed option, to make sure people are aware they are using an approximation likely to cause issues at larger scales.

I'm happy to do the work to implement the warning, but wanted confirmation before creating the pull request.

hadley:
This problem happens for basically any stat and a non-Cartesian coordinate system so I don’t think singling out one combination is the right approach.

mwaldstein:
Generally agree.

Part of the difficulty is that there are a lot of examples in the wild of using these functions for map heatmaps, particularly at the city level. A novice (read: me) with low understanding of how the densities are calculated would readily assume that since the functions work together, you can apply them to global data.

Another option could be to just add an explicit message to the documentation highlighting that by "2d" it means "Cartesian" 2d.

I was looking at the minimum change (geom_density_2d + coord_map) but a more thorough choice would be to mimic geom_hex and geom_raster and stop on any non-Cartesian for geom_density_2d.

mwaldstein:
One more data point in the discussion - blocking geom_density_2d on non-Cartesian coordinates would break many [ggmap](https://github.com/dkahle/ggmap) examples, which likes to show off density plots.

paleolimbot:
I think that this is a `StatContour` problem (used by `StatDensity2d`), which assumes evenly-spaced x and y values, but doesn't give any warnings if this is not the case. The warning in `GeomRaster$setup_data()` could probably be generalized:

https://github.com/tidyverse/ggplot2/blob/e2bdf85929603591387c2938625121b600f3e84d/R/geom-raster.r#L51-L80

[...]

PR #3439, opened 2019-07-13
Use the isoband package to calculate contour lines and bands

paleolimbot:
This is a PR to address #3044. While the desired behaviour (calculating contours at draw time that can be labeled) can't be done immediately because `Geom`s don't have access to non-position scales, a good start is to use @clauswilke's [isoband package](http://github.com/clauswilke/isoband) to calculate contour lines (as opposed to `grDevices::contourLines()`) in `stat_contour()`. The isoband package also has a function to return contour polygons, which are implemented in this PR as `stat_contour_filled()` and `geom_contour_filled()`. 

Additional refactoring steps were:

- The `complete` parameter is not used in the current calculation and hasn't been used for about 4 years (estimated from the git blame view), so I removed it.
- I added function parameters and documentation for `bins`, `binwidth` and `breaks`. These appeared in examples but were not function arguments and were not documented.

Things I am not sure about:

- The calculation of breaks in this way is different than the calculation of bins and breaks everywhere else in ggplot2, and I wonder if they should be updated in this PR. In particular, letting breaks be a function I think would be helpful.
- I'm not sure about how I implemented `geom_contour_filled()`. There is no `GeomContourFilled`, it just calls `layer()` with `stat = "contour_filled"` and `geom = "polygon"`.
- I'm not sure what the corner cases are for `isoband::iso(band|contour)`, but I think I should probably add more tests to make sure inappropriate/weird inputs are handled properly (such as for #2702).


clauswilke:
I had a quick look and didn't see anything that concerns me. I'll be traveling extensively over the next two weeks, starting today. Therefore, you probably shouldn't wait for a more careful review from me, unless there's anything you feel uncertain about.

DESCRIPTION
```diff
@@ -55,7 +55,8 @@ Suggests:
     sf (>= 0.7-3),
     svglite (>= 1.2.0.9001),
     testthat (>= 2.1.0),
-    vdiffr (>= 0.3.0)
+    vdiffr (>= 0.3.0),
+    isoband
```
hadley: Should be in alphabetical order

PR #3443, opened 2019-07-17
Change from `unit` to `margin` on line 133

Myfanwy:
Currently the documentation for `plot.margin` (line 133 of `theme.r`) reads: 

> "plot.margin | margin around entire plot (`unit` with the sizes of the top, right, bottom, and left margins)".

Although this suggests that `unit()` is required for the `plot.margin` specification, using it throws an error:
```
ggplot(mtcars) +
  geom_point(aes(x = disp, y = hp)) +
  theme(plot.margin = unit(50, "pt"))

**> Error in `[.unit`(theme$plot.margin, 2) : 
  index out of bounds ('unit' subsetting)**
```
...while this does not:

```
ggplot(mtcars) +
  geom_point(aes(x = disp, y = hp)) +
  theme(plot.margin = margin(50, 0, 0, 0))
```

### DIFF
--- a/R/theme.r
+++ b/R/theme.r
@@ -130,7 +130,7 @@
 #'   "top", "topright", "left", "right", "bottomleft", "bottom", "bottomright)
 #'   or a coordinate. If a string, extra space will be added to accommodate the
 #'   tag.
-#' @param plot.margin margin around entire plot (`unit` with the sizes of
+#' @param plot.margin margin around entire plot (`margin` with the sizes of
 #'   the top, right, bottom, and left margins)
 #'
 #' @param strip.background,strip.background.x,strip.background.y


clauswilke:
You misunderstood the current documentation. `plot.margin` takes a unit vector of length 4. Your current PR is not correct, but feel free to think about how the documentation could be improved so it's less likely to be misunderstood.

``` r
library(ggplot2)

ggplot(mtcars) +
  geom_point(aes(x = disp, y = hp)) +
  theme(plot.margin = unit(c(50, 0, 0, 0), "pt"))
```



<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Myfanwy:
Ah. I think I would have been less likely to misunderstand the documentation if the phrase "concatenated vector" were added, and/or if the `reexports` documentation page had an example of a call to `unit()` with a concatenated vector, as you have specified above (currently all the examples on the `theme` man page with a `unit` call use vectors of length 1).  I could create a new PR if you think I am not an isolated case?  I may be though :)

Myfanwy:
I also see that the reason that the call to `margin()` works is because `margin()` simply plugs the values in to `unit()` properly.

Myfanwy:
Closing this because it's not really a problem with ggplot2/theme.R (and likely isn't a problem in general)

PR #3471, opened 2019-08-05
Do not change order of positive and negative values in PositionStack

### LINKED ISSUE #3390: geom_area with negative y not working in ggplot2 3.2.0
In `ggplot2 3.1.1` the area under the curve handles negative values
```r
library(ggplot2)

set.seed(115)

ggplot(
    data.frame(
        x = 1:100,
        y = cumsum(rnorm(n = 100, mean = 0, sd = 3))
    ),
    aes(x, y)) +
    geom_area(fill = "pink", alpha = 0.5) +
    geom_path()
```
image

In `ggplot2 3.2.0` something goes wrong:
```r
library(ggplot2)

set.seed(115)

ggplot(
    data.frame(
        x = 1:100,
        y = cumsum(rnorm(n = 100, mean = 0, sd = 3))
    ),
    aes(x, y)) +
    geom_area(fill = "pink", alpha = 0.5) +
    geom_path()
```
image

thomasp85:
fixes #3390 

@karawoo I have checked that this fix does not cause regressions in #1552. Basically what this does is avoid positive and negative y-values to get reordered, not fiddle with the group-based reordering

### DIFF
--- a/R/position-stack.r
+++ b/R/position-stack.r
@@ -195,7 +195,7 @@ PositionStack <- ggproto("PositionStack", Position,
       )
     }
 
-    rbind(neg, pos)
+    rbind(neg, pos)[match(seq_len(nrow(data)), c(which(negative), which(!negative))),]
   }
 )
 
--- a/tests/testthat/test-position-stack.R
+++ b/tests/testthat/test-position-stack.R
@@ -21,11 +21,11 @@ test_that("negative and positive values are handled separately", {
   p <- ggplot(df, aes(x, y, fill = factor(g))) + geom_col()
   dat <- layer_data(p)
 
-  expect_equal(dat$ymin[dat$x == 1], c(-1, 0, 1))
-  expect_equal(dat$ymax[dat$x == 1], c(0, 1, 2))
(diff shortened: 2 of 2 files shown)


R/position-stack.r
```diff
@@ -195,7 +195,7 @@ PositionStack <- ggproto("PositionStack", Position,
       )
     }
 
-    rbind(neg, pos)
+    rbind(neg, pos)[match(seq_len(nrow(data)), c(which(negative), which(!negative))),]
```
clauswilke: I don't see immediately that this is always correct. Can there be problems with `NA`s? Or will `NA`s never reach this point?

thomasp85: An `NA` value in ymax is pretty nonsensical and would also have broken the previous code. I can put in some guarding if you’d like nonetheless

clauswilke: No strong opinion either way. Just make sure to look into why the CI builds fail.

thomasp85: The failures are due to the downstream changes in sf, which have been fixed in master but not yet merged into the `v3.2.1-rc` branch

Issue #1371, opened 2015-10-15
Use alpha for lines of geom_density(), geom_bar(), and so on


yutannihilation:
Hi, thanks for this great package!

I saw many people (including me) are confused at the behavior of `alpha` when used in Geoms which draw areas or rects (e.g. `geom_desity()`). So, let me ask if this is your design decision or not.

When I draw a chart with `alpha`, I expect that the whole layer, the line and fill, will become transparent. But the line stays clear red.

```r
ggplot(diamonds, aes(carat)) +
    geom_density(fill = "red", colour = "red", alpha = 0.1)
```
![plot3](https://cloud.githubusercontent.com/assets/1978793/10501293/cf987d82-731a-11e5-8ade-a8a462af878b.png)

I expect the one like this.

![plot1](https://cloud.githubusercontent.com/assets/1978793/10501292/cb700022-731a-11e5-8c30-db8ff67ff6d0.png)

I know we can draw the latter chart by using `colour=alpha("red", 0.1)`, but I believe `alpha` is a transparency param for **the entire layer**, not only for the line or the fill. For me, this is a matter of semantics. Are there any reason not to use `alpha` for `colour`?

The possible implementation is very simple. We just need to change those two lines to `col = alpha(aes$colour, aes$alpha)`:

https://github.com/hadley/ggplot2/blob/bbc1b5ad3c0f55e8809d4fe1a9bd2783eb9fa29c/R/geom-rect.r#L55
https://github.com/hadley/ggplot2/blob/bbc1b5ad3c0f55e8809d4fe1a9bd2783eb9fa29c/R/geom-ribbon.r#L95

If this looks good to you, I'm happy to submit PR :smile:

hadley:
It was a deliberate design decision. But now I can't remember why :frowning: I think I had a good reason though.

yutannihilation:
Hmm..., thanks for your quick answer, anyway.

I found the behavior of `alpha` have been this way from the time it was born, but no background is described in the commit message.

https://github.com/hadley/ggplot2/commit/36d5ad2e2a0a3d9df2c527ddfa7452ed030beb7d

Okay, I trust in you this time. You always have a good reason; the very fact can be a good reason for me to close this issue :+1:. Please close this issue anytime!

[...]

Issue #1523, opened 2016-01-27
breaking change: alpha now impacts the line of stat_ based geoms.


antoine-lizee:
This has changed following #1371, and I think (personally) that this is the wrong decision. 
Pros: it's slightly more 'natural' to have alpha impacting both the line and fill (?)
Cons: the line texture is modified by the alpha parameter in a way that is irreversible, and thus you need a different geom for the common use case (to me) of having a solid line with a transparent filling.
Here is an example:
```
ggplot(diamonds %>% filter(cut %in% c("Very Good", "Ideal")), aes(x = price, fill = cut)) + theme_bw() + 
  geom_histogram(binwidth = 1000) +
  geom_density(aes(x = price, y = ..count.. * 1000), color = "NA", alpha = 0.3) +
  geom_density(aes(x = price, y = ..count.. * 1000), color = "black", fill = NA) +
  facet_grid(cut ~ ., scales = "free_y", space = "free_y")
```
I would vote for reverting this behavior, as I'm pretty sure it's breaking more graphing code now than it would if reverted for 2.1.

The downside of the old behavior is that you need to change the color directly if you want something lighter.

antoine-lizee:
A solve-it-all solution would be to have two different `alpha` parameters?

joanwa:
yes, please either revert or two different alpha parameters. This change broke many of my graphs...

antoine-lizee:
Thanks!

yutannihilation:
I'm sorry for those who suffered with the change I proposed :sob:
Hope someday some great person will discover some great solution which has more easier `alpha` semantics both for beginners and those who already got used to ggplot2....

dhimmel:
+1 for separate parameters for `line_alpha` and `fill_alpha` (or just `alpha` as it is currently).

antoine-lizee:
To answer requests from @dhimmel and others that needs lines representing densities *with alpha*, `geom_density` is the wrong tool as it is designed to mainly show an area. To get a line, use directly `stat_density(geom = "line", alpha = 0.1, ...)`:
```
ggplot(diamonds %>% filter(cut %in% c("Very Good", "Ideal")), aes(x = price, fill = cut)) + theme_bw() + 
  geom_histogram(binwidth = 1000) +
  stat_density(aes(x = price, y = ..count.. * 1000), geom = "line", color = "black", alpha = 0.3) +
  facet_grid(cut ~ ., scales = "free_y", space = "free_y")
```

[...]

PR #3485, opened 2019-08-15
Allow select alpha to be applied on fill (default), colour, or both

atusy:
This PR is related to https://github.com/tidyverse/ggplot2/issues/1371

In the discussion in #1371, a change was attempted to apply alpha both on `colour` and `fill`.
In this PR, we can select alpha to be applied on `fill` (default), `colour`, or `both`.
The default selection is `fill` and thus is safe in terms of backward compatibility.

Currently, `alpha_to` is implemented for `geom_polygon`, `geom_rect`, and `geom_ribbon`.
It works in some more (e.g., `geom_density`), but not all (`geom_bar`).
So I need some more to work on.

Before I update other functions, I'd like to ask @hadley if you like my PR.

## Examples

Varying `alpha_to` by `fill`, `colour`, and `both`

``` r
library(ggplot2)
d <- data.frame(xmin = 0, xmax = 1, ymin = 0:4, ymax = 1:5)
g <- ggplot(d) +
  aes(
    xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax,
    colour = ymin, alpha = ymin
  )
g + geom_rect(size = 3) # alpha_to = "fill"
```

``` r
g + geom_rect(size = 3, alpha_to = "colour")
```

``` r
g + geom_rect(size = 3, alpha_to = "both")
```

<sup>Created on 2019-08-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/geom-polygon.r
+++ b/R/geom-polygon.r
@@ -87,7 +87,7 @@ geom_polygon <- function(mapping = NULL, data = NULL,
                          na.rm = FALSE,
                          show.legend = NA,
                          alpha_to = c("fill", "colour", "color", "both")) {
-  alpha_to <- standardise_aes_names(match.arg(alpha_to))
+  alpha_to <- standardize_alpha_to(match.arg(alpha_to))
   layer(
     data = data,
     mapping = mapping,
--- a/R/geom-polygon.r
+++ b/R/geom-polygon.r
@@ -15,6 +15,8 @@
 #' @export
 #' @inheritParams layer
 #' @inheritParams geom_point
+#' @param alpha_to
+#'  Whether to apply alpha to "fill" (default), "colour" ("color"), or "both".
 #' @param rule Either `"evenodd"` or `"winding"`. If polygons with holes are
 #' being drawn (using the `subgroup` aesthetic) this argument defines how the
 #' hole coordinates are interpreted. See the examples in [grid::pathGrob()] for
@@ -84,7 +86,8 @@ geom_polygon <- function(mapping = NULL, data = NULL,
                          ...,
                          na.rm = FALSE,
                          show.legend = NA,
-                         inherit.aes = TRUE) {
+                         alpha_to = c("fill", "colour", "color", "both")) {
+  alpha_to <- standardise_aes_names(match.arg(alpha_to))
   layer(
     data = data,
     mapping = mapping,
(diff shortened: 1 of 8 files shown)


clauswilke:
I feel certain this is the wrong API. It's better (and easier to implement!) to set up a hierarchy of aesthetics. This is done e.g. in ggridges, where `point_alpha` can be used to override the application of `alpha` specifically to points. 

``` r
library(ggplot2)
library(ggridges)

# without `point_alpha`
ggplot(iris, aes(x=Sepal.Length, y=Species)) +
  geom_density_ridges(
    jittered_points = TRUE,
    position = position_points_jitter(width = 0.05, height = 0),
    point_shape = '|', point_size = 3, alpha = 0.4,
  )
#> Picking joint bandwidth of 0.181
```



``` r

# with `point_alpha`
ggplot(iris, aes(x=Sepal.Length, y=Species)) +
  geom_density_ridges(
    jittered_points = TRUE,
    position = position_points_jitter(width = 0.05, height = 0),
    point_shape = '|', point_size = 3, point_alpha = 1, alpha = 0.4,
  )
#> Picking joint bandwidth of 0.181
```



<sup>Created on 2019-08-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Whether we want this in ggplot2 proper is a separate question though. I don't see a strong need. For all geoms provided by ggplot2 one can always achieve the same effect by using two separate geoms (e.g., `geom_area()` and `geom_line()`).

Another relevant discussion: #1523

atusy:
> better (and easier to implement!) to set up a hierarchy of aesthetics

Sound like you are right.
Your idea is much more flexible because alpha of color and fill can be based on different variables.

>  achieve the same effect by using two separate geoms

For `geom_density`, its true that `geom_area()` and `geom_line()` works.
What about for `geom_rect`?
Is there any `geom` which just draws border of rectangles?
Maybe `geom_path` is the one, but is not as easy as using `geom_rect`.

yutannihilation:
>  Maybe `geom_path` is the one, but is not as easy as using `geom_rect`.

But, it's still possible. I agree it would be nice to have some option to make alpha affect on lines, but I don't see a good reason to do so, considering the difficulty of making changes on these fundamental geoms and maintaining them. As we saw on #1371, I think it's too late to change....

atusy:
I see...

yutannihilation:
You can create an extension package, though.

PR #3495, opened 2019-08-21
Remove lazyeval

yutannihilation:
This PR removes lazyeval from Imports.

I didn't notice, but lazyeval package is no longer used since v3.1.0. More precisely, we only imports `f_eval()` from lazyeval but it was removed by this commit: https://github.com/tidyverse/ggplot2/commit/d9f2cd4b6848c8b86a78ae63cf2fd52d1e124f9a#diff-43ce5f73e89026d3e9ec337c66087370


thomasp85:
Weird - I was sure I removed it completely together with plyr... good catch

PR #3505, opened 2019-08-27
Added feature to separately color the middle line of crossbar and box…

capebulbs:
…plot

To realize the minimalist style of boxplot like the one inside
http://biochemres.com/beautiful-minimalist-boxplots-with-r-and-ggplot2
Instead of adding extra geom_segment over geom_boxplot, now the color of middle bar can be colored separately by setting middle.colo(u)r. And this feature can be turned off by setting separate.middle = FALSE, the middle bar will be colored with the outline of boxplot/crossbar by the col parameter, which acts as what ggplot2 does currently

### DIFF
--- a/R/geom-boxplot.r
+++ b/R/geom-boxplot.r
@@ -44,6 +44,10 @@
 #'   it only hides them, so the range calculated for the y-axis will be the
 #'   same with outliers shown and outliers hidden.
 #'
+#' @param separate.middle Controls whether to color the middle bar separately, 
+#'   default is `TRUE`, if `FALSE`, the middle bar is colored with the outline of the boxplot
+#' @param middle.colour, middle.color Set the color of middle bar separately, 
+#'   only effects when separate.middle is `TRUE`
 #' @param notch If `FALSE` (default) make a standard box plot. If
 #'   `TRUE`, make a notched box plot. Notches are used to compare groups;
 #'   if the notches of two boxes do not overlap, this suggests that the medians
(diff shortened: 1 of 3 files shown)


capebulbs:
sample code:
`# also supported in legends
ggplot(mpg, aes(displ, hwy, col = drv, fill = drv)) + geom_boxplot(middle.color = "white")
`


`# turn off this feature by setting separate.middle = FALSE
ggplot(mpg, aes(displ, hwy, col = drv)) + geom_boxplot(middle.color = "white", separate.middle = FALSE)`

capebulbs:
> Please take a look at this comment and the surrounding conversation:
> [#3485 (comment)](https://github.com/tidyverse/ggplot2/pull/3485#issuecomment-521651724)

I've read the #3485, and I think my pull request is not like it, I just want to make the plot look better, the three newly-added arguments are not mapped in `aes`.

yutannihilation:
I think a geom can have colour-related parameters outside of the hierarchy of aesthetics, as `GeomBoxplot` already has `outlier.colour`. But... I rather feel your `middle.colour` is not the case; it should be in the hierarchy, if it is to be implemented.

clauswilke:
In my opinion, moving color parameters outside of the hierarchy of aesthetics is almost never warranted. `outlier.colour` was added a long time ago, before we fully understood how to implement hierarchical aesthetics. Now that we know how it works and that it works fine, it's unclear that keeping it as a parameter only is the right choice.

yutannihilation:
Ah, agreed. So, do you mean this PR should consider moving `outlier.colour` into the hierarchy first, and then implement `middle.colour` in the hierarchy?

clauswilke:
That should at least be under consideration, yes.

capebulbs:
OK, I'll try to put it into the hierarchy of aesthetics, thank you for your suggestions!

clauswilke:
Closing this PR.

@capebulbs: If you still plan to implement this feature along the lines outlined in my comments, please open a new PR.

PR #3506, opened 2019-08-28
Removing direction constraint from geoms

thomasp85:
Some geoms have directionality constraint, e.g. `geom_histogram()` that requires binning along the x-axis. The gospel has been to use `coord_flip()` if one wanted binning along the y-axis, but that comes with a set of limitations. Further the constraint seems arbitrary from a user perspective. 

This PR explores the possibility of making relevant geoms work in either direction. Hopefully this can completely remove the need for ggstance.

#### Geoms, stats, and positions under consideration

- [x] `geom_bar`
- [x] `geom_histogram`
- [x] `geom_violin`
- [x] `geom_col`
- [x] `geom_linerange()`
- [x] `geom_pointrange()`
- [x] `geom_errorbar()`
- [x] `geom_crossbar()`
- [x] `geom_boxplot()`
- [x] `geom_density()`
- [x] `geom_area()`
- [x] `geom_ribbon()`
- [x] `geom_line()`
- [x] `geom_smooth()`
- [x] `stat_bin()`
- [x] `stat_boxplot()`
- [x] `stat_count()`
- [x] `stat_density()`
- [x] `stat_ydensity()` (consider renaming this for clarity)
- [x] `stat_summary()`
- [x] `position_dodge()`
- [x] `position_dodge2()`
- [x] `position_fill()`
- [x] `position_stack()`
- [x] `position_jitterdodge()`

Please add to this list if anything comes up. @clauswilke when we have settled on this it would be great if ggridges could be updated to reflect this as well


thomasp85:
One thing to consider is how to deal with ambiguity in directional geoms. `geom_col()` for instance could be used in situations where the data can be interpreted in either direction. Should we add a `direction` argument that default to `NA` (i.e. we try to guess) but can be used to overwrite guessing

thomasp85:
This should also deprecate `geom_errorbarh`

thomasp85:
This is very much WIP at the moment as I’m exploring what is possible. Don’t expect the current state of the code to prevail.

The overarching strategy is to let the stats (and geoms) sniff out the wanted orientation from the mapping of the aesthetics. Many are quite easy such as histograms where you just look at whether x or y is mapped. Others need inspection of the data types mapped to x and y to figure out which one is discrete. Others still are fully ambiguous and we’ll need to figure out what to do there.

The plan and hope is that it should “just work” by mapping the variables to the right aesthetics, but whether this is possible is open

clauswilke:
Auto-detection + optional manual override should work fine.

thomasp85:
yeah, that was my plan as well... don't know if we want manual override for all or just the ones where ambiguity exist though

thomasp85:
@clauswilke can you remember why you created the `GeomCol` class? It seems identical to the `GeomBar` class it subclasses...

thomasp85:
One annoying side effect of this is that many of these stats/geoms have to have their `required_aes` field removed as they can take two non-overlapping sets of aesthetics to work. Would be nice if this could somehow be specified so we don't loose the self-documenting feature of the field (along with the error handling)

thomasp85:
One way to address the problem of removing required aesthetics will be to add another field to geoms and stats (`required_aes_alt`, or something) that will be checked and handled along with the primary field... What say you @hadley ?

clauswilke:
I didn't create `GeomCol`, I'm just the last one who touched it. I looked at the code right now and compared it to `GeomBar` and I don't see any differences. Maybe they were different at some point and converged.

thomasp85:
Oh yeah. It was @hrbrmstr. Got it mixed up

hadley:
Maybe you could make all these `Geoms` inherit from (say) `GeomFlippable` and then it could provide custom required aesthetic logic? 

I don't see an obvious way to update the documentation generation, ideally we'd special case something where it would generate "*x* or *y*". 

Or maybe a better option would be to introduce some special sentinel value `required_aes = c("x|y")`?

mjskay:
Can I offer some unsolicited thoughts after seeing this on Twitter? :) I recently implemented something similar in tidybayes (the multiple-orientations-one geom part, not the part that chooses orientation automatically --- I love the idea of having that sniffed out automatically and am planning to add it now too, with an API consistent with whatever is decided here).

I came to a similar solution as @thomasp85 : have a parameter like `direction` (I called mine `orientation`) and then write the code to use either `x` or `y` based columns depending on that parameter. I did do things a little differently from there: I created a somewhat hackish function [define_orientation_variables](https://github.com/mjskay/tidybayes/blob/c903bf04d900fb2387acea322502ffc02b270b00/R/geom.R#L50) that I can call at the top of any function that needs to interact with the data. This allowed me to rewrite my geoms very easily; code that looked like this:

```r
draw_panel = function(self, data, panel_params, coord) {
  ...
  data$y = data$ymin + justification * data$height
  ...
}
```

Became:

```r
draw_panel = function(self, data, panel_params, coord, orientation) {
  define_orientation_variables(orientation)
  ...
  data[[y]] = data[[ymin]] + justification * data[[height]]
  ...
}
```

Which saved me a bunch of `switch`es and code duplication and other headaches when rewriting those geoms. Now I just write the geoms as if they are have a horizontal orientation in my mind and `define_orientation_variables` takes care of the vertical case. On the other hand, this is perhaps a very weird style of API that produces code that works for me internally in one package and might not be appropriate for a broadly-used API like ggplot, but I wonder if something similar in spirit could be used here to make building such geoms similarly straightforward.

A few other things:

- I like the idea of having a parameter to control this explicitly with default value `NA` meaning "automatic", as @thomasp85 suggested. FWIW as a single data point about discoverability, that's the interface I had assumed it would take after hearing about it on Twitter but before seeing the code.
- I think that "orientation" is a better word for this than "direction". "direction" connotes more... well, _direction_ to me, like "up" or "down", not "horizontal" or "vertical", which I think of as orientations. [This answer](https://english.stackexchange.com/a/85632) on English stack exchange seems to agree with my intuition. I think not using "direction" here would also be more consistent with the use of "direction" elsewhere in the ggplot API (e.g. the `direction` parameter in `scale_colour_brewer` controls the direction of the color mapping in the palette, not its orientation, which is meaningless).
- +1 to @hadley's suggestion of a GEOM to inherit from. Relatedly, it strikes me that detect_direction (or what have you) would also make more sense as a method instead of a function.

thomasp85:
Thanks for chiming in @mjskay... Some quick thoughts:

- I don't feel good about your `define_orientation_variables()` solution. It is way too magicking and I'll usually refrain from creating environment-changing side effects for the benefit of less typing. Some tooling around this is needed (maybe the current `switch_position()`, maybe something else), but it should not do *funny stuff*
- `orientation` sounds good to me
- I'm almost 100% certain that this cannot be abstracted away into a virtual `Geom` subclass that you can just subclass and everything will be fine. It might be doable with the `Stat` changes, but will require more effort than it is worth IMO

@Hadley the `"x|y"` nomenclature sounds good to me, but will require some thought in how it is implemented. What would be really nice would be to define two separate sets of aesthetics that needs to be defined, e.g. `x`, `ymin`, and `ymax` *or* `y`, `xmin`, and `max`, instead of `x` or `y`, `xmin` or `ymin`, and `xmax` or `ymax`

clauswilke:
Why not assign the orientation variables to a list and then use that, rather than messing with the parent frame?

``` r
  ov <- define_orientation_variables(orientation)
  ...
  data[[ov$y]] <- data[[ov$ymin]] + justification * data[[ov$height]]
```

mjskay:
@thomasp85 makes total sense. What about something along these lines to try to reduce boilerplate and switches:

```r
    params$orientation <- self$detect_orientation(data, params$orientation)
    data$orientation = params$orientation
    ...
  }
```

Then when you actually need to use any variables affected by orientation, something like:

```r
draw_panel = function(self, data, panel_params, coord) {
  o = self$get_orientation_variables(data$orientation[[1]])
  ...
  data[[o$y]] = data[[o$ymin]] + justification * data[[o$height]]
  ...
}
```

Then GeomFlippable would provide default implementations of `setup_data`, `detect_orientation`, and `get_orientation_variables`, and in 99% of cases making a flippable geom would just require using the `o = self$get_orientation_variables(data$orientation[[1]])` boilerplate in any function that uses flippable aesthetics (you wouldn't even need to add the boilerplate in `setup_data`, you would just call the GeomFlippable class's implementation.

thomasp85:
Maybe it is just me, but that doesn't seem easier than what is currently done...?

Almost every geom and stat that this affects has their own `setup_data()` method so I'm not convinced making a parent class for this makes any sense. There is potentially also differences in how the sniffing needs to be done for different stats/geoms...

thomasp85:
Another thing to consider is that this is not just about geoms... stats and positions need exactly the same tooling so I think it makes sense to offer than as functions available to all rather than as ggproto methods

mjskay:
I see your point about stats and geoms. I think the value of having a `switch_orientation` or `get_orientation_variables` or whatever as a method is that it is used in multiple places in the geom and it may need to be modified for different geoms. E.g. in your examples you aren't switching "width" and "height" aesthetics; in the geoms I am writing I need to. If a default `get_orientation_variables` defined everything except "width" and "height" it would be easy for me to simply override it, call up to the parent to get everything the default provides and add on the couple of additional aesthetics I need. Then if I (or someone else) subclasses that geom the method correctly sets the orientation-sensitive aesthetics. A similar argument could apply with a `detect_orientation` method when subclassing geoms.

If it is desirable to have both a function and method (method for extensibility, function to reduce code duplication), you could have a function which the default implementation of the corresponding methods in geoms, stats, and positions use.

W.r.t. simplicity, I quite like the simplicity of using `switch_position` on `names(data)` in your solution compared to a `get_orientation_variables` solution. On the other hand, I see more `switch`es and code duplication in your `setup_data` than I think should be necessary (and which is not necessary using something like `get_orientation_variables`). The other advantage of the latter is that it will not cause bugs if accidentally done twice, where I could anticipate some bugs due to accidentally double-using `switch_position` (or at least, you have to carefully keep track of where in a code sequence it has been applied).

thomasp85:
I see what you say - the switches are just there as an alternative to switch_position. It would probably make sense to align these implementations

thomasp85:
Regarding the need to switch width with height, there really is no need. There is only a requirement to switch the aesthetics that are recognised as positional aesthetics (unless you plan to provide both a height and width aesthetic to the user, but I would argue that is bad design)

clauswilke:
Regarding the switching of names: I think this is safe in cases where the renamed data frame is immediately consumed, as in the `draw_group()` function of a geom, e.g.:
https://github.com/tidyverse/ggplot2/blob/a13ae074e4a4ab19d2f4372a7a851344e8f02ab4/R/geom-boxplot.r#L202-L209

However, if the data frame with renamed columns were to be handed off to other parts of ggplot2, it'd a recipe for confusion. Not sure if that is the case anywhere in the code. I looked at the `compute_group()` function for `stat_boxplot()` and it switches the names and then switches them back, so that's safe also:
https://github.com/tidyverse/ggplot2/blob/a13ae074e4a4ab19d2f4372a7a851344e8f02ab4/R/stat-boxplot.r#L91-L131

yutannihilation:
> AFAIK, `geom_col()` should just be `StatIdentity` and `GeomBar`, right?

I don't fully remember, but I was stuck with the subtle difference about whether `width` can be mapped or not when I attempted https://github.com/tidyverse/ggplot2/pull/3194. Personally, now we have good reason to change the existing behaviour for consistency, but we should be aware this will be a (small) breaking change.

thomasp85:
@yutannihilation but the two implementations are completely identical, so I can't see how it could break anything to deprecate GeomCol?

thomasp85:
@clauswilke yes, the renaming have to be confined within each function. OTOH it is a way easier approach when retrofitting an existing geom as you don't need to change anything in the existing logic... I'll go through another cleanup of my current approach and then we can discuss from there

yutannihilation:
Sorry, you are right... I was confused about #3194. If we want to make `geom_col()` ignore `width` aesthetic, (IMHO) the implementation should differ from `geom_bar()` in that it would ignore `data$width`. But, yeah, we don't break anything if we remove `GeomCol` at the moment. Please forget my comment, sorry.

smouksassi:
@cpsievert how would this auto direction detection impact the auto plotly conversion plotly:ggplotly. ggplotly did not support ggstance and ggstance did not plug into plotly extension framework

mjskay:
> Regarding the need to switch width with height, there really is no need. There is only a requirement to switch the aesthetics that are recognised as positional aesthetics (unless you plan to provide both a height and width aesthetic to the user, but I would argue that is bad design)

Ah, I encountered this when trying to make a geom compatible with `ggstance::position_dodgev()`, which uses `height` instead of `width`. Will the replacement here (I assume something like `position_dodge(orientation = "vertical")` still use `width`?

cpsievert:
@smouksassi it'll likely break a decent amount of logic, but it'd also likely to simplify some of `ggplotly()`'s internals

thomasp85:
@mjskay I would not touch the name of that parameter... I think `width` is still descriptive even when the bars are tipped

thomasp85:
I have updated the code for a consistent approach. During the rewrite and experimentation I discovered an additional issue with assigning to the correct aesthetic names directly (as proposed by @mjskay) that is kind of a deal breaker. A lot of the underlying algorithms that gets called in within the different `setup_*`/`compute_*`/`draw_*` methods expects data in a x-major format and would thus require a rewrite as well, with increased complexity.

thomasp85:
Okay, all obvious candidates for this treatment has been taken care off, except one: `stat_summary()`... As such it isn't complicated but it exposes orientation in its argument names e.g. `fun.ymin`, and I'm not sure how we should deal with this gracefully... any ideas?

yutannihilation:
For stat_summary, I think it's okay to create a new, more general interface if possible, as long as we still can provide a backward-compatible behavior when the user specify such a direction-dependent argument. Since the existing names are direction-dependent, I guess the new names won't conflict.

thomasp85:
I've decided to include `geom_line()` and `geom_smooth()`... there is no way to guess the orientation from the input so these will rely solemnly on the user-specified orientation parameter

thomasp85:
@batpigandme can I get you to look at the wording in the *Orientation* section in `geom_bar()`? I feel it is a bit off but can't put my finger on it...

R/geom-bar.r
```diff
@@ -19,6 +19,16 @@
 #' [position_fill()] shows relative proportions at each `x` by stacking the bars
 #' and then standardising each bar to have the same height.
 #'
+#' @section Orientation:
+#' This geom treats each axis differently and can thus have two orientations.
```
batpigandme:
```suggestion
#' This geom treats each axis differently and, thus, can have two orientations.
```

R/geom-bar.r
```diff
@@ -19,6 +19,16 @@
...
+#' Often the orientation is easily deducable from a combination of the given
```
batpigandme:
```suggestion
#' Often the orientation is easy to deduce from a combination of the given
```

R/geom-bar.r
```diff
@@ -19,6 +19,16 @@
...
+#' mappings and the types of positional scales in use. Thus, ggplot2 will by
+#' default try to guess which orientation the layer should have. Under rare
+#' circumstances the orinetation is ambiguous and guessing may fail. In that
```
batpigandme:
```suggestion
#' circumstances, the orientation is ambiguous and guessing may fail. In that
```

R/geom-bar.r
```diff
@@ -19,6 +19,16 @@
...
+#' case the orientation can be given directly using the `orientation` parameter,
```
batpigandme:
```suggestion
#' case, the orientation can be specified directly using the `orientation` parameter,
```

R/geom-bar.r
```diff
@@ -19,6 +19,16 @@
...
+#' which can be either `"x"` or `"y"`. The value gives the axis that the geom
```
batpigandme: Checking that I understand this last sentence correctly: The `orientation` value (either `"x"` or `"y"`) should be the axis you want the geom to run along. 
Right?

thomasp85: yeah

thomasp85:
Ok, last piece of the puzzle: `stat_summary()`. We have decided to rename the arguments from `fun.ymin`, `fun.y`, and `fun.ymax` to `fun.min`, `fun`, and `fun.max`. But, what should we expect the output of `fun.data` to be? Currently nothing gets enforced, but should we expect the function to output correctly named columns or should we flip it automatically (e.g. expect it to behave as it should unflipped)?

The easiest is to expect the function to output data with `y`, `ymin`, and `ymax` columns even in the case of flipped axes but will this be weird?

thomasp85:
It was really a no-brainer as all of the supplied function returns data in x-orientation format so this is what we continue to support

hadley:
@thomasp85 can you give me a few pointers as to where to focus my effort?

thomasp85:
The main logic lies in `has_flipped_aes()` so checking the assumptions in there would be first priority. Also, documentation of the feature and whether it is understandable

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+#'
+#' `has_flipped_aes()` is used to sniff out the orientation of the layer from
+#' the data. It has a range of arguments that can be used to finetune the
+#' sniffing based on what the data should look like. `flip_data()` will switch
+#' the column names of the data so that it looks like x-oriented data.
+#' `flipped_names()` provides a named list of aesthetic names that corresponds
+#' to the orientation of the layer.
+#'
+#' @param data The layer data
+#' @param params The parameters of the `Stat`/`Geom`. Only the `orientation`
```
hadley: Why take a list then?

Second and subsequent lines should be indented by two spaces

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+#' parameter will be used.
+#' @param main_is_orthogonal If only `x` or `y` are present does they correspond
```
hadley: I think an example would help make this concrete

hadley: In fact, I think adding an example to all the parameters would help (i.e. just mention which existing geom the behaviour corresponds to)

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+#' is not flipped. If `NA` this check will be ignored.
+#' @param range_is_orthogonal If `xmin`/`xmax` or `ymin`/`ymax` is present do
+#' they correspond to the main orientation or reverse. If `NA` this check will
+#' be ignored.
+#' @param group_has_equal Is it expected that grouped data has either a single
+#' `x` or `y` value that will correspond to the orientation.
+#' @param ambiguous Is the layer ambiguous in its mapping by nature. If so, it
+#' will only be flipped if `params$orientation == "y"`
+#' @param main_is_continuous If there is a discrete and continuous axis, does
+#' the continuous one correspond to the main orientation.
```
hadley:
```suggestion
#' the continuous one correspond to the main orientation?
```

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+#' @export
+#' @keywords internal
+#' @name omnidirection
+#'
+has_flipped_aes <- function(data, params = list(), main_is_orthogonal = NA,
+                            range_is_orthogonal = NA, group_has_equal = FALSE,
+                            ambiguous = FALSE, main_is_continuous = FALSE) {
+  # Is orientation already encoded in data?
+  if (!is.null(data$flipped_aes)) {
+    return(data$flipped_aes[1])
```
hadley:
```suggestion
    return(data$flipped_aes[[1])
```

Should use `[[` to make it clear that you're extracting a single value

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+    return(("x" %in% names(data) || "x" %in% names(params)) == main_is_orthogonal)
+  }
+
+  has_x <- !is.null(data$x)
+  has_y <- !is.null(data$y)
+
+  # Does each group have a single x or y value
+  if (group_has_equal) {
+    if (has_x) {
+      x_groups <- vapply(split(data$x, data$group), function(x) length(unique(x)), integer(1))
```
hadley: Is this slow?

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+  if (length(x_aes_pos) > 0) {
+    aesthetics[x_aes_pos] <- y[x_aes[x_aes_pos]]
+  }
+  if (length(y_aes_pos) > 0) {
+    aesthetics[y_aes_pos] <- x[y_aes[y_aes_pos]]
+  }
+  aesthetics
+}
+
+#' Utilities for working with omnidirecitonal layers
```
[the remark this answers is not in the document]
thomasp85: good point

R/utilities.r
thomasp85: The main motivation is dual... It makes the call more concise, and it allows adding further controls without rewriting all geom/stat code

R/utilities.r
thomasp85: potentially, which is also why it is only called if requested, and even so only fairly low down in the  stack (the call would on most cases exit before reaching this state)

thomasp85: I've moved it down even lower for good measure

R/utilities.r
thomasp85: I've added a section with in depth description and pointer to where it is used. Let me know if this addresses your comment sufficiently

R/utilities.r
```diff
@@ -388,3 +400,179 @@ parse_safe <- function(text) {
...
+#' `has_flipped_aes()` is used to sniff out the orientation of the layer from
+#' the data. It has a range of arguments that can be used to finetune the
+#' sniffing based on what the data should look like. `flip_data()` will switch
+#' the column names of the data so that it looks like x-oriented data.
+#' `flipped_names()` provides a named list of aesthetic names that corresponds
+#' to the orientation of the layer.
+#'
+#' @param data The layer data
+#' @param params The parameters of the `Stat`/`Geom`. Only the `orientation`
+#' parameter will be used.
```
hadley: Style: these lines need to be indented by two spaces

R/utilities.r
hadley: I still think you need to mention in the documentation an example of a geom for each option.

thomasp85:
Haven’t I provided examples for all options in the section ?

PR #3511, opened 2019-08-30
Add stat_count_2d

thomasp85:
This is a PR for discussing if this functionality should be in ggplot2 (otherwise I'll just add it to ggforce).

This PR implements `stat_count_2d()`, an equivalent to to `stat_count()` but for counting values that co-occur on both the x and y axis. This is valuable, both when plotting two discrete variables against each other, but also when working with continuous data that has been discretised by rounding etc. Motivating example:

```r
ggplot(iris) + 
  stat_count_2d(aes(Petal.Length, Petal.Width, colour = Species))
```
image

Right now the stat will also reorder data so that large points are behind small points, but I'd like to have that as a position adjustment instead (sorta like a `position_stack()` in the z-direction)

what say you?


yutannihilation:
Is this different from `stat_sum()`?

``` r
library(ggplot2)

ggplot(iris) + 
  stat_sum(aes(Petal.Length, Petal.Width, colour = Species))
```



<sup>Created on 2019-09-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Ah, no. Probably not (except it also calculate a proportion in line with `stat_count`)

yutannihilation:
OK, I feel `stat_count_2d()` sounds better than `stat_sum()` (I periodically forget what this stat is for...). So, +1 for this PR.

thomasp85:
I obviously forgot about it too🙂

thomasp85:
As most functionality already exist in `stat_sum()` I'm closing. I don't think the added functionality warrants confusing the API more

jan-glx:
maybe instead it is possible to clear up the API a bit?
This is how I wish it was:
`stat_count` - as it is now
`stat_sum` - same as current `stat_count` but `weight` required and without default
`stat_count_2d` - same as current `stat_sum`
`stat_sum_2d` -  same as current `stat_sum` but `weight` required and without default
Not all wishes come true but maybe it's something we *can* do?

`stat_sum` could become smart and behave like `stat_count` if either `x` or `y` is given and as before if both are given. This might not break existing code (unless it relies on ggplot2 throwing an error).

In any case, your trouble (and mine (but at least I googled `stat_count_2d` before finishing the implementation)) would likely have been avoided by having documentation links to `stat_sum` in `stat_count` and `stat_bin_2d`...

PR #3519, opened 2019-09-06
Improve error message for geom_density() when using y

### LINKED ISSUE #3514: Fix error message for geom_density() when using y
Hi, 
The error message for geom_density() when using x and y comes up and says y is missing. My suggestion would be to change the error when x and y are present but stat is not identity to "y is present, did you mean to use stat = 'identity'" or something like that. Also adding an example that uses x and y to https://ggplot2.tidyverse.org/reference/geom_density.html would be great.  

``` r
x<-data.frame(value=seq(0,1,by=0.2), count=rnorm(6, mean = 10) )

library(ggplot2)
#> Warning: package 'ggplot2' was built under R version 3.6.1

ggplot(x)+geom_density(aes(x=value, y=count))
#> Error: geom_density requires the following missing aesthetics: y

ggplot(x)+geom_density(aes(x=value, y=count), stat = 'identity')
```

<sup>Created on 2019-09-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Close #3514.

This PR make `StatDensity` raise a consistent error with other stats like `StatCount`.

``` r
#> Loading ggplot2

library(ggplot2)

d <- base::data.frame(x = 1:3)

ggplot(d) +
  geom_density(aes(x, x))
#> Error: stat_density() must not be used with a y aesthetic.
```

### DIFF
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -66,6 +66,13 @@ StatDensity <- ggproto("StatDensity", Stat,
   required_aes = "x",
   default_aes = aes(y = stat(density), fill = NA, weight = NULL),
 
+  setup_params = function(data, params) {
+    if (!is.null(data$y)) {
+      stop("stat_density() must not be used with a y aesthetic.", call. = FALSE)
+    }
+    params
+  },
+
   compute_group = function(data, scales, bw = "nrd0", adjust = 1, kernel = "gaussian",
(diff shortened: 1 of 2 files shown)


thomasp85:
I'm going to close as #3506 makes it obsolete

PR #3529, opened 2019-09-11
Enable geom_ribbon() to draw separate lines for upper and lower intervals

### LINKED ISSUE #3503: new geom suggestion: separate lines for upper and lower intervals
I'm wondering if there's room for a new geom that can produce the following graph

```r
library(tidyverse)

# Example data
df <- tibble(y = c(0, 0, 0, 0, 0, 1, 1.5, 1.7, 1.8), 
             upper_y = y + 0.2, 
             lower_y = y - 0.2,
             year = -4:4)

p <- ggplot(df, aes(year, y)) + geom_line()

# geom_errorline would simplify this
p + 
  geom_line(aes(y = upper_y), linetype = 2) +
  geom_line(aes(y = lower_y), linetype = 2)

```

image

So instead of having to use `geom_line` twice, you could instead do

```r
ggplot(df, aes(year, y, ymax = upper_y, ymin = lower_y)) + geom_errorline()

```

and this would fit more neatly into the idea of mapping the upper and lower intervals to the `ymin` and `ymax` aesthetics. 

I imagine there's a way to achieve that same plot with `geom_ribbon` or `geom_area` but I haven't been able to do so so I'm guessing it wouldn't be straightforward even if there is. 

Here is a prototype I wrote a few years ago: [https://github.com/weiyangtham/econothemes](https://github.com/weiyangtham/econothemes)

Thanks!

yutannihilation:
Fix #3503

This PR changes `GeomRibbon` from a normal polygon to a polygon + two lines; while `GeomRibbon` is implemented as a polygon, considering that the nature of this geom is to indicate the upper and the lower range along X axis, I believe it's natural that `GeomRibbon` comes with two lines.

Though this is a breaking change, I expect there's very few users using `GeomRibbon` as a closed line. But, I'm not fully confident... If this expectation would fail, I'll withdraw this PR and create a new geom.

``` r
devtools::load_all("~/Documents/repo/ggplot2")
#> Loading ggplot2

df <- tibble::tibble(
  y = c(0, 0, 0, 0, 0, 1, 1.5, 1.7, 1.8), 
  upper_y = y + 0.2, 
  lower_y = y - 0.2,
  year = -4:4
)

# can divide the ribbon
df[4, "upper_y"] <- NA 
df[4, "lower_y"] <- NA

ggplot(df, aes(year, y)) + geom_line() + 
  geom_ribbon(aes(ymax = upper_y, ymin = lower_y),
              linetype = 2, fill = "red", alpha = 0.1, colour = "black")
```

<sup>Created on 2019-09-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/geom-ribbon.r
```diff
@@ -108,15 +108,29 @@ GeomRibbon <- ggproto("GeomRibbon", Geom,
...
       default.units = "native",
       gp = gpar(
         fill = alpha(aes$fill, aes$alpha),
+        col = NA
+      )
+    )
+
+    munched_lines <- munched
+    # increment the IDs of the lower line
+    munched_lines$id <- munched_lines$id <- rep(c(0, max(ids, na.rm = TRUE)), each = length(ids))
```
yutannihilation: Probably I typoed `+` as `<-` here...

yutannihilation:
Thanks! I'm wondering how this breaking change should be described on NEWS. I'll wait for Hadley's comment.

hadley:
Can you explain how this would break existing code? I think the main change is that it won't draw vertical lines on either end, right?

thomasp85:
It's visually breaking, not something that makes any old code throw an error... I think we had a couple of those for the 3.2.0 release as well. But the NEWS will get rewritten come release so don't worry to much about it right now

hadley:
Yeah, I wouldn't call this a breaking change then.

thomasp85:
It just occurred to me that this is related to and will influence request for only drawing upper line in `geom_area()`. As `geom_area()` subclasses `geom_ribbon()` it could/should be solved here as well. I suggest adding a parameter to `GeomRibbon` to control whether `upper` or `both` lines are drawn.

We should make sure that the correct line is drawn in geom_area if it dips or is fully below 0

yutannihilation:
> Yeah, I wouldn't call this a breaking change then.

Thanks, I got it.


> I suggest adding a parameter to `GeomRibbon` to control whether `upper` or `both` lines are drawn.

Thanks, I'll consider this. But, I didn't think about the impact on `GeomArea` carefully, let me check...

yutannihilation:
Hmm... after some thinking, now I'm not sure if this is OK for `geom_area()`. How do you feel?

``` r
devtools::load_all("~/Documents/repo/ggplot2")
#> Loading ggplot2

huron <- base::data.frame(year = 1875:1972, level = as.vector(LakeHuron))
h <- ggplot(huron, aes(year))
h + geom_area(aes(y = level), alpha = 0.5, colour = "black", size = 2)
```



<sup>Created on 2019-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

We probably need these 3 types of Geoms here.

* `GeomRibbon` with separate lines on the upper and the lower
* `GeomArea` with a closed outline (as-is)
* `GeomArea` with separate lines on the upper and the lower

thomasp85:
I think `geom_area()` should only have the upper line drawn. This is how most people envision it and is a common request. You can potentially make the behaviour a optional so the old full stroking is accessible.

I don't think anyone wants area plots with upper **and** lower

yutannihilation:
> I don't think anyone wants area plots with upper **and** lower

After some thinking, now I agree with you here.

I think there are some options to implement this. For example:

1. `upper_colour` and `lower_colour`, which overrides `colour` for the upper line and the lower line individually (c.f. https://github.com/tidyverse/ggplot2/pull/3485#issuecomment-521651724)
2. Add an option to choose how the line should be drawn (e.g. "both", "upper", "lower", or "polygon").

I'll use the option 2 because it might be good to provide an option to fall back to the previous behavour (polygon).

yutannihilation:
> We should make sure that the correct line is drawn in geom_area if it dips or is fully below 0

It seems I need to add some tweaks on `geom_area()`, though I'm yet to figure out why these differ.

``` r
#> Loading ggplot2

d <- base::data.frame(x = 1:10, y = 5 - 1:10)

p1 <- ggplot(d, aes(x, ymin = 0, ymax = y)) +
  geom_ribbon(alpha = 0.3, colour = "red", size = 3, outline.type = "upper")

p2 <- ggplot(d, aes(x, y)) +
  geom_area(alpha = 0.3, colour = "red", size = 3, outline.type = "upper")

patchwork::wrap_plots(p1, p2)
```



<sup>Created on 2019-10-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Probably, I need another version of `PositionStack` to complete this? I can flip `ymin` and `ymax` when `ymax` is negative in `GeomArea$draw_group()`, but I feel it should be done on Position's side.

thomasp85:
This is not the domain of positions IMO... Pretty sure the reason for what you are seeing is that `geom_area()` has a `pmin()`/`pmax()` call on the y range somewhere before plotting. Maybe we can just remove that?

yutannihilation:
> `geom_area()` has a `pmin()`/`pmax()` call on the y range somewhere before plotting

I too thought so at first, but `GeomArea` does nothing special at the drawing stage.

https://github.com/tidyverse/ggplot2/blob/40e8b6094e23cad19c8f06182af6730a963c205f/R/geom-ribbon.r#L170-L187

I think the call of `pmax()` and `pmin()` in question is done in `pos_stack()`, which is used in `PositionStack$compute_panel()`.

https://github.com/tidyverse/ggplot2/blob/40e8b6094e23cad19c8f06182af6730a963c205f/R/position-stack.r#L220-L221

thomasp85:
damn, you are probably right... that is inconvenient

Maybe position_stack should keep track of which values it flip and switch them back in the end

thomasp85:
I've created a PR that addresses the observed problems with geom_area

I think we should have a switch for the new behaviour, both in geom_ribbon and geom_area... there will be some breaking visual changes and giving people a way to opt out will be good. I think the new behaviour should be default, though

thomasp85:
@yutannihilation can you check that everything is as it should be after the merge of #3673? If so I think we can finish this off

R/geom-ribbon.r
```diff
@@ -130,23 +138,48 @@ GeomRibbon <- ggproto("GeomRibbon", Geom,
...
+    if (identical(outline.type, "legacy")) {
+      return(ggname("geom_ribbon", g_poly))
+    }
+
+    munched_lines <- munched
+    # increment the IDs of the lower line
+    munched_lines$id <- switch(outline.type,
+      both = munched_lines$id + rep(c(0, max(ids, na.rm = TRUE)), each = length(ids)),
+      upper = munched_lines$id + rep(c(0, NA), each = length(ids)),
+      abort(paste("inlvaid outline.type:", outline.type))
```
thomasp85:
```suggestion
      abort(glue("invalid outline.type: {outline.type}"))
```

yutannihilation:
Thanks, confirmed the problem is now fixed after #3673.

``` r
#> Loading ggplot2

d <- base::data.frame(x = 1:10, y = 5 - 1:10)

p1 <- ggplot(d, aes(x, ymin = 0, ymax = y)) +
  geom_ribbon(alpha = 0.3, colour = "red", size = 3, outline.type = "upper")

p2 <- ggplot(d, aes(x, y)) +
  geom_area(alpha = 0.3, colour = "red", size = 3, outline.type = "upper")

patchwork::wrap_plots(p1, p2)
```



<sup>Created on 2019-12-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


> I think we should have a switch for the new behaviour, both in geom_ribbon and geom_area... there will be some breaking visual changes and giving people a way to opt out will be good. I think the new behaviour should be default, though. Am I right in that the current implementation doesn't allow for old-style full stroking?

Sorry, I missed this comment. Yes, I agree with you and the users can specify `"legacy"` option for the old-style full stroking. I'll add a NEWS item to describe the change briefly. Do I need to add some examples for this?

https://github.com/tidyverse/ggplot2/blob/7c6d3d5f1809af2599f4d524b5107eb01d0844bc/R/geom-ribbon.r#L25-L27

yutannihilation:
I'm merging this. Thanks for reviewing multiple times!

PR #3534, opened 2019-09-19
Finer control over aesthetic evaluation

thomasp85:
This PR implements a parallel to `stat()`, called `mod()` that allows you to mark aesthetics for evaluation just before they are send to the draw method in the geom. At this point the data has been mapped so expressions inside `mod()` will work with mapped data.

The prime use case for this is for tying colour and fill values together:

```r
library(ggplot2)
library(colourspace) # for lighten()
p <- ggplot(mpg, aes(class, hwy))
p + geom_boxplot(aes(colour = class, fill = mod(lighten(colour, 0.4))))
```
image

As can be seen the legend needs some work, but I open now so we can discuss this feature


R/aes-calculated.r
```diff
@@ -54,6 +60,14 @@ is_calculated <- function(x) {
     stop("Unknown input:", class(x)[1])
   }
 }
+is_modifier <- function(x) {
+  if (!is.call(x)) return(FALSE)
```
hadley: Can you please rewrite to use the same branches as above? It's a pattern that make it clear that you're recursing over an expression tree

thomasp85: Sure... The reason I didn't was because `is_calculated()` seemed to only be this complex because it had to deal with the old dotted notation

thomasp85:
any good suggestions on name? `mapped()` or `modify_mapped()` springs to mind. The former is a bit more in line with `stat()`

clauswilke:
I like `mapped()`. I wouldn't use `modify_mapped()` because the mapping doesn't have to be modified. Sometimes it may be a good idea to just reuse a mapping, for example because the first one is quite complex. E.g., in the example below, being able to reuse the `colour` mapping without retyping everything might be nice. As an alternative to `mapped()`, you could also consider `remap()`.

``` r
library(ggplot2)

  geom_boxplot(aes(colour = ifelse(class %in% c("pickup", "suv", "minivan"), "large car", "small car"))) +
  scale_color_discrete(name = NULL)
```



<sup>Created on 2019-09-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
changed name to `mapped()` but if anything better comes along I'm not married to that name. Fixed legend so the updated aesthetics get reflected:

```r
  geom_boxplot(aes(colour = class, fill = mapped(alpha(colour, 0.4))))
```
image

yutannihilation:
Cool, this solves the frequent request like #3485.

Minor question, do we need to be tolerant for `color` here?

``` r
  geom_boxplot(aes(color = class, fill = mapped(alpha(color, alpha))), alpha = 0.4)
#> Error in as.character(col) %in% "0": object 'color' not found
```

thomasp85:
Yeah — I thought about that as well. I think we should

clauswilke:
That's an excellent point. Aesthetics names get standardized as described/implemented here:
https://github.com/tidyverse/ggplot2/blob/b5606624cbd3cc5bb09fe2e008d1fada82daf44e/R/aes.r#L150-L163

Ideally the same rules should apply inside `mapped()`. Maybe it's possible to apply `standardise_aes_names()` to all symbols inside the `mapped()` expression?

thomasp85:
I have implemented a way to recurse through aesthetic expressions ago substitute names to standard aesthetic nomenclature. We should consider this carefully though as I don't think many people are aware of the old-style support and may be surprised that their variables are getting renamed..?

thomasp85:
So... this PR has experienced a little feature creep 🙂

I've added a new control function as well called `stage()` which effectively lets you remap aesthetics between stat, geom, and mapped. This address recurring needs to decouple the output of stats from the mapping of geoms. The `stat()` and `mapped()` functions persists as shortcuts when multiple mappings are not needed.

API (continuing with the same example):
```r
ggplot(mpg, aes(x = stage(stat = class, geom = x + 0.5), y = hwy)) +
  geom_boxplot(aes(colour = class, fill = mapped(alpha(color, alpha))), alpha = 0.4)
```
image

thomasp85:
@lionel- can I get you to have a look at the meta-programming stuff and make sure I've not done something terribly stupid?

has2k1:
For the name, as this functionality is similar to `hjust` and `vjust` of `geom_text` how about `aes_adjust`.

R/aes-calculated.r
```diff
@@ -36,7 +73,12 @@ is_dotted_var <- function(x) {
 is_calculated_aes <- function(aesthetics) {
   vapply(aesthetics, is_calculated, logical(1), USE.NAMES = FALSE)
 }
-
+is_mapped_aes <- function(aesthetics) {
```
lionel-: These semantics feel a bit sloppy to me because they allow users to affect the interpretation of an expression by using `mapped()` or `stage()` deep in the call, even though that applies to the whole expression, not subparts. It seems it would be cleaner to require the qualifier to be at top-level.

Then you could check for a `mapped()` call at top level, and bind `mapped` to a function that fails with an error message in the function-level of the data mask.

R/aes-calculated.r
```diff
@@ -24,6 +26,41 @@
...
+#'   geom_boxplot(aes(colour = class, fill = mapped(alpha(colour, 0.4))))
+#'
+mapped <- function(x) {
+  x
+}
+#' @export
+stage <- function(stat = NULL, geom = NULL, mapped = NULL) {
+  stat
+}
+stage_geom <- function(stat = NULL, geom = NULL, mapped = NULL) {
```
lionel-: Nice pattern!

R/aes-calculated.r
lionel-: On a side note, I find it confusing to have `is_` predicate return length != 1 vectors. I think an `is_` function should always be safe to use in `if ()` (length 1, no missing value). In rlang I use the `are_` prefix for vectorised predicates (but this convention wasn't picked up anywhere else I think).

R/aes-calculated.r
```diff
@@ -54,6 +96,40 @@ is_calculated <- function(x) {
     stop("Unknown input:", class(x)[1])
   }
 }
+is_mapped <- function(x) {
+  if (is.atomic(x)) {
```
lionel-: I would merge all the `FALSE` branches into one here. I don't think the `is.pairlist()` branch is needed.

R/aes-calculated.r
thomasp85: The whole `is_mapped` and `is_mapped_aes` setup is borrowed completely from the way `stat()` calls are handled. @Hadley do you think there is reason to rewrite them both?

clauswilke:
On old-style support: I think `pch` at least is reasonably common in the wild. See e.g. [here](https://community.rstudio.com/t/adding-a-legend-to-an-existing-ggplot/24380/3) for an example I found within 2 minutes of googling. If we want to deprecated old-style names, we should probably do it in two stages, where we first just add a warning that informs people they're using old-style names and that support for this may go away in the future.

On the name of the function `stage()`: I don't find this name particularly intuitive. What is a "stage" in a geom? I also feel that a verb would be better than a noun. (I assume you mean "stage" as a noun, not a verb, here). I don't yet have the best proposal, but something like `map_at()` might work. I'll ponder some more.

thomasp85:
`stage` is indeed meant as a verb. We are staging multiple mapping of the same aesthetic

clauswilke:
> `stage` is indeed meant as a verb.

Ah, Ok.

thomasp85:
As for old-style name. I do t suggest we deprecate them. But we need to be careful about this as there is a difference in supporting them as argument names and substituting their use inside expressions. The later will mask any variable that coincide with an old-style name

R/geom-.r
```diff
@@ -123,6 +123,32 @@ Geom <- ggproto("Geom",
       data[names(missing_eval)] <- missing_eval
     }
 
+    if (length(modifiers) != 0) {
+      env <- new.env(parent = baseenv())
```
hadley: Since we already rely on rlang, I think using `env()` would make the code simpler here:

```
env <- env(baseenv(), mapped = mapped)
stage_mask <- env(emptyenv(), stage = stage_mapped)
```

R/aes-calculated.r
```diff
@@ -24,6 +26,41 @@
...
+#' @export
+#' @examples
+#' ggplot(mpg, aes(class, hwy)) +
+#'   geom_boxplot(aes(colour = class, fill = mapped(alpha(colour, 0.4))))
+#'
+mapped <- function(x) {
+  x
+}
+#' @export
+stage <- function(stat = NULL, geom = NULL, mapped = NULL) {
```
hadley: Where are these arguments documented? I think the docs should include a brief description of the order in which they are applied.

R/aes-calculated.r
```diff
@@ -24,6 +26,41 @@
...
+  stat
+}
+stage_geom <- function(stat = NULL, geom = NULL, mapped = NULL) {
+  geom
+}
+stage_mapped <- function(stat = NULL, geom = NULL, mapped = NULL) {
+  mapped
```
hadley: Is `NULL` the right default here? I would've thought you'd want `waiver()`, but maybe I'm forgetting where that's used. The related change to `check_nondata_cols()` feels a bit suspicious to me.

thomasp85: `waiver()` is not really used for aesthetics anywhere in the code and would require similar changes to `check_nondata_cols()` as well as additional places. `NULL` is the right call

NAMESPACE
```diff
@@ -387,6 +387,7 @@ export(layer_scales)
 export(layer_sf)
 export(lims)
 export(map_data)
+export(mapped)
```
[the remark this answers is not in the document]
thomasp85: I think `mapped()` (or the functionality it provides) will be used more than stage, so IMO it makes sense to provide a `stat()` parallel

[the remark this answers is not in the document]
clauswilke: I agree with Thomas. I'd like to have access to something like `mapped()`. `stage()` is sufficiently complicated that in most cases I would probably not bother using it. (Though it's good to have the option.)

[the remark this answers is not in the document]
hadley: If we're going to have a collection of these functions, I think we need to think about the name a bit more — I think a common prefix would be appropriate.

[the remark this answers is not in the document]
thomasp85: I think the ship has sailed for that since we already have `stat()`... Open for something other than `mapped()` but I don't think we can meaningfully introduce prefixes at this point.

R/aes-calculated.r
thomasp85: ok, so I'll provide simplified parsers for `mapped()` (or whatever it ends up being called) and `stage()`

NAMESPACE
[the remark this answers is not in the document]
hadley: In that case, I think the second argument to `stage()` can't be `geom`, since if we did add an explicit function for that case, we couldn't call it `geom()`.

I also wouldn't object to `stage_mapped()`, `stage_stat()`, etc. Although then I guess we'd have to soft deprecate `stage()`?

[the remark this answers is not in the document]
thomasp85: The current `stage()` function needs some rethinking. Currently the `stat` argument corresponds to default evaluation, and the `geom` argument corresponds to `stat()` evaluation which is counterintuitive. The problem is that `stat()` is intended to denote the mapping happens after stats have been computed whereas `stage()` refers to mapping of the next stage. Maybe the signature should be `stage(start, after_stat, after_scale)` with corresponding `after_stat()` and `after_scale()` functions (soft-deprecating `stat()`)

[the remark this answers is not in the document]
hadley: I rather like that idae

R/aes-calculated.r
thomasp85: Do we agree that something like 
```r
is_scaled <- function(x) {
  is.call(x) && identical(x[[1]], quote(after_scale))
}
```
should be sufficient?

hadley: Given that we use rlang, you can use `is_call(x, "after_scale")`

thomasp85: oh, that is handy

thomasp85: but doesn't work on quosures... @lionel- is that a bug? The documentation says it will match name on the res if `x` is a formula - shouldn't the same be applied to quosures?

thomasp85:
```r
t <- quo(test())
is_call(t)
#> [1] TRUE
is_call(t, "test")
#> [1] FALSE
```
```

Comment (thomasp85) [reply R/aes-calculated.r]: hmm... it doesn't work for formulas either
```r
t <- ~test()
is_call(t, "test")
#> [1] FALSE
```

lionel-: This is fixed in dev rlang, `is_call()` consistently returns `FALSE` on quosures, forcing you to special-case them as appropriate.

thomasp85: will the new slang be out soon? 🙂

lionel-: Next week.

thomasp85: great

thomasp85: Just tried with GitHub rlang - still doesn't work... Can you confirm the above example still fails?

lionel-: I get the expected results:

```r
t <- quo(test())

is_call(t)
#> [1] FALSE

is_call(t, "test")
#> [1] FALSE
```

thomasp85: Ah... I misunderstood... I thought that `FALSE` was an error and that it would be fixed with next release...

thomasp85: so... would this be the correct way to test for whether a quosure is of a specific call?
```r
is_call(quo_squash(t), 'test')
```
or are there any gotchas to that approach?

lionel-: You should almost never use `quo_squash()`, its name is meant to be scary ;). `quo_get_expr()` is the way to extract an expression.

- If you know you have a quosure, use `quo_is_call()` or `is_call(quo_get_expr())`.

- If you'd like to automatically unwrap quosures and formulas (this can be tricky and cause unexpected results), use `is_call(get_expr())`.

lionel-:
```r
get_expr(quote(foo()))
#> foo()

get_expr(~ foo())
#> foo()

get_expr(quo(foo()))
#> foo()
```

hadley:
Can you write a NEWS bullet please? I find them very helpful for giving me the PR context (which I've usually forgotten, and I'd prefer not to have to re-read the entire thread)

thomasp85:
news bullet added @Hadley

thomasp85:
can you approve if you trust me on the implementation 🙂

PR #3537, opened 2019-09-23
Implement GeomSf$handle_na()

### LINKED ISSUE #3536: Regression: geom_sf() is now too aggressive in removing rows
I just encountered this problem where `geom_sf()` removes polygons that are plotted without outline. This is a regression in current ggplot2 master. I assume it's due to #3491.

cc @yutannihilation

``` r
library(ggplot2)
library(sf)

# make outline of the world in robinson coordinates
lats <- c(90:-90, -90:90, 90)
longs <- c(rep(c(180, -180), each = 181), 180)

robin_outline <- 
  list(cbind(longs, lats)) %>%
  st_polygon() %>%
  st_sfc(crs = "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs") %>% 
  st_sf() %>%
  st_transform(crs = "+proj=robin +lat_0=0 +lon_0=0 +x0=0 +y0=0")

# plot with line
ggplot(robin_outline) +
  geom_sf(fill = "#56B4E950", color = "black")
```

``` r

# plot without line
ggplot(robin_outline) +
  geom_sf(fill = "#56B4E950", color = NA)
#> Warning: Removed 1 rows containing missing values (geom_sf).
```

<sup>Created on 2019-09-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Fix #3536

### DIFF
--- a/R/geom-sf.R
+++ b/R/geom-sf.R
@@ -96,9 +96,22 @@ GeomSf <- ggproto("GeomSf", Geom,
                     alpha = NA,
                     stroke = 0.5
                   ),
-
-  non_missing_aes = c("size", "shape", "colour"),
                   
+  handle_na = function(self, data, params) {
+    type <- sf_types[sf::st_geometry_type(data$geometry)]
+    
+    if (all(type == "point")) {
+      non_missing_aes <- c("size", "shape", "colour")
+    } else {
+      non_missing_aes <- character()
+    }
+    
+    remove_missing(data, params$na.rm,
+      c(self$required_aes, non_missing_aes),
+      snake_class(self)
+    )
+  },
+
   draw_panel = function(data, panel_params, coord, legend = NULL,
                         lineend = "butt", linejoin = "round", linemitre = 10) {
     if (!inherits(coord, "CoordSf")) {
--- a/tests/testthat/test-geom-sf.R
+++ b/tests/testthat/test-geom-sf.R
@@ -81,19 +81,36 @@ test_that("geom_sf() removes rows containing missing aes", {
     colour = c("red", NA)
   )
 
-  p <- ggplot(pts) + geom_sf()
+  p1 <- ggplot(pts) + geom_sf()
   expect_warning(
-    expect_identical(grob_xy_length(p + aes(size = size)), c(1L, 1L)),
+    expect_identical(grob_xy_length(p1 + aes(size = size)), c(1L, 1L)),
     "Removed 1 rows containing missing values"
   )
   expect_warning(
-    expect_identical(grob_xy_length(p + aes(shape = shape)), c(1L, 1L)),
+    expect_identical(grob_xy_length(p1 + aes(shape = shape)), c(1L, 1L)),
     "Removed 1 rows containing missing values"
   )
   # default colour scale maps a colour even to a NA, so identity scale is needed to see if NA is removed
   expect_warning(
-    expect_identical(grob_xy_length(p + aes(colour = colour) + scale_colour_identity()),
+    expect_identical(grob_xy_length(p1 + aes(colour = colour) + scale_colour_identity()),
                      c(1L, 1L)),
     "Removed 1 rows containing missing values"
   )
+
+  plgns <- sf::st_sf(
+    geometry = sf::st_sfc(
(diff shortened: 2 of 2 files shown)


yutannihilation:
No idea why builds on 3.3~3.5 fails... https://travis-ci.org/tidyverse/ggplot2/jobs/588242890#L4107-L4120

clauswilke:
It seems strange to me to only handle points and to only handle them when all rows contain exclusively points. I suspect there needs to be a separate treatment for sf types point, line, and other, and it'll probably require a reimplementation of `remove_missing()` for this specific case.

thomasp85:
I think the pragmatic approach would be to handle this in `sf_grob()` as we are already inspecting each rows geometry type there. @yutannihilation does that make sense to you?

yutannihilation:
Yes, it makes sense because it's only sf that can access to the individual geometry including the ones in `GEOMETRYCOLLECTION`. I was thinking how I can implement in the way Claus suggested above, but I couldn't see how to do make ggplot2 generate the same plot in the following two `ggplot()` codes...:

``` r
library(sf)
library(ggplot2)

pt <- st_point(1:2)

outer <- matrix(c(0,0,10,0,10,10,0,10,0,0),ncol=2, byrow=TRUE)
pl <- st_polygon(list(outer))

sf1 <- st_sf(geometry = st_sfc(pt, pl))
ggplot(sf1) + geom_sf(fill = "red", colour = NA)

sf2 <- st_sf(geometry = st_sfc(st_geometrycollection(list(pt, pl))))
ggplot(sf2) + geom_sf(fill = "red", colour = NA)
```

In my understanding, in both case, the point should be removed, but the polygon should not. However, it seems difficult to handle it on ggplot2's side.

thomasp85:
`sf_grob()` is a ggplot2 function, so you can do it there... before it passes the data on to `st_as_grob()`

yutannihilation:
Hmm, sorry, I confused it with `st_as_grob()`...

clauswilke:
I think that what all of this shows is that our current approach to generating sf grobs is broken. We are not handling nested geometries correctly. See the following reprex.

``` r
library(sf)
library(ggplot2)

pt <- st_point(1:2)

outer <- matrix(c(0,0,10,0,10,10,0,10,0,0),ncol=2, byrow=TRUE)
pl <- st_polygon(list(outer))

# point is shown
sf1 <- st_sf(geometry = st_sfc(pl, pt))
ggplot(sf1) + geom_sf()
```



``` r

# point is not shown, because GeomPolygon$default_aes doesn't contain a shape value
sf2 <- st_sf(geometry = st_sfc(st_geometrycollection(list(pl, pt))))
ggplot(sf2) + geom_sf()
```



``` r

# now it's shown (but notice size difference from first plot)
GeomPolygon$default_aes$shape <- 19
ggplot(sf2) + geom_sf()
```



<sup>Created on 2019-09-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think the options are to either implement our own version of `sf::st_as_grob()` or work with the sf package authors to change how graphical parameters are handed to that function and how missing values are handled. It may be easier to implement our own function.

thomasp85:
Well, I'm the author of the implementation in sf, but better support for GeometryCollection is orthogonal to this issue I think.

I don't see why we do not filter in `sf_grobs()` it is a natural place to do it as all necessary information is available there

clauswilke:
> I don't see why we do not filter in `sf_grobs()` it is a natural place to do it as all necessary information is available there

I think the point is this: If we want to remove, say, all points that have a color of `NA`, and to do it right, we'd have to recursively descend into GeometryCollections to make sure we catch all cases. At that point, we might as well reimplement all of `sf::st_as_grob()` because it'll be a comparable amount of code anyways.

thomasp85:
I disagree. GeometryCollections is really a niche case due to bad performance and poor support for styling so reimplementing the whole grob conversation for that is overkill. I’ll throw together a poc for what I have in mind tomorrow so I can show what I’m getting at

yutannihilation:
Sorry, I didn't follow the discussion. Thank you Thomas for #3546, I'm closing this PR in favor of yours.

Issue #3183, opened 2019-03-13
gplot2 3.1.0 on a fresh setup does not install `hexbin` package


pachamaltese:
gplot2 3.1.0 on a fresh setup does not install `hexbin` package

``` r
library(tidyverse)
library(modelr)

diamonds2 <- diamonds %>%
  filter(carat <= 2.5) %>%
  mutate(lprice = log2(price), lcarat = log2(carat))

ggplot(diamonds2, aes(lcarat, lprice)) +
  geom_hex(bins = 50)
#> Warning: Computation failed in `stat_binhex()`:
#> Package `hexbin` required for `stat_binhex`.
#> Please install and try again.
```

![](https://i.imgur.com/VeuS0Ki.png)

<sup>Created on 2019-03-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
This is on purpose. hexbin is listed as a "suggested" package, which means it's not required for ggplot2 to function but if you have it installed it unlocks additional features.

PR #3541, opened 2019-09-25
hexbinDependency

hvaret:
Hi,

I just came across an error message when trying to use `geom_hex()` while the hexbin package is not installed (already described in #3183).

Instead of installing manually hexbin and then let ggplot2 load the entire package using `try_require()` (as with the maps package in #3126), I slightly modified ggplot2 to only import the required functions, and here is a pull request if you think it could make things smoother.

Best regards


thomasp85:
The current design is on purpose—we do not want to have hard dependencies on packages that are only required for select functionality

Issue #3483, opened 2019-08-13
Sizing geom_sf point data with variable containing NAs


daniel-barnett:
Trying to plot point data with `geom_sf()` that has a mapping to a variable that contains NAs creates an error. This occurs both in the CRAN 3.2.1 and development versions. 

``` r
library(ggplot2)
packageVersion("ggplot2")
#> [1] '3.2.1.9000'

# From geom_sf() examples in docs
nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"), quiet = TRUE)
nc_3857 <- sf::st_transform(nc, "+init=epsg:3857")

nc_3857$mid <- sf::st_centroid(nc_3857$geometry)

ggplot(nc_3857) +
  geom_sf(aes(geometry = mid, size = AREA))
```

![](https://i.imgur.com/HQQavPm.png)

``` r

nc_3857$AREA[10] <- NA

ggplot(nc_3857) +
  geom_sf(aes(geometry = mid, size = AREA))
#> Error in checkNA("fontsize"): mixture of missing and non-missing values for fontsize
```

![](https://i.imgur.com/mfYiRYf.png)

``` r

sessionInfo()
#> R version 3.6.1 (2019-07-05)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 7 x64 (build 7601) Service Pack 1
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_New Zealand.1252 
#> [2] LC_CTYPE=English_New Zealand.1252   
#> [3] LC_MONETARY=English_New Zealand.1252
#> [4] LC_NUMERIC=C                        
#> [5] LC_TIME=English_New Zealand.1252    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] ggplot2_3.2.1.9000
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.2         pillar_1.4.2       compiler_3.6.1    
#>  [4] highr_0.8          class_7.3-15       tools_3.6.1       
#>  [7] digest_0.6.20      evaluate_0.14      tibble_2.1.3      
#> [10] gtable_0.3.0       pkgconfig_2.0.2    rlang_0.4.0       
#> [13] DBI_1.0.0          yaml_2.2.0         xfun_0.8          
[...]

PR #3546, opened 2019-10-01
Handle NA gracefully in geom_sf

### LINKED ISSUE #3536: Regression: geom_sf() is now too aggressive in removing rows

thomasp85:
Fixes #3536

This is an alternative implementation to the fix in #3537 that respects the different geometry types interpretation of NA. As the problem of NA's is unique to the vectorised constructors there is no need to handle geometrycollections in any special way as they will always be treated as scalars and grid accepts a single NA in `gpar()`

``` r
library(ggplot2)
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.1.3, PROJ 4.9.3

pt <- st_point(1:2)

outer <- matrix(c(0,0,10,0,10,10,0,10,0,0),ncol=2, byrow=TRUE)
pl <- st_polygon(list(outer))

# point is shown
sf1 <- st_sf(geometry = st_sfc(pl, pt))
ggplot(sf1) + geom_sf()
```

``` r

# point is still shown as expected
sf2 <- st_sf(geometry = st_sfc(st_geometrycollection(list(pl, pt))))
ggplot(sf2) + geom_sf()
```

``` r

# NA is handled fine for collections
ggplot(sf2) + geom_sf(colour = NA)
```

``` r

# as well as for vectorised geometries
sf3 <- st_sf(geometry = st_sfc(pt, st_point(3:4)), one_na = c(1, NA))
ggplot(sf3) + geom_sf(aes(size = one_na))
```

<sup>Created on 2019-10-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


R/utilities.r
```diff
@@ -107,6 +100,10 @@ remove_missing <- function(df, na.rm = FALSE, vars = names(df), name = "",
 
   df
 }
+detect_missing <- function(df, vars, finite = FALSE) {
+  vars <- intersect(vars, names(df))
+  !cases(df[, vars, drop = FALSE], if (finite) is_finite else is_complete)
```
clauswilke: Is this line correct? Where does `is_complete` come from?

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* `geom_sf()` now removes rows can't be plotted due to `NA` aesthetics (#3546, @thomasp85)
```
clauswilke: "removes rows *that* can't be plotted"

clauswilke:
Could you also verify that the example from #3483 still works?

R/utilities.r
thomasp85: It's a function defined in `utilitites.r`. `detect_missing()` is more or less lifted from `remove_missing()` so that `sf_grobs()` can use the same logic as `handle_na()`

thomasp85:
>Could you link to the original bug report in one of the commit messages: #3536
It is linked at the top of the PR. Why would you want it in a commit message (that will get squashed away during merge)?

>Could you also verify that the example from #3483 still works?
I've done that already (but forgot to mention). Do you want a repress of it?

R/geom-sf.R
```diff
@@ -133,16 +131,30 @@ default_aesthetics <- function(type) {
...
+    is_collection <- is_collection[!remove]
+  }
   defaults <- list(
     GeomPoint$default_aes,
     GeomLine$default_aes,
     modify_list(GeomPolygon$default_aes, list(fill = "grey90", colour = "grey35"))
   )
+  defaults[[4]] <- modify_list(
+    defaults[[3]],
+    rename(GeomPoint$default_aes, c("size" = "point_size", "fill" = "point_fill"))
```
yutannihilation:
```suggestion
    rename(GeomPoint$default_aes, c(size = "point_size", fill = "point_fill"))
```

yutannihilation:
One minor problem of this approach to move the process to outside of `handle_na()` is that `na.rm` no longer controls whether or not to warn, we need to choose always be quiet or always warn. I think the latter should be slightly better, but am not fully sure what should be done here.

```
> test_check("ggplot2")

── 1. Failure: geom_sf() removes rows containing missing aes (@test-geom-sf.R#85

`expect_identical(...)` did not produce any warnings.

── 2. Failure: geom_sf() removes rows containing missing aes (@test-geom-sf.R#89

`expect_identical(...)` did not produce any warnings.

── 3. Failure: geom_sf() removes rows containing missing aes (@test-geom-sf.R#94

`expect_identical(...)` did not produce any warnings.
```
(https://travis-ci.org/tidyverse/ggplot2/jobs/591938837#L1219-L1227)

PR #3555, opened 2019-10-07
Fix remove_missing() call in Stat$compute_layer()

### LINKED ISSUE #3548: boxplot computation fails when setting scale limits
Before 10fa0014d6be81a7e641b8c2c3b48e1df1b7e2f5, this code produced the following image:

```r
ggplot(PlantGrowth, aes(y = weight)) +
  geom_boxplot() +
  scale_y_continuous(limits = c(5, 7.5))
```

image

After 10fa0014d6be81a7e641b8c2c3b48e1df1b7e2f5, it now produces a blank plot with this warning:

```r
Warning message:
Computation failed in `stat_boxplot()`:
missing values and NaN's not allowed if 'na.rm' is FALSE
```

cc @thomasp85

thomasp85:
Fix #3548 

The old call to `remove_missing()` in `Stat$compute_layer()` did not factor in the new `x|y` notation allowed in `Stat$required_aes` for bidirectional layers. This caused missing data to not be removed correctly. This PR fixes the problem

### DIFF
--- a/R/stat-.r
+++ b/R/stat-.r
@@ -76,8 +76,15 @@ Stat <- ggproto("Stat",
       snake_class(self)
     )
 
+    # Make sure required_aes consists of the used set of aesthetics in case of
+    # "|" notation in self$required_aes
+    required_aes <- intersect(
+      names(data),
+      unlist(strsplit(self$required_aes, "|", fixed = TRUE))
+    )
+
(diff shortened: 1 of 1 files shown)


thomasp85:
This fixes the issue. The `StatBoxplot` code itself is correct

PR #3562, opened 2019-10-10
Add missing verb

mine-cetinkaya-rundel:
"makes it easy to specific colours" = "makes it easy to **choose** specific colours"


clauswilke:
Thanks!

Issue #3560, opened 2019-10-09
Limits with NA cause errors when there is no data


paleolimbot:
The following should create a blank plot without erroring, but does not after #3426. This should probably be fixed in the same PR as #3448.

``` r
library(ggplot2)
ggplot(data.frame(x = numeric(0)), aes(x)) + 
  geom_point() +
  scale_x_continuous(limits = c(0, NA))
#> Warning in rep(yes, length.out = len): 'x' is NULL so the result will be
#> NULL
#> Error in ans[ypos] <- rep(yes, length.out = len)[ypos]: replacement has length zero
```

The offending code is here (where `self$limits` is not `NULL` and `self$range$range` is `NULL`):

https://github.com/tidyverse/ggplot2/blob/10fa0014d6be81a7e641b8c2c3b48e1df1b7e2f5/R/scale-.r#L439-L452

I think that `Scale$get_limits()` needs another case for limits containing `NA` (or `Scale$is_emtpy()` should return `TRUE` for these limits). @thomasp85 any opinion on either of these solutions?

Issue #3558, opened 2019-10-07
Regression: labels and breaks are not the same length


smouksassi:
As reported in this discussion and reprex:

https://github.com/tidyverse/ggplot2/pull/3398#issuecomment-539164374

Opening a new issue to keep track of it.

paleolimbot:
Thanks @smouksassi  - this was identified in a number of reverse dependency failures as @thomasp85 mentioned. I'm using this as my reprex:

``` r
library(ggplot2)
ggplot(mpg, aes(cty, hwy)) +
  geom_point() +
  scale_x_continuous(
    breaks = c(0, 20, 40),
    labels = c("0", "20", "40")
  )
#> Error: Breaks and labels are different lengths
```

I changed how breaks were "censored" by the scales by accident. I think the solution is to move the censoring to `guide_train()` rather than have the scales do it, but there might be something easier. I'm happy to fix this, but it will take a few days.

Issue #3448, opened 2019-07-22
Funcational limits on non-position discrete scales refer to non-existent field `trans`


paleolimbot:
Thanks to #3426 we can now use functional limits on discrete position scales. However, when using them on non-position discrete scales, there is an error:

``` r
library(ggplot2)
ggplot(mpg, aes(displ, hwy, col = class)) +
  geom_point() +
  scale_colour_discrete(limits = rev)
#> Error in self$trans$transform(self$limits(self$trans$inverse(self$range$range))): attempt to apply non-function
```

The issue is that the default implementation of `Scale$get_limits()` refers to `self$trans`, which isn't defined for discrete scales. One way to fix this is to move the default implementation to `ScaleContinuous` and move the implementation in `ScaleDiscretePosition` that was added in #3426 to `ScaleDiscrete`.

PR #3426, opened 2019-07-08
Allowed reversing on discrete scales #3115

AnneLyng:
resolved #3115

PR #3566, opened 2019-10-11
Fix scale- and coord-related regressions

paleolimbot:
This PR fixes two regressions I introduced in #3398 and #3426. The first affected plots where labels and breaks were specified manually (#3558):

```r
library(ggplot2)
ggplot(mpg, aes(cty, hwy)) +
  geom_point() +
  scale_x_continuous(
    breaks = c(0, 20, 40),
    labels = c("0", "20", "40")
  )
#> Error: Breaks and labels are different lengths
```

This was because I censored breaks at the wrong place when I introduced `ViewScale`s. This is fixed in this PR.

The second involves plots where there is no data and NA limits (#3560). I think NA limits should be considered "empty" when there is no data, which is how I implemented it in this PR. I also made sure that functional limits work on all discrete scales (#3448), and that NA limits aren't replaced with data limits on discrete scales (I don't even know what that would look like but probably has induced more than one spurious error). Feel free to debate how I should implement either of these.

```r
library(ggplot2)
ggplot(data.frame(x = numeric(0)), aes(x)) + 
  geom_point() +
  scale_x_continuous(limits = c(0, NA))
#> Warning in rep(yes, length.out = len): 'x' is NULL so the result will be
#> NULL
#> Error in ans[ypos] <- rep(yes, length.out = len)[ypos]: replacement has length zero
```

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -444,10 +444,9 @@ Scale <- ggproto("Scale", NULL,
     if (is.null(self$limits)) {
       self$range$range
     } else if (is.function(self$limits)) {
-      # if limits is a function, it expects to work in data space
-      self$trans$transform(self$limits(self$trans$inverse(self$range$range)))
+      self$limits(self$range$range)
     } else {
-      ifelse(is.na(self$limits), self$range$range, self$limits)
+      self$limits
     }
   },
 
@@ -559,6 +558,22 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
 
   rescale = function(self, x, limits = self$get_limits(), range = limits) {
     self$rescaler(x, from = range)
+  },
+
+  get_limits = function(self) {
+    if (self$is_empty()) {
+      return(c(0, 1))
+    }
+
+    if (is.null(self$limits)) {
+      self$range$range
+    } else if (is.function(self$limits)) {
+      # if limits is a function, it expects to work in data space
+      self$trans$transform(self$limits(self$trans$inverse(self$range$range)))
+    } else {
+      # NA limits for a continuous scale mean replace with the min/max of data
+      ifelse(is.na(self$limits), self$range$range, self$limits)
+    }
   },
 
   dimension = function(self, expand = expansion(0, 0), limits = self$get_limits()) {
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -534,6 +534,12 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     self$range$train(x)
   },
 
+  is_empty = function(self) {
+    has_data <- !is.null(self$range$range)
+    has_limits <- is.function(self$limits) || (!is.null(self$limits) && all(is.finite(self$limits)))
+    !has_data && !has_limits
+  },
+
   transform = function(self, x) {
      new_x <- self$trans$transform(x)
      axis <- if ("x" %in% self$aesthetics) "x" else "y"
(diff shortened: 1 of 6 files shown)


R/scale-.r
```diff
@@ -534,6 +533,12 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     self$range$train(x)
   },
 
+  is_empty = function(self) {
+    has_data <- !is.null(self$range$range)
+    has_limits <- is.function(self$limits) || (!is.null(self$limits) && all(is.finite(self$limits)))
+    !has_data && !has_limits
+  },
+
```
thomasp85: Can this be added to the `Scale` class instead or will it break `ScaleDiscrete`?

R/scale-.r
```diff
@@ -555,6 +560,22 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
...
+      self$range$range
+    } else if (is.function(self$limits)) {
+      # if limits is a function, it expects to work in data space
+      self$trans$transform(self$limits(self$trans$inverse(self$range$range)))
+    } else {
+      # NA limits for a continuous scale mean replace with the min/max of data
+      ifelse(is.na(self$limits), self$range$range, self$limits)
+    }
+  },
+
```
thomasp85: Is this just the same as `Scale$get_limits()`?

paleolimbot: It's different...`Scale$get_limits()` doesn't do anything to `NA` limits, and doesn't apply the function to untransformed limits and then transform the result (because transforms only exist in continuous scales). I'm not sure what was happening with NA limits in discrete scales before.

thomasp85: Can you test if it breaks anything in ScaleDiscrete? I don’t like having such a small method overwriting for no reason

R/scale-.r
paleolimbot: Whoops...I think that adding this implementation of `is_empty()` to `Scale` will cause errors in `ScaleDiscrete`.

R/scale-.r
paleolimbot: The reason is that passing functions as limits for discrete scales errored before (#3448), because `Scale$trans` is not defined. I've added tests for limit handling for continuous scales, and there are no failing tests otherwise...would adding a similar suite of tests for discrete scales be a reasonable way to resolve this?

thomasp85:
I think I was just trying to be too clever about the implementation... this is prob the way to go

cpsievert:
After https://github.com/tidyverse/ggplot2/pull/3566/commits/e2e343fd84e4e94e0959a87ae341df86a699035f, it's now possible for `break_positions()` and `get_breaks()` to return `NA` values, is that intentional? It does raise an issue for some logic in plotly and shiny as both of them rely on these values.

```r
library(ggplot2)

df <- data_frame(
  time = factor(c("Lunch","Dinner"), levels = c("Lunch","Dinner")),
  total_bill = c(14.89, 17.23)
)

gg <- ggplot(data = df, aes(x = time, y = total_bill, group = 1)) +
  geom_line() +
  ylim(0, max(df$total_bill))

gg <- ggplot_build(gg)

gg$layout$panel_params[[1]][["y"]]$break_positions()
#> [1] 0.04545455 0.30926502 0.57307550
#> [4] 0.83688598         NA
```

thomasp85:
@cpsievert @paleolimbot what is the status on the last bit of discussion? It sounds like ggplot2 is fixed in the new approach - has plotly/shiny been updated to work with it?

cpsievert:
plotly/shiny haven't yet been updated but I'll have a closer look within the next 24 hours

cpsievert:
Oh, forgot to follow-up --- there's a workaround for this on `plotly@master` now

PR #3567, opened 2019-10-14
Add new position_nudgestack()

ThomasKnecht:
The position_nudgestack() is meant to plot stacked bar-plots between the x-axis tick marks. This is especially useful for time series for styling reasons. position_nudgestack() is a simple combination of the two positions: position_nudge() and position_stack().

I created this position since I needed to create time series plots where it was wished that the stacked bars are between monthly x-axis ticks (mainly because of style reasons). Furthermore, I saw that other people might also be interested in the combination of these two positions:

https://stackoverflow.com/questions/55730693/ggplot2-combine-positioning-commands-position-nudge-and-position-stack

https://stat.ethz.ch/pipermail/r-help/2008-October/178162.html


thomasp85:
I think such functionality is better placed in an extension package. You can try to submit a PR to one of the general purpose ones eg ggforce or ggalt, or create your own

Issue #3039, opened 2018-12-15
theme(x, y) versus theme(x) + theme(y)


ptoche:
I had always assumed that the following were equivalent. They are not. Is it a feature of theming that all the options must be set in ``theme()`` in one pass? 

If so, may I suggest adding an example like below to: https://ggplot2.tidyverse.org/reference/theme.html

``` r
library(reprex)

library(ggplot2)
ggplot(mtcars, aes(wt, mpg)) +
    geom_point() +
    theme_void() + 
    theme(axis.line.x = element_line(color = "blue"),
          axis.ticks.x = element_line(color = "red"),
          axis.ticks.length = unit(1, "cm")) 
```

![](https://i.imgur.com/MT2ctVw.png)

<sup>Created on 2018-12-15 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

``` r
library(reprex)
library(ggplot2)
ggplot(mtcars, aes(wt, mpg)) +
    geom_point() + 
    theme_void() + 
    theme(axis.line.x = element_line(color = "blue")) +
    theme(axis.ticks.x = element_line(color = "red")) +
    theme(axis.ticks.length = unit(1, "cm")) 
```

![](https://i.imgur.com/K86sagq.png)

<sup>Created on 2018-12-15 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

ptoche:
Perhaps the same issue: To display the axis line, I need to set ``axis.ticks.length``. Change ``unit(1, "cm")`` to ``unit(0, "cm")`` or drop that last line altogether and the axis line disappears. 

    ggplot(mtcars, aes(wt, mpg)) +
        geom_point() +
        theme_void() + 
        theme(axis.line.x = element_line(color = "blue"),
               axis.ticks.length = unit(1, "cm"))

clauswilke:
This is a bug with `theme_void()`. It declares itself to be a complete theme but misses elements. This keeps coming up in various ways and the correct way to fix it is to define all theme elements in `theme_void()`. See e.g. #2723.

``` r
library(ggplot2)
p1 <- ggplot(mtcars, aes(wt, mpg)) +
  geom_point() +
  theme_classic() + 
  theme(axis.line.x = element_line(color = "blue"),
        axis.ticks.x = element_line(color = "red"),
        axis.ticks.length = unit(1, "cm")) 

p2 <- ggplot(mtcars, aes(wt, mpg)) +
[...]

PR #3570, opened 2019-10-14
Clean up theme addition

clauswilke:
Here is an attempt at fixing #3039. Recall: That issue shows that adding theme elements piecewise behaves differently from adding them all at once. The problem is that when adding a theme to a plot object, the current implementation pulls missing info from the default theme, but this is not the case when adding a theme to a theme. In my mind, this behavior is inconsistent. The addition should behave the same in both cases. The solution is to delete the plot-specific update function.

This is the reprex from #3039, which now works just fine:
``` r
library(ggplot2)

p <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
p1 <- p + theme_light() + 
  theme(axis.line.x = element_line(color = "blue")) +
  theme(axis.ticks.x = element_line(color = "red"))

p2 <- p + theme_light() +
  theme(axis.line.x = element_line(color = "blue"),
        axis.ticks.x = element_line(color = "red"))

identical(p1$theme, p2$theme)
#> [1] TRUE

p1
```

``` r
p2
```

<sup>Created on 2019-10-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup> 

However, this is potentially a breaking change (and in fact, some unit tests currently break), because partially updating root elements of a theme doesn't work anymore:
(**Update:** This has now been fixed, see below.)
``` r
library(ggplot2)

# this now breaks but works in current ggplot2 by magically pulling info from default theme
  geom_point() +
  theme(text = element_text(color = "red"))
#> Error in FUN(X[[i]], ...): Theme element 'text' has NULL property: family, face, size, hjust, vjust, angle, lineheight, margin, debug

# this still works
  geom_point() +
  theme(axis.title = element_text(color = "red"))
```

<sup>Created on 2019-10-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think the new behavior is correct. If we're messing with a root element, we should have to specify it completely. However, if that's considered to be bad, we could also fix the problem at this point in the code:
https://github.com/tidyverse/ggplot2/blob/115c3960d0fd068f1ca4cfe4650c0e0474aabba5/R/theme.r#L558-L565
by pulling values from the ggplot2 default theme where available instead of issuing an error.

### DIFF
--- a/R/theme.r
+++ b/R/theme.r
@@ -519,8 +519,8 @@ calc_element <- function(element, theme, verbose = FALSE) {
       return(el_out) # no null properties, return element as is
     }
 
-    # if we have null properties, try to fill in from theme_gray()
-    el_out <- combine_elements(el_out, theme_gray()[[element]])
+    # if we have null properties, try to fill in from theme_grey()
+    el_out <- combine_elements(el_out, ggplot_global$theme_grey[[element]])
     nullprops <- vapply(el_out, is.null, logical(1))
     if (!any(nullprops)) {
       return(el_out) # no null properties remaining, return element
--- a/R/theme.r
+++ b/R/theme.r
@@ -440,26 +440,13 @@ add_theme <- function(t1, t2, t2name) {
       call. = FALSE)
   }
 
-  # If t2 is a complete theme, just replace everything with the new theme
-  if (is_theme_complete(t2))
+  # If t2 is a complete theme or t1 is NULL, just return t2
+  if (is_theme_complete(t2) || is.null(t1))
     return(t2)
 
   # Iterate over the elements that are to be updated
   for (item in names(t2)) {
-    x <- t1[[item]]
-    y <- t2[[item]]
-
-    if (is.null(x) || inherits(x, "element_blank")) {
-      # If x is NULL or element_blank, then just assign it y
-      x <- y
-    } else if (is.null(y) || is.character(y) || is.numeric(y) || is.unit(y) ||
-               is.logical(y) || inherits(y, "element_blank")) {
-      # If y is NULL, or a string or numeric vector, or is element_blank, just replace x
-      x <- y
-    } else {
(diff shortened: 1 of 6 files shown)


clauswilke:
@hadley @thomasp85 Could you provide feedback on this? Should I proceed along these lines? And should I implement the workaround for `NULL` values in root elements, as explained at the end?

clauswilke:
After some more thinking, I decided that filling in root elements from the default theme is probably the right way to go. The required changes to make this happen were much more involved than I expected, but I think overall the code is moving in the right direction and getting cleaner. There's now only one theme addition function, and it consists of only a few lines of code. Also, all the unit tests work again (two made incorrect assumptions under the new model and required fixing), including all visual tests without modification. The following reprex works again also.

``` r
library(ggplot2)

  geom_point() +
  theme(text = element_text(color = "red"))
```



<sup>Created on 2019-10-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/theme.r
```diff
@@ -557,23 +511,31 @@ calc_element <- function(element, theme, verbose = FALSE) {
...
-        paste(names(nullprops)[nullprops], collapse = ", "))
+    nullprops <- vapply(el_out, is.null, logical(1))
+    if (!any(nullprops)) {
+      return(el_out) # no null properties, return element as is
     }
 
-    if (verbose) message("nothing (top level)")
-    return(theme[[element]])
+    # if we have null properties, try to fill in from theme_gray()
+    el_out <- combine_elements(el_out, theme_gray()[[element]])
```
thomasp85: Should we cache this in ggplot_globals? Building up `theme_gray()` for each element (or those that needs it at least) seems quite resource-needy

clauswilke: Yes, good point, will do.

Issue #3574, opened 2019-10-18
report a bug in ScaleBinned$get_breaks()


microly:
hi, @thomasp85 
Firstly, thank you for your hard works to improve ggplot2!

I think there is a small bug in ScaleBinned$get_breaks(), which is in:
https://github.com/tidyverse/ggplot2/blob/115c3960d0fd068f1ca4cfe4650c0e0474aabba5/R/scale-.r#L991-L993

As I known, end users has no way to set the value of ScaleBinned$n_bins, which has a NULL value by default.
So, when the breaks is a function provided by users, the function's second parameter will always set to be NULL. 
As a consequence, users cannot get meaningful breaks by set breaks as a function in the scales, such as scale_colour_stepsn.  

This is a simple example:
```
df <- data.frame(
    x = runif(100),
    y = runif(100),
    z = rnorm(100)
)

brk <- function(limits, n_bins) {
    if (is.null(n_bins))
        stop("n_bins is always set to be NULL, users cannot implement useful function for breaks.")
    else cat("users can set the value of n_bins, and get meaningful breaks using the values of limits and n_bins.")
}

library(ggplot2)

ggplot(df, aes(x, y)) +
    geom_point(aes(colour = z)) +
    scale_colour_stepsn(colours = terrain.colors(10), breaks = brk)

# Error in f(...) : 
# n_bins is always set to be NULL, users cannot implement useful function for breaks. 
```

I will  opening a pull request for this bug, please consider it, thanks!

PR #3575, opened 2019-10-18
fix a bug in ScaleBinned$get_breaks()

microly:
this pull request try to fix the bug in issue #3574 . 

I set n.breaks as the second parameter of breaks function.
if the n.breaks is not set by users, it will set to the default value 5, which is the same default as trans objects.

n.breaks can be set by end users.
So, after this fix, users can get meaningful breaks by set breaks as a function and get meaningful breaks using the values of limits and n.breaks.

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -989,8 +989,15 @@ ScaleBinned <- ggproto("ScaleBinned", Scale,
         self$limits <- self$trans$transform(limits)
       }
     } else if (is.function(self$breaks)) {
-      n.breaks <- self$n.breaks %||% 5 # same default as trans objects
-      breaks <- self$breaks(limits, n.breaks)
+      if ("n.breaks" %in% names(formals(environment(self$breaks)$f))) {
+        n.breaks <- self$n.breaks %||% 5 # same default as trans objects
+        breaks <- self$breaks(limits, n.breaks)
+      } else {
+        if (!is.null(self$n.breaks)) {
+          warning("Ignoring n.breaks. Use a breaks function that supports setting number of breaks", call. = FALSE)
+        }
+        breaks <- self$breaks(limits)
+      }      
     } else {
       breaks <- self$breaks
     }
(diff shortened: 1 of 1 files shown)


thomasp85:
I think we'd need to check if the breaks function support two arguments before calling it with `n.breaks`, as the assumptions up until now has been that a breaks function only take a single argument

microly:
I think the breaks function with two arguments can work well.
The code block below is a simple example.

The brk function can be also used in other scales whose breaks function only need a single argument, because n.breaks has a default value.

```
df <- data.frame(
    x = runif(100),
    y = runif(100),
    z = rnorm(100)
)

brk <- function(limits, n.breaks = NULL) {
    n.breaks <- n.breaks %||% 5
    breaks <- seq(limits[1], limits[2], length.out = n.breaks + 2)
    breaks <- breaks[-c(1, length(breaks))]
}

library(ggplot2)

ggplot(df, aes(x, y)) +
    geom_point(aes(colour = z)) +
    scale_colour_stepsn(colours = terrain.colors(10), breaks = brk)

ggplot(df, aes(x, y)) + geom_point() + scale_x_continuous(breaks = brk)
```

thomasp85:
What I mean is that your change will mean that if you pass in a breaks function that only accepts a single argument it will fail. This runs counter to the expectations of the breaks function

microly:
oh, I know what you mean now~

maybe, we can handle it the same way as:
https://github.com/tidyverse/ggplot2/blob/115c3960d0fd068f1ca4cfe4650c0e0474aabba5/R/scale-.r#L955-L956

thomasp85:
yeah, something like that

microly:
OK, thanks for your reply!
I learn very much from your codes and packages, thanks! 

How can I finish the work? should I open a new pull request?

clauswilke:
@microly Just make additional changes to the code, commit, push, and the PR will update. Don't open a new PR.

microly:
@clauswilke 
Thanks for your reply.
I am mot familiar with github, but I will try my best!

By the way, cowplot is one of the most important package in my daily work, thanks for your great 
 work!

microly:
@thomasp85 
@clauswilke    
I have made an update, please check it, thanks!

thomasp85:
Thanks!

microly:
thanks for your work!

PR #3579, opened 2019-10-19
Add value matching to breaks in manual_scale

### LINKED ISSUE #2429: Order of colors in case variable is 'character'
Hi!

I am using ggplot2 quite often, and I ran into a problem concerning the ordering of colors. So I am setting the breaks and the color values in `scale_fill_manual()`, and I would expect that the first color "black" would be assigned to the first break ('small') and so on. This seems to happen when the grouping variable is data type = character. I suppose you transform the variable into a factor internally, or something like that. 
Is this the intended behaviour, because for me this is really confusing?!

```r
test <- data.frame(a=as.character(gl(n = 4, length = 8, k = 2, labels = letters[1:4])), 
                   b=as.character(gl(n = 2, length = 8, k = 1, labels = c("small",  "big"))),
                   values = sample(10:300, 8))


pl1 <- ggplot(test, aes(x=a, y=values, fill=b))
pl1 + geom_bar(stat="identity", position="dodge") + 
scale_fill_manual("test", breaks=c("small", "big") , values = c("black","tomato2"))

```

Thank you!!
Manuel

idno0001:
Fixes #2429.

As noted in the issue, currently `values` can be matched with a named vector. The below examples show (1) this as a baseline, (2) how behaviour changes as `breaks` are added, and the new behaviour (3) when `values` is not a named vector.

``` r
library(ggplot2)

df <- data.frame(x = c("data_red", "data_black"))

p <- ggplot(df, aes(x, 1, fill = x)) + 
  geom_col()

# Named values, no breaks
p + scale_fill_manual(values = c("data_red" = "red", "data_black" = "black"))
```

``` r
# Named values with breaks
p + scale_fill_manual(values = c("data_red" = "red", "data_black" = "black"),
                      breaks = c("data_red", "data_black"))
```

``` r
# New behaviour: values matched to breaks
p + scale_fill_manual(values = c("red", "black"),
                      breaks = c("data_red", "data_black"))
```

<sup>Created on 2019-10-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


clauswilke:
Just to be sure this actually works as it should, could you provide a reprex where the data values are not identical to the color names? E.g., use this as your data:
``` r
df <- data.frame(x = c("data_red", "data_black"))
```

idno0001:
That's a good point; I've updated the reprex as you suggested.

clauswilke:
I've looked over the patch. It looks good to me, but we should get a second opinion from somebody else on the team.

A few additional items we will need:
- A news item that describes the new behavior. Importantly, this can change the visual output for some plots where people have relied on the previous, unordered behavior. (Not sure how to assess how big of a problem this will be.)
- A unit test for the new behavior. Probably something similar to this one:
https://github.com/tidyverse/ggplot2/blob/115c3960d0fd068f1ca4cfe4650c0e0474aabba5/tests/testthat/test-scale-manual.r#L3-L7

clauswilke:
Oh, also, put "Fixes #2429." into your next commit, so this is properly recorded in the git log.

idno0001:
I've added the news item and a couple of unit tests. Thanks for the feedback again!

NEWS.md
```diff
@@ -84,6 +84,12 @@
 
 * Increase the default `nbin` of `guide_colourbar()` to place the ticks more precisely (#3508, @yutannihilation).
 
+* Modified `manual_scale` to match `values` with the order of `breaks` whenever
```
clauswilke: Please write "`manual_scale()` now matches `values` with ..." Note the parentheses at the end of `manual_scale()`.

NEWS.md
```diff
@@ -84,6 +84,12 @@
...
+  `values` is an unnamed vector. Previously, unnamed `values` would be matched
+  to the limits of the scale and ignore the order of any `breaks` provided. Note
```
clauswilke: I think it should be "match with" instead of "match to".

tests/testthat/test-scale-manual.r
```diff
@@ -67,3 +67,17 @@ test_that("generic scale can be used in place of aesthetic-specific scales", {
 
   expect_equal(layer_data(p1), layer_data(p2))
 })
+
+test_that("named values do not match with breaks in manual scales", {
+   s <- scale_fill_manual(values = c("data_red" = "red", "data_black" = "black"),
+                          breaks = c("data_black", "data_red"))
+   s$train(c("data_black", "data_red"))
```
clauswilke: Please use tidyverse indenting:
``` r
s <- scale_fill_manual(
  values = c("data_red" = "red", "data_black" = "black"),
  breaks = c("data_black", "data_red")
)
```
(Also in the next test.)

NEWS.md
```diff
@@ -84,6 +84,12 @@
...
+  that this may change the appearance of plots that previously relied on the
+  unordered behaviour. (#2429, @idno0001)
```
clauswilke: Please move the period to the very end of the line: "unordered behaviour (#2429, @idno0001)."

idno0001:
All done. Give me a shout if there's anything else.

NEWS.md
```diff
@@ -84,6 +84,12 @@
 
 * Increase the default `nbin` of `guide_colourbar()` to place the ticks more precisely (#3508, @yutannihilation).
 
+* Modified `manual_scale()` to match `values` with the order of `breaks`
```
clauswilke: Sorry, one last nitpick: I think it's better to write "`manual_scale()` now matches ..." to emphasize the new behavior rather than the fact the code was modified.

R/scale-manual.r
```diff
@@ -124,6 +124,18 @@ manual_scale <- function(aesthetic, values = NULL, ...) {
     force(values)
   }
 
+  # order values according to breaks
+  args <- list(...)
+  if (is.vector(values) && is.null(names(values))
+      && "breaks" %in% names(args)) {
+    if (length(args[["breaks"]]) != length(values)) {
```
thomasp85: I'm not particularly fond of this setup. It will force all dots to get evaluated which may have unintended consequences. Is there a reason why `breaks` was not added as a formal argument?

clauswilke: I missed this. Good point. I agree.

@idno0001 You should add a `breaks` argument with default `waiver()` (just like here: https://github.com/tidyverse/ggplot2/blob/5e388e143f3d24bbeb6ce6b892cdb91deb5af6a2/R/scale-.r#L158-L159) and then explicitly hand this over to `discrete_scale()`. And the correct way to check whether breaks are set or not is then to test with `is.waive()`, I think. https://github.com/tidyverse/ggplot2/blob/40e8b6094e23cad19c8f06182af6730a963c205f/R/utilities.r#L180

idno0001: Thanks for the comments and apologies for the delayed response, I've made the suggested changes. Initially I didn't want to make such 'big' changes, but I see how it makes sense to add `breaks` as an argument.

thomasp85:
Thank you for your contribution!

PR #3580, opened 2019-10-21
the default value of ScaleBinned$n.breaks should be 4, not 5.

microly:
The default value of ScaleBinned$n.breaks was set to the same as trans objects.
The default value of trans objects can map data to 5 colours in plot and legend, which is a direct-viewing color display.

However, when n.breaks is set to 5, ScaleBinned will map the data to **6 colours**!
6 colours display is not so good as 5 colours. 

This occur in the codes:
https://github.com/tidyverse/ggplot2/blob/115c3960d0fd068f1ca4cfe4650c0e0474aabba5/R/scale-.r#L914
and
https://github.com/tidyverse/ggplot2/blob/115c3960d0fd068f1ca4cfe4650c0e0474aabba5/R/scale-.r#L928

So I think the default value of ScaleBinned$n.breaks should be 4, not 5, then it can works the same way as trans objects.

@thomasp85 , please consider it, thanks!

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -961,7 +961,7 @@ ScaleBinned <- ggproto("ScaleBinned", Scale,
           breaks <- self$trans$breaks(limits)
         }
       } else {
-        n.breaks <- self$n.breaks %||% 5 # same default as trans objects
+        n.breaks <- self$n.breaks %||% 4 # same default as trans objects
         breaks <- seq(limits[1], limits[2], length.out = n.breaks + 2)
         breaks <- breaks[-c(1, length(breaks))]
       }


clauswilke:
Could you provide a reprex that demonstrates the problem with the current code base?

microly:
@clauswilke 
Sure!

I have not used reprex before.
I will learn it and provide a demonstration soon.

microly:
@clauswilke
I try to use the reprex package.
However, because of the government regulation of Internet, I cannot connect to "imgur.com" from the mainland of China.
So I have to supply graphic examples by hand.

batpigandme:
Ah @microly, the GFW issue with reprex/imgur is unfortunate, and known. 
https://github.com/tidyverse/reprex/issues/220#issuecomment-436334660
In lieu of running an actual reprex, could you give the full code for your example so one of us could run it locally?

microly:
Here are the examples:

By now, The default value of ScaleBinned$n.breaks is 5.

Scale_fill_fermenter (which based on ScaleBinned) has six colours in the legend, which is different from scale_fill_distiller (which based on ScaleContinuous).

After setting ScaleBinned$n.breaks to 4, it works the same way like scale_fill_distiller, both of which have 4 colours in their legends.

```
library(ggplot2)
library(tibble)

ggplot(tibble(x = 1:10, y = x, z = x), aes(x, y, fill = z)) +
    geom_point(shape = 21, size = 10) +
    scale_fill_fermenter(trans = "identity", nice.breaks = F,
                         n.breaks = 5,
                         guide = "legend")
```
p1
```
ggplot(tibble(x = 1:10, y = x, z = x), aes(x, y, fill = z)) +
    geom_point(shape = 21, size = 10) +
    scale_fill_fermenter(trans = "identity", nice.breaks = F,
                         n.breaks = 4,
                         guide = "legend")
```
p2
```
ggplot(tibble(x = 1:10, y = x, z = x), aes(x, y, fill = z)) +
    geom_point(shape = 21, size = 10) +
    scale_fill_distiller(trans = "identity",
                         guide = "legend")
```
p3

microly:
@batpigandme                
Thanks for your kind reply!
I past the example codes above. 
I think it can be saw now.

thomasp85:
We are pending an update to the scales package before we can actually pass `n.breaks` forward. The current values are not expected to make it to the release, but we will wait to change it until it has an effect

microly:
OK, thanks for your reply~

thomasp85:
I am closing this. The default of 5 makes sense as you'd generally want a break in the middle of your range instead of having a bin that spans over the middle. You shouldn't use the legend guide with a binned scale as it does not communicate the actually scale. Use the 'bins' guide instead

Issue #3576, opened 2019-10-18
Incorrect tick positions of the secondary axis


Ilia-Kosenkov:
While building a derivative of `facet_wrap`, I stacked several plots, produced from `mtcars` data set, and noticed, that, with `sec.axis` set, top axis ticks of the bottom plot do not **exactly** align with the bottom ticks of the top plot. (See figures below).

![1](https://user-images.githubusercontent.com/8782986/67085951-7c140f80-f1a8-11e9-8045-e6b1eebd1ef4.png)

And a zoomed-in version.

![2](https://user-images.githubusercontent.com/8782986/67085952-7f0f0000-f1a8-11e9-9523-14a0ab64722d.png)

The issue does not seem to be related to any of my modifications and can be easily reproduced in base `ggplot2`. 
In the reprex I build a simple plot with a secondary x axis, then check grobs of each axis and retrieve the `x` coord of `polyLine` grob. The difference, while small, is clearly seen. 

``` r

library(ggplot2, quietly = TRUE, warn.conflicts = FALSE)
library(dplyr, quietly = TRUE, warn.conflicts = FALSE)
library(grid, quietly = TRUE, warn.conflicts = FALSE)


# Simple plot with secondary, duplicated, x-axis
ggplot(mtcars, aes(hp, mpg)) +
    geom_point() +
    scale_x_continuous(sec.axis = dup_axis()) -> plt

# Converting to gtable to get each grob
plt %>% ggplot_build %>% ggplot_gtable -> tbl
tbl$layout %>% mutate(rowid = 1:n()) -> layout
# Indices of grobs
top_ind <- layout %>% filter(name == "axis-t") %>% pull(rowid)
bot_ind <- layout %>% filter(name == "axis-b") %>% pull(rowid)

# Axis grob is nested, so go deep until we hit the `polyLine` and its `x` coordinates
tbl$grobs[[top_ind]]$children[[2]]$grobs[[2]]$x -> top_tick_pos
# For another axis, slightly different order
tbl$grobs[[bot_ind]]$children[[2]]$grobs[[1]]$x -> bot_tick_pos
# Compare
print(top_tick_pos)
#> [1] 0.199199199199199native 0.199199199199199native 0.52052052052052native 
#> [4] 0.52052052052052native  0.841841841841842native 0.841841841841842native
print(bot_tick_pos)
#> [1] 0.199646643109541native 0.199646643109541native 0.520880179890781native
[...]

PR #3581, opened 2019-10-23
Better secondary axis tick positioning

Ilia-Kosenkov:
This PR suggests a solution that addresses issue #3576.
When plotting the secondary axis, tick positions of the secondary should be transformed to the range of the primary axis (`AxisSecondary$break_info()`). To achieve this, a grid (`old_range`) is generated (of the size `AxisSecondary$detail`, defaulted to `1000`). This grid is then transformed to the secondary axis scale (named `full_range`). Each tick of the secondary axis is transformed to the primary's scale by finding the `index` of the closest value in the `full_range` (`index <- which.min(abs(full_range - x))`) and adopting `old_range[index]` as tick position in the primary axis scale. 

It is easy to understand that if `full_range` does not contain values equal to the secondary axis ticks, transformation is not accurate and the quality of this transformation depends on the grid size. This is observed in #3576, when duplicated axis (`dup_axis`) cannot properly position its ticks.

A simple alternative solution is to introduce linear interpolation. Instead of using the closest grid value, it is better to use two grid values and interpolate in between to increase the quality of the transformation. This works exceptionally well in the case of linear transformations of the secondary axis, including duplicated axes (`dup_axis`).

The method works well for different combinations of primary/secondary transformations, including when one axis is monotonically increasing while another is decreasing. There are no checks for boundary cases, i.e. when ticks are found at the edges of the grid, which I assume never happens.

### DIFF
--- a/R/axis-secondary.R
+++ b/R/axis-secondary.R
@@ -201,12 +201,32 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
       range_info <- temp_scale$break_info()
 
       # Map the break values back to their correct position on the primary scale
-      old_val <- lapply(range_info$major_source, function(x) which.min(abs(full_range - x)))
-      old_val <- old_range[unlist(old_val)]
+      old_val <- sapply(range_info$major_source, function(x) {
+                which_less <- full_range < x
+                index_lower <- which(which_less & (abs(full_range - x) == min(abs(full_range[which_less] - x))))
+
+                which_greater <- full_range > x
+                index_upper <- which(which_greater & abs(full_range - x) == min(abs(full_range[which_greater] - x)))
+
+                index <- c(index_lower, index_upper)
+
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
Hi @paleolimbot. 
I came up with such a solution while looking through the `ggplot2`'s internal code related to #3576.
Now that you know the source of this problem, you can probably find something more reliable & appropriate than this simple interpolation. 

I find secondary axes extremely useful and believe that `ggplot2` will benefit from better secondary ticks.

Also, as far as I understand, the test data/template plots were generated with poorly positioned secondary ticks, so I am not sure how to properly test my changes. It seems that all secondary axis-related tests have failed.
It can be useful to implement a test similar to what I showed in #3576 - i.e. duplicating an axis and comparing tick positions (in units) on both sides.

PR #3585, opened 2019-10-25
Explicitly define all theme elements

### LINKED ISSUE #3584: Missing elements in theme definitions
All themes miss some elements in their definitions (i.e., elements are implicitly set to `NULL` rather than explicitly set to `NULL`). I think all themes provided by ggplot2 should explicitly define all elements. This will become more important if we want to allow extensible themes (#2540, #2784), because we'll have to distinguish between elements a theme knows about but wants to be `NULL` and elements a theme doesn't know about that need to be filled in from the theme defaults. Missing element definitions also have led to unexpected problems in the past, e.g. #2724.

Elements that I notice are missing:
- In all themes: some of `panel.grid`, `panel.grid.major`, `panel.grid.minor`, `panel.grid.major.x`, `panel.grid.minor.x`, `panel.grid.major.y`, `panel.grid.minor.y`
- In `theme_void()`: Too many to list, including parts of `axis.*`, `legend.*`, `panel.*`, `strip.*`, `plot.background`.

As an aside, we should recommend as a general practice to external theme developers that they always start with an existing ggplot2 theme and then just modify the parts they want to change, just like we do in most cases:
https://github.com/tidyverse/ggplot2/blob/6f5ffea70e80ce3aa076fb9da8a6b0ec7ebfbc51/R/theme-defaults.r#L247-L267

clauswilke:
This PR closes #3584 by making sure every theme explicitly defines every possible theme element. The strategy is to simply set every element that hasn't been defined explicitly to `NULL`. This can be done programmatically in a few lines of code.

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -236,7 +236,7 @@ theme_grey <- function(base_size = 11, base_family = "",
   )
 
   # make sure all elements are set to NULL if not explicitly defined
-  theme_all_null() %+replace% t
+  ggplot_global$theme_all_null %+replace% t
 }
 #' @export
 #' @rdname ggtheme
@@ -513,7 +513,7 @@ theme_void <- function(base_size = 11, base_family = "",
   )
 
   # make sure all elements are set to NULL if not explicitly defined
-  theme_all_null() %+replace% t
+  ggplot_global$theme_all_null %+replace% t
 }
 
 
@@ -647,7 +647,7 @@ theme_test <- function(base_size = 11, base_family = "",
   )
 
   # make sure all elements are set to NULL if not explicitly defined
-  theme_all_null() %+replace% t
+  ggplot_global$theme_all_null %+replace% t
 }
 
 theme_all_null <- function() {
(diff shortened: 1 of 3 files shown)


R/theme-defaults.r
```diff
@@ -639,4 +645,19 @@ theme_test <- function(base_size = 11, base_family = "",
...
+  # set all elements in the element tree to NULL
+  elements <- sapply(
+    names(ggplot_global$element_tree),
+    function(x) NULL,
+    simplify = FALSE, USE.NAMES = TRUE
+  )
+
+  args <- c(elements, list(complete = TRUE))
+  do.call(theme, args)
 }
```
thomasp85: Could we cache this theme as well?

clauswilke: Sure, will do that tomorrow.

clauswilke: Done. I'm also wondering whether this function should be exported, so external package developers can use the same trick, or whether we should rather encourage them to always start with one of the usable ggplot2 themes.

thomasp85: I think it is better to guide them towards modifying an existing theme... They can always base it off `theme_void()` if they want a clean slate

clauswilke: Sounds good. In any case, I think we need a vignette or other document to provide best practices recommendations for themes. I'll probably write something like that when I'm done with the theme reworking.

clauswilke: I did a quick check. It looks like all the most popular theme packages already always start from an existing theme. In fact, I couldn't find any example that didn't do this.

PR #3591, opened 2019-10-27
Change minor_breaks interface

### LINKED ISSUE #3583: More flexibility for minor_breaks
Should be able to:

* specify number of breaks between major tick marks (e.g. 10 minor breaks per major break in a log10 scale)
* specify size of break (e.g. a break every 0.2 units)

Given that `minor_breaks` can already be a function, it might just be a matter of writing a couple of little helper functions.

hadley:
To pass in both limits and major breaks. This will make it much easier to create useful families of minor break functions in scales.

Fixes #3583

@thomasp85 — if you think this approach is reasonable I'll work on a NEWS bullet. I think documentation will need to happen in a separate PR because all of `breaks`, `labels`, and `minor_breaks` need revision to point to the reader towards the scales package.

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -24,7 +24,8 @@
 #'   - `waiver()` for the default breaks (one minor break between
 #'     each major break)
 #'   - A numeric vector of positions
-#'   - A function that given the limits returns a vector of minor breaks.
+#'   - A function passed the limits and major breaks, which should return a
+#'     vector of minor breaks.
 #' @param labels One of:
 #'   - `NULL` for no labels
 #'   - `waiver()` for the default labels computed by the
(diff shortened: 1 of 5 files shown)


hadley:
This helps a lot, but there's a major drawback: the breaks always start within the limits, so this doesn't generate any minor breaks between the minimum and first break, and the last break and the maximum.

You can see the problem with a log scale with 10 minor breaks between each major break:

Screenshot 2019-10-27 at 9 53 21 AM

hadley:
But maybe I can handle that by also using the limits, and doing something more sophisticated than just using a fixed number of breaks for log scales.

hadley:
Having briefly explored the possibility of using the range, I think your suggestion would be a better approach @paleolimbot.

thomasp85:
@Hadley where are we on this?

hadley:
Lets hold on this as it needs more changes to be fully useful.

TyStark:
minor_breaks
Hello, I am new to contributing to open source, but I was trying to contribute a solution to this exact situation with minor_breaks on a log10 scale when I found this PR. It kept coming up at work and there seemed to be no convenient solution. The attached image is from a function I wrote that is supplied to minor_breaks that generalizes to any axis on a log10 scale. As you can see, it includes minor breaks past the last major breaks. I'm not exactly sure how to implement this, it doesn't seem like I can simply create a variable set to my function in scale-.r or scale-continuous.r, but with some help I may be able to help get this functionality into ggplot2.

thomasp85:
@Hadley is it viable to get this into the next release?

hadley:
@thomasp85 I have no idea

teunbrand:
>  Breaks would still need to be censored before they are drawn, but this could be moved to guide_train(). 

I explored this for a bit, but there are plenty of places outside the guides where that approach gives issues, mostly in coords when dealing with the panel grid.

I think the change in the PR improves the current situation, and any 'full' solution would include these changes as well.
Therefore, I'll suggest to merge this PR under the banner of incremental improvement and deal with the out-of-bounds breaks separately once we find a satisfactory solution to that problem.

PR #3593, opened 2019-10-28
Fix guide_bins() when the data has no size

### LINKED ISSUE #3582: Report a bug of guide-bins: it cannot work with geom_sf.
The guide-bins seems cannot work with geom_sf.

Some examples:

For geom_point, both "coloursteps" and "bins" guides work well:
```
library(ggplot2)
library(tibble)

ggplot(tibble(x = 1:10, y = x, z = x), aes(x, y, fill = z)) +
    geom_point(shape = 21, size = 10) +
    scale_fill_stepsn(guide = "coloursteps", colours = terrain.colors(10))
```
p1

```
ggplot(tibble(x = 1:10, y = x, z = x), aes(x, y, fill = z)) +
    geom_point(shape = 21, size = 10) +
    scale_fill_stepsn(guide = "bins", colours = terrain.colors(10))
```
p2

But, for geom_sf, only "coloursteps" can work. 
There seems to be a bug in guide-bins.
```
nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"))

ggplot() +
    geom_sf(aes(fill = AREA), data = nc) +
    scale_fill_stepsn(guide = "coloursteps", colours = terrain.colors(10))
```
p3
```
ggplot() +
    geom_sf(aes(fill = AREA), data = nc) +
    scale_fill_stepsn(guide = "bins", colours = terrain.colors(10))

# Error in do.call("cbind", lapply(guide$geoms, function(g) g$data$size/10))[seq_len(n_keys),  : 
# subscript out of bounds
```

yutannihilation:
Fix #3582

The result of `do.call("cbind", lapply(guide$geoms, function(g) g$data$size / 10))` can be an empty matrix, which cannot be subsetted. I guess the branch `if (nrow(key_size_mat) == 0 || ncol(key_size_mat) == 0) {` is prepared to handle this.

### DIFF
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -322,6 +322,8 @@ guide_gengrob.bins <- function(guide, theme) {
     lapply(guide$geoms, function(g) g$data$size / 10)
   )
 
+  # key_size_mat can be an empty matrix (e.g. the data doesn't contain size
+  # column), so subset it only when it has any rows and columns.
   if (nrow(key_size_mat) == 0 || ncol(key_size_mat) == 0) {
     key_size_mat <- matrix(0, ncol = 1, nrow = n_keys)
   } else {
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -320,10 +320,12 @@ guide_gengrob.bins <- function(guide, theme) {
 
   key_size_mat <- do.call("cbind",
     lapply(guide$geoms, function(g) g$data$size / 10)
-  )[seq_len(n_keys), , drop = FALSE]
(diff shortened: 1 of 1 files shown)


clauswilke:
Do you want to add a small comment about the empty matrix issue to the code? I generally think that whenever we identify these kinds of potential problems, it makes sense to point them out in the code, so somebody else doesn't repeat the mistake in the future.

yutannihilation:
Thanks, I will.

yutannihilation:
Thanks.

PR #3594, opened 2019-10-28
Better error message when trying to add a function to a plot.

### LINKED ISSUE #3571: Suggestion for more informative error message when () missing from geom_*
If you omit the `()` at the end of a geom, you get the following error:

``` r
library(ggplot2)

ggplot(mtcars, aes(y = mpg, group = vs)) +
  geom_boxplot
#> Error: Don't know how to add geom_boxplot to a plot
```

I think the error would be more informative if it continued on to say _"Did you mean `geom_boxplot()`?"_

clauswilke:
This closes #3571:

``` r
library(ggplot2)
ggplot() + geom_point
#> Error: Don't know how to add geom_point to a plot. Did you mean geom_point()?
```

<sup>Created on 2019-10-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/plot-construction.r
```diff
@@ -100,6 +100,13 @@ ggplot_add.data.frame <- function(object, plot, object_name) {
   plot
 }
 #' @export
+ggplot_add.function <- function(object, plot, object_name) {
+  stop(
+    "Don't know how to add ", object_name, " to a plot. Did you mean ",
+    object_name, "()?", call. = FALSE
+  )
```
yutannihilation: Do we use `abort()`? Or will it be done in #3526?

NAMESPACE
```diff
@@ -105,6 +106,7 @@ S3method(makeContent,labelgrob)
 S3method(makeContext,dotstackGrob)
 S3method(merge_element,default)
 S3method(merge_element,element)
+S3method(merge_element,element_blank)
```
yutannihilation: These diff seems to come from #3570, but I'm OK whether or not they are included in this PR.

R/plot-construction.r
clauswilke: As of now, we're using `stop()`. See e.g. here (purposefully not using permalink):
https://github.com/tidyverse/ggplot2/blob/master/R/plot-construction.r#L89-L92

NAMESPACE
clauswilke: Good catch. I opened a separate PR to disentangle these, see #3595.

R/plot-construction.r
yutannihilation: OK.

PR #3596, opened 2019-10-28
Add after_script to ensure the documents are regenerated

yutannihilation:
It's a common mistake to forget to run `devtools::documents()`, but it's hard to detect them by reviewing on GitHub. It would be nice if it can be detected on Travis CI. This PR adds an `after_script` for this; the error might be false positive so it can be ignored by manually checking the CI results.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ after_success:
 
 after_script:
   # make sure the documents are regenerated
-  - Rscript -e 'devtools::documents()'
+  - Rscript -e 'devtools::document()'
   - |
     if [ -n "$(git diff --name-only man)" ]; then
       echo "Did you forget to regenerated the documents? (This error might be ignorable)."
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,15 @@ env:
 after_success:
   - Rscript -e 'covr::codecov()'
 
+after_script:
+  # make sure the documents are regenerated
+  - Rscript -e 'devtools::documents()'
+  - |
+    if [ -n "$(git diff --name-only man)" ]; then
+      echo "Did you forget to regenerated the documents? (This error might be ignorable)."
+      exit 1
+    fi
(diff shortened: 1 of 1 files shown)


clauswilke:
Presumably the automatic check on this is going to fail until #3595 is merged? This will be a good test case.

yutannihilation:
> Presumably the automatic check on this is going to fail until #3595 is merged?

Ah, yes! Let's see if this works as expected...

yutannihilation:
Ouch...

> Error: 'documents' is not an exported object from 'namespace:devtools'

clauswilke:
It's `devtools::document()`.

yutannihilation:
Sorry, I misunderstood about the build stages. `after_script` cannot make the build fail. Let me think if I can insert the check into `script`.

> The exit code of after_success, after_failure, after_script, after_deploy and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as failed. (https://docs.travis-ci.com/user/job-lifecycle/#breaking-the-build)

yutannihilation:
Hmm, it's actually failed. I'm confused...

https://travis-ci.org/tidyverse/ggplot2/jobs/604197809#L1227-L1232

yutannihilation:
@clauswilke It might take some time to figure out the proper settings here, so you don't need to wait for this to merge #3595.

yutannihilation:
My guess is that explicit `exit <non-zero status>` works while execution of some script that exits non-zero status (e.g. `Rscript -e 'devtools::documents()'`) is ignored. But, I'm not sure if we can rely on this behaviour.

yutannihilation:
They say I should not use `exit`. Probably I need to think about a better approach to do this in R, so I'm closing this for now.

https://docs.travis-ci.com/user/job-lifecycle/#how-does-this-work-or-why-you-should-not-use-exit-in-build-steps

PR #3597, opened 2019-10-29
Handle a quosure properly in strip_dots() and is_calculated()

### LINKED ISSUE #3552: rlang warnings in ggplot2 unit tests
When I run the ggplot2 unit tests, I get two rlang warnings. Does anybody else see this? Is this a misconfiguration on my end or is something not quite right with the test or the code base?

```
==> devtools::test()

Loading ggplot2
Testing ggplot2
✔ |  OK F W S | Context
✔ |   1   2   | Adding plot elements
───────────────────────────────────────
test-add.R:4: warning: mapping class is preserved when adding uneval objects
Subsetting quosures with `[[` is deprecated as of rlang 0.4.0
Please use `quo_get_expr()` instead.
This warning is displayed once per session.

test-add.R:4: warning: mapping class is preserved when adding uneval objects
Subsetting quosures with `[` is deprecated as of rlang 0.4.0
Please use `quo_get_expr()` instead.
This warning is displayed once per session.
───────────────────────────────────────
✔ |  14       | test-aes-calculated.r
✔ |   9       | Aesthetics (grouping) [0.4 s]
✔ |  10       | Aes - setting values [0.5 s]
...
```

Strangely, I cannot reproduce these warnings in a reprex:
``` r
library(ggplot2)
library(testthat)

# the test from test-add.R, line 4
p <- ggplot(mtcars) + aes(wt, mpg)
expect_identical(class(p$mapping), "uneval")
```

<sup>Created on 2019-10-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Fix #3552

Currently, `strip_dots()` and `is_calculated()` treat a quosure as a call, and it surprisingly works for now, but it's fragile. A quosure should be treated as a quosure...

### DIFF
--- a/R/aes-calculated.r
+++ b/R/aes-calculated.r
@@ -42,6 +42,8 @@ is_calculated <- function(x) {
     FALSE
   } else if (is.symbol(x)) {
     is_dotted_var(as.character(x))
+  } else if (is_quosure(x)) {
+    is_calculated(quo_get_expr(x))
   } else if (is.call(x)) {
     if (identical(x[[1]], quote(stat))) {
       TRUE
@@ -66,6 +68,12 @@ strip_dots <- function(expr) {
     } else {
       expr
     }
+  } else if (is_quosure(expr)) {
+    # strip dots from quosure and reconstruct the quosure
+    expr <- new_quosure(
+      strip_dots(quo_get_expr(expr)),
+      quo_get_env(expr)
+    )
   } else if (is.call(expr)) {
     if (identical(expr[[1]], quote(stat))) {
       strip_dots(expr[[2]])


clauswilke:
Oh, maybe one minor issue: Can you amend the commit message so it states that the commit closes  #3552?

yutannihilation:
Sure, I tweaked the commit message. But, I'm not familiar with this manner. Why should it be written in addition to the description of the pull request?

clauswilke:
I thought that github was automatically closing issues only based on commit messages, but now I see this [may not be true.](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) In any case, I think it makes sense to put it into the commit message, because that is part of the actual git log. If at some point in the future you stumble across this commit and you wonder why it was made then you can see directly from the commit message that it was related to a specific issue. If it's only mentioned in the PR, then you'd have to first dig out the PR that contributed the commit and then read through the discussion of that PR to find the same info.

yutannihilation:
Thanks, makes sense. I tend to rewrite (or remove) the individual commit message when I "Squash and merge" on GitHub, but I'll keep in mind to include the original issue number in the message from now on!

thomasp85:
General advice is really just to put it in the PR description as that will close it. Any commit messages will potentially be removed during squashing anyway so it doesn’t make sense to have it there

yutannihilation:
Yeah, ultimately, commit messages are not what we can rely on, but a useful commit message is still useful (e.g. Linux's commit messages are really cool). So, I just decided to do my best to write them a bit more informative. It's my personal decision and I don't think we can enforce it as a rule here, though.

PR #3628, opened 2019-11-16
Typo fix: Remove double "just" in documentation

lwjohnst86:
:)


yutannihilation:

lwjohnst86:
Done :) I only updated the change in the rd, not everything that was updated (the roxygen2 version was higher for me, so it changed a lot of things). Hope that is fine :)

PR #3629, opened 2019-11-17
WIP: Theme caching

clauswilke:
This is an idea for theme caching. We can calculate the entire theme right before rendering, so that we don't have to calculate theme elements on the fly over and over. It makes little difference in speed for rendering a single plot, but it speeds up gganimate by about 5% for a realistic example.

The code I'm using to test:
``` r
library(ggplot2)
library(gganimate)

p <- ggplot(mtcars, aes(factor(cyl), mpg, color = factor(cyl))) + 
  geom_boxplot() + 
  ggtitle("Title here", subtitle = "Subtitle here") +
  transition_states(gear)
```

The output from benchmarking:
```
> bench::mark(animate(p), iterations = 5)
theme caching turned off
                                                                                 
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
# A tibble: 1 x 13
  expression   min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr> <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 animate(p)   23s  23.6s    0.0425    53.2MB     2.38     5   280      1.96m
# … with 4 more variables: result <list>, memory <list>, time <list>, gc <list>
Warning message:
Some expressions had a GC in every iteration; so filtering is disabled. 

> bench::mark(animate(p), iterations = 5)
theme caching turned on
                                                                                 
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
Frame 100 (100%)
Finalizing encoding... done!
# A tibble: 1 x 13
  expression   min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr> <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 animate(p) 21.4s  22.7s    0.0447    61.9MB     2.33     5   261      1.87m
# … with 4 more variables: result <list>, memory <list>, time <list>, gc <list>
Warning message:
Some expressions had a GC in every iteration; so filtering is disabled. 
```

Note: gganimate requires a small modification, see here: https://github.com/clauswilke/gganimate/commit/442a471c886fe8bc5880cb2fa3cba93f3c06038a

@thomasp85 Do you think this is worthwhile pursuing? I had hoped for a bigger speed increase, but 5% is not nothing.


clauswilke:
The unit tests fail in part because the legend drawing code breaks theme inheritance, e.g. here:
https://github.com/tidyverse/ggplot2/blob/f418e92bd8c249759e7af55625f562a845f53460/R/guide-legend.r#L358-L362
That's a bit of a problem, but we could work around it by excluding problematic theme elements from caching.

thomasp85:
Unless we can make it generally stable I don’t think we should introduce this complexity for a 5% gain

clauswilke:
Yes, I had hoped this would have a bigger effect. Let's shelve this until the rest of the rendering pipeline is so fast that this is actually a meaningful improvement. :-)

PR #3637, opened 2019-11-23
Upgrade to roxygen2 7.0.1

hadley:
Including converting all uses of `\%` back to `%`, and removing roxygen2 tags that the new warnings indicated were superfluous.


hadley:
Hmmm that's a little concerning. @thomasp85 can you please take a look? 3 unit tests are failing but only for R-release on linux

yutannihilation:
Actually tests fails on my local. I think we can just update the visual cases, but I don't know well about the details of the latest release of scales package, so please confirm this is expected or not. I create another PR to fix this: https://github.com/tidyverse/ggplot2/pull/3638

clauswilke:
The scales package changed the code used to calculate color ramps (see here: https://github.com/r-lib/scales/commit/eb24db71358a6cff9bef274c44f7310e5f6b4fbd). As a result, I would expect all visual tests with color ramps to fail. I would recommend visually checking that the new color ramps look fine and then accepting the new test cases.

yutannihilation:
Oh, thanks for identifying the commit.

> I would recommend visually checking that the new color ramps look fine

Yes, I've checked on my local. I'll post the comparison on #3638.

hadley:
Thanks all! I've merged the latest master into this branch so hopefully the next round of checks will pass.

PR #3641, opened 2019-11-25
Update link in docu

katrinleinweber:
Hello :-) Is it intended that this link directs to the package-level docu, instead of the relevant function?


yutannihilation:

katrinleinweber:
Will do in the next few days. Thanks for the hint!

katrinleinweber:
Sure :-) Like this?

katrinleinweber:
You're welcome & thanks for merging :-)

PR #3643, opened 2019-11-25
Correct a spelling mistake in the documatation of continuous_scale

microly:
Correct a spelling mistake in this line:

https://github.com/tidyverse/ggplot2/blob/5dfe49829d9b0b35f9cf0783d9915c6112f88eec/R/scale-.r#L52

It should be "scales, which always use [scales::rescale()]. ".


yutannihilation:

microly:
Thanks for your reply!
I am not familiar with regenerating the documents, but I will try it in my free time.

microly:
Is it ok now?
@yutannihilation

microly:
Thanks for merging!

Issue #3653, opened 2019-12-04
Annotation_map does not work with coord_sf


markpayneatwork:
When using`coord_sf()` to change the projection of a map, the background layer produced by  `annotation_map()` is not getting reprojected.

First, a plot  without changing projection, using the annotation_map layer to give more context to North Carolina
```r
library(ggplot2)
library(sf)
nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"), quiet = TRUE)

ggplot() +
  annotation_map(map_data("state"),fill="antiquewhite",colour="darkgrey")+
  geom_sf(data = nc)
```
![image](https://user-images.githubusercontent.com/4990832/70135829-1a8dfe80-168b-11ea-96d1-6734c4480b6b.png)

However, changing the projection looses the annotation layer, but reprojects the geometry layer
```r
ggplot() +
  annotation_map(map_data("state"),fill="antiquewhite",colour="darkgrey")+
  geom_sf(data = nc)+
  coord_sf(crs = st_crs(3347))
```
![image](https://user-images.githubusercontent.com/4990832/70135882-32658280-168b-11ea-8042-80ee3df9e653.png)

It looks to me like the annotation_layer is not being reprojected.

clauswilke:
This happens because `annotation_map()` is old code that predates `geom_sf()`. Only sf layers are reprojected. While it would be nice to reproject other layers, there are numerous technical hurdles that make me think it's not going to happen anytime soon.

At this point, the best way forward is probably to amend the documentation accordingly.

thomasp85:
@clauswilke has this been fixed by #3659?

clauswilke:
Yes, correct, just need to set the appropriate default crs. This particular case is actually included in the examples, so I think we're good as is.

``` r
library(ggplot2)
library(sf)
#> Linking to GEOS 3.8.1, GDAL 3.2.1, PROJ 7.2.1
nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"), quiet = TRUE)

ggplot() +
  annotation_map(map_data("state"),fill="antiquewhite",colour="darkgrey")+
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347), default_crs = st_crs(4326))
```

![](https://i.imgur.com/Z2gQ2Rh.png)

<sup>Created on 2022-05-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Issue #3654, opened 2019-12-04
Enable crs transformations of non-sf layers


clauswilke:
The recent issue #3653 prompted me to think a little deeper about mixing sf and non-sf layers. Currently, we can mix these layers if the non-sf layers are properly projected.

``` r
library(ggplot2)

nc <- sf::st_read(system.file("shape/nc.shp", package = "sf"), quiet = TRUE)

df <- data.frame(
  lat = 35.76667,
  long = -78.63333,
  name = "Raleigh"
)

# works, crs is longlat
ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1)
```

![](https://i.imgur.com/WojXNRY.png)

``` r
# doesn't work, non-sf layers are not transformed
ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264)
```

![](https://i.imgur.com/xtsmO0e.png)

``` r
# manually transforming coordinates works
df_trans <- data.frame(
  lat = 734165.51156997,
  long = 2108774.28357641,
  name = "Raleigh"
)

ggplot(df_trans, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264)
```

![](https://i.imgur.com/UrxqijU.png)

<sup>Created on 2019-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It would be great (and make life easier in many ways) if non-sf layers could also be transformed by `coord_sf()`. However, this poses a problem: The sf layers carry information about their crs, and therefore `coord_sf()` can figure out if and how they should be transformed. The non-sf layers don't carry that information, and thus whatever `coord_sf()` does will be wrong in some cases. We could add a `crs` argument to all layers, but that seems ugly since it breaks orthogonality between different parts of the API.

[...]

PR #3655, opened 2019-12-04
WIP: sf-ify regular layers

clauswilke:
First attempt at implementing the idea proposed in #3654.

``` r
library(ggplot2)

df <- data.frame(
  lat = 35.76667,
  long = -78.63333,
  name = "Raleigh"
)

# doesn't work, non-sf layers are not transformed
ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264)
```

``` r

# with_sf() fixes it
ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  with_sf(geom_point()) +
  with_sf(geom_text(aes(label = name), hjust = -.1, vjust = 1.1)) +
  coord_sf(crs = 2264)
```

<sup>Created on 2019-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


markpayneatwork:
One thing I can see that is missing from the `with_sf()` approach is the ability to tell `with_sf` what the default actually is. Maybe this could be supplied as an argument?

clauswilke:
@paleolimbot Since we don't have a good way of chaining stats, I'm not sure that using a stat just for the purpose of a coordinate transformation is the right way to go. Also not sure how that would play out with geoms such as `geom_hline()` and `geom_vline()` which need to know about the extent of the coord. But I'd be interested to hear other opinions. @yutannihilation @thomasp85?

@markpayneatwork Yes, of course. This is a work in progress. Source crs can be specified now.

``` r
library(ggplot2)

df <- data.frame(
  lat = 35.76667,
  long = -78.63333,
  name = "Raleigh"
)

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  with_sf(
    geom_point(),
    geom_text(aes(label = name), hjust = -.1, vjust = 1.1)
  ) +
  coord_sf(crs = 2264)
```



``` r


# transformed data
df_trans <- data.frame(
  lat = 734165.51156997,
  long = 2108774.28357641,
  name = "Raleigh"
)

ggplot(df_trans, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  with_sf(
    geom_point(),
    geom_text(aes(label = name), hjust = -.1, vjust = 1.1),
    crs = 2264
  ) +
  coord_sf(crs = 2264)
```



<sup>Created on 2019-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Thinking some more about this, I don't think a stat can do this.

``` r
library(ggplot2)

ggplot(nc) +
  geom_sf(size = 0.1, fill = "white") +
  with_sf(
    geom_hline(yintercept = 34:36),
    geom_vline(xintercept = c(-84, -80, -76))
  ) +
  coord_sf(crs = 2264)
```



<sup>Created on 2019-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Or this.

``` r
library(ggplot2)
library(readr)
library(ggforce)

df <- read_csv(file = "name,population,lat,long
Charlotte,827097,35.227,-80.843
Raleigh,451066,35.772,-78.639
Greensboro,285342,36.073,-79.792
Durham,257636,35.994,-78.899
Winston-Salem,241218,36.1,-80.244")

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  with_sf(
    geom_point(),
    geom_mark_hull()
  ) +
  coord_sf(crs = 2264)
```



<sup>Created on 2019-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Fix #3653.

``` r
library(ggplot2)
library(sf)

ggplot() +
  with_sf(
    annotation_map(map_data("state"), fill="antiquewhite", colour="darkgrey")
  ) +
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347))
```



<sup>Created on 2019-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

microly:
This is my solution:

Firstly, get the data from non-sf layer.
```
library(dplyr)
library(sf)
library(purrr)

# suppose the data is:
data <- data.frame(cl = letters[1:3],
                   x = c(-76, -77, -78), y = c(34, 35, 36))
data
# data
#  cl   x  y
# 1  a -76 34
# 2  b -77 35
# 3  c -78 36
```

Secondly, do the work.

```
data_other <- data %>% select(-x, -y)

data_xy_new <- data %>% 
  select(x, y) %>% 
  rowwise() %>% 
  mutate(geometry = map2(x, y, ~st_point(x = c(.x, .y), dim = "XY"))) %>% 
  ungroup() %>% 
  st_sf(crs = 4267) %>% 
  st_transform("+init=epsg:3857") %>% 
  st_coordinates() %>% 
  as.data.frame() %>% 
  rename(x= X, y = Y)

data_new <- cbind(data_other, data_xy_new)

data_new
#   cl        x       y
# 1  a -8460238 4028826
# 2  b -8571564 4163903
# 3  c -8682888 4300642
```

Finally, replace the origin data by data_new.

clauswilke:
@microly Your code example shows how to transform the data manually, but it doesn't address the issue of how we get ggplot2 to do this automatically.

clauswilke:
@hadley Are you Ok with this addition to the grammar?

microly:
Yes, right.
I am not on my working computer now.
So I supply a temp solution, which may be just the core of a full solution.


 
---Original---

clauswilke:
@microly It's off-topic, though. This is the discussion thread for a pull request that already provides a solution to the problem, including code exactly like what you posted.

microly:
got it~
I am sorry...
 
---Original---

yutannihilation:
Is it possible to let `CoordSf` transform all data in this way if the user wants it? For example:

``` r
ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264, transform_non_sf_layers = TRUE)
```

I feel `with_*()` syntax is a bit overspec when we need just a single modification; I think the main advantage of `with_*()` is that it can be nested to express the series of modifications (e.g. `with_blur(with_sobel(...))`).

clauswilke:
@yutannihilation The problem I see is that the `transform()` function of `coord_sf()` doesn't know anything about the layer that calls it, and therefore doesn't know what to assume about the coordinates coming in: 
https://github.com/tidyverse/ggplot2/blob/913e936ce4abfd933897ca9e5425dea76bd7d596/R/coord-sf.R#L43-L58

So you'd have to first transform all incoming sf layers to long-lat or similar in `setup_data()`, and then transform again to the target crs in `transform()`. But you also have to transform all incoming sf layers to the target crs at some point before `transform()` is called, because you need to figure out the limits of the plot. So you're going from one transformation per sf layer to three. I don't know if this can be avoided.

yutannihilation:
> because you need to figure out the limits of the plot

If we assume most of the usage of non-sf layers are for annotation, this step might not be necessary, but your explanation makes sense, thanks. Hmm...

clauswilke:
Another way to think about the problem: The current `coord_sf()` is essentially a Cartesian coordinate system operating in projected coordinates. To have regular geoms work natively with longitude/latitude data, we'd have to implement a non-Cartesian coordinate system operating in non-projected coordinates. These two coordinate systems are sufficiently different that we'd end up with two coords, e.g. `coord_sf()` and `coord_sf2()` (or whatever better name somebody can come up with). While this is an approach that could be implemented, I'm not sure it's less confusing.

clauswilke:
Thanks for the suggestion, Dewey. This may work, though I foresee some issues with limits. In what CRS would they be specified? Do we need a third parameter, `limits.crs`, that defines the limits CRS to be used? If limits are specified in long-lat, do we run into problems with non-rectangular areas and/or areas that cross the international dateline?

hadley:
IIRC my original intention was for other layers to be automatically transformed, assuming that the coordinates were specified in lat and long, and this is supported by one of the examples:

```R
ggplot(nc) +
  geom_sf() +
  annotate("point", x = -80, y = 35, colour = "red", size = 4)
```

I'd really prefer if we could perform this transformation automatically — but I don't know enough about the constraints to know whether or not this is possible.

clauswilke:
@hadley I think the main constraint is the API: We'd have to add an argument to the `transform()` function of `coord_sf()`, so that the calling layers can specify what CRS they are assuming. The least invasive change would be to require the layers that don't want to be transformed to announce so. This would only require changes in `geom_sf()` and related geoms. It's a breaking change though for any downstream packages that have defined their own sf-related geoms.

clauswilke:
Oh, and it's also a breaking change for any code currently in existence that mixes `geom_sf()` with other geoms using pre-transformed data, though I'm not sure whether people commonly do this.

hadley:
@clauswilke I think those breakages should be minor, and would be worth it (esp since I'm pretty sure that this is what I always intended)

clauswilke:
@hadley Ok, let me take a stab at this and then we can see what it looks like.

clauswilke:
@hadley It turns out changing the `transform()` API doesn't work after all, because geoms don't in general know what data they are drawing. E.g., in `stat_sf_coordinates(geom = "point")`, the geom has no idea it is working on properly transformed sf data.

Thus, the only other alternative I see is to transform everything to a common CRS first and then transform a second time right before drawing. Maybe I'll take a stab at that to see where it leads.

microly:
What is about that with_sf works with data? Like: 

ggplot() +
  geom_sf(data = nc) +
  geom_point(aes(x = long, y = lat), data = with_sf(df, crs = the_crs_of_df))

In essence, it is a issue about transforming data, so we can only wrap the data...
Please forgive me if this is a foolish idea...

clauswilke:
Thanks @paleolimbot, I think your comment just cut the Gordian knot. We can continue treating geometry objects as we currently do, no change required, and we adopt the convention that regular position data is in long-lat (to be configurable in `coord_sf()`). Problems arise when geometry data is converted into regular position data, as `stat_sf_coordinates()` does, but it can just transform the geometry data as a first step. The final open question is whether we can get `coord_sf()` and `stat_sf_coordinates()` to talk to each other, so one can know what the other's default crs is, but worst case scenario both would take a default crs argument.

clauswilke:
@paleolimbot See #3659.

clauswilke:
Closing this as it has been superseded by #3659.

PR #3657, opened 2019-12-05
address duplicate aesthetic listings in documentation, re #3618

### LINKED ISSUE #3618: Duplicated aesthetics on some documentations of Geom/Stat
Probably due to the change of the specification of `required_aes` (#3506), some documentations of Geom/Stat now show both **"x or y"** and, **"x"** and **"y"** separately on their Aesthetics section. This seems a bit confusing, so it's nice if we can fix this.

For example, https://ggplot2.tidyverse.org/dev/reference/geom_linerange.html#aesthetics:

screenshot of `geom_linerange.html`

carywreams:
Fix #3618

```
    in utilities-help.r::rd_aesthetics_item, split required
    aesthetics specified with pipe operator (e.g. "x|y")
    into separate items so that they may be excluded from
    the list of all aesthetics prior to rendering the
    documentation of parameters.
```

My thoughts on approach included in #3618 comments.
Also have concerns about results of ```testthat``` run, but results are identical to master branch.

### DIFF
--- a/R/utilities-help.r
+++ b/R/utilities-help.r
@@ -1,4 +1,3 @@
-
 rd_aesthetics <- function(type, name) {
   obj <- switch(type,
     geom = check_subclass(name, "Geom", env = globalenv()),
@@ -22,7 +21,9 @@ rd_aesthetics <- function(type, name) {
 rd_aesthetics_item <- function(x) {
   req <- x$required_aes
   req <- sub("|", "} \\emph{or} \\code{", req, fixed = TRUE)
-  all <- union(req, sort(x$aesthetics()))
+  req_aes <- unlist(strsplit(x$required_aes, "|", fixed = TRUE))
+  optional_aes <- setdiff(x$aesthetics(),req_aes)
+  all <- union(req, sort(optional_aes))
 
   ifelse(all %in% req,
     paste0("\\strong{\\code{", all, "}}"),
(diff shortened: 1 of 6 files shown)


carywreams:
Well, damn.
I'm off the map now.
Here, there be monsters.

lil help ?

R/utilities-help.r
```diff
@@ -22,7 +21,9 @@ rd_aesthetics <- function(type, name) {
 rd_aesthetics_item <- function(x) {
   req <- x$required_aes
   req <- sub("|", "} \\emph{or} \\code{", req, fixed = TRUE)
-  all <- union(req, sort(x$aesthetics()))
+  req_aes <- unlist(strsplit(x$required_aes, "|", fixed = TRUE))
+  optional_aes <- setdiff(x$aesthetics(),req_aes)
```
yutannihilation:
```suggestion
  optional_aes <- setdiff(x$aesthetics(), req_aes)
```

carywreams:
Pull request description change?

yutannihilation:
Thanks, perfect! I'm merging this but will wait for a while for other comments.

carywreams:
Thanks.

When you have a moment, I'd like some more information on how to know whether to ignore the check failures that happened here -- or what I could have done to better prep it.

yutannihilation:
No worries, it's not your fault. In my rough understanding, codecov complains because this PR adds lines to the location where we don't have any unit tests. You can add tests to improve the test coverage, but I don't think it's worth doing for this documentation-only function.

yutannihilation:
Merging now. Thanks!

PR #3659, opened 2019-12-06
Implement default crs for non-sf objects in coord_sf().

clauswilke:
Second attempt at implementing the idea proposed in #3654. Based on discussion in #3655.

The main issue that remains to be solved is proper setting of x and y scale limits. 

``` r
library(ggplot2)
library(readr)
library(ggforce)
library(sf)

df <- data.frame(
  lat = 35.76667,
  long = -78.63333,
  name = "Raleigh"
)

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264)
```

``` r

ggplot(nc) +
  geom_sf(size = 0.1, fill = "white") +
  geom_hline(yintercept = 34:36) +
  geom_vline(xintercept = c(-84, -80, -76)) +
  coord_sf(crs = 2264)
```

``` r

df <- read_csv(file = "name,population,lat,long
Charlotte,827097,35.227,-80.843
Raleigh,451066,35.772,-78.639
Greensboro,285342,36.073,-79.792
Durham,257636,35.994,-78.899
Winston-Salem,241218,36.1,-80.244")

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_mark_hull() +
  coord_sf(crs = 2264)
```

``` r

ggplot() +
  annotation_map(map_data("state"), fill = "antiquewhite", colour = "darkgrey") +
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347))
```

``` r

# this works except the scales can't figure out the correct limits

ggplot(nc) +
  geom_sf() +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = st_crs(3347))
```

``` r

ggplot(nc) +
  stat_sf_coordinates(geom = "point") +
    coord_sf(crs = st_crs(3347))
```

<sup>Created on 2019-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/coord-sf.R
```diff
@@ -375,20 +393,44 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
...
+      is.null(target_crs) || is.null(source_crs) || is.null(data) ||
+      !all(c("x", "y") %in% names(data))) {
+    return(data)
+  }
+
+  sf_data <- sf::st_sfc(
+    sf::st_multipoint(cbind(data$x, data$y)),
+    crs = source_crs
+  )
+  sf_data_trans <- sf::st_transform(sf_data, target_crs)[[1]]
```
paleolimbot: I forget whether it's safe to assume that `data$x` and `data$y` are all finite. `st_transform()` fails if there are any non-finite coordinates.  (see  `sf::st_transform(sf::st_sfc(sf::st_multipoint(cbind(c(1, 1), c(2,  NA))), crs = 4326), crs = 3857)`). If it's not safe to assume that these values are finite, you could implement something like I did in  `ggspatial::xy_transform()`.

clauswilke: I think infinite values could occur and would have to be handled somehow. I'm more worried about getting the scale limits right at this point though.

clauswilke:
I haven't figured out yet how to get the scale limits correct. I'll provide some notes as food for thought.

Currently, the scale limits for geometry columns are calculated in `StatSf`, which calculates a bounding box for each object and then adds columns `xmin`, `xmax`, `ymin`, `ymax` to the data:
https://github.com/tidyverse/ggplot2/blob/16ed4d00c82dbc20f70da893506e093371ba664c/R/stat-sf.R#L6-L14

This trick works nicely if we're exclusively working in projected coordinates, but it doesn't if non-sf layers are in non-projected coordinates, as then the data units don't match. We could back-transform the bounding boxes to long-lat coordinates, but I'm worried that that would lead to strange limit choices in cases where meridians and parallels are not mostly aligned with the y and x axis, respectively. As much as possible, we should perform limit calculations in transformed coordinates.

Since `CoordSf` has final say in the limits, maybe it's somehow possible to carry the bounding box info of geometry objects until the point where the final panel params are set up, and only merge then with the limits the scales have determined from the position data:
https://github.com/tidyverse/ggplot2/blob/16ed4d00c82dbc20f70da893506e093371ba664c/R/coord-sf.R#L128-L147

I'm just not sure yet how to do this, in particular in a way that works in faceted graphs. On the flip side, free scales don't seem to be currently supported anyways, so maybe that's not a regression.
``` r
library(ggplot2)

ggplot(nc[1:5, ]) + geom_sf() + facet_wrap(~NAME)
```



``` r
ggplot(nc[1:5, ]) + geom_sf() + facet_wrap(~NAME, scales = "free")
#> Error: coord_sf doesn't support free scales
```



<sup>Created on 2019-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Maybe `StatSf` can register the bounding boxes with `CoordSf`, which then can use them when it determines the final scale limits.

clauswilke:
I think I've got limits working, but it required quite a bit of fiddling and communication back-and-forth between the stat and the coord.

@paleolimbot Could you take a look?

``` r
library(ggplot2)
library(sf)

ggplot(nc) +
  geom_sf() +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = st_crs(3347))
```



``` r

ggplot(nc) +
  stat_sf_coordinates(geom = "point") +
    coord_sf(crs = st_crs(3347))
```



``` r

# scale limits are given in longitude/latitude
ggplot(nc) +
  geom_sf() +
  scale_x_continuous(limits = c(-86, -74)) +
  coord_sf(crs = 3347)
```



``` r

# coord limits are given in projected coordinates
ggplot(nc) +
  geom_sf() +
  coord_sf(xlim = c(6800000, 7900000), crs = 3347)
```



<sup>Created on 2019-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Some more testing of this code. First, I wanted to see whether it works in cases that span the international dateline. The answer is yes as long as an appropriate CRS is chosen. This is reasonable.

``` r
library(tidyverse)
library(rnaturalearth)
library(rnaturalearthdata)
library(sf)

world <- ne_countries(scale = "medium", returnclass = "sf")
fiji_polys <- world %>%
  filter(name == "Fiji") %>%
  pull(geometry) %>%
  st_cast("POLYGON") %>% 
  st_sf() %>%
  mutate(label = letters[1:n()])

# doesn't work, long-lat across the dateline causes problems 
ggplot(data = fiji_polys) +
  geom_sf()
```



``` r

# works
ggplot(data = fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460)
```



``` r

# works even with separate groups on either
# side of the date line
ggplot(data = fiji_polys) +
  geom_sf(aes(fill = label)) +
  coord_sf(crs = 3460)
```



<sup>Created on 2019-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Second, I wanted to see how coordinate expansion is affected by non-sf layers. The answer is that coordinate expansion can change depending on whether the default crs of non-sf layers is set equal to the primary crs or not. I assume there's no way around this. When the default crs is long-lat, then the x and y scales are trained on these values, and the resulting scale limits can be different from what we would choose in an appropriate projection.

``` r
library(ggplot2)
library(patchwork)

p1 <- ggplot(nc) + geom_sf() +
  coord_sf(crs = 3347)
p2 <- ggplot(nc) + geom_sf() +
  coord_sf(crs = 3347, default_crs = 3347)

# when only sf layers are used, the panel expansion does not
# depend on the default crs
p1 / p2
```



``` r

# however, when there are non-sf layers also, the panel expansion
# depends on the default crs
(p1 + stat_sf_coordinates()) / (p2 + stat_sf_coordinates())
```



<sup>Created on 2019-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
I think this generally works, but there could be some implications for `geom_stars()` from the stars package. @edzer, could you comment/take a look?

clauswilke:
We don't currently allow `scales = "free"` in any coord that enforces a fixed aspect ratio, and it's not clear that we ever will. But in general, I'd prefer a solution that can scale separately for different panels.

I like the idea of a `scale_geometry()`, I just don't know enough about scales to know how to implement that. Could you give me some pointers and/or take a stab on top of my PR? Would it make sense to create a branch and develop this effort there until it's done?

clauswilke:
Yes, that's what I was expecting was going to be problematic when I looked over the code yesterday. At some point, I think we'll have to rethink hard-coded position scales(*), but it's not something I want to tackle now.

(*) For example, in some cases a third dimension would be useful, e.g. to control z-ordering of elements, or to natively implement something like ggtern.

clauswilke:
There is indeed a problem with writing changes back to the coord. If we reuse the coord in multiple plots, limits can get messed up. This could be addressed by cloning the coord when adding it, but I'll first want to see if there's a different approach.

``` r
library(ggplot2)
library(patchwork)

# works as expected
p1 <- ggplot(nc[1, ]) + geom_sf() + coord_sf(crs = 3347)
p2 <- ggplot(nc[2, ]) + geom_sf() + coord_sf(crs = 3347)
p1 + p2
```



``` r

# doesn't work, limits in second plot are wrong
crd <- coord_sf(crs = 3347)
p1 <- ggplot(nc[1, ]) + geom_sf() + crd
p2 <- ggplot(nc[2, ]) + geom_sf() + crd
p1 + p2
```



<sup>Created on 2019-12-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

If we decide the coord needs to be cloned, this would have to happen here:
https://github.com/tidyverse/ggplot2/blob/48660e134982bd6ba6e00bd59b73db6a1ed8132c/R/plot-construction.r#L139-L149

clauswilke:
I was able to avoid interference between different copies of the same coord, by resetting internal parameters to a defined state whenever the plot build begins.

``` r
library(ggplot2)
library(patchwork)

# parameters are regenerated with every plot build, no
# sharing of parameters between plots
crd <- coord_sf(crs = 3347)
p1 <- ggplot(nc[1, ]) + geom_sf() + crd
p2 <- ggplot(nc[2, ]) + geom_sf() + crd
p1 + p2
```



<sup>Created on 2019-12-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
@edzer FYI, with the latest version of my code, if you set `default_crs = NULL` in `coord_sf()` everything should behave exactly as before. But without that option, I assume there will be cases that have issues. The right way to resolve this would likely involve implementing an sf-aware version of `geom_raster()` and/or `geom_rect()`, but I don't understand the `geom_stars()` code well enough to grasp all the implications.

R/stat-sf.R
```diff
@@ -3,12 +3,49 @@
...
+      # register geometric center of each bbox, to give regular scales
+      # some indication of where shapes lie
+      bbox_trans <- sf_transform_xy(
+        list(
+          x = 0.5*(bbox[["xmin"]] + bbox[["xmax"]]),
+          y = 0.5*(bbox[["ymin"]] + bbox[["ymax"]])
+        ),
+        coord$get_default_crs(),
+        geometry_crs
+      )
```
paleolimbot: I probably missed this in the discussion, but is there a reason you're not training the regular scales with the whole extent? The approach I use in ggspatial for rasters is to use `st_make_grid(what = "corners") %>% st_transform() %>% st_bbox()`.

clauswilke: I want the regular scales to be as little involved as possible in determining the final extent, because for very non-linear or rotated coordinate systems taking the extremes in longitude and latitude and turning them into a box and projecting that will lead to an extent that is way too big. At the same time, I need to give the regular scales some numbers to work with, otherwise they'll set their ranges to `c(0, 1)` and that will mess things up even more. 

See the last two examples in this comment:
https://github.com/tidyverse/ggplot2/pull/3659#issuecomment-562818034
Specifically, in the top row of the very last set of plots, the final extent is a bit too large because the regular scales have seen too wide of an extent in longitude and latitude. This effect would be much more severe if we gave them the whole long-lat extent for every geometry.

paleolimbot: I see! Good call.

clauswilke:
@paleolimbot:
> My vote would be to do projection of non-sf layers early (e.g., in Layer$compute_aesthetics() or Stat$compute_layer()). That approach makes it hard to interpret position aesthetics that aren't x or y, but might be much simpler.

I don't think it is technically possible. The only coordinates that can be projected reliably are (x, y) pairs. If you have, for example, `x`, `xmin`, `xmax`, `y`, `ymin`, `ymax`, you can't transform them because you don't necessarily know which x position should be matched with which y position (and possibly, the same x position will get matched with multiple y positions or vice versa).

I'm leaning towards enabling the features in this PR but not turning them on by default. Where I think they will be very helpful is in limit specifications of scales and `coord_sf()`. (Still need to do some work for the latter.) I've found it rather cumbersome in plots I've made that I have to specify limits in projected coordinates, which are entirely unintuitive and difficult to figure out without some external tools. Why shouldn't ggplot2 users be able to specify limits directly in simple long-lat ranges, at least if they flip an appropriate switch?

yutannihilation:
>  Why shouldn't ggplot2 users be able to specify limits directly in simple long-lat ranges, at least if they flip an appropriate switch?

I'm wondering if https://github.com/tidyverse/ggplot2/issues/3698#issuecomment-571678370 is a related issue and I feel it might not be possible to switch between projected and non-projected transparently. (Sorry, this might be just I can't follow the discussion here yet...)

clauswilke:
I think the issue in #3698 would be easy to resolve with the final modification to `coord_sf()` that I'm planning. Just need to sit down and finish it.

yutannihilation:
Got it, thanks.

clauswilke:
I discussed this issue today in person with @hadley, @thomasp85, @dpseidel, and all were generally in favor of this approach. I'll finish this off in the next few days so it can be merged.

clauswilke:
I'm quite happy with this PR now and I think it's ready for review.

@paleolimbot @yutannihilation @dpseidel @thomasp85 If you could try this out and see if you can break things that would be terrific. I'll post a bunch of examples tomorrow, but you can also scroll through this PR and try out the various examples. Most important recent change: Coord limits are now also defined in the default crs, which means long-lat by default.

clauswilke:
As promised, a final set of examples for this PR.

1. Mixing regular geoms and sf geoms.

``` r
library(tidyverse)
library(ggforce)
library(sf)

df <- data.frame(
  lat = 35.76667,
  long = -78.63333,
  name = "Raleigh"
)

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264)
```



``` r

ggplot(nc) +
  geom_sf(size = 0.1, fill = "white") +
  geom_hline(yintercept = 34:36) +
  geom_vline(xintercept = c(-84, -80, -76)) +
  coord_sf(crs = 2264)
```



``` r

df <- read_csv(file = "name,population,lat,long
Charlotte,827097,35.227,-80.843
Raleigh,451066,35.772,-78.639
Greensboro,285342,36.073,-79.792
Durham,257636,35.994,-78.899
Winston-Salem,241218,36.1,-80.244")

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_mark_hull() +
  coord_sf(crs = 2264)
```



``` r

ggplot() +
  geom_sf(data = nc) +
  geom_tile(
    aes(x, y, width = 2, height = 2),
    data = data.frame(x = -79, y = 35),
    alpha = 0.5
  ) +
  coord_sf(crs = 3347)
```



``` r

# coord_sf() now plays nicely with annotation_map()
ggplot() +
  annotation_map(map_data("state"), fill = "antiquewhite", colour = "darkgrey") +
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347))
```



``` r

# limits are set in long-lat
ggplot() +
  annotation_map(map_data("state"), fill = "antiquewhite", colour = "darkgrey") +
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347), xlim = c(-82, -76), ylim = c(34, 37))
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
2. `stat_sf_coordinates()` works. Implementing this stat was slightly tricky in the new framework, because it needs to backtransform to the default crs.

``` r
library(ggplot2)
library(sf)

ggplot(nc) +
  geom_sf() +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = 3347)
```



``` r

ggplot(nc) +
  geom_sf() +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = 3347, default_crs = NULL)
```



``` r

# limits in long-lat
ggplot(nc) +
  geom_sf() +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = 3347, xlim = c(-82, -78))
```



``` r

# limits in projected coordinates
lims <- sf_transform_xy(list(x = c(-82, -78), y = c(36, 36)), 3347, 4326)
lims
#> $x
#> [1] 7149345 7528995
#> 
#> $y
#> [1] 12311.57 83974.19
ggplot(nc) +
  geom_sf() +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = 3347, xlim = lims$x, default_crs = NULL)
```



``` r

# stat_sf_coordinates without geom_sf()
ggplot(nc) +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = 3347)
```



``` r

ggplot(nc) +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = 3347, default_crs = NULL)
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
3. Caveats on limits. Adjusting limits in one direction can have unexpected results for limits
in the other direction. There isn't really any way around this if we want to allow limits in non-projected coordinates. For reliable results, limits need to be specified for both longitude and latitude.

``` r
library(ggplot2)
library(sf)
library(patchwork)

p1 <- ggplot(nc) +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = st_crs(3347))

p2 <- ggplot(nc) +
  stat_sf_coordinates(geom = "point") +
  coord_sf(crs = st_crs(3347), xlim = c(-84.3, -75.8))

p1 / p2
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
4. More caveats on limits. Things can go wonky across the international date line. However, things can generally be made to work with the appropriate projection.

``` r
library(tidyverse)
library(rnaturalearth)
library(rnaturalearthdata)
library(sf)

world <- ne_countries(scale = "medium", returnclass = "sf")
fiji_polys <- world %>%
  filter(name == "Fiji") %>%
  pull(geometry) %>%
  st_cast("POLYGON") %>% 
  st_sf() %>%
  mutate(label = letters[1:n()])

# doesn't work, long-lat across the dateline causes problems 
ggplot(fiji_polys) +
  geom_sf()
```



``` r

# works
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460)
```



``` r

# doesn't work, not mathematically well defined
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460) + 
  ylim(-22, -12)
#> Warning: Projection of x or y limits failed.
#> Consider working in projected coordinates by setting `default_crs = NULL` in `coord_sf()`.
```



``` r

# expansion past the dateline works
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460, xlim = c(177.5, 179.9), ylim = c(-20, -16)) +
  scale_x_continuous(expand = c(.8, 0))
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
5. Statefulness does not bleed through among plots. Finally, even though `coord_sf()` now has a state, to keep track of bounding boxes of sf objects, this state is properly cleared with every rebuild of a plot. It doesn't bleed through to other plots that may be using the same coord object.

``` r
library(ggplot2)
library(patchwork)

# parameters are regenerated with every plot build, no
# sharing of parameters between plots
crd <- coord_sf(crs = 3347)
p1 <- ggplot(nc[1, ]) + geom_sf() + crd
p2 <- ggplot(nc[2, ]) + geom_sf() + crd
p1 + p2
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
I calculate limits in one dimension at the mid-point along the other dimension. That's why the first few examples all have their corners cut off. I could use a different heuristic (e.g., use the maximum and minimum extents in all possible combinations), but that heuristic tends to produce limits that are far too large. We could also make the limits heuristic configurable, but maybe that's overkill.

If we go ahead with this change, there's probably still room for the `geom_spatial_*()` variants, exactly because they would get the limits right in the more obscure cases. Ultimately, the only way to guarantee limits are correct is to calculate them in projected coordinates, and that's not possible with my approach. Note that you'd likely have to modify your `geom_spatial_*()` functions somewhat to play nicely with the new coord version, similar to the changes I made to `stat_sf()` and `stat_sf_coordinates()`.

Any idea what's happening in the ggrepel example?

clauswilke:
Actually, a different and maybe more important question: Do all your ggspatial examples work as is if you set `default_crs = NULL` in `coord_sf()`? They should, I think.

clauswilke:
Ah, ggrepel fails because some of the points are outside the drawing area. If they're all inside ggrepel works fine. And I just had an idea for a better limit heuristic. It's somewhat complicated, though, so not sure if it's worth it.

clauswilke:
I've implemented an alternative limits strategy and have made it configurable, as the one that works for your cases doesn't work well for the North Carolina examples I had been working with. I think it's reasonable to have a few different options.

Note 1: hline and vline don't work properly in highly non-linear coordinate systems. Not sure this can be fixed easily.

Note 2: The ggrepel example still doesn't work, even though now all points are visible.

``` r

library(ggplot2)

# Canada!
tile_df <- expand.grid(
  xmin = seq(-140, -52, by = 20),
  ymin = seq(40, 70, by = 10)
)

tile_df$xmax <- tile_df$xmin + 20
tile_df$ymax <- tile_df$ymin + 10
tile_df$x <- (tile_df$xmin + tile_df$xmax) / 2
tile_df$y <- (tile_df$ymin + tile_df$ymax) / 2
tile_df$width <- 20
tile_df$height <- 10

# rectangles

p <- ggplot(
  tile_df,
  aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax, x = x, y = y)
) +
  coord_sf(crs = 3979, lims_method = "box")


p + geom_rect()
```



``` r
p + geom_tile(aes(height = 7.5, width = 5))
```



``` r

# segments

cities <- data.frame(
  x = c(-63.58595, 116.41214, 13.50, -149.75),
  y = c(44.64862, 40.19063, 52.51, 61.20),
  city = c("Halifax", "Beijing", "Berlin", "Anchorage")
)

cities$xend <- cities$x[c(2, 4, 1, 3)]
cities$yend <- cities$y[c(2, 4, 1, 3)]

p <- ggplot(cities, aes(x, y, xend = xend, yend = yend)) +
  geom_point() +
  # view of the north pole
  coord_sf(crs = 3995, lims_method = "box")

p
```



``` r
p + geom_segment()
```



``` r
p + geom_segment() + coord_sf(crs = 3857)
#> Coordinate system already present. Adding new coordinate system, which will replace the existing one.
```



``` r

p +
  geom_hline(
    aes(yintercept = y, col = city)
  ) +
  geom_vline(
    aes(xintercept = x, lty = city)
  )
```



``` r

p + geom_text(aes(label = city))
```



``` r

p + ggrepel::geom_text_repel(aes(label = city))
#> Error in sf::st_point(c(x, y)): is.numeric(x) is not TRUE
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
I fixed the ggrepel issue.

``` r
library(ggplot2)
library(ggrepel)

cities <- data.frame(
  x = c(-63.58595, 116.41214, 13.50, -149.75),
  y = c(44.64862, 40.19063, 52.51, 61.20),
  city = c("Halifax", "Beijing", "Berlin", "Anchorage")
)

ggplot(cities, aes(x, y)) +
  annotation_map(
    map_data("world"),
    fill = "antiquewhite", colour = "darkgrey", size = 0.1
  ) +
  geom_point() +
  geom_text_repel(aes(label = city), box.padding = unit(20, "pt")) +
  coord_sf(crs = 3995, lims_method = "box")
```



<sup>Created on 2020-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

edzer:
Hi Claus, impressive work! I have a question. `sf` objects have a CRS, and can be transformed to a CRS chosen by ggplot2, or defined in `coord_sf()`. But in the `data.frame` objects with coordinates, like the one in
```r
ggplot() +
  geom_sf(data = nc) +
  geom_tile(
    aes(x, y, width = 2, height = 2),
    data = data.frame(x = -79, y = 35),
    alpha = 0.5
  ) +
  coord_sf(crs = 3347)
```
what is the implicit assumption of the CRS of the `data.frame`, and can this be modified (e.g. when the datum is not WGS84)?

clauswilke:
Edzer, yes, it can be changed. That's why I added the parameter `default_crs` to `coord_sf()`. It sets the crs that is assumed for all data that doesn't come with its own crs (i.e., non-sf data, limits, etc.). I'll prepare an example later today.

clauswilke:
@edzer See examples below. If you set `default_crs = NULL`, it interprets values in non-sf layers as being provided in projected coordinates for the final plot (e.g. as set by `crs`). However, you can choose any combination of crs's. The second example takes data in projected coordinates and then plots as long-lat in WGS84.

@paleolimbot If coord munching is slow, maybe we should look into that separately, since it's being used all over ggplot.

``` r
library(ggplot2)
library(sf)

ggplot() +
  geom_sf(data = nc) +
  geom_tile(
    aes(x, y, width = 240000, height = 200000),
    data = data.frame(x = 7458481, y = -53026),
    alpha = 0.5
  ) +
  coord_sf(crs = 3347, default_crs = NULL)
```



``` r

ggplot() +
  geom_sf(data = nc) +
  geom_tile(
    aes(x, y, width = 240000, height = 200000),
    data = data.frame(x = 7458481, y = -53026),
    alpha = 0.5
  ) +
  coord_sf(crs = 4326, default_crs = 3347)
```



<sup>Created on 2020-02-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

edzer:
And you segmented the sides of the tile before transforming so that they're no longer straight in the second plot, right?

clauswilke:
Edzer, that's how it works, yes. You can see it more clearly in the first example in [this comment.](https://github.com/tidyverse/ggplot2/pull/3659#issuecomment-581242176)

edzer:
As @paleolimbot's examples show, it's a tough decision in which space to draw straight lines; I addressed this problem here: https://edzer.github.io/sdsc19/#16 . Somehow, to me, unprojected coordinates are more naturally seen as "on a sphere", where great circles seem most natural because rotation invariant. Then, the entire GIS legacy essentially lives in equidistant rectangular - draw lines in lon/lat=x/y=2D.

clauswilke:
Edzer, that's what my implementation would do by default, correct? Take unprojected coordinates and interpret them as lying on a sphere.

Any other option is possible, e.g. users can always work in projected coordinates or explicitly with sf objects, but that may require more knowledge or thought. My goal is to implement an API that creates reasonable results with minimal effort but allows the expert to do whatever they may want to do.

clauswilke:
Ah, now I get it, I'm doing equirectangular rather than equidistant rectangular. Well, for non-sf layers, I think that's the only way we can do it without adding even more overhead. For GIS nerds that care about such details, they should probably work exclusively with sf layers and projected coordinates.

edzer:
Yes, sorry for confusing, I should have said "cylindrical equidistant" (Iliffe & Lott), or "equirectangular" (wikipedia) rather than "equidistant rectangular" (which doesn't seem to exist). The point is: you interpolate points on R2, and not on S2. It has nothing to do with nerdiness, but with solutions that break on antimeridians and poles (R2), or not (S2). Work on this (in sf) is planned to start soon (see [here](https://github.com/r-spatial/global/blob/master/proposal/isc-proposal.pdf)).

clauswilke:
Yes, I get it now. In any case, for non-sf geoms, the segmentation is actually handled by the geoms themselves (that's just how ggplot2 works), so they'll always create straight lines in whatever coordinate system they were given the input data. I don't see any technical way around this at this time. I think the best way to proceed is to add a sentence or two about this to the documentation, and to recommend that people pre-segment their straight lines if the auto-segmentation produces undesirable results.

clauswilke:
@edzer I wrote expanded documentation that includes a paragraph about straight lines:
https://github.com/tidyverse/ggplot2/pull/3659/commits/c5dd56a1fe88569047e90be417addbd8b39a0f1e
Could you look it over and let me know if this seems reasonable?

@thomasp85 @yutannihilation Would you mind reviewing the code?

edzer:
@clauswilke I think the documentation is good.

R/stat-sf.R
```diff
@@ -3,12 +3,52 @@
...
+        xmin = bbox[["xmin"]], xmax = bbox[["xmax"]],
+        ymin = bbox[["ymin"]], ymax = bbox[["ymax"]]
+      )
+
+      # to represent the location of the geometry in default coordinates,
+      # we take the mid-point along each side of the bounding box and
+      # backtransform
+      bbox_trans <- sf_transform_xy(
+        list(
+          x = c(rep(0.5*(bbox[["xmin"]] + bbox[["xmax"]]), 2), bbox[["xmin"]], bbox[["xmax"]]),
```
yutannihilation: Maybe it would be easier to read if we use `$` instead of `[[`?

edzer: if `bbox` comes from `sf::st_bbox`, it is a numeric vector, not a list.

yutannihilation:
I still don't think I understand the details of this change, but generally looks good to me!

One corner case I've come up with my mind is that the case when users mistakenly mix some discrete values on the sf-coordinates. I think we can ignore this and address in a separate issue because it doesn't work at the moment either, but let me comment here quickly.

``` r
library(ggplot2)

nc <- sf::read_sf(system.file("shape/nc.shp", package="sf"))
wrong_df <- data.frame(x = "a")

# v3.3.0rc
ggplot(nc) +
  geom_sf() +
  geom_point(aes(x, x), wrong_df)
```



``` r
# this PR
#> Loading ggplot2
ggplot(nc) +
  geom_sf() +
  geom_point(aes(x, x), wrong_df)
#> Warning in mean.default(xlim): argument is not numeric or logical: returning NA
#> Warning in seq.default(xlim[1], xlim[2], length.out = 20): NAs introduced by
#> coercion
#> Error in seq.default(xlim[1], xlim[2], length.out = 20): 'from' must be a finite number
```

R/stat-sf.R
yutannihilation: Oh, thanks. But this is what I'm seeing...:

``` r
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.2.1

a <- st_sf(a = 1:2, geom = st_sfc(st_point(0:1), st_point(1:2)), crs = 4326)
bbox <- st_bbox(a)
is(bbox)
#> [1] "bbox"
bbox$xmin
#> xmin 
#>    0
```

<sup>Created on 2020-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

edzer: You are right - a case of over-engineering from my side...

clauswilke: So is this official API I can rely on? Should I change the code from `[[` to `$` as suggested?

R/coord-sf.R
```diff
@@ -408,15 +617,16 @@ sf_rescale01_x <- function(x, range) {
...
 #' @inheritParams coord_cartesian
 #' @export
 #' @rdname ggsf
 coord_sf <- function(xlim = NULL, ylim = NULL, expand = TRUE,
-                     crs = NULL, datum = sf::st_crs(4326),
+                     crs = NULL, default_crs = sf::st_crs(4326),
+                     datum = sf::st_crs(4326),
                      label_graticule = waiver(),
-                     label_axes = waiver(),
+                     label_axes = waiver(), lims_method = "cross",
```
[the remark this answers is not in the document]
clauswilke: Thanks, will do.

clauswilke:
@yutannihilation The error you see with your example is generated inside of this function, which tries to construct a bounding box for axis limits:
https://github.com/clauswilke/ggplot2/blob/98433331298cefd10de3c9b395bf56dbf51a96e2/R/coord-sf.R#L534-L562

The simplest solution seems to be to run `is.finite()` on the input and error out if any values are found that are not finite. The error message could be something like: "Scale limits cannot be mapped onto spatial coordinates. Are you mapping the right data columns to the `x` and `y` aesthetics?"

Does this sound reasonable to you, or do you think the code should recover and produce something like your example from 3.3.0rc?

clauswilke:
How's this?

``` r
library(ggplot2)

nc <- sf::read_sf(system.file("shape/nc.shp", package="sf"))
wrong_df <- data.frame(x = "a")

ggplot(nc) +
  geom_sf() +
  geom_point(aes(x, x), wrong_df)
#> Error: Scale limits cannot be mapped onto spatial coordinates. Are you mapping the right data columns to the `x` and/or `y` aesthetics?
```



<sup>Created on 2020-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/stat-sf.R
yutannihilation: Sorry, though this seems convenient, now I don't feel it's very readable to use `$` for subsetting an element from a numeric vector. Please ignore, I just mistook `bbox` for a list, guessing from `[[`.

yutannihilation:
> Does this sound reasonable to you, or do you think the code should recover and produce something like your example from 3.3.0rc?

Thanks, I think the code should fail. The error message sounds fine, but this part feels a bit cryptic to me.

> Are you mapping the right data columns to the `x` and `y` aesthetics?

Maybe we should show some hints about what data is "right"? Let me think a bit...

clauswilke:
> Maybe we should show some hints about what data is "right"? Let me think a bit...
The tricky bit is that it's difficult to enumerate all ways in which data could be wrong. E.g., I could write "Are you mapping character values or a factor to the `x` and/or `y` aesthetics", but that would leave out dates and whatever else people might incorrectly map.

clauswilke:
Alternatively, I could not give a hint at all, and just leave it at "Scale limits cannot be mapped onto spatial coordinates." Maybe that's the better approach.

yutannihilation:
> The tricky bit is that it's difficult to enumerate all ways in which data could be wrong.

Yeah, I understand it's difficult... So, giving no hint at all sounds reasonable to me, at least at the point of this PR.

hadley:
@clauswilke would you mind writing a quick overview of the PR, updated based on the discussion so far? Otherwise it's going to be hard for me to find the time to read the whole thread.

clauswilke:
@hadley Sure. Give me a day or two. I'll also resolve the merge conflicts. I'll ping you when I'm ready.

mdsumner:
gee this is fab only just found it, thanks @clauswilke !!!!!!!!!!

clauswilke:
@hadley As promised, a summary of the discussion for this PR. I'll write this in a series of comments, so that I don't have to craft one massive comment all at once.

First, what does this PR do? From the news item:

> `coord_sf()` now has an argument `default_crs` that specifies the coordinate reference system (CRS) for non-sf layers and scale/coord limits. This argument defaults to the World Geodetic System 1984 (WGS84), which means x and y positions are interpreted as longitude and latitude. This is a potentially breaking change for users who use projected coordinates in non-sf layers or in limits. Setting `default_crs = NULL` recovers the old behavior. Further, authors of extension packages implementing `stat_sf()`-like functionality are encouraged to look at the source code of `stat_sf()`'s `compute_group()` function to see how to provide scale-limit hints to `coord_sf()` (@clauswilke, #3659).

Here are some basic examples. We can now easily add points or text in long-lat coordinates.

``` r
library(tidyverse)
library(sf)

df <- data.frame(
  lat = 35.76667,
  long = -78.63333,
  name = "Raleigh"
)

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_text(aes(label = name), hjust = -.1, vjust = 1.1) +
  coord_sf(crs = 2264)
```



The framework is general and extends to essentially any geom. Here is an example with `geom_mark_hull()` from ggforce.

``` r
library(ggforce)

df <- read_csv(file = "name,population,lat,long
Charlotte,827097,35.227,-80.843
Raleigh,451066,35.772,-78.639
Greensboro,285342,36.073,-79.792
Durham,257636,35.994,-78.899
Winston-Salem,241218,36.1,-80.244")

ggplot(df, aes(x = long, y = lat)) +
  geom_sf(data = nc, size = 0.1, fill = "white", inherit.aes = FALSE) +
  geom_point() +
  geom_mark_hull() +
  coord_sf(crs = 2264)
```



And here is an example from `annotation_map()`, which uses `geom_polygon()` to draw maps and until now hasn't played nicely with `coord_sf()`.

``` r
ggplot() +
  annotation_map(map_data("state"), fill = "antiquewhite", colour = "darkgrey") +
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347))
```



The limits (both for scales and for the coord) are set in the default coordinate system, i.e., long-lat by default.

``` r
ggplot() +
  annotation_map(map_data("state"), fill = "antiquewhite", colour = "darkgrey") +
  geom_sf(data = nc) +
  coord_sf(crs = st_crs(3347), xlim = c(-82, -76), ylim = c(34, 37))
```



<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
While I developed this PR a number of issues came up, which I'll list here and then address one by one:

1. We have to keep track of two different coordinate systems, the default one (`default_crs`) and the one into which we're projecting (`crs`). This breaks some assumptions in standard ggplot2.

2. We have to appropriately segmentize straight lines.

3. Setting limits in default coordinates poses unique challenges, because the coordinates are not cartesian after projection.

4. In general, things can get wonky in weird coordinates, e.g. when we want to plot a pole or the international date boundary.

5. Extension developers will need to write more complex code to write sf-aware geoms and stats. This requires additional helper functions.

clauswilke:
**Point 1: Two different coordinate systems**

For geometry objects, we want to keep track of their correct limits in projected coordinates. But scales are now trained in default coordinates. Thus, we somehow have to keep track of projected limits outsides of the regular scales. I handled this by implementing a call-back to `coord_sf()` to record the bounding boxes of all geometry objects:
https://github.com/clauswilke/ggplot2/blob/e4cdcd67498e53885a3124b13959acf5aef53560/R/stat-sf.R#L18-L24

Ultimately, the correct way to handle this would be to implement generic position scales, so that we could add a geometry scale that can be trained on geometry objects. However, that solutions is out of reach at this time.

clauswilke:
**Point 2: Appropriately segmentize straight lines.**

I'm segmentizing lines in the default coordinate system and then transform the segments. With the default setting, this generates horizontal and vertical lines that follow parallels and meridians, a behavior that many people might expect:

``` r
library(tidyverse)
library(sf)

ggplot(nc) +
  geom_sf(size = 0.1, fill = "white") +
  geom_hline(yintercept = 34:36) +
  geom_vline(xintercept = c(-84, -80, -76)) +
  coord_sf(crs = 2264)
```



<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

However, @edzer pointed out that this is not always the correct approach, and that we might want to segmentize, for example, along great circles. Doing this is beyond the scope of this PR though (and likely beyond the scope of a simple ggplot2 coord), so I addressed this issue by pointing it out in the documentation: 
https://github.com/clauswilke/ggplot2/blob/e4cdcd67498e53885a3124b13959acf5aef53560/R/geom-sf.R#L43-L51

clauswilke:
**Point 3: Limits**

Specifying limits in default coordinates (either through the min/max values in the data, or via `xlim` etc.) is complicated, because how these limits translate into regions in projected coordinates can be unintuitive, in particular for very curved coordinate systems. There is no algorithm that always works well, so I implemented a couple of different approaches that have different strengths and weaknesses. The default choice is relative conservative, i.e., it is unlikely to make the limits too large, but it tends to make limits too narrow for very non-linear projections.

``` r
library(ggplot2)

tile_df <- expand.grid(
  xmin = seq(-140, -52, by = 20),
  ymin = seq(40, 70, by = 10)
)

tile_df$xmax <- tile_df$xmin + 20
tile_df$ymax <- tile_df$ymin + 10
tile_df$x <- (tile_df$xmin + tile_df$xmax) / 2
tile_df$y <- (tile_df$ymin + tile_df$ymax) / 2
tile_df$width <- 20
tile_df$height <- 10

p <- ggplot(
  tile_df,
  aes(xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax, x = x, y = y)
) + geom_tile(aes(height = 7.5, width = 5))

p + coord_sf(crs = 3979) # default limit method is "cross"
```



``` r
p + coord_sf(crs = 3979, lims_method = "box")
```



``` r
p + coord_sf(crs = 3979, lims_method = "orthogonal")
```



<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
**Point 4: Things can get wonky in weird coordinate systems**

The reason for why was already explained under Points 2 and 3. Here are a few more examples with limits:
``` r
library(tidyverse)
library(rnaturalearth)
library(rnaturalearthdata)
library(sf)

world <- ne_countries(scale = "medium", returnclass = "sf")
fiji_polys <- world %>%
  filter(name == "Fiji") %>%
  pull(geometry) %>%
  st_cast("POLYGON") %>% 
  st_sf() %>%
  mutate(label = letters[1:n()])

# doesn't work, long-lat across the dateline causes problems 
ggplot(fiji_polys) +
  geom_sf()
```



``` r

# doesn't work, due to dateline boundary
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460)
```



``` r

# setting `default_crs = NULL` always fixes things
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460, default_crs = NULL)
```



``` r

# changing the limits method also works
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460, lims_method = "geometry_bbox")
```



``` r

# works
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460, lims_method = "orthogonal")
```



``` r

# works (not sure why)
ggplot(fiji_polys) +
  geom_sf() +
  coord_sf(crs = 3460, lims_method = "orthogonal") + 
  xlim(-170, 170) +
  ylim(-22, -12)
```



<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

And now an example with straight lines. These plots probably don't look the way we would want them to.

``` r
library(ggplot2)

cities <- data.frame(
  x = c(-63.58595, 116.41214, 13.50, -149.75),
  y = c(44.64862, 40.19063, 52.51, 61.20),
  city = c("Halifax", "Beijing", "Berlin", "Anchorage")
)

cities$xend <- cities$x[c(2, 4, 1, 3)]
cities$yend <- cities$y[c(2, 4, 1, 3)]

p <- ggplot(cities, aes(x, y, xend = xend, yend = yend)) +
  geom_point() +
  # view of the north pole
  coord_sf(crs = 3995, lims_method = "box")

p + geom_segment()
```



``` r

p +
  geom_hline(
    aes(yintercept = y, col = city)
  ) +
  geom_vline(
    aes(xintercept = x, lty = city)
  )
```



<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

However, surprisingly, the following works just fine.

**Note: this does not use `geom_sf()` at all, only `coord_sf()`!**

``` r
library(ggplot2)
library(ggrepel)

cities <- data.frame(
  x = c(-63.58595, 116.41214, 13.50, -149.75),
  y = c(44.64862, 40.19063, 52.51, 61.20),
  city = c("Halifax", "Beijing", "Berlin", "Anchorage")
)

ggplot(cities, aes(x, y)) +
  annotation_map(
    map_data("world"),
    fill = "antiquewhite", colour = "darkgrey", size = 0.1
  ) +
  geom_point() +
  geom_text_repel(aes(label = city), box.padding = unit(20, "pt")) +
  coord_sf(crs = 3995, lims_method = "box")
```



<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
**Point 5: Additional helper functions**

This point applies primarily to a function I needed to write that makes it easy to transform coordinates among different coordinate systems, without having to worry about creating and then unpacking sf objects. I called this function `sf_transform_xy()`:
https://github.com/clauswilke/ggplot2/blob/e4cdcd67498e53885a3124b13959acf5aef53560/R/coord-sf.R#L470-L517

This is how it works:
``` r
library(ggplot2)

# location of cities in NC by long (x) and lat (y)
data <- data.frame(
  city = c("Charlotte", "Raleigh", "Greensboro"),
  x =  c(-80.843, -78.639, -79.792),
  y = c(35.227, 35.772, 36.073)
)

# transform to projected coordinates
data_proj <- sf_transform_xy(data, 3347, 4326)
data_proj
#>         city       x         y
#> 1  Charlotte 7275499 -60169.66
#> 2    Raleigh 7474260  44384.13
#> 3 Greensboro 7357835  57438.27

# transform back
sf_transform_xy(data_proj, 4326, 3347)
#>         city       x      y
#> 1  Charlotte -80.843 35.227
#> 2    Raleigh -78.639 35.772
#> 3 Greensboro -79.792 36.073
```

<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This function is not only useful for extension writers, it can also be useful to endusers who want to, e.g., calculate plot limits in projected coordinates. However, it is somewhat outside the normal scope of what ggplot2 does. So, two questions:

1. Are we Ok exporting this or should I keep it internal for now?
2. If we export it, is the current name fine, or should I change it?

This was the last comment in this series. I have addressed everything I could think of.

clauswilke:
Thanks! I think I'll wait merging this until after the 3.3.1 release. It's too much of a change for a minor release.

I agree with the documentation issues. I've also thought about breaking the man page into two, one for `geom_sf()` and one for `coord_sf()`, because it's starting to get confusing which parts of the text belong to which function.

clauswilke:
This is now merged. Thanks to everybody for the extensive feedback. Let's see how this plays out as people start testing this in the wild.

Issue #3468, opened 2019-08-03
print method for scales does not handle function as limits argument well


teunbrand:
I've found a very minor inconvenience when printing scales. When passing a function to the limits arguments in a `scale_*` , the printing method of the scale throws an error. See reprex below.

``` r
library(ggplot2)
g <- ggplot(iris, aes(Sepal.Width, Sepal.Length,
                      colour = Petal.Width)) +
  geom_point()

goodprint <- g + scale_colour_continuous(limits = range(iris$Petal.Width))
badprint <- g + scale_colour_continuous(limits = function(lims) {1 * lims})

print(goodprint$scales$scales[[1]])
#> <ScaleContinuous>
#>  Range:  
#>  Limits:  0.1 --  2.5

print(badprint$scales$scales[[1]])
#> <ScaleContinuous>
#>  Range:
#> Error in zero_range(range): x must be length 1 or 2
```

<sup>Created on 2019-08-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

While not really a problem for most applications, it would take only a minor adjustment to the printing function to fix the error. Below you'll find an option for a fix.

```r
scale_print <- function(self, ...) {
  # Same as current
  show_range <- function(x) {
      paste0(formatC(x, digits = 3), collapse = " -- ")
    }
  cat("<", class(self)[[1]], ">\n", sep = "")
  cat(" Range:  ", show_range(self$range$range), "\n", sep = "")
  # Only the following bit is different
  if (is.function(self$limits)) { 
    cat(" Limits: function()\n")
  } else {
    cat(" Limits: ", show_range(self$dimension()), "\n", sep = "")
  }
}

badprint$scales$scales[[1]]$print <- scale_print

print(badprint$scales$scales[[1]])
#> <ScaleContinuous>
#>   Range:  
#>   Limits: function()
```

paleolimbot:
This no longer errors, but @teunbrand's option is more informative:

``` r
library(ggplot2)
g <- ggplot(iris, aes(Sepal.Width, Sepal.Length,
                      colour = Petal.Width)) +
  geom_point()

goodprint <- g + scale_colour_continuous(limits = range(iris$Petal.Width))
badprint <- g + scale_colour_continuous(limits = function(lims) {1 * lims})

print(goodprint$scales$scales[[1]])
#> <ScaleContinuous>
#>  Range:  
#>  Limits:  0.1 --  2.5
print(badprint$scales$scales[[1]])
[...]

PR #3664, opened 2019-12-12
Add trans field to ScaleContinuous

thomasp85:
The ScaleContinuous implementation has never specified a `trans` field, despite its methods referencing the field. This has not really been a problem as it is always subclassed, but recent changes mean that it's `print` method will throw an error now.

This PR adds an identity trans object as it's trans field and makes the error go away


yutannihilation:
Is this related? #3468

PR #3666, opened 2019-12-12
Better handle position aesthetics in params

thomasp85:
revdepchecks uncovered an edge case in direction detection if position aesthetics are given outside aes(). This PR correctly looks in both data and params when figuring out direction

### DIFF
--- a/R/geom-linerange.r
+++ b/R/geom-linerange.r
@@ -99,7 +99,7 @@ GeomLinerange <- ggproto("GeomLinerange", Geom,
 
   setup_params = function(data, params) {
     params$flipped_aes <- has_flipped_aes(data, params, range_is_orthogonal = TRUE)
-    if (!(params$flipped_aes || all(c("x", "ymin", "ymax") %in% names(data)))) {
+    if (!(params$flipped_aes || all(c("x", "ymin", "ymax") %in% c(names(data), names(params))))) {
       stop("Either, `x`, `ymin`, and `ymax` or `y`, `xmin`, and `xmax` must be supplied", call. = FALSE)
     }
     params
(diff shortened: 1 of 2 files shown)


R/geom-linerange.r
```diff
@@ -99,7 +99,7 @@ GeomLinerange <- ggproto("GeomLinerange", Geom,
 
   setup_params = function(data, params) {
     params$flipped_aes <- has_flipped_aes(data, params, range_is_orthogonal = TRUE)
-    if (!(params$flipped_aes || all(c("x", "ymin", "ymax") %in% names(data)))) {
+    if (!(params$flipped_aes || all(c("x", "ymin", "ymax") %in% c(names(data), names(params))))) {
```
clauswilke: Just wondering why it tests for `x`, `ymin`, `ymax` but not `y`, `xmin`, `xmax` in `params`. Is this correct?

thomasp85: if `has_flipped_aes()` returns true it means that it contains y, xmin, and xmax

clauswilke: Ah, Ok, maybe add a quick comment.

R/geom-linerange.r
```diff
@@ -113,6 +113,7 @@ GeomLinerange <- ggproto("GeomLinerange", Geom,
   },
 
   draw_panel = function(data, panel_params, coord, flipped_aes = FALSE) {
+    browser()
```
clauswilke: This looks left over. That's probably why the CI fails.

thomasp85: Oh shit — thanks😅

PR #3672, opened 2019-12-16
Rename size aesthetic for line-based geoms

thomasp85:
This is a draft PR that seeks to rename the `size` aesthetic whenever it refers to the width of a stroke. There are two problems with the current setup:

1. It conflates two distinct aesthetics making it difficult to control each one separately when working with geoms that contain both points and lines/polygons, e.g. boxplot, linerange, and sf.

2. As the default size scale is area based, scaling of stroke width is not linear by default, and it is needed to use `scale_radius()` to change that which is super counterintuitive

For the most part we can change this without backward compatibility issues as `size` can be renamed to `linewidth` if present in `setup_data()`. A few geoms have need for both as discussed in 1.) and we can not do any inference there meaning that there might be visual breakage in these cases. The pros outweigh the cons considerably though (IMO).

The aesthetic name `linewidth` has been chosen as both `width` and `stroke` is already in use. `linewidth` also falls in line with the `linetype` aesthetic.

Please use this PR for discussions


thomasp85:
Team - I would like to get this into the next release as it will be primarily focused on cleaning up code, syntax, and internal, and I feel the current behaviour is a major logical issue... Are there any objections to this change in API?

clauswilke:
Will this require all extension packages to make changes to their code? I like the idea conceptually, but I think we need to think carefully about how the transition will work. If downstream packages need to change their code, can they easily support the new and the old ggplot2 at the same time?

thomasp85:
That is a good question to bring up... This will not break any extension packages (I think), but it will also not magically transform them into understanding `linewidth` as an aesthetic. It is not possible to do an automatic renaming since `size` is still a valid aesthetic in some cases...

So there will be an API consistency issue between ggplot2 and the extension packages in terms of aesthetic naming until they are updated.

However, I think our dev community has generally reacted quite fast on these types of updates

thomasp85:
The implementation of this has now been improved significantly from the point-of-view of updating geoms and ensuring a smooth transition.

In order to upgrade a geom implementation to use `linewidth` instead of `size` all you need to do is to change `size` to `linewidth` in the `default_aes` field, and set the `rename_size` field to TRUE. Further, you of course need to update your code wherever it refers to `size` and change it to `linewidth`

What you get for free when setting `rename_size = TRUE` is that ggplot2 will automatically rename any use of the size aesthetic on the fly, inherit `size` from the global mapping if no `linewidth` mapping exist, allow `scale_size()` to be used with the `size` aesthetic in the geom etc. It will further emit a message to update call to use `linewidth`. I'm fairly certain this means that there won't be much code breaking and that we can gently coerce users into using the new aesthetic, while also making it very easy for extension developers to update their geoms.

thomasp85:
I will go through adding unit tests for all the affected geoms etc, but before that, I'd like some input about the current approach @hadley @clauswilke @yutannihilation

R/layer.r
```diff
@@ -125,6 +125,12 @@ layer <- function(geom = NULL, stat = NULL,
 
   # Warn about extra params and aesthetics
   extra_param <- setdiff(names(params), all)
+  # Take care of size->linewidth renaming in layer params
+  if (geom$rename_size && "size" %in% extra_param && !"linewidth" %in% mapped_aesthetics(mapping)) {
+    aes_params <- c(aes_params, params["size"])
+    extra_param <- setdiff(extra_param, "size")
+    cli::cli_inform("Using {.arg size} as {.arg linewidth}")
```
hadley, line 132: I think this message needs to be a bit more clear that you should now use `linewidth` instead of `size`.

hadley (approved):
LGTM!

R/layer.r
thomasp85, line 132: Set frequency

R/geom-sf.R
```diff
@@ -200,10 +201,15 @@ sf_grob <- function(x, lineend = "butt", linejoin = "round", linemitre = 10,
   fill <- x$fill %||% defaults$fill[type_ind]
   fill <- alpha(fill, alpha)
   size <- x$size %||% defaults$size[type_ind]
-  point_size <- ifelse(is_collection, x$size %||% defaults$point_size[type_ind], size)
+  linewidth <- x$linewidth %||% defaults$linewidth[type_ind]
+  point_size <- ifelse(
+    is_collection,
+    x$size %||% defaults$point_size[type_ind],
+    ifelse(is_point, size, linewidth)
+  )
```
yutannihilation, line 209: I might be wrong, but I think this can be left as is. `point_size` is used only for calculating `fontsize`, so `size` is appropriate here, isn't it?

R/scale-linewidth.R
```diff
@@ -0,0 +1,85 @@
+#' Scales for line width
+#'
+#' `scale_linewidth` scales the width of lines and polygon strokes. Due to
+#' historical reasons, it is also possible to control this with the `size`
+#' aesthetic, but using `linewidth` is encourage to clearly differentiate area
+#' aesthetics from stroke width aesthetics.
```
yutannihilation, line 6: I'm not sure if there are any valid usage, but maybe the document or NEWS should describe how to mimic `scale_size_area()` for linewidth?

Also, probably the document on `scale_size()`'s side should be modified as well.

yutannihilation (commented):
Looks great in general, though I'm not yet sure what this change will mean to the extension packages. Minor comments.

R/scale-linewidth.R
thomasp85: I don't think we should add that - line width scaling is one-dimensional and adding anything describing it differently will just confuse users.

I'll have a look at the `scale_size()` docs

R/geom-sf.R
thomasp85: No, this is necessary to avoid erroring

R/layer.r
thomasp85: How is this?
```
> ggplot(airquality) + geom_line(aes(Day, Ozone, group = Month, size = Month))
Renaming size aesthetic to linewidth
ℹ linewidth now has its own aesthetic instead of conflating it with size
ℹ While you can continue to use size it is highly advices to adapt your code to the new aesthetic
```

R/scale-linewidth.R
thomasp85: Hmm, what do you propose for the `scale_size()` docs - I don't think it contains anything wrong... Just a short notice about using `scale_linewidth()` for strokes?

R/geom-sf.R
yutannihilation: I see.

R/scale-linewidth.R
yutannihilation: > I don't think we should add that - line width scaling is one-dimensional and adding anything describing it differently will just confuse users.

I see. I was thinking about those who need backward-compatibility, but such users can probably keep using `scale_size()`, so please ignore.

>  what do you propose for the `scale_size()` docs - I don't think it contains anything wrong... Just a short notice about using `scale_linewidth()` for strokes?

Yes, something like that. Since `size` can still be used for strokes after this pull request gets merged, I think we should clarify when to use which. I think `scale_size()` is one of the good place for such a notice.

thomasp85: yeah, backwards compatibility would be insisting on using a bad scaling for the aesthetic so I think it's ok to not make that easy

yutannihilation (approved):
Thanks for adding a note. Looks good!

yutannihilation:
Just curious, will this eventually change the behavior of `size` on Geoms that actually has both `size` and `linewidth`? I know there's no breaking change here, but #4872 made me think this might need some consideration to avoid ambiguity. 

For example, in the case of `geom_pointrange()`, it's not clear what to do when we want to increase only the size of the points. Does specifying only `size` achieve this? Or, will this increase both the size of the points and that of the lines? I feel the former behavior is right, but it would be a breaking change.

``` r
geom_pointrange(size = 10)
```

thomasp85:
At this point the ambiguous geoms will not differentiate but I think we would do that down the line after proper deprecation

Issue #3607, opened 2019-11-04
Allow PositionStack to specify the directions on negatives and positives seperately


yutannihilation:
Context: I need this to finish https://github.com/tidyverse/ggplot2/pull/3529#issuecomment-542493052

`PositionStack` allows to flip the stacking direction by `reverse` parameter, but it forces the data on the positive side and on the negative side have the same direction.

https://github.com/tidyverse/ggplot2/blob/528a3740e60520194e3bf4546a5b93e19e37676b/R/position-stack.r#L191-L204

However, it's not a rare case we want to stack them in the opposite direction (i.e. reverse the direction only on the negative side) to draw data properly. `geom_area()` is the very example for this. I wonder if `PositionStack` can accept these.

For the interface, I don't come up with a nice idea yet, but it would be something like:

``` r
geom_area(
  position = position_stack(reverse = c(positive = FALSE, positive = TRUE))
)
```

or

``` r
geom_area(
  position = position_stack(reverse = TRUE, mirrored = TRUE)
)
```

yutannihilation:
Superseded by #3673

PR #3673, opened 2019-12-16
Don't switch relationship of ymin/ymax during stacking

thomasp85:
This PR addresses the problem uncovered in #3529 where position_stack changes the relation of ymin/ymax during stacking, making it difficult to recover the correct line to draw. It is less drastic than what #3607 proposes, and simply ensure minimal changes to the incoming data

### DIFF
--- a/R/position-stack.r
+++ b/R/position-stack.r
@@ -221,10 +221,11 @@ pos_stack <- function(df, width, vjust = 1, fill = FALSE) {
   } else {
     max_is_lower <- rep(FALSE, nrow(df))
   }
-  df$ymin <- pmin(heights[-n], heights[-1])
-  df$ymax <- pmax(heights[-n], heights[-1])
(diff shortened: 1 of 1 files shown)


R/position-stack.r
```diff
@@ -216,10 +216,16 @@ pos_stack <- function(df, width, vjust = 1, fill = FALSE) {
   if (fill) {
     heights <- heights / abs(heights[length(heights)])
   }
-
-  df$ymin <- pmin(heights[-n], heights[-1])
-  df$ymax <- pmax(heights[-n], heights[-1])
-  df$y <- (1 - vjust) * df$ymin + vjust * df$ymax
+  if (!is.null(df$ymin) && !is.null(df$ymax)) {
```
thomasp85:
```suggestion
# We need to preserve ymin/ymax order. If ymax is lower than ymin in input, it should remain that way
  if (!is.null(df$ymin) && !is.null(df$ymax)) {
```

PR #3676, opened 2019-12-18
Regenerate docs after #3579

yutannihilation:
Not sure why, but I see this diff when I regenerate docs on the current master. Probably there's some changes on roxygen's behaviour after #3579...?


yutannihilation:
It seems this is already fixed by https://github.com/tidyverse/ggplot2/commit/47a06183c4aa1fe8c4d8c68921c148e6fb9e0cdd.

PR #3680, opened 2019-12-19
Restore data order after collide

### LINKED ISSUE #3498: geom_area not working with negative values and two categories in 3.2.1
---

I am using geom_area for stacked area chart showing cumulative results for categories.
For some series the cumulative result turns negative from time to time and this have been working nicely with versions of ggplot2 until 3.2.0 was released. Some things have been fixed with the 3.2.1 release but I experince an issue when I have more than one category and using the category for the fill. I notice an ealier issue #3390 with similar problem but that has been closed now or ?

```r
 packageVersion("ggplot2")
[1] ‘3.2.1’

# with one serie  (category A) it works as I expect

date <- as.Date(c("2019-07-31","2019-08-01","2019-08-02","2019-08-03"))
category <-rep(c("A"),4)
variable <-rep(c("cum_result"),4)
value <- c(0,2,3,-4)
data<- data.frame(date,category,variable,value)

ggplot(data, aes(x=date,y=value)) +
  geom_area(aes(x=date,y=value,fill=category),stat = "identity")
```

image

```r
# addding a additional category B with values on the variable  shows the issue

date <- as.Date(c("2019-07-31","2019-07-31","2019-08-01","2019-08-01","2019-08-02","2019-08-02","2019-08-03","2019-08-03"))
category <-rep(c("A","B"),4)
variable <-rep(c("cum_result"),8)
value <- c(0,0,
           2,1,
           3,6,
           -4,3)

data<- data.frame(date,category,variable,value)

ggplot(data, aes(x=date,y=value)) +
  geom_area(aes(x=date,y=value,fill=category),stat = "identity")
```

image

thomasp85:
Fix #3498 This seems to be the last leftover from moving the area/ribbon reorder into setup_data like the rest of the geoms.

This change make sure that data returned from `collide()` is in the same order is it came in.

I think a general rule should be that positions should never change the order of data, only the content. There are situations where this is not possible, but as a general rule of thumb I think it is sane

### DIFF
--- a/R/position-collide.r
+++ b/R/position-collide.r
@@ -38,10 +38,11 @@ collide <- function(data, width = NULL, name, strategy,
   # Reorder by x position, then on group. The default stacking order reverses
   # the group in order to match the legend order.
   if (reverse) {
-    data <- data[order(data$xmin, data$group), ]
+    ord <- order(data$xmin, data$group)
   } else {
-    data <- data[order(data$xmin, -data$group), ]
+    ord <- order(data$xmin, -data$group)
   }
+  data <- data[ord, ]
 
   # Check for overlap
   intervals <- as.numeric(t(unique(data[c("xmin", "xmax")])))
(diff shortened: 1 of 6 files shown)


yutannihilation:
Btw, a side question, even after this (and #3673), stacking negative values still doesn't seem very nice (the blue area at period 4-5 below) though I believe this is just the same as what we saw before version 3.2.0. Do you think there's still a chance to fix this? Or, should this be treated as the known limitation of `geom_ribbon()`? Sorry, I'm behind the discussion around this issue...

``` r
library(ggplot2)

d <- data.frame(
  x = c(1:5, 1:5),
  y = c(1, 2, 0, -1, -2, 3, 2, 0, 1, -1),
  g = rep(c("a", "b"), each = 5)
)

ggplot(d, aes(x, y)) +
  geom_area(aes(fill = g), alpha = 0.3, position = "stack")
```



<sup>Created on 2019-12-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #3685, opened 2019-12-22
Fix limits argument of scale_{colour,fill}_manual()

### LINKED ISSUE #3262: scale_{colour,fill}_manual(limits = ...) doesn't limit the scale
(I found this during looking at the examples on [`scale_manual`](https://ggplot2.tidyverse.org/reference/scale_manual.html#examples).)

It seems `scale_{colour,fill}_manual(limits = ...)` doesn't limit the scale actually (but limit the legend).

``` r
library(ggplot2)
library(patchwork)

p <- ggplot(mtcars, aes(factor(cyl))) +
  geom_bar(aes(fill = factor(cyl)))

# this drops 6
p1 <- p + scale_fill_discrete(limits = c("4", "8"))

# this doesn't
cols <- c("8" = "red", "4" = "blue", "6" = "darkgreen", "10" = "orange")
p2 <- p + scale_fill_manual(values = cols, limits = c("4", "8"))

p1 / p2
```

<sup>Created on 2019-04-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

yutannihilation:
Fix #3262

At the time of https://github.com/tidyverse/ggplot2/issues/3262#issuecomment-488263110, I thought this should be fixed in `manual_scale()`. But it turns out this is a bug of `ScaleDiscrete`; the if block below uses `limits` only when `pal` is named:

https://github.com/tidyverse/ggplot2/blob/e6cbcc1b52d3627e21922d2ee297ce3900ac096f/R/scale-.r#L778-L783

I see no reason to ignore `limits` here. (This code has been here since `Scale` was first implemented (https://github.com/tidyverse/ggplot2/commit/1877b6d70a48a4fa6fc5125d20d9ac12620a09c4), so there might have been some limitation?) This PR fixes the issue by limiting the `pal` first before matching it against the supplied values.


yutannihilation:
Thanks

PR #3687, opened 2019-12-22
Use EPSG code instead of proj4string "+init=epsg:XXXX" in test and example

### LINKED ISSUE #3686: PROJ.6 warning in test and example
I see this warning with PROJ 6.

```
──────────────────────────────────────────────────
test-geom-sf.R:162: warning: geom_sf_text() and geom_sf_label() draws correctly
GDAL Message 1: +init=epsg:XXXX syntax is deprecated. It might return a CRS with a non-EPSG compliant axis order.
──────────────────────────────────────────────────
```

c.f. https://proj.org/development/migration.html#backward-incompatibilities

yutannihilation:
Fix #3686


yutannihilation:
Thanks!

PR #3691, opened 2020-01-03
Improved docs for register_theme_elements()

clauswilke:
@hadley As requested. I wasn't sure whether I should use the prefix `ggxyz` in the code example as well, I can think of reasons for both yes and no.


clauswilke:
Ok, done.

PR #3694, opened 2020-01-07
Fix ScaleBinnedPosition for new axis implementation

thomasp85:
I just came across a spectacular failure of the new bin scale in combination with the new axis implementation. This PR adresses the two issues: Make sure guide is set to `waiver()` instead of `"none"`. And make sure that breaks are not mapped by the scale unless it is a discrete scale. The old approach worked before because mapping with a continuous position scale was a no-op, but this is not true for the bin scale and it would throw the axis completely off.

### DIFF
--- a/R/guides-axis.r
+++ b/R/guides-axis.r
@@ -80,7 +80,7 @@ guide_train.axis <- function(guide, scale, aesthetic = NULL) {
   } else if (length(breaks) == 0) {
     guide$key <- empty_ticks
   } else {
-    mapped_breaks <- if (scale$scale_is_discrete) scale$map(breaks) else breaks
+    mapped_breaks <- if (scale$is_discrete()) scale$map(breaks) else breaks
     ticks <- new_data_frame(setNames(list(mapped_breaks), aesthetic))
     ticks$.value <- breaks
     ticks$.label <- scale$get_labels(breaks)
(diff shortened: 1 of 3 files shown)


R/guides-axis.r
```diff
@@ -80,7 +80,8 @@ guide_train.axis <- function(guide, scale, aesthetic = NULL) {
   } else if (length(breaks) == 0) {
     guide$key <- empty_ticks
   } else {
-    ticks <- new_data_frame(setNames(list(scale$map(breaks)), aesthetic))
+    mapped_breaks <- if (scale$scale_is_discrete) scale$map(breaks) else breaks
```
paleolimbot: I think `scale$is_discrete()` would be better here (`scale_is_discrete` wasn't intended to be accessed directly in a `ViewScale`)

thomasp85:
I don't completely agree with your analysis. Breaks are the same between continuous and binned scales, i.e. a numeric along a continuous interval. It is the discrete scale that is the odd one as it converts the non-numeric values into the same type as that of continuous and binned. This is also the only reason why a mapping in needed and it makes sense to do it based on a discrete-check.

As bins don't necessarily have the same size I don't think it makes sense to have the default at the center - it would not be possible to deduce the correct border position solemnly on the centers

PR #3703, opened 2020-01-08
Make bin guides compatible with discrete scales

thomasp85:
This PR makes the new bin guides work with discrete scales if the breaks are given in a `cut()` compliant format (e.g. `(4, 19]`).

This, most importantly, makes the bin guides work with `geom_contour_filled()` but also gives a venue for users to pre-bin their data and still benefit from the new guides.

Example:

``` r
library(ggplot2)

volcano_long <- data.frame(
  x = as.vector(col(volcano)),
  y = as.vector(row(volcano)),
  z = as.vector(volcano)
)

ggplot(volcano_long, aes(x, y, z = z)) + 
  geom_contour_filled(aes(fill = stat(level)), bins = 8) + 
  scale_fill_brewer()
```

``` r

ggplot(volcano_long, aes(x, y, z = z)) + 
  geom_contour_filled(aes(fill = stat(level)), bins = 8) + 
  scale_fill_brewer(guide = guide_colorsteps())
```

<sup>Created on 2020-01-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -16,6 +16,18 @@
 #' @param show.limits Logical. Should the limits of the scale be shown with
 #'   labels and ticks.
 #'
+#' @section Use with discrete scale:
+#' This guide is intended to show binned data and work together with ggplot2's
+#' binning scales. However, it is sometimes desirable to perform the binning in
+#' a separate step, either as part of a stat (e.g. [stat_contour_filled()]) or
+#' prior to the visualisation. If you want to use this guide for discrete data
+#' the levels must follow the naming scheme implemented by [base::cut()]. This
+#' means that a bin must be encoded as `"(<lower>, <upper>]"` with `<lower>`
+#' giving the lower bound of the bin and `<upper>` giving the upper bound
+#' (`"[<lower>, <upper>)"` is also accepted). If you use [base::cut()] to
+#' perform the binning everything should work as expected, if not, some recoding
+#' may be needed.
+#'
 #' @return A guide object
 #' @family guides
 #' @export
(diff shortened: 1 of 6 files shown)


thomasp85:
Failure is just standard travis random failure

R/guide-bins.R
```diff
@@ -123,13 +123,25 @@ guide_train.bins <- function(guide, scale, aesthetic = NULL) {
...
   # in the key data frame, use either the aesthetic provided as
   # argument to this function or, as a fall back, the first in the vector
   # of possible aesthetics handled by the scale
   aes_column_name <- aesthetic %||% scale$aesthetics[1]
+
+  if (is.numeric(breaks)) {
+    limits <- scale$get_limits()
+    all_breaks <- c(limits[1], breaks, limits[2])
+    bin_at <- all_breaks[-1] - diff(all_breaks) / 2
+  } else {
```
clauswilke: The `else` branch is sufficiently complicated that a comments would be helpful. In particular, explain that when `breaks` is not numeric it's expected to be in the `(<lower>, <upper>]` format. I was very confused by the regex until I figured that out. Also, should this format be mentioned somewhere in the documentation?

R/guide-colorsteps.R
```diff
@@ -54,30 +54,46 @@ guide_colorsteps <- guide_coloursteps
...
       return()
-    limits <- scale$get_limits()
-    all_breaks <- c(limits[1], breaks, limits[2])
-    bin_at <- all_breaks[-1] - diff(all_breaks) / 2
+    }
+    if (is.numeric(breaks)) {
+      limits <- scale$get_limits()
+      all_breaks <- c(limits[1], breaks, limits[2])
+      bin_at <- all_breaks[-1] - diff(all_breaks) / 2
+    } else {
```
clauswilke: See my previous comment. Add some explanation to the code here as well.

thomasp85:
You are probably right about everything. I'll add some comments and improve the docs

Issue #3387, opened 2019-07-02
`geom_density()` for bounded data


echasnovski:
Currently `geom_density()` directly uses `density()` to compute kernel density estimation. Due to its nature, it has "extending property": output density can imply that values outside the range of input data are possible (with default "gaussian" kernel). This is a realistic practical setup, but there are cases when this is not true and data is bounded (for example, when only positive values are possible). It can be a good idea to support kernel density estimation on this type of bounded data with new `bounds` argument of `geom_density()`.

In my opinion, one of the methods most easiest to implement, understand, and teach, is "reflection" method. There is one possible description on [this page](https://ned.ipac.caltech.edu/level5/March02/Silverman/Silver2_10.html). Basically, boundary correction is done by doing "standard" kernel density estimation first, and then "reflecting" tails outside of desired interval to be inside. Densities inside and outside of desired interval are added together in "symmetric fashion": `d(x) = d_f(x) + d_f(l - (x-l)) + d_f(r + (r-x))`, where `d_f` is density of input, `d` is density of output, `l` and `r` are left and right edges of desired interval.

I made some quick and dirty changes to ggplot2 for demonstration. `stat_density()` gets `bounds` argument with default value of `c(-Inf, Inf)`. Here are some examples of proposed functionality:

```r
library(tibble)

set.seed(101)

ggplot(tibble(x = runif(100)), aes(x)) +
  geom_density() +
  geom_density(bounds = c(0, 1), color = "blue", ) +
  stat_function(data = tibble(x = c(0, 1)), fun = dunif, color = "red")
```

![](https://i.imgur.com/aYGG5Fs.png)

```r
ggplot(tibble(x = rexp(100)), aes(x)) +
  geom_density() +
  geom_density(bounds = c(0, Inf), color = "blue") +
  stat_function(data = tibble(x = c(0, 5)), fun = dexp, color = "red")
```

![](https://i.imgur.com/OyzYbJl.png)

echasnovski:
Here are more appropriate plots with plot limits completely representing current behavior:

``` r
library(ggplot2)

set.seed(101)

ggplot(data.frame(x = runif(100)), aes(x)) +
  geom_density() +
  geom_density(color = "blue", bounds = c(0, 1)) +
  stat_function(data = data.frame(x = c(0, 1)), fun = dunif, color = "red")
```

![](https://i.imgur.com/RNNCHGR.png)

``` r

ggplot(data.frame(x = rexp(100)), aes(x)) +
[...]

PR #3708, opened 2020-01-08
Set outline type for geom_density as geom_area

thomasp85:
The new stroking of `geom_ribbon()` and `geom_area()` didn't make it to `geom_density()` This PR adds it and make it behave like `geom_area()`

### DIFF
--- a/R/geom-density.r
+++ b/R/geom-density.r
@@ -12,6 +12,7 @@
 #'   See [geom_violin()] for a compact density display.
 #' @inheritParams layer
 #' @inheritParams geom_bar
+#' @inheritParams geom_ribbon
 #' @param geom,stat Use to override the default connection between
 #'   `geom_density` and `stat_density`.
 #' @export
@@ -56,7 +57,9 @@ geom_density <- function(mapping = NULL, data = NULL,
                          na.rm = FALSE,
                          orientation = NA,
                          show.legend = NA,
-                         inherit.aes = TRUE) {
+                         inherit.aes = TRUE,
+                         outline.type = 'upper') {
+  outline.type <- match.arg(outline.type, c("both", "upper", "lower", "legacy"))
 
   layer(
     data = data,
@@ -69,6 +72,7 @@ geom_density <- function(mapping = NULL, data = NULL,
     params = list(
       na.rm = na.rm,
       orientation = orientation,
+      outline.type = outline.type,
       ...
     )
   )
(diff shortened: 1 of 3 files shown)


yutannihilation:
You need to add `geom_density` here to avoid argument checks:

https://github.com/tidyverse/ggplot2/pull/3529/commits/9c21ed72db8acaefc4f04db021a438cccc636ef5

yutannihilation:
Hmm, you are right. Hope this is OK this time...

yutannihilation:
@thomasp85 @hadley 
Now I feel this is not OK after looking at the examples on [`geom_density()`](https://ggplot2.tidyverse.org/dev/reference/geom_density.html), a geom with no fill by default. Note that these figures are drawn against master and doesn't reflect this pull request, so it's just my fault in #3529. Sorry I didn't aware of this visual breakage.

I want to propose renaming option `"legacy"` to `"full"` and making it default for `geom_density()`. What do you think?

Screenshot from 2020-01-09 08-42-24

thomasp85:
I am honestly against. The full stroking of density plots has always looked ugly to my eye. I’m for the renaming, though

yutannihilation:
> The sides and lower bound in a density plot is not really data, and stroking them just creates visual noise IMO

Thanks, agreed. After some investigation, I now agree that drawing the upper lines only is the right thing as base R does so. 

``` r
plot(density(diamonds$carat, adjust = 5))
```



What makes me uneasy with upper-only `geom_density()` is that the line doesn't start from 0 by default, but it seems because `geom_density()` provides a narrower plot range than `plot(density())` (#3387 might be related?). So, you are definitely right in that the side strokes are "visual noise" that have hided this problem. I feel the right one below should be drawn by default. I'll file a new issue for this.

``` r
p <- ggplot(diamonds, aes(carat)) +
  geom_density(adjust = 5)

patchwork::wrap_plots(
  p + ggtitle("(default)"),
  p + ggtitle("(expanded)") + expand_limits(x = -0.5)
)
```



> I’m for the renaming, though

Sure.

thomasp85:
I agree that the rightmost is more pleasing, but I think that ggplot2 has historically erred to the side of keeping the data-range. This is also done by default in violin plots. Extrapolating is something you have to do wilfully, as it creates data from thin air

yutannihilation:
Sounds fair, but let me dig into the history...

PR #3710, opened 2020-01-08
don't drop ordering from ordinal factors

thomasp85:
When I wrote out ply, I inadvertently dropped support for ordered factors. This is such a niche data structure that it didn't set of any alarms, but I noticed it when I wanted `stat_contour_filled()` to return levels as ordered.

This PR fixes the regression by not converting ordered factors into factors when binding data.frames


clauswilke:
Can you add a unit test for this case?

thomasp85:
sure thing 🙂

Issue #3688, opened 2019-12-23
Feature request: type stability in internals, specifically at `unlist()` operations


teunbrand:
I was experimenting with having S3 classes based on the `vctrs` package as xy coordinates and extending the scales to support these alternative coordinates (link to [specific use case](https://github.com/teunbrand/ggnomics/issues/32) ). However, I ran into the problem that unlist operations in ggplot code drop relevant attributes and recursively flatten the data. This would be preventable if unlist operations on S3 classes preserve type stability.

One of the pieces of code that I'm struggling with, is updating/training limits, which needs to go through the `scale_apply()` function, wherein [data is unlisted](https://github.com/tidyverse/ggplot2/blob/660aad2db2b3495ae0d8040915a40d247133ffc0/R/layout.R#L309-L311).

I've asked a [question on stackoverflow](https://stackoverflow.com/questions/59401011/unlist-behaviour-on-vctr-rcrd-class-objects) to try and get around this from the vctrs end that illustrates my struggle with unlist and `vctrs` classes. I'm also aware that the `vctrs` package has its own [syntax](https://github.com/r-lib/vctrs/issues/395) for unlist operations.

I think that type stability in general would be a good thing, and would allow for a richer set of ggplot extensions that support `vctrs`-based S3 classes. Therefore, I'd like to propose more type stable equivalents of `unlist()` in the ggplot internals.

yutannihilation:
Could you provide a reprex (or at least a pseudo-reprex)? I feel it might make some sense to use vctrs, but, since I don't have knowledge about `GRanges`, I can't imagine why it's bothering you...

teunbrand:
Sure thing! I've adapted the code below from the question on stack overflow, but added in another field. This is the behaviour I encounter when using a `vctrs_rcrd` derivative with `unlist()`, and the same happens in for example the `scale_apply()` function in ggplot:

``` r
library(vctrs)

# Setting up a constructor
facvec <- function(f) {
  vec_assert(f, partial_factor())
  new_rcrd(fields = list(f1 = f, f2 = f), class = "facvec")
}

# Setting up how facvec's should be combined
# Based on: https://vctrs.r-lib.org/articles/s3-vector.html
vec_ptype2.facvec <- function(x, y, ...) UseMethod("vec_ptype2.facvec", y)
vec_ptype2.facvec.default <- function(x, y, ..., x_arg = "x", y_arg = "y") {
[...]

PR #3711, opened 2020-01-09
More type stability for scale_apply()

teunbrand:
This PR follows from the discussion at #3688 in order that the proposed changes could be reviewed here.

In brief:
It could be convenient if scales would be able to work with vectors other than numeric or character, for example S3 vectors based on the vctrs package. To get there in a way that can be useful, these vectors would need to be type stable and preserve attributes. Currently, the only barrier to type stability in not-extendable parts of ggplot is the `scale_apply()` function, as the `unlist()` there drops attributes. The idea is then to replace `unlist()` with a `do.call("c", ...)`, which does preserve type and attributes, when the classes of the `pieces` elements have an appropriate `c` and subsetting methods.

It has been hard to design a concise unit test without using the vctrs package and `scale_apply()` being so deeply burrowed into ggplot code, so if anyone has suggestions for improvements feel welcome to share!


tests/testthat/test-scales.r
```diff
@@ -319,3 +319,58 @@ test_that("limits with NA are replaced with the min/max of the data for continuo
...
+  `[.baz` <- function(x, i) {
+    attris <- attributes(x)
+    x <- unclass(x)[i]
+    attributes(x) <- attris
+    x
+  }
+
+  # These S3 methods won't work within testthat, so assign to global environment
+  assign("c.baz", `c.baz`, global_env())
+  assign("[.baz", `[.baz`, global_env())
```
yutannihilation: I'm not familiar with this kind of test, but it seems we can use `rlang::local_binding()` for temporal assignment on the global environment. e.g. https://github.com/r-lib/vctrs/blob/ab84679b816b903cbafa9aa042ec1c9c98d385a5/tests/testthat/helper-s3.R#L9-L11

teunbrand: Ah excellent, I was hoping such a thing existed but didn't know it did!

teunbrand:
I don't think the Travis CI error necessarily relates to this PR. Is there a way to restart it?
This is probably the relevant error message:

```
* checking examples ... ERROR

Running examples in ‘ggplot2-Ex.R’ failed

The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")

> ### Name: annotation_map

> ### Title: Annotation: a maps

> ### Aliases: annotation_map

> 

> ### ** Examples

> 

> if (require("maps")) {

+ usamap <- map_data("state")

+ 

+ seal.sub <- subset(seals, long > -130 & lat < 45 & lat > 40)

+ ggplot(seal.sub, aes(x = long, y = lat)) +

+   annotation_map(usamap, fill = "NA", colour = "grey50") +

+   geom_segment(aes(xend = long + delta_long, yend = lat + delta_lat))

+ }

Loading required package: maps

Error: Unknown colour name: NA

Execution halted
```

yutannihilation:
Ah, yes, it's not your fault and already fixed by #3714 on master. You can ignore (or merge the current master).

clauswilke:
The only concern that I have is that `NA`s are now handled differently, as you mentioned in a comment somewhere. Can you elaborate on this some more? Can you create an example with `NA`s, so we can evaluate whether it may be a problem or not?

@paleolimbot Any comments from your side?

teunbrand:
Well, `NA`s as input aren't going to be handled different necessarily, it's the function that produces less `NA`s due to out-of-bounds subsetting operations.

I found out about these `NA`s while I was in the debug window after calling `debug(ggplot2:::scale_apply)` and evaluating the `lapply()`-bit of that function in the console (otherwise it would go straight to exitting the function after debugging).

To see that these `NAs` appear, you could run the following. I ran this on a computer with the old version of `scale_apply`:

``` r
library(ggplot2)

df <- data.frame(
  x = 1:3,
  y = 1:3
)

plot <- ggplot(df, aes(x, y)) + geom_point()
scale <- list(scale_x_continuous())

ggplot2:::scale_apply(df, "x", "train", c(1,1,1), scale)
#> [[1]]
#> [1]  1  3 NA
```

<sup>Created on 2020-01-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think the `scale_apply()` input is realistic, because if you do the following next in RStudio:

```
debug(ggplot2:::scale_apply)

ggplot_build(plot)
```

You can see what the input to the `scale_apply()` function is under normal circumstances, by inspecting/printing the arguments by their names in the console, while in debug mode.

The reason I think these `NA`s come into being is because the `pieces` variable is subsetted by an out-of-bounds index, analogous to the following:
```
pieces <- list(c(1, 3))
scale_index <- 1:3
unlist(pieces)[order(unlist(scale_index))]
```

Obviously the ggplot code is able to deal with these produced `NA`s without any problem, but the out-of-bounds subsetting presents a problem for S3 vctrs-derivatives, since this is not allowed by default. In order to subset within bounds, these `NA`s will be produced no longer. Also, the ggplot code also appears to be fine without these `NA`s, otherwise I would have expected some unit- or visual- tests to fail. I think they are just the result of a small oversight and don't serve a purpose anywhere.

clauswilke:
@paleolimbot If this looks good to you, do you want to approve and merge?

R/layout.R
```diff
@@ -307,7 +307,8 @@ scale_apply <- function(data, vars, method, scale_id, scales) {
     })
     # Join pieces back together, if necessary
     if (!is.null(pieces)) {
```
paleolimbot: I can't see any possible way that `pieces` could be `NULL` here (`lapply` never returns `NULL`). In the name of making this function a tiny bit easier to understand for the next person, could you remove the `if()` wrapper here (and the "if necessary" comment, making sure all the tests still pass)?

teunbrand: Sure thing, good point!

teunbrand:
Thanks, awesome!

PR #3720, opened 2020-01-10
Fixes #3718

### LINKED ISSUE #3718: Theme-related bug in release candidate
I just found an annoying theme-related bug in the release candidate. Not sure if there's still time to fix it. I don't want to risk messing anything else up.

The issue is: When rendering axis labels, any element properties that are not available in `element_text()` will get stripped out in this code block:
https://github.com/tidyverse/ggplot2/blob/94b5a16ce0b508e4ad67ae1aa504660221830327/R/guides-axis.r#L195-L199

The intent is clearly to work with elements other than `element_text()` (even the comments state this), but in practice it doesn't work.

Reprex:
``` r
library(ggplot2)
library(ggtext)

ggplot(mpg, aes(class)) + 
  geom_bar() +
  scale_x_discrete(
    name = "vehicle *class*",
    labels = c("2seater", "*compact*", "midsize", "**minivan**", "pickup", "<sub>sub</sub>compact", "suv")
  ) +
  theme(
    axis.title.x = element_markdown(
      linetype = 1, fill = "cornsilk", r = unit(5, "pt"), padding = margin(4, 4, 4, 4)
    ),
    axis.text.x = element_markdown(
      # these properties get stripped
      linetype = 1, fill = "cornsilk", r = unit(5, "pt"), padding = margin(4, 4, 4, 4)
    )
  )
```

<sup>Created on 2020-01-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

(The axis tick labels should be enclosed in boxes, just like the axis title. This doesn't happen, because `linetype`, `fill`, etc. get stripped. We can see, though, that the tick labels are rendered as markdown, as requested.)

The problem is that `merge_element()` doesn't keep any properties of the old element that aren't defined in the new element:
https://github.com/tidyverse/ggplot2/blob/47a06183c4aa1fe8c4d8c68921c148e6fb9e0cdd/R/theme.r#L633-L640

A fix is theoretically simple, I just don't know immediately off the top of my head whether `merge_element()` is behaving incorrectly or whether the code that overrides axis text properties should not use `merge_element()` to combine the elements.

@paleolimbot @thomasp85 Thoughts? Can this be fixed before the release? I'll let it sit for a few hours and then see if I can prepare a PR.

clauswilke:
This fixes #3718. I am confident about this fix. It doesn't touch `merge_element()` at all, only makes a minor change in `draw_axis()`, and simply copies over the relevant properties manually instead of merging the elements. Reprex below shows that things work as expected with this fix.

``` r
library(ggplot2)
library(ggtext)

ggplot(mpg, aes(class)) + 
  geom_bar() +
  scale_x_discrete(
    name = "vehicle *class*",
    labels = c("2seater", "*compact*", "midsize", "**minivan**", "pickup", "<sub>sub</sub>compact", "suv")
  ) +
  theme(
    axis.title.x = element_markdown(
      linetype = 1, fill = "cornsilk", r = unit(5, "pt"), padding = margin(4, 4, 4, 4)
    ),
    axis.text.x = element_markdown(
      linetype = 1, fill = "cornsilk", r = unit(5, "pt"), padding = margin(4, 4, 4, 4)
    )
  )
```

<sup>Created on 2020-01-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/guides-axis.r
+++ b/R/guides-axis.r
@@ -193,10 +193,18 @@ draw_axis <- function(break_positions, break_labels, axis_position, theme,
   # override label element parameters for rotation
   if (inherits(label_element, "element_text")) {
     label_overrides <- axis_label_element_overrides(axis_position, angle)
-    # label_overrides is always an element_text(), but in order for the merge to
-    # keep the new class, the override must also have the new class
-    class(label_overrides) <- class(label_element)
-    label_element <- merge_element(label_overrides, label_element)
+    # label_overrides is an element_text, but label_element may not be;
+    # to merge the two elements, we just copy angle, hjust, and vjust
+    # unless their values are NULL
+    if (!is.null(label_overrides$angle)) {
+      label_element$angle <- label_overrides$angle
+    }
+    if (!is.null(label_overrides$hjust)) {
+      label_element$hjust <- label_overrides$hjust
+    }
+    if (!is.null(label_overrides$vjust)) {
+      label_element$vjust <- label_overrides$vjust
+    }
   }
 
   # conditionally set parameters that depend on axis orientation


clauswilke:
@thomasp85 Thanks! Just to be clear: This PR is against the RC branch. Is it still Ok to merge? And if yes, should I make the same PR against master also or just wait until the fix makes it way from RC to master? (It's not urgent to get this into master, but I'd like to have the fix in the release if possible.)

thomasp85:
Yes, it’s fine to have in rc. You are modifying unreleased code so there are no breaking changes to worry about. Just wait for rc to be merged into master. No need to dual commit

PR #3733, opened 2020-01-15
make sure aes renaming is reflected in data

thomasp85:
This issue is based on failing revdep check of ggspectra and its reverse dependencies

Currently aes name standardisation changes every occurrence of "color" to "colour", not just the exact string "color". This has not been a problem before, but with the new aesthetic evaluation we have chosen to substituted names inside the expressions used in `after_stat()` and `after_scale()` and this causes problems if the expression references a US spelling column in the data.

This PR fixes it by only matching exactly to "color", but I'm unsure about whether there will be unforeseen consequences of this change. Another approach would be to substitute the names of the columns as well

### DIFF
--- a/R/aes.r
+++ b/R/aes.r
@@ -158,7 +158,7 @@ print.uneval <- function(x, ...) {
 #' @export
 standardise_aes_names <- function(x) {
   # convert US to UK spelling of colour
-  x <- sub("^color$", "colour", x, fixed = TRUE)
+  x <- sub("color", "colour", x, fixed = TRUE)
 
   # convert old-style aesthetics names to ggplot version
   revalue(x, ggplot_global$base_to_ggplot)
--- a/R/aes.r
+++ b/R/aes.r
@@ -158,7 +158,7 @@ print.uneval <- function(x, ...) {
 #' @export
 standardise_aes_names <- function(x) {
   # convert US to UK spelling of colour
-  x <- sub("color", "colour", x, fixed = TRUE)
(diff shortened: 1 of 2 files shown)


R/aes.r
```diff
@@ -158,7 +158,7 @@ print.uneval <- function(x, ...) {
 #' @export
 standardise_aes_names <- function(x) {
   # convert US to UK spelling of colour
-  x <- sub("color", "colour", x, fixed = TRUE)
+  x <- sub("^color$", "colour", x, fixed = TRUE)
```
yutannihilation:
```suggestion
  x <- sub("^color$", "colour", x)
```

or simply

``` r
x[x == "color"] <- "colour"
```

thomasp85: I've decided on the alternative approach, as we actually rely on the arbitrary color renaming

thomasp85:
I've pivoted on this and changed the fix to instead rename the data columns... I'm not 100% sure that this will not have consequences for extensions, but we'll have to wait for a new revdep check

R/aes.r
yutannihilation: Sure.

clauswilke:
I think this is the right approach. The rule "ggplot2 always considers US and UK spelling of `colour` to be the same thing, in all contexts" is clear and simple to remember. If there truly is an extension package that uses a column `color` and a column `colour` and expects them to be treated differently, I think it's reasonable to ask them to reconsider their column naming approach.

Issue #3674, opened 2019-12-17
Release ggplot2 3.3.0


thomasp85:
Prepare for release: 
 
* [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_ggplot2.html) 
* [x] `devtools::check(remote = TRUE, manual = TRUE)` 
* [x] `devtools::check_win_devel()` 
* [x] `rhub::check_for_cran()` 
* [x] `revdepcheck::revdep_check(num_workers = 4)` 
* [x] Update `cran-comments.md` 
* [x] [Polish NEWS](https://style.tidyverse.org/news.html#news-release) 
* [x] Review pkgdown reference index for, e.g., missing topics 
* [x] Draft blog post 
 
Submit to CRAN: 
 
* [x] `usethis::use_version('minor')` 
* [x] `devtools::submit_cran()` 
* [x] Approve email 
 
Wait for CRAN... 
 
* [x] Accepted :tada: 
* [x] `usethis::use_github_release()` 
* [x] `usethis::use_dev_version()` 
* [x] Finish blog post 
* [x] Tweet 
* [x] Add link to blog post in pkgdown news menu

thomasp85:
We are not beginning the release process just yet, but I'm opening this so we have a venue for coordination. The hard deadline for a release is RStudio::conf in late January, but I think we can easily get it out before that.

@clauswilke, @yutannihilation, @Hadley, @paleolimbot, @karawoo can you tag any open issues you deem necessary to get into this release. I don't think we should add any more features to the milestone, but if there is something of utmost importance, please make your case 🙂

We have already run a preliminary revdepcheck and fixed a couple of regressions. There are a number of maintainers that will need to update their packages because they depend on the old axis data structure. @Hadley will send out mails to these.

We will run an additional revdepcheck when all code in the milestone has been merged in.

Don't work on this over the holidays if you can help it. You all deserve a break!

clauswilke:
I think issue #3557 would be a good candidate to complete before the release, though maybe it's too big of a job. It's about enabling markdown formatting in strip labels. I plan to release ggtext as soon as ggplot2 3.3.0 is out, and people will want to use it, and will find out that they can't use it for strip labels.

The argument against finishing up #3557: I looked at the code and didn't find an easy way to do this. Maybe somebody else wants a challenge? :-)

thomasp85:
I can try to look at it. If I fold I don’t think it is reasonable to wait for it

[...]

PR #3735, opened 2020-01-15
WIP: Allow expressions in facet specs even when some variables are not available on some layers

### LINKED ISSUE #2963: Reference lines break facet_wrap when facets are the product of a function
I want to add some reference lines to a faceted plot. The lines should be the same in every facet. I want to use a custom value not mapped from the data. It works just fine as long as I use plain variables for faceting, but when I mix or alter them, I get an error, which depends on the operation I perform.

I think the problem lies in the fact that the plot wants to compute facets with for the non-mapped layer, but the data doesn't exist. I don't think geom_line has a problem, but rather facet_wrap.

```
## Example 1 ------------------------------
## Works
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(am))

## Doesn't work
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(2 * am)) # <-- there's an operation involved

## Example 2 ------------------------------
## Works
ggplot(mtcars, aes(cyl)) +
  geom_bar() +
  geom_vline(xintercept = 4.5) +
  facet_wrap(vars(am, vs))

## Doesn't work
ggplot(mtcars, aes(cyl)) +
  geom_bar() +
  geom_vline(xintercept = 4.5) +
  facet_wrap(vars(paste(am, vs))) # <-- there's an operation involved

## Easiest fix I can think of: move the operation to a mutate
mtcars %>% 
  mutate(facet = paste(am, vs)) %>% 
  ggplot(aes(cyl)) +
  geom_bar() +
  geom_vline(xintercept = 4.5) +
  facet_wrap(vars(facet))
```

yutannihilation:
Fix #2963

Not all variables in a facet spec are available on all the layers. A solution suggested on https://github.com/r-lib/rlang/issues/888#issuecomment-573691024 is:

> bind the same symbols inside each panel, and when the symbols are undefined use an active binding to throw a typed error? And then you'd only catch these particular errors in the `tryCatch()`

This PR injects such an active bindings on the column names of all the plot data so that we can let `eval_facet()` fail gracefully.

Note that this PR removes `env` argument of `eval_facet()` to avoid confusion. I believe all facet specs are converted to quosures and bare expressions are not allowed, which means `env` is always ignored on `eval_tidy()`.

``` r
devtools::load_all("~/repo/ggplot2/")
#> Loading ggplot2

# works fine with expressions
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(2 * am))
```

``` r

# works fine with external variables
two <- 2
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(two * am))
```

``` r

# raises an error when the expression refers to some non-existent variable
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(no_such_variable * am))
#> Error in eval_tidy(facet, data): object 'no_such_variable' not found

# special case: raises a friendlier error when the expression is a symbol
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(no_such_variable))
#> Error: At least one layer must contain all faceting variables: `no_such_variable`.
#> * Plot is missing `no_such_variable`
#> * Layer 1 is missing `no_such_variable`
#> * Layer 2 is missing `no_such_variable`
```

<sup>Created on 2020-01-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

----
TODO:
- [ ] Add a NEWS bullet
- [ ] Consider using `warn_for_aes_extract_usage()` here as well
- [ ] Consider if it's really OK to add a param to `Facet` (I bet it's OK)

### DIFF
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -436,14 +436,11 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
     return(out)
   }
 
-  # Create an environment for data mask
+  # Key idea: use active bindings so that column names missing in this layer
+  # but present in others raise a custom error
   env <- new_environment(data)
-
-  # Bind all possible column names that remains undefined to raise a custom error
-  # so that we can detect and ignore the case when a variable is missing from the
-  # layer data but exists in other layer
   missing_columns <- setdiff(possible_columns, names(data))
-  undefined_error <- function(e) abort("", class = "ggplot2_undefined_aes_error")
+  undefined_error <- function(e) abort("", class = "ggplot2_missing_facet_var")
   bindings <- rep_named(missing_columns, list(undefined_error))
   env_bind_active(env, !!!bindings)
 
@@ -451,11 +448,9 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
   mask <- new_data_mask(env)
   mask$.data <- as_data_pronoun(mask)
 
-  # Do not treat the cases as errors when it refers to a column name unavailable
-  # in the layer data
   tryCatch(
     eval_tidy(facet, mask),
-    ggplot2_undefined_aes_error = function(e) NULL
+    ggplot2_missing_facet_var = function(e) NULL
   )
 }
 
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -423,17 +423,6 @@ eval_facets <- function(facets, data, possible_columns = NULL) {
   new_data_frame(tibble::as_tibble(vars))
 }
 eval_facet <- function(facet, data, possible_columns = NULL) {
-  if (quo_is_symbol(facet)) {
-    facet <- as.character(quo_get_expr(facet))
-
-    if (facet %in% names(data)) {
-      out <- data[[facet]]
-    } else {
-      out <- NULL
-    }
-    return(out)
-  }
-
   # clone the env in order to prevent side effects (hopefully)
   env <- env_clone(quo_get_env(facet))
 
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -444,8 +444,7 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
   # layer data but exists in other layer
   missing_columns <- setdiff(possible_columns, names(data))
   undefined_error <- function(e) abort("", class = "ggplot2_undefined_aes_error")
-  bindings <- rep_along(missing_columns, list(undefined_error))
-  names(bindings) <- missing_columns
+  bindings <- rep_named(missing_columns, list(undefined_error))
   env_bind_active(env, !!!bindings)
 
   # Create a data mask and install a data pronoun manually (see ?new_data_mask)
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -443,10 +443,9 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
   # so that we can detect and ignore the case when a variable is missing from the
   # layer data but exists in other layer
   missing_columns <- setdiff(possible_columns, names(data))
-  bindings <- lapply(
-    set_names(missing_columns),
-    function(...) function(e) abort("", class = "ggplot2_undefined_aes_error")
-  )
+  undefined_error <- function(e) abort("", class = "ggplot2_undefined_aes_error")
+  bindings <- rep_along(missing_columns, list(undefined_error))
+  names(bindings) <- missing_columns
   env_bind_active(env, !!!bindings)
 
   # Create a data mask and install a data pronoun manually (see ?new_data_mask)
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -418,12 +418,11 @@ is_facets <- function(x) {
 # when evaluating an expression, you want to see any errors. That does
 # mean you can't have background data when faceting by an expression,
 # but that seems like a reasonable tradeoff.
-eval_facets <- function(facets, data, env = globalenv(), possible_columns = NULL) {
-  vars <- compact(lapply(facets, eval_facet, data, env = env, possible_columns = possible_columns))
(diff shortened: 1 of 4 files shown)


yutannihilation:
Note that this cannot handle data pronoun at the moment. Probably I did wrong about environment ordering and I should define active bindings on the data mask directly. Will fix...

``` r
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(.data$am))
#> Error: Column `am` not found in `.data`
```

R/facet-.r
```diff
@@ -433,7 +436,25 @@ eval_facet <- function(facet, data, env = emptyenv()) {
     return(out)
   }
 
-  eval_tidy(facet, data, env)
+  # clone the env in order to prevent side effects (hopefully)
+  env <- env_clone(quo_get_env(facet))
+
+  # create a env with active bindings
+  cushioning_env <- child_env(env_parent(env))
```
clauswilke: From the documentation of `child_env()`:
> `child_env()` is in the questioning stage. It is redundant now that `env()` accepts parent environments.

yutannihilation: Oops, thanks!

yutannihilation:
Now `.data` works.

``` r
devtools::load_all("~/repo/ggplot2/")
#> Loading ggplot2

ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(.data$am))
```



<sup>Created on 2020-01-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
I think this is basically ready for review bit, since this is not a one in v3.3.0, let's leave this for a while. Any comments are welcome.

hadley:
This seems like an elegant approach to fixing the problem. What are the potential risks to including this in 3.3.0? It feels fairly low risk to me.

yutannihilation:
Thanks, I just hesitated to add a new one to the milestones while [the hard deadline](https://github.com/tidyverse/ggplot2/issues/3674#issuecomment-566446232) is approaching, but I'd love to finish this PR if there's still some more time.

I don't have any big concern about this change because the most cases will be handled by this existing logic (though I'm not yet sure whether we should keep this or not):

``` r
  if (quo_is_symbol(facet)) {
    facet <- as.character(quo_get_expr(facet))

    if (facet %in% names(data)) {
      out <- data[[facet]]
    } else {
      out <- NULL
    }
    return(out)
  }
```

Note that this means we might still need `is_symbol(strip_dots(quo_get_expr(facet)))` to issue a consistent warning for the case of `.data$no_such_columns` as `no_such_columns` (see the details below).

<details>

``` r
library(ggplot2)

p <- ggplot(mpg) + geom_bar(aes(cyl))

# friendly error
p + facet_wrap(vars(no_such_column))
#> Error: At least one layer must contain all faceting variables: `no_such_column`.
#> * Plot is missing `no_such_column`
#> * Layer 1 is missing `no_such_column`

# fine, but unfriendly error
p + facet_wrap(vars(.data$no_such_column))
#> Error: Column `no_such_column` not found in `.data`

# not sure if this should be allowed or not
no_such_column <- 1
p + facet_wrap(vars(no_such_column))
#> Error: At least one layer must contain all faceting variables: `no_such_column`.
#> * Plot is missing `no_such_column`
#> * Layer 1 is missing `no_such_column`
```

</details>

@clauswilke @thomasp85
What do you think about including this in 3.3.0?

clauswilke:
I think it's somewhat risky to introduce this kind of change this late in the release process, even if we're pretty sure it'll be fine. A safe approach would be to add the patch to a branch of the release branch, run a revdep check on that, and only merge into the release branch if no new regressions are found.

hadley:
Let me give this a thorough review, and I'll contemplate where this might change existing behaviour. I think if it's low risk we can merge, and then re-run the revdeps; if that reveals any problems we can back out and try again for the next version.

hadley:
@yutannihilation I made a couple of tiny style changes, but otherwise this is lovely

@clauswilke my analysis suggests that this change should only affect a small set of plots that previously errored. I suggest we merge, and then re-run revdepchecks just to be safe.

@thomasp85 does that sound ok with you? I would like to get this into this release so we can over a comprehensive approach to tidyeval that works the same way in ggplot2 and dplyr (1.0.0)

hadley:
@yutannihilation actually I failed to push my changes into your branch; can you please make sure it's editable? (there's should be a checkbox somewhere on this page that only you can see)

thomasp85:
This looks good. I’m fine with trying a merge into the rc branch

yutannihilation:
@hadley 
Thanks, I made the branch editable (sorry it seems I forgot to check the box...). Could you try again?

hadley:
@yutannihilation done, thanks!

yutannihilation:
@hadley
Thanks for your review and improvement!

@thomasp85 
I created #3757, a new PR for the rc branch. If you are already working on merging this into master, please feel free to close this.

clauswilke:
@hadley Sounds good.

hadley:
I forgot that we can't easily modify this PR to merge on the RC, so lets close this PR. Further discussion can happen on #3757

PR #3740, opened 2020-01-16
Handle label functions that return character(0)

thomasp85:
This PR addresses an issue in scales. If a label function converts an `NA` break to a `character(0)` label we will get an error later on when we enlist the label list. This is fixed here in the `ScaleContinuous$get_labels()` function by converting such list elements to empty strings (`""`) instead.

It can be argued that this should get fixed in `scales::label_number_auto()` (and perhaps it should as well), but I think it makes sense to have a guard in ggplot2 as well, since otherwise this problem will be carried all the way to `guide_axis()` and create an unintelligible error there.

### DIFF
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -692,6 +692,12 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
       labels[vapply(labels, length, integer(1)) == 0] <- ""
       # Make sure each element is scalar
       labels <- lapply(labels, `[`, 1)
+
+      if (any(vapply(labels, is.language, logical(1)))) {
+        labels <- do.call(expression, labels)
+      } else {
+        labels <- unlist(labels)
+      }
     }
 
     labels
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -690,6 +690,8 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     if (is.list(labels)) {
       # Guard against list with empty elements
       labels[vapply(labels, length, integer(1)) == 0] <- ""
+      # Make sure each element is scalar
+      labels <- lapply(labels, `[`, 1)
     }
 
     labels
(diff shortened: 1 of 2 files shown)


R/scale-.r
```diff
@@ -687,6 +687,10 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     if (length(labels) != length(breaks)) {
       abort("Breaks and labels are different lengths")
     }
+    if (is.list(labels)) {
+      # Guard against list with empty elements
+      labels[vapply(labels, length, integer(1)) == 0] <- ""
```
hadley: What happens if the list contains elements of length > 1?

thomasp85: It would fail in the same way, I guess... but I think it is more likely that we have functions that inadvertently convert `NA` input to `character(0)`.

I can check for `> 1` as well and just take the first element to be sure.

thomasp85: done... do you think this should also be fixed in scales, or is it behaving as it should?

thomasp85:
It surfaced in scales with
```r
> demo_log10(c(1, 1e7), labels = label_number_auto())
# scale_x_log10(labels = label_number_auto())
#  Error in `$<-.data.frame`(`*tmp*`, ".label", value = c("10", "1 000",  : 
#   replacement has 4 rows, data has 5 
```

R/scale-.r
hadley: It's definitely suspicious that scales is returning something of the wrong size, but I don't remember what I was thinking in the scales code. Probably worth merging this expedient fix and filing a scales issue so we don't forget to fix upstream.

thomasp85:
My reason for placing it where I have is to avoid all axis guides (assuming extensions will be made) has to make this correction. Basically I think it is fair that the output of `get_labels()` returns sensible output instead of relying on downstream functions cleaning it up

PR #3746, opened 2020-01-20
Strip .data from default labels

hadley:
@thomasp85 this ensures that when you use the `.data` pronoun with ggplot2, it doesn't adversely affect the labels.

@paleolimbot — this should make your talk slightly simpler 😄

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -162,8 +162,7 @@ fail.
 * `stat_summary()` and related functions now support rlang-style lambda functions
   (#3568, @dkahle).
 
-* `strip_dots()` strips the data masking pronoun `.data` to make better
-  axis labels.
+* The data mask pronoun, `.data`, is now stripped from default labels.
 
 * Addition of partial themes to plots has been made more predictable;
   stepwise addition of individual partial themes is now equivalent to
--- a/NEWS.md
+++ b/NEWS.md
@@ -162,6 +162,9 @@ fail.
 * `stat_summary()` and related functions now support rlang-style lambda functions
   (#3568, @dkahle).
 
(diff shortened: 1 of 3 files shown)


hadley:
I don't _think_ the `plotenv` should matter here, since all the environment information is stored in the quosures inside the aesthetics. The function signature is mildly confusing because the default `env` should never actually be used since it's always overridden by the quosure environment; it's only need to get recursion off the ground.

hadley:
If you're concerned about it, I can just leave it without a default — if it fails (which based on my understanding of the code, shouldn't happen), it'll be caught by the `tryCatch()` so you'll just get a slightly less optimal label.

thomasp85:
I have no problem with it, but it is just slightly unclear that evaluation happens in the correct environment without a comment. So for future generations of me, I'd like a small comment

thomasp85:
If you remove the default, the argument is basically unused all the way through the recursion. Why have the argument at all then? I feel I'm missing something about this code...

thomasp85:
Ah... now I see. Disregard my last comment

NEWS.md
```diff
@@ -162,6 +162,9 @@ fail.
 * `stat_summary()` and related functions now support rlang-style lambda functions
   (#3568, @dkahle).
 
+* `strip_dots()` strips the data masking pronoun `.data` to make better
```
yutannihilation: `strip_dots()` is not an exported function, so I think it might better to write more simply like "The data mask pronoun `.data` is now stripped from default labels."

hadley: Good idea, thanks!

PR #3757, opened 2020-01-22
Allow expressions in facet specs even when some variables are not available on some layers

### LINKED ISSUE #2963: Reference lines break facet_wrap when facets are the product of a function

yutannihilation:
(Cherry-picked version of #3735 for v3.3.0-rc branch)

---

Fix #2963

### Background

Not all variables in a facet spec are available on all the layers. A solution suggested on https://github.com/r-lib/rlang/issues/888#issuecomment-573691024 is:

> bind the same symbols inside each panel, and when the symbols are undefined use an active binding to throw a typed error? And then you'd only catch these particular errors in the `tryCatch()`

### Main change

This PR injects such an active bindings on the column names of all the plot data so that we can let `eval_facet()` fail gracefully.

`eval_facet()` evaluates facet specs in these two steps:

1. When the spec is a symbol (the most common case), check if the layer data contains the same name of column. If it does, subset the column and return it. If it doesn't, return `NULL`. In this case, no error happens even if the symbol is not available on all the layers.
2. When the spec is a expression, evaluate it. In this case, an error is raised on the layer where the symbol is not available.

This PR changes only step 2, so I expect this doesn't affect most of the existing code. Note that, while step 2 can also handle symbols, we still need step 1 to issue a friendly error ("At least one layer must contain all faceting variables") when no layer contain the variable.

``` r
devtools::load_all("~/repo/ggplot2/")
#> Loading ggplot2

# works fine with expressions
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(2 * am))
```

``` r
# works fine with external variables
two <- 2
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(two * am))
```

``` r
# raises an error when the expression refers to some non-existent variable
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(no_such_variable * am))
#> Error in eval_tidy(facet, mask): object 'no_such_variable' not found

# special case: raises a friendlier error when the expression is a symbol
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(no_such_variable))
#> Error: At least one layer must contain all faceting variables: `no_such_variable`.
#> * Plot is missing `no_such_variable`
#> * Layer 1 is missing `no_such_variable`
#> * Layer 2 is missing `no_such_variable`

# a data pronoun can be handled
ggplot(mtcars, aes(mpg, cyl)) +
  geom_point() +
  geom_vline(xintercept = 20) +
  facet_wrap(vars(.data$am))
```

<sup>Created on 2020-01-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### Minor change

* This PR removes `env` argument of `eval_facet()`. This is not necessary, but done in order to avoid confusion. I believe all facet specs are converted to quosures and bare expressions are not allowed, which means `env` is always ignored on `eval_tidy()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -168,6 +168,9 @@ fail.
   stepwise addition of individual partial themes is now equivalent to
   addition of multple theme elements at once (@clauswilke, #3039).
 
+* Facets now don't fail even when some variable in the spec are not available
+  in all layers (@yutannihilation, #2963).
+
 # ggplot2 3.2.1
 
 This is a patch release fixing a few regressions introduced in 3.2.0 as well as
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -436,14 +436,11 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
     return(out)
   }
 
-  # Create an environment for data mask
+  # Key idea: use active bindings so that column names missing in this layer
+  # but present in others raise a custom error
   env <- new_environment(data)
-
-  # Bind all possible column names that remains undefined to raise a custom error
-  # so that we can detect and ignore the case when a variable is missing from the
-  # layer data but exists in other layer
   missing_columns <- setdiff(possible_columns, names(data))
-  undefined_error <- function(e) abort("", class = "ggplot2_undefined_aes_error")
+  undefined_error <- function(e) abort("", class = "ggplot2_missing_facet_var")
   bindings <- rep_named(missing_columns, list(undefined_error))
   env_bind_active(env, !!!bindings)
 
@@ -451,11 +448,9 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
   mask <- new_data_mask(env)
   mask$.data <- as_data_pronoun(mask)
 
-  # Do not treat the cases as errors when it refers to a column name unavailable
-  # in the layer data
   tryCatch(
     eval_tidy(facet, mask),
-    ggplot2_undefined_aes_error = function(e) NULL
+    ggplot2_missing_facet_var = function(e) NULL
   )
 }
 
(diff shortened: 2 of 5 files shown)


yutannihilation:
May I merge this? I'll keep this open for a day or so in case we need further discussion. Feel free to merge if you need to run revdepchecks early.

thomasp85:
You can merge at your leisure - we won't run revdeps before after the conf anyway

yutannihilation:
Sure.

Issue #3745, opened 2020-01-20
“Computation failed in `stat_count()` when using loadRDS


nick-youngblut:
I often save my ggplot objects with `saveRDS()` so that I can load & plot them latter in new sessions. This works quite well, except when I use `geom_bar()`, which creates the following warning:

```
Warning message:
“Computation failed in `stat_count()`:
Please use `data_frame()` or `new_data_frame()` instead of `data.frame()` for better performance. See the vignette "ggplot2 internal programming guidelines" for details.”
```

...and results in an empty plot. 

### Session 1:
```
library(dplyr)
library(ggplot2)
out_file = 'ggplot_obj.RDS'
mpg %>%
    ggplot(aes(class)) +
    geom_bar() %>%
    saveRDS(file=out_file)
```

### Session 2:
```
library(dplyr)
library(ggplot2)
out_file = 'ggplot_obj.RDS'
p = loadRDS(out_file)
plot(p)
```

## sessionInfo

```
R version 3.6.2 (2019-12-12)
Platform: x86_64-conda_cos6-linux-gnu (64-bit)
Running under: Ubuntu 18.04.3 LTS

Matrix products: default
BLAS/LAPACK: /ebio/abt3_projects/Georg_animal_feces/envs/tidyverse/lib/libopenblasp-r0.3.7.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] RColorBrewer_1.1-2 ggplot2_3.2.1      tidyr_1.0.0        dplyr_0.8.3       

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.3       plyr_1.8.5       pillar_1.4.3     compiler_3.6.2  
 [5] base64enc_0.1-3  tools_3.6.2      zeallot_0.1.0    digest_0.6.23   
[...]

PR #3758, opened 2020-01-23
Move the check on misuse of data.frame() to tests

yutannihilation:
#3526 introduced the static analysis of codes here.

https://github.com/tidyverse/ggplot2/blob/214f3148d8a9a25ce80859645dbef38f9632b4fa/tests/testthat/test-conditions.R#L3-L21

This approach seems very clean and extensive because this catches the misuse of the prohibited function even if it's on the rare code path. I think the check on `data.frame()` should be done in the same way instead of overwriting the function. One bonus of this is that we'll be less likely to face such mysterious issues as #3745.


clauswilke:
I like this approach, but I'd like to suggest that we should also test the test (i.e., include a true positive). Add a file to the testthat directory that contains the prohibited functions and check that the counts of prohibited functions for that file are correct.

yutannihilation:
Thanks, I added some simple tests.

thomasp85:
@yutannihilation can we merge this in in its current form?

yutannihilation:
Yes, since this is just about testing, I think it's safe to merge this.

thomasp85 (approved):
LGTM

yutannihilation:
Thanks

PR #3759, opened 2020-01-24
Faster scale_apply

thomasp85:
When handling large datasets (>1e6 rows), scale_apply ends up being the main bottleneck (around 50% of the rendering time) This is all due to a call to `factor()` on the vector identifying which positional scale each row belongs two (this happens 8 time in total, as the scales are trained and mapped twice for both x and y).

This PR simplifies the factor construction as well as bypass it completely in the case of only a single scale (which is the predominant case). It speeds up the step 1000-fold for single panel plots and 2-3 fold otherwise

### DIFF
--- a/R/layout.R
+++ b/R/layout.R
@@ -296,10 +296,7 @@ scale_apply <- function(data, vars, method, scale_id, scales) {
     abort("`scale_id` must not be `NA`")
   }
 
-  scale_index <- unname(split(
-    seq_along(scale_id),
-    factor(scale_id, levels = seq_along(scales))
-  ))
+  scale_index <- split_with_index(seq_along(scale_id), scale_id, length(scales))
 
   lapply(vars, function(var) {
     pieces <- lapply(seq_along(scales), function(i) {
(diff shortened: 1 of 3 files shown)


hadley:
This is a v. crude comparison, but it suggests that vctrs is unlikely to get us much more performance here:

``` r
x <- 1:1e7
n <- 5
g <- sample(1:n, length(x), replace = TRUE)

split_thomas <- function(x, g, n) {
  g <- as.integer(g)
  attributes(g) <- list(levels = as.character(seq_len(n)), class = "factor")
  unname(split(x, g))
}

bench::mark(
  unname(split(x, factor(g))),
  split_thomas(x, g, n = n),
  vctrs::vec_group_loc(g),
  check = FALSE
)
#> # A tibble: 3 x 6
#>   expression                       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unname(split(x, factor(g)))    552ms    552ms      1.81   471.3MB     3.62
#> 2 split_thomas(x, g, n = n)      158ms    224ms      4.76    76.3MB     4.76
#> 3 vctrs::vec_group_loc(g)        160ms    176ms      5.72   179.9MB     3.81
```

<sup>Created on 2020-01-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #3764, opened 2020-01-29
Fix single value detection in has_flipped_aes()

yutannihilation:
These lines are to detect the case "the axes is a single value," but, in actual, they check whether all values are a fixed value`1` or not.

https://github.com/tidyverse/ggplot2/blob/5e6b1e5daa1599661bf5841353a067e9eee88161/R/utilities.r#L594-L600

This causes different behaviours between the case of `1` and that of another value. This PR fixes the check to `zero_range(range())`.

``` r
library(ggplot2)

df <- data.frame(value = 1:3, group = c("a", "a", "b"))
l <-  stat_summary(fun = "mean", geom = "bar", position = "dodge")
p1 <- ggplot(df, aes(value, 1, fill = group)) + l
p2 <- ggplot(df, aes(value, 0, fill = group)) + l

patchwork::wrap_plots(p1, p2, ncol = 1)
```

<sup>Created on 2020-01-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/utilities.r
+++ b/R/utilities.r
@@ -592,10 +592,10 @@ has_flipped_aes <- function(data, params = list(), main_is_orthogonal = NA,
     return(y_is_int != main_is_continuous)
   }
   # Is one of the axes a single value
-  if (all(x == 1)) {
+  if (zero_range(range(x, na.rm = TRUE))) {
     return(main_is_continuous)
   }
-  if (all(y == 1)) {
+  if (zero_range(range(y, na.rm = TRUE))) {
     return(!main_is_continuous)
   }
   # If both are discrete like, which have most 0 or 1-spaced values


thomasp85:
The I tend of that check is really to catch when boxplots are created with no x-aesthetic at which point they’ll get a 1 automatically. I’ll have to think about the repercussions of this change a bit and whether it is the right way to go about it

yutannihilation:
Got it, thanks for the context. Then, if this requires some time to consider, should I close this and resend against the master? This pull request is just that I was not immediately sure what the comment says.

thomasp85:
Your reprex suggest that the check is too eager, not the other way around... I'll experiment a bit

yutannihilation:
> Your reprex suggest

To be clear, my reprex in the description shows the problematic behaviour of the current RC version. Sorry if it's confusing.

thomasp85:
Closing in favour of #3800

PR #3768, opened 2020-01-31
improve documentation for xlim and ylim in limits.r

### LINKED ISSUE #3701: Improve documentation of arguments to `xlim`/`ylim`
Currently, `xlim` and `ylim` are documented together with `lims`.  Thus, the argument `...` to all three functions is described as:

    A name-value pair. The name must be an aesthetic, and the value must be either a length-2 numeric, a character, a factor, or a date/time.

Which is not correct for `xlim`/`ylim`.

Suggested text:

```
For `xlim` and `ylim`: Two numeric values, specifying the left/lower limit and the right/upper limit of the scale. If the larger value comes first, the scale will be reversed. You can leave one value as NA to compute from the range of the data.

For `lims`: A name-value pair. The name must be an aesthetic, and the value must be either a length-2 numeric, a character, a factor, or a date/time.

A numeric value will create a continuous scale. If the larger value comes first, the scale will be reversed. You can leave one value as NA to compute from the range of the data.

A character or factor value will create a discrete scale.

A date-time value will create a continuous date/time scale.
```

Alternative would be to move some of this to the details section.

sushmitavgopalan16:
closes #3701

### DIFF
--- a/R/limits.r
+++ b/R/limits.r
@@ -4,11 +4,18 @@
 #' individual scales. Note that, by default, any values outside the limits
 #' will be replaced with `NA`.
 #'
-#' @param ... For `xlim` and `ylim`: Two numeric values, specifying the left/lower limit and the right/upper limit of the scale. If the larger value comes first, the scale will be reversed. You can leave one value as NA to compute from the range of the data.
-#'For `lims`: A name-value pair. The name must be an aesthetic, and the value must be either a length-2 numeric, a character, a factor, or a date/time.
-#'A numeric value will create a continuous scale. If the larger value comes first, the scale will be reversed. You can leave one value as NA to compute from the range of the data.
-#'A character or factor value will create a discrete scale.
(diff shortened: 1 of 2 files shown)


R/limits.r
```diff
@@ -4,16 +4,12 @@
 #' individual scales. Note that, by default, any values outside the limits
 #' will be replaced with `NA`.
 #'
-#' @param ... A name-value pair. The name must be an aesthetic, and the value
-#'   must be either a length-2 numeric, a character, a factor, or a date/time.
+#' @param ... For `xlim` and `ylim`: Two numeric values, specifying the left/lower limit and the right/upper limit of the scale. If the larger value comes first, the scale will be reversed. You can leave one value as NA to compute from the range of the data.
```
clauswilke: Could you please manually word-wrap the text so that individual lines fit onto the screen? And also, for all lines but the one starting with `@param`, please add two spaces of indentation.

PR #3772, opened 2020-01-31
updating the manual word wrap

melimore86:
updating manual word wrap as per Claus recommendation, updating text for stat_density, the trim argument, found text in https://www.rdocumentation.org/packages/ggplot2/versions/1.0.1/topics/stat_density


clauswilke:
I think there was a misunderstanding. Can you track me down and we talk it over in person.

Issue #3756, opened 2020-01-22
Improve geom_histogram documentation


thomasp85:
Mention the possible use of `scale_x_binned()` + `geom_bar()` as an alternative to `geom_histogram()`, with tick marks in between the bars. Also add an example of it to the examples

yutannihilation:
This seems already fixed by #3773.

PR #3773, opened 2020-01-31
Added description of scale_x_binned to geom_histogram

kelseymoty:
Fixes Issue #3756


kelseymoty:
@thomasp85 those changes should be added now!

PR #3775, opened 2020-01-31
Change structure of error message to correct grammar

### LINKED ISSUE #3602: Small grammar error in error messages
As error messages are parametrised, some end up having a grammatical error:

```r
theme(axis.line.x = "wrong!")
#> Error in (function (el, elname)  : 
#>   Element axis.line.x must be a element_line object.
```

Should be: *Element axis.line.x must be **an** element_line object*. This could be solved by checking the first character after the ' a ', although it's dependent on the sound, not the vowel itself, and it would go wrong e.g. here:

```r
theme(legend.spacing.x = "wrong!")
#> Error in (function (el, elname)  : 
#>   Element legend.spacing.x must be a unit object.
```
... which is of course already correct.

Maybe then only check for `^(a|e|i|o)`? Or perhaps change to:
```
Element {...} must be an object of type {...}.
```

wendtke:
Fixes #3602

### DIFF
--- a/R/theme-elements.r
+++ b/R/theme-elements.r
@@ -524,7 +524,7 @@ validate_element <- function(el, elname, element_tree) {
     if (!is.unit(el) && length(el) == 4)
       abort(glue("Theme element `{elname}` must be a unit vector of length 4."))
   } else if (!inherits(el, eldef$class) && !inherits(el, "element_blank")) {
-      abort(glue("Element `{elname}` must be an object of type `{eldef$class}`."))
+      abort(glue("Theme element `{elname}` must be an object of type `{eldef$class}`."))
   }
   invisible()
 }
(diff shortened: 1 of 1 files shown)


R/theme-elements.r
```diff
@@ -524,7 +524,7 @@ validate_element <- function(el, elname, element_tree) {
     if (!is.unit(el) && length(el) == 4)
       abort(glue("Theme element `{elname}` must be a unit vector of length 4."))
   } else if (!inherits(el, eldef$class) && !inherits(el, "element_blank")) {
-      abort(glue("Theme element `{elname}` must be an `{eldef$class}` object."))
+      abort(glue("Element `{elname}` must be an object of type `{eldef$class}`."))
```
clauswilke: Please write "Theme element", not just "Element", as in the current code.

Issue #3533, opened 2019-09-19
Suggestion: improve error message on wrong use of scale_x_datetime


adisarid:
The following code yields an error message 
```
date_issues <- tibble(dates = lubridate::ymd(c("2019-09-19", "2019-09-20", "2019-09-21")), values = 1:3)

ggplot(date_issues, aes(x = dates, y = values)) + 
   geom_line() + 
   scale_x_datetime(date_minor_breaks = "1 day")
```
The error is:
```
Error: Invalid input: time_trans works with objects of class POSIXct only
```

The reason is that we are using the wrong scale function. We should be using `scale_x_date` instead of `scale_x_datetime`. 

A "suggestion flavored" error message would help the user understand and fix the issue. E.g.:

```
Error: Invalid input: time_trans works with objects of class POSIXct only. 
   Did you consider using scale_x_date() instead of scale_x_datetime()?
```

lazappi:
I have had a go at doing this in a branch here https://github.com/lazappi/ggplot2/tree/issue-3533. My solution was to catch the error in `ScaleContinuous` and modify the message. I could submit this as a pull request but not sure what the developers think of the solution? My main concern is that if the error occurs for some other reason then the new message won't be appropriate.

PR #3777, opened 2020-01-31
Added helpful error message for scale_x_datetime

kelseymoty:
Possible solution for addressing issue #3533 to point users to `scale_x_date()`


thomasp85:
We need to think more about this. Relying on the exact wording of the error message is going to be too brittle, but there is currently no way to implement it as trans objects are not subclassed. We will revisit this when the required changes has been made to the scales package

PR #3779, opened 2020-01-31
ggplot2#3626 minor text issue fix

melimore86:
Updating text of stat_density to text located in https://www.rdocumentation.org/packages/ggplot2/versions/1.0.1/topics/stat_density. Updated text to show the TRUE and FALSE parameters in the argument `trim`.


melimore86:
fixes issue ggplot2 3626

Issue #3455, opened 2019-07-25
filled arrowhead not showing in legend


RABxx:
As discussed here: https://stackoverflow.com/questions/57193353/show-filled-arrow-in-legend-ggplot

I am trying to plot a line segment with an arrow on the end, and make it appear in the legend. I can do this with the following code:

```
# sample data
dat <- data.frame(
  x = as.factor(1:10),
  y = c(20,30,13,37,12,50,31,2,40,30),
  z = rep('a', 10)
)

# basic plot
ggplot(dat) +
  geom_segment(
    aes(x = x, xend = x, y = 0, yend = y+15, linetype = z), 
    arrow = arrow(length = unit(0.25, 'cm'), type = 'closed'),
    size = 0.7
  ) 
```

but the arrowhead is empty, not filled. I tried overriding the `aes` in a `guide_legend` argument, but nothing changed. Is this a bug? Or am I missing something for how to make the arrowhead filled?

output:

![image](https://user-images.githubusercontent.com/48337093/61849944-3ec81a80-aef6-11e9-9ae0-48932c0a0b75.png)

clauswilke:
It's a bug. The legend drawing code doesn't set `fill`:
https://github.com/tidyverse/ggplot2/blob/141c77537ea4f54a7949ed8f23fd02c50172960e/R/legend-draw.r#L138-L143
Unlike the actual line drawing code:
https://github.com/tidyverse/ggplot2/blob/141c77537ea4f54a7949ed8f23fd02c50172960e/R/geom-segment.r#L119-L126

As a workaround, you can copy the legend-drawing code, add a line that sets the `fill` value, and then use that as your custom legend glyph as is possible with the latest ggplot2 release.

RABxx:
thanks for the response. I'm not quite sure how to do that last step though, can you give example code?

clauswilke:
I provided an example on SO.

PR #3781, opened 2020-01-31
Pass fill value to legend key icons

808sAndBR:
Resolves: #3455 

Below is reprex of this working:

```
# sample data
dat <- data.frame(
  x = as.factor(1:10),
  y = c(20,30,13,37,12,50,31,2,40,30),
  z = rep('a', 10)
)

# basic plot
ggplot(dat) +
  geom_segment(
    aes(x = x, xend = x, y = 0, yend = y+15, linetype = z), 
    arrow = arrow(length = unit(0.25, 'cm'), type = 'closed'),
    size = 0.7
  ) 
```

New output:
image

Old output:
image

### DIFF
--- a/R/legend-draw.r
+++ b/R/legend-draw.r
@@ -136,12 +136,12 @@ draw_key_path <- function(data, params, size) {
 
   segmentsGrob(0.1, 0.5, 0.9, 0.5,
     gp = gpar(
-       col = alpha(data$colour %||% data$fill %||% "black", data$alpha),
-       fill = alpha(params$arrow.fill %||% data$colour
-                    %||% data$fill %||% "black", data$alpha),
-       lwd = (data$size %||% 0.5) * .pt,
-       lty = data$linetype %||% 1,
-       lineend = "butt"
+      col = alpha(data$colour %||% data$fill %||% "black", data$alpha),
+      fill = alpha(params$arrow.fill %||% data$colour
+                   %||% data$fill %||% "black", data$alpha),
+      lwd = (data$size %||% 0.5) * .pt,
+      lty = data$linetype %||% 1,
+      lineend = "butt"
     ),
     arrow = params$arrow
   )
--- a/R/legend-draw.r
+++ b/R/legend-draw.r
@@ -137,6 +137,7 @@ draw_key_path <- function(data, params, size) {
   segmentsGrob(0.1, 0.5, 0.9, 0.5,
     gp = gpar(
       col = alpha(data$colour %||% data$fill %||% "black", data$alpha),
+      fill = alpha(data$colour %||% data$fill %||% "black", data$alpha),
       lwd = (data$size %||% 0.5) * .pt,
       lty = data$linetype %||% 1,
       lineend = "butt"
--- a/R/legend-draw.r
+++ b/R/legend-draw.r
@@ -137,7 +137,7 @@ draw_key_path <- function(data, params, size) {
   segmentsGrob(0.1, 0.5, 0.9, 0.5,
     gp = gpar(
       col = alpha(data$colour %||% data$fill %||% "black", data$alpha),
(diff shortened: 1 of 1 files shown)


clauswilke:
I have two concerns about this:

1. `geom_segment()` has a parameter `arrow.fill`. We should see if we can carry it through to the legend and use if defined.

2. I'm wondering whether this draw key is used by any geoms that meaningfully use a `fill` aesthetic, and whether in this case filling the arrow in the way it is done now will create strange results. Possibly we need a new draw key specifically for `geom_segment()`.

thomasp85:
I don't think we can do this for geom_segment specifically. Any segment and path type geom can have arrowheads and they should behave as expected.

Your concern about the arrow.fill argument is valid though... is there a straightforward way to pass this through?

808sAndBR:
I believe that this new version passes the arrow.fill through as expected, but let me know if it looks questionable.

For the second concern. I ran this version of legend_draw on everywhere that I saw importing it:

```
geom_sf
geom_errorbar
geom_path
geom_segment
geom_rug
geom_hline
```
None of the outputs seems strange to me, but I have them in a Rmd if you want to swing by and see if you notice anything strange.

clauswilke:
Yes, ignore my 2nd concern, it doesn't apply.

However, for my 1st concern, a little more needs to be done. First, `arrow.fill` should be the first choice for `fill`, not the second, as it is the case in the geom. See here:
https://github.com/tidyverse/ggplot2/blob/7e6d1259237d9a2b48b8156e62ada2c47c411f31/R/geom-segment.r#L116
and here:
https://github.com/tidyverse/ggplot2/blob/7e6d1259237d9a2b48b8156e62ada2c47c411f31/R/geom-segment.r#L121

Second, I doubt `data$arrow.fill` will work. I think it has to be `params$arrow.fill`. In either case, I think it's important that you run a test to check whether this actually works.

Finally, I wouldn't discount the fill variable, so the final sequence should read:
`params$arrow.fill %||% data$colour %||% data$fill %||% "black"`

808sAndBR:
Thanks for all the extra info, it's been super helpful for understanding how this all works behind the scenes.

Ahhhh, I see I think it picking up "black" from the back of the options which made me think it was working as expected. 

I have added some more robust test cases (below) to  make sure this is working as expected. I think it looks right to me, but let me know if anything seems off.

```{r}
# sample data
dat <- data.frame(
  x = as.factor(1:10),
  y = c(20,30,13,37,12,50,31,2,40,30),
  z = rep('a', 10)
)

# basic plot
ggplot(dat) +
  geom_segment(
    aes(x = x, xend = x, y = 0, yend = y+15, linetype = z), 
    arrow = arrow(length = unit(0.25, 'cm'), type = 'closed'),
    size = 0.7
  ) 
```

image

```{r}
ggplot(dat) +
  geom_segment(
    aes(x = x, xend = x, y = 0, yend = y+15, linetype = z), 
    arrow = arrow(length = unit(0.25, 'cm'), type = 'closed'),
    arrow.fill = 'red',
    size = 0.7
  ) 
```
image

```{r}
ggplot(dat) +
  geom_segment(
    aes(x = x, xend = x, y = 0, yend = y+15, linetype = z, colour='yellow'), 
    arrow = arrow(length = unit(0.25, 'cm'), type = 'closed'),
    size = 0.7
  ) 
```

image

```{r}
ggplot(dat) +
  geom_segment(
    aes(x = x, xend = x, y = 0, yend = y+15, linetype = z, colour='yellow'), 
    arrow = arrow(length = unit(0.25, 'cm'), type = 'closed'),
    arrow.fill = 'blue',
    size = 0.7
  ) 
```

image

clauswilke:
Looks great, thanks! 

Going forward, I suggest you check out the [reprex package,](https://reprex.tidyverse.org/articles/articles/magic-reprex.html) so you don't have to copy and paste screenshots.

PR #3790, opened 2020-02-01
adds small section to ggsave() on writing to device

### LINKED ISSUE #3755: add section on writing to a device in ggsave doc
A small section describing how you would save a ggplot without using `ggsave()` might make sense, as it will help users in situations where ggsave cannot be used (e.g. if wanting to use a special graphic device)

djnavarro:
Inserts a short section in the ggsave docs on saving graphics files without ggsave. Fixes #3755


R/save.r
```diff
@@ -13,6 +13,16 @@
 #' containing the `%` sign, use `%%`. For example, `filename = "figure-100%%.png"`
 #' will produce the filename `figure-100%.png`.
 #'
+#' @section Writing to a graphics device:
+#'
+#' In most cases `ggsave()` is the simplest way to save your plot, but
+#' sometimes it is necessary to save the plot by writing directly to a
+#' graphics device. For example, to create a PNG file, you could first
```
clauswilke: I think there's a small break in logic here. "...sometimes it is necessary to save the plot ... For example, ..." At this point I expect an example of *why* it is necessary to save the plot, but instead I'm given an example of how to get this done.

You could either provide an example of why one might want to do this, or write a better transition. To write a better transition, you could add: "In this case, you can just open a regular R graphics device, print the plot, and then close the device. For example, ..."

R/save.r
```diff
@@ -13,6 +13,16 @@
...
+#' open the device with a command like `png("mtcars.png")`. Next, you
+#' would create the ggplot with something like
+#' `ggplot(mtcars, aes(mpg, wt)) + geom_point()` and then use `dev.off()`
```
clauswilke: Should this state explicitly that the plot needs to be printed? `p <- ggplot(...); print(p)`

clauswilke:
Also, please merge in master so the CI doesn't break anymore.

R/save.r
djnavarro: The updated version uses a transition similar to your suggestion, and moves the code illustrating the logic to the examples section. I think this version is clearer than the previous one

R/save.r
djnavarro: Oops, yes. The revised version does this explicitly in the examples section

djnavarro:
I may have made a mistake with the merging: I'm still learning the workflow. What I _attempted_ to do is fetch from upstream/master, merge the upstream changes into my write_to_device branch along with my changes. But (a) I'm not 100% sure that this is what you meant, and (b) I may have made a mistake when trying to do this.

My sincere apologies if I have made an error :slightly_smiling_face:

PR #3800, opened 2020-02-06
Make sure boxplot check doesn't polute other stats/geoms

thomasp85:
Superseedes #3764 

This PR introduces a new switch in `has_flipped_aes()` which confines the single value check to those requesting it (currently only boxplot)

### DIFF
--- a/R/stat-boxplot.r
+++ b/R/stat-boxplot.r
@@ -59,7 +59,9 @@ StatBoxplot <- ggproto("StatBoxplot", Stat,
   },
 
   setup_params = function(data, params) {
-    params$flipped_aes <- has_flipped_aes(data, params, main_is_orthogonal = TRUE, group_has_equal = TRUE)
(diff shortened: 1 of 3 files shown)


yutannihilation:
Ah, now I understand your comment on #3764. The example I used was the one that should *not* be flipped, and I confirmed this PR handles them correctly. Thanks.

``` r
df <- data.frame(value = 1:3, group = c("a", "a", "b"))

#> Loading ggplot2

l <-  stat_summary(fun = "mean", geom = "bar", position = "dodge")
p1 <- ggplot(df, aes(value, 1, fill = group)) + l
p2 <- ggplot(df, aes(value, 0, fill = group)) + l

patchwork::wrap_plots(p1, p2, ncol = 1)
```



<sup>Created on 2020-02-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
I have no idea how I convinced myself it defaulted to `1` 🤷‍♂️

PR #3801, opened 2020-02-06
Catch errors when applying after_scale mods to legend keys

### LINKED ISSUE #3729: Legends should not use after_scale()ed values
(Originally commented at https://github.com/tidyverse/ggplot2/issues/3725#issuecomment-573745316)

Currently, legends reflect the value of `after_scale()`. But, it's not always possible because the expression of `after_scale()` might refer to the variable unavailable at the stage of drawing legends.

For example, I think it is a proper usage of `stage()` to modify scaled `colour`s by combining other aesthetics. But, the following example has two problems:

1. `x` is not available on drawing legends
2. even if we manage to make `x` available, it's not clear what legends should be drawn because the scaled `colour`s are discrete but the `after_scale()`ed values are continuous.

Especially because of the problem 2., I think legends should use the raw scaled values, not the `after_scale()`ed ones.

``` r
library(ggplot2)

set.seed(88)
d <- data.frame(x = runif(9), y = runif(9), g = rep(c("a", "b", "c"), each = 3))
p <- ggplot(d) +
  geom_point(aes(x, y, colour = stage(g, after_scale = alpha(colour, x))), size = 10)

# fail
p
#> Error in alpha(colour, x): object 'x' not found

# ok
p + theme(legend.position = "none")
```

<sup>Created on 2020-01-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Fixes #3729 

This PR catches any errors happening when applying `after_scale()` modifications to legend keys. If it errors it will simply discard them all with a warning.

It may be done more fine grained, but not without a lot of weird refactoring of the code to accommodate this edge-case. This PR fixes the surprise of the legend disappearing and gives a meaningful warning. If we get a lot of requests for better support (I doubt it) we can revisit it in the future

### DIFF
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -200,7 +200,16 @@ guide_geom.bins <- function(guide, layers, default_mapping) {
       n <- vapply(layer$aes_params, length, integer(1))
       params <- layer$aes_params[n == 1]
 
-      data <- layer$geom$use_defaults(guide$key[matched], params)
+      aesthetics <- layer$mapping
+      modifiers <- aesthetics[is_scaled_aes(aesthetics) | is_staged_aes(aesthetics)]
+
+      data <- tryCatch(
+        layer$geom$use_defaults(guide$key[matched], params, modifiers),
+        error = function(...) {
+          warn("Failed to apply `after_scale()` modifications to legend")
+          layer$geom$use_defaults(guide$key[matched], params, list())
+        }
+      )
     } else {
       data <- layer$geom$use_defaults(NULL, layer$aes_params)[rep(1, nrow(guide$key)), ]
     }
--- a/R/guide-legend.r
+++ b/R/guide-legend.r
@@ -262,7 +262,13 @@ guide_geom.legend <- function(guide, layers, default_mapping) {
       aesthetics <- layer$mapping
       modifiers <- aesthetics[is_scaled_aes(aesthetics) | is_staged_aes(aesthetics)]
 
-      data <- layer$geom$use_defaults(guide$key[matched], params, modifiers)
+      data <- tryCatch(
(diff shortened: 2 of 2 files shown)


yutannihilation:
Looks good, but let me confirm.

> This PR fixes the surprise of the legend disappearing and gives a meaningful warning.

This doesn't fix "the legend disappearing," but fixes the drawing failure; I pointed two problems on #3729 and the second case is not fixed by this PR. Is this what you intended?

``` r
set.seed(88)
d <- data.frame(x = runif(9), y = runif(9), g = rep(c("a", "b", "c"), each = 3))

#> Loading ggplot2

# Case 1: fails with the current master
ggplot(d) +
  geom_point(aes(x, y, colour = stage(g, after_scale = alpha(colour, x))), size = 10)
#> Warning: Failed to apply `after_scale()` modifications to legend
```



``` r

# Case 2: does not fail, but loses the colours
ggplot(d) +
  geom_point(aes(x, y, size = x + y,
                 colour = stage(g, after_scale = ifelse(size > 4, colour, "grey"))))
```



<sup>Created on 2020-02-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
That was bad wording on my part. I got confused by the first example showing a plot without a legend...

Yes - this is the correct behaviour...

yutannihilation:
Thanks, then I'm approving this PR. I think your concern is valid.

> It may be done more fine grained, but not without a lot of weird refactoring of the code to accommodate this edge-case.

We would need to complicate things just to solve edge cases. We might need to consider this a bit deeper in future if many people use this in some extreme ways, but I agree this fix is the best at this point.

Issue #3717, opened 2020-01-09
better documentation for map related functions?


moodymudskipper:
Here are several comments, They're linked but I'm happy to split them into several issues if it's preferable.

## `geom_map()`

The titles and descriptions of `geom_map` are : 

> Polygons from a reference map
> This is pure annotation, so does not affect position scale

I think this could be clearer, `geom_map` wraps the functionalities of `geom_polygon` to plot maps. the difference being that we provide a reference map through the `map` argument, which takes care of the `x` and `y` aesthetics and allows us to use a more convenient `data` argument.

The description of the `map` argument doesn't mention the `map_data()` function (only an example does), tough it will probably be useful in most cases. Instead it says : 

> Data frame that contains the map coordinates. This will typically be created using fortify() on a  spatial object. It must contain columns x or long, y or lat, and region or id.

## `fortify()`

If we follow to fortify's page we read : 

> Rather than using this function, I now recommend using the broom package, which implements a much wider range of methods. fortify may be deprecated in the future.

It doesn't say what `fortify()` does, says we shouldn't use it anyway, and refers to *broom* without telling us which functions from *broom* we should use (we might think it's just `broom::fortify()` but it doesn't exist).

## `map_data()`

The `map` argument is described as : 

> name of map provided by the maps package. These include maps::county(), maps::france(), maps::italy(), maps::nz(), maps::state(), maps::usa(), maps::world(), maps::world2().

But none of these functions exist.

From examples we can infer that we can use a string such as "state", but the doc seems to hint that we might have more choice than what's in this list, and doesn't tell us where to look. The following argument hints that we can look at `?maps::map`, and there we find that we can infer the list of available maps by looking `help(package='maps')`. I think it is very convoluted and we would benefit from having these maps enumerated in the *ggplot2* doc.

## `coord_map()`

The `projection` argument is forwarded to `mapproj::mapproject` so I understand that you decide to redirect the user there, but again, once there we need to read and scroll to get a list of possible values and it all adds up for a user who "just wants to draw a map", as this one : https://twitter.com/calvinmdewilde/status/1215116524798636032.

So here as well, could we have in *ggplot2*'s doc the list of available projections ?

## a map vignette ?

[...]

PR #3802, opened 2020-02-06
Update docs for fortify

bkkkk:


yutannihilation:
I agree `fortify()`'s doc needs to be updated. But, maybe we need to discuss about the current status of `fortify()` first? I think `fortify()` is still useful as part of the ggplot2's API; it's just not useful for general use of converting any object to data.frame.

bkkkk:
@yutannihilation this was brought up in issue #3717, I thought I’d get a legs up on resolving part of it. I agree that the whole fortify family in ggplot2 would benefit from a look-over. How should we proceed? Maybe create a separate issue from #3717 to hold this discussion?

yutannihilation:
Hmm, sorry I think I need some time to answer...

Basically, there are roughly 3 types of objects that `fortify()` converts.

1. model object (I think this is why the first argument is named `model`)
2. sp object
3. data.frame-alikes (e.g. `grouped_df`)

I think category 1 is superseded by `broom::augment()` (for `lm` object) and `broom::tidy()` (for multcomp objects), and category 2 is by `geom_sf()`. What I said "is still useful as part of the ggplot2's API" above is category 3.

#3717 is only about sp objects, so we need another issue for broader discussion. Let me open a new one later.

Issue #3807, opened 2020-02-07
ggsave() can't be used with devices taking `file` as argument instead of `filename`


bwiernik:
In `ggsave()`, if `device` is specified and the device takes the `file` argument instead of `filename`, ggsave fails with an error: `unused argument (filename = filename)`

```r
library(ggplot)
p <- qplot(cars[[1]], cars[[2]])
ggsave("plot.pdf", p, device = pdf)
```

For the `grDevices` devices that take `file`, ggsave customizes the call to use `file` instead, but this doesn't work for devices from other packages, such as `devEMF::emf` or `Cairo::CairoSVG`.

This appears to be caused by these changes to remove partial argument matching.
https://github.com/tidyverse/ggplot2/issues/2355
https://github.com/tidyverse/ggplot2/pull/2364

It would be nice if ggsave automatically caught this error, then retried with `file` instead of `filename`.

PR #3810, opened 2020-02-07
Permit function devices to use `file` argument

bwiernik:
Devices like `devEMF::emf` or `Cairo::CairoSVG` use `file` instead of `filename`. They currently fail when provided in `device` to ggsave. This fixes that.

Closes https://github.com/tidyverse/ggplot2/issues/3807

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,9 @@ ggplot2 1.0.1
 * Improvements to order of colours and legend display for continuous colour
   scales extracted from colorbrewer palettes by `scale_*_distiller()` (@jiho, 1076)
 
+ * Support graphics devices that use the `file` argument instead of `fileneame` 
+   in `ggsave()` (@bwiernik, 3810)
+
 ggplot2 1.0.0
 ----------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


R/save.r
```diff
@@ -149,8 +149,12 @@ plot_dev <- function(device, filename = NULL, dpi = 300) {
   force(filename)
   force(dpi)
 
-  if (is.function(device))
-    return(device)
+  if (is.function(device)) {
+    if ("file" %in% names(formals(device))) {
+      dev <- function(filename, ...) device(file = filename, ...)
+      return(dev)
+    } else return(device)
```
yutannihilation: According to https://style.tidyverse.org/syntax.html#inline-statements:

> Function calls that affect control flow (like `return()`, `stop()` or continue) should always go in their own `{}` block:

```suggestion
    } else {
      return(device)
    }
```

bwiernik:
Okay, fixed the styling and added a NEWS bullet as requested.

NEWS
```diff
@@ -6,6 +6,9 @@ ggplot2 1.0.1
 * Improvements to order of colours and legend display for continuous colour
   scales extracted from colorbrewer palettes by `scale_*_distiller()` (@jiho, 1076)
 
+ * Support graphics devices that use the `file` argument instead of `fileneame` 
+   in `ggsave()` (@bwiernik, 3810)
+
```
yutannihilation: Ah, sorry, we are now using `NEWS.md` instead of `NEWS`. Could you move this to `NEWS.md`?

yutannihilation:
Thanks!

PR #3813, opened 2020-02-10
reference binwidth instead of width in documentation

evanhaldane:
`width` is deprecated, yet the documentation for using `center`/`boundary` still reference it instead of `binwidth`.


yutannihilation:

yutannihilation:
Thanks!

PR #3820, opened 2020-02-13
Bump scales dependency

zeehio:
Scales 1.0.0 fixed a bug https://github.com/r-lib/scales/pull/81. The bug can lead to wrong plots when they are defined inside for loops, because some scales 0.5.0 functions  evaluated some arguments when the plot was rendered instead of when the plot was created.

I contributed the fix for that bug some years ago and I hit myself with it last week because an old version of scales was still installed in a computer from a friend.

Most CI tests happen with the latest version of the pkg dependencies, so I guess it may make sense to upgrade to the latest 1.1.0, but that is up to you, I guess.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,7 @@ Imports:
     MASS,
     mgcv,
     rlang (>= 0.3.0),
-    scales (>= 0.5.0),
+    scales (>= 1.0.0),
     stats,
     tibble,
     withr (>= 2.0.0)


clauswilke:
We don't normally bump package dependencies just because an upstream package has a bug. Dependency constraints make statements about compatibility. I.e., `scales (>= 0.5.0)` means that ggplot2 cannot be used with a scales package of a lower version, because of some sort of API mismatch. (E.g., ggplot2 is calling a scales function that doesn't exist in prior versions of scales.)

zeehio:
Oh, I did not know about that policy. Thanks for the information.

I wrote a package (https://github.com/sipss/AlpsNMR) that Imports ggplot2 and it was using it in a way that generates wrong plots with scales 0.5.0 (no errors, just wrong colours if I recall correctly).

Either my package Imports `scales  >= 1.0.0` without using any of its functions directly (getting at least a NOTE) or I implement some version checking and ask users to upgrade scales if needed.

It's not a big issue, and there are lots of workarounds available. If you happen to have a recommended practice or suggestion for my situation I will be very happy to consider it. I thought bumping the dependency version in ggplot2 was the best for everyone, but I guess having a wide range of dependencies has its advantages as well.

If my use case is not convincing enough, feel free to close this PR or I will do it myself tomorrow otherwise. I understand your decision and I have alternatives to choose from.

clauswilke:
Yes, it sounds like scales >= 1.0 is a hard dependency for your package, so you should enforce it in your Description file. If you're not explicitly using any scales functions anywhere in your package, you'll have to manually import one (using `@importFrom`) to keep CRAN check happy.

Issue #3827, opened 2020-02-20
More configurable continuous colorscale defaults


cpsievert:
`scale_[fill/colour]_[continuous/binned]` have a `type` argument which provides an entrypoint for customizing the scale's default. Currently, `type` only supports `gradient` and `viridis`. It seems reasonable that it could also support an arbitrary function, which would make the default much more configurable. For example:

```r
library(ggplot2)
scale_colour_custom <- function(...) {
  scale_colour_gradient(..., low = "white", high = "green")
}
options(ggplot2.continuous.colour = scale_colour_custom)
qplot(1:10, 1:10, color = 1:10)
```

![image](https://user-images.githubusercontent.com/1365941/74889056-67f24180-5345-11ea-8c19-ea4ced2f27e7.png)

cpsievert:
Done in #3828

fcecinati:
The issue has been closed, but the example above is still not working with the provided syntax in my session (R4.0 ggplot2 3.3.1), returning the followign error:
```
> library(ggplot2)
> scale_colour_custom <- function(...) {
+   scale_colour_gradient(..., low = "white", high = "green")
+ }
> options(ggplot2.continuous.colour = scale_colour_custom)
> qplot(1:10, 1:10, color = 1:10)
Error in switch(type, gradient = scale_colour_gradient(...), viridis = scale_colour_viridis_c(...),  : 
  EXPR must be a length 1 vector
```

If a different syntax is needed, could you please provide an example?

Screenshot of my sessionInfo()
![image](https://user-images.githubusercontent.com/8179868/113259876-8b681980-92c5-11eb-87a6-dd2c8cd84916.png)

PR #3828, opened 2020-02-20
Type argument in scale_[fill/colour]_[continuous/binned] can accept a function

cpsievert:
See #3827


hadley:
This seems reasonable to me.

thomasp85:
@cpsievert can you add a small note in NEWS.md?

cpsievert:
@thomasp85 good to merge? Those build failures look like false positives

thomasp85:
Thanks!

PR #3832, opened 2020-02-23
Add axis-reversed stat_ecdf

jgjl:
Add an option to reverse axis y->x instead of x->y to acomodate tail-latency plots, i.e.: https://github.com/ixy-languages/ixy-languages/blob/master/img/latency-hdr-hist-1.png (context: https://github.com/ixy-languages/ixy-languages).
In this kind of plot the probability is plotted along the horizontal axis while the values are plotted along the vertical axis. Hence, in comparison to the default stat_ecdf function, the axis are reversed.
This specific step could also be achived by applying coord_flip. However, in this case the reversed-logarithmic transformation of the x-axis cannot be applied. Hence, in my understanding, switching axis in stat_ecdf is the only way to get there. 

I will supply the complementary transformation object for the x-axis separately outside of the ggplot2 package.

The proposed implementation is a little simplistic in that it just adds a version where input and output variables are reversed. An alternative would be to allow the input and output variables to be freely choosen. I am happy to supply such a version if it would be preferrable.


clauswilke:
Since ggplot2 3.3.0 (about to be released), we have a framework for this type of problem. Please see for example how `geom_density()`/`stat_density()` are now implemented.

jgjl:
stat_ecdf now uses the now aes switching scheme and was rebased on top on master.

clauswilke:
@jgjl Just an FYI: There's no need to rebase. Merge is fine. We squash-merge all PRs anyways.

jgjl:
> @jgjl Just an FYI: There's no need to rebase. Merge is fine. We squash-merge all PRs anyways.

Should I create a new PR without the rebase?

clauswilke:
No, no need to change this now. I just wanted to point out that rebasing is not necessary, since it can be a lot of unnecessary extra work.

clauswilke:
@jgjl I think you'll have to create a new PR. This PR now lists 60 commits and 49 files changed and it just looks like things went wrong somewhere along the way. Also, I suggest you open an issue first describing the feature you'd like to add, and then you can make a new PR implementing that feature.

PR #3833, opened 2020-02-24
Configurable qualitative color scales

### LINKED ISSUE #3826: Improvements to qualitative color scaling
`scale_[fill/color]_discrete` currently default to `scale_[fill/color]_hue`. While this is a great default when there are a large number of levels to encode, it's not a great default there are just a few (i.e., it's not colour-blind safe or aesthetically pleasing compared to carefully chosen qualitative palettes like colorbrewer or [Okabe-Ito](https://jfly.uni-koeln.de/color/)). In the event there isn't already a "blessed" ggplot2 scale (e.g., `scale_colour_brewer()`), the usual way to override the default is with a custom qualitative colorscale is to use `scale_[fill/color]_manual()`, but this also isn't ideal for situations where you don't want to make any assumptions about the number of levels to encode (i.e.,  `scale_[fill/color]_manual()` errors if there are more levels than color codes).

I'd like to propose that `scale_[fill/color]_discrete` gain a new argument, say `codes`, that would accept a vector of color codes. In the event `codes` is longer than the number of levels to encode, it would effectively be `scale_[fill/color]_manual`; otherwise, it would effectively be `scale_[fill/color]_hue`. Moreover, I think it would make sense for `codes` to default to something sensible like Okabe-Ito.

It seems the implementation for this would be relatively straight-forward ([see here](https://github.com/cpsievert/ggplot2/commit/bafa5361eb602bdee2cdfcfb97661d4eb9979bef) for a proof-of-concept). If there's interest in pursuing this idea, I'd be happy to start a PR.

cpsievert:
Closes #3826 by:

1. Allowing control over discrete colour scaling defaults (i.e., `scale_[fill/colour]_discrete()`) through a new `type` argument, which defaults to`ggplot2.discrete.[fill/colour]` options.
2. Adding the ability to supply a collection of color codes to `type` and using those codes if (and only if) there are more codes than levels. In addition to a character vector, you can also supply a list of vectors to `type` and it will choose the minimum length vector that can support the number of levels to encode, which provides a convenient way to switch the default palette based on the number of levels (see examples below).
~~3. Having `type` default to a colour-blind safe colour palette, such as Okabe-Ito. This will lead to 'breaking' changes, but those who wish to have the old behavior can do: `options(ggplot2.discrete.fill = scale_fill_hue, ggplot2.discrete.colour = scale_colour_hue)`~~

(**UPDATE:** this PR won't change the default, but a future major release likely will) 

### Examples

``` r
library(ggplot2)
# Template function for creating densities grouped by a variable
cty_by_var <- function(var) {
  ggplot(mpg, aes(cty, colour = factor({{var}}), fill = factor({{var}}))) +
    geom_density(alpha = 0.2)
}

# Define custom palettes for when there are 1-2, 3, or 4-6 levels
opts <- options(
  ggplot2.discrete.fill = list(
    c("skyblue", "orange"),
    RColorBrewer::brewer.pal(3, "Set2"),
    RColorBrewer::brewer.pal(6, "Accent")
  )
)
# Two levels
cty_by_var(year)
```

``` r
# Three levels (Set2)
cty_by_var(drv)
```

``` r
# Five levels (Accent)
cty_by_var(fl)
#> Warning: Groups with fewer than two data points have been dropped.
#> Warning in max(ids, na.rm = TRUE): no non-missing arguments to max; returning -
#> Inf
```

``` r
# Seven levels (scale_fill_hue)
cty_by_var(class)
```

``` r
options(opts)
```

<sup>Created on 2020-02-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### TODO

- [ ] Unit tests (visual and `scale_[fill/colour]_qualitative()`)


R/scale-hue.r
```diff
@@ -63,3 +62,120 @@ scale_fill_hue <- function(..., h = c(0, 360) + 15, c = 100, l = 65, h.start = 0
...
+# prismatic::check_color_blindness(okabeIto)
+okabeIto <- c(
+  "#E69F00",
+  "#56B4E9",
+  "#009E73",
+  "#F0E442",
+  "#0072B2",
+  "#D55E00",
+  "#CC79A7",
+  "#999999"
```
clauswilke: I've received feedback from colorblind people that gray is not that great of a choice and that black would be better. See also: https://github.com/jrnold/ggthemes/blob/master/data-raw/theme-data/colorblind.yml (And note black is the first color there. Not sure if I like it or not, but it definitely works better for people with CVD.)

cpsievert: Interesting, thanks. I did (somewhat selfishly) avoid using black as I was hoping for a default that could work well in both 'light' and 'dark' mode (so I can use it for the [upcoming auto-theming feature in shiny](https://github.com/rstudio/shiny/pull/2740)). Now that I think of it though, maybe it'd be better to switch between black/white based on the background color?  It'd be easy enough for me to do that for my purposes, but maybe that's worth doing in ggplot2 itself (i.e., use white instead of black if `plot.background` is dark)

clauswilke: This touches on something that I have said for a long time: I think color scales should be settable from the theme. That makes much more sense to me than setting them via global options.

clauswilke: See #2691.

yutannihilation: Can we agree to use black for now, and improve the support for dark theme after #2691 gets ready?

cpsievert: ok with me

thomasp85:
@clauswilke and @cpsievert where are we on this? If we want to have this in 3.3.1 we will need to get this finished off in a way where it is non-breaking (i.e. doesn't default to okabe-ito, but the regular behaviour). I'm fine with having a breaking change in this for the release after that.

Should we add the infrastructure in 3.3.1 or just wait for the whole PR until after the patch release?

clauswilke:
I think it's in @cpsievert's court to finish this off. He had indicated a one-line change that would make this non-breaking, but I think this change hasn't been applied yet: https://twitter.com/cpsievert/status/1255585845915955200?s=20

@cpsievert Could you commit this change and resolve the conflicts etc? And I'm still not sure whether this comment (https://github.com/tidyverse/ggplot2/issues/3826#issuecomment-589968166) needs further addressing.

DESCRIPTION
```diff
@@ -56,6 +56,7 @@ Suggests:
     nlme,
     profvis,
     quantreg,
+    RColorBrewer,
```
clauswilke, line 59: Is there a reason we need this? I thought we access the brewer scales through the {scales} package, which already depends on {RColorBrewer}.

cpsievert: It's to avoid an R CMD check warning (since my example is using `RColorBrewer`)

clauswilke: Ah, I see it now, yes, agreed.

cpsievert:
Ok, this should be ready for a closer look now

Issue #3835, opened 2020-02-25
Feature request: more lenient check_nondata_cols


teunbrand:
Currently data.frame columns are dropped when they don't satisfy the checks of `check_nondata_cols` which looks like the following:

```
check_nondata_cols <- function(x) {
  idx <- (vapply(x, function(x) is.null(x) || rlang::is_vector(x), logical(1)))
  names(x)[which(!idx)]
}
````

The reason I would like it to be more lenient is such that vector-like columns which technically aren't vectors could be accepted, specifically for the virtual `S4Vectors::Vector-class` (and non-virtual classes containing this). This would open up channels to have Bioconductor work with ggplot2 more naturally, provided some extra infrastructure is implemented at the scale/coordinate level (which could befit a ggplot extention package).

I would propose to change `check_nondata_cols` to the following:

```
check_nondata_cols <- function(x) {
  idx <- (vapply(x, function(x) is.null(x) || rlang::is_vector(x) || inherits(x, "Vector"), logical(1)))
  names(x)[which(!idx)]
}
```

Thanks you for reading and considering!

**Comment by clauswilke** [comment]
I don't think the intend was ever to exclude meaningful data columns. This was just to filter out nonsensical things, such as functions.

However, the question is why `rlang::is_vector(x)` fails. Maybe it requires modification? Could you provide a simple reprex where you create an object of a type that would fail the check but can be reasonably considered as a vector?

**Comment by teunbrand** [comment]
I think `rlang::is_vector()` (and also `base::is.vector()`) always return `FALSE` for many bioconductor S4 classes since their internal representation is not as a vector but most relevant data is stored in slots, akin to S3 attributes, that often behave in a parallel way to vectors.

Below is a minimal reprex, keep in mind though that the Bioconductor S4 `Vector` virtual class actually has methods that make it behave the way vectors behave, while this reprex does not.

``` r
library(ggplot2)

setClass("fakeinteger", representation = list(x = "integer"), 
         prototype = list(x = integer()))

y <- new("fakeinteger", x = 1:5)

ggplot2:::check_nondata_cols(list(y = y))
#> [1] "y"
```

<sup>Created on 2020-02-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Below is a more realistic reprex for two actual bioconductor classes that behave near identical to vectors.

``` r
# BiocManager::install("S4Vectors")
suppressPackageStartupMessages(library(S4Vectors))
library(ggplot2)

[...]

Issue #2609, opened 2018-05-15
Relax 1d atomic vectors or lists requirement


hadley:
As it's breaking many revdeps so will likely have deleterious impact on user code

hadley:
@lionel- can you take on this one? I'm not going to have time since I'm teaching Thursday and Friday.

The problem arises from the use of `as_tibble()` in three places. On the whole, I think the increased strictness of tibble is a good thing but it causes problems with some base R functions which 1d arrays like (like `scale()`, as in #2374). I think the simplest solution probably to switch to `tibble::new_tibble()`, but it requires a little thought.  Any solution should include a PR that ensures this works:

```R
df <- data.frame(x = 1:10)
df$y <- scale(1:10)
  
p <- ggplot(df, aes(x, y)) + geom_point()
layer_data(p)
```

Beware that `data.frame(x = scale(1:10))` strips the dim attribute that causes the problem.

It'd be great if you could check a couple of the revdeps that fail to make sure this actually fixes the problem (easiest way is to search https://docs.google.com/spreadsheets/d/1UZIFwt3j3XjnRGOMNJZvpal_CAj7Hz8aNXJt41m3JjM/edit#gid=0 for "atomic vector")

Issue #2374, opened 2017-12-18
Current development version breaks compatibility with scale() output, e.g. in qq-plots


jemus42:
Supplying `scale()`- standardised values to e.g. the `sample` aesthetic for the use of qq-plots does work in the current CRAN version, however, it does not work anymore with the current GitHub version.  
I looked through the `NEWS.md` to find any mention of a possible breaking change, but could not find anything that seemed to indicate that this change is intended.

Previous behavior:
```
# install.packages("ggplot2")

library(ggplot2)
ggplot(data = mtcars, aes(sample = scale(mpg))) +
  geom_qq()
```

Works fine.

Behavior with current GitHub version:

```
# devtools::install_github("tidyverse/ggplot2")

library(ggplot2)
ggplot(data = mtcars, aes(sample = scale(mpg))) +
  geom_qq()

# Error: Column `sample` must be a 1d atomic vector or a list
```

I'm aware that `scale` returns a 1d-matrix instead of an atomic vector, so I know I can circumvent this issue by using `as.numeric(scale(mpg))`, but this is simply a case of not wanting to change old code if I don't have to.

Tested versions of ggplot:

- CRAN: ` ggplot2      * 2.2.1      2016-12-30 CRAN (R 3.4.0) `
- GitHub: `ggplot2      * 2.2.1.9000 2017-12-18 Github (tidyverse/ggplot2@bfff1d8)`

batpigandme:
As you can see below, the difference between `scale(mtcars$mpg)` and `as.numeric(scale(mtcars$mpg))` is that the former includes attributes (scaled:center, and scaled:scale), while the latter does not. 

`scale` itself is actually comes from base, so I don't think it's something that can be changed here. 

``` r

scale(mtcars$mpg)
#>              [,1]
#>  [1,]  0.15088482
#>  [2,]  0.15088482
#>  [3,]  0.44954345
#>  [4,]  0.21725341
#>  [5,] -0.23073453
#>  [6,] -0.33028740
#>  [7,] -0.96078893
#>  [8,]  0.71501778
#>  [9,]  0.44954345
#> [10,] -0.14777380
#> [11,] -0.38006384
#> [12,] -0.61235388
#> [13,] -0.46302456
#> [14,] -0.81145962
#> [15,] -1.60788262
#> [16,] -1.60788262
#> [17,] -0.89442035
#> [18,]  2.04238943
#> [19,]  1.71054652
#> [20,]  2.29127162
[...]

PR #3837, opened 2020-02-26
Let check_nondata_cols allow Vector class (#3835)

teunbrand:
This PR comes from the discussion over at #3835. 
The change in code does only one thing: it add `inherits(x, "Vector")` to the allowed data columns check.
This would allow a large number of classes that exhibit vector-like behaviour from Bioconductor to be valid data columns.

I was experimenting around with this, and realised that this PR is necessary but not sufficient to let the S4 classes reach the scale transforming stage of plot building. I'd like to reach that stage of plot building since the scales are extendable and I can work with the S4 classes from there onwards.

What I would additionally like is to drop/replace the `as_tibble()` in `as_gg_data_frame()` since it also throws errors with S4 classes. Relevant code here: https://github.com/tidyverse/ggplot2/blob/53edc51bee60ae768d54d6d96e6437f44d07a16a/R/utilities.r#L379-L393 
I ran the unit tests with the code changed from `new_data_frame(tibble::as_tibble(x))` to `new_data_frame(x)` and this both didn't seem to break any tests and seemed sufficient for the S4 classes to reach the scales. Disclaimer: I couldn't install the sf package, so these tests were skipped. Relevant issues to consider should be #2609 and #2374.

Thanks for reading and considering!


clauswilke:
The `as_tibble()` call has bothered me for a while, because we're outsourcing our data validation to an external package, which may have a different agenda. If a check is needed there, I think it should also be based on the `check_nondata_cols()` function or a variation thereof. (I think in the code base, the `as_tibble()` call was there before `check_nondata_cols()` was added.)

However, it would be good to get input from @thomasp85 and/or @yutannihilation. There may be something I'm missing.

teunbrand:
I've meanwhile managed to also run the tests with the sf package, which don't seem to be affected.

yutannihilation:
> If a check is needed there, I think it should also be based on the `check_nondata_cols()` function or a variation thereof. (I think in the code base, the `as_tibble()` call was there before `check_nondata_cols()` was added.)

Sounds good to me! I've been wondering how we can replace `as_tibble()` for a long time ( https://github.com/tidyverse/ggplot2/issues/3018#issuecomment-455834049)...

teunbrand:
I might not be particularly aware what function the `as_tibble` call is. The particular check that tibble does that hinders the S4 classes is:

``` r
print(tibble:::is_1d_or_2d) # is a check in tibble:::check_valid_cols
#> function (x) 
#> {
#>     (is_vector(x) && !needs_dim(x)) || is.data.frame(x) || is.matrix(x)
#> }
#> <bytecode: 0x4272488>
#> <environment: namespace:tibble>
```

I tried to explore what behaviour would be different if we'd remove the `as_tibble`, and found two cases that would act different.

```
# heads up: this data.frame does not print prettily
df <- ggplot2:::new_data_frame(list(x = matrix(1:9, 3)))

ggplot(df, aes(col(x), row(x), fill = x)) +
  geom_tile()
```
The above would throw an error with `as_tibble()` and yields a normal plot without it (since the `validate_column_vec()` function already linearises the matrix).

Also the following case:

```
df <- data.frame(x = 1)
# Manual assignment because data.frame constructor automatically converts to POSIXct 
df$y <- as.POSIXlt(Sys.time(), "GMT")

ggplot(df, aes(x, y)) +
  geom_point()
```

Returns this error with `as_tibble()`:
>  Error: Column `y` is a date/time and must be stored as POSIXct, not POSIXlt.

And this error without `as_tibble()`:
 > Error: Invalid input: time_trans works with objects of class POSIXct only 

I think the matrix case is actually kind of convenient and the error messages about POSIXlt are both indicating what class is causing trouble. Would it be acceptable to remove the `as_tibble()` call in `as_gg_data.frame()`?

clauswilke:
Let's assess the need for `as_tibble()` systematically. I can find only two cases where `as_gg_data_frame()` is used. First in `Geom$use_defaults()`:
https://github.com/tidyverse/ggplot2/blob/660aad2db2b3495ae0d8040915a40d247133ffc0/R/geom-.r#L120-L124

Second in `Layer$compute_aesthetics()`:
https://github.com/tidyverse/ggplot2/blob/660aad2db2b3495ae0d8040915a40d247133ffc0/R/layer.r#L265-L268 

I think the `as_tibble()` check isn't needed in either. In `Geom$use_defaults()`, the check is run on the default aesthetic values of geoms, and I think it's reasonable to assume that they are valid quantities. Also, the `else` branch in the same code doesn't run the check. In `Layer$compute_aesthetics()`, we already have the check for non-data columns, so I don't think we need to check again:
https://github.com/tidyverse/ggplot2/blob/660aad2db2b3495ae0d8040915a40d247133ffc0/R/layer.r#L237-L246

Am I missing a use of this function?

@thomasp85 Do you want to chime in? You wrote the `as_tibble()` line originally.

clauswilke:
While we're at it, I also found this, which is essentially equivalent to a call to `as_gg_data_frame()`:
https://github.com/tidyverse/ggplot2/blob/660aad2db2b3495ae0d8040915a40d247133ffc0/R/facet-.r#L420-L423

teunbrand:
Thank you very much Claus for assessing `as_tibble` across the entirety of ggplot. I've run the unit tests without the `as_tibble` in facet-.r too and it appears from my point of view that it wouldn't break anything. 

I'm going to go ahead with removing the `as_tibble()` from `as_gg_data_frame()` in this PR; it doesn't really make sense to merge this when the S4 classes will throw errors a few lines of code later. It would be really cool for me if this would be merged, but if there are decent reasons to keep `as_tibble()` that can't easily be overcome, I'll understand.

clauswilke:
Yes, this looks good to me. I think the code in facet-.r needs to be handled separately, since we'd probably want an input check there.

clauswilke:
@thomasp85 Unless you have strong objections, I'd say let's go ahead with this. It seems the worst case scenario is a less informative error message in some strange corner case, and if that arises we can deal with it then.

yutannihilation:
Don't we need a NEWS bullet for this?

clauswilke:
I guess a news item would be good, so people are more broadly aware that this now works.

@teunbrand Could you add a 1-2 sentence news item in `NEWS.md`?

teunbrand:
Sure thing. I didn't know how to do that without merging in the current master first, but I hope it is correct. Thanks everyone for the help!

clauswilke:
Merging in the current master is correct. However, I'd like to ask you to rephrase the news item. Usually we try to not focus on the technical details but instead write something that makes sense to the user. So in this case, I think it should be something like "ggplot2 now accepts data columns of type `Vector`, which are widely used in the Bioconductor project." The `as_tibble()` is a technical detail that I wouldn't put into the changelog at all.

teunbrand:
Yes that indeed sounds more informative. Thanks!

clauswilke:
Thanks!

PR #3842, opened 2020-02-28
Use rlang to generate quosure list

lionel-:
Avoids making assumptions about internals of quosure lists. Fixes a test failure with dev rlang.

PR #3845, opened 2020-02-29
Include #3842 in the RC branch

yutannihilation:


yutannihilation:
Will merge this without approval after the CI passes, as this change is already approved on #3842.

yutannihilation:
Ah, thanks for approving!

PR #3850, opened 2020-03-03
Avoid stripping .data pronoun during evaluation

thomasp85:
This fixes a regression in the `.data` stripping code for default labels that would remove it before evaluation as well


thomasp85:
Classic example of forgetting to save the file before committing :-)

PR #3856, opened 2020-03-05
Merge RC back into master

thomasp85:
v3.3.0 is on CRAN 🎉

This PR merges the release branch back into master


R/geom-ribbon.r
```diff
@@ -22,8 +22,8 @@
 #' @inheritParams layer
 #' @inheritParams geom_bar
 #' @param outline.type Type of the outline of the area; `"both"` draws both the
-#'   upper and lower lines, `"upper"` draws the upper lines only. `"legacy"`
-#'   draws a closed polygon around the area.
+#'   upper and lower lines, `"upper"`/`"lower"` draws the either lines only.
```
clauswilke: Something seems to have gone wrong here, "draws the either lines only" doesn't make sense. Maybe "draws the respective lines only"?

R/stat-boxplot.r
```diff
@@ -59,7 +59,9 @@ StatBoxplot <- ggproto("StatBoxplot", Stat,
   },
 
   setup_params = function(data, params) {
-    params$flipped_aes <- has_flipped_aes(data, params, main_is_orthogonal = TRUE, group_has_equal = TRUE)
+    params$flipped_aes <- has_flipped_aes(data, params, main_is_orthogonal = TRUE,
+                                          group_has_equal = TRUE,
+                                          main_is_optional = TRUE)
```
clauswilke: Tidyverse style indenting?

R/geom-ribbon.r
clauswilke: Something seems to have gone wrong here: "draws the either lines only". Maybe write "draws the respective lines only"?

R/stat-boxplot.r
thomasp85: Isn't this in line with the style guide?

R/geom-ribbon.r
thomasp85:
```suggestion
#'   upper and lower lines, `"upper"`/`"lower"` draws the respective lines only.
```

R/stat-boxplot.r
yutannihilation: If we follow the tidyverse style guide, I believe these likes should be:

``` r
code <- "
    params$flipped_aes <- has_flipped_aes(data, params, main_is_orthogonal = TRUE,
                                          group_has_equal = TRUE,
                                          main_is_optional = TRUE)
"

styler::style_text(code)
#> Warning: Could not use colored = TRUE, as the package prettycode is not
#> installed. Please install it if you want to see colored output or see `?
#> print.vertical` for more information.
#> 
#> params$flipped_aes <- has_flipped_aes(data, params,
#>   main_is_orthogonal = TRUE,
#>   group_has_equal = TRUE,
#>   main_is_optional = TRUE
#> )
```

<sup>Created on 2020-03-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

But, it seems the guide is unclear about this point (I guess the "bad" example here describes somehow related rule, but there seems no corresponding text):

https://style.tidyverse.org/syntax.html#long-lines

hadley: I'd probably write:

```R
#> params$flipped_aes <- has_flipped_aes(
#>   data, params,
#>   main_is_orthogonal = TRUE,
#>   group_has_equal = TRUE,
#>   main_is_optional = TRUE
#> )

Issue #3846, opened 2020-03-01
stat_density2d doesn't work nicely with geom_polygon


JoFAM:
As per the title. If I use `stat_density2d(aes(fill = stat(level)), geom = "polygon")` the outer regions are completely warbled. 
```r
data(airquality)
airQ <- filter_all(airquality, ~!is.na(.))
ggplot(airQ, aes(x = Solar.R, y = Temp)) +
  geom_point() +
  stat_density2d(aes(fill = stat(level)),
                 geom = "polygon")
```
gives
![Wrongplot](https://user-images.githubusercontent.com/809906/75631397-e2974880-5bf2-11ea-866c-f41c0c784862.png)

thomasp85:
The coming release will make you happy 🙂

JoFAM:
@thomasp85 I hoped to be able to use this for the practical exercises tomorrow, but that's pushing it a bit probably? 😃 

Definitely looking forward to it.

thomasp85:
You can install the v3.3.0-rc branch and run with it

clauswilke:
Looking into this now, I'm not sure this is fixed yet. We can draw filled contours from a raster grid of data, but we can't run the density estimation and then draw the filled contour, as far as I can tell.

``` r
library(tidyverse)

data(airquality)

airQ <- filter_all(airquality, ~!is.na(.))

ggplot(airQ, aes(x = Solar.R, y = Temp)) +
  stat_density_2d(aes(fill = after_stat(level)), geom = "polygon")
```

![](https://i.imgur.com/A0uoBCx.png)

<sup>Created on 2020-03-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.6.0 (2019-04-26)
#>  os       macOS Mojave 10.14.6        
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       America/Chicago             
#>  date     2020-03-01                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                            
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 3.6.0)                    
[...]

PR #3864, opened 2020-03-07
Implement filled 2d density contours

clauswilke:
The missing feature from 3.3.0, as described in #3846.

This is work in progress. Still to do:

- [x] Implement `geom_density2d_filled()` analogously to `geom_contour_filled()`
- [x] Usage examples
- [x] Make `nlevel` work for `geom_density2d_filled()`.
- [x] Unit tests
- [x] News item
- [x] Fully document computed variables
- [x] Improve documentation of `stat_contour_filled()`


clauswilke:
@thomasp85 @yutannihilation I've run into a weird issue trying to implement filled density contours, and I was hoping you might see what the problem is. In brief, setting `fill = after_stat(level)` in `default_aes` doesn't work, even though it works when provided explicitly in the geom's `aes()` call.

The relevant code is here: https://github.com/clauswilke/ggplot2/blob/bee68db107e49876e3c478a5d8f024b122f6478d/R/geom-density2d.r#L121-L126

``` r
library(tidyverse)

data(airquality)
airQ <- filter_all(airquality, ~!is.na(.))

# this works fine
ggplot(airQ, aes(x = Solar.R, y = Temp)) +
  geom_density2d_filled(aes(fill = after_stat(level)), alpha = 0.7)
```



``` r

# but setting `fill = after_stat(level)` in `default_aes` doesn't work
ggplot(airQ, aes(x = Solar.R, y = Temp)) +
  geom_density2d_filled(alpha = 0.7)
#> Error in after_stat(level): object 'level' not found
```

<sup>Created on 2020-03-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Never mind. Apparently this works only from inside a stat.

``` r
library(tidyverse)

data(airquality)
airQ <- filter_all(airquality, ~!is.na(.))

ggplot(airQ, aes(x = Solar.R, y = Temp)) +
  geom_density2d_filled(alpha = 0.7) +
  geom_density2d(size = 0.25, color = "black") +
  geom_point(size = 1)
```



<sup>Created on 2020-03-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
@thomasp85 How's this?

``` r
library(ggplot2)

  geom_contour_filled() + 
  facet_wrap(~waiting < 70)
```



``` r

set.seed(4393)
dsmall <- diamonds[sample(nrow(diamonds), 1000), ]
d <- ggplot(dsmall, aes(x, y))

d + geom_density_2d_filled() +
  facet_wrap(~cut)
```



``` r

d + geom_density_2d_filled(contour_var = "ndensity") +
  facet_wrap(~cut)
```



``` r

d + geom_density_2d_filled(contour_var = "count") +
  facet_wrap(~cut)
```



``` r

d + geom_density_2d(aes(color = after_stat(level))) + 
  facet_wrap(~cut)
```



``` r

d + geom_density_2d(aes(color = after_stat(level)), contour_var = "ndensity") + 
  facet_wrap(~cut)
```



``` r

d + geom_density_2d(aes(color = after_stat(level)), contour_var = "count") + 
  facet_wrap(~cut)
#> Warning: stat_contour(): Zero contours were generated
```



<sup>Created on 2020-03-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
This looks good. The `compute_layer` approach was exactly what I had in mind...

It hurts a little that `contour_var` cannot be an aesthetic, but there is no way around that if it all has to happen in a stat...

Some part of me really wants to make the contouring happen in the geom so that anything that ends up as a raster hightmap after the stat can be contoured... but then we end up with the issues of colour scaling the output of contouring –– no singular best solution it seems

clauswilke:
> Some part of me really wants to make the contouring happen in the geom so that anything that ends up as a raster hightmap after the stat can be contoured... but then we end up with the issues of colour scaling the output of contouring –– no singular best solution it seems

If the geom could get the height data from the stat and the binning information from the scale then it might work. We could always in the future add such geoms in addition to `geom_contour()`, as I did in [ggisoband](https://github.com/clauswilke/ggisoband) (which conceptually does what you want, it just doesn't quite work right in all cases).

clauswilke:
I think this is ready for review now.

R/geom-contour.r
```diff
@@ -123,3 +124,11 @@ GeomContour <- ggproto("GeomContour", GeomPath,
     alpha = NA
   )
 )
+
+#' @rdname ggplot2-ggproto
+#' @format NULL
+#' @usage NULL
+#' @export
+#' @include geom-polygon.r
+GeomContourFilled <- ggproto("GeomContourFilled", GeomPolygon)
```
thomasp85: What is the purpose of this? Do we have any precedence for creating ggproto aliases?

R/stat-density-2d.r
```diff
@@ -10,34 +17,51 @@
...
                             adjust = c(1, 1),
                             na.rm = FALSE,
                             show.legend = NA,
                             inherit.aes = TRUE) {
+  if (isTRUE(contour_type == "bands")) {
+    stat <- StatDensity2dFilled
+  } else {
+    stat <- StatDensity2d
+  }
+
```
thomasp85: I'm not particularly fond of this setup. Basically I don't think `contour_type` should be a user facing parameter but simply a quality of `StatDensity2d` and `StatDensity2dFilled`. One of the reasons is that if this parameter is not set explicitly e.g. by doing something weird like `geom_polygon(stat = 'density2dfilled')` the code will behave like `density2d`, but also because having a switch in the constructor that turns a layer into a new type of layer is not really in line with how we usually do it

R/geom-contour.r
clauswilke: This is needed to auto-generate documentation for aesthetics. We probably should do it more consistently across ggplot2. See here:
https://github.com/clauswilke/ggplot2/blob/743bad010638fdccaed4c0cfa0258e0879e1263c/R/geom-density2d.r#L9-L10

R/stat-density-2d.r
clauswilke: Agreed. I think this is a hold-over from my first attempt at writing this, which muddled everything together and would have required a lot of code duplication for these two stats. Now, with the new setup, I can do it very cleanly and easily. I'll rewrite.

clauswilke:
The relationship between `stat_bin2d()` and `stat_density_2d()` is equivalent to `stat_bin()` and `stat_density()`. One creates a histogram, by making regular-sized bins and simply counting how many observations fall into each bin. The other does a kernel-density estimation. Both are useful, and they are distinct.

thomasp85:
You are right. I got bin2d confused

R/stat-density-2d.r
clauswilke: Done. There was really no need for `contour_type`.

Note that the failing CI build for R devel has nothing to do with this PR but is due to #3885.

R/geom-contour.r
thomasp85: Ah... that seems bad, but is not really the fault of this PR

R/stat-density-2d.r
```diff
@@ -60,30 +115,80 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
 
   required_aes = c("x", "y"),
 
+  extra_params = c(
+    "na.rm", "contour", "contour_var",
+    "bins", "binwidth", "breaks"
+  ),
+
+  # stat used for contouring
+  contour_stat = StatContour,
```
thomasp85: I don't know whether I like that a stat contains another stat as a field. Why not just have it as the equivalent of the old parameter (i.e. a character string) and do the switch as you did before?

My main gripe was that this switch was user accessible, not that it was being done

clauswilke: fixed.

PR #3868, opened 2020-03-08
Ignore empty layers when facetting

### LINKED ISSUE #3853: Facet_grid with empty "data=" in a second geom: empty(data) returns NA and error message is wrong
Someone puts a second facet in a facet_grid, but forgets to add a data argument:

```r
ggplot() + geom_point(aes(x=Sepal.Width, y=Sepal.Length), data=iris) +
  geom_line(aes(x = whatever, y=whatever)) + 
  facet_grid(rows=vars(Species))
```
Should say something like "whatever not found", but instead you get "Error in if (empty(data)) { : missing value where TRUE/FALSE needed".

yutannihilation:
Fix #3853 

This code fails because the facet cannot handle the empty second layer, which should just be ignored.

``` r
library(ggplot2)

ggplot() +
  geom_point(aes(x = Sepal.Width, y = Sepal.Length), data = iris) +
  geom_line() + 
  facet_grid(rows=vars(Species))
#> Error in if (empty(data)) {: missing value where TRUE/FALSE needed
```

I think this is a bug of `empty()`; it should return `TRUE` for `waiver()`, not `NA`, because the `waiver()` represents the absense of data.


NEWS.md
```diff
@@ -9,6 +9,8 @@
 * Added an `outside` option to `annotation_logticks()` that places tick marks
   outside of the plot bounds. (#3783, @kbodwin)
 
+* Facets now handles layers containing no data (@yutannihilation, #3853).
```
clauswilke: "Facets now handle"

tests/testthat/test-facet-.r
```diff
@@ -90,6 +90,18 @@ test_that("facets split up the data", {
   expect_equal(d1$PANEL, factor(1:3))
 })
 
+test_that("facets handle empty layers", {
+  p <- ggplot() + geom_point(aes(x, y), df) + geom_line()
```
clauswilke: The way `df` is being reused in this file is a bit strange. Initially it's defined outside of `test_that()` and reused many times, but later on it's overwritten inside various `test_that()` alls. It makes it difficult to figure out what `df` actually is at any point in the code. Maybe refactor through a combination of moving the creation of `df` into `test_that()` calls and combining some tests into one  `test_that()` call so you don't have to recreate `df` over and over?

yutannihilation: Makes sense. I'll try it.

yutannihilation:
Moved the creation of `df` into each test. Does this looks good?

yutannihilation:
Thanks for reviewing.

PR #3872, opened 2020-03-09
Remove a mysterious dot and regenerate docs

yutannihilation:
Remove a dot and regenerate docs (It seems some document changes are not reflected to the actual docs yet)


yutannihilation:
Thanks

PR #3880, opened 2020-03-10
Avoid split-combine with only one grouping var

thomasp85:
This PR is a performance improvement. It removes the need to split and combine the data inside `compute_*()` and `draw_*()` functions when there is only a single panel (and group)...

It uses the `"n"` attribute assigned by the id function to quickly figure out if the grouping is all a single group. This attribute disappears if the grouping column gets manipulated so it should be safe. For panels it uses the fact the PANEL column is encoded as a factor and simply checks the length of `levels`. Testing indicate that any check that relies on comparing the values in the grouping prior to split-combine, e.g. `all(x[1] == x)` will offset the performance gain, so checking has been confined to the cases where it can be determined instantaneously.

(A side effect of this PR is that it fixes a few edge cases where PANEL did not end up encoded as a factor)

The performance gain of this is very plot dependent, but will especially be seen in single panel plots with large data where it can give a 10-20% performance boost. For other plots the performance gain will be negligible.

As `dapply()` is not exported an no unit tests break, I consider this change fairly safe, but would like to discuss if there is any hesitations as to the approach


R/compat-plyr.R
```diff
@@ -359,19 +359,43 @@ rbind_dfs <- function(dfs) {
...
+#' @export
+single_value.default <- function(x, ...) {
+  # This is set by id() used in creating the grouping var
+  isTRUE(attr(x, "n") == 1)
+}
+#' @export
+single_value.factor <- function(x, ...) {
+  # Panels are encoded as factor numbers
+  lev <- levels(x)
+  length(lev) == 1 && lev[1] == "1"
```
hadley: `identical(x, "1")` would be equivalent and faster, I think.

R/compat-plyr.R
```diff
@@ -359,19 +359,43 @@ rbind_dfs <- function(dfs) {
...
+  UseMethod("single_value")
+}
+#' @export
+single_value.default <- function(x, ...) {
+  # This is set by id() used in creating the grouping var
+  isTRUE(attr(x, "n") == 1)
+}
+#' @export
+single_value.factor <- function(x, ...) {
+  # Panels are encoded as factor numbers
```
hadley: Needs a comment saying that `NA`s are ignored, but shouldn't matter because they never occur in panels.

R/compat-plyr.R
```diff
@@ -359,19 +359,43 @@ rbind_dfs <- function(dfs) {
...
   }))
 }
+#' @export
+single_value <- function(x, ...) {
+  UseMethod("single_value")
+}
+#' @export
+single_value.default <- function(x, ...) {
+  # This is set by id() used in creating the grouping var
+  isTRUE(attr(x, "n") == 1)
```
hadley: I think `identical(attr(x, "n"), 1L)` would be slightly nicer here

R/compat-plyr.R
```diff
@@ -359,19 +359,43 @@ rbind_dfs <- function(dfs) {
...
     res <- modify_list(unclass(vars), unclass(res))
     new_data_frame(res[intersect(c(fallback_order, names(res)), names(res))])
+  }
+
+  # Shortcut when only one group
+  if (all(vapply(grouping_cols, single_value, logical(1)))) {
+    return(apply_fun(df))
+  }
+
+  ids <- id(grouping_cols, drop = drop)
```
hadley: This is likely to be significantly faster if we use the vctrs equivalent.

thomasp85: Yes... I think most of the compatibility-plyr file will be erased when we decide to go full on with vctrs

PR #3882, opened 2020-03-10
ignore min/max variables when detecting orientation in geom_bar

### LINKED ISSUE #3866: geom_bar() is broken!
Adding an aesthetic `ymin` or `ymax` makes `geom_bar()` do silly things.
This is true on both the devel-version and new CRAN version.
It did not use to do that.

``` r
library(ggplot2)

d <- data.frame(x = "A", y = 1, ymin = 0.5)

ggplot(d) +
  aes(y = y, x = x, ymin = ymin) +
  geom_bar(stat = "identity")
```

<sup>Created on 2020-03-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Fixes #3866 

The check for `range_is_orthogonal` in `geom_bar()` was unneeded and resulted in failure with edge cases

### DIFF
--- a/R/geom-bar.r
+++ b/R/geom-bar.r
@@ -119,7 +119,7 @@ GeomBar <- ggproto("GeomBar", GeomRect,
   non_missing_aes = c("xmin", "xmax", "ymin", "ymax"),
 
   setup_params = function(data, params) {
-    params$flipped_aes <- has_flipped_aes(data, params, range_is_orthogonal = FALSE)
+    params$flipped_aes <- has_flipped_aes(data, params)
     params
   },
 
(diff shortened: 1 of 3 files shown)


yutannihilation:
> could you fix the document

I mean I just want you to remove this sentence since `geom_bar()` no longer uses this option after this PR gets merged.

```
 #'   ... `FALSE` is less 
 #'   prevalent but can be seen in [geom_bar()] where it may encode the span of 
 #'   each bar. 
```

yutannihilation:
@thomasp85 May I tweak the document and merge this?

thomasp85:
Yes — sure. I’ve been elsewhere for a while

yutannihilation:
Sure.

PR #3883, opened 2020-03-10
calculate z.range in stat_contour and stat_contour_filled on the full layer data

### LINKED ISSUE #3875: Filled contours don't play well with faceting
This is technically a bug report against my PR #3864, but the bug isn't caused by that PR, it's just revealed by it. The same bug exists with `geom_contour_filled()`, just requires more code to generate a reprex.

I have noticed that when I draw filled density contours across facets, the color levels in the legend get jumbled. (Carefully look at the levels in the legend, and you'll see they're not even in order.)

``` r
library(ggplot2)

set.seed(4393)
dsmall <- diamonds[sample(nrow(diamonds), 1000), ]

ggplot(dsmall, aes(x, y)) +
  geom_density_2d_filled() +
  facet_grid(. ~ cut)
```

What happens is that the breaks are calculated separately for each facet, and then they don't match across facets. It's possible to work around this by manually specifying breaks, but I think the default settings should work with facets.

Fundamentally, the problem is that the `level` variable for contour bands is an ordered factor specifying the bin boundaries:
https://github.com/tidyverse/ggplot2/blob/db58364f9d95b7596249feb01b34420b7b4483b5/R/stat-contour.r#L111

By contrast, the `level` variable for contour lines is a numeric:
https://github.com/tidyverse/ggplot2/blob/db58364f9d95b7596249feb01b34420b7b4483b5/R/stat-contour.r#L86

I think instead we should return a numeric value for `level` in all cases and by default map that to fill for filled contours. We can still return the ordered factor as well as an option, under a different name. The numeric value I would use for `level` would be the mid-point between the lower and the upper level boundary. This is consistent with other binning approaches, e.g. the placement of bars in histograms.

The result would be something like the following (but without the `aes(fill = ...)` statement needed):
``` r
ggplot(dsmall, aes(x, y)) +
  geom_density_2d_filled(aes(fill = after_stat(level_high))) +
  facet_grid(. ~ cut) +
  scale_fill_viridis_c()
```

This will be consistent with how contour lines work already:
``` r

ggplot(dsmall, aes(x, y)) +
  geom_density_2d(aes(color = after_stat(level))) +
  facet_grid(. ~ cut) +
  scale_color_viridis_c()
```

If we agree on this approach I'll incorporate it into #3864.

thomasp85:
Fixes #3875 

This PR fixes the issue as discussed in the thread. The issue is also present in the standard `stat_contour` where the levels of the contour lines would not be equivalent across panels

## Old buggy behaviour

``` r
library(ggplot2)
  geom_contour_filled() + 
  facet_wrap(~waiting < 70)
```

<sup>Created on 2020-03-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

## Working example with PR

``` r
library(ggplot2)
  geom_contour_filled() + 
  facet_wrap(~waiting < 70)
```

<sup>Created on 2020-03-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

### DIFF
--- a/R/stat-contour.r
+++ b/R/stat-contour.r
@@ -70,15 +70,17 @@ stat_contour_filled <- function(mapping = NULL, data = NULL,
 #' @usage NULL
 #' @export
 StatContour <- ggproto("StatContour", Stat,
-
+  setup_params = function(data, params) {
+    params$z.range <- range(data$z, na.rm = TRUE, finite = TRUE)
+    params
+  },
   required_aes = c("x", "y", "z"),
   default_aes = aes(order = after_stat(level)),
 
-  compute_group = function(data, scales, bins = NULL, binwidth = NULL,
+  compute_group = function(data, scales, z.range, bins = NULL, binwidth = NULL,
                            breaks = NULL, na.rm = FALSE) {
 
-    z_range <- range(data$z, na.rm = TRUE, finite = TRUE)
-    breaks <- contour_breaks(z_range, bins, binwidth, breaks)
+    breaks <- contour_breaks(z.range, bins, binwidth, breaks)
 
     isolines <- xyz_to_isolines(data, breaks)
     path_df <- iso_to_path(isolines, data$group[1])
@@ -95,14 +97,17 @@ StatContour <- ggproto("StatContour", Stat,
 #' @usage NULL
 #' @export
 StatContourFilled <- ggproto("StatContourFilled", Stat,
-
+  setup_params = function(data, params) {
+    params$z.range <- range(data$z, na.rm = TRUE, finite = TRUE)
+    params
+  },
   required_aes = c("x", "y", "z"),
   default_aes = aes(order = after_stat(level), fill = after_stat(level)),
 
-  compute_group = function(data, scales, bins = NULL, binwidth = NULL, breaks = NULL, na.rm = FALSE) {
+  compute_group = function(data, scales, z.range, bins = NULL, binwidth = NULL,
+                           breaks = NULL, na.rm = FALSE) {
 
-    z_range <- range(data$z, na.rm = TRUE, finite = TRUE)
-    breaks <- contour_breaks(z_range, bins, binwidth, breaks)
+    breaks <- contour_breaks(z.range, bins, binwidth, breaks)
 
     isobands <- xyz_to_isobands(data, breaks)
     names(isobands) <- pretty_isoband_levels(names(isobands))


clauswilke:
I think this breaks the use of `StatContour` by `StatDensity2d`:

https://github.com/tidyverse/ggplot2/blob/435ac040dfe441818b69bf55a4cf7a442c7cb2ac/R/stat-density-2d.r#L71-L80

We don't know the z range of inferred densities until we have inferred them.

Does the following reprex work with your branch?
``` r
library(ggplot2)

set.seed(4393)
dsmall <- diamonds[sample(nrow(diamonds), 1000), ]
ggplot(dsmall, aes(x, y)) + 
  geom_density_2d(aes(colour = after_stat(level))) + 
  facet_grid(. ~ cut)
```



<sup>Created on 2020-03-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Continuing discussion from here: https://github.com/tidyverse/ggplot2/issues/3875#issuecomment-597116549

> Moving to numeric will loose information and remove the possibility of using binned guides which is a very bad move

Just to be clear: I think we should return both. So this is primarily a discussion about what the default mapping for `fill` should be, and whether the column called `level` should be numeric or categorical. Maybe the compromise is to keep `level` numeric for consistency but by default map to the column that returns the factor, whatever its name? :-)

Also, unless I'm missing something, it's fairly easy to go from the numeric limits to the ordered factor. It's much harder the other way round, because we don't have a simple inverse to `glue()`.

``` r
library(ggplot2)
library(glue)

set.seed(4393)
dsmall <- diamonds[sample(nrow(diamonds), 1000), ]

ggplot(dsmall, aes(x, y)) +
  geom_density_2d_filled(aes(fill = after_stat(bin))) +
  scale_fill_viridis_d(name = NULL)
```



``` r
  
ggplot(dsmall, aes(x, y)) +
  geom_density_2d_filled(
    aes(fill = after_stat(ordered(glue("({level_low}, {level_high}]"))))
  ) +
  scale_fill_viridis_d(name = NULL)
```



<sup>Created on 2020-03-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
I just realized that there may be a clean way to resolve all of these issues at the same time. Let me give it a shot.

thomasp85:
Well, if you return `level_low` and `level_high` then no information is lost, but a single numeric level will not be able to capture the binning range. We should continue to default map to the ordered factor. There is no mapping to a single numeric variable that is truthful to the data.

You are correct that your example with density_2d doesn't work with this PR. I'm sure that is fixable, but I'll let you work on your solution for now

clauswilke:
The latest iteration of my PR #3864 now incorporates the approach you proposed here and fixes all associated issues. See example below. Could you take a look and let me know whether I'm on the right track? If yes I'll clean up the code and rewrite the documentation.

``` r
library(ggplot2)

  geom_contour_filled() + 
  facet_wrap(~waiting < 70)
```



``` r

set.seed(4393)
dsmall <- diamonds[sample(nrow(diamonds), 1000), ]
d <- ggplot(dsmall, aes(x, y))

d + geom_density_2d_filled() +
  facet_wrap(~cut)
```



``` r

d + geom_density_2d_filled(contour_var = "ndensity") +
  facet_wrap(~cut)
```



``` r

d + geom_density_2d_filled(contour_var = "count") +
  facet_wrap(~cut)
```



``` r

d + geom_density_2d(aes(color = after_stat(level))) + 
  facet_wrap(~cut)
```



``` r

d + geom_density_2d(aes(color = after_stat(level)), contour_var = "ndensity") + 
  facet_wrap(~cut)
```



``` r

d + geom_density_2d(aes(color = after_stat(level)), contour_var = "count") + 
  facet_wrap(~cut)
#> Warning: stat_contour(): Zero contours were generated
```



<sup>Created on 2020-03-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
The key idea of my revised approach is to implement a true chaining of the 2d-density stat and the contour stat, by reimplementing `compute_layer()`.

https://github.com/clauswilke/ggplot2/blob/6546535fc46797a5520de1adcfe4c7c1d281ae0f/R/stat-density-2d.r#L94-L123

thomasp85:
Closing this now. All in all we are on the same page and I'm happy with #3864

PR #3862, opened 2020-03-06
Enable Github Action

yutannihilation:
Fix #3760, also fixes #3702.

## Major changes

* Enable the following GitHub Actions:
    - R-CMD-check: Run `R CMD check` on Windows, macOS, and Ubuntu with the last 5 minor versions of R.
    - pr-command: Commenting `/document` will trigger `devtools::document()` automatically (example: https://github.com/yutannihilation/ggplot2/pull/3#issuecomment-596099457).
    - pkgdown: build the pkgdown site.
* ~~Remove Travis and AppVeyor settings~~ Stop building pkgdown site by Travis (Now I feel it's safe to leave them for now and remove when we confirm GitHub Actions CI does the right thing. But, pkgdown site should not be built twice, so I tweaked `.travis.yml` only about this point)

Diffs between the default YAML files are following:

<details>

``` diff
diff --git a/.github/workflows/R-CMD-check.yaml b/.github/workflows/R-CMD-check.yaml
index 0d76f6e0..075d5167 100644
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -18,18 +18,23 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: '3.6'}
-          - {os: macOS-latest, r: '3.6'}
-          - {os: macOS-latest, r: 'devel'}
-          - {os: ubuntu-16.04, r: '3.2', cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.3', cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.4', cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.5', cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.6', cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
+          - {os: windows-latest, r: '3.6', vdiffr: true}
+          - {os: macOS-latest, r: '3.6', vdiffr: true}
+          - {os: macOS-latest, r: 'devel', vdiffr: false}
+          - {os: ubuntu-16.04, r: '3.2', vdiffr: false, cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
[...]

PR #3895, opened 2020-03-15
Temporarily disable Test coverage step on GitHub Action CI

yutannihilation:
`covr::codecov()` keeps failing on GitHub Action CI.

```
Error in curl::curl_fetch_memory(url, handle = handle) : 
  HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
Calls: <Anonymous> ... request_fetch -> request_fetch.write_memory -> <Anonymous>
Execution halted
##[error]Process completed with exit code 1.
```
https://github.com/tidyverse/ggplot2/runs/509138337?check_suite_focus=true#step:13:14

I tried some tweaks on my forked repo (https://github.com/yutannihilation/ggplot2/pull/4/), but none of these worked. As the other repos (e.g. [roxygen2](https://github.com/r-lib/roxygen2/blob/2a068b181bc15ec55545ddd752f5341471f5e74a/.github/workflows/R-CMD-check.yaml#L83-L86)) works without error, this seems specific to ggplot2. So, I think it takes some time to figure out what's happening here, so this should be disabled for now.

Fortunately(?), I forgot to remove the test coverage step on Travis CI in #3862, so we can rely on it for a while.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -97,7 +97,8 @@ jobs:
           name: ${{ runner.os }}-r${{ matrix.config.r }}-results
           path: check
 
-      - name: Test coverage
-        if: matrix.config.os == 'macOS-latest' && matrix.config.r == '3.6'
-        run: covr::codecov()
-        shell: Rscript {0}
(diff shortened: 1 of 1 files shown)


yutannihilation:
Thanks for approving.

PR #3907, opened 2020-03-22
fix stat_contour() for irregular data grids.

### LINKED ISSUE #3906: geom_contour incompatible with scale_x_log10 and scale_y_log10
`geom_contour` seems to be incompatible with log scales, generated with `scale_x_log10` or `scale_y_log10` (or both).

```r
library(ggplot)
v + geom_contour() + scale_x_log10()
```

This produces a warning message, below, and an empty plot.

```
Warning message:
Computation failed in `stat_contour()`:
Number of x coordinates must match number of columns in density matrix.
```

Use of `scale_y_log10` yields a similar result, with the warning message about `y` coordinates.

clauswilke:
Closes #3906.

``` r
library(ggplot2)

v <- ggplot(faithfuld, aes(waiting, eruptions, z = density)) +
  geom_point(aes(color = density)) + geom_contour(color = "white")

v
```

``` r
v + scale_x_log10()
```

<sup>Created on 2020-03-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/stat-contour.r
+++ b/R/stat-contour.r
@@ -206,14 +206,14 @@ xyz_to_isobands <- function(data, breaks) {
 #'
 isoband_z_matrix <- function(data) {
   # Convert vector of data to raster
-  x_pos <- as.integer((data$x - min(data$x)) / resolution(data$x, FALSE))
-  y_pos <- as.integer((max(data$y) - data$y) / resolution(data$y, FALSE))
+  x_pos <- as.integer(factor(data$x, levels = sort(unique(data$x))))
+  y_pos <- as.integer(factor(data$y, levels = sort(unique(data$y))))
 
-  nrow <- max(y_pos) + 1
-  ncol <- max(x_pos) + 1
+  nrow <- max(y_pos)
+  ncol <- max(x_pos)
 
   raster <- matrix(NA_real_, nrow = nrow, ncol = ncol)
-  raster[cbind(nrow - y_pos, x_pos + 1)] <- data$z
+  raster[cbind(y_pos, x_pos)] <- data$z
 
   raster
 }
--- a/tests/testthat/test-stat-contour.R
+++ b/tests/testthat/test-stat-contour.R
@@ -13,6 +13,25 @@ test_that("contouring sparse data results in a warning", {
   expect_warning(ggplot_build(p), "Zero contours were generated")
 })
 
+test_that("contouring irregularly spaced data works", {
+  tbl <- expand.grid(x = c(1, 10, 100, 1000), y = 1:3)
+  tbl$z <- 1
+  tbl[c(6, 7), ]$z <- 10
+  p <- ggplot(tbl, aes(x, y, z = z)) + geom_contour(breaks = c(4, 8))
+
+  # we're testing for set equality here because contour lines are not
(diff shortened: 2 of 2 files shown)


R/stat-contour.r
```diff
@@ -206,14 +206,14 @@ xyz_to_isobands <- function(data, breaks) {
...
+  y_pos <- as.integer(factor(data$y, levels = sort(unique(data$y))))
 
-  nrow <- max(y_pos) + 1
-  ncol <- max(x_pos) + 1
+  nrow <- max(y_pos)
+  ncol <- max(x_pos)
 
   raster <- matrix(NA_real_, nrow = nrow, ncol = ncol)
-  raster[cbind(nrow - y_pos, x_pos + 1)] <- data$z
+  raster[cbind(y_pos, x_pos)] <- data$z
```
paleolimbot: This looks great! I wrote this based on the code in `GeomRaster` because it was faster...before it was:

```r
raster <- tapply(data$z, data[c("x", "y")], identity)
```

(I found that [here](https://github.com/tidyverse/ggplot2/commit/b771c502a61bb9be1ab7ab613429cbc8a1bf43fa) in case you want to look at the context)

It might be worth checking whether that is faster (I suspect your way is better)

tests/testthat/test-stat-contour.R
```diff
@@ -10,7 +10,7 @@ test_that("a warning is issued when there is more than one z per x+y", {
 test_that("contouring sparse data results in a warning", {
   tbl <- data_frame(x = c(1, 27, 32), y = c(1, 1, 30), z = c(1, 2, 3))
   p <- ggplot(tbl, aes(x, y, z = z)) + geom_contour()
-  expect_warning(ggplot_build(p), "Number of x coordinates must match")
+  expect_warning(ggplot_build(p), "Zero contours were generated")
 })
```
paleolimbot: Could you add a test for log-transformed data? I suspect that log scales worked before my PR, but there was no test to let me know that my changes broke anything.

R/stat-contour.r
clauswilke: `tapply()` is ~40% slower.

As an aside, we're doing a lot of work wrangling the data into the right format for isoband and then back out. It'd probably be more efficient to just write an isoband function that takes the data in the format preferred by ggplot2 and returns it in a format that can be used directly. But that's for another day.

tests/testthat/test-stat-contour.R
clauswilke: Ironically, I think it was this test. :-) The "Number of x coordinates must match" error is generated by isoband when the input data is of the wrong format. In any case, I'll add a test that does create a contour, to make this more explicit.

clauswilke: The test is now there, if you want to take a quick look.

paleolimbot: Great! Good call on set equality...I remember my graphical tests failing and this is a great solution.

clauswilke: It's what I'm using in the isoband package also. It was quite a challenge to write tests that wouldn't fail on random architectures.
https://github.com/wilkelab/isoband/blob/master/tests/testthat/test-isolines.R

PR #3916, opened 2020-03-30
removed expect_equal(2 * 2, 4)

rerickson-usgs:
removed a seemingly  unneeded expect_equal(2 * 2, 4). Please let me know if this test serves a purpose. I was reading through this project to see good examples of testthat.


clauswilke:
Thanks. This is the example test in the usethis template, it should have been removed.

PR #3922, opened 2020-04-01
Run visual tests on CI

cpsievert:
I'm pretty sure vdiffr hasn't been running on Travis CI since https://github.com/tidyverse/ggplot2/commit/9fdf452d34be3986c778c9f4306ce96c7c14af8e

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ matrix:
   - r: release
     env: VDIFFR_RUN_TESTS=true
     env: VDIFFR_LOG_PATH="../vdiffr.Rout.fail"
+    before_install:
+      - echo $VDIFFR_RUN_TESTS
   - r: 3.5
   - r: 3.4
   - r: 3.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ matrix:
     # avoind errors due to world clock API (hopefully can be removed in future)
     env: _R_CHECK_SYSTEM_CLOCK_=false
   - r: release
-    env: VDIFFR_RUN_TESTS=true
-    env: VDIFFR_LOG_PATH="../vdiffr.Rout.fail"
+    env:
+      - VDIFFR_RUN_TESTS=true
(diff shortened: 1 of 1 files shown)


cpsievert:
BTW, f3d4087 is here to just show `VDIFFR_TUN_TESTS` isn't set properly https://travis-ci.org/github/tidyverse/ggplot2/jobs/669725114#L1195

clauswilke:
Yeah, vdiffr silently not running and missing all sorts of regressions is one of my recurring nightmares. Is there any way to automatically check whether it's actually run? This might require some change to vdiffr itself. If we could query whether vdiffr thinks it's running we could write a sentinel file somewhere and then test in CI whether the file has been created / has the right contents.

cpsievert:
I'm not sure, perhaps @lionel- can offer advice?

yutannihilation:
Wow, I didn't notice this... But, visual tests do run on GitHub Action CI, right?

cpsievert:
> tests do run on GitHub Action CI, right?

I'm not sure, I don't see anything immediately wrong with the workflow fille

yutannihilation:
Confirmed the visual tests definitely run on GitHub Actions CI.

diff: https://github.com/yutannihilation/ggplot2/pull/5/files
test failure: https://github.com/yutannihilation/ggplot2/pull/5/checks?check_run_id=554963096#step:11:105

lionel-:
Maybe vdiffr could signal a special condition class when a test is not run. Then a calling (not exiting so other tests keep running) condition handler could transform the condition to an error that would be caught by testthat.

With R 4.0 it would be easy to install do this with a global condition handler, without any change in the ggplot2 sources. On older R we'd need to install the handler around `test_check()` with `withCallingHandlers()`, conditionally on some environment variable indicating that the tests should really be run.

clauswilke:
@yutannihilation Yes, go ahead.

yutannihilation:
Sure.

yutannihilation:
Thanks @cpsievert for finding this.

PR #3931, opened 2020-04-07
Created new Geoms geom_shadowpath()`, `geom_shadowline()` and `geom_s…

marcmenem:
I have created new geoms similar to geom_path, geom_line and geom_step that first draw a shadow beneath the lines to plot in order to get a more aesthecalliy pleasing effect when several lines overlap.

This is my frst time contributing to an open source project, and I am not sure this is the best place to contribute, or whether it would be more apropriate to create an extension with my code rather than contribute to ggplot2, please let me know.


clauswilke:
This is more appropriate for an extension package.

yutannihilation:
I agree with Claus. Though `geom_shadowpath()` seems an interesting experimentation, ggplot2 will never have a new Geom unless there's a strong reason that it should belong to ggplot2.

PR #3942, opened 2020-04-11
Don't modify legend key settings in geom_sf() during plot build.

### LINKED ISSUE #3941: Legend keys from geom_sf() change upon repeat printing of the same plot
I'm having a super weird issue with `ggsave` and, after [sharing it at RStudio community](https://community.rstudio.com/t/legend-keys-from-geom-sf-in-pdfs-created-using-ggsave-dont-match-what-is-displayed-in-the-plots-window-nor-the-output-from-rstudios-export/60722/4), Mara Averick suggested I open an issue here. So here I am 😅 

Basically, the legends of plots I generate using `ggsave` do not match what is displayed in RStudio's "Plots" window nor the legends of plots I generate using RStudio's "Export" option. This is happening for maps created using `geom_sf`.

Here's an example:

``` r
library(rnaturalearth)
library(ggplot2)

world <-
  rnaturalearth::ne_countries(scale = "medium", returnclass = "sf") # creates a r object  called "world" with spatial data

ggplot2::ggplot(data = world) + #starts plot with "world" as its data
  geom_sf(aes(color = continent), show.legend = "point") # draws map with colored continent outlines
```

<sup>Created on 2020-04-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

As you can see, the map above has points as its legend key. That is the output I wanted to get, however, when I use `ggsave("ggsave_test.pdf")` to save the map as a pdf file this is what I get:
image

The weird thing is that this problem seems to be exclusive to ggsave since saving the plot using RStudio's "Export" functionality or knitting a Rmarkdown document result in the plot having points as its legend key.

This happens for .ps, .pdf, .jpeg, .tiff, .png, .bmp, and .svg files both when leaving `ggsave()`'s plot argument  blank (e.g., when it uses `last_plot()` like in the reprex above), and when using the saved map as the argument for plot. There's an exception though I used `purrr::walk` to test the different file formats and the .eps file that was generated using the named map as ggsave()'s "plot" argument has the points as its key ! The plots generated using ggsave(filename = here::here("ggsave_no_loop.eps"), plot = map), and `purrr::walk` without the named map as plot's argument have the squares as their keys though.

I think this might be a macOS-specific issue (or even something that is happening only in my computer) because this problem doesn't happen in rstudio.cloud.

Cheers and thanks for your amazing work!

clauswilke:
Fixes #3941.

``` r
library(ggplot2)

p <- ggplot(nc[1:3, ]) +
  geom_sf(aes(color = NAME), show.legend = "point")
p
```

``` r
p
```

<sup>Created on 2020-04-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/layer-sf.R
+++ b/R/layer-sf.R
@@ -20,7 +20,7 @@ layer_sf <- function(geom = NULL, stat = NULL,
     legend_key_type <- NULL
   }
 
-  # inherit from LayerSf class to add `legend_type` slot
+  # inherit from LayerSf class to add `legend_key_type` slot
   layer_class <- ggproto(NULL, LayerSf, legend_key_type = legend_key_type)
 
   layer(
--- a/R/layer-sf.R
+++ b/R/layer-sf.R
@@ -13,11 +13,21 @@ layer_sf <- function(geom = NULL, stat = NULL,
                      position = NULL, params = list(),
                      inherit.aes = TRUE, check.aes = TRUE, check.param = TRUE,
                      show.legend = NA) {
+  if (is.character(show.legend)) {
+    legend_key_type <- show.legend
+    show.legend <- TRUE
+  } else {
+    legend_key_type <- NULL
+  }
+
+  # inherit from LayerSf class to add `legend_type` slot
+  layer_class <- ggproto(NULL, LayerSf, legend_key_type = legend_key_type)
+
   layer(
     geom = geom, stat = stat, data = data, mapping = mapping,
     position = position, params = params, inherit.aes = inherit.aes,
     check.aes = check.aes, check.param = check.param,
-    show.legend = show.legend, layer_class = LayerSf
+    show.legend = show.legend, layer_class = layer_class
   )
 }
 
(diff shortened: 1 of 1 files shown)


yutannihilation:
Looks good, but as I don't quite remember the discussion on #3646, l need some more time to see the context...

yutannihilation:
After reading the comments around https://github.com/tidyverse/ggplot2/pull/3646#discussion_r353483822, I remembered I once proposed a similar approach, but it uses `param$legend` directly. Here's a version for this PR:

https://github.com/yutannihilation/ggplot2/commit/2b6317dd48f4be2b4a8cbdff0d1907cddac55544

Do you think we need to introduce a new parameter on `LayerSf` rather than recycling `legend` (or some other parameters in `geom_params`)?

clauswilke:
I *think* writing to `params$legend` directly can create a warning of an unknown parameter if you're using the layer with a geom that doesn't have a `legend` parameter.

yutannihilation:
I see. So, can we do this in `geom_sf()`?

yutannihilation:
I mean, if this is specific to geoms that have a `legend` parameter, it makes sense to do it in `geom_sf()`. If this is more general one, this should be done in `layer_sf()`. (And I'm getting to feel the latter is right...)

PR #3943, opened 2020-04-11
feat: add DocSearch as recommended by pkgdown

s-pace:
👋 team,

I am working on DocSearch. [We have an integration for pkgdown](https://pkgdown.r-lib.org/). We thought it is a shame you can not also used this search.

This PR will add DocSearch to the documentation website. It will allow a user to have a learn-as-you-type experience by displaying results thanks to a dropdown in a live way.

Let me know if you need anything.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -2,6 +2,10 @@ url: https://ggplot2.tidyverse.org
 
 template:
   package: tidytemplate
+  params:
+      docsearch:
+        api_key: e1261f3df95d2d4704c626ff11a4b536
+        index_name: tidyverse_ggplot2
 
 development:
   mode: auto


yutannihilation:
DocSearch was added by #3370 and then removed by #3374. I don't know about the detail, but it seems DocSearch didn't work at that time.

@batpigandme 
Do you think this is worth trying again?

s-pace:
It is now working yes

batpigandme:
Yeah, if it's working, then, sure thing.

s-pace:
It is https://docsearch.algolia.com/playground/?indexName=tidyverse_ggplot2&apiKey=e1261f3df95d2d4704c626ff11a4b536

yutannihilation:
Thanks, so the previous mistake was that we used the same key (`tidyverse`) over multiple repositories, right? Then, I have no concern.

One more question. Can we use this API key? It seems there are already several API keys used among tidyverse and r-lib repositories, but I don't know how they are managed...

* https://github.com/search?q=docsearch+org%3Atidyverse+filename%3Apkgdown.yml
* https://github.com/search?q=docsearch+org%3Ar-lib+filename%3Apkgdown.yml

s-pace:
Yes it is one key for a website

yutannihilation:
Thanks.

yutannihilation:
What I'm still not sure is that which is better to use the API key here or the new one someone in the tidyverse team issues, but I think it's fine to merge this PR anyway as we can replace the key anytime.

I'll wait for some days before merging this. If you have any other concern, please comment.

yutannihilation:
Merged. Thanks @s-pace for catching this :)

Issue #3839, opened 2020-02-27
Strange behavior of "group" aesthetic in geom_dotplot


gael-millot:
Is this behavior expected?
```r
# conversion of the vs and am columns of mtcars data frame into factors
mtcars$vs <- factor(mtcars$vs, labels = c("A", "B"))
mtcars$am <- factor(mtcars$am, labels = c("G", "H"))

# test of geom_dotplot using vs, mpg and am columns of mtcars
ggplot2::ggplot()+ggplot2::geom_dotplot(
    data = mtcars, 
    mapping = ggplot2::aes(x = vs, y = mpg, group = am), 
    position = ggplot2::position_dodge(), 
    binaxis = "y", 
    stackdir = "center",
    show.legend = TRUE
)
```
![1](https://user-images.githubusercontent.com/61543802/75440031-3be54a80-595b-11ea-91d4-9014c2b27c40.png)

Because other aesthetics gives the appropriate result, for instance when ```group``` is replaced by ```fill```:
```r
ggplot2::ggplot()+ggplot2::geom_dotplot(
    data = mtcars, 
    mapping = ggplot2::aes(x = vs, y = mpg, fill= am), 
    position = ggplot2::position_dodge(), 
    binaxis = "y", 
    stackdir = "center",
    show.legend = TRUE
)
```
![2](https://user-images.githubusercontent.com/61543802/75439883-ee68dd80-595a-11ea-9642-45d3c0c5ecaa.png)

The legend does not appear with ```group``` aesthetic, as expected, but the dot positions seem totally incoherent. It would be great if I could use ```group``` with the correct dot positioning.

Thanks for your help !

Best.

gael-millot:
Of note, ```group``` works fine when replacing the ```am``` category by ```vs``` (a single qualitative variable plotted instead of two):

```r
ggplot2::ggplot()+ggplot2::geom_dotplot(
    data = mtcars, 
    mapping = ggplot2::aes(x = vs, y = mpg, group = vs), 
    position = ggplot2::position_dodge(), 
    binaxis = "y", 
    stackdir = "center", 
    show.legend = TRUE
)
```
![3](https://user-images.githubusercontent.com/61543802/75441684-6389e200-595e-11ea-8dcd-3c3b12bd819f.png)

As if, with two categories, ```group``` takes the "average" position of ```A``` and ```B``` and then separates correctly between ```G``` and ```H```...

yutannihilation:
> Is this behavior expected?

[...]

PR #3950, opened 2020-04-16
Other examples added

gael-millot:
The end of the page points to https://github.com/tidyverse/ggplot2/issues/3839 to describe special behaviors of group

### DIFF
--- a/R/aes-group-order.r
+++ b/R/aes-group-order.r
@@ -72,5 +72,6 @@
 #' # we again need to override the default grouping for that layer with aes(group = Subject)
 #' boysbox <- boysbox + geom_boxplot()
 #' boysbox + geom_line(aes(group = Subject), colour = "blue")
-#' }
+#' 
+#' # Particular behaviors of group are detailed here https://github.com/tidyverse/ggplot2/issues/3839
 NULL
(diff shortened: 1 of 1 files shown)


R/aes-group-order.r
```diff
@@ -72,5 +72,6 @@
 #' # we again need to override the default grouping for that layer with aes(group = Subject)
 #' boysbox <- boysbox + geom_boxplot()
 #' boysbox + geom_line(aes(group = Subject), colour = "blue")
-#' }
```
yutannihilation, line 75: You cannot delete this `}`.

yutannihilation:
Ah, sorry, please ignore my comment above. It seems I need to run `/ document` after merging this into master...

By the way, what I meant on https://github.com/tidyverse/ggplot2/issues/3839#issuecomment-595291390 was to add some example *code*, not a link to the discussion. I think this PR is better than nothing, but this might not very handy for users.

thomasp85:
@gael-millot are you interested in fixing this up by summarising the points in #3839 rather than linking to it?

gael-millot:
Hi @thomasp85. I sent a new pull request. Thanks.

PR #3961, opened 2020-04-25
Test R 4.0 on CI

yutannihilation:
Now R 4.0.0 is released, ~~let's drop support for R 3.2 (c.f. we support only 4 previous versions of R https://www.tidyverse.org/blog/2019/04/r-version-support/).~~ let's add R 4.0 runners on CI

In addition, this PR contains the following tweaks:

* Remove a workaround related to xml2 installation problem, as the new version of xml2 was released on CRAN
* Add `cran:libgit2` PPA to avoid conflicts related to `libcurl4-gnutls-dev` (c.f. https://github.com/r-hub/sysreqsdb/issues/77#issuecomment-620025428)

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -22,14 +22,14 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: '3.6', vdiffr: true}
-          - {os: macOS-latest, r: '3.6', vdiffr: true}
-          - {os: macOS-latest, r: 'devel', vdiffr: false}
-          - {os: ubuntu-16.04, r: '3.2', vdiffr: false, cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.3', vdiffr: false, cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.4', vdiffr: false, cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04, r: '3.5', vdiffr: false, cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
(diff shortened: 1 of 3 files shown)


jimhester:
I generally think it is better to be explicit with what versions you are checking rather than using the aliases.

One reason is that you will have to bump the cache numbers to invalidate the cache when the versions change anyway, so if you have to edit the yaml anyway might as well make it very clear what versions you are checking against.

yutannihilation:
Thanks! It makes sense. I was wondering why `R-CMD-check.yaml` refers to the `matrix.config.r` instead of the actual R version (e.g. we can include the `version` in `depends.Rds`), but now I understand the design.

yutannihilation:
Aw, true. Sorry, probably I should have asked Hadley and Thomas.

>     1. Do we want to change the R version requirement for an upcoming minor release (presumably ggplot2 3.3.1 should come out soon)?

Yes, since the second next release (currently expected as "v3.4.0") should be distant, I thought we should be clear about dropping the support in the next version to prevent confusion.

>     2. Do we always increase the version in Depends when we drop supported versions? As far as I can tell, there's no change in the code that causes ggplot2 to break for R 3.2, we're just not checking anymore.

Yes, I think we are supposed to increase it even if there's no actual breakage, but I might misunderstand the rule of tidyverse.

@hadley @thomasp85 
As R 4.0.0 is released, all tidyverse packages are going to bump the R version requirement, right? If so, I want to include it in the upcoming v3.3.1. Does it sound good to you? Or, do we need some strategy about the release schedules?

jimhester:
No, you don't need to bump the minor version of the required version of R unless there is something you use in ggplot2 that would require it.

yutannihilation:
Oh, sorry. I misunderstood the process... Then, should I keep the R 3.2 runner to know when we actually need to drop the support?

hadley:
@yutannihilation Yes. Generally what we do is keep the dependency and runners as is (just adding a new one for the latest version), and then only drop when we have to (i.e. when you use a function or package that needs 3.3). We should really document this process somewhere 😬

yutannihilation:
@hadley 
OK, thanks for clarifying! I'll do this better next year...

@clauswilke
Sorry for confusing you. I'll restore the minimum requirement and the CI runners.

Issue #3962, opened 2020-04-25
Shortcut for `scale_y_continuous(expand = expansion(c(0, 0.05))`


thomas-neitmann:
When creating histogram or bar charts the standard expansion factor of  `0.05` creates an unsighty gap between the bars and the x axis.

```
library(ggplot2)
data(mtcars)
ggplot(mtcars, aes(hp)) +
  geom_histogram()
```

![image](https://user-images.githubusercontent.com/11788080/80287984-0ed3cf00-8735-11ea-8f62-96140c2788a4.png)

To remove this gap currently one has to add `scale_y_continuous(expand = expansion(c(0, 0.05))` to the plot. This is cumbersome to type, easy to forget and hard to grasp for beginners.

I'd love to have a shortcut for that, something like:
```
scale_y_tight <- function(...) {
  scale_y_continuous(expand = expansion(0, 0.05), ...)
}
```

This is in the same spirit as `ylab()` or `ylim()`.

Thinking a bit further maybe this could even be an option:

```
exact_ylim <- function(...) {
  scale_y_continuous(limits = c(...), expand = expansion())
}
```
This would force the axis limits to be exactly those requested which is not the case currently with `ylim()`.

clauswilke:
I think this is a good idea, but I'd propose a different interface. We already have `ylab()` and `ylim()`, so it would make sense to add something like `yexpand()`. It could take either string arguments (e.g. `"both"`, `"none"`, `"upper"`, `"lower"`), which would apply default expansions on both side, on neither side, or on the upper or lower end only, or numerical arguments similar to `expansion()`.

Not sure how to implement this though. I'll have to look into how `ylab()` and `ylim()` communicate their arguments to the relevant scale.

thomas-neitmann:
I like the idea of  `yexpand()` but then the question arises what should be the default for the argument. For the use case I have in mind it should be `"upper" ` which feels kind of unintuitive because you want to get rid of the lower expansion.

clauswilke:
I've thought about this a bit and at this time I'm not sure how to implement this. `xlab()` and `xlim()` are implemented using entirely different mechanisms, and both seem ad-hoc. In general, we don't have a good approach to feeding in additional parameters into a scale that exists already. The same problems would arise if we wanted to, for example, set the breaks separately from the color mapping in a color scale. This has broader implications for other issues, such as default scales via theme (#2691), and we may have to think carefully about a good solution to this issue.

[...]

PR #3973, opened 2020-04-29
WIP: Register default scales via themes 

### LINKED ISSUE #2691: Register default scales via themes

Alanocallaghan:
Resolve #2691 

Add `default.scales` component of themes.


R/scale-type.R
```diff
@@ -25,7 +24,10 @@ find_scale <- function(aes, x, env = parent.frame()) {
 # Look for object first in parent environment and if not found, then in
 # ggplot2 namespace environment.  This makes it possible to override default
 # scales by setting them in the parent environment.
-find_global <- function(name, env, mode = "any") {
+find_global <- function(name, env, mode = "any", theme_ = theme()) {
```
clauswilke, line 27: We don't normally use argument names ending in `_`, as in `theme_`? Does `theme = theme()` not work here?

R/scale-type.R
```diff
@@ -25,7 +24,10 @@ find_scale <- function(aes, x, env = parent.frame()) {
...
+  if (!is.null(theme_[["default.scales"]][[name]]) && mode(theme_[["default.scales"]][[name]]) == mode) {
+    return(theme_[["default.scales"]][[name]])
```
clauswilke, line 29: Theme elements should generally be retrieved with `calc_element()`, not by direct lookup.

clauswilke: Also, you're doing the same lookup three times. I think it makes more sense to do it once and assign to a temporary variable.

R/scale-type.R
Alanocallaghan: I got `promise already under evaluation`, otherwise I'd have gone with that. I can do `theme=ggplot2::theme()` instead. See reprex:

``` r
x <- "Hello world"
f <- function(get = get("x")) {
  print(get)
}
f()
#> Error in print(get): promise already under evaluation: recursive default argument reference or earlier problems?
```

R/scale-type.R
Alanocallaghan: Ah thanks, not poked around in internals much so I'm likely to miss things like that.

R/scale-type.R
```diff
@@ -1,4 +1,4 @@
-find_scale <- function(aes, x, env = parent.frame()) {
+find_scale <- function(aes, x, env = parent.frame(), theme) {
```
clauswilke, line 1: I think `theme` should have a default here. I think `theme = NULL` makes sense.

R/scale-type.R
```diff
@@ -25,7 +24,13 @@ find_scale <- function(aes, x, env = parent.frame()) {
 # Look for object first in parent environment and if not found, then in
 # ggplot2 namespace environment.  This makes it possible to override default
 # scales by setting them in the parent environment.
-find_global <- function(name, env, mode = "any") {
+find_global <- function(name, env, mode = "any", theme = ggplot2::theme()) {
```
clauswilke, line 27: Thinking some more about this, I don't understand the point of `theme = ggplot2::theme()`. Just write `theme = NULL`.

R/scale-type.R
```diff
@@ -25,7 +24,13 @@ find_scale <- function(aes, x, env = parent.frame()) {
...
+  ## there is presumably a more elegant way to test presence and mode in a list.
```
clauswilke, line 28: Not sure what you mean. The code looks fine as is.

R/scale-type.R
```diff
@@ -25,7 +24,13 @@ find_scale <- function(aes, x, env = parent.frame()) {
...
+  default_scales <- calc_element("default.scales", theme)
+  scale <- default_scales[[name]]
+  if (!is.null(scale) & mode(scale) == mode) {
```
clauswilke, line 31: Please use `&&` instead of `&` and write `isTRUE(mode(scale) == mode)`.

R/plot-build.r
```diff
@@ -44,6 +44,8 @@ ggplot_build.ggplot <- function(plot) {
     out
   }
 
+  plot$theme <- plot_theme(plot)
```
clauswilke, line 47: This should be commented, just like the other parts of the code in this function.

R/theme-elements.r
```diff
@@ -480,6 +480,7 @@ el_def <- function(class = NULL, inherit = NULL, description = NULL) {
   strip.placement.y   = el_def("character", "strip.placement"),
   strip.switch.pad.grid = el_def("unit"),
   strip.switch.pad.wrap = el_def("unit"),
+  default.scales      = el_def("list"),
```
clauswilke, line 483: It's probably best to add a specific element class here. PR #2749 adds `element_geom()`, so maybe this one here could be called `element_scales()`. But I'd like to have input from @yutannihilation or @thomasp85 here.

R/scale-type.R
clauswilke: As I commented elsewhere, just write `theme = NULL`.

clauswilke:
Apologies, my picture is evolving as I'm thinking about this: I don't think you should modify `find_global()`. That makes no sense. Instead, move the theme lookup code into `find_scale()`. If `find_scale()` doesn't find an appropriate scale in the theme, then it can try the global lookup with the unmodified `find_global()`. Makes much more sense.

R/theme-elements.r
clauswilke: Also, if we add a specific element class then we check then and there that the element values are functions and then we can avoid the `mode()` check in `find_global()` (to be moved into `find_scale()`).

R/scale-type.R
Alanocallaghan: Comment is related to a previous commit and should be ignored

Alanocallaghan:
Thanks, yeah I figured `find_global` was the wrong place entirely. `find_scale` makes sense, there's no reason to look in the theme in general when doing lookup.

clauswilke:
@Alanocallaghan I realized there's a different issue and it suggests that this PR may be premature. We don't have a good way of feeding additional parameters into an existing scale, such as breaks, break labels, or expansion parameters. Without this mechanism, setting a default scale in a theme has limited utility and can actually cause confusion, when people want to set the breaks and cannot do this while also retaining the color scheme their theme defines. Issue #3962 is related and may have to be resolved first.

@baptiste It seems to me that you're proposing an entirely new way of defining scales. Your proposal may make sense, but I'd like to see a proof-of-concept implementation to really understand what exactly the proposal is and how it'd interact with all the currently existing ggplot2 code. I'd also want to make sure that it works for shape, size, alpha, etc. scales in addition to color scales.

Issue #3978, opened 2020-05-01
names of facet_grid() panels in gtable (panel-x-y) are inconsistent


fabern:
The names of the `grobs` in the `gtable` representing the panels created with `facet_grid()` are inconsistent. I would expect them to follow the pattern `panel-[row.nr]-[col.nr]`, but they don't.

``` r
library(ggplot2)
library(grid)
library(gtable)
dat <- data.frame(x = c(1, 1, 1, 1, 1, 1), 
                  y = c(1, 1, 1, 1, 1, 1), 
                  a = c("a1", "a1", "a1", "a2", "a2", "a2"), 
                  b = c("b1", "b2", "b3", "b1", "b2", "b3"))
p <- ggplot(dat, aes(x=x,y=y)) + 
  geom_point() + 
  facet_grid(vars(a),vars(b))

# Annotate facets with grob name:
g <- ggplotGrob(p)
gtable_panel_positions <-  g$layout[grepl("panel-", g$layout$name),]
for (it in seq_len(nrow(gtable_panel_positions))) {
  g <- gtable_add_grob(g,
                       grobs = textGrob(label=gtable_panel_positions[it,"name"]),
                       t = gtable_panel_positions[[it,"t"]],
                       l = gtable_panel_positions[it,"l"])
}
grid.newpage()
grid.draw(g)
```

![](https://i.imgur.com/8XdM70h.png)

<sup>Created on 2020-05-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
```

fabern:
Below example illustrates the issue in an alternative way by using `gtable_filter()`:

``` r
library(ggplot2)
library(grid)
library(gtable)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
dat <- data.frame(x = c(1, 1, 1, 1, 1, 1), 
                  y = c(1, 1, 1, 1, 1, 1), 
                  a = c("a1", "a1", "a1", "a2", "a2", "a2"), 
                  b = c("b1", "b2", "b3", "b1", "b2", "b3"))
p <- ggplot(dat, aes(x=x,y=y)) + 
  geom_point() + 
  facet_grid(vars(a),vars(b))

# Example B (requires dplyr)
p %>%
  ggplotGrob() %>%
[...]

PR #3979, opened 2020-05-01
Renamed panel titles in gtable of plot with facet_grid()

fabern:
This fixes the inconsistent naming in the `gtable` of `facet_grid()` panels to result in the pattern `panel-[row.nr]-[col.nr]`.

Reprex in issue: https://github.com/tidyverse/ggplot2/issues/3978


thomasp85:
Thanks

PR #3984, opened 2020-05-02
Make stat_function() work with empty input data.

### LINKED ISSUE #3983: stat_function() doesn't work without dummy data
I would have expected the second and third example below to work, but it doesn't.

``` r
library(ggplot2)

# works
ggplot(data.frame(x = c(-5, 5)), aes(x)) + 
  stat_function(fun = dnorm)
```

``` r

# doesn't work
ggplot() + xlim(-5, 5) + 
  stat_function(fun = dnorm)
```

``` r

# doesn't work
ggplot() + 
  stat_function(fun = dnorm, xlim = c(-5, 5))
```

<sup>Created on 2020-05-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The problem seems to be that `stat_function()` doesn't return any data in the second and third case, or the data gets somehow deleted. It would be nice if we could solve this before the 3.3.1 release, I assume (hope!) it's a simple fix.

``` r
library(ggplot2)

p <- ggplot(data.frame(x = c(-5, 5)), aes(x)) + 
  stat_function(fun = dnorm)
head(layer_data(p))
#>      x            y PANEL group colour size linetype alpha
#> 1 -5.0 1.486720e-06     1    -1  black  0.5        1    NA
#> 2 -4.9 2.438961e-06     1    -1  black  0.5        1    NA
#> 3 -4.8 3.961299e-06     1    -1  black  0.5        1    NA
#> 4 -4.7 6.369825e-06     1    -1  black  0.5        1    NA
#> 5 -4.6 1.014085e-05     1    -1  black  0.5        1    NA
#> 6 -4.5 1.598374e-05     1    -1  black  0.5        1    NA

p <- ggplot() + xlim(-5, 5) + 
  stat_function(fun = dnorm)
layer_data(p)
#> data frame with 0 columns and 0 rows

p <- ggplot() + 
  stat_function(fun = dnorm, xlim = c(-5, 5))
layer_data(p)
#> data frame with 0 columns and 0 rows
```

<sup>Created on 2020-05-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
Fixes #3983.

### DIFF
--- a/R/geom-function.R
+++ b/R/geom-function.R
@@ -18,7 +18,7 @@
 #'   geom_function(fun = dnorm, colour = "red")
 #'
 #' # To plot functions without data, specify range of x-axis
-#' base <- ggplot(data.frame(x = c(-5, 5)), aes(x))
+#' base <- ggplot() + xlim(-5, 5)
 #' base + geom_function(fun = dnorm)
 #' base + geom_function(fun = dnorm, args = list(mean = 2, sd = .5))
 #'
@@ -45,8 +45,11 @@ geom_function <- function(mapping = NULL, data = NULL, stat = "function",
                           position = "identity", ..., na.rm = FALSE,
                           show.legend = NA, inherit.aes = TRUE) {
   # Warn if supplied data is going to be overwritten
-  if (!is.null(data) && identical(stat, "function")) {
-    warn("`data` is not used by stat_function()")
+  if (identical(stat, "function")) {
+    if (!is.null(data)) {
+      warn("`data` is not used by stat_function()")
+    }
+    data <- ensure_nonempty_data
   }
 
   layer(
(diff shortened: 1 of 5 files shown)


clauswilke:
Still needs an additional unit test or two for the new code paths.

clauswilke:
The checks currently fail because I forgot to rebuild the documentation.

clauswilke:
This is ready for review now.

R/geom-function.R
```diff
@@ -45,8 +45,11 @@ geom_function <- function(mapping = NULL, data = NULL, stat = "function",
                           position = "identity", ..., na.rm = FALSE,
                           show.legend = NA, inherit.aes = TRUE) {
   # Warn if supplied data is going to be overwritten
-  if (!is.null(data) && identical(stat, "function")) {
-    warn("`data` is not used by stat_function()")
+  if (identical(stat, "function")) {
```
thomasp85, line 48: You can in theory provide a ggproto object directly... This check will fail if a user for some reason pass in `StatFunction` (I know this is an esoteric issue)

clauswilke: Yes, I think it's fine, since the worst case scenario is no warning.

PR #3998, opened 2020-05-15
Make discrete scale determination stable

### LINKED ISSUE #3873: `geom_col` mishandles x-axis `Date` with decimals
`geom_col` does not handle `Date` on the x-axis when these contain decimals. Note that the input is still a valid `Date` object. This currently breaks `incidence::plot.incidence` as reported in [this issue](https://github.com/reconhub/incidence/issues/119).

``` r
library(ggplot2)
x <- data.frame(date = Sys.Date() + 1:10, 
                y = 1:10)


## this works
ggplot(x) + geom_col(aes(x = date, y = y))
```

``` r

## this does not
ggplot(x) + geom_col(aes(x = date + 0.5, y = y))
```

``` r


## not sure if it affects other geoms; geom_point is fine:
ggplot(x) + geom_point(aes(x = date + 0.5, y = y))
```

<sup>Created on 2020-03-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #3932: Unexpected (?) orientation with geom_col
I'm using ggplot2 v3.3.0.9000. The orientation of the `geom_col` bars is not correct when values of x are numeric and less than 1. Specifying `orientation = "x"` does what I would expect should be default behaviour. From reading the documentation (https://ggplot2.tidyverse.org/reference/geom_bar.html), it does state `x` is the default oreintation but that doesn't seem to be the case here.

``` r
library(tidyverse)

df <-
data.frame(
           x = c(0.1, 0.2, 0.3),
           y = c(10L, 20L, 30L)
      )
  
ggplot(df) +
  geom_col(aes(x = x, y = y))
```

``` r


ggplot(df) +
  geom_col(aes(x = x * 10, y = y))
```

``` r

ggplot(df) +
  geom_col(aes(x, y), orientation = "x")
```

### LINKED ISSUE #3939: geom_bar not working with yearmon date format
In v3.3.0 of ggplot2 the handling of date-type variables on the x-axis is inconsistent in geom_bar(). Specifically, when using zoo's yearmon class for dates, the plot fails to render correctly. (I was not able to produce an MWE without zoo.) The month variable does not render properly on the axis and the data plotted is structured incorrectly. 

The bug does not appear when using geom_line().

```r

library(zoo)
library(ggplot2)


df_mwe <- structure(list(month = structure(c(1997.41666666667, 1997.5, 1997.5, 1997.58333333333, 1997.58333333333, 
                                              1997.66666666667, 1997.66666666667, 1997.75, 1997.75, 1997.83333333333, 
                                              1997.83333333333, 1997.91666666667, 1997.91666666667, 1998, 1998, 
                                              1998.08333333333, 1998.08333333333, 1998.16666666667, 1998.16666666667, 
                                              1998.25, 1998.25), class = "yearmon"), 
                         
                          value = c(29480, 147850, 20430, 162930, 25510, 236670, 31340, 140940, 
                                    21070, 178170, 31750, 209430, 37840, 244560, 44960, 259310, 
                                    47700, 277570, 43860, 167700, 21120)), row.names = c(NA, 21L), class = "data.frame")


df_mwe$month <- as.yearmon(df_mwe$month)

#' fails:
ggplot(df_mwe) +
  geom_bar(aes(x=month, y= value),  stat='identity')

#' works fine
ggplot(df_mwe) +
  geom_line(aes(x=month, y= value),   stat='identity')

#' works fine
df_mwe$month <- as.Date(df_mwe$month)
ggplot(df_mwe) +
  geom_bar(aes(x=month, y= value),      stat='identity')

```
geom_bar with yearmon:
image

geom_line with yearmon:
image

geom_bar with Date
image

### LINKED ISSUE #3957: geom_boxplot() on continuous x-axis is flipped unexpectedly(?)
Still not sure if this should be flipped or not, but let me file this issue here.

``` r
library(ggplot2)

d <- dplyr::filter(diamonds, carat < 3)
ggplot(d, aes(x = carat, y = price)) +
  geom_boxplot(aes(group = cut_width(carat, 0.1)))
```

<sup>Created on 2020-04-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #3991: geom_boxplot throws error when all y are 0 and an NA exists
`Error in if (all(y == 0)) { : missing value where TRUE/FALSE needed`

Followed the error down with browser:
 `geom_boxplot` [line 136](https://github.com/tidyverse/ggplot2/blob/660aad2db2b3495ae0d8040915a40d247133ffc0/R/geom-boxplot.r#L136) >
 `layer`  [line 134](https://github.com/tidyverse/ggplot2/blob/e9b9946786dae861dea1c352f2ac7b2a837b5f82/R/layer.r#L134) > ggproto
I'm not sure where the offending line implicated by the error is as that as far as I could follow it with `debug`.
Most of the other `geom_*` seem to effectively omit NA values, and so I thought that it might be helpful to report this exception.
```r
library(ggplot2)
df <- data.frame(d = c(0, 0, NA, 0), dt = seq(Sys.Date() - 3, Sys.Date(), by = "1 days"))
ggplot(df, aes(x = dt)) +
  geom_boxplot(aes(y = d))
```

thomasp85:
Fix #3991, Fix #3957, Fix #3939, Fix #3932, Fix #3873

This PR addresses the many spurious edge cases in orientation determination introduced in 3.3.0. Instead of guessing based on whether the position values are "discrete-like", the discrete scale now returns a numeric subclass identifying the values as coming from a discrete mapping. I've implemented the minimal interface for this information to be carried through the data pipeline, but once we adopt vctrs this is a clear case for that.

@hadley can you make sure that my subclassing doesn't commit any cardinal sins


R/scale-discrete-.r
```diff
@@ -129,3 +128,26 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
     new
   }
 )
+
+# TODO: This is a clear candidate for vctrs once we adopt it
+new_mapped_discrete <- function(x) {
+  class(x) <- c("mapped_discrete", "numeric")
```
hadley, line 134: Just for safety, this should check that `x` is numeric

R/scale-discrete-.r
```diff
@@ -129,3 +128,26 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+#' @export
+c.mapped_discrete <- function(..., recursive = FALSE) {
+  new_mapped_discrete(c(unlist(lapply(list(...), unclass))))
+}
+#' @export
+`[.mapped_discrete` <- function(x, ..., drop = TRUE) {
+  new_mapped_discrete(NextMethod())
+}
+#' @export
+`[<-.mapped_discrete` <- function(x, ..., value) {
```
hadley, line 147: I don't think this is correct in general, but it should be fine for internal use. You might want to check that `value` is numeric just to be safe.

tests/testthat/test-build.r
```diff
@@ -25,8 +25,8 @@ test_that("position aesthetics are coerced to correct type", {
   l2 <- ggplot(df, aes(x, z)) + geom_point() + scale_x_discrete()
   d2 <- layer_data(l2, 1)
 
-  expect_is(d2$x, "integer")
-  expect_is(d2$y, "integer")
+  expect_is(d2$x, "mapped_discrete")
```
hadley, line 28: Should change this to `expect_s3_class()` since you've touched it

R/scale-discrete-.r
thomasp85: this is mapped after `[<-.Date`, but I will (as mentioned) port this to vctrs once we embrace it

thomasp85:
there shouldn't be any downstream consequences. this change is basically circumventing all the guesswork that was part of the last implementation and thus avoids the false positives. As the class is a very thin wrapper around numeric the worst that could happen is that the class gets dropped and the orientation will default to the old behaviour

yutannihilation:
Does this mean a subclass of `ScaleDiscretePosition` need to call `ggproto_parent(ScaleDiscretePosition, self)$map(x, limits)` if it implements its own `map()` method? It seems there's no such a package on CRAN at the moment ([search result](https://github.com/search?q=org%3Acran+ScaleDiscretePosition&type=Code)), so it won't be a big problem anyway, though.

thomasp85:
They don’t _need_ to, but if they want the automatic orientation deduction they should

PR #4013, opened 2020-05-19
Add `bounds` argument to `geom_density()`

echasnovski:
This is a PR for #3387 and serves as a place for discussion and feedback about possible change.

The goal is to **add `bounds` argument to `geom_density()` and `stat_density()`, which will allow users to specify known constraints on data to obtain more "realistic" density estimation**. Change to computed output of `stats::density()` is made using relatively simple "reflection method of boundary correction" *roughly* taken from "Simple boundary correction for kernel density estimation” by M.C. Jones, 1993 (you can look at free [pdf-file of presentation based on this article](http://www.buch-kromann.dk/tine/nonpar/Simple%20boundary%20correction,%20Jones%201993.pdf), in particular, page 7).

Currently this PR serves two purposes:

- Show that this is a completely non-breaking update. All previous behavior is explicitly preserved.
- Allow everyone to "play" with it.

If agreed that this is a desirable change, at least the following should also be made (which I am happy to do):

- Figure out when and how to notify user about "bad bounds". For example, when some or all data points are outside of `bounds` interval, or simply it doesn't represent a valid interval.
- Add examples of `bounds` usage.

### DIFF
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -15,6 +15,8 @@
 #'   not line-up, and hence you won't be able to stack density values.
 #'   This parameter only matters if you are displaying multiple densities in
 #'   one plot or if you are manually adjusting the scale limits.
+#' @param bounds Known lower and upper bounds for estimated data. Default
+#'   `c(-Inf, Inf)` means that there are no (finite) bounds.
 #' @section Computed variables:
 #' \describe{
 #'   \item{density}{density estimate}
@@ -35,6 +37,7 @@ stat_density <- function(mapping = NULL, data = NULL,
                          n = 512,
                          trim = FALSE,
                          na.rm = FALSE,
+                         bounds = c(-Inf, Inf),
                          orientation = NA,
                          show.legend = NA,
                          inherit.aes = TRUE) {
@@ -54,6 +57,7 @@ stat_density <- function(mapping = NULL, data = NULL,
       n = n,
       trim = trim,
       na.rm = na.rm,
+      bounds = bounds,
       orientation = orientation,
       ...
     )
@@ -84,7 +88,8 @@ StatDensity <- ggproto("StatDensity", Stat,
   extra_params = c("na.rm", "orientation"),
 
   compute_group = function(data, scales, bw = "nrd0", adjust = 1, kernel = "gaussian",
-                           n = 512, trim = FALSE, na.rm = FALSE, flipped_aes = FALSE) {
+                           n = 512, trim = FALSE, na.rm = FALSE, bounds = c(-Inf, Inf),
+                           flipped_aes = FALSE) {
     data <- flip_data(data, flipped_aes)
     if (trim) {
       range <- range(data$x, na.rm = TRUE)
@@ -93,7 +98,8 @@ StatDensity <- ggproto("StatDensity", Stat,
     }
 
     density <- compute_density(data$x, data$weight, from = range[1],
-      to = range[2], bw = bw, adjust = adjust, kernel = kernel, n = n)
+      to = range[2], bw = bw, adjust = adjust, kernel = kernel, n = n,
+      bounds = bounds)
     density$flipped_aes <- flipped_aes
     flip_data(density, flipped_aes)
   }
@@ -101,7 +107,8 @@ StatDensity <- ggproto("StatDensity", Stat,
 )
 
 compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
-                            kernel = "gaussian", n = 512) {
+                            kernel = "gaussian", n = 512,
+                            bounds = c(-Inf, Inf)) {
   nx <- length(x)
   if (is.null(w)) {
     w <- rep(1 / nx, nx)
@@ -122,8 +129,15 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
     ), n = 1))
   }
 
-  dens <- stats::density(x, weights = w, bw = bw, adjust = adjust,
-    kernel = kernel, n = n, from = from, to = to)
+  if (all(is.infinite(bounds))) {
+    dens <- stats::density(x, weights = w, bw = bw, adjust = adjust,
+                           kernel = kernel, n = n, from = from, to = to)
+  } else {
+    dens <- stats::density(x, weights = w, bw = bw, adjust = adjust,
+                           kernel = kernel, n = n)
+
+    dens <- reflect_density(dens = dens, bounds = bounds, from = from, to = to)
+  }
 
   new_data_frame(list(
     x = dens$x,
@@ -133,4 +147,27 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
     count =   dens$y * nx,
     n = nx
   ), n = length(dens$x))
+}
+
+reflect_density <- function(dens, bounds, from, to) {
+  if (all(is.infinite(bounds))) {
+    return(dens)
+  }
+
+  f_dens <- stats::approxfun(
+    x = dens$x, y = dens$y, method = "linear", yleft = 0, yright = 0
+  )
+
+  out_x <- intersection_grid(dens$x, bounds, from, to)
+  out_y <- f_dens(out_x) + f_dens(bounds[1] + (bounds[1] - out_x)) +
+    f_dens(bounds[2] + (bounds[2] - out_x))
+
+  list(x = out_x, y = out_y)
+}
+
+intersection_grid <- function(grid, bounds, from, to) {
+  left <- max(from, bounds[1])
+  right <- min(to, bounds[2])
+
+  seq(from = left, to = right, length.out = length(grid))
 }
--- a/man/geom_density.Rd
+++ b/man/geom_density.Rd
@@ -30,6 +30,7 @@ stat_density(
   n = 512,
   trim = FALSE,
   na.rm = FALSE,
+  bounds = c(-Inf, Inf),
   orientation = NA,
   show.legend = NA,
   inherit.aes = TRUE
@@ -111,6 +112,9 @@ range of that group: this typically means the estimated x values will
 not line-up, and hence you won't be able to stack density values.
 This parameter only matters if you are displaying multiple densities in
 one plot or if you are manually adjusting the scale limits.}
+
+\item{bounds}{Known lower and upper bounds for estimated data. Default
+\code{c(-Inf, Inf)} means that there are no (finite) bounds.}
 }
 \description{
 Computes and draws kernel density estimate, which is a smoothed version of


echasnovski:
Examples of usage:

``` r
library(ggplot2)
library(tibble)

set.seed(101)

ggplot(tibble(x = runif(100)), aes(x)) +
  geom_density() +
  geom_density(bounds = c(0, 1), color = "blue", ) +
  stat_function(fun = dunif, color = "red")
```



``` r

ggplot(tibble(x = rexp(100)), aes(x)) +
  geom_density() +
  geom_density(bounds = c(0, Inf), color = "blue") +
  stat_function(fun = dexp, color = "red")
```



<sup>Created on 2020-05-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
@clauswilke do you have time to take a look at this - I don't have any problems with the functionality but I'm quite rusty on density estimation code

thomasp85:
This is not pressing btw - it will not be part of the patch release

clauswilke:
Sure, assigned to myself.

@echasnovski I will review within the next few weeks or so. If I don't, please ping me to remind me.

echasnovski:
`ping @clauswilke`

clauswilke:
Apologies, please ping again after May 10. I have two more weeks of teaching and it's using up nearly all my time.

echasnovski:
Sure, no worries :)

echasnovski:
@clauswilke, how about `ping` number 2?

R/stat-density.r
```diff
@@ -122,8 +129,15 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
     ), n = 1))
   }
 
-  dens <- stats::density(x, weights = w, bw = bw, adjust = adjust,
-    kernel = kernel, n = n, from = from, to = to)
+  if (all(is.infinite(bounds))) {
```
clauswilke, line 132: Please add some comments here explaining the thought process.

R/stat-density.r
```diff
@@ -134,3 +148,26 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
     n = nx
   ), n = length(dens$x))
 }
+
+reflect_density <- function(dens, bounds, from, to) {
```
clauswilke, line 152: Please add some comments explaining what the function does and what the parameters are.

R/stat-density.r
```diff
@@ -134,3 +148,26 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
...
+  )
+
+  out_x <- intersection_grid(dens$x, bounds, from, to)
+  out_y <- f_dens(out_x) + f_dens(bounds[1] + (bounds[1] - out_x)) +
+    f_dens(bounds[2] + (bounds[2] - out_x))
+
+  list(x = out_x, y = out_y)
+}
+
+intersection_grid <- function(grid, bounds, from, to) {
```
clauswilke, line 168: Please add some comments explaining what the function does and what the parameters are.

R/stat-density.r
```diff
@@ -15,6 +15,8 @@
 #'   not line-up, and hence you won't be able to stack density values.
 #'   This parameter only matters if you are displaying multiple densities in
 #'   one plot or if you are manually adjusting the scale limits.
+#' @param bounds Known lower and upper bounds for estimated data. Default
+#'   `c(-Inf, Inf)` means that there are no (finite) bounds.
```
clauswilke, line 19: There should be some explanation somewhere in the documentation of what happens when the bounds are finite. And what happens if data points fall outside the set bounds?

clauswilke (commented):
Generally looks reasonable. But please add some comments to document the logic, and also explain the logic in the exported documentation.

R/stat-density.r
echasnovski: This was more of an open question to you. What do you think is the best behavior here? It mostly depends on the implied severity of `bounds` argument.

Currently no check is done. It estimates density and then tries to "reflect" it at edges. There are some extreme cases (like when bounds and data range not intersect at all), which I decided to tackle after the decision about `bounds` severity.

I can suggest at least three options:
- Give a warning when at least one point is outside of `bounds`. Proceed as nothing happened. Implies that `bounds` is essentially for visualization purpose.
- Remove points outside of `bounds` before or after estimating density. This implies that `bounds` is a data generation assumption and tries to fix it.
- Give an error if at least one point is outside. This also implies data generation but stops early. I personally like this the most.

echasnovski: This was more of an open question to package authors. What do you think is the best behavior here? It mostly depends on the implied severity of bounds argument.

Currently no check is done. It estimates density and then tries to "reflect" it at edges. There are some extreme cases (like when bounds and data range not intersect at all), which I decided to tackle after the decision about bounds severity.

I can suggest at least three options:
- Give a warning when at least one point is outside of bounds. Proceed as nothing happened. Implies that bounds is essentially for visualization purpose.
- Remove points outside of bounds before or after estimating density. This implies that bounds is a data generation assumption and tries to fix it.
- Give an error if at least one point is outside. This also implies data generation but stops early. I personally like this the most.

clauswilke:
With this PR, would the `bounds` argument generally be preferred over `from` and `to`? I notice that `from` and `to` are not documented, so that's maybe a good thing.

echasnovski:
> With this PR, would the `bounds` argument generally be preferred over `from` and `to`? I notice that `from` and `to` are not documented, so that's maybe a good thing.

As I understand it, `from` and `to` are not arguments here. They are computed based on data range ([here](https://github.com/tidyverse/ggplot2/blob/acaa98e4cfcd922a737f35c16ec3dd6079aab3dc/R/stat-density.r#L95-L96)) and result into computed density being clipped at those edges.

Proposed `bounds` and pair of computed `from` and `to` result into two completely different behaviors. While `from` and `to` don't affect estimated density (they simply ignore outer tails which result into plot not having total square of 1), `bounds` affects estimated density by "reflecting" those tails inside bounded segment.

thomasp85:
@clauswilke are you interested in finishing this off with @echasnovski ?

echasnovski:
I have already forgotten the details of it :(
But if you (plural) think it is a worthwhile addition to {ggplot2}, I'll try to put some time into it.

clauswilke:
Same here. Forgotten the details, but I think it's reasonable to revive.

R/stat-density.r
echasnovski: Done.

echasnovski:
I addressed previous comments. All other work seems to be better done after consensus about how severe `bounds` argument should be (see comments).

thomasp85:
@clauswilke I'll let you decide on the final points for this PR

clauswilke:
@echasnovski Sorry, I missed your update from May 12. Please always make sure to tag me, as I don't currently have the time to read every ggplot2 notification that comes my way but I try to pay attention to cases where I'm tagged.

Regarding bounds, I'm not generally in favor of any approach that errors out when it doesn't have to. Sometimes users want to do weird stuff for valid reasons, and we shouldn't prevent them from doing so. I think issuing a warning but otherwise performing a reasonable action is the way to go. So maybe remove points outside of bounds and warn that this has happened? This would be similar to the treatment of limits in scales functions.

echasnovski:
> @echasnovski Yes, some tests would be good. Just because there is no test currently for stat_density() doesn't mean we shouldn't add one. It would actually be a good idea to test the full density estimation. Create a small data set, run it through ggplot2, extract the data with layer_data(), then also perform the density estimation manually in the test script, and compare that the two are the same.

@clauswilke, this showed to be not trivial. The reason is that `ggplot()` outputs data frame with grid range equal to data range, while `density()` makes its own, extended range. I ended up testing functional approximation at some test sample.

Other than that, I added snapshot tests for different `bounds` and how `stat_density()` handles data outside of `bounds`. This feels like enough test cases while not being too restrictive.

Also I didn't manage to incorporate weights into density estimation. Here is what I got:

```r
ggplot(mtcars, aes(mpg, weight = cyl)) + stat_density()
Warning message:
The following aesthetics were dropped during statistical transformation: weight
ℹ This can happen when ggplot fails to infer the correct grouping structure in the data.
ℹ Did you forget to specify a `group` aesthetic or to convert a numerical variable into a factor? 
```

> Also test that no adjustments are made when no boundaries are set.

Not entirely sure what you mean here. No boundary correction with default `boundary` is tested at equivalence of `geom_density()` and plain `stats::density()`.

tests/testthat/test-stat-density.R
```diff
@@ -1,3 +1,54 @@
...
+  }
+
+  mpg_min <- min(mtcars$mpg)
+  mpg_max <- max(mtcars$mpg)
+
+  # Density should be around twice higher at finite boundary
+  expect_snapshot(make_density(c(-Inf, Inf)))
+  expect_snapshot(make_density(c(mpg_min, Inf)))
+  expect_snapshot(make_density(c(-Inf, mpg_max)))
+  expect_snapshot(make_density(c(mpg_min, mpg_max)))
```
clauswilke, line 31: Is there a reason that you're using snapshots here? Couldn't you just test numerically that density is approximately twice higher at finite boundary? It doesn't have to be that accurate. You could just test that it's > 1.9x and < 2.1x, for example.

echasnovski: Snapshots seem to provide more full testing of the feature. This is not only about being twice as higher at finite boundary, but about a specific way of dealing with boundary correction (via density reflection). I can make snapshots smaller by setting smaller `n`.

All in all, I'd keep them. Will do as you ask, of course. So should I not use snapshots?

echasnovski:
> Regarding weights, did you check whether the computation was correct? I see no reason that it shouldn't. The warning occurs because I missed to add the following line in my recent PR, could you add it to stat_density()?

Yep, you are right. `weight` is used and the warning is gone after adding `dropped_aes = "weight"`. Should I keep it in this PR?

tests/testthat/test-stat-density.R
clauswilke: I don't like snapshots because if they break at some point in the future nobody will know what the correct result should have been. If you want to test the full reflection, then the right way to do this is to calculate the reflection again in the test and compare numerically. Yes, it may seem silly to run the same calculation in two locations and compare they're the same, but it actually is a good check for the correctness of the implementation in the main library.

clauswilke:
Yes, please keep `dropped_aes = "weight"`. I should have added it in my recent PR and I forgot. It makes sense to add it to yours, in particular if you add a test for weighted density estimations (but even if you don't, no strong opinion on my end).

tests/testthat/test-stat-density.R
echasnovski: Updated test to not use snapshots.

clauswilke:
This looks good to me now. Only one last thing: Please add a news item to the top of: https://github.com/tidyverse/ggplot2/blob/main/NEWS.md
You can model it after the other ones there. We generally want to link the issue or PR that contributed the new feature and the author of the feature.

clauswilke (approved):

thomasp85 (approved):
LGTM

echasnovski:
@clauswilke, @thomasp85, thank *you*. The pleasure is mine to finally give back to *the* R package for visualization.

PR #4030, opened 2020-05-26
Add description of undocumented parameters

### LINKED ISSUE #3946: Add undocumented parameter "bins=" to geom_density_2d docs
The parameter "bins" in

 ``` geom_density_2d(aes(x,y)  , bins=6)```

works as expected, but is not documented anywhere.

johan-ejstrud:
Close #3946

I have borrowed the parameter description from [geom-contour](https://github.com/tidyverse/ggplot2/blob/2b03e4723f0b91875667ffb4f952ad6e471ac9f8/R/geom-contour.r#L16).

(This is my first pull request for ggplot2, and I'm not familiar with roxygen2, so apologies if I have missed something obvious.)

### DIFF
--- a/R/stat-density-2d.r
+++ b/R/stat-density-2d.r
@@ -5,6 +5,7 @@
 #' @param contour_var Character string identifying the variable to contour
 #'   by. Can be one of `"density"`, `"ndensity"`, or `"count"`. See the section
 #'   on computed variables for details.
+#' @inheritDotParams geom_contour bins binwidth breaks
 #' @param n Number of grid points in each direction.
 #' @param h Bandwidth (vector of length two). If `NULL`, estimated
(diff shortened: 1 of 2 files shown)


yutannihilation:
Thanks, but you cannot add documents here because there's no such arguments actually. So, you need to use `@inheritDotParams` to include them in dots.

https://blog.rstudio.com/2017/02/01/roxygen2-6-0-0/

johan-ejstrud:
Thank @yutannihilation! I have updated to use `@inheritDotParams` now.

The CI failing confused me quite a bit, but I can see there's another PR open to fix that.

yutannihilation:
Thanks! Please ignore the failures as they are not related to this change.

PR #4032, opened 2020-05-27
Fix labeller to pass .default for lookups (#4031)

### LINKED ISSUE #4031: facet labellers not calling default
I have a plot using facet_grid, that I want to set the facet labels with.
The faceting variables are stored as integers, and I am trying to use lookup tables to specify some expressions, and then pass those to label_parsed.

As I read the docs for labeller(), the lookup table should be passed to .default.

`.default  Default labeller for variables not specified. Also used with lookup tables or non-labeller functions.`

However, this doesn't seem to be happening (note labels on rows.)

I'm not sure if this is a documentation problem, me misunderstanding something, or a problem with the labeller.

---

``` r
library(ggplot2)
tdf <- data.frame(x=c(1,2,3,4), y=c(4,3,2,1), a=c(0,1,0,1), b=c(1, 2, 2, 1))
ggplot(tdf, aes(x, y))+geom_point()+
  facet_grid(a~b, labeller=labeller(a=c(`0`="a^2", `1`="a[2]"), b=c(`1`="One", `2`="Two"), .default=label_parsed))
```

<sup>Created on 2020-05-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

---

If I instead use `.rows` and `.cols` to name the arguments to labeller, I get the expected result.
(The labels on the rows are now passed to label_parsed)

``` r

library(ggplot2)
tdf <- data.frame(x=c(1,2,3,4), y=c(4,3,2,1), a=c(0,1,0,1), b=c(1, 2, 2, 1))
ggplot(tdf, aes(x, y))+geom_point()+
  facet_grid(a~b, labeller=labeller(.rows=c(`0`="a^2", `1`="a[2]"), .cols=c(`1`="One", `2`="Two"), .default=label_parsed))
```

<sup>Created on 2020-05-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

waltersom:
Pass .default to as_labeller even when margin_labeller is null, such as
when using named arguments for a lookup table.

This should make the behaviour match the documentation.

Fixes #4031


clauswilke:
While we're touching this code, I think we should also consider whether `.multi_line` should be forwarded to `as_labeller` as well.

waltersom:
@karawoo I have made the requested changes.

Also @clauswilke, I was looking at also passing `.multi_line`, however this was more complicated, and more work is required for that. A problem is that if `.multi_line==FALSE`, when using `label_parsed`, `collapse_label_lines` is eventually called on the output, which converts any `expression()`s to characters. Further, `label_parsed` adds `list()` around the argument to `expression()`, which I don't fully understand.

karawoo:
Hmm, I see what you mean @waltersom. The `.multi_line` issue is more complicated. Using the example from the original issue with current ggplot2, `.multi_line = FALSE` gives the plot below, and this PR doesn't change that:

``` r
library("ggplot2")

tdf <- data.frame(
  x = c(1, 2, 3, 4),
  y = c(4, 3, 2, 1),
  a = c(0, 1, 0, 1),
  b = c(1, 2, 2, 1)
)

ggplot(tdf, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ a + b,
    labeller = labeller(
      a = c(`0` = "a^2", `1` = "a[2]"),
      b = c(`1` = "One", `2` = "Two"),
      .default = label_parsed,
      .multi_line = FALSE
    )
  )
```



If we just passed along `.multi_line` to the line you've edited, it becomes:

``` r
ggplot(tdf, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ a + b,
    labeller = labeller(
      a = c(`0` = "a^2", `1` = "a[2]"),
      b = c(`1` = "One", `2` = "Two"),
      .default = label_parsed,
      .multi_line = FALSE
    )
  )
```




I'd be inclined to merge this and open a separate issue for the `.multi_line` problem, but what do you think @clauswilke?

clauswilke:
This seems fine by me. The only reservation I have is that I don't know where we are in the current release cycle. Is it fine to merge into master or should we wait?

karawoo:
Yeah, let's wait until this release is finished

karawoo:
(CI failures are unrelated)

yutannihilation:
(For failures, macOS runner will keep failing for a while until the R 4.0.2 installer is reliably available.)

Issue #3893, opened 2020-03-14
coord_polar() documentation: make clear 'start' depends on 'direction'?


david-romano:
The current documentation simply states that `start` is measured in radians from 12 o'clock, but it doesn't say that 'start' is measured relative to the direction chosen -- could that be added in the documentation.

thomasp85:
Sure... would you be interested in submitting a PR?

brianrice2:
I submitted a pull request since this was simple and inactive for a while, and agree seemed good for a first issue. Feedback welcome. Thanks.

PR #4039, opened 2020-06-01
Add details to start parameter of coord_polar (#3893).

brianrice2:
Motivation: current documentation for coord_polar() does not mention that 'start' offset can be applied either clockwise or anticlockwise. Added note in docs that the offset is applied depending on orientation specified by 'direction' parameter (1 = clockwise, -1 = anticlockwise).

### DIFF
--- a/R/coord-polar.r
+++ b/R/coord-polar.r
@@ -4,7 +4,8 @@
 #' are a stacked bar chart in polar coordinates.
 #'
 #' @param theta variable to map angle to (`x` or `y`)
-#' @param start offset of starting point from 12 o'clock in radians
+#' @param start offset of starting point from 12 o'clock in radians. Offset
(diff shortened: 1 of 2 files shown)


R/coord-polar.r
```diff
@@ -4,7 +4,8 @@
 #' are a stacked bar chart in polar coordinates.
 #'
 #' @param theta variable to map angle to (`x` or `y`)
-#' @param start offset of starting point from 12 o'clock in radians
+#' @param start offset of starting point from 12 o'clock in radians. Offset
```
yutannihilation, line 7:
```suggestion
#' @param start Offset of starting point from 12 o'clock in radians. Offset
```

brianrice2: Agree. good catch!

yutannihilation:
Thanks!

PR #4042, opened 2020-06-02
Update NEWS to include discrete color scale options from #3833

cpsievert:
This should have been done as a part of #3833 (sorry!)


yutannihilation:
Thanks!

PR #4046, opened 2020-06-06
Fix CI failures

### LINKED ISSUE #4045: CI tests failing
Since the last two commits, various integration tests are failing. However, it's somewhat random, with (for example) on Travis three succeeding and four failing. 

Screen Shot 2020-06-05 at 3 34 39 PM

The most common error is the following:
```
* creating vignettes ... ERROR
Error: processing vignette 'extending-ggplot2.Rmd' failed with diagnostics:
'vignetteInfo' is not an exported object from 'namespace:tools'
Execution halted
```

Not sure if any of the two most recent commits have actually messed anything up. I don't quite see how.

@yutannihilation Any ideas?

yutannihilation:
Fix #4045 

We need two fixes here:

* For the macOS (devel) runner, bump cache version to clear the cache
* For the Linux (<3.6) runner, install the dev version of rmarkdown package. This will be unnecessary after the next version of rmarkdown is released on CRAN.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -77,6 +77,11 @@ jobs:
           sysreqs=$(Rscript -e "cat(sysreqs::sysreq_commands('DESCRIPTION'))")
           sudo -s eval "$sysreqs"
 
+      # Remove this when the new rmarkdown is released on CRAN
+      - name: "Workaround for rstudio/rmarkdown#1831"
+        if: runner.os == 'Linux' && runner.r < '3.6'
+        run: Rscript -e "remotes::install_github('rstudio/rmarkdown')"
+
       - name: Install system dependencies on macOS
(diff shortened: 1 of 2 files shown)


yutannihilation:
Hmm, is something wrong with R 4.0.1 installer?

yutannihilation:
Ah, it seems this is the cause. Let's wait for a while.

https://github.com/r-lib/actions/issues/116#issuecomment-640129789

yutannihilation:
Thanks!

> Are we planning to keep travis and appveyor now that we're using github actions too?

At the time of enabling GitHub Actions CI (https://github.com/tidyverse/ggplot2/pull/3862#issuecomment-595819282), the things were very unstable so I felt it was better to keep Travis and AppVayor until we're sure GitHub Actions CI is reliable enough. But, it might be time to remove them now. I'll create another PR for this. Thanks for catching.

PR #4059, opened 2020-06-12
Use parentheses

krlmlr:
when referring to a function in documentation.

Only looked at `` at the start of a line and at surroundings. I can hunt for other instances if you think it's useful.

```sh
sed -r -i "" 's/(^#'"'"' `[a-z_]+[a-z_0-9]*)`/\1()`/' R/*
```


clauswilke:
Looks good to me. If you can find other instances that'd be great, otherwise we can just merge this one.

krlmlr:
Thanks. If you merge this I can do a second PR.

PR #4065, opened 2020-06-16
Support ptype argument for scale() functions

krlmlr:
so that vectors can apply their own formatting to a scale. See the labels on the x axis in the example.

Applying the vector class "after the fact" on numeric breaks works for this use case. Am I missing a better solution to the problem?

I've also tried maintaining the vector class so that the `breaks` are of the same type as the data, that seems very difficult. Perhaps the `ptype` should be passed to the `labels` function by default instead?

``` r
library(tidyverse)

scale_x_formattable <- function(..., ptype) {
  labels <- function(x) {
    out <- ptype[rep(NA_integer_, length(x))]
    out[] <- x
    out
  }

  ggplot2::scale_x_continuous(..., labels = labels)
}

scale_type.formattable <- function(x) "formattable"

x <- formattable::percent((1:10) / 100)

data.frame(x = x, y = 1:10) %>%
  ggplot(aes(x = x, y = y)) %>%
  + geom_point()
```

<sup>Created on 2020-06-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


krlmlr:
Solved in https://github.com/r-lib/pillar/pull/404.

PR #4069, opened 2020-06-18
Remove a workaround about rmarkdown on R <3.6

yutannihilation:
The new version of rmarkdown was released on CRAN and it now contains the fix for `tools::vignetteInfo()` errors.  The workaround is no longer needed.


yutannihilation:
Thanks. Yeah, I don't think the version of the rmarkdown matters much at the moment as there are few differences between the CRAN version and the GitHub one. Let's leave the version as it is this time.

Issue #4072, opened 2020-06-20
undefined columns selected with groupS aes in custom geom


tdhock:
Hi again. I'm the maintainer of directlabels::geom_dl and I'm trying to figure out a way to let the user optionally specify a grouping variable to use, and to detect when the user specifies that in my draw_panel method. Originally I was trying to use aes(group) for this purpose but that does not work, see #4070.

Now I am considering using another aes for this purpose, like aes(groupS), but I'm observing the following error:
```
Error in `[.data.frame`(data, "group") : undefined columns selected
```
I expected that I could use any custom aes, but this one is not working.

This happens because of a bug in grouping.r, which currently tests for absence of group column via
```
  if (is.null(data$group)) {
```
however that test will not work if there is another column/aes (like groupS) that starts with group. The fix is to change it to 
```
  if (is.null(data[["group"]])) {
```
Here is a reprex using a standard ggplot2 geom 
```
library(ggplot2)
ggplot()+
  geom_point(aes(
    Petal.Length, Petal.Width, groupS=1),
    data=iris)
#> Warning: Ignoring unknown aesthetics: groupS
#> Error in `[.data.frame`(data, "group"): undefined columns selected
```

yutannihilation:
Hmm, partial matching bites... Thanks, good catch.

Issue #3725, opened 2020-01-13
after_scale fails to find some aesthetic variables


atusy:
I am highlighting some points by a combination of `stage()` and `after_scale()`.
I found that `after_scale` can find aesthetic variables, but not variables in the original data.
It is fine, if that's a design.
As a workaround, I let `after_scale` find the variable in the data by mapping it to an aesthetic variable. 
It works in some case, but not always.
As far as I tried, it seems the error is related to discrete variables.

``` r
library(ggplot2)
g <- ggplot(mtcars) +
    aes(wt, mpg, highlight = am == 1) +
    geom_point(
        aes(colour = stage(cyl, after_scale = ifelse(highlight, colour, 'gray'))),
        shape = 21
    )

# Okay
g
```

![](https://i.imgur.com/v7wlkZO.png)

``` r

# Okay
g + aes(fill = am)
```

![](https://i.imgur.com/P78dORo.png)

``` r

# Error
g + aes(fill = as.factor(am))
#> Error in ifelse(highlight, colour, "gray"): object 'highlight' not found

# Error
ggplot(iris) +
    aes(Sepal.Length, Sepal.Width, highlight = Species == "setosa") +
    geom_point(
        aes(
            colour = stage(
                Species, after_scale = ifelse(highlight, colour, 'gray')
            )
        )
    )
#> Error in ifelse(highlight, colour, "gray"): object 'highlight' not found
```

<sup>Created on 2020-01-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

yutannihilation:
The error seems to occur on drawing legends. I don't get what's the expected behaviour here...

``` r
library(ggplot2)

g <- ggplot(mtcars) +
  aes(wt, mpg, highlight = am == 1) +
  geom_point(
    aes(colour = stage(cyl, after_scale = ifelse(highlight, colour, 'gray'))),
    shape = 21
  )

g + aes(fill = as.factor(am)) + theme(legend.position = "none")
```

![](https://i.imgur.com/uHr3v75.png)

<sup>Created on 2020-01-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Defining a random aesthetic (`highlight`) will invariantly lead to undefined behaviour. There is no way to know how the stat and geom will treat it or carry it over. This has always been true and was also the case for the old `stat()` and `..var..` notations

[...]

PR #4074, opened 2020-06-20
Fix aes(groupS)

tdhock:
hi this is my proposed fix for #4072, along with a test that fails using current master.

### DIFF
--- a/tests/testthat/test-aes-grouping.r
+++ b/tests/testthat/test-aes-grouping.r
@@ -24,10 +24,10 @@ test_that("one group per combination of discrete vars", {
 })
 
 test_that("no error for aes(groupS)", {
-  df <- data_frame(x = df$a, y = df$b, groupS = 1)
-  g <- add_group(df)
+  df2 <- data_frame(x = df$a, y = df$b, groupS = 1)
+  g <- add_group(df2)
 
-  expect_equal(nrow(df), 4L)
+  expect_equal(nrow(g), nrow(df2))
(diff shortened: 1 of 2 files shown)


tests/testthat/test-aes-grouping.r
```diff
@@ -23,6 +23,12 @@ test_that("one group per combination of discrete vars", {
   expect_equal(groups(plot), 4)
 })
 
+test_that("no error for aes(groupS)", {
+  plot <- ggplot(df, aes(a, b, groupS=1)) + geom_point()
+  g <- ggplotGrob(plot)
+  expect_is(g, "grob")
```
yutannihilation, line 29: Could you test `add_group()` directly with `expect_error(..., NULL)`? Using an unknown aes is not a guaranteed usage (c.f. https://github.com/tidyverse/ggplot2/issues/3725#issuecomment-573716059), so you need to create a custom Geom that actually requires `groupS`, but I think it's a bit too heavy to test this issue.

yutannihilation: If you need some help, I'd be happy to provide the test code.

yutannihilation:
@tdhock Could you fix the test case?

tdhock:
sorry I have lots of other deadlines coming up right now so I won't be able
to fix right away.

yutannihilation:
I see. Sorry to hurry you, maybe I can take care of the test.

tdhock:
great

PR #4076, opened 2020-06-20
Update utilities-tidy-eval.R and use as_label()

### LINKED ISSUE #4063: Use as_label() instead of quo_name()
https://github.com/tidyverse/ggplot2/blob/111258238fb0932d390a4fb56a1e49026376f4e2/R/facet-.r#L216-L217

yutannihilation:
Close #4063 

What I did to create this PR was:

* update `utils-tidy-eval.R` by `usethis::use_tidy_eval()`
* add `@export quo_name` for backward-compatibility, considering we've been exporting it for 2 years with a documentation using it
* replace `quo_name()` with `as_label()` in the code and the docs

As a result, this PR changes the following two points:

* adds some new re-exported functions `:=`, `as_name()`, and `as_label()`.
* removes some re-exported functions `quo()`, `quos()`, `enexpr()`, `enexprs()`, `ensym()`, `ensyms()`; they are different than `quo_name()` in that no documentations of ggplot2 refer to them, so I believe it's OK to remove them.


yutannihilation:
@lionel- 
Sorry to bother you. Could you take a look when you have time? I believe this PR is safe to merge, but I want to confirm if this is consistent with the latest direction of rlang things.

yutannihilation:
Thanks! OK, let's keep other functions as well.

PR #4079, opened 2020-06-21
Deprecate qplot()

### LINKED ISSUE #3956: Remove stat and position arguments from qplot() and formaly deprecate the function
`stat` and `position` were deprecated 4.5 years ago ([v2.0.0](https://ggplot2.tidyverse.org/news/index.html#deprecated-features)), so I think it's time to remove them. I think we have no development bandwidth to fix such tricky bugs like https://github.com/tidyverse/ggplot2/issues/3670.

Also, can we agree on deprecating `qplot()` itself? Though we've never discussed this formally, I believe the use of `qplot()` is discouraged, which virtually means it's deprecated.

yutannihilation:
Close #3956 

### TODO

- [x] Add `.Deprecate()` to `qplot()`
- [x] Remove `stat` and `position` and throw errors
- [x] Tweak the document of `qplot()`
- [x] Remove the use of `qplot()` in tests

### Remaining quesitons

- [x] ~~How should we rewrite `translate_qplot_ggplot` and `translate_qplot_lattice`?~~ (I decided to just delete them)
- [ ] Do we need some instruction to get rid of `qplot()`?
- [ ] How should we communicate with the extension package authors?


thomasp85:
Team - should we pull the trigger on this? I don't think any more coercion without deprecation will move remaining users so there is no point in waiting longer. It is either deprecation of `qplot()` or leaving it in in perpetuity...

yutannihilation:
> throw a warning when somebody uses `qplot()`

This pull request was intended for this.

clauswilke:
Then I'm in support.

thomasp85:
Yeah, this will have to go through the proper and very very long deprecation process... but it will allow us to better ignore the function during development

thomasp85 (approved):
This looks good - had no idea we were using plot so much in the tests 😬

yutannihilation:
Yeah, I too didn't expect I would have to modify so many lines... Thanks for reviewing, I'm merging this.

PR #4062, opened 2020-06-14
Test: Fix Rd cross-ref warnings on R-devel

yutannihilation:
(This PR is only for testing if the new document generation can pass the checks on r-devel)

* `devtools::install_github("r-lib/roxygen2")`
* ~~Re-generate the document of scales package and installed it~~ (This has been addressed on roxygen2's side)
* Update the documents of `utilities-tidy-eval.R` to [the latest one](https://github.com/r-lib/usethis/blob/master/inst/templates/tidy-eval.R) (Note that I copied only the paragraphs about the functions that are currently re-exported. Let's update it again after the release of v3.3.2. c.f.  #4063)
* Re-generate document of ggplot2

PR #4091, opened 2020-06-25
Use sf::sf_project() to transform spatial coordinates.

### LINKED ISSUE #4090: Use sf_project() in function sf_transform_xy()
Per @edzer (https://twitter.com/edzerpebesma/status/1276159543152705543?s=20), `sf::sf_project()` should do what we need and should be faster than the method currently implemented in `sf_transform_xy()` (which generates a list of separate points and then transforms them one by one).

clauswilke:
Closes #4090.

### DIFF
--- a/R/coord-sf.R
+++ b/R/coord-sf.R
@@ -470,11 +470,15 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
 #' Transform spatial position data
 #'
 #' Helper function that can transform spatial position data (pairs of x, y
-#' values) among coordinate systems.
+#' values) among coordinate systems. This is implemented as a thin wrapper
+#' around [sf::sf_project()].
 #'
 #' @param data Data frame or list containing numerical columns `x` and `y`.
 #' @param target_crs,source_crs Target and source coordinate reference systems.
 #'   If `NULL` or `NA`, the data is not transformed.
+#' @param authority_compliant logical; `TRUE` means handle axis order authority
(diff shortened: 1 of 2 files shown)


clauswilke:
@yutannihilation What's the situation with roxygen2 now? I processed documentation with the latest roxygen2 from github and got a ton of changes in numerous `.Rd` files. I didn't commit those here, though.

clauswilke:
Also, the check on R 3.2 complains about an unknown package sf in the checking of cross-references. Is this an issue? Do we still support R 3.2?

yutannihilation:
You need to install `devtools::install_github("r-lib/roxygen2#1109")` as 
https://github.com/r-lib/roxygen2/pull/1109#issuecomment-649466170 is not merged yet.
(I mistakenly wrote `devtools::install_github("r-lib/roxygen2")` on #4062, sorry if you got confused by this)

yutannihilation:
> it seems this is the first time to link to a package that is not available on R 3.2

A small correction. multcomp and Hmisc are such packages, but they doesn't seem to be a problem because they *supposed to* exist on R 3.2, even though they cannot be installed. In the case of sf, as it explicitly requires R >= 3.3, it doesn't exist.

```
❯ checking Rd cross-references ... WARNING
  Packages unavailable to check Rd xrefs: ‘multcomp’, ‘Hmisc’
  Unknown package ‘sf’ in Rd xrefs
```

clauswilke:
The other alternative here would be to just not add a cross reference to the documentation. In any case, I'll merge now, and we can worry about the cross reference later.

PR #4053, opened 2020-06-09
Stop using Travis CI and AppVeyor

yutannihilation:
It has been 3 months since we enabled GitHub Actions CI (#3862). While we still need to take care GitHub Actions CI time to time, it seems to catch the unintended changes reliably just as Travis CI and AppVeyor have been doing for years. So, I think it's okay to stop using Travis CI and AppVeyor now.

PR #4096, opened 2020-06-26
Ignore Rd xref warnings on R 3.2

### LINKED ISSUE #4095: CI failure on R 3.2
After #4091, R 3.2 runner fails with this warning. It seems cross-referencing to packages that are not on the CRAN index (sf is not available on R 3.2 as it requires R >= 3.3) is not allowed.

```
❯ checking Rd cross-references ... WARNING
  Packages unavailable to check Rd xrefs: ‘multcomp’, ‘Hmisc’
  Unknown package ‘sf’ in Rd xrefs
```

We can avoid this by setting `_R_CHECK_RD_XREFS_` to `FALSE`.

https://github.com/wch/r-source/blob/8d7ac4699fba640d030703fa010b66bf26054cbd/src/library/tools/R/check.R#L5988-L5989
https://github.com/wch/r-source/blob/8d7ac4699fba640d030703fa010b66bf26054cbd/src/library/tools/R/check.R#L2187-L2201

c.f. https://cloud.r-project.org/doc/manuals/r-patched/R-ints.html#index-_005fR_005fCHECK_005fRD_005fXREFS_005f

yutannihilation:
Close #4095 

According to [R-internals](https://cloud.r-project.org/doc/manuals/r-patched/R-ints.html#index-_005fR_005fCHECK_005fRD_005fXREFS_005f), we can avoid the warning by setting  `_R_CHECK_RD_XREFS_` to `FALSE`.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -22,15 +22,15 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: '4.0',   vdiffr: true}
-          - {os: macOS-latest,   r: '4.0',   vdiffr: true}
-          - {os: macOS-latest,   r: 'devel', vdiffr: false}
-          - {os: ubuntu-16.04,   r: '4.0',   vdiffr: true,  cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
-          - {os: ubuntu-16.04,   r: '3.6',   vdiffr: false, cran: "https://demo.rstudiopm.com/all/__linux__/xenial/latest"}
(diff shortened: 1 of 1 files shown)


yutannihilation:
Hmm, seems we cannot leave `_R_CHECK_RD_XREFS_` as empty...

```
* checking Rd metadata ... OK
Error in if (dir.exists("man") && R_check_Rd_xrefs) { : 
  missing value where TRUE/FALSE needed
Execution halted
```

yutannihilation:
Thanks.

> Should they be disabled?

Yes. I asked Thomas and Hadley to disable it (assuming they have admin privilege of this repo) on https://github.com/tidyverse/ggplot2/pull/4053#issuecomment-646908967, but it seems my voice doesn't reach them yet...

PR #4098, opened 2020-06-26
Style code used in examples

### LINKED ISSUE #4092: Code style in examples hard to read in reference docs
---
Really minor but the code could be easier to read in some of the examples. 

In the [sec_axis() example](https://ggplot2.tidyverse.org/reference/sec_axis.html#examples
) it looks ok if I copy and paste the code but in the help menu and the ref docs, it's hard to see where the parentheses start and end and the piping of the plot elements

When copied, it's easier to read:
```r
ggplot(df, aes(x = dx, y = price)) + geom_line() +
  scale_x_datetime("Date", date_labels = "%b %d",
                   date_breaks = "6 hour",
                   sec.axis = dup_axis(name = "Time of Day",
                                       labels = scales::time_format("%I %p")))
```
I'd like for it to look like this:
```r
ggplot(df, aes(x = dx, y = price)) +
  geom_line() +
  scale_x_datetime(
    name = "Date",
    date_labels = "%b %d",
    date_breaks = "6 hour",
    sec.axis = dup_axis(
      name = "Time of Day",
      labels = scales::time_format("%I %p")
    )
  )

```

rjake:
Fixes #4092 

This is my first PR for `ggplot2`. This will use the tidyverse style in the example code. I tried to check all docs. It was mentioned that a bigger audit might be done via `styler` at a later date and so I only picked a few types of coding style to change. If the code looked like this
```r
 p + scale_fill_brewer(direction = -1) 
```
I left it be. 
I only moved to new lines when there was more than one pipe present
```r
 p + scale_fill_brewer(direction = -1) + theme_dark()

#becomes 
 p + 
  scale_fill_brewer(direction = -1) + 
  theme_dark()

```
when the arguments weren't indented under the parentheses, or when the line had two functions `ggplot() + geom...`
```r
ggplot(mpg, aes(cty, hwy) + geom_point()

# becomes
ggplot(mpg, aes(cty, hwy) + 
  geom_point()
```
There were some updates that `roxygen2` (dev version) made that I committed at the end
```
\item The default (\code{\link[scales:oob]{scales::censor()}}) replaces out of
\item The default (\code{\link[scales:censor]{scales::censor()}}) replaces out of bounds values with \code{NA}.
```

I also had 2 failed tests that seemed unrelated to my code changes
image

### DIFF
--- a/R/geom-function.R
+++ b/R/geom-function.R
@@ -18,14 +18,20 @@
 #'   geom_function(fun = dnorm, colour = "red")
 #'
 #' # To plot functions without data, specify range of x-axis
-#' base <- ggplot() + xlim(-5, 5)
+#' base <-
+#'   ggplot() +
+#'   xlim(-5, 5)
+#'
 #' base + geom_function(fun = dnorm)
+#'
 #' base + geom_function(fun = dnorm, args = list(mean = 2, sd = .5))
 #'
 #' # The underlying mechanics evaluate the function at discrete points
 #' # and connect the points with lines
 #' base + stat_function(fun = dnorm, geom = "point")
+#'
 #' base + stat_function(fun = dnorm, geom = "point", n = 20)
+#'
 #' base + geom_function(fun = dnorm, n = 20)
 #'
 #' # Two functions on the same plot
@@ -35,11 +41,14 @@
 #'
 #' # Using a custom anonymous function
 #' base + geom_function(fun = function(x) 0.5*exp(-abs(x)))
+#'
 #' base + geom_function(fun = ~ 0.5*exp(-abs(.x)))
 #'
 #' # Using a custom named function
 #' f <- function(x) 0.5*exp(-abs(x))
+#'
 #' base + geom_function(fun = f)
+#'
 #' @export
 geom_function <- function(mapping = NULL, data = NULL, stat = "function",
                           position = "identity", ..., na.rm = FALSE,
--- a/R/geom-text.r
+++ b/R/geom-text.r
@@ -64,12 +64,19 @@
 #' p + geom_text(size = 10)
 #'
 #' # Set aesthetics to fixed value
-#' p + geom_point() + geom_text(hjust = 0, nudge_x = 0.05)
-#' p + geom_point() + geom_text(vjust = 0, nudge_y = 0.5)
-#' p + geom_point() + geom_text(angle = 45)
+#' p +
+#'   geom_point() +
+#'   geom_text(hjust = 0, nudge_x = 0.05)
+#' p +
+#'   geom_point() +
+#'   geom_text(vjust = 0, nudge_y = 0.5)
+#' p +
+#'   geom_point() +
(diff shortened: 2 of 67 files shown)


clauswilke:
The sf tests don't fail in the continuous integration runs. If tests fail for you locally you may have to update sf.

yutannihilation:
Thanks for the contribution!

> There were some updates that `roxygen2` (dev version) made that I committed at the end

You probably are using the older version of scales. These Rd files are now `oob` as of version 1.1.1, more precisely, since this commit:

https://github.com/r-lib/scales/commit/e800430d1b61db3941ee48572d0fae4db685bd61#diff-cf40208acff59ab6d96fa89d3c1f9a39

yutannihilation:
@rjake Could you update scales package and regenerate the docs?

R/geom-function.R
```diff
@@ -18,14 +18,20 @@
 #'   geom_function(fun = dnorm, colour = "red")
 #'
 #' # To plot functions without data, specify range of x-axis
-#' base <- ggplot() + xlim(-5, 5)
+#' base <-
+#'   ggplot() +
+#'   xlim(-5, 5)
```
yutannihilation, line 23: One question is a like break after `<-` should be added or not, but I think either is fine.

c.f. https://style.tidyverse.org/pipes.html#assignment-2

R/geom-text.r
```diff
@@ -80,18 +87,26 @@
...
+#'     aes(label = paste(wt, "^(", cyl, ")", sep = "")),
+#'     parse = TRUE
+#'   )
 #'
 #' # Add a text annotation
 #' p +
 #'   geom_text() +
-#'   annotate("text", label = "plot mpg vs. wt", x = 2, y = 15, size = 8, colour = "red")
+#'   annotate(
+#'     "text", label = "plot mpg vs. wt",
```
yutannihilation, line 107: This `"text",` can be placed in the same line of `annotate(`, but this is also fine.

> Short unnamed arguments can also go on the same line as the function name, even if the whole function call spans multiple lines.
https://style.tidyverse.org/syntax.html#long-lines

yutannihilation:
Thanks!

R/geom-text.r
rjake: I've noticed that Ctrl + Enter sometimes errors out at the bottom if the pattern is mismatched about code touching the parentheses or not

```r
# fine
annotate("text", 
         label = "plot mpg vs. wt")

# fine
annotate(
  "text",
  label = "plot mpg vs. wt"
)

# not fine
annotate("text",
  label = "plot mpg vs. wt"
)
```

Here's an example
```r
library(tidyverse)

mpg %>% 
  mutate(
    x1 = "ok here") %>% 
  mutate(
    x2 = "not ok") # use Ctrl + Enter in this line
````

image

yutannihilation: Hmm, thanks for the information. I hope this will be fixed on RStudio...

PR #4109, opened 2020-07-01
Deprecate guides(... = FALSE)

### LINKED ISSUE #4097: Can we deprecate guides(... = FALSE)?
I happened to know `guides()` accepts `FALSE` as an option not to show the legend. e.g https://stackoverflow.com/a/14604540

``` r
p + guides(colour = FALSE)
```

I think none of the official documentation refers to this usage, and no test is broken if I remove this `isFALSE(guide)`, so I think this is safe to remove. At least we should start encouraging people to use `"none"` as it is the supposed way. The only concern is that I cannot understand what this comment exactly means.... This has been here for 9 years since the point where the S3 guide system was born (https://github.com/tidyverse/ggplot2/commit/0795c6f1e3f3d4d3b9afc0aaea176d3e7a24a7f0)...

https://github.com/tidyverse/ggplot2/blob/b4e3aeae772bea75298734a89d28c1efdb24da82/R/guides-.r#L187-L190

yutannihilation:
Fix #4097

`guides(... = FALSE)` was a very old syntax and is not mentioned in any documentations. This should be deprecated to inform the users about the recommended way (`guides(... = "none")`).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,9 @@
   
 * Updated style for example code (@rjake, #4092)
 
+* Using `guides(<aes> = FALSE)` to suppress legends is now deprecated. Please
+  use `guides(<aes> = "none")` instead (@yutannihilation, #4094).
+
 # ggplot2 3.3.2
 This is a small release focusing on fixing regressions introduced in 3.3.1.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,8 +7,10 @@
 
 * Only drop groups in `stat_ydensity()` when there are fewer than two data points and throw a warning (@andrewwbutler, #4111).
 
-* Using `guides(<scale> = FALSE)` to remove a guide is now deprecated. Please
-  use `guides(<scale> = "none")` instead (@yutannihilation, #4094).
+* It is not deprecated to specify `guides(<scale> = FALSE)` or
+  `scale_*(guide = FALSE)` to remove a guide. Please use 
+  `guides(<scale> = "none")` or `scale_*(guide = "none")` instead 
+  (@yutannihilation, #4094).
 
 # ggplot2 3.3.2
 This is a small release focusing on fixing regressions introduced in 3.3.1.
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,8 +5,8 @@
   
 * Updated style for example code (@rjake, #4092)
 
-* Using `guides(<aes> = FALSE)` to suppress legends is now deprecated. Please
-  use `guides(<aes> = "none")` instead (@yutannihilation, #4094).
+* Using `guides(<scale> = FALSE)` to remove a guide is now deprecated. Please
+  use `guides(<scale> = "none")` instead (@yutannihilation, #4094).
 
 # ggplot2 3.3.2
 This is a small release focusing on fixing regressions introduced in 3.3.1.
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 
 * Only drop groups in `stat_ydensity()` when there are fewer than two data points and throw a warning (@andrewwbutler, #4111).
 
-* It is not deprecated to specify `guides(<scale> = FALSE)` or
+* It is now deprecated to specify `guides(<scale> = FALSE)` or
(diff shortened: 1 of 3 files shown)


yutannihilation:
@clauswilke 
Could you review this when you have time, as you are familiar with `guides()` well? I believe this change is safe, but one thing I'm not sure is how often `guides(<scale> = FALSE)` is actually used. Have you ever seen this usage before? If you feel this is so common a usage and isn't worth changing, I might withdraw this PR.

R/guides-.r
```diff
@@ -190,10 +197,12 @@ guides_train <- function(scales, theme, guides, labels) {
...
       guide <- resolve_guide(output, scale, guides)
 
-      # this should be changed to testing guide == "none"
-      # scale$legend is backward compatibility
-      # if guides(XXX=FALSE), then scale_ZZZ(guides=XXX) is discarded.
-      if (identical(guide, "none") || isFALSE(guide) || inherits(guide, "guide_none")) next
+      if (identical(guide, "none") || inherits(guide, "guide_none")) next
+
+      if (isFALSE(guide)) {
+        warn('It is deprecated to specify `FALSE` to remove a guide. Please use `"none"` instead.')
```
clauswilke, line 203: Is there a realistic usage pattern by which we would reach this branch? What happens when somebody adds `scale_*(guide = FALSE)`? Is it caught here or in the other check?

R/guides-.r
```diff
@@ -69,6 +69,13 @@ guides <- function(...) {
     if (is.list(args[[1]]) && !inherits(args[[1]], "guide")) args <- args[[1]]
     args <- rename_aes(args)
   }
+
+  idx_false <- vapply(args, isFALSE, FUN.VALUE = logical(1L))
+  if (any(idx_false)) {
```
clauswilke, line 74: `any()` can return `NA`. Better write `isTRUE(any(...)`.

yutannihilation: Thanks.

R/guides-.r
yutannihilation: > `scale_*(guide = FALSE)`

Ah, yes. It reaches here. Maybe I should warn it on `(discrete|continuous|binned)_scale()`? What I was in mind is that there might be some extension package that creates guide object without `guides()`. Not sure if there's any.

clauswilke: Why not just write "It is deprecated to specify `guide = FALSE` to remove a guide. Please use `guide = "none"` instead."?

tests/testthat/test-guides.R
```diff
@@ -524,3 +524,18 @@ test_that("coloursteps guide can be styled correctly", {
...
+
+test_that("a warning is generated when guides(<scale> = FALSE) is specified", {
+  df <- data_frame(x = c(1, 2, 4),
+                   y = c(6, 5, 7))
+
+  expect_warning(g <- guides(colour = FALSE), "`guides(<scale> = FALSE)` is deprecated.", fixed = TRUE)
+  expect_equal(g[["colour"]], "none")
+
+  # manually create a invalid guide
+  g[["colour"]] <- FALSE
```
clauswilke, line 536: I would change the test so it uses an existing scale function. The current version seems a bit contrived.

R/guides-.r
yutannihilation: I see what you mean. Makes sense, thanks.

tests/testthat/test-guides.R
yutannihilation: Sure.

NEWS.md
```diff
@@ -7,8 +7,10 @@
 
 * Only drop groups in `stat_ydensity()` when there are fewer than two data points and throw a warning (@andrewwbutler, #4111).
 
-* Using `guides(<scale> = FALSE)` to remove a guide is now deprecated. Please
-  use `guides(<scale> = "none")` instead (@yutannihilation, #4094).
+* It is not deprecated to specify `guides(<scale> = FALSE)` or
```
clauswilke, line 10: I think you meant "now" not "not".

yutannihilation: Yes. Thanks for catching.

yutannihilation:
Thanks for reviewing!

PR #4115, opened 2020-07-03
Doesn't try to combine labels if labels are NULL

eliocamp:
I encountered this strange error. It took me a while to get a reprex because it is incredibly elusive. 

``` r
library(ggplot2)

grid <- expand.grid(x = seq(0, 360, by = 2.5), y = seq(-90, -20, by = 2.5))
grid$z <- with(grid, x^2*y)

ggplot(grid, aes(x, y)) +
  geom_contour(aes(z = z)) +
  coord_polar() +
  scale_x_continuous(breaks = seq(0, 360, by = 30), labels = NULL)
#> Error in labels[[n]] <- combined: attempt to select less than one element in integerOneIndex
```

The issue was that `coord_polar()` tries to combine labels even if they are `NULL`. I added a check. All tests succeed and the bug is fixed.

### DIFF
--- a/R/coord-polar.r
+++ b/R/coord-polar.r
@@ -281,7 +281,7 @@ CoordPolar <- ggproto("CoordPolar", Coord,
     # Combine the two ends of the scale if they are close
     theta <- theta[!is.na(theta)]
     ends_apart <- (theta[length(theta)] - theta[1]) %% (2*pi)
-    if (length(theta) > 0 && ends_apart < 0.05) {
+    if (length(theta) > 0 && ends_apart < 0.05 & !is.null(labels)) {
       n <- length(labels)
       if (is.expression(labels)) {
         combined <- substitute(paste(a, "/", b),
--- a/R/coord-polar.r
+++ b/R/coord-polar.r
@@ -281,7 +281,7 @@ CoordPolar <- ggproto("CoordPolar", Coord,
     # Combine the two ends of the scale if they are close
     theta <- theta[!is.na(theta)]
     ends_apart <- (theta[length(theta)] - theta[1]) %% (2*pi)
-    if (length(theta) > 0 && ends_apart < 0.05 & !is.null(labels)) {
+    if (length(theta) > 0 && ends_apart < 0.05 && !is.null(labels)) {
       n <- length(labels)
       if (is.expression(labels)) {
         combined <- substitute(paste(a, "/", b),


R/coord-polar.r
```diff
@@ -281,7 +281,7 @@ CoordPolar <- ggproto("CoordPolar", Coord,
     # Combine the two ends of the scale if they are close
     theta <- theta[!is.na(theta)]
     ends_apart <- (theta[length(theta)] - theta[1]) %% (2*pi)
-    if (length(theta) > 0 && ends_apart < 0.05) {
+    if (length(theta) > 0 && ends_apart < 0.05 & !is.null(labels)) {
```
clauswilke, line 284: I think you want to use `&&` here, not `&`.

eliocamp: Ups. Fixed. 
Thanks.

clauswilke:
A test seems to be failing on R devel. Could you see if you can figure out why?

@yutannihilation Is there a reason we check R devel only on macOS? Makes it difficult to figure out whether a problem is related to R devel or to macOS. I think we should check at least on ubuntu also.

eliocamp:
Some visual test unrelated to this change are failing. Might have to do with fonts or some other platform-specific reason?

```
══ testthat results  ═══════════════════════════════════════════════════════════
[ OK: 1515 | SKIPPED: 19 | WARNINGS: 1 | FAILED: 58 ]
1. Error: aesthetics are drawn correctly (@test-aes.r#160) 
2. Error: alpha is drawn correctly (@test-aes.r#175) 
3. Error: segment annotations transform with scales (@test-annotate.r#30) 
4. Error: cartesian coords draws correctly with limits (@test-coord-cartesian.R#25) 
5. Error: cartesian coords draws correctly with clipping on or off (@test-coord-cartesian.R#56) 
6. Error: USA state map drawn (@test-coord-map.R#7) 
7. Error: coord_map scale position can be switched (@test-coord-map.R#16) 
8. Error: polar coordinates draw correctly (@test-coord-polar.r#96) 
9. Error: basic coord_trans() plot displays both continuous and discrete axes (@test-coord-transform.R#99) 
1. ...
```

clauswilke:
Looks like the failing tests are vdiffr tests. Maybe some graphics related detail has changed in R devel?

yutannihilation:
Because R devel is not available on Linux.

c.f. https://github.com/r-lib/actions/issues/49#issuecomment-593455162

yutannihilation:
Let me handle the CI issue on #4116. I'm sure the failure is unrelated to this PR.

clauswilke:
Thanks!

Issue #4118, opened 2020-07-04
Option to turn strip clipping off


teunbrand:
Hi everyone,

I recently found the need to turn the clipping off, but I can't easily do that without either (1) writing a new facet function that builds strips as I want or (2) edit the strips in the gtable.

I found a similar question [here on so](https://stackoverflow.com/questions/49740215/ggplot-facet-grid-label-cut-off/49744051#49744051).
A while back, there was a move to make clipping more configurable (#2536), but this has not been extended to strips. Also there was an issue with strip clipping (#2772), which turned it on for all strips; however my intention would be to make this configurable from e.g. the theme.

One reason besides turning off label clipping is to make the strips flush with either axis lines, panel backgrounds and panel borders (the latter two when the coord clipping is turned off).

An example of the behaviour I love to see configurable is below. The intersection between the strip and the axis line is not ideal, and the `size` parameter appears halved for the strip.

``` r
library(ggplot2)

ggplot(mpg, aes(displ, hwy)) +
  geom_point() +
  facet_wrap(~ drv) +
  theme(
    strip.background = element_rect(colour = "red", size = 5),
    axis.line.y.left = element_line(colour = "blue", size = 5)
  )
```

![](https://i.imgur.com/96k25Fk.png)

<sup>Created on 2020-07-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Thanks for considering!

teunbrand:
Is this idea fundamentally flawed or would it be OK if I were to draft a PR on this subject and discuss from there?

clauswilke:
I don't think it's fundamentally flawed. Could you put together a PR so we can see what would be required?

teunbrand:
Sure thing!

stemangiola:
Thanks for this feature request. Is there an ETA for this feature to be merged?

PR #4125, opened 2020-07-09
Strip clipping optional

teunbrand:
This PR follows from issue #4118.

I think the changes proposed here are the minimal ones required to go make strip clipping an option from the theme setting.

Briefly, `strip.clip` becomes a defined, logical theme-element in the element tree. Also it has gained a corresponding argument in `theme()`, and has defaults in 'root' themes (theme_grey / theme_void / theme_test). In downstream strip rendering, the clipping argument to `assemble_strips()` is taken from the theme instead of immutably `"on"`.

A thing I was unsure about, was wether to use `calc_element()` for consistency, or to take the strip clipping setting directly from the theme, as there isn't anything really anything to inherit from as far as I could tell.

Here is a code example and visual comparison:

``` r
library(ggplot2)
library(patchwork)

# Default clipping is TRUE
p1 <- ggplot(mpg, aes(displ, hwy)) +
  geom_point() +
  facet_wrap(~ drv) +
  theme(
    strip.background = element_rect(colour = "red", size = 5),
    axis.line.y.left = element_line(colour = "blue", size = 5)
  )

p2 <- p1 + theme(strip.clip = FALSE)
(p1 / p2)
```

<sup>Created on 2020-07-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Thanks for considering!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,11 @@
 
 * Only drop groups in `stat_ydensity()` when there are fewer than two data points and throw a warning (@andrewwbutler, #4111).
 
+* It is now deprecated to specify `guides(<scale> = FALSE)` or
+  `scale_*(guide = FALSE)` to remove a guide. Please use 
+  `guides(<scale> = "none")` or `scale_*(guide = "none")` instead 
+  (@yutannihilation, #4094).
+
 # ggplot2 3.3.2
 This is a small release focusing on fixing regressions introduced in 3.3.1.
 
--- a/R/guides-.r
+++ b/R/guides-.r
@@ -69,6 +69,13 @@ guides <- function(...) {
     if (is.list(args[[1]]) && !inherits(args[[1]], "guide")) args <- args[[1]]
     args <- rename_aes(args)
   }
+
+  idx_false <- vapply(args, isFALSE, FUN.VALUE = logical(1L))
+  if (isTRUE(any(idx_false))) {
+    warn('`guides(<scale> = FALSE)` is deprecated. Please use `guides(<scale> = "none")` instead.')
+    args[idx_false] <- "none"
+  }
+
   structure(args, class = "guides")
 }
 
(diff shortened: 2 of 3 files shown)


clauswilke:
Not sure whether this should be a theme option or just an argument to `facet_grid()`. It's an argument to `coord_cartesian()`.

@thomasp85 Any opinion on this?

If we use a theme element, please use `calc_element()` to calculate it. Also, I think the value of the setting should be taken as a string, `"on"`, `"inherit"`, or `"off"`, as is used throughout grid.

teunbrand:
> Not sure whether this should be a theme option or just an argument to facet_grid(). It's an argument to coord_cartesian().

My thinking on this was that defining strip clipping in the theme would automatically transfer the setting to `facet_wrap()` as well as other facet configurations in extension packages that use the `build_strip()` function to render strips. Also it would allow us to define it in a global theme that applies the setting to subsequent plots. 

I agree it would be more consistent to follow `coord_cartesian()` behavior. On the other hand, I feel that clipping the panel is a data-decision that belongs in the coordinate system, whereas strip clipping is more of a 'personal taste'-decision that doesn't directly affect the interpretation of a plot.

> Also, I think the value of the setting should be taken as a string, "on", "inherit", or "off", as is used throughout grid.

The "on" or "off" strings make sense to me, I'm slightly less comfortable with "inherit" (the whole gtable system is effectively invisible for many users). If we would take the theme element route, where would be a natural place to check the validity of the character argument?

clauswilke:
> The "on" or "off" strings make sense to me, I'm slightly less comfortable with "inherit" (the whole gtable system is effectively invisible for many users).

With forthcoming improvements in clipping, we should actually not normally use "off" but rather "inherit" as the default for no clipping. "off" turns all clipping off.

> If we would take the theme element route, where would be a natural place to check the validity of the character argument?

If you think it needs to be checked, check it right before you hand it off to grid.

teunbrand:
> With forthcoming improvements in clipping

Ah yes, excellent point. I'm excited about the expanded capabilities of grid!
I've changed the type from logical to character, defaulting to "inherit"; it indeed seems more appropriate.
If anyone wants to chime in on where this argument should be specified (theme or facet), let me know.

PR #4127, opened 2020-07-10
Handle when `geometry` is not a symbol

### LINKED ISSUE #4122: Bug in geom_sf()?  Unable to convert call to string when processing geometry aesthetic at print.
### Summary of issue

An example similar to `p2` is used in the {ggformula} package but is now breaking.  It appears that when rendering the plot, `as_string(as_name())` fails on the geometry aesthetic when it has the form `data$variable`.  Simply storing the variable as a vector "solves" the problem; see `p1`.

Is this a bug or an intended change in behavior?

### ggplot2 examples

``` r
suppressPackageStartupMessages(library(ggplot2))
suppressPackageStartupMessages(library(sf))
suppressPackageStartupMessages(library(ggformula))
suppressPackageStartupMessages(library(dplyr))
MI <- sf::st_as_sf(maps::map("county", "michigan", plot = FALSE, fill = TRUE))
MIgeom <- MI$geom

# this works
p1 <- 
  ggplot(ggformula::MIpop %>% arrange(county)) + 
  geom_sf(aes(fill = log10(population), geometry = MIgeom), color = "white") 
p1
```

``` r

# this fails
p2 <- 
  ggplot(ggformula::MIpop %>% arrange(county)) + 
  geom_sf(aes(fill = log10(population), geometry = MI$geometry), color = "white") 
p2
#> Error: Can't convert a call to a string
```

### ggformula examples

For completeness, here are the {ggformula} examples that led to this.  I see that these are not quite identical (although the same issues arise).  See the geometry aesthetics displayed near the end.

```{r}
p3 <- gf_sf(
  data = MIpop %>% dplyr::arrange(county),
  fill = ~ log10(population), 
  geometry = ~ MIgeom, color = "white")
p3
```

``` r
p4 <- gf_sf(
  data = MIpop %>% dplyr::arrange(county),
  fill = ~ log10(population), 
  geometry = ~ MI$geom, color = "white")
p4
#> Error: Can't convert a call to a string
```

``` r
p1$mapping
#> Aesthetic mapping: 
#> <empty>
p2$mapping
#> Aesthetic mapping: 
#> <empty>
p3$mapping
#> Aesthetic mapping: 
#> * `fill`     -> `log10(population)`
#> * `geometry` -> `MIgeom`
p4$mapping
#> Aesthetic mapping: 
#> * `fill`     -> `log10(population)`
#> * `geometry` -> `MI$geom`

rlang::as_string(rlang::as_name(p3$mapping$geometry))
#> [1] "MIgeom"
rlang::as_string(rlang::as_name(p4$mapping$geometry))
#> Error: Can't convert a call to a string
```

<sup>Created on 2020-07-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Fix #4122 

Currently, `LayerSf$setup_layer()` uses `as_name(self$mapping$geometry)` in the assumption that `geometry` is a quosure of a symbol, which is not always the case. Possible fix would be

1. actually evaluate it
2. give up guessing the type nicely

and I chose option 2. because option 1. seems too much for this rare use case.

``` r
#> Loading ggplot2

ggplot(nc) +
  geom_sf(aes(fill = AREA, geometry = sf::st_centroid(geometry)))
#> Warning in st_centroid.sfc(geometry): st_centroid does not give correct
#> centroids for longitude/latitude data

#> Warning in st_centroid.sfc(geometry): st_centroid does not give correct
#> centroids for longitude/latitude data
```

<sup>Created on 2020-07-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/layer-sf.R
+++ b/R/layer-sf.R
@@ -52,7 +52,7 @@ LayerSf <- ggproto("LayerSf", Layer,
       self$geom_params$legend <- "polygon"
 
       # now check if the type should not be polygon
-      if (!is.null(self$mapping$geometry)) {
+      if (!is.null(self$mapping$geometry) && quo_is_symbol(self$mapping$geometry)) {
         geometry_column <- as_name(self$mapping$geometry)
         if (inherits(data[[geometry_column]], "sfc")) {
           sf_type <- detect_sf_type(data[[geometry_column]])
--- a/tests/testthat/test-geom-sf.R
+++ b/tests/testthat/test-geom-sf.R
@@ -76,6 +76,13 @@ test_that("geom_sf() determines the legend type from mapped geometry column", {
     ggplot(d_sf) + geom_sf(aes(geometry = g_line, colour = "a"))
   )
   expect_identical(p$plot$layers[[1]]$geom_params$legend, "line")
+
+  # If `geometry` is not a symbol, layer_sf() gives up guessing the legend type,
+  # and falls back to "polygon"
+  p <- ggplot_build(
+    ggplot(d_sf) + geom_sf(aes(geometry = identity(g_point), colour = "a"))
+  )
+  expect_identical(p$plot$layers[[1]]$geom_params$legend, "polygon")
 })
 
 test_that("geom_sf() removes rows containing missing aes", {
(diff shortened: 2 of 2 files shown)


yutannihilation:
Thanks.

PR #4128, opened 2020-07-11
Document flipped computed variables

### LINKED ISSUE #4124: Bug in setting fill of horizontal boxplots
I suspect I have found a bug in `geom_boxplot()` when trying to set the color of the box to be a computed variable. ie.

```r
> library(ggplot2)

> #Horizontal boxes with fill doesn't work with computed variables 
> ggplot(iris)+  geom_boxplot(aes(y=Species,x=Petal.Width,fill=..middle..)) 
Error in FUN(X[[i]], ...) : object 'middle' not found 
> ggplot(iris)+   geom_boxplot(aes(y=Species,x=Petal.Width,fill=..lower..))
 Error in FUN(X[[i]], ...) : object 'lower' not found
>
```
Other related variants on this theme seem to work ok

````r
library(ggplot2)
#Vertical boxes with fill works with computed variables
ggplot(iris)+
  geom_boxplot(aes(x=Species,y=Petal.Width,fill=..middle..))

#Horizontal boxes without fill works
ggplot(iris)+
  geom_boxplot(aes(y=Species,x=Petal.Width))

#Horizontal boxes with categorical variables works
ggplot(iris)+
  geom_boxplot(aes(y=Species,x=Petal.Width,fill=Species))
````
Running ggplot Version: 3.3.2.9000 on Windows, built from the master branch on Github 2020-07-09 16:14:23 UTC

### PR DESCRIPTION
**Pull request by yutannihilation**
Fix #4124

These four stat have variables whose names differ depending on the orientation.

- `stat_boxplot()`: `middle`, `lower`, and `upper`. This feels a bit tricky at least in two points; the flipped names are asynmetric (e.g. `middle` vs `xmiddle`, not `ymiddle` vs `xmiddle`), and `notchlower` and `notchupper` doesn't have their flipped form.
- `stat_smooth()`: `y`, `ymin`, and `ymax`
- `stat_function()` ~~`stat_ecdf()`~~ : these generate both `x` and `y`, so the documentation would be cryptic like below, so I don't address them in this pull request (at least they don't have undocumented computed variables)

```
#' `stat_function()` computes the following variables:
#' \describe{
#'   \item{x or y}{x or y values along a grid}
#'   \item{y or x}{value of the function evaluated at corresponding x or y}
#' }
```

edit: I just remembered `stat_ecdf()` doesn't support flipping (c.f. #4005)

### DIFF
--- a/R/stat-boxplot.r
+++ b/R/stat-boxplot.r
@@ -4,13 +4,13 @@
 #' @section Computed variables:
 #' \describe{
 #'   \item{width}{width of boxplot}
-#'   \item{ymin}{lower whisker = smallest observation greater than or equal to lower hinge - 1.5 * IQR}
-#'   \item{lower}{lower hinge, 25% quantile}
+#'   \item{ymin or xmin}{lower whisker = smallest observation greater than or equal to lower hinge - 1.5 * IQR}
+#'   \item{lower or xlower}{lower hinge, 25% quantile}
 #'   \item{notchlower}{lower edge of notch = median - 1.58 * IQR / sqrt(n)}
-#'   \item{middle}{median, 50% quantile}
+#'   \item{middle or xmiddle}{median, 50% quantile}
 #'   \item{notchupper}{upper edge of notch = median + 1.58 * IQR / sqrt(n)}
-#'   \item{upper}{upper hinge, 75% quantile}
-#'   \item{ymax}{upper whisker = largest observation less than or equal to upper hinge + 1.5 * IQR}
+#'   \item{upper or xupper}{upper hinge, 75% quantile}
+#'   \item{ymax or xmin}{upper whisker = largest observation less than or equal to upper hinge + 1.5 * IQR}
 #' }
 #' @export
 stat_boxplot <- function(mapping = NULL, data = NULL,
--- a/R/stat-boxplot.r
+++ b/R/stat-boxplot.r
@@ -2,6 +2,7 @@
 #' @param coef Length of the whiskers as multiple of IQR. Defaults to 1.5.
 #' @inheritParams stat_identity
 #' @section Computed variables:
+#' `stat_boxplot()` provides the following variables, some of which depend on the orientation:
 #' \describe{
 #'   \item{width}{width of boxplot}
 #'   \item{ymin or xmin}{lower whisker = smallest observation greater than or equal to lower hinge - 1.5 * IQR}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: R/stat-boxplot.r [hunk 1]
```diff
@@ -4,13 +4,13 @@
...
+#'   \item{ymin or xmin}{lower whisker = smallest observation greater than or equal to lower hinge - 1.5 * IQR}
+#'   \item{lower or xlower}{lower hinge, 25% quantile}
 #'   \item{notchlower}{lower edge of notch = median - 1.58 * IQR / sqrt(n)}
-#'   \item{middle}{median, 50% quantile}
+#'   \item{middle or xmiddle}{median, 50% quantile}
 #'   \item{notchupper}{upper edge of notch = median + 1.58 * IQR / sqrt(n)}
-#'   \item{upper}{upper hinge, 75% quantile}
-#'   \item{ymax}{upper whisker = largest observation less than or equal to upper hinge + 1.5 * IQR}
+#'   \item{upper or xupper}{upper hinge, 75% quantile}
+#'   \item{ymax or xmin}{upper whisker = largest observation less than or equal to upper hinge + 1.5 * IQR}
```
clauswilke, line 13: "or `xmax`"

R/stat-boxplot.r
```diff
@@ -4,13 +4,13 @@
 #' @section Computed variables:
```
clauswilke, line 4: I think it would be helpful to add a sentence stating that the computed variables depend on the orientation.

R/stat-smooth.r
```diff
@@ -31,9 +31,9 @@
 #'   function defined by `method`.
 #' @section Computed variables:
```
clauswilke, line 32: As above: state that calculated variables depend on orientation.

R/stat-boxplot.r
yutannihilation: Oops, thanks...

R/stat-boxplot.r
yutannihilation: Sure.

yutannihilation:
Added this sentence. Does this sound good?

> `stat_boxplot()` provides the following variables, some of which depend on the orientation:

clauswilke:
I think this is fine. Maybe put the "or" in italics just as is the case for the aesthetics.

clauswilke:
The "or" between computed variables, I mean.

yutannihilation:
Makes sense, I put them in Italic.

clauswilke:
Oh, this was merged already. Looks like my browser had cached an old version.

PR #4129, opened 2020-07-11
Refer to ggplot_global for x and y aesthetics variables

yutannihilation:
I happened to find `scale_{x,y}_binned()` doesn't include `x0` and `y0`, which I think is unintended. Probably we should refer to `ggplot_global` directly in these cases. Hard coding the same list on 4 places is a bit hard to maintain.


yutannihilation:
Sure. I'll file a new issue for `ggplot_global` and let's discuss there.

PR #4145, opened 2020-07-23
typos; stick to en-GB; prefer project to website

stragu:
Use ColorBrewer project name rather than obsolete website name.

Not sure what is meant by "use blender variants" in the last example. Should it be "use binned variants"?

### DIFF
--- a/R/scale-brewer.r
+++ b/R/scale-brewer.r
@@ -1,15 +1,15 @@
-#' Sequential, diverging and qualitative colour scales from colorbrewer.org
+#' Sequential, diverging and qualitative colour scales from ColorBrewer
 #'
 #' @description
-#' The `brewer` scales provides sequential, diverging and qualitative
+#' The `brewer` scales provide sequential, diverging and qualitative
 #' colour schemes from ColorBrewer. These are particularly well suited to
 #' display discrete values on a map. See \url{http://colorbrewer2.org} for
 #' more information.
 #'
 #' @note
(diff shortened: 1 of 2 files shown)


stragu:
I don't believe any of the failed checks are related to these minor fixes.

R/scale-brewer.r
```diff
@@ -1,15 +1,15 @@
...
 #'
 #' @description
-#' The `brewer` scales provides sequential, diverging and qualitative
+#' The `brewer` scales provide sequential, diverging and qualitative
 #' colour schemes from ColorBrewer. These are particularly well suited to
 #' display discrete values on a map. See \url{http://colorbrewer2.org} for
 #' more information.
 #'
 #' @note
-#' The `distiller` scales extend brewer to continuous scales by smoothly
```
yutannihilation, line 10: I feel this should not be removed.

stragu: I removed it because it is repeated later on in the same sentence: "...to a continuous scale." If you read it out loud I think you'll find it is quite wordy / repetitive.

The sentence needs to at least be reworded, as "extends brewer to continuous scales" is not the best wording, in my opinion.

Alternatively:

> The `distiller` scales are continuous scales created by smoothly interpolating 7 colours from any `brewer` palette.

yutannihilation: Ah, sorry, I missed it. I see.

yutannihilation:
Thanks!

PR #4147, opened 2020-07-25
Fix CI failures

yutannihilation:
## Summary of the changes

* Skip some tests if quantreg package is not available
* Replace macOS + R-devel runner and with Ubuntu-16.04 + R-devel runner
* Install binary packages only on macOS
* Remove the step to add libgit PPA

### quantreg on R < 3.6

As of version 5.61 (released on July 7, 2020), quantreg package imports [conquer package](https://cran.r-project.org/web/packages/conquer/index.html), which requires R >= 3.6. So, quantreg now virtually requires R >= 3.6.

Diff: https://github.com/cran/quantreg/commit/16f49417c4f7ce6308ba660aba56ee70319cd396#diff-35ba4a2677442e210c23a00a5601aba3R31

#### R-devel

It fails to setup R for an unknown error. Now that r-lib/actions provide R-devel also to Linux runners, let's use `ubuntu-16.04` for R-devel. I feel the toolchain of macOS is too franky, and as ggplot2 is not a package that needs compilation, I feel the macOS + R-devel is not worth keeping.

https://github.com/r-lib/actions/issues/49#issuecomment-665047147

#### R 4.0

`brew install gdal` fails with the following error. I set the `pkgType` option on macOS to avoid building packages at all.

```
==> Installing gdal dependency: gcc
==> Pouring gcc-10.1.0.catalina.bottle.tar.gz
##[error]Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/gfortran
Target /usr/local/bin/gfortran
already exists. You may want to remove it:
  rm '/usr/local/bin/gfortran'

To force the link and overwrite all conflicting files:
  brew link --overwrite gcc

To list all files that would be deleted:
  brew link --overwrite --dry-run gcc

Possible conflicting files are:
/usr/local/bin/gfortran -> /usr/local/gfortran/bin/gfortran
```

### libgit2 PPA

(This is not about failures, just for cleanup)

As https://github.com/r-lib/actions/pull/97 is already merged, we can just remove this from the YAML.

PR #4153, opened 2020-07-30
Fix a typo in pr-command.yaml

yutannihilation:
My mistake in #4147...


yutannihilation:
No, it's me who should say sorry as this is just my fault... Thanks for approving many times!

Issue #4158, opened 2020-08-03
ggproto docs about Geom are wrong?


burchill:
From #3509, it seems that `setup_params()` has been added to the Geom object, but the docs (i.e., for `ggplot2-ggproto`) explicitly say that the Geom doesn't have this method.

yutannihilation:
Thanks, the docs seem incorrect. Would you like to contribute a pull request?

burchill:
I can do it, but I don't think I know enough about Geoms/ggproto to do it well. Should I just delete the whole paragraph?

yutannihilation:
Yes, I think you can just delete it. Until https://github.com/tidyverse/ggplot2/pull/3506, we haven't found a nice usage of `setup_params()` on Geom objects, but now we know we actually need the method.

yutannihilation:
Fixed by #4161

PR #4160, opened 2020-08-04
Doc fix: Geoms no longer said to not have `setup_params()`

burchill:
As per #4158, the docs used to say that Geoms do not have the `setup_params()` method. This change removes that faulty information


burchill:
Oops, I'll make a cleaner pull request--this contains old commits from a separate pull request I made a while ago

PR #4161, opened 2020-08-04
Docs no longer say Geoms don't have setup_params() method

burchill:
Per #4158, the docs used to say that Geoms don't have the `setup_params()` method, which is no longer true (#3506). I've deleted this paragraph.


yutannihilation:
Thanks! Sorry, looking at this again, I now feel we can delete only the last sentence and keep the other words as it seems they are still useful. What do you think?

burchill:
You're the expert here! (And I agree).

PR #4164, opened 2020-08-06
Set file background from plot theme

### LINKED ISSUE #4057: ggsave should use plot theme for default background
I think @djnavarro tweeted about this recently, and I hit it too. It seems like in a situation like this:

```R
plot %>%
  ...
  theme(
   plot.background = element_rect(fill = NA, colour = NA)
  )

ggsave("ratios.png", width = 8, height = 4.5, bg = NA)
```

`ggsave()` could automatically figure out what `bg` should be and pass it along to the graphics device.

karawoo:
Closes #4057. @clauswilke I took your proposed approach and calculated the theme element to get the background color. I imagine people will still typically want to edit their theme to remove the plot border.

I'm opening this as a draft because I think it needs a test but I'm not sure how best to write one. Is there a way to have vdiffr compare two files rather than comparing a plot object against a file? I imagine there is and that vdiffr is doing this under the hood normally, but I tinkered with it a bit and couldn't figure it out.

``` r
library("ggplot2")

  geom_point() +
  coord_fixed() + # using coord_fixed() just to emphasize what's the plot and what's the image background
  theme(plot.background = element_rect(fill = "forestgreen"))

ggsave("~/p_new.png", p)
#> Saving 7 x 5 in image
knitr::include_graphics("~/p_new.png")
```

``` r

ggsave("~/p_override.png", p, bg = "cyan")
#> Saving 7 x 5 in image
knitr::include_graphics("~/p_override.png")
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,7 +63,8 @@ Suggests:
     sf (>= 0.7-3),
     svglite (>= 1.2.0.9001),
     testthat (>= 2.1.0),
-    vdiffr (>= 0.3.0)
+    vdiffr (>= 0.3.0),
+    xml2
 Enhances: sp
 License: GPL-2 | file LICENSE
 URL: http://ggplot2.tidyverse.org, https://github.com/tidyverse/ggplot2
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* `ggsave()` now sets the default background to match the fill value of the
+  `plot.background` theme element (@karawoo, #4057)
+
 * Extended `stat_ecdf()` to calculate the cdf from either x or y instead from y only (@jgjl, #4005).
 
 * Fixed a bug in `labeller()` so that `.default` is passed to `as_labeller()`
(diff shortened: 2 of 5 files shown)


clauswilke:
You could write the png to a temp file, read it back in, plot as a `rasterGrob()` with grid, and then vdiffr can compare that.

karawoo:
I couldn't get the `rasterGrob()` method to work, so instead of writing a visual test I'm saving to svg and finding the background rect with xpath. This involves adding xml2 to Suggests which maybe we don't want to do for just one test. If you all prefer I can write a regex instead, that may be ok for this purpose.

clauswilke:
What was the problem with the `rasterGrob()`approach? Maybe it can be fixed? Could you post the code you tried?

karawoo:
Thanks @clauswilke, and you feel ok about adding xml2 to Suggests?

clauswilke:
I realized that writing an image and reading it back in to plot would also require an additional dependency (e.g. png), so it's a wash. And in that case just writing to and reading from svg (rather than writing a png, reading it in, plotting to svg, and then comparing that svg to a reference) is much more efficient. And in any case, vdiffr imports xml2, so we have this dependency already, implicitly.

PR #4188, opened 2020-09-03
#4173 lambda functions in discrete scales & facets

### LINKED ISSUE #4173: Allow passing arguments to the custom labelling functions
In the discrete scales (e.g. `scale_y_discrete()`), we are allowed to specify a vector of values, NULL, `waiver()` or any function that takes the breaks and returns the labels desired. The last option is especially neat as it gives the user plenty of possibilities. However, the breaks are passed only as the first argument of the function. When I want to gain any control, I have to "wrap" my function with `function(x) my_function(arg_1, arg_2)`. In other words, labels argument of a scale layer can only handle function "name/object" and not a function call.

Another and quite a similar issue concerns `facet_wrap` and `facet_grid` layers; this time, a special `labeller `function is introduced for labelling and there is even documented `as_labeller` function. This works great, but again, when I construct `my_function` labeller, it cannot be specified as a function call with arguments. Again, when wrapped with `as_labeller(function(x) my_function(arg_1, arg_2))` it does the job as intended.

Is there any workaround that I am missing? If not, I would highly appreciate the option to pass additional arguments to my labelling function. Maybe it could be accomplished with some `labs_args` list user can specify within a scale/facet layer or possibly, it would be even nicer to have `purrr `style recipe, like `~ my_function(.x, arg_1, arg_2)`.

Thanks a lot!

netique:
Fix #4173 

second PR, previous one was failing tests... now it should work as intended


clauswilke:
In the future, please don't open a new PR just because something needs correcting in the previous PR. You can just push additional commits and the PR will update.

R/labeller.r
```diff
@@ -426,7 +427,15 @@ labeller <- function(..., .rows = NULL, .cols = NULL,
...
   }
+
   dots <- list(...)
+
+  if (length(dots) == 1) {
+    if (is.formula(dots[[1]]) || is.function(dots[[1]]) && !is_labeller(dots[[1]])) {
+      return(as_labeller(dots[[1]]))
+    }
+  }
+
```
thomasp85, line 438: Can you elaborate on this? This seems orthogonal to the intent of the PR

netique: My bad, the chunk was meant to be deleted. I presume I was thinking about allowing to pass a lambda function directly as a `labeller()` argument back then, but that's way broader topic. From documetation, it is clear (to some extent) that unnamed arguments in `labeller()` is treated by `.default` labeller and `as_labeller()` should be used instead. Deleted in the upcoming commit to this PR branch.

R/labeller.r
```diff
@@ -311,6 +311,8 @@ as_labeller <- function(x, default = label_value, multi_line = TRUE) {
       x(labels)
     } else if (is.function(x)) {
       default(lapply(labels, x))
+    } else if (is.formula(x)) {
+      default(lapply(labels, as_function(x)))
```
thomasp85, line 315: Is there a reason why you pivoted on the earlier approach of converting x to a function once at the top?

thomasp85:
@netique We recently merged in another PR that touched upon some of the same areas - can I get you to update this PR so it incorporates the changes in the master branch?

netique:
> @netique We recently merged in another PR that touched upon some of the same areas - can I get you to update this PR so it incorporates the changes in the master branch?

Sure, done.

R/scale-.r
```diff
@@ -184,6 +184,7 @@ discrete_scale <- function(aesthetics, scale_name, palette, name = waiver(),
 
   check_breaks_labels(breaks, labels)
 
+  if (is.formula(labels)) labels <- as_function(labels)
```
teunbrand: Sorry this is mostly my fault for getting in your way with #4427, but the `allow_lambda()` statements below already capture formulas as functions, so this line has become unnecessary. Apologies!

R/labeller.r
```diff
@@ -433,7 +435,10 @@ labeller <- function(..., .rows = NULL, .cols = NULL,
   if (!is.null(keep.as.numeric)) {
     .Deprecated(old = "keep.as.numeric")
   }
+
   dots <- list(...)
+
+
```
thomasp85, line 441: Can I get you to revert these superfluous changed

netique: My apologies. Fixed in https://github.com/tidyverse/ggplot2/pull/4188/commits/6d488b6716c5228bc68f9fc1841888abbffa1565.

thomasp85: No apology needed - thanks for your work on this

thomasp85 (approved):

thomasp85:
Last thing - can I get you to add a news bullet about the change

thomasp85:
Thanks!

PR #4194, opened 2020-09-06
Backtransform data before mapping statistics

### LINKED ISSUE #4155: Possible bug in `stage()`/`after_stat()` with scale transformations.
I think I might have found a bug with either the scales or the `stage()`/`after_stat()` functions in combination with dates. 
I came across this when making a custom stat where I wanted to put `default_aes = aes(xmin = after_stat(start), xmax = after_stat(end))` in the ggproto methods, where `start` and `end` are computed variables.

What happens is that when I try to apply a `stage()` or `after_stat()` to a date variable, it gives me an error that the date transformation is invalid. I don't think it is a problem with stats, as it also shows up with `stat_identity()`.
The attempted plot below is not very interesting, but it is the most minimal example I could come up with.

``` r
library(ggplot2)

ggplot(economics) +
  geom_point(aes(x = stage(date, after_stat = x), 
                 y = unemploy))
#> Error: Invalid input: date_trans works with objects of class Date only
```

The error above doesn't happen for integer or numeric data.
```
ggplot(economics) +
  geom_point(aes(x = stage(as.integer(date), after_stat = x), 
                 y = unemploy))
```

And also the error can be rescued by formatting the output as a date again, but this seems overly verbose and seems difficult to anticipate in a custom stat.
``` r
ggplot(economics) +
  geom_point(aes(x = stage(date, after_stat = structure(x, class = "Date")), 
                 y = unemploy))
```

The same is true for other geoms and aesthetics as well. Here are similar examples with `after_stat()` instead of `stage()`.

``` r
ggplot(economics) +
  geom_segment(aes(x = date, xend = after_stat(x + 30),
                   y = unemploy, yend = unemploy))
#> Error: Invalid input: date_trans works with objects of class Date only
```

```
ggplot(economics) +
  geom_segment(aes(x = as.integer(date), xend = after_stat(x + 30),
                   y = unemploy, yend = unemploy))
```

``` r
ggplot(economics) +
  geom_segment(aes(x = date, xend = after_stat(structure(x + 30, class = "Date")),
                   y = unemploy, yend = unemploy))
```

<sup>Created on 2020-07-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Fix #4155

On https://github.com/tidyverse/ggplot2/issues/4155#issuecomment-673932456, I wrote 

> The staged values should be bypass the transformation as probably it is already transformed

but, I was wrong. The transformation that the data needs to be bypassed is not the one after mapping the calculated variables, but the one *before* mapping. Otherwise, if we do another calculation over the calculated value, the result is still wrong even if it skip the transformation at the end of `Layer$map_statistic()`.

Here's an example. When evaluating `after_stat`, the actual scaled value of `x` is `4`, and the result of `x / 2` will be `2`, which will locate at `4` in sqrt scale. But, in actual, `x / 2` is `8`.

``` r
d <- data.frame(value = 16)

ggplot(d) +
  geom_point(aes(stage(value, after_stat = x / 2), 0))
```

The transformation we want to bypass is this one. But, the operations that come after this (i.e. training and mapping positions) depends on the scaled values, so we cannot skip this transformation.

https://github.com/tidyverse/ggplot2/blob/6b8dba07cfe2bc7a01d3545c7615a60c984c07c7/R/plot-build.r#L60-L61

After some thinking, I come to a conclusion that the values need to be back-transformed before evaluating `after_stat`. There might be cleaner mechanism for this, but I believe this is a realistic solution at the moment...

-------

``` r
library(ggplot2)

d <- data.frame(value = 16)

# Current behaviour

ggplot(d) +
  geom_point(aes(value / 2, 0), colour = "green", size = 10) +
  geom_point(aes(stage(value, after_stat = x / 2), 1), colour = "purple", size = 10) +
  scale_x_sqrt(limits = c(0, 16), breaks = c(2, 4, 8))
```

``` r
# This pull request

#> ℹ Loading ggplot2

ggplot(d) +
  geom_point(aes(value / 2, 0), colour = "green", size = 10) +
  geom_point(aes(stage(value, after_stat = x / 2), 1), colour = "purple", size = 10) +
  scale_x_sqrt(limits = c(0, 16), breaks = c(2, 4, 8))
```

<sup>Created on 2022-05-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


clauswilke:
Something makes me uncomfortable about this. At a minimum, I think this needs much more documentation to really explain the logic at each step.

thomasp85:
will this not potentially break every plot that has ever used `stat()` or `..aes..`?

I must admit that I think it is better to document what has happened to the data up until the `after_stat` phase and simply let that be the end of it...

clauswilke:
@thomasp85 Something like this PR may very well be needed. I've dealt with a similar problem in coords. However, I think we should do some more due diligence before merging.

It'd be a breaking change, but I think the cases where this would hit would be quite rare, since it requires a combination of using `stat()` and using a scale with transformation, and the latter is usually applied to positions and the former is usually not. I'm not sure I've ever made a plot that meets this criterion (except my reprex below).

I'm providing another reprex, using the current ggplot2 (without this patch), that shows that the current behavior is confusing and inconsistent with how everything else in ggplot2 works.

    library(ggplot2)
    library(ggridges)

    df <- data.frame(x = rexp(100))

    # works as expected
    ggplot(df, aes(x, y = 0, fill = stat(x) < 1)) +
      geom_density_ridges_gradient()
    #> Picking joint bandwidth of 0.287




    # confusing result. why are all points red?
    ggplot(df, aes(x, y = 0, fill = stat(x) < 1)) +
      geom_density_ridges_gradient() +
      scale_x_log10()
    #> Picking joint bandwidth of 0.194




    # we have to transform x according to the scale before things work,
    # that seems strange from a user's perspective
    ggplot(df, aes(x, y = 0, fill = stat(x) < 0)) +
      geom_density_ridges_gradient() +
      scale_x_log10()
    #> Picking joint bandwidth of 0.194



<sup>Created on 2020-09-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
I'm not convinced, but I'm also not 100% against. Scale transformations are applied as the very first thing in the data pipeline which is something there is value in teaching and understanding. It seems weird to me to cherry-pick parts of the data-pipeline and undue them at certain points...

Again, not saying this shouldn't be done, but I'd be weary of doing this since, as you note yourself, this is an extreme edge case and the change required is not nothing...

clauswilke:
Another example. This one I find even more disconcerting, because it produces a plot that is objectively wrong.

    library(ggplot2)
    library(ggridges)

    df <- data.frame(x = rexp(100))

    ggplot(df, aes(x, y = 0, fill = stat(x))) +
      geom_density_ridges_gradient() +
      scale_x_log10() +
      scale_fill_viridis_c()
    #> Picking joint bandwidth of 0.171



<sup>Created on 2020-09-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Well, it is only objectively wrong because we strip `stat()` from the guide title 😉

clauswilke:
A possible compromise could be to leave `after_stat()` as is but add a new function that explicitly backtransforms. E.g. `after_stat_bt()`. Maybe somebody can come up with a better name.

thomasp85:
I'd much rather explore this (or make it an argument to `after_stat()`). In general I'd be very skeptic from a performance point of view to transform all aesthetics back and forth for no reason at all in 99.9% of all plot cases

yutannihilation:
I too feel unhappy about this PR, and half of the purposes of this is to share the uncomfortableness with you so that you'll come up with some superseding solution, as you always did :)

> I've dealt with a similar problem in coords.

Yeah, actually it made me think back-transformation is needed here as well.

> In general I'd be very skeptic from a performance point of view to transform all aesthetics back and forth for no reason at all in 99.9% of all plot cases

Probably we can easily skip the back-transformation by checking if the trans is identity? Besides, the back-transformation only happens when there's some Stat (otherwise it returns early), so I don't think it make up 99.9%.

Whether or not we end up adopting this, I think this is worth breaking change as the name `after_stat()` doesn't sound right, especially, now that we have `after_scale()`, which makes the users think `after_stat()` is before scale. And the actual breakage would be rare.

clauswilke:
> Whether or not we end up adopting this, I think this is worth breaking change as the name `after_stat()` doesn't sound right, especially, now that we have `after_scale()`, which makes the users think `after_stat()` is before scale. And the actual breakage would be rare.

I'm not suggesting changing these names again, but I did realize today that I've been always confused by `after_scale()` because it probably should be called `after_mapping()`. The data gets transformed (some may think of this as scaling the data), then the statistical transformations are applied, and finally the data gets mapped.

thomasp85:
> Besides, the back-transformation only happens when there's some Stat (otherwise it returns early), so I don't think it makes up 99.9%

Sorry - I confused myself as to when this function would get called.

> I'm not suggesting changing these names again, but I did realize today that I've been always confused by after_scale() because it probably should be called after_mapping(). The data gets transformed (some may think of this as scaling the data), then the statistical transformations are applied, and finally the data gets mapped.

I'm sure `after_mapping()` would be more confusing since we use the mapping term for a different purpose in the API (assigning data to aesthetics). I feel `after_scale()` is correct. This step happens after all scaling operations (transformation, censoring, mapping etc) has completed. I personally don't think these names in any way imply that all operations in the scale or the stat happens in one chunk, which is what I think all this confusing is from.

thomasp85:
@clauswilke what kind of related issues have you face in coords?

clauswilke:
For geoms such `geom_hline()` or `geom_vline()`, I needed to backtransform the range to make these geoms work correctly:
https://github.com/tidyverse/ggplot2/blob/ac2b5a7bb460179e0cc8c4b3204795317dcfb9b8/R/geom-hline.r#L46-L48

Getting this all right was quite tricky, because in many places the code wasn't very clear about whether it needed a regular range or a backtransformed range. It took quite a while to disentangle all of this. I tried to document this here:
https://github.com/tidyverse/ggplot2/blob/ac2b5a7bb460179e0cc8c4b3204795317dcfb9b8/R/coord-.r#L16-L30

yutannihilation:
> I'm sure `after_mapping()` would be more confusing since we use the mapping term for a different purpose in the API (assigning data to aesthetics). I feel `after_scale()` is correct.

I agree with this part. While the word "scale" might have different meanings in different places in ggplot2, probably it still has cleaner meaning than "mapping," which is too general.

thomasp85:
Team - should we revive this for the upcoming release? I don't think my stance has changed all that much, but I'd be happy to consider an argument — I don't think a new function is the correct way because there would be no way to port this over to `stage()`

clauswilke:
I just reread the entire thread and I'm not sure I have a useful opinion either way. Maybe it would be a good idea to bring in @teunbrand, since he filed the original issue that brought this up.

If we want to go forward with this, I think it needs an empirical approach. Implement a version of the suggested idea and see if it breaks things or has performance implications.

R/layer.r
```diff
@@ -299,6 +299,9 @@ Layer <- ggproto("Layer", NULL,
     # evaluation (since the evaluation symbols gets renamed)
     data <- rename_aes(data)
 
+    # data needs to be non-scaled
+    data_orig <- scales_backtransform_df(plot$scales, data)
```
teunbrand, line 303: I think you might be able to alleviate Thomas' concern that this is applied to 99% of plots, by executing this line after the early exit at the `if (length(new) == 0) return(data)` line.

teunbrand:
I think we might even further reduce the impact of this PR on performance if we selectively backtransform the relevant columns, and leave other columns as-is.

``` r
# Suppose these are the aesthetics to evaluate
new <- aes(x = after_stat(y + 1), colour = 2, fill = Species)

# Extract all variables occuring in the aesthetics
vars <- unlist(lapply(new, all.vars), use.names = FALSE)

# Only backtransform variables that occur
data_orig <- scales_backtransform_df(scales, df[intersect(names(df), vars)])
```

R/layer.r
yutannihilation: Agreed, thanks!

yutannihilation:
Thanks!

> ```r
> vars <- unlist(lapply(new, all.vars), use.names = FALSE)
> ```

I think this works in most cases, there's chance that some variable is referenced indirectly (e.g. `get("x")`). While it's probably not a good practice, I think we should not limit the variables for safety.

yutannihilation:
If the concern is the performance, I think I addressed it. `scales_transform_df()` can be left as is, but I added the same workaround to skip non-transforming trans for consistency.

thomasp85:
@yutannihilation is this ready for review?

yutannihilation:
Yes.

thomasp85 (approved):
Can you add a news bullet indicating that this is a breaking change (though quite niche)

otherwise it's good to merge

yutannihilation:
Thanks! Let me find some good explanation for NEWS.

yutannihilation:
Added a test and a NEWS item.

PR #4216, opened 2020-10-01
Update README.Rmd

devdesai16:
Updated installation process in README

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -34,6 +34,9 @@ install.packages("ggplot2")
 # Or the development version from GitHub:
 # install.packages("devtools")
 devtools::install_github("tidyverse/ggplot2")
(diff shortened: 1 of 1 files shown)


yutannihilation:
Hi, are you very new to R? You almost always need to `library()` a package to use it, and I think it's what we can assume an ordinal R user knows, so I don't think this is worth describing explicitly. I'm not sure if you have good intentions or you're just doing a hacktoberfest as pointed on https://github.com/kubernetes-sigs/kubespray/pull/6775#issuecomment-702082697 (really sorry if it's not true), let me close this.

Note that, though I reject this pull request, we always welcome small documentation improvements (including fixing a typo) if it's meaningful. Feel free to create another pull request if you find something that needs to be improved :)

Issue #4189, opened 2020-09-04
Feature request: support for tibble aesthetics


davidchall:
I'm developing a ggplot2 extension where it would be helpful to pass a tibble column as an aesthetic.

As a simple motivating example, you could imagine a `geom_box()` layer with a `bounds` aesthetic that expects a tibble with columns xmin, ymin, xmax and ymax. Then I could simply do `geom_box(bounds = bbox)` and the layer will use the nested columns to draw the box.

To do this, we need to make use of nested tibbles (i.e. a tibble column within a tibble).

``` r
library(tidyverse)

dat <- tibble(a = tibble(x = 1:5, y = 6:10))

# works fine
ggplot(dat) +
  geom_point(aes(x = a$x, y = a$y))
```

So far, so good. But these aesthetics are still just vectors. Let's try using the tibble column `a`.

(Obviously, `geom_point()` doesn't expect a tibble column for its `x` aesthetic, but it _does_ generate the relevant error.)

``` r
ggplot(dat) +
  geom_point(aes(x = a, y = a$y))
#> Don't know how to automatically pick scale for object of type tbl_df/tbl/data.frame. Defaulting to continuous.
#> Error: Aesthetics must be either length 1 or the same as the data (5): x
```

The warning can be removed with a new `scale_type`. But the error is generated by this line: https://github.com/tidyverse/ggplot2/blob/813d0bd8c7d4df3b3cba170de54233076b346c1c/R/geom-.r#L211

The length of a data frame is the number of columns (2) instead of the number of rows (5), so we get this error. This is the same problem that `vctrs::vec_size()` addresses.

Would it be possible to simply avoid this error by replacing `length()` with `nrow()`, `NROW()` or `vec_size()`? Or would there be other repercussions?

yutannihilation:
Thanks. I'm not sure if this is really doable, but I think data frame column ("nested tibble," or "packed column"? I don't know the proper term for this...) should be supported. This might be a nice reason to start importing vctrs.

davidchall:
To understand the motivation with a more fleshed out example, please take a look at [this vignette](https://davidchall.github.io/ggip/articles/ggip.html). This package solves the problem using an vctrs-based class that behaves effectively like a data frame whose `length()` returns `vec_size()`.

teunbrand:
Disclaimer: the following is mostly out of selfish reasons, so feel free to dismiss entirely.

[...]

PR #4224, opened 2020-10-08
Use NROW() instead of length() in aesthetics checking.

teunbrand:
In response to #4189, I'm creating this PR to see if any checks fail.
They didn't fail when I ran the unit tests locally.

However, there are a few things to note, as just changing `length()` to `NROW()` is not sufficient to allow for tibble aesthetics.
I'll post some details about this in the relevant issue.


teunbrand:
I'll close this because I don't think the changes here are insufficient to help the original issue.

PR #4227, opened 2020-10-11
Add missing period at the end of sentence

mine-cetinkaya-rundel:
The error currently reads something like the following:

```
Error: `data` must be a data frame, or other object coercible by `fortify()`, not an S3 object with class uneval
Did you accidentally pass `aes()` to the `data` argument?
```

This PR adds a period at the end of the first sentence.

### DIFF
--- a/R/fortify.r
+++ b/R/fortify.r
@@ -41,7 +41,7 @@ fortify.grouped_df <- function(model, data, ...) {
 fortify.default <- function(model, data, ...) {
   msg <- paste0(
     "`data` must be a data frame, or other object coercible by `fortify()`, ",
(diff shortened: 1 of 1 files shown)


clauswilke (approved):
For a moment I thought this needed updated documentation files, but of course it doesn't.

yutannihilation:
The failure on R-devel seems irrelevant to this pull request (I'll fix later). Can we merge this?

> Graphics API version mismatch

clauswilke:
I think so. I was just running out the door when I approved this so it fell through the cracks.

yutannihilation:
Thanks!

PR #4229, opened 2020-10-12
Add `oob` arguments to date/datetime position scales.

### LINKED ISSUE #4199: Datetime scales don't support out of bounds (oob) arguments
I've noticed that date/time scales do not support the `oob` argument that the continuous scales have, while I do think this might be a useful feature.

Let's say I have the following graph, of which I would want the x-axis labelling:

``` r
library(ggplot2)

  geom_line() +
  scale_x_date(limits = as.Date(c("1970-01-01", "2010-01-01")))
#> Warning: Removed 93 row(s) containing missing values (geom_path).
```

But I'd like the line to look like the following:

``` r


ggplot(economics, aes(as.numeric(date), unemploy)) +
  geom_line() +
  scale_x_continuous(
    limits = as.numeric(as.Date(c("1970-01-01", "2010-01-01"))),
    oob = scales::oob_squish
  )
```

<sup>Created on 2020-09-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This does not currently seem possible. However, as the `ggplot2:::datetime_scale()` function passes the `...` to `continuous_scale()` (which *does* use `oob`), it would seem most of the infrastructure for `oob` arguments is already there.

Thanks for considering!

teunbrand:
This PR aims to fix #4199 by adding `oob` arguments to date and datetime position scales that are passed along to `ggplot2:::datetime_scale()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,9 @@
   `scale_*(guide = FALSE)` to remove a guide. Please use 
   `guides(<scale> = "none")` or `scale_*(guide = "none")` instead 
   (@yutannihilation, #4094).
+  
+* Date and datetime position scales support out-of-bounds (oob) arguments to 
+  control how limits affect data outside those limits (@teunbrand, #4199).
 
 # ggplot2 3.3.2
 This is a small release focusing on fixing regressions introduced in 3.3.1.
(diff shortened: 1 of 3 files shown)


yutannihilation (approved):
Looks good to me! Could you also add a NEWS bullet for this?

teunbrand:
Thanks, sure thing!

yutannihilation:
Thanks!

Issue #3816, opened 2020-02-11
Clarify the current status of fortify()


yutannihilation:
(Originally commented at https://github.com/tidyverse/ggplot2/pull/3802#issuecomment-583852613)

In my understanding, there are 2 purposes of `fortify()`.

1. User-facing utility to convert an object to data.frame, which otherwise the users don't know how to.
2. API for developers that let the custom objects handled nicely in the internal of ggplot2 by tweaking the object at the very first stage.

For purpose 1, ggplot2 now provides 2 types of methods, both of which are becoming less important:

* Convert model objects, which is superseded broom package. In addition, ggfortify covers much wider range of objects.
* Convert geospatial objects (of sp and maps), which is being superseded by sf package and `geom_sf()`.

So, we can and should deprecate (or at least discourage) this type of usages.

On the other hand, for purpose 2, I think `fortify()` still plays a somehow important role. For example, `fortify.tbl()` ensures `tbl` is `collect()`ed. 

https://github.com/tidyverse/ggplot2/blob/0137a4d7a25f6ffa3c80cd3e9626a9b0dbc22482/R/fortify.r#L19-L24

### Suggestions

I suggest

- [ ] Deprecate `fortify.lm()`, `fortify.glht()`, `fortify.confint.glht()`, `fortify.summary.glht()` and `fortify.cld()`. Replace examples using `broom::augment()` and `broom::tidy()`.
- [ ] Deprecate `fortify.<sp-object>` and `fortify.map()` after ensuring `geom_sf()` supersedes `geom_map()`.
- [ ] Merge the docs of `fortify()` to `autoplot()` and `autolayer()`, and explain how developers of some custom objects can implements these methods so that the users can use ggplot2 to visualize the objects painlessly (#3721).

Note that this issue is not very high priority as maintaining `fortify()` doesn't seem a heavy burden, at least at the moment. I filed this issue just because it seems we need some place to discuss to avoid confusion. Anyway, I believe it's a right move to encourage users to use broom and sf.

hadley:
This sounds like a good proposal to me, but it'll need to wait until a major/minor release.

teunbrand:
I'm considering the docs bullet adressed by #5745

yutannihilation:
> * [ ]  Deprecate `fortify.<sp-object>` and `fortify.map()`

[...]

PR #4239, opened 2020-10-14
add fortify for gamlss objects

hasandiwan:


yutannihilation:
Thanks for your attempt, but we don't want to add a new `fortify()` method to ggplot2 unless there's very strong reason. We recommend using `broom::tidy()` for this kind of usages. See #3816.

PR #4242, opened 2020-10-15
Do not use RStudio Package Manager for R-devel

yutannihilation:
All visual tests fail on R-devel with this error:

```
 > test_check("ggplot2")
── 1. Error: aesthetics are drawn correctly (@test-aes.r#160)  ─────────────────
Graphics API version mismatch
Backtrace:
 1. vdiffr::expect_doppelganger(...)
 2. vdiffr:::writer(fig, testcase, title)
 3. vdiffr:::svglite(file, user_fonts = get_aliases())
 4. vdiffr:::svglite_(...)
```
(<https://github.com/tidyverse/ggplot2/pull/4227/checks?check_run_id=1239216870#step:10:132>)

I tried clearing cache, but it still failed. I bet this is because some binary on RStudio package manager is outdated. This might be solved by just waiting for some package to get rebuilt, but it kept failing at least for this 2 weeks, so I think there's little hope. This pull request removes RStudio Package Manager on R-devel runner and clears cache.

This pull request adds these changes:

* Stop using RStudio Public Package Manager on R-devel runner
* Remove workaround for systemfonts package as it's available again
* Add a workaround for xml2 (it seems not available on R 3.2)

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -10,7 +10,7 @@ name: R-CMD-check
 
 # Increment this version when we want to clear cache
 env:
-  cache-version: v3
+  cache-version: v4
 
 jobs:
   R-CMD-check:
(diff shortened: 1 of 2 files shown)


yutannihilation:
Hmm, xml2 cannot be installed on R 3.2...?

yutannihilation:
This should be ready to merge now.

For the behaviour of RStudio Public Package Manager, I asked on RStudio Community:
https://community.rstudio.com/t/how-frequently-are-r-devel-binaries-rebuilt-on-rstudio-public-package-manager/84926

yutannihilation:
I got the answer.

> RStudio Package Manager doesn't build binary packages for R-devel yet. I think the GitHub Actions workflows currently install R 4.0 binaries for R-devel on Linux, which is probably okay in most cases.
(https://community.rstudio.com/t/how-frequently-are-r-devel-binaries-rebuilt-on-rstudio-public-package-manager/84926/2)

Since this isn't okay in our case, sticking with source seems a right choice.

karawoo (approved):
LGTM, thanks for all your work keeping the gh actions workflows working

yutannihilation:
Thanks!

Unfortunately, the master fails because, 4 days ago, the digest package increased the minimum required R version to 3.3... This time, we might need to consider dropping support for R 3.2. I'll file a new issue for this.

https://github.com/cran/digest/blame/d385e0ba04673903d026ac483da9e9a2f1f3f55c/DESCRIPTION#L23

PR #4244, opened 2020-10-20
Fix saving plots with no background

### LINKED ISSUE #4212: ggsave to svg fails for certain themes
ggsave to svg operation is failing when certain themes are used:

```r
require(ggplot2)
#> Loading required package: ggplot2
p = qplot(Sepal.Length, data=iris, geom='density') + theme_void()
ggsave(p, filename = 'test.svg', w=8, h=6)
#> Error in svglite_(file, bg, width, height, pointsize, standalone, aliases): Expecting a single string value: [type=NULL; extent=0].
```

I've tested {ggplot} and {ggthemes} and the following show this error:
- ggplot::theme_void
- ggplot::theme_minimal
- ggthemes::theme_map
- ggthemes::theme_tufte
- ggthemes::theme_solid

I'm running macOS 10.15.5, R 4.0.2, ggplot 3.3.2.9000.

karawoo:
Fixes #4212. When the `plot.background` itself is NULL, `ggsave()` will set the device bg to `"transparent"`. This seems to work for devices that support transparent backgrounds. JPEGs and PDFs will have a white background, which seems ok.

### DIFF
--- a/R/save.r
+++ b/R/save.r
@@ -88,6 +88,9 @@ ggsave <- function(filename, plot = last_plot(),
   }
   if (is_null(bg)) {
     bg <- calc_element("plot.background", plot_theme(plot))$fill
+    if (is_null(bg)) {
+      bg <- "transparent"
+    }
   }
   old_dev <- grDevices::dev.cur()
   dev(filename = filename, width = dim[1], height = dim[2], bg = bg, ...)
--- a/R/save.r
+++ b/R/save.r
@@ -87,10 +87,7 @@ ggsave <- function(filename, plot = last_plot(),
     filename <- file.path(path, filename)
   }
   if (is_null(bg)) {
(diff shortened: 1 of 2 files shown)


R/save.r
```diff
@@ -88,6 +88,9 @@ ggsave <- function(filename, plot = last_plot(),
   }
   if (is_null(bg)) {
     bg <- calc_element("plot.background", plot_theme(plot))$fill
+    if (is_null(bg)) {
```
clauswilke, line 91: Maybe just write:
```r
bg <- calc_element("plot.background", plot_theme(plot))$fill %||% "transparent"
```

karawoo: Oh of course, thanks, will update

clauswilke (approved):

Issue #3174, opened 2019-03-05
Issue with span for geom_smooth() on ggplot2 v3.1.0.9000


sabahzero:
Hello,

I too am having issues with the span option of geom_smooth()

I've tried installing the older version, no dice
https://github.com/tidyverse/ggplot2/issues/1292

My data does default to gam, and have also tried setting the method to loess. Still no difference
https://github.com/tidyverse/ggplot2/issues/2870
https://ggplot2.tidyverse.org/reference/geom_smooth.html

Many thanks in advance!

clauswilke:
Please provide a reproducible example following the guidelines laid out here:
https://reprex.tidyverse.org/articles/reprex-dos-and-donts.html

And please create it with the reprex package as explained here:
https://reprex.tidyverse.org/articles/articles/magic-reprex.html

paleolimbot:
I can't seem to replicate this. It is possible that when using a large dataset (like one that would default to `method = "gam"`), changing the `span` parameter might not change the smooth that much, even when setting `method = "loess"`.

``` r
library(ggplot2)
ggplot(mpg, aes(hwy, cty)) +
  geom_smooth(method = "loess", span = 0.75, col = "red") +
  geom_smooth(method = "loess", span = 10, col = "blue") +
  geom_smooth(method = "loess", span = 0.3, col = "purple")
```

![](https://i.imgur.com/KTF86B0.png)

<sup>Created on 2019-05-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

sabahzero:
Hi all,

It seems I'm just now seeing this. Thanks Claus for the quick reply.

Yes, I have about 3000 data points. I figured out a way around it and took a different path, but will update if I encounter an issue again.

Issue #4236, opened 2020-10-14
Why are you re-licensing ggplot2?


slowkow:
@hadley said:

> We are systematically re-licensing tidyverse and r-lib packages to use the MIT license, to make our package licenses as clear and permissive as possible. To do so, we need the approval of all copyright holders, which I have found by reviewing contributions from all all non-RStudio contributors.

https://github.com/tidyverse/ggplot2/issues/4232#issue-719518370

I asked:

> Could I please ask if you might provide some context about the reason for the change of license?

I'm afraid my comment was not visible, so I'm posting a new issue here, where I hope ggplot2 users and contributors might learn more about the plan to re-license. I would like to please ask if someone might be willing to provide some context regarding the reason for the change of license. Is that OK?

ggplot2 is available under the terms of the Free Software Foundation’s GNU General Public License (GPL). https://github.com/tidyverse/ggplot2/blob/master/LICENSE

That's the [same license that the R Project uses](https://www.r-project.org/about.html):

> R is available as Free Software under the terms of the Free Software Foundation’s GNU General Public License in source code form.

For those of you who are not familiar with software licenses, you might consider reading [this recent post](https://plausible.io/blog/open-source-licenses) from a company that re-licensed their code from from the MIT to a newer licensing scheme called GNU Affero General Public License V3 (AGPLv3).

For example, one reason to re-license code from GPL to MIT might be to give permission for a corporation that is interested to take the ggplot2 (and other tidyverse and r-lib) code and use it in closed-source proprietary products.

clauswilke:
Here's my perspective. But first, I'd like to emphasize that I had nothing to do with this decision, and didn't even know until the issues were filed. I also have absolutely no financial interest in any company that could benefit from relicensing.

In theory, the GPL protects you from cheaters that use the free code without contributing back to the community. In practice, this doesn't work, and the people actually hurt by the GPL are other free software developers that have good intention but that for some reason or another need to work in a licensing environment that is incompatible with the GPL.

The problem with the idea that the GPL protects you is that companies that are willing to cheat the system aren't particularly disturbed by violating the GPL. There are plenty of high profile examples, most recently Tesla violating the licensing requirements of the Linux kernel. And these cases never go anywhere. You can sue the company, and it costs you a lot of money, and usually the best-case outcome is that the company just posts a current dump of their code base somewhere, so they're in compliance temporarily, and then they go right at it again violating the license.

[...]

Issue #4231, opened 2020-10-12
Relicensing ggplot2 (1/3)


hadley:
We are systematically re-licensing tidyverse and r-lib packages to use the MIT license, to make our package licenses as clear and permissive as possible. To do so, we need the approval of all copyright holders, which I have found by reviewing contributions from all all non-RStudio [contributors](https://github.com/tidyverse/ggplot2/graphs/contributors). @808sAndBR, @Ahobert, @AliciaSchep, @alistaire47, @andrasberczi, @andrewdolman, @andrewwbutler, @AnneLyng, @annennenne, @Ax3man, @baptiste, @bfgray3, @bjreisman, @botanize, @BrianDiggs, @brianrice2, @brianwdavis, @brodieG, @burchill, @bwiernik, @carywreams, @CateGitau, @cderv, @clairemcwhite, @clauswilke, @connorlewis, @coolbutuseless, @daniel-wells, @DanRuderman, @dariobuttari, @dariyasydykova, @davharris, @davidmorrison, @djmurphy420, @djnavarro, @dkahle, @dpseidel, @dylan-stark, @econandrew, @edzer, @Elesday, @eliocamp, @erocoar, @fabern, @foo-bar-baz-qux, @haleyjeppson, @has2k1, @hbuschme, @hlendway, @hrbrmstr, would you permit us to re-license ggplot2 with the MIT license? If so, please comment "I agree" below.

* [x] [@808sAndBR](https://github.com/tidyverse/ggplot2/commits?author=808sAndBR)  
* [x] [@Ahobert](https://github.com/tidyverse/ggplot2/commits?author=Ahobert)  
* [x] [@AliciaSchep](https://github.com/tidyverse/ggplot2/commits?author=AliciaSchep)  
* [x] [@alistaire47](https://github.com/tidyverse/ggplot2/commits?author=alistaire47)  
* [x] [@andrasberczi](https://github.com/tidyverse/ggplot2/commits?author=andrasberczi)  
* [x] [@andrewdolman](https://github.com/tidyverse/ggplot2/commits?author=andrewdolman)  
* [x] [@andrewwbutler](https://github.com/tidyverse/ggplot2/commits?author=andrewwbutler)  
* [x] [@AnneLyng](https://github.com/tidyverse/ggplot2/commits?author=AnneLyng)  
* [x] [@annennenne](https://github.com/tidyverse/ggplot2/commits?author=annennenne)  
[...]

PR #4261, opened 2020-11-09
Clarify that span is only used for loess in stat_smooth

nalimilan:
It can be surprising that `span` argument has no effect when number of observations is large (cf. https://github.com/tidyverse/ggplot2/issues/3174).

### DIFF
--- a/R/stat-smooth.r
+++ b/R/stat-smooth.r
@@ -25,8 +25,8 @@
 #' @param level Level of confidence interval to use (0.95 by default).
 #' @param span Controls the amount of smoothing for the default loess smoother.
 #'   Smaller numbers produce wigglier lines, larger numbers produce smoother
-#'   lines. Only used with loess, i.e. when `method="loess"`,
-#'   or when `method=NULL` (the default) and there are fewer than 1,000
+#'   lines. Only used with loess, i.e. when `method = "loess"`,
+#'   or when `method = NULL` (the default) and there are fewer than 1,000
 #'   observations.
 #' @param n Number of points at which to evaluate smoother.
 #' @param method.args List of additional arguments passed on to the modelling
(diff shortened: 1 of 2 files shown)


yutannihilation:
You need to edit the original text and then re-generate the docs by `devtools::document()`.

https://github.com/tidyverse/ggplot2/blob/a132727e9de06dc43828b53de0257aa3fb76b33d/R/stat-smooth.r#L26-L28

nalimilan:
Ah of course, sorry!

R/stat-smooth.r
```diff
@@ -25,7 +25,9 @@
 #' @param level Level of confidence interval to use (0.95 by default).
 #' @param span Controls the amount of smoothing for the default loess smoother.
 #'   Smaller numbers produce wigglier lines, larger numbers produce smoother
-#'   lines.
+#'   lines. Only used with loess, i.e. when `method="loess"`,
```
yutannihilation, line 28:
```suggestion
#'   lines. Only used with loess, i.e. when `method = "loess"`,
```

R/stat-smooth.r
```diff
@@ -25,7 +25,9 @@
...
+#'   or when `method=NULL` (the default) and there are fewer than 1,000
```
yutannihilation, line 29:
```suggestion
#'   or when `method = NULL` (the default) and there are fewer than 1,000
```

yutannihilation (commented):
Thanks, almost looks good to me! I added minor comments about spacing.

One more thing, Tidyverse team is trying to change the license of ggplot2 to more permissive one (the MIT license) to minimize the problems and confusions arising from license incompatibilities. If you are fine that your contribution will be merged to ggplot2 and get licensed under MIT, would you mind commenting "I agree" on #4231?

For more details, you can join the discussion on #4236, or wait for their upcoming blog post.

yutannihilation (approved):

yutannihilation:
Thanks!

PR #4265, opened 2020-11-15
Fix `after_scale()` with plot mappings

### LINKED ISSUE #4260: after_scale() in the initial ggplot() call
Consider the `after_scale()` example from `?aes_eval`:

``` r
library(ggplot2) # tidyverse/ggplot2@a132727

  geom_boxplot(aes(colour = class, fill = after_scale(alpha(colour, 0.4))))
```

Moving the colour and fill specification form the geom to the initial `ggplot()` call causes the fill to not be mapped at all:

``` r
ggplot(mpg, aes(class, hwy, colour = class, fill = after_scale(alpha(colour, 0.4)))) +
  geom_boxplot()
```

yutannihilation:
Fix #4260

~~This feels a bit uneasy to repeat `defaults(layer$mapping, plot$mapping)` everywhere, but I don't come up with a nice idea... Any suggestions are welcome.~~ This pull request moves `defaults(layer$mapping, plot$mapping)` to `setup_layer()` so that we don't need to do it in individual methods that refer to the mapping.

``` r
#> Loading ggplot2

ggplot(mpg, aes(class, hwy, colour = class, fill = after_scale(alpha(colour, 0.4)))) +
  geom_boxplot()
```

<sup>Created on 2020-11-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


has2k1:
> This feels a bit uneasy to repeat `defaults(layer$mapping, plot$mapping)` everywhere, but I don't come up with a nice idea... Any suggestions are welcome.

This is against the grain of the mainly functional style in the codebase, but somewhere in a "setup" function of the layer you could resolve the effective mapping for each layer, kind of similar to how a copy of the data is generated for each layer at the start of the plot build.

```R
if (self$inherit.aes) {
   self$mapping <- defaults(self$mapping, plot$mapping)
}
```

yutannihilation:
Thanks, it sounds nice, but I'm not yet sure when is safe to modify the layer...

has2k1:
> Thanks, it sounds nice, but I'm not yet sure when is safe to modify the layer...

Here

https://github.com/tidyverse/ggplot2/blob/a132727e9de06dc43828b53de0257aa3fb76b33d/R/plot-build.r#L34-L35

Just like there is certainty about the data at that point, there is also certainty about the aesthetics.

clauswilke:
I'm not sure I fully understand the problem, but would a fix fit into the `setup_layer()` function?

https://github.com/tidyverse/ggplot2/blob/a132727e9de06dc43828b53de0257aa3fb76b33d/R/layer.r#L205-L207

See e.g. what we do for sf-specific layers:
https://github.com/tidyverse/ggplot2/blob/a132727e9de06dc43828b53de0257aa3fb76b33d/R/layer-sf.R#L35-L71

We could just finalize the layer mapping in that function and then remove this code from `map_statistic()`:
https://github.com/tidyverse/ggplot2/blob/a132727e9de06dc43828b53de0257aa3fb76b33d/R/layer.r#L290-L296

has2k1:
> but would a fix fit into the `setup_layer()` function?

Yes it would. `setup_layer` for `sf` resolves my concern about keeping methods as pure functions.

yutannihilation:
What I'm not confident is, whether there is really no case (including extension packages) when we need to distinguish inherited mappings and layer's original mappings. But, considering we already do tweak the mappings in `setup_layer()`, I now feel it might be okay. Thanks for the suggestion.

yutannihilation:
This isn't a very big problem, but, interestingly, adding these lines in `setup_layer()` breaks `LayerSf`. Since `defaults()` doesn't preserve the class of `uneval`, `$<-.uneval` method won't triggered when `self$mapping$geometry <- sym(geometry_col)`.

thomasp85:
@yutannihilation where are we on this? Your last comment indicates that it is still not done?

yutannihilation:
Ah, sorry I forgot to add a comment, but the concern in my last comment was addressed by this commit. 

https://github.com/tidyverse/ggplot2/pull/4265/commits/08de90d4454a6f6c27f102b7ea281e94390c4e64

So, I think this is ready now (except for the conflict).

thomasp85 (approved):
LGTM

yutannihilation:
Thanks, I'll merge this after confirming the CI passes

Issue #4149, opened 2020-07-28
Setting `options(ggplot2.discrete.fill)` overrides for aes(colour) with bad colourmap


apoorvalal:
I know the option to set colour palette is in its early stages, so wanted to flag an error I noticed.

Looks like setting `ggplot2.discrete.fill` overwrites the colourmap for `colour` (which should be unchanged based on the documentation?) with a cmap that is both ugly and unintelligible (and it hides the legend too for some reason).

```r
``` r
library(ggplot2)
(p1 = ggplot(mpg, aes(class)) + geom_bar(aes(fill = drv)))
```

![](https://i.imgur.com/MgFayHj.png)

``` r
# unrelated plot 
(p2 = ggplot(mtcars, aes(x = wt,  y = mpg)) +
  geom_point(aes(colour = as.factor(cyl))))
```

![](https://i.imgur.com/k2z7IRB.png)

``` r
# this should only apply to p1
options(ggplot2.discrete.fill    = function() scale_fill_viridis_d())
# as it does
p1
```

![](https://i.imgur.com/uJwnRtD.png)

``` r
# but it also mysteriously overrides p2's colourmap 
p2
```

![](https://i.imgur.com/wxdrvUa.png)

I'm running ggplot 3.3.2 in R 3.6.3 on a machine running ubuntu 18.04 LTS.

yutannihilation:
Thanks, confirmed. Overriding is for convenience and it works perfectly when a vector of colors is supplied (or specifying `aesthetics = c("colour", "fill") in the scale function`. But, it also leads us to unexpected behaviours like the one you encountered. here's another bad example. I agree this needs to be improved somehow.

``` r
library(ggplot2)

p <- ggplot(mpg, aes(class)) + geom_bar(aes(fill = drv))

# Correct, and works as expected
withr::with_options(
  list(ggplot2.discrete.fill = function() scale_fill_viridis_d()),
  print(p)
)
```

![](https://i.imgur.com/uCm4bNC.png)

``` r

# Incorrect, and fails with a cryptic error
withr::with_options(
  list(ggplot2.discrete.fill = function() scale_colour_viridis_d()),
  print(p)
)
#> Error: Unknown colour name: f
```

thomasp85:
@yutannihilation do you have any thoughts on how to best approach this?

yutannihilation:
I was thinking if it's possible to check `aesthetics` of the scale and tweak it (with some warning). For example:

``` r
[...]

PR #4277, opened 2020-11-30
Allow scale_{colour/fill}_ordinal() to be configured via new ggplot2.ordinal.{colour/fill} global options. 

### LINKED ISSUE #4148: New ggplot2.discrete.fill option doesn't work with ordered factor columns
I have found what I think is an issue with the new ggplot2.discrete.fill options, although it may be my understanding of what exactly "discrete" means and this may be by design. When using the option it works correctly for character columns, or as in the example factor(char_column)'s, but does not appear to work for already existing factor columns as this uses the default colours instead.

As I said this may be by design in that factors are treated different in plotting and I haven't realised, but this isn't clear in documentation anywhere I looked. I am using ggplot2 version 3.3.2, and R version 3.6.1. 

``` r
library(ggplot2)
okabe <- c("#E69F00", "#56B4E9", "#009E73", "#F0E442", "#0072B2", "#D55E00", "#CC79A7")

cty_by_var <- function(var) {
  ggplot(mpg, aes(cty, colour = {{var}}, fill = {{var}})) +
    geom_density()
}

price_by_var <- function(var) {
  ggplot(diamonds, aes(price, colour = {{var}}, fill = {{var}})) +
    geom_density()
}

price_by_var2 <- function(var) {
  ggplot(diamonds, aes(price, colour = as.character({{var}}), fill = as.character({{var}}))) +
    geom_density()
}

# Works correctly with okabe colours 
withr::with_options(
  list(ggplot2.discrete.fill = okabe),
  print(cty_by_var(class))
)
```

``` r

# Works incorrectly, using the default colours
withr::with_options(
  list(ggplot2.discrete.fill = okabe),
  print(price_by_var(cut))
)
```

``` r

# Works correctly with okabe colours 
withr::with_options(
  list(ggplot2.discrete.fill = okabe),
  print(price_by_var2(cut))
)
```

<sup>Created on 2020-07-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

cpsievert:
Closes #4148. 

Could be seen as a follow-up of sorts to #3833, which added the ability to configure qualitative (i.e., non-ordinal) colourscale defaults. This PR adds a similar ability to configure ordinal colourscale defaults via new `ggplot2.ordinal.{colour/fill}` options:

```r
library(ggplot2)

p <- ggplot(diamonds, aes(price, colour = cut, fill = cut)) + geom_density()
p
```

image

```r
withr::with_options(
  list(ggplot2.ordinal.fill = viridisLite::cividis(10)), p
)
```

image

### DIFF
--- a/R/zxx.r
+++ b/R/zxx.r
@@ -3,7 +3,15 @@
 #' @export
 #' @rdname scale_viridis
 #' @usage NULL
-scale_colour_ordinal <- scale_colour_viridis_d
+scale_colour_ordinal <- function(..., type = getOption("ggplot2.ordinal.colour", getOption("ggplot2.ordinal.fill"))) {
+  type <- type %||% scale_colour_viridis_d
+  if (is.function(type)) {
+    type(...)
+  } else {
+    discrete_scale("colour", "ordinal", ordinal_pal(type), ...)
+  }
+}
+
 
 #' @export
 #' @rdname scale_viridis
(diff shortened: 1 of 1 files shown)


yutannihilation:
Thanks, this basically looks good to me.

thomasp85:
This seems to deviate a bit from what we do the other places, where we don't fall back through first *colour* then *fill* (or vice versa). Any reason for the change?

I'm not against it per se, but I think we need to be consistent

cpsievert:
It's consistent with `scale_{fill/colour}_discrete` (added in #3833).

I don't feel super strongly about it, but seems pretty convenient to set both defaults at once, and probably what you want 90% of the time.

yutannihilation:
> but it does seem pretty convenient to set both defaults at once, and probably what you want 90% of the time.

I agree, but, in the case of specifying a function, it might be undesirable. 

https://github.com/tidyverse/ggplot2/issues/4149#issuecomment-664780922

We need to improve an error message, at least, but probably it's outside of the scope of this pull request.

thomasp85:
@cpsievert i was probably only looking at the continuous implementation. So it seems we have already shipped two different setups😔. Let's keep this in line with discrete as is and think about how we improve on it in the future

thomasp85 (approved):

thomasp85:
Thanks!

PR #4287, opened 2020-12-07
Adds order to guide hash

### LINKED ISSUE #4280: Hash for guide_train should take colours into account
When using [ggnewscale](https://github.com/eliocamp/ggnewscale), legends can disappear if they have the same title and breaks:

``` r
library(ggplot2)
library(ggnewscale)

data <- expand.grid(y = 1:4, x = 1:4)
data$z <- c("a", "b")

ggplot(data, aes(x, y)) +
  geom_tile(data = ~.x[.x$x == 1, ], aes(fill = z)) +
  scale_fill_brewer(palette = 1) +
  new_scale_fill() +
  geom_tile(data = ~.x[.x$x == 2, ], aes(fill = z)) +
  scale_fill_brewer(palette = 2) 
```

What's going on is that ggplot2 is merging both guides because they have the same title and labels. Changing the name of each scale makes them distinct. 

``` r
ggplot(data, aes(x, y)) +
  geom_tile(data = ~.x[.x$x == 1, ], aes(fill = z)) +
  scale_fill_brewer(name = "1", palette = 1) +
  new_scale_fill() +
  geom_tile(data = ~.x[.x$x == 2, ], aes(fill = z)) +
  scale_fill_brewer(name = "2", palette = 2) 
```

The problem is that `guide_train.legend` computes the hash using the title, labels, direction and name: 

https://github.com/tidyverse/ggplot2/blob/12e61dfcf9151c2094536294a8ef42ddffdca9ef/R/guide-legend.r#L229

In the first example, all of those characteristics are the same, and the guides are merged. Ideally, the hash should include information on the colours included in `key[[aes_column_name]]`

This small hack is a small proof of concept of how it would work:

``` r
my_guide <- function(...) {
  guide <- guide_legend(...)
  class(guide) <- c("my_guide", class(guide))
  guide
}

guide_train.my_guide <- function (guide, scale, aesthetic = NULL) {
  guide <- NextMethod("guide_train")
  
  guide$hash <-  with(guide, 
                      digest::digest(list(title, 
                        key$.label, direction, name, key[, 1]  # should be `key[[aes_column_name]]`
                                                 )))
  guide
}

ggplot(data, aes(x, y)) +
  geom_tile(data = ~.x[.x$x == 1, ], aes(fill = z)) +
  scale_fill_brewer(palette = 1, guide = my_guide()) +
  new_scale_fill() +
  geom_tile(data = ~.x[.x$x == 2, ], aes(fill = z)) +
  scale_fill_brewer(palette = 2, guide = my_guide()) 
```

eliocamp:
This is a way to control which guides can be merged together. 

Closes #4280

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,10 @@
 
 * ggplot2 now requires R >= 3.3 (#4247).
 
+* The hash of `guide_legend()` now includes information on the `order` argument. This means
+that guides with different orders will not be merged. It also means that legend order in 
+plots with multiple legends may change. (@eliocamp, #4280)
+
 # ggplot2 3.3.2
 This is a small release focusing on fixing regressions introduced in 3.3.1.
 
--- a/R/guide-legend.r
+++ b/R/guide-legend.r
@@ -226,7 +226,7 @@ guide_train.legend <- function(guide, scale, aesthetic = NULL) {
   guide$key <- key
   guide$hash <- with(
     guide,
-    digest::digest(list(title, key$.label, direction, name))
+    digest::digest(list(title, key$.label, direction, name, order))
   )
   guide
 }
--- a/tests/figs/deps.txt
+++ b/tests/figs/deps.txt
@@ -1,3 +1,3 @@
 - vdiffr-svg-engine: 1.0
-- vdiffr: 0.3.1
+- vdiffr: 0.3.3.9000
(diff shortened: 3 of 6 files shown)


thomasp85:
I'm going to close this as I'm not a fan of how this adds a non-obvious side-effect to the `order` argument... We can continue discussions about a possible alternative approach in #4280

PR #4294, opened 2020-12-09
Compare facet params instead of the whole objects in tests

yutannihilation:
(As probably some of you also got the email about this,) `all.equal()` now compares environments on R-devel. And actually the GitHub Action runner on R-devel kept failing from some days ago. This pull request fixes it by narrowing the comparing from the whole ggproto objects to only the meaningful elements.

Note that, this behavior is consistent with that of `expect_equal()` in testthat 3rd edition, which uses `waldo::compare()` and it has no option to ignore the environment generally. If we stick with 2nd edition, we can use `check.environment = FALSE` as a workaround, but probably we will want to move to 3rd edition at some point, so I didn't choose `check.environment`.

### DIFF
--- a/tests/testthat/test-facet-.r
+++ b/tests/testthat/test-facet-.r
@@ -125,10 +125,10 @@ test_that("facet_grid() accepts vars()", {
   expect_identical(grid$params$rows, quos(foo = foo))
   expect_identical(grid$params$cols, quos(bar = bar))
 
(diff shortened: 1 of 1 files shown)


hadley (approved):
Thanks for doing this!

yutannihilation:
Thanks!

PR #4303, opened 2020-12-27
Fixed typo in map_data() documentation

### LINKED ISSUE #4278: Map_data() documentation has what seems like a typo
This is not a complicated problem. In the reference page for the function `map_data()` ([here](https://ggplot2.tidyverse.org/reference/map_data.html)), the example has the following call:
```r
ggplot(choro, aes(long, lat)) +
  geom_polygon(aes(group = group, fill = assault)) +
  coord_map("albers",  at0 = 45.5, lat1 = 29.5)
}
```
And then later, there is something slightly different.

However, the call `coord_map()` should have `lat0` instead of `at0` as an argument (in both examples):
``` r
ggplot(choro, aes(long, lat)) +
  geom_polygon(aes(group = group, fill = assault)) +
  coord_map("albers",  lat0 = 45.5, lat1 = 29.5)
}
```
It works either way, since the name doesn't seem to matter here, but it's rather confusing.

I hope I'm doing this right. I've never used GitHub before, so maybe I'm wrong about something (maybe I'm filing it in the wrong place)... I just think this should be pointed out somewhere.

heathgilham:
Fix #4278

### DIFF
--- a/R/fortify-map.r
+++ b/R/fortify-map.r
@@ -66,13 +66,13 @@ fortify.map <- function(model, data, ...) {
 #' choro <- choro[order(choro$order), ]
 #' ggplot(choro, aes(long, lat)) +
 #'   geom_polygon(aes(group = group, fill = assault)) +
-#'   coord_map("albers",  at0 = 45.5, lat1 = 29.5)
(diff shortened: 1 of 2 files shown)


karawoo (approved):
Thank you!

PR #4310, opened 2021-01-07
Toggle the logic of skipping visual tests

### LINKED ISSUE #4293: Use vdiffr conditionally
I also got this email as I'm the maintainer of a package that uses vdiffr. It says we need to fix this until 2021-01-12. On master, the use of vdiffr is already made conditional by #4192, so it's just patching this commit to the released version.

> Suggested packages should be used conditionally: see §1.1.3.1 of
'Writing R Extensions'.  Some of the requirements of vdiffr are hard to
install on a platform without X11 such as M1 Macs: see the logs at
https://www.stats.ox.ac.uk/pub/bdr/M1mac/.

> In some cases there are other suggested packages not used conditionally:
you can check all of them by setting environment variable
`_R_CHECK_DEPENDS_ONLY_=true`  -- see
https://cran.r-project.org/doc/manuals/r-devel/R-ints.html#Tools .

> Please correct ASAP and before 2021-01-12 to safely retain the package
on CRAN.

@thomasp85 @hadley 
Are you already planning a patch release for this? Please let me know if there's anything I can help.

yutannihilation:
Fix #4293

At the time of implementing #4192, I was afraid that visual tests would be silently ignored on the CI (and we would oversight some bugs), so I chose the following logic:

* When vdiffr is not available, do raise an error unless `VDIFFR_RUN_TESTS` envvar is explicitly set to *`false`*

but this ends up failure on CRAN checks. So, this pull request toggle the logic to the following for safety. This should fix the CRAN checks.

* When vdiffr is not available, do NOT raise an error unless `VDIFFR_RUN_TESTS` envvar is explicitly set to *`true`*

### DIFF
--- a/tests/testthat/helper-vdiffr.R
+++ b/tests/testthat/helper-vdiffr.R
@@ -1,12 +1,12 @@
-# vdiffr ignores failures when
-#   - VDIFFR_RUN_TESTS is "false" (on Travis CI with older versions and dev version of R)
-#   - CI is not set (on CRAN)
+# By default, if vdiffr is not installed, all visual tests are skipped unless
+# VDIFFR_RUN_TESTS is explicitly set to "true", which should be the case only on
+# a GitHub Actions CI runner with stable version of R.
 
 if (requireNamespace("vdiffr", quietly = TRUE)) {
   expect_doppelganger <- vdiffr::expect_doppelganger
 } else {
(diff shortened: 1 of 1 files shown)


yutannihilation:
CI failures are irrelevant to this pull request. #4311will update the visual cases to fix these.

clauswilke (approved):

Issue #4324, opened 2021-01-24
`geom_contour` documentation is misleading


dmurdoch:
The documentation for `geom_contour` says:

> Contouring
> tends to work best when \code{x} and \code{y} form a (roughly) evenly
> spaced grid.

but in fact, as issue #4320 and its discussion describes, there are much more stringent requirements. Exactly what the requirements are is a little unclear to me; the code to do it is in `isoband_z_matrix()` here:  https://github.com/tidyverse/ggplot2/blob/9741da5050f81b7b5c012c56d02f45fc93d68f89/R/stat-contour.r#L224  It's something like "Contouring requires the `(x, y)` pairs to be a large subset of the grid of all possible pairs of unique values of `x` and `y`, with no repeats."  

It would also be helpful to add an example using `akima::interp()`, the usual way to produce the grid of values from irregular data.

clauswilke:
As I said in #4320, the true condition is that you have a matrix of z values whose locations can be identified with marginal vectors of x and y values. It's Ok for some z values to be missing, though, as long as there are corresponding `NA`s in the matrix of z values.

I agree the documentation should spell this out clearly.

dmurdoch:
I think your description is too strict.  For example, this works:
``` r
# First, start with a complete grid
df <- expand.grid(x=1:10, y=1:10)
df$z <- with(df, x^2 + y^2)

# Take the subset of lower z values
df <- subset(df, z < 100)

# Shuffle it
df <- df[sample(nrow(df)),]

head(df)
#>    x y  z
#> 37 7 4 65
#> 42 2 5 29
#> 52 2 6 40
#> 34 4 4 32
#> 7  7 1 50
#> 1  1 1  2
library(ggplot2)
ggplot(df, aes(x, y, z=z)) + geom_contour()
```

![](https://i.imgur.com/0ba944G.png)

<sup>Created on 2021-01-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

clauswilke:
I'm pretty sure what's happening is that ggplot2 expands this input into a matrix of z values whose locations can be identified with marginal vectors of x and y values, replacing missing values with `NA`. Because that's the input that the isoband algorithm requires. I would know, I wrote it. :-)

[...]

PR #4311, opened 2021-01-07
Update vdiffr cases for sf 0.9-7 and fix forbidden "brew cask"

yutannihilation:
It seems sf 0.9-7 (published on CRAN yesterday) introduced a small change on axis labels. This pull request is to update the cases.

### Before
before

### After
after

---
edit: This pull request also fixes the following error on macOS runner due to the recent change on Homebrew

```
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
Error: Process completed with exit code 1.
```

PR #4325, opened 2021-01-25
Improve docs for geom_contour as discussed in issue #4324

dmurdoch:


dmurdoch:
The failures I'm seeing in the tests don't seem to be related to the changes I made.  I did add an example to `?geom_contour`; would that require some change to reference figures?

yutannihilation:
Yes, we can ignore those failures; they are due to a slight change introduced in the recent version of sf (c.f. #4311). Sorry for the confusion.

yutannihilation (approved):
Looks good!

@clauswilke 
I think this is fine to merge. This is only a Suggest dependency and is just for documentation so we can remove this easily if we found any problem in future. Any concerns?

clauswilke:
No concerns about the external dependency. I haven't carefully checked the various CI failures.

yutannihilation:
Thanks, for R-CMD-check, failures of windows-latest and ubuntu-16.04 is due to visual cases, which #4311 should fix. macOS-latest errors on setting up XQuartz, I guess this is a temporary one. For test-coverage, it fails at the same XQuartz setup. None seems caused by this pull request.

yutannihilation:
Thanks for contributing!

Issue #4357, opened 2021-03-02
Poor compatibility with R 4.1 pipe |>


bwiernik:
The planned native R pipe `|>` in R 4.1 currently interacts poorly with `ggplot()`. `|>` prohibits any symbols with special parsing rules on the RHS of the pipe, including `+`, `(`, and `{`. That makes it really hard to work with `\>` and `ggplot()` in a way that feels natural. See reprex below for some approaches that fail and work. 

Enabling a more natural workflow would require either an update to the ggplot2 API to make it pipeable or a change in the `|>` parsing rules to permit, for example, `(` or `{` on the RHS of `|>`. Do you have some connections with the R development team to raise this issue? 

```r
stopifnot(R.version$major == "4" & R.version$minor == "1.0")
library(ggplot2)

# Drop-in replacement of %>% with |>
## Doesn't work
mtcars |>
  ggplot() + 
      aes(x = as.factor(cyl), y = mpg, color = as.factor(cyl)) +
      geom_jitter(height = 0, width = .2) + 
      stat_summary(fun.data = mean_se, size = 1, color = "black") +
      theme_classic()

# Wrapping ggplot() chain in parentheses
## Doesn't work
mtcars |>
  (ggplot() + 
      aes(x = as.factor(cyl), y = mpg, color = as.factor(cyl)) +
      geom_jitter(height = 0, width = .2) + 
      stat_summary(fun.data = mean_se, size = 1, color = "black") +
      theme_classic()
  )

# Wrapping ggplot() chain in squiggly brackets
## Doesn't work
mtcars |>
  {ggplot() + 
      aes(x = as.factor(cyl), y = mpg, color = as.factor(cyl)) +
      geom_jitter(height = 0, width = .2) + 
      stat_summary(fun.data = mean_se, size = 1, color = "black") +
      theme_classic()
  }

# Wrap pipe chain and initial ggplot() in parentheses
## Works, but confusing
(mtcars |>
  ggplot()) + 
      aes(x = as.factor(cyl), y = mpg, color = as.factor(cyl)) +
      geom_jitter(height = 0, width = .2) + 
      stat_summary(fun.data = mean_se, size = 1, color = "black") +
      theme_classic()

# Write ggplot call as an anonymous function
## Works, but confusing
mtcars |>
  dplyr::mutate(cyl = as.factor(cyl)) |>
  (\(d) ggplot(d) + 
      aes(x = cyl, y = mpg, color = cyl) +
[...]

PR #4359, opened 2021-03-02
Allow compatibility with the base pipe for R 4.1

MyKo101:
Currently, the base pipe is incompatible with ggplot due to the precedence of the syntax surrounding `|>`. Any specially parsed operators (such as `+`) are not permitted on the right hand side of the pipe operator. Since the `%+%` is parsed _after_ the `|>` is parsed, it can be exploited.

It now acts as a ternary function when in a pipeline, the first argument is the piped object (presumably a data.frame), the second argument is all of chain of gg function calls before the last one, and the third argument is the final gg function call.

If the third argument is missing, then the function acts just as the standard `+` operator (using `+.gg()`), meaning it is backwards compatible with prior code and can act as a drop-in when users want to use `|>` instead of `%>%`.

Note that since the `|>` does not use the `.` substitute as the `%>% does, the behaviour of inserting as the _first_ argument (usually in a `ggplot()` call) is consistent.

Examples:

Does not work:
```
mpg |>
  ggplot() +
  aes(displ,hw) +
  geom_point() +
  geom_path()
```

Does work:
```
mpg |>
  ggplot() %+%
  aes(displ,hw) %+%
  geom_point() %+%
  geom_path()
```

Parsed as:
```
`%+%`(
  mpg, 
  ggplot() %+% aes(displ,hw) %+% geom_point(), 
  geom_path()
)
```


yutannihilation:
I don't think this makes sense. In my understanding, ggplot2 had never advertised the usage with pipe, and, personally, I don't feel

``` r
mpg |>
  ggplot() %+%
  aes(displ,hw) %+%
  geom_point() %+%
  geom_path()
```

is more elegant than

``` r
d <- mpg

ggplot(d) +
  aes(displ,hw) +
  geom_point() +
  geom_path()
```

That said, I agree it would be nice if there's some workaround (though I'm afraid there's no such things). Let's continue the discussion on #4357.

PR #4360, opened 2021-03-02
Export `add()` as alias for `+`

bwiernik:
This facilitates ggplot2 usage with the R 4.1 native pipe `|>`, which prohibits `+` on the RHS.

```r
diamonds |>
  ggplot(diamonds, aes(carat, price)) |>
  add(geom_point())
```

Also enables a workflow involving the magrittr pipe:

```r
diamonds %>%
  ggplot(diamonds, aes(carat, price)) %>%
  add(geom_point())
```

cf. https://github.com/tidyverse/ggplot2/issues/4357
cf. https://github.com/tidyverse/ggplot2/pull/4359

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,8 @@
 
 * ggplot2 now requires R >= 3.3 (#4247).
 
+* Export `add()` as a alias for `+`, to facilitate ggplot2 usage with the native R pipe `|>` (@bmwiernik, #4357)
+
 # ggplot2 3.3.3
 This is a small patch release mainly intended to address changes in R and CRAN.
 It further changes the licensing model of ggplot2 to an MIT license.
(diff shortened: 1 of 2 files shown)


yutannihilation:
I don't think this makes sense. In my understanding, ggplot2 had never advertised the usage with pipe, and users can import `magrittr` to use `add()` anytime if they want.

That said, I agree it would be nice if there's some workaround (though I'm afraid there's  no such things). Let's continue the discussion on #4357.

bwiernik:
Just to be clear, ggplot() can be used at the end of a magrittr pipe chain. It can't be used at the end of a base R pipe chain.

yutannihilation:
Yes, I understand what you mean, and actually I myself sometimes uses `%>%`. Still, I don't think it's the "official" usages. None of the code in ggplot2's documentation or in ggplot2-book uses it in that way.

Anyway I believe we can agree at least that we need discussion before jumping into creating a pull request directly, considering there's two different solutions proposed in just a single day :)

PR #4363, opened 2021-03-03
Fix scale_color_discrete documentation link

jashapiro:
When using ggplot2, I noticed that the documentation for `scale_color_discrete()` was pointing to `scale_colour_hue()` rather than `scale_colour_discrete()`.

This PR updates the `@rdname` link to the correct parent function.


yutannihilation (approved):
Good catch, thanks!

PR #4375, opened 2021-03-17
Use rlang::check_installed()

### LINKED ISSUE #4344: Better error message when using coord_map() without {maproj} installed
It seems that the {mapproj} package is required for `coord_map()` to work. The documentation for the function does talk about {mapproj} but doesn't explicitly say that the package is required.

Using the example from the function documentation this is what I get:

```
library(ggplot2)
library(maps)
nzmap <- ggplot(map_data("nz"), aes(x = long, y = lat, group = group)) +
  geom_polygon(fill = "white", colour = "black")
nzmap + coord_map()
``` 

`Error in loadNamespace(name) : there is no package called ‘mapproj‘`

When running this function in isolation it's easy enough to understand I need to install the package. However when this is part of a large codebase that runs on one machine and then is attempted to run on another machine, the error has no indication of where it came from. When uploading a shiny app to shinyapps.io, shinyapps did not pick up that it needs to install {mapproj}, and looking at the error log did not make it clear where in my app {mapproj} is needed.

### LINKED ISSUE #4370: Using rlang::check_installed()
rlang 0.4.10 introduced `check_installed()` to both check that a suggested package is installed and, if not, offer to the user to do so for them. I have found this to be very ergonomic so far and suspect it would be a boon to R newcomers to use this function in ggplot2 (after having some students trip on hexbin just this week). 

I'd be happy to prepare a PR. It would require a bump in the rlang dependency, so I'm not sure if that's an issue for you all.

malcolmbarrett:
This PR replaces various `requireNamespace()` + `abort()` checks with the new rlang function `check_installed()`, which offers to install the suggested package then proceeds or aborts. This PR also bumps the rlang dependency to 0.4.10

Screen Shot 2021-03-17 at 3 23 00 PM

I didn't use the `reason` argument in any `check_installed()` calls, which can add a bit of context. I'd be happy to add that, if desired.

Closes #4370, fixes #4344

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Imports:
     isoband,
     MASS,
     mgcv,
-    rlang (>= 0.3.0),
+    rlang (>= 0.4.10),
     scales (>= 0.5.0),
     stats,
     tibble,
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,10 @@
 
 * ggplot2 now requires R >= 3.3 (#4247).
 
+* ggplot2 now uses `rlang::check_installed()` to check if a suggested package is
+  installed, which will offer to install the package before continuing (#4375, 
+  @malcolmbarrett)
+
 # ggplot2 3.3.3
 This is a small patch release mainly intended to address changes in R and CRAN.
 It further changes the licensing model of ggplot2 to an MIT license.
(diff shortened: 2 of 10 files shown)


yutannihilation:
Thanks, I think we can use `reason` argument on where ggplot2 already uses `try_require()`, which contains the reason.

malcolmbarrett:
Yeah, I agree. Added! It now looks like this, depending on the function:

Screen Shot 2021-03-17 at 7 17 05 PM

yutannihilation (approved):
Thanks, looks good! Do you think this needs a news item?

malcolmbarrett:
✔️ Added to news

Thanks! 🙏

PR #4384, opened 2021-03-21
Fix blank strip text with multilayered strips

### LINKED ISSUE #4050: strip.text.y = element_blank() yields an error when there are multiple layers of strip labels
Hello,
I was trying to remove the strip labels. I found that setting `strip.text.x = element_blank()` works when there is only one layer of strip labels and yields an error when there are multiple layers of labels. Is there any way to remove the strip labels in the latter case? Many thanks!

```r
# This works:
p1 <- ggplot(mtcars, aes(x = mpg, y = wt)) + geom_point() +
 facet_grid(vs + am ~ gear)  + theme(strip.text.x = element_blank())

#This does not work:
p2 <- ggplot(mtcars, aes(x = mpg, y = wt)) + geom_point() +
 facet_grid(vs + am ~ gear)  + theme(strip.text.y = element_blank())
#Error in gtable_add_grob(panel_table, strips$y$right, panel_pos_rows$t,  : 
#  Not all inputs have either length 1 or same length same as 'grobs'
```

teunbrand:
This PR aims to fix #4050, wherein an error occurs in `facet_grid()` when the strip text is a blank element and there are two variables determining the strips. This also caused an inelegant warning in `facet_wrap()` that should now also be remedied.
Instead of returning an `n` by `m` matrix of `zeroGrob`s, the `assemble_strips()` internal function now returns an `n`-length list of `zeroGrob`s regardless of `m`. This then satisfies the `FacetWrap/Grid$draw_panel()` methods' assumption that the length of the strip-grobs is equal to the number of panels (for `facet_wrap()`) or columns/rows (for `facet_grid()`).

The following should now work without warnings or errors:

``` r
library(ggplot2)

df <- data.frame(
  x = 1, y = 1, a = "A", b = c("B", "C")
)

p <- ggplot(df, aes(x, y)) +
  geom_point() +
  theme(strip.text = element_blank())

# Was warning before, now doesn't warn
p + facet_wrap(~ a + b)

# Was error before, now runs fine
p + facet_grid(~ a + b)
```


yutannihilation (approved):
Looks good, thanks!

Could you add a news item? As at least two users faced this problem, I think this is worth a news.

teunbrand:
Yes, sure thing. Thanks for approving!

yutannihilation:
Thanks!

PR #4332, opened 2021-01-28
 Update geom-text.r

larspijnappel:
As proposed, here's a PR for the documentation improvement regarding the unsupported `check_overlap` argument for `geom_label()` (#4331).

If this is not sufficient, please advise me on the necessary improvements, as this is my first contribution.

Issue #4331, opened 2021-01-28
check_overlap argument not accepted in geom_label()


larspijnappel:
The `geom_label()` is not accepting the `check_overlap` argument, although it appears to be a valid option according to the `geom_label` documentation (`ggplot2` version 3.3.3):

> Arguments `check_overlap` 	
If `TRUE`, text that overlaps previous text in the same layer will not be plotted. `check_overlap` happens at draw time and in the order of the data. Therefore data should be arranged by the label column before calling __`geom_label()`__ or __`geom_text()`__.

In another section of this documentation, there's no explicit reference that this specific argument is not supported:
> geom_label()
Currently `geom_label()` does not support the `angle` aesthetic and is considerably slower than `geom_text()`. The `fill` aesthetic controls the background colour of the label.

#### `geom_label()`
The plot is generated with a warning:
```r
mpg %>% 
  ggplot(aes(displ, hwy)) +
  geom_label(aes(label = class), check_overlap = T)
#> Warning: Ignoring unknown parameters: check_overlap
```
![image](https://user-images.githubusercontent.com/12656725/106171413-cb116880-6191-11eb-8b00-ff021558659b.png)

#### `geom_text()`
* `geom_text()` default behavior: overlapping text
```r
library(tidyverse)

mpg %>% 
  ggplot(aes(displ, hwy)) +
  geom_text(aes(label = class))
```
![image](https://user-images.githubusercontent.com/12656725/106170959-4e7e8a00-6191-11eb-9e7a-755572f46c72.png)

* `geom_text()` with `check_overlap = TRUE`: overlapping text removed
```r
mpg %>% 
  ggplot(aes(displ, hwy)) +
  geom_text(aes(label = class), check_overlap = TRUE)
```
![image](https://user-images.githubusercontent.com/12656725/106170800-22fb9f80-6191-11eb-9881-a2667057dd52.png)

clauswilke:
Yes, the documentation is incorrect. Feel free to submit a PR to fix this.

yutannihilation:
Fixed by #4389

PR #4389, opened 2021-03-25
Update geom-text.r

larspijnappel:
As proposed, here's a PR for the documentation improvement regarding the unsupported `check_overlap` argument for `geom_label()` (#4331).

Note: this request replaces the previous PR #4332 as it can't be completed due to loss of the previous fork.


thomasp85:
Can I get you to run roxygen to also update the docs

larspijnappel:
No problem, although that would be another first time for me as well :-)

So, in my local branch I performed `devtools::document()` and the `man\geom_text.Rd` has been updated successfully. 

Do I just `pr_push()` again to complete all of it?

thomasp85:
Yes - every change can be added to the PR using `pr_push()`

NAMESPACE
```diff
@@ -489,6 +489,7 @@ export(scale_color_fermenter)
 export(scale_color_gradient)
 export(scale_color_gradient2)
 export(scale_color_gradientn)
+export(scale_color_gray)
```
thomasp85, line 492: Can I get you to remove the two new exports in the NAMESPACE... It is not your fault, but another PR did not get oxygenised and this is now causing your PR to fail check. Removing these lets us handle it in a different PR

NAMESPACE
```diff
@@ -536,6 +537,7 @@ export(scale_fill_fermenter)
 export(scale_fill_gradient)
 export(scale_fill_gradient2)
 export(scale_fill_gradientn)
+export(scale_fill_gray)
```
thomasp85, line 540: same as above

thomasp85 (changes requested):

thomasp85 (approved):
Thanks!

PR #4403, opened 2021-04-08
Make sure jitter is only calculated once

### LINKED ISSUE #2941: Vertical line and crossbar jitter being applied differently.
I have this issue - basically the different components of the geom_errorbar have jitter being applied to them differently. I posted this on stack-exchange and someone there seemed to think it might be a bug.

```
library(tidyverse)
```
Basic data set:
```
df <- structure(list(
Test = c("A", "B", "C", "D", "A", "C", "D"), 
mean = c(1, 100.793684, 1, 1, 51.615601, 1, 2.456456), 
sd = c(1, 2.045985, 1, 1, 4.790053, 1, 4.250668), 
lower = c(2, 102.839669, 2, 2, 56.405654, 2, 6.707124), 
upper = c(0, 98.747699, 0, 0, 46.825548, 0, -1.79421)), 
row.names = c(NA, -7L), class = c("tbl_df", "tbl", "data.frame"))
```
Now the code I am using:

```
subplot <- ggplot(df, aes(x = Test, y = mean)) +
  geom_point(aes(x= Test, y = mean), 
                position = position_jitter(width = 0.2, height = 0.2))+
  geom_errorbar(aes(ymin = lower, ymax = upper),
                width = 0.1,
                position = position_jitter(width = 0.2, height = 0.2)) 
subplot
```
mine_ex

Thanks.

thomasp85:
Fix #2941 

This PR changes the jitter position implementation to calculate a jitter once for `x` and `y` and then apply that to all position aesthetics

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Fix a bug in `position_jitter()` where different jitters would be applied to 
+  different position aesthetics of the same axis (@thomasp85, #2941)
+
 * `coord_sf()` now has an argument `default_crs` that specifies the coordinate
   reference system (CRS) for non-sf layers and scale/coord limits. This argument
   defaults to the World Geodetic System 1984 (WGS84), which means x and y positions
--- a/R/position-jitter.r
+++ b/R/position-jitter.r
@@ -77,9 +77,9 @@ PositionJitter <- ggproto("PositionJitter", Position,
     trans_y <- if (params$height > 0) function(x) jitter(x, amount = params$height)
 
     # Make sure x and y jitter is only calculated once for all position aesthetics
-    x_aes <- intersect(names(data), ggplot_global$x_aes)
+    x_aes <- intersect(ggplot_global$x_aes, names(data))
     x <- if (length(x_aes) == 0) 0 else data[[x_aes[1]]]
-    y_aes <- intersect(names(data), ggplot_global$y_aes)
+    y_aes <- intersect(ggplot_global$y_aes, names(data))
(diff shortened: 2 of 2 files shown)


clauswilke:
I'm not entirely sure yet this works in all edge cases. I believe `jitter()` will look at the data to consider how much jitter to apply. This gets overwritten with the `amount` argument, but if `amount` is `NULL` or 0 then `jitter()` again looks at the data. At a minimum this requires a warning in the release notes, I believe.

A safer approach might be to use some actual x or y data columns as dummy data, if available, and then subtract them from the dummy data before applying the jitter. Get's very circuitous, though.

clauswilke:
Looks like a long time ago I came up with a different way to solve this issue. See if you like this approach:
https://github.com/clauswilke/ggplot2/blob/ed9be70c14e67e5448ac9f11aadce2cfb868764d/R/position-jitter.r#L75-L91

thomasp85:
Yeah, I saw that. I still prefer to calculate one jitter value per dimension and apply that. You are right that we need to extract real values for it but that is trivial

clauswilke (approved):
Looks good, but I would flip the order in the `intersect()` command to get `x` if available.

``` r
intersect(c("xmin", "x", "y"), c("x", "xmin"))[1]
#> [1] "xmin"
intersect(c("x", "xmin"), c("xmin", "x", "y"))[1]
#> [1] "x"
```

<sup>Created on 2021-04-09 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

thomasp85:
Ah - good catch... I thought I was doing the right way🙈

PR #4404, opened 2021-04-08
FAQ vignettes for ggplot2

mine-cetinkaya-rundel:
These were created based on an analysis of the most commonly asked questions about ggplot2 on SO. A summary of these can be found [here](https://connect.rstudioservices.com/tidyverse-so) (requires login).

A few notes:

- If using a package other than ggplot2 in the answers, I've indicated the package name with `::`.
- There is a minimal amount of custom CSS on top fo the FAQ pages that ensure better spacing for readability.
- I think we should add alt text to the images but I'd like to do that after a round of review to avoid rewriting them.

There are some questions that were commonly asked that I didn't include. Happy to add these in if we think it would be good to do so (and I might ask for guidance on most elegant solutions for some).

- Questions not included because I'm not sure of the most elegant solution:
  - [How can I set the background color of facet labels based on another variable?](https://stackoverflow.com/questions/19440069/ggplot2-facet-wrap-strip-color-based-on-variable-in-data-set)
  - [How can I show the x and y axis on all panels in a faceted plot?](https://stackoverflow.com/questions/22116518/add-x-and-y-axis-to-all-facet-wrap)
  - [How can I place a legend for each facet?](https://stackoverflow.com/questions/14840542/place-a-legend-for-each-facet-wrap-grid-in-ggplot2)
  - [How can I add floating axis labels in facet_wrap plot](https://stackoverflow.com/questions/13297155/add-floating-axis-labels-in-facet-wrap-plot) -- this seems to already work and, to me, placing them at the bottom doesn't look very good
- Questions that I wasn't sure which FAQ page to place in and didn't seem worth creating an individual page for
  - [How can I use ggplot2 in a function](https://stackoverflow.com/questions/10004847/making-plot-functions-with-ggplot-and-aes-string) -- I think this is well covered in the vignette on [using ggplot2 in packages](https://stackoverflow.com/questions/10004847/making-plot-functions-with-ggplot-and-aes-string)
  - [How can I change the whisker definition in a box plot?](https://stackoverflow.com/questions/4765482/changing-whisker-definition-in-geom-boxplot)
- Question I thought was out of scope but could easily add to barplots FAQ if we decide it's in scope:
  - [How can I make an animated, racing bar plot?](https://stackoverflow.com/questions/53162821/animated-sorted-bar-chart-with-bars-overtaking-each-other)


vignettes/faq-reordering.Rmd
```diff
@@ -0,0 +1,231 @@
...
+The following bar plot shows the number of cars that fall into each `class` category.
+Classes are ordered alphabetically.
+You might prefer them to be ordered by the number of cars in each class.
+
+```{r}
+ggplot(mpg, aes(y = class)) +
+  geom_bar()
+```
+
+To do this, you should first calculate the number of cars that fall into each `class` category.
```
clauswilke, line 58: There's various ways to avoid manually calculating the sums, and in my experience (teaching a class on this right now) students constantly confuse `geom_bar()` and `geom_col()` so I would try to avoid switching right away.

```r
ggplot(mpg, aes(y = fct_infreq(class))) +
  geom_bar()

ggplot(mpg, aes(y = fct_rev(fct_infreq(class)))) +
  geom_bar()

ggplot(mpg, aes(y = fct_reorder(class, class, length))) +
  geom_bar()
```

mine-cetinkaya-rundel: Oh interesting. I've had the opposite experience, but might be a function of teaching dplyr and ggplot2 together (already doing lots of `count()` so they expect to use it). I'll update to `fct_infreq()` since it's basically designed for that, I think the cognitive load on `fct_reorder(class, class, length)` is even higher.

yutannihilation:
These packages need to be included in Suggests.

```
Error: Error: processing vignette 'faq-axes.Rmd' failed with diagnostics:

there is no package called 'ggtext'
```

```
Error: Error: processing vignette 'faq-bars.Rmd' failed with diagnostics:

there is no package called 'tidyr'
```

```
Error: Error: processing vignette 'faq-reordering.Rmd' failed with diagnostics:

there is no package called 'forcats'
```

thomasp85:
I'm not sure we want them in Suggests, to be honest... @mine-cetinkaya-rundel is it imperative that this is a vignette that is distributed with thee package, or could it be an article on the ggplot2 site?

mine-cetinkaya-rundel:
I think this is up to you -- also mainly why I left this as a draft PR for now, was going to discuss with @hadley as well and then flip the switch.

My thoughts:

- I think these will be most helpful for people googling such questions, so articles on ggplot2 site could be sufficient.
- If distributed with the package they might be helpful to those who are not always working online, and hence not able to google/SO things as easily. But I imagine this is not a large proportion of users.
- So I'm not sure of the burden of additional dependencies (even if in Suggests only) vs. the potential benefit of offline access.
- I tried to be selective in non ggplot2 packages I use -- they're either tidyverse or those that I think we can reasonably be confident will be kept in good shape.

Let me know if you have a preference one way or another and I'm happy to update as such. If going the article on site only route, I'd love a pointer on how we indicate that in the package (a link to another package that uses that would be sufficient).

clauswilke:
I frequently find it helpful to put vignettes into the `.Rbuildignore`. In this way, they're in the github repo, they show up in the pkgdown site, but they're excluded from the R check and from CRAN. I think that's a good compromise. I would assume people are much more likely to look for documentation materials on the package website than on CRAN. (If I google "ggplot2" the first hit I get is ggplot2.tidyverse.org, so as long as the article is there I'd find it.)

hadley:
Or you can use `use_article()` to put them in a subdirectory — we'd still need to declare these packages somewhere though, so that pkgdown can have them installed them.

mine-cetinkaya-rundel:
I've made the FAQs into articles and added a `Config/Needs/Website` field to `DESCRIPTION` for the additional packages needed to build the articles

thomasp85 (approved):

thomasp85:
Thanks @mine-cetinkaya-rundel !

PR #4299, opened 2020-12-17
Add missing US spellings for “scale_grey”

klmr:
As per #1901: [“gray” is the predominant US spelling](https://www.merriam-webster.com/dictionary/gray). The purpose of this PR is to increase discoverability for en-US speakers as well as non-native English speakers whose first exposure is to US English spelling rather than UK/AU/NZ.

This PR keeps the existing, less common spelling `scale_color_grey` for backwards compatibility.

Issue #1901, opened 2016-11-13
scale_color_gray


sfeuerriegel:
ggplot2 usually supports commands in both American and British English. I came across one exception: "scale_color_gray" that appears to be missing.

http://docs.ggplot2.org/0.9.2.1/scale_grey.html#

hadley:
This seems like a good idea, but unfortunately it's not high enough on my priority list that I will implement it myself. However, I'd be happy to review a PR.

PR #4407, opened 2021-04-09
Add last grey/gray spelling combination

thomasp85:
This is not really the best, and I'm open to alternatives.

ggplot2 originally had `scale_colour_grey()`, and `scale_color_grey()`, with the former mixing up English/American spelling of colour and grey. A recent PR added `scale_color_gray()` which makes the tests fail as it looks for whether all English colour scales have an American equivalent. This PR adds `scale_colour_gray()` which fixes the test, but leaves a sour taste in my mouth. Dream scenario would be to only have `scale_colour_grey()` and `scale_color_gray()` as those names would be internally consistent, but I don't think we can remove the original mixed one as it would cause major breakage...

The alternative is to change the tests to remove one of the versions but that seems almost just as bad...

Thoughts


hadley:
Why don't we just remove `scale_color_gray()`?

clauswilke:
Counter-argument: I can never remember whether it's grey or gray and use them interchangingly.

yutannihilation:
I too can't remember it, but always use tab completion so it's better if there are fewer candidates. (But I have no strong opinion here)

clauswilke (approved):
I'm approving this but don't have a strong opinion one way or another.

thomasp85:
@hadley, it was recently added because it was lacking and US spelling users requested it

yutannihilation:
@hadley 
A bit more context is that you indicated that adding `scale_color_gray` is "a good idea," which aren't implemented just because we don't have the development bandwidth. This was the reasoning of the PR (#4299).

> This seems like a good idea, but unfortunately it's not high enough on my priority list that I will implement it myself. However, I'd be happy to review a PR.
https://github.com/tidyverse/ggplot2/issues/1901#issuecomment-274978940

hadley:
@yutannihilation that's just my standard response. I wouldn't read anything into it. 

I think reverting #4299 is the simplest fix — `scale_color_gray()` (no matter how you spell it) is relatively rarely used, and the spelling difference is small enough that I don't think it's really causing any problems.

yutannihilation:
I see... Yeah, I personally agree with reverting it.

clauswilke:
Fine with reverting also. Maybe I'll finally learn the correct spelling. :-)

hadley:
Its A for America (grAy) and E for England (grEy) 😄

clauswilke:
And what do they use in Australia?

thomasp85:
ok, I'll revert instead

PR #4411, opened 2021-04-12
Revert "Add missing US spellings for “scale_grey” (#4299)"

thomasp85:
This reverts commit a8352072a048249224f6e517e817e6ac72477c47.

As discussed, adding the additional spelling is more trouble than it is worth so this PR reverts it


clauswilke (approved):
Had this made it into a release already? If yes there should be a comment in the changelog. If no then it's all good.

thomasp85:
No this got merged in two weeks ago

PR #4412, opened 2021-04-12
Reflect more options to viridis_pal()

yutannihilation:
(Depends on https://github.com/r-lib/scales/pull/306)

The dev version of viridisLite has 3 more options, so we need to update the document once it's released. This pull request might be a bit too early and I'll reflect the released version of the document. The main purpose of this is to notify here about the very fact there will be such additions.

https://sjmgarnier.github.io/viridisLite/reference/viridis.html

### DIFF
--- a/man/scale_viridis.Rd
+++ b/man/scale_viridis.Rd
@@ -99,16 +99,26 @@ labels and so forth.}
 \item{alpha}{The alpha transparency, a number in [0,1], see argument alpha in
 \code{\link[grDevices]{hsv}}.}
 
-\item{begin}{The (corrected) hue in [0,1] at which the viridis colormap begins.}
+\item{begin}{The (corrected) hue in [0,1] at which the color map begins.}
 
-\item{end}{The (corrected) hue in [0,1] at which the viridis colormap ends.}
+\item{end}{The (corrected) hue in [0,1] at which the color map ends.}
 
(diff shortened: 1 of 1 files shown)


thomasp85:
can you make sure that the generated docs are up to date with viridis and then this is good to go

thomasp85 (approved):

yutannihilation:
This is up to date with viridisLite package, but ahead of scales package. As our document refers to `scales::viridis_pal()`, not directly to viridisLite's functions, we need to wait for scales to be updated (i.e. r-lib/scales#306 gets merged) and released. Otherwise, the changes in this pull request will be just reverted when someone run `devtools::document()` with the released version of scales.

thomasp85:
I see... thanks

z3tt:
Ah, looking through the issues first would have saved me lots of time while trying to trace down the documentation and why it is not up2date even though the options are mentioned in the viridisLite package.

Any chance this is going to be updated soon?

yutannihilation:
I'm afraid no chance, because the version of the scales package is not released on CRAN yet.

hadley:
Are we still waiting for a scales release?

yutannihilation:
Yes.

hadley:
Ok, I'll add that to my to do list.

yutannihilation:
Thanks for releasing the new version of scales! This should be ready to merge now.

PR #4413, opened 2021-04-12
Move seed setup in `position_jitter()` to `setup_params()` from the constructor

### LINKED ISSUE #2507: `position = "jitter"` and `position = position_jitter()` have different behavior
I noticed today that `position = "jitter"` is not the same as `position = position_jitter()`, and the behavior described in the documentation for `position = position_jitter()` cannot be obtained for `position = "jitter"`. The underlying issue is a ggproto behavior that has caught me off guard many times (see below).

First the reprex. We start with a simple plot with jitter:
```{r}
df <- data.frame(x = c(1, 2, 1, 2),
                 y = c(1, 1, 2, 2))

set.seed(1234)
ggplot(df, aes(x, y)) + geom_point(position = "jitter")
```
screen shot 2018-04-01 at 12 11 41 am

The following code produces different jitter:
```{r}
set.seed(1234)
ggplot(df, aes(x, y)) + geom_point(position = position_jitter())
```
screen shot 2018-04-01 at 12 11 51 am

However, this reproduces the jitter from the first example:
```{r}
set.seed(1234)
ggplot(df, aes(x, y)) + geom_point(position = position_jitter(seed = NULL))
``` 
screen shot 2018-04-01 at 12 11 58 am

The problem is that `position = "jitter"` does not use the default arguments defined in `position_jitter()`, and therefore `seed` is set to `NULL`, not to `NA`, when we use `position = "jitter"`. This same problem happens generally with ggproto, e.g. when we call a geom with `geom_*()` vs. `geom = "*"`, and it has bitten me many times in different scenarios. 

There are two ways to fix this particular problem:
1. The simple way: Don't set `seed = NA` as default in `position_jitter()`.

2. The complicated way: Fix the ggproto behavior so the default arguments apply when objects are specified by their name.

thomasp85:
Fix #2507 

This PR implements @paleolimbot suggested fix. I think it is brittle to rely on calling the constructor, because it is also valid (but weird) to pass in PositionJitter directly

I think we should aim for all ggproto objects to be self-sufficient, i.e. have next to no logic in the constructors

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Make sure position_jitter creates the same jittering independent of whether it
+  is called by name or with constructor (@thomasp85, #2507)
+
 * Fix a bug in `position_dodge2()` where `NA` values in thee data would cause an
   error (@thomasp85, #2905)
 
--- a/R/position-jitter.r
+++ b/R/position-jitter.r
@@ -46,10 +46,6 @@
 #'   geom_point(position = jitter) +
 #'   geom_point(position = jitter, color = "red", aes(am + 0.2, vs + 0.2))
 position_jitter <- function(width = NULL, height = NULL, seed = NA) {
-  if (!is.null(seed) && is.na(seed)) {
-    seed <- sample.int(.Machine$integer.max, 1L)
-  }
-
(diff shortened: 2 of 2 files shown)


clauswilke (approved):
Looks good to me.

PR #4416, opened 2021-04-13
move width computation to `setup_params()`

### LINKED ISSUE #2047: `geom_bar` inconsistently handling date values
Description
===========

There are unexpected plot results when specifying the fill aesthetic in `geom_bar` when the `x` aesthetic is a Date or POSIXct value. Below I have listed examples using Date and POSIXct objects, respectively. Similar data represented with these two classes results in rather different plots, see below.

Date Examples
=============

The following examples use Date objects produced with `make_date()`.

1 month with 2 fill values
--------------------------

In this example January fill values are TRUE and FALSE, February and March fill values are only FALSE.

``` r
library(ggplot2)
library(lubridate)

d1 <- data.frame(
  dates = rep(make_date(year = 2017, month = 1:3), each = 2),
  highlight = c(TRUE, FALSE, FALSE, FALSE, FALSE, FALSE)
)

# this unexpectedly changes the width of one of January's bars
ggplot(d1, aes(x = dates)) +
  geom_bar(aes(fill = highlight))
```

    ## Warning: position_stack requires non-overlapping x intervals

unnamed-chunk-1-1

1 month with distinct fill value
--------------------------------

In this example January fill values are only TRUE, both February and March are only FALSE.

``` r
library(ggplot2)
library(lubridate)

d2 <- data.frame(
  dates = make_date(year = 2017, month = 1:3),
  highlight = c(TRUE, FALSE, FALSE)
)

# only one bar for January, but as above it is thin
ggplot(d2, aes(x = dates)) +
  geom_bar(aes(fill = highlight))
```

unnamed-chunk-2-1

3+ fill values
--------------

This example introduces a third fill value to the highlight column.

``` r
library(ggplot2)
library(lubridate)

d3 <- data.frame(
  dates = make_date(year = 2017, month = 1:3),
  highlight = c(1, 2, 3)
)

# all bars are now equally thin, additional breaks added to x-axis
ggplot(d3, aes(x = dates)) +
  geom_bar(aes(fill = factor(highlight)))
```

unnamed-chunk-3-1

POSIXct Examples
================

The following examples use POSIXct objects produced with `make_datetime()`.

1 month with distinct fill value
--------------------------------

In this example, when January fill is only FALSE, February and March are only TRUE, the January bar does not show. The bar may be too thin to see.

``` r
library(ggplot2)
library(lubridate)

d4 <- data.frame(
  datetimes = make_datetime(year = 2017, month = 1:3),
  highlight = c(TRUE, FALSE, FALSE)
)

# January bar is no longer thin, instead missing 
ggplot(d4, aes(x = datetimes)) +
  geom_bar(aes(fill = highlight))
```

unnamed-chunk-4-1

3+ fill values
--------------

Similar the 3+ fill value example above, however in this example the bars are not equally thin, they are all missing.

``` r
library(ggplot2)
library(lubridate)

d5 <- data.frame(
  datetimes = make_datetime(year = 2017, month = 1:3),
  highlight = c(1, 2, 3)
)

# no bars
ggplot(d5, aes(x = datetimes)) +
  geom_bar(aes(fill = factor(highlight)))
```

unnamed-chunk-5-1

Conclusion
==========

The `geom_bar` fill aesthetic is not properly handling Date and POSIXct objects. I am not weighing in yet on whether bars ought to be thin or wide, rather I am hoping to iron out `geom_bar` so Date and POSIXct values are handled consistently. I hope I have not misunderstood how `geom_bar` is intended to handle date values.

I will look under the hood to try and identify the problem. For now I am not sure of a work around.

thomasp85:
Fix #2047

This PR makes sure that width is calculated globally to avoid surprise behaviour outlined in #2047. There is a chance that this will break someones code, but I think it is pretty low


hadley:
Is this computation done per facet?

thomasp85:
No, globally. I'm not married to either but it seems we usually do these kind of computations globally

hadley (approved):
This seems safe to me, but I don't have strong sense of what the consequences are. But I can't imagine that you want different bar widths in different groups or facets.

clauswilke:
I think the case to check is a faceted plot where there are different numbers of bars and axes are set to free. I can't think of a dataset off the top of my head to make such a plot, but I have definitely encountered them. The current behavior is to adjust the bar width to fill up the available space. I'm not sure if that's good or bad, but we should at least check it. Such plots certainly do exist.

If I have some time later today I'll try to make an example.

smouksassi:
``` r
library(ggplot2)
ggplot(mtcars,aes(cyl,fill=as.factor(vs)))+
   geom_bar()+
  facet_grid(~gear,scales="free" )
#> Warning: position_stack requires non-overlapping x intervals

#> Warning: position_stack requires non-overlapping x intervals
```



<sup>Created on 2021-04-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

clauswilke:
I meant something like this. Will it change? Also, notice how `geom_bar()` and `geom_col()` produce the same output. That should probably be retained, whatever the output is.

I have no strong opinion on whether bar widths should be the same or not in this type of a plot, I just wanted to point out that I've seen plots like this one, more than once.

``` r
library(tidyverse)
library(nycflights13)

d1 <- flights %>% 
  select(carrier) %>% 
  mutate(group = 1)

d2 <- flights %>%
  select(carrier) %>%
  mutate(
    carrier = as.character(fct_lump_n(carrier, n = 10)),
    group = 2
  )

d3 <- flights %>%
  select(carrier) %>%
  mutate(
    carrier = as.character(fct_lump_n(carrier, n = 5)),
    group = 3
  )

d4 <- flights %>%
  select(carrier) %>%
  mutate(
    carrier = as.character(fct_lump_n(carrier, n = 2)),
    group = 4
  )

d <- rbind(d1, d2, d3, d4)

d %>%
  ggplot(aes(carrier)) + 
  geom_bar() + 
  facet_wrap(~group, scales = "free_x")
```



``` r
d %>%
  count(carrier, group) %>%
  ggplot(aes(carrier, n)) + 
  geom_col() + 
  facet_wrap(~group, scales = "free_x")
```



<sup>Created on 2021-04-13 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

thomasp85:
@clauswilke the output doesn't change because the width is encoded in the scale domain, i.e. a width of 0.9 seems bigger in a panel with fewer categories
image

thomasp85:
For completeness, this is @smouksassi's example rendered with this patch 
image

PR #4418, opened 2021-04-14
Make sure that expressions are reencoded as expressions when collapsed

### LINKED ISSUE #4084: Combined facet labels aren't parsed when .multi_line = FALSE
(previously: https://github.com/tidyverse/ggplot2/pull/4032#issuecomment-643531350)

Facet labels that are combined with `.multi_line = FALSE` don't show up as parsed when using `label_parsed`:

``` r
library("ggplot2")

dat <- data.frame(x = 1, y = 1, a = "a", b = 1)

ggplot(dat, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ a + b,
    labeller = labeller(
      a = c("a" = "a^2"),
      .default = label_parsed,
      .multi_line = FALSE
    )
  )
```

<sup>Created on 2020-06-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Compare to `.multi_line = TRUE`, which looks as it should:

``` r
ggplot(dat, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ a + b,
    labeller = labeller(
      a = c("a" = "a^2"),
      .default = label_parsed,
      .multi_line = TRUE
    )
  )
```

One issue is the `.multi_line` argument doesn't get passed along here: https://github.com/tidyverse/ggplot2/blob/7d05fa3b6cda40fe645fe04090485275a0666979/R/labeller.r#L440

However changing that line alone makes things worse:

``` r
ggplot(dat, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ a + b,
    labeller = labeller(
      a = c("a" = "a^2"),
      .default = label_parsed,
      .multi_line = FALSE
    )
  )
```

thomasp85:
Fix #4084 

There is one "issue" in that parsing is either, or. Right now it will re-add parsing if *any* label is an expression which could potentially lead to errors if there is a mix of expressions and not (I'm not sure if that is possible) and the non-expressions cannot be parsed...

Alternatively we could only do this if *all* labels are expressions...

``` r
library("ggplot2")

dat <- data.frame(x = 1, y = 1, a = "a", b = 1)

ggplot(dat, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ a + b,
    labeller = labeller(
      a = c("a" = "a^2"),
      .default = label_parsed,
      .multi_line = FALSE
    )
  )
```

<sup>Created on 2021-04-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Fix bug in `labeller()` where parsing was turned off if `.multiline = FALSE`
+  (@thomasp85, #4084)
+
 * Fix bug in `geom_dotplot()` where dots would be positioned wrong with 
   `stackgroups = TRUE` (@thomasp85, #1745)
 
--- a/R/labeller.r
+++ b/R/labeller.r
@@ -91,7 +91,7 @@ collapse_labels_lines <- function(labels) {
   is_exp <- vapply(labels, function(l) length(l) > 0 && is.expression(l[[1]]), logical(1))
   out <- do.call("Map", c(list(paste, sep = ", "), labels))
   label <- list(unname(unlist(out)))
-  if (any(is_exp)) {
+  if (all(is_exp)) {
     label <- lapply(label, function(l) list(parse(text = paste0("list(", l, ")"))))
   }
   label
--- a/R/labeller.r
+++ b/R/labeller.r
@@ -88,8 +88,13 @@
 NULL
 
 collapse_labels_lines <- function(labels) {
+  is_exp <- vapply(labels, function(l) length(l) > 0 && is.expression(l[[1]]), logical(1))
   out <- do.call("Map", c(list(paste, sep = ", "), labels))
-  list(unname(unlist(out)))
+  label <- list(unname(unlist(out)))
+  if (any(is_exp)) {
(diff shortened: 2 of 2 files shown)


thomasp85:
I decided to switch to `all()` instead of `any()` as it seems the safest

karawoo (approved):
This seems good to me. It is possible to have a mix of expressions and non-expressions (below). But at that point the label is the combined text of both the labels, so if they aren't all expressions then I think it's a reasonable limitation to not try to parse them.

``` r
library("ggplot2")

dat <- data.frame(
  x = 1,
  y = 1,
  fruit = "apple",
  veg = "carrot",
  stringsAsFactors = TRUE
)

ggplot(dat, aes(x, y)) +
  geom_point() +
  facet_grid(
    ~ fruit + veg,
    labeller = labeller(
      fruit = c("apple" = "a^2"),
      veg = label_both,
      .default = label_parsed,
      .multi_line = FALSE
    )
  )
```



<sup>Created on 2021-04-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

thomasp85:
yeah, it is not possible to mix it without a very big rewrite of the strip rendering system which I don't think is warranted at this point

PR #4419, opened 2021-04-14
Only apply limits if they are not NA in plot

### LINKED ISSUE #4027: default xlim and ylim in qplot are not supportd when non missing
See below. I think it's against the principles laid out in the tidyverse principle book, it can be fixed by using `c(NA_real_, NA_real_)` as a default. 

The issue arose as I was writing a wrapper around `qplot` and it choked when passing  the parameters `xlim = xlim, ylim = ylim`.

``` r
library(ggplot2)

# works
qplot(Sepal.Width, Sepal.Length, data = iris)
```

``` r

# doesn't work
qplot(Sepal.Width, Sepal.Length, data = iris, xlim = c(NA,NA))
#> Error in UseMethod("limits"): no applicable method for 'limits' applied to an object of class "logical"

# works
qplot(Sepal.Width, Sepal.Length, data = iris, xlim = c(NA_real_,NA_real_))
```

<sup>Created on 2020-05-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Fix #4027 

This PR simply catches the case where the user supplies `c(NA, NA)` as limits and ignores it

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Fix a bug in `qplot()` when supplying `c(NA, NA)` as axis limits 
+  (@thomasp85, #4027)
+
 * Fix bug in `geom_dotplot()` where dots would be positioned wrong with 
   `stackgroups = TRUE` (@thomasp85, #1745)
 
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good to me

PR #4420, opened 2021-04-14
Add example showing how to use the fill argument in stat_function()

Bisaloo:
Output:

image

I believe this new example is useful because it might not be straightforward how to use the `fill` aesthetic with `geom_function()`/`stat_function()`.

My first try was to simply add `fill = "blue"` but it doesn't work because the `geom_line()` doesn't understand `fill`.

### DIFF
--- a/R/geom-function.R
+++ b/R/geom-function.R
@@ -32,6 +32,8 @@
 #'
 #' base + stat_function(fun = dnorm, geom = "point", n = 20)
 #'
+#' base + stat_function(fun = dnorm, geom = "density", color = "blue", fill = "blue", alpha = 0.5)
+#'
 #' base + geom_function(fun = dnorm, n = 20)
 #'
 #' # Two functions on the same plot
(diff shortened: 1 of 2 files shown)


R/geom-function.R
```diff
@@ -32,6 +32,8 @@
 #'
 #' base + stat_function(fun = dnorm, geom = "point", n = 20)
 #'
+#' base + stat_function(fun = dnorm, geom = "density", color = "blue", fill = "blue", alpha = 0.5)
```
thomasp85, line 35: any reason why not using `geom = "polygon"` - it seems more general...

Bisaloo: No, no specific reason. `geom = "density"` seemed a little bit more intuitive since the example is about plotting a probability density function. But I have no strong feelings on this.

thomasp85: I think we should change it

Bisaloo: Done

thomasp85 (approved):

thomasp85:
Thanks!

PR #4422, opened 2021-04-15
Extent the range of `geom_abline()` outside the panel range

### LINKED ISSUE #4024: Can geom_abline() draw lines a bit longer than the actual Coord range?
Someone told me `geom_abline()` doesn't look nice when we increase the `size`. To be fair, `geom_function()` seems better at the first example, but it has the same problem when the range of the scale is shorter or equal to the coord's range. I think this is intentional, but it might be nice if there's some option to draw lines to outside of the plot area.

(Btw, maybe this is a good chance to replace `GeomAbline` with `GeomFunction`...?)

``` r
library(ggplot2)
library(tibble)
library(patchwork)

p <- ggplot(tibble(x = c(-5, 5), y = c(-4, 6))) +
  geom_line(aes(x, y), size = 10, color = "#cc0000") +
  coord_fixed(xlim = c(-1, 1), ylim = c(-1, 1), expand = FALSE) +
  theme_bw()

p1 <- p +
  geom_abline(slope = 1, intercept = 1, size = 10, color = "#000088", alpha = 0.7) +
  ggtitle("abline")

p2 <- p +
  geom_function(fun = ~ .x + 1, size = 10, color = "#000088", alpha = 0.7) +
  ggtitle("function")

p1 * p2
```

<sup>Created on 2020-05-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

``` r
library(ggplot2)

ggplot() +
  geom_function(fun = ~ .x + 1, size = 10, color = "#000088", alpha = 0.7) +
  xlim(-1, 1) +
  ggtitle("function") + 
  coord_fixed(xlim = c(-1, 1), ylim = c(-1, 1), expand = FALSE) +
  theme_bw()
```

<sup>Created on 2020-05-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

thomasp85:
Fix #4024

This PR only fixes the issue in `geom_abline()` leaving `geom_function()` as it is because it is not really feasible to fix it right now/for this

### DIFF
--- a/R/geom-abline.r
+++ b/R/geom-abline.r
@@ -127,7 +127,7 @@ GeomAbline <- ggproto("GeomAbline", Geom,
   draw_panel = function(data, panel_params, coord) {
     ranges <- coord$backtransform_range(panel_params)
 
-    if (T || coord$clip == "on" && coord$is_linear()) {
+    if (coord$clip == "on" && coord$is_linear()) {
       # Ensure the line extends well outside the panel to avoid visible line
       # ending for thick lines
       ranges$x <- ranges$x + c(-1, 1) * diff(ranges$x)
--- a/R/geom-abline.r
+++ b/R/geom-abline.r
@@ -127,6 +127,12 @@ GeomAbline <- ggproto("GeomAbline", Geom,
   draw_panel = function(data, panel_params, coord) {
     ranges <- coord$backtransform_range(panel_params)
 
+    if (T || coord$clip == "on" && coord$is_linear()) {
+      # Ensure the line extends well outside the panel to avoid visible line
+      # ending for thick lines
+      ranges$x <- ranges$x + c(-1, 1) * diff(ranges$x)
+    }
+
     data$x    <- ranges$x[1]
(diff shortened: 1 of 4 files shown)


R/geom-abline.r
```diff
@@ -127,6 +127,12 @@ GeomAbline <- ggproto("GeomAbline", Geom,
   draw_panel = function(data, panel_params, coord) {
     ranges <- coord$backtransform_range(panel_params)
 
+    if (T || coord$clip == "on" && coord$is_linear()) {
```
yutannihilation, line 130: What's the intention of this `T`...?

thomasp85: 😵 Thank god for you reviewing it

yutannihilation: Ah, for debugging. I see.

yutannihilation:
Looks good. Could you update visual cases?

thomasp85:
yes, of course :-)

yutannihilation (approved):

PR #4425, opened 2021-04-16
[CI] Switch to pak

### LINKED ISSUE #4424: pkgdown CI error
https://github.com/tidyverse/ggplot2/runs/2360094189#step:9:464
```
-- RMarkdown error -------------------------------------------------------------
Quitting from lines 372-382 (faq-axes.Rmd) 
Error: Error in loadNamespace(name) : there is no package called 'ggtext'
--------------------------------------------------------------------------------
Error: Error: Failed to render RMarkdown
```

yutannihilation:
Fix #4424

Test build: https://github.com/yutannihilation/ggplot2/runs/2361972393

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -22,50 +22,57 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: '4.0',   vdiffr: true }
-          - {os: macOS-latest,   r: '4.0',   vdiffr: true }
-          - {os: ubuntu-18.04,   r: 'devel', vdiffr: false}
-          - {os: ubuntu-18.04,   r: '4.0',   vdiffr: true,    rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.6',   vdiffr: false,   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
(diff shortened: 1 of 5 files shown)


thomasp85:
Thanks - there's a new GHA that uses pak and does all of this automatically, so we'll be migrating to that instead

yutannihilation:
Hmm, yeah, using pak seems better. This pull request still fails to build the pkgdown site...

```
ERROR: dependencies ‘gtable’, ‘isoband’, ‘scales’ are not available for package ‘ggplot2’
```

yutannihilation:
Switched to pak and this seems to work.

https://github.com/yutannihilation/ggplot2/runs/2362084213

thomasp85:
Since you are as far as I am and have a better feeling for the CI scripts do you mind updating all the scripts to use pak, i.e. run `usethis::use_tidy_github_actions()` with the dev version of usethis and make sure to not remove stuff we have added?

yutannihilation:
Oh, sure. I will.

yutannihilation:
It seems `pak::pkg_deps()` doesn't work when there's unavailable packages. We need to fix pak first...

yutannihilation:
It turned out pak is not ready to use in our repo. I'll think if we can add some (probably not very simple) workaround or give up for now.

thomasp85:
Ok, let's just fix the pkgdown build for now then... will you roll back the last commits?

yutannihilation:
Reverted. I'll file a new issue for migrating `R-CMD-check.yaml` to pak when it's ready.

thomasp85 (approved):

Issue #4103, opened 2020-06-30
Accept tidyverse tilde functions


davidsjoberg:
It would be neat to be able to use the same "short" version of defining function as in `purrr` in `ggplot2` as well.

---

Example would be to be able to write:

```r
ggplot(mtcars, aes(drat, qsec)) + 
  geom_line() +
  scale_y_continuous(labels = ~format(., nsmall = 1))
```

Today you need to do like this:

```r
ggplot(mtcars, aes(drat, qsec)) + 
  geom_line() +
  scale_y_continuous(labels = function(x) format(x, nsmall = 1))
```

Thanks! :)

yutannihilation:
Yes, I think all the arguments that accepts functions should accept tilde notation. It's just we don't have enough development bandwidth to do it. Pull requests are welcome!

davidsjoberg:
Which parameters accept functions?

yutannihilation:
This now works. Probably fixed by #4188.

``` r
library(ggplot2)

ggplot(mtcars, aes(drat, qsec)) + 
  geom_line() +
  scale_y_continuous(labels = ~format(., nsmall = 1))
```

![](https://i.imgur.com/HO7TCpx.png)

<sup>Created on 2022-07-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #4427, opened 2021-04-17
purrr style lambda notation in scale arguments

teunbrand:
This PR aims to enable purrr style lambda notation for scale arguments that accept functions (`limits`, `breaks`, `minor_breaks`, `labels`, `rescaler` and `oob`). It does so by converting formulas to functions in the scale constructors. Thomas mentioned that lambda notation should work in https://github.com/tidyverse/ggplot2/issues/4173#issuecomment-683742992, but the notation is currently not enabled. 

This PR has partial overlap in scope and in functions altered with #4188. Notable differences are that this PR doesn't touch facet labellers, but applies the lambda notation to more arguments and in all three types of scales (continuous, binned and discrete).

Thank you for considering!


teunbrand:
I came across issue #4103 as well, which seemed related and worth mentioning.

yutannihilation (approved):
Looks great!

@thomasp85 
Can we have this in 3.3.4?

yutannihilation:
@thomasp85
I'm merging this if there's no concern. This supersedes some part of #4188, and as #4188 is in the milestones for v3.3.4, I think it's good to include this in v3.3.4.

thomasp85:
Sorry for lack of response, go ahead and merge

yutannihilation:
It's fine. Thanks for confirming!

PR #4431, opened 2021-04-19
Make annotate_logticks coord_flip-aware

### LINKED ISSUE #3954: conflict of `coord_flip()` with `annotation_logticks()`
This behaves exactly like the closed issue #881, still here in ggplot2 version 3.3.0

Using a bit modified [example](https://ggplot2.tidyverse.org/reference/annotation_logticks.html)

```r
a <- ggplot(msleep, aes(bodywt, brainwt)) +
 geom_point(na.rm = TRUE) +
 scale_x_log10(
   breaks = scales::trans_breaks("log10", function(x) 10^x),
   labels = scales::trans_format("log10", scales::math_format(10^.x))
 ) +
 scale_y_log10(
   breaks = scales::trans_breaks("log10", function(x) 10^x),
   labels = scales::trans_format("log10", scales::math_format(10^.x))
 )
```
This works:
```r
a + coord_flip()
```
And this works:
```r
a + annotation_logticks()
```
 Now,
```r
a + coord_flip() + annotation_logticks()
```

returns error: `Error in unit(xticks$x, "native") : 'x' and 'units' must have length > 0`

thomasp85:
Fix #3954 

This PR ensures that `annotate_logticks()` works in the presence of `coord_flip()`. It does so by flipping around the ranges and aesthetic names.

Important notice: With this PR, `sides = "b"` will not get translated to `sides = "l"`. Since the `sides` argument is positional and not tied to aesthetic names it makes sense to me to leave it alone

### DIFF
--- a/R/annotation-logticks.r
+++ b/R/annotation-logticks.r
@@ -126,12 +126,14 @@ GeomLogticks <- ggproto("GeomLogticks", Geom,
                         mid = unit(0.2, "cm"), long = unit(0.3, "cm"))
   {
     ticks <- list()
+    flipped <- inherits(coord, "CoordFlip")
+    x_name <- if (flipped) "y" else "x"
+    y_name <- if (flipped) "x" else "y"
 
     # Convert these units to numbers so that they can be put in data frames
     short <- convertUnit(short, "cm", valueOnly = TRUE)
     mid   <- convertUnit(mid,   "cm", valueOnly = TRUE)
     long  <- convertUnit(long,  "cm", valueOnly = TRUE)
-
 
     if (grepl("[b|t]", sides)) {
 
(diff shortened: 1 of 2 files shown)


R/annotation-logticks.r
```diff
@@ -126,13 +126,15 @@ GeomLogticks <- ggproto("GeomLogticks", Geom,
                         mid = unit(0.2, "cm"), long = unit(0.3, "cm"))
   {
     ticks <- list()
+    flipped <- inherits(coord, "CoordFlip")
```
yutannihilation, line 129: This feels a bit too specific to this problem. Are there any other way to check if the coord is flipped? (I guess the answer is no, but let me just confirm...)

thomasp85: The answer is no :-)... We have these annoying specific logic pertaining to coord_flip around the code base and there is not really much to do about it at this time

We can discuss a more throughout fix at a later stage but that is beyond this fix

yutannihilation: I see, thanks for clarifying.

yutannihilation (approved):

PR #4433, opened 2021-04-19
Create the requested number of bins in `bin2d_breaks()`

### LINKED ISSUE #3824: stat_summary_bin gives 1 bin more than specified.
stat_summary_bin() consistently gives one more bin than specified in the function call

```r
ggplot(data=mtcars, aes(y=mpg,x=wt)) +
  stat_summary_bin(fun.y = "mean", geom = "point",
                   bins = 3)
```

This shows 4 points. If you change bins to 4, it'll show 5 points and so on.

thomasp85:
Fix #3824 

This PR fixes the reported bug by removing the last bin if it lies completely outside the the scale range

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Fix a bug in `stat_summary_bin()` where one more than the requested number of
+  bins would be created (@thomasp85, #3824)
+
 * Fix bug in `guide_coloursteps()` that would repeat the terminal bins if the
   breaks coincided with the limits of the scale (@thomasp85, #4019)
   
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good.

``` r
#> ℹ Loading ggplot2

x <- seq(0, 1, length = 1e4)
y <- x + rnorm(length(x))
dt <- data.frame(x, y)

# NOT OK
ggplot(dt, aes(x, y)) +
  geom_point(colour = "grey80") +
  stat_summary_bin(fun = mean, bins = 10, geom = "point", colour = "red")
```



<sup>Created on 2021-04-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

PR #4439, opened 2021-04-21
Remove warning when creating secondary axis with no minor breaks

### LINKED ISSUE #4368: scale_y_continuous() 'sec.axis' and Warning min(x) max(x)
In a ggplot I am using a secondary Y-axis to display additionnal labels to my plot.
I use it like this :
```
scale_y_continuous(
      expand = c(0, 0),
      breaks = seq(0, N.rows, length.out = n.grad),
      labels = function(x) format(x, digits = 2, scientific = TRUE),
      sec.axis = sec_axis(
        trans = ~.,
        breaks = dt.data[right.y.hiden == FALSE]$right.y.label.pos,
        labels = dt.data[right.y.hiden == FALSE]$sample.string))
```
in `sec_axis()` breaks and labels are queried in a data.table `dt.data`:

- If multiple breaks and labels are selected from the data.table -> no problems.
- If there is only 1 break and 1 label queried from the data.table, displaying the plot returns this warning (I guess this is the matching translation from the language I use):
```
Warning messages:
1: In min(x) : no non-missing arguments to min; returning Inf
2: In max(x) : no non-missing arguments to max; returning -Inf
```
The plot returned is fine : it totally matches my expectation.
But I don't understand why these warnings are printed.

So I basically have to questions:
1. Why those warnings are printed when there is only 1 break & label ?
2. Is there something I should mention in `sec_axis()` or elsewhere to get rid of these warnings ?

I don't know if it is reproducible, but from what I experienced, if you give just a single value to `breaks` and `labels` in `sec_axis()` such has:
`sec.axis = sec_axis(trans = ~., breaks = 125, labels = "awesome single label")`
This should trigger the warnings.

thomasp85:
Fix #4368 

This PR implements the fix proposed by @yutannihilation in the issue

### DIFF
--- a/R/axis-secondary.R
+++ b/R/axis-secondary.R
@@ -213,30 +213,36 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
       range_info <- temp_scale$break_info()
 
       # Map the break values back to their correct position on the primary scale
-      old_val <- lapply(range_info$major_source, function(x) which.min(abs(full_range - x)))
-      old_val <- old_range[unlist(old_val)]
-      old_val_trans <- scale$trans$transform(old_val)
+      if (!is.null(range_info$major_source)) {
+        old_val <- lapply(range_info$major_source, function(x) which.min(abs(full_range - x)))
+        old_val <- old_range[unlist(old_val)]
+        old_val_trans <- scale$trans$transform(old_val)
 
-      old_val_minor <- lapply(range_info$minor_source, function(x) which.min(abs(full_range - x)))
-      old_val_minor <- old_range[unlist(old_val_minor)]
-      old_val_minor_trans <- scale$trans$transform(old_val_minor)
+        # rescale values from 0 to 1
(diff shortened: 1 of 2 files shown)


R/axis-secondary.R
```diff
@@ -213,30 +213,36 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
...
-      old_val_minor_trans <- scale$trans$transform(old_val_minor)
+        # rescale values from 0 to 1
+        range_info$major[] <- round(
+          rescale(
+            scale$map(old_val_trans, range(old_val_trans)),
+            from = range
+          ),
+          digits = 3
+        )
+      }
```
yutannihilation, line 229: Maybe we need a similar `else` branch here for `old_val_trans`?

thomasp85: weird - I was sure I made it both places

yutannihilation (approved):
Looks good

PR #4441, opened 2021-04-22
Make sure guide_axis can modify labels in coord_flip

### LINKED ISSUE #4289: dup_axis guide parameters don't apply to the right axis with coord_flip
If you use `dup_axis()` to duplicate an axis and don't want to repeat the title of the axis, then flip the coordinates with `coord_flip` you have the "new" x axis title that disappears and not the y axis one (as expected). If you change the guide argument to something like `angle = 45` then it correctly applies to the duplicated axis.

Sorry if the problem is not explained very well, here is a reprex demonstrating it.

``` r
library(ggplot2)

mtcars %>% 
  ggplot(aes(x = mpg, y = cyl)) +
  geom_point() +
  scale_x_continuous(sec.axis = dup_axis(guide = guide_axis(title = NULL))) +
  coord_flip()
```

thomasp85:
Fix #4289 

This PR changes the order of calls in `labels()` to make sure that any guide_axis parts gets correctly applied

### DIFF
--- a/R/coord-flip.r
+++ b/R/coord-flip.r
@@ -62,7 +62,7 @@ CoordFlip <- ggproto("CoordFlip", CoordCartesian,
   },
 
   labels = function(labels, panel_params) {
-    flip_axis_labels(CoordCartesian$labels(labels, panel_params))
+    CoordCartesian$labels(flip_axis_labels(labels), panel_params)
   },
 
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
An elegant minimal solution to this! Looks great to me.

PR #4442, opened 2021-04-22
Ensure all keys in `guide_bins(..., reverse = TRUE)`

### LINKED ISSUE #4210: Top key missing in guide_bins when reverse = TRUE
When `guide_bins()` is used with the `reverse` param set to `TRUE`, the key for the top bin is missing (and it seems the entire scale is shifted down in the key).

Tested on current dev `ggplot2`.

```r
library(ggplot2)
df <- data.frame(a = 1:100, b = 1:100)

ggplot(df) +
  geom_point(aes(colour = b, x = a, y = b)) +
  guides(colour = guide_bins(reverse = T))
```
26d9d09b-64e1-4a34-b666-be1f3fe000cb

thomasp85:
Fix #4210 

This fixes a bug when reversing the bin guide because the terminal row that held irrelevant `NA` values ended up as the first row

### DIFF
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -144,6 +144,7 @@ guide_train.bins <- function(guide, scale, aesthetic = NULL) {
 
   if (is.numeric(breaks)) {
     limits <- scale$get_limits()
+    breaks <- breaks[!breaks %in% limits]
     all_breaks <- c(limits[1], breaks, limits[2])
     bin_at <- all_breaks[-1] - diff(all_breaks) / 2
   } else {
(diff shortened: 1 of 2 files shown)


R/guide-bins.R
```diff
@@ -163,7 +164,12 @@ guide_train.bins <- function(guide, scale, aesthetic = NULL) {
   key$.label <- scale$get_labels(all_breaks)
   guide$show.limits <- guide$show.limits %||% scale$show_limits %||% FALSE
 
-  if (guide$reverse) key <- key[nrow(key):1, ]
+  if (guide$reverse) {
+    key <- key[rev(seq_len(nrow(key))), ]
+    # Move last row back to last
+    aesthetics <- setdiff(names(key), ".label")
+    key[, aesthetics] <- key[c(seq_len(nrow(key))[-1], 1), aesthetics]
+  }
```
yutannihilation, line 172: Can we reverse the bins before appending `NA` at the bottom?

``` r
  bin_at <- scale$map(bin_at)
  if (guide$reverse) {
    bin_at <- rev(bin_at)
  }
  key <- new_data_frame(setNames(list(c(bin_at, NA)), aes_column_name))
  key$.label <- scale$get_labels(all_breaks)
```

thomasp85: The issue is that `bin_at` and the derivatives gets calculated at two different branches so we would have to duplicate the logic there

yutannihilation: > `bin_at` and the derivatives gets calculated at two different branches

I meant something like this, not inside the `if` branches. 

https://github.com/tidyverse/ggplot2/compare/master...yutannihilation:poc/pr-4442

Since I needed some minutes to figure out what the added code does, I was hoping to make the logic simple, but it seems my attempt is no shorter than this pull request... Please ignore my comments.

yutannihilation (approved):

PR #4449, opened 2021-04-26
Use systematic naming for geom_bin_2d(), just like geom_density_2d()

clauswilke:
For `geom_density_2d()`, we have versions with and versions without underscore before the 2d. But for `geom_bin2d()`, we don't currently have those, even though we do for `geom_stat_2d()`. This PR fixes this, by adding `geom_bin_2d()`.


clauswilke:
Not sure why the text coverage fails. If anybody has an idea, please let me know.

thomasp85 (approved):
LGTM - nice catch

clauswilke:
> LGTM - nice catch

It helps to teach a semester-long course on this material and to try to get everything consistent in the slides. :-)

yutannihilation:
> Not sure why the text coverage fails. If anybody has an idea, please let me know.

Sorry, it's my fault. #4454 will fix.

PR #4450, opened 2021-04-26
Improve color scale documentation

### LINKED ISSUE #4415: Binned scales with manual color choices
I tried the other day to use a binned scale where I would set the color gradient by manually picking colors, and I found myself unable to do so. (At least with only investing a moderate amount of time, reading the docs and trying a few different things.) Is there a way currently to do this? If so, could it be added to the docs? If not, could we add something like `scale_fill_binned_manual()` or `scale_fill_binned_gradientn()` for this purpose?

I also note that `scale_fill_binned()` is not documented at all right now. This is the man page that comes up for it: https://ggplot2.tidyverse.org/reference/scale_colour_continuous.html

For a reprex, how do I make the binned version of the following plot?

``` r
library(ggplot2)

ggplot(faithfuld, aes(waiting, eruptions, fill = density)) +
  geom_tile() +
  scale_fill_gradientn(
    colors = c("red", "green", "blue")
  )
```

<sup>Created on 2021-04-12 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

@thomasp85 It would be great if some minimal version of this issue could be addressed for the 3.3.4 release, as binned scales with manual color choices are likely an important feature to anybody wanting to use binned scales in practice.

clauswilke:
Improve color scale documentation. Closes #4415.


R/scale-colour.r
```diff
@@ -3,7 +3,12 @@
 #' Colour scales for continuous data default to the values of the
 #' `ggplot2.continuous.colour` and `ggplot2.continuous.fill` options. These
 #' [options()] default to `"gradient"` (i.e., [scale_colour_gradient()] and
-#' [scale_fill_gradient()])
+#' [scale_fill_gradient()]). Note that both continuous and binned scales
+#' are configured with the same options values. In other words,
+#' the option `ggplot2.continuous.colour` controls the behavior of both
+#' `scale_colour_continuous()` and `scale_colour_binned()`. To manually set
```
thomasp85, line 9: Wouldn't it be more correct to say that the binned scale falls back to `ggplot2.continuous.colour` if `ggplot2.binned.colour` is not set?

clauswilke:
@thomasp85 I have fixed the issue you pointed out and generally expanded the documentation of `scale_colour_continuous()` and related scales.

thomasp85:
This is kind of orthogonal to this PR, so sorry for hijacking it, but it occurs to me that the fallback to `ggplot2.continuous.colour` only should happen if that value is not a scale... you set `ggplot2.continuous.colour` to e.g. `scale_colour_distiller()` it would inadvertently transform `scale_colour_binned()` to a continuous scale

clauswilke:
I agree with your comment, but that's a lot of logic for a default argument. Should I move that logic into the main body of the function? I don't like complex code execution in default arguments anyways.

thomasp85:
I think it is already way too much to have in the signature so it is better moved inside the function...

But we should do this consistently then, maybe in a new PR

clauswilke:
That’s fine, I can do it. Can we merge this PR first?

thomasp85 (approved):
Yes, let us just merge this in

PR #4446, opened 2021-04-24
Fix tests/testthat/test-prohibited-functions.R

yutannihilation:
Fix #4445

PR #4454, opened 2021-04-27
Fix failures on test coverage CI

yutannihilation:
(Fix https://github.com/tidyverse/ggplot2/pull/4446#issuecomment-826799440)

Tests in `tests/testthat/test-prohibited-functions.R` needs to parse R files to do static analysis. But, files are not always available, so this test might fail and `covr::coverage()` is the case.

https://github.com/tidyverse/ggplot2/blob/bb8f960356a55fe2e72cfd0569179835818f00d8/tests/testthat/test-prohibited-functions.R#L49-L51

This pull requests let the test skip on `covr::coverage()`, and also on CRAN for safety.


thomasp85 (approved):

thomasp85:
Thanks!

PR #4456, opened 2021-04-28
Don't switch scale type when using defaults

clauswilke:
Addresses this comment: https://github.com/tidyverse/ggplot2/pull/4450#issuecomment-828231861

Reprex:
``` r
library(ggplot2)

v <- ggplot(faithfuld, aes(waiting, eruptions, fill = density)) +
  geom_tile()

options(ggplot2.continuous.fill = "viridis")
v
```

``` r
v + scale_fill_binned()
```

``` r
options(ggplot2.continuous.fill = scale_fill_viridis_c)
v
```

``` r
v + scale_fill_binned()
```

<sup>Created on 2021-04-28 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

### DIFF
--- a/R/scale-colour.r
+++ b/R/scale-colour.r
@@ -161,6 +161,9 @@ scale_fill_binned <- function(...,
 # helper function to make sure that the provided scale is of the correct
 # type (i.e., is continuous and works with the provided aesthetic)
 check_scale_type <- function(scale, name, aesthetic) {
+  if (!is.ggproto(scale) || !inherits(scale, "Scale")) {
+    abort(glue("The `type` argument of `{name}()` must return a continuous scale for the {aesthetic} aesthetic. The provided object is not a scale function."))
+  }
   if (!isTRUE(aesthetic %in% scale$aesthetics)) {
     abort(glue("The `type` argument of `{name}()` must return a continuous scale for the {aesthetic} aesthetic. The provided scale works with the following aesthetics: {glue_collapse(scale$aesthetics, sep = ', ')}"))
   }
--- a/R/scale-colour.r
+++ b/R/scale-colour.r
@@ -78,7 +78,7 @@ scale_colour_continuous <- function(...,
   type <- type %||% "gradient"
 
   if (is.function(type)) {
-    type(...)
+    check_scale_type(type(...), "scale_colour_continuous", "colour")
   } else if (identical(type, "gradient")) {
     scale_colour_gradient(...)
   } else if (identical(type, "viridis")) {
@@ -95,7 +95,7 @@ scale_fill_continuous <- function(...,
   type <- type %||% "gradient"
 
   if (is.function(type)) {
-    type(...)
+    check_scale_type(type(...), "scale_fill_continuous", "fill")
   } else if (identical(type, "gradient")) {
     scale_fill_gradient(...)
   } else if (identical(type, "viridis")) {
@@ -110,7 +110,7 @@ scale_fill_continuous <- function(...,
 scale_colour_binned <- function(...,
                                 type = getOption("ggplot2.binned.colour")) {
   if (is.function(type)) {
-    type(...)
+    check_scale_type(type(...), "scale_colour_binned", "colour")
   } else {
     type_fallback <- getOption("ggplot2.continuous.colour", default = "gradient")
     # don't use fallback from scale_colour_continuous() if it is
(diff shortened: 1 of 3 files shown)


clauswilke:
FYI: The ubuntu/ R3.3 run failed at the setting-up-R stage. I'm not going to restart all CI runs just for that.

thomasp85:
I think we should use this occasion to also add checks that the defaulted functions have the correct type and aesthetics, i.e. if the default is a function, expect the return value and see if it is a continuous scale for `scale_colour_continuous()` and if `colour` is part of the aesthetics

clauswilke:
Just to be clear: Do you mean adding something like the following?

```
scale <- type(...)
if (!"colour" %in% scale$aesthetics | isTRUE(scale$is_discrete())) {
  stop("invalid scale function")
}
scale
```

thomasp85:
Yes, exactly... I feel this would be a good catch

clauswilke:
In fact, it would have helped me the other day when I was trying to figure out how to use this argument and was trying to provide a palette function. :-)

Will do today or tomorrow.

clauswilke:
Ok, done I think.

R/scale-colour.r
```diff
@@ -104,29 +108,68 @@ scale_fill_continuous <- function(...,
...
+# type (i.e., is continuous and works with the provided aesthetic)
+check_scale_type <- function(scale, name, aesthetic) {
+  if (!is.ggproto(scale) || !inherits(scale, "Scale")) {
+    abort(glue("The `type` argument of `{name}()` must return a continuous scale for the {aesthetic} aesthetic. The provided object is not a scale function."))
+  }
+  if (!isTRUE(aesthetic %in% scale$aesthetics)) {
+    abort(glue("The `type` argument of `{name}()` must return a continuous scale for the {aesthetic} aesthetic. The provided scale works with the following aesthetics: {glue_collapse(scale$aesthetics, sep = ', ')}"))
   }
+  if (isTRUE(scale$is_discrete())) {
+    abort(glue("The `type` argument of `{name}()` must return a continuous scale for the {aesthetic} aesthetic, but the provided scale is discrete."))
```
thomasp85, line 171: Shouldn't we use this for the default discrete scale as well, in which case `scale$is_discrete()` should match some input argument?

thomasp85: Related issue for discrete scales: #4149

clauswilke: Ok, I'll look into that issue and see if I can fix it also.

thomasp85: Sorry for pilling onto what was effectively a documentation change🙂

clauswilke: @thomasp85 Do you want an error or a warning when the scale doesn't contain the right aesthetic? @yutannihilation suggested a warning [here,](https://github.com/tidyverse/ggplot2/issues/4149#issuecomment-827576936) I have currently implemented an error.

clauswilke: I'm slightly worried that issuing a warning and modifying the list of aesthetics can create weird cases when somebody maps two different variables to `color` and `fill`, but maybe that's a sufficiently obscure case that it doesn't matter. And there'll still be a warning that something isn't right.

yutannihilation: Yes, I too am worried, to be honest... The justification is that the option has a different semantics than usual `+ scale_{colour,fill}_*` in that it's used both for `fill` and for `colour` by default, in contrast to that `+ scale_{colour,fill}_*` isn't. Specifying a `scale_*` function there causes the conflict between two semantics, so apparently it needs some tweak to soften the conflict. On the other hand, the error might be a good lesson to learn `aesthetics` argument, which many users don't know how useful it is yet. I have no idea which is better.

clauswilke: It looks like this is done differently for continuous and discrete scales. Continuous scales don't look at the defaults of other aesthetics. Maybe we should merge the current PR as is and defer a broader cleanup of this issue to the next major release, when we wanted to somewhat change how scales work anyways.

yutannihilation: I see. Sounds good to me.

thomasp85: We can merge this, but I'm very tempted to also fix the discrete scales for this release. The current behaviour is a mistake I think...

At least it should be avoided when the defaults are functions and only when a palette is provided

thomasp85: But to be frank that would probably be more confusing than simply not inheriting across aesthetics at all

thomasp85 (approved):

Issue #4064, opened 2020-06-16
Annotate all facets with axis ticks and labels for fixed scales


clauswilke:
By default, interior facets are not drawn with axis ticks or labels.

``` r
library(ggplot2)

df <- data.frame(
  x = rnorm(900),
  y = rnorm(900),
  a = rep(letters[1:9], each = 10)
)

ggplot(df, aes(x, y)) +
  geom_point() +
  facet_wrap(vars(a))
```

![](https://i.imgur.com/t3JJimc.png)

However, if we set `scales = "free"`, we do get per-facet axis ticks and labels, but now each facet also has its own scale range.

``` r
ggplot(df, aes(x, y)) +
  geom_point() +
  facet_wrap(vars(a), scales = "free")
```

![](https://i.imgur.com/hyS4ECI.png)

I find that it's often desirable to have fixed scales but have each facet have its own axis annotations. This can currently be achieved by setting `scales = "free"` but then manually setting the scale limits.

``` r
ggplot(df, aes(x, y)) +
  geom_point() +
  facet_wrap(vars(a), scales = "free") +
  xlim(-4, 4) + ylim(-4, 4)
```

![](https://i.imgur.com/PnjPESc.png)

I think it would be a good idea to add an option that draws axis annotations for all facets even if scales are fixed. I routinely wish I had this option.

clauswilke:
I believe these lines are responsible for removing the axis annotations, and we would simply have to modify the `if` conditions to make this feature work.

https://github.com/tidyverse/ggplot2/blob/7e9e9d61b4406b21825d60f07821601a3157d656/R/facet-wrap.r#L289-L296

smouksassi:
Might be worth to look at what is possible with lemon : `facet_rep_grid`
https://cran.r-project.org/web/packages/lemon/vignettes/facet-rep-labels.html

thomasp85:
Yeah, this is not something that is particularly hard to do. I've never found the need, personally, but if you feel for it we can add a switch

clauswilke:
Yes, I think it'd be very useful. Of course I had hoped it would transfer over when calling ggplotly on the resulting object, and that's not the case. Oh well, one step at a time.

clauswilke:
I think the main challenge will be to come up with a good name for the argument. We currently have `scales`. Maybe `axes`? `axes = "all"` draws all, `axes = "all_x"` draws all for x, `axes = "margins"` only draws axes on the margins?

[...]

Issue #3329, opened 2019-05-15
Convert guide system to ggproto


clauswilke:
The guide system is the last remaining major ggplot2 component that has not been converted to ggproto. As a consequence, it doesn't match well with the rest of the ggplot2 code base, and it is difficult to extend by external package maintainers. We should consider converting it over, possibly before building additional components that would use the current system, as in #3322.

paleolimbot:
Yes! I was surprised that guides were S3 when I opened up that code. I think it will be less work to do this after #3322, which is to say, after all the guides use the same system. I'd be happy to take this on as well, but would prefer to do #3322 first.

teunbrand:
The ggplot2 book [softly discourages](https://ggplot2-book.org/extensions.html#new-guides) extending ggplot's guide system. However, I'm pretty keen on making some guide extensions, so I'd be willing to spend some time experimenting. What are the current thoughts on moving the guide system to ggproto? I've heard on the grapevines that the entire secondary axis mechanism might move to guides, for example. Other than simply wrapping guide code into ggproto methods, what would need to change?

thomasp85:
@teunbrand I can expand a bit on why I've yet to build up the stamina to do this. If it were a simple rewrite of the current guide methods into a ggproto system it would be quite quick and we would have done it years ago. However, I want this to be an opportunity to rethink the guide implementation in a better way. Currently there is so much code duplication between the different legend guides in terms of how to lay out the grobs, calculate ticks etc. One could even go so far as to say that the keys in a colorbar guide is actually an instance of an axis guide and should use that implementation - but can that be generalised to e.g. the legend guide? There are quite a number of questions about how to do this properly in a manner that doesn't constraint future guide development but allows maximum code reuse.

This is not to deter you from taking on the task. I would be happy if you had the stamina for it, and I would love to set off some time to discuss and guide you through it so you end up with a PR that we agree with the direction on.

teunbrand:
Thanks for your thoughts Thomas, they make a lot of sense to me. From an extension POV, it would be useful too if (in particular the drawing) code was more modular in some ways, which would align with the perspective of reducing code duplication if these modules can be re-used in different guides. For example, if you'd like to make ticks for the minor breaks, you wouldn't want to touch the axis lines or labels parts. I'll have a walk through the code at some point and take stock of what is the same/very similar or very different between various guides.

thomasp85:
[...]

PR #4467, opened 2021-05-05
Axes at interior panels #4064

teunbrand:
This PR aims to implement drawing axes at interior facets proposed in #4064. It implements this for `facet_wrap()` as well as for `facet_grid()`. While the changes for `facet_wrap()` are quite minimal, for `facet_grid()` some additional code was borrowed from `facet_wrap()` to draw additional axes. 

A few discussion points:
* With regards to the main challenge (the name of the argument), I went with `draw.axes` because it felt somewhat more intuitive than just `axes` to me. I'm not particularly attached to the name though.
* I also think the unit tests might still be improved, they now require drawing four plots which doesn't make the tests particularly lightweight.
* Lastly, https://github.com/tidyverse/ggplot2/issues/4064#issuecomment-646564182 mentions an option for omitting the labels at interior axes. I could add this, but wanted to ask if the maintainers approve of this first.

Two small demonstrations:
``` r
library(ggplot2)

df <- data.frame(
  x = 1, y = 1, facet = LETTERS[1:4]
)

ggplot(df, aes(x, y)) +
  geom_point() +
  facet_wrap(vars(facet), draw.axes = "all")
```

``` r
df <- data.frame(
  x = 1:4, y = 1:4, 
  fx = c("A", "A", "B", "B"),
  fy = c("c", "d", "c", "d")
)

ggplot(df, aes(x, y)) +
  geom_point() +
  facet_grid(vars(fy), vars(fx), draw.axes = "all")
```

<sup>Created on 2021-05-05 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>


smouksassi:
When I need this functionality I rely on:
https://cran.r-project.org/web/packages/lemon/vignettes/facet-rep-labels.html
It offers different ways to specify how axes are repeated

teunbrand:
I have also some version of this in an extension package of mine:
https://teunbrand.github.io/ggh4x/articles/Facets.html#extended-facets-1
But I'll gladly make it unnecessary through incorporation in ggplot2.

smouksassi:
Agree having this in ggplot2 will make it more mainstream I use ggh4x a lot especially the facet_nested ! Thank you !

thomasp85:
@clauswilke any thoughts on this - I'm in two minds

clauswilke:
I like this feature. My workaround, which I use all the time, is to set scales to free and then impose scale limits. This has the same effect visually but is a bit confusing, as it requires setting two contradictory options. I'd also say that being able to omit labels for interior plots would be useful.

I haven't reviewed the code, though.

hadley:
The current interface seems like a reasonable compromise to me; it meaningfully improves the display options available, without getting bogged down into too many details. However, I do think `axes` fits better with the current argument names than `draw.axes`.

@teunbrand do you mind changing to `axes` and then merging/rebasing against main?

@clauswilke do you want to review this? or should I?

teunbrand:
Thanks Hadley! I'll take a look in a few days time once I've got some to spare myself.

teunbrand:
> option for omitting the labels at interior axes

Revisiting this topic now a while later, I don't think there currently exists a clean path towards omitting labels. We can edit the axis grob in the `draw_panels()` method, but it feels a bit messy. While I think it is useful to be able to omit labels, the messy approach might be more difficult to maintain in the future, and the faceting code is already quite complicated. 

What I think is a neater solution would be to create an argument in the `render_axes()` function that propagates all the way down to `ggplot2:::guide_gengrob.axis()`, that controls the drawing of the label parts. I could give this a go, but it might be a good idea to postpone this and consider implementing #3329 first.

teunbrand:
So this PR has been barren for a while, but with ggproto guides, we can now also implement the censoring of labels. I've called the argument that controls this `axis_labels` but perhaps there is a more intuitive name.

``` r
#> ℹ Loading ggplot2

base <- ggplot(mtcars, aes(mpg, disp)) +
  geom_point()

base + facet_grid(vars(cyl), vars(vs), axes = "all", axis_labels = "margins")
```



``` r
base + facet_wrap(vars(cyl, vs), axes = "all", axis_labels = "margins")
```



<sup>Created on 2023-05-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
Yeah it looks like it should if no interior axes are drawn or the interior axes are drawn with labels. The issue is that the space reserved for axes is calculated as the maximum size for all the axes in a particular row or column. We could try to calculate the size of only the relevant axes that are in between panels, but this would get pretty gnarly and the drawing code already isn't the easiest to work with.

teunbrand:
Now with much better spacing for hidden labels:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  guides(x.sec = "axis", y.sec = "axis") +
  facet_wrap(vars(cyl, vs), axes = "all", axis_labels = "margins")
```



<sup>Created on 2023-10-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
A small update on how this PR now interacts with newer features.

When using stacked axes, only the first axis in the stack is rendered at interior panels without labels if so directed.

``` r
#> ℹ Loading ggplot2

p <- ggplot(mtcars, aes(mpg, disp)) +
  geom_point() +
  facet_wrap(vars(cyl, vs), axes = "all", axis_labels = "margins")

p + guides(x = guide_axis_stack("axis", "axis"),
           y = guide_axis_stack("axis", "axis"))
```



An exterior r-axis in `coord_radial()` is treated as typical axes, i.e. labels are censored.

``` r
p + coord_radial()
```



Whereas interior r-axes are rendered in full.

``` r
p + coord_radial(end = 1.5 * pi)
```



<sup>Created on 2023-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/facet-grid-.R
```diff
@@ -112,7 +122,8 @@ facet_grid <- function(rows = NULL, cols = NULL, scales = "fixed",
                        space = "fixed", shrink = TRUE,
                        labeller = "label_value", as.table = TRUE,
                        switch = NULL, drop = TRUE, margins = FALSE,
-                       facets = deprecated()) {
+                       facets = deprecated(), axes = "margins",
+                       axis_labels = "all") {
```
thomasp85, line 126: we should continue to use `.` separation in argument names

thomasp85 (changes requested):

R/facet-grid-.R
```diff
@@ -367,17 +407,53 @@ FacetGrid <- ggproto("FacetGrid", Facet,
...
+        unit(apply(axes$x$top, 1, max_height, value_only = TRUE), "cm"),
+        name = "axis-t", z = 3
+      )
+      axes$x$bottom <- matrix(axes$x$bottom[x_axis_order],
+                              nrow = nrow, ncol = ncol, byrow = TRUE)
+      panel_table <- weave_tables_row(
+        panel_table, axes$x$bottom, 0,
+        unit(apply(axes$x$bottom, 1, max_height, value_only = TRUE), "cm"),
+        name = "axis-b", z = 3
+      )
```
thomasp85, line 425: Can this be abstracted away if we are using it multiple places in the code base (both for x and y and in facet_grid and facet_wrap?

R/facet-grid-.R
```diff
@@ -112,7 +122,8 @@ facet_grid <- function(rows = NULL, cols = NULL, scales = "fixed",
                        space = "fixed", shrink = TRUE,
                        labeller = "label_value", as.table = TRUE,
                        switch = NULL, drop = TRUE, margins = FALSE,
-                       facets = deprecated()) {
+                       facets = deprecated(), axes = "margins",
+                       axis.labels = "all") {
```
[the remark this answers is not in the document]
teunbrand: Good catch

R/facet-grid-.R
teunbrand: I'd like this too, but the actual facet_wrap code has very involved code dealing with empty panels (restoring axes for non-empty panels and getting the correct axis sizes). This is right between the `matrix(axes$...)` and `weave_tables_*()` steps. We can skip all of this because facet_grid doesn't have empty panels. So while the approach is the same, it would be tricky to unify the actual code.

thomasp85: could we at least abstract it away for x and y?

teunbrand: I'll see what I can do

teunbrand: I figured that `facet_wrap()` didn't need to measure the axis before dealing with empty panels, so axis measurement and adding it to the table is now wrapped in a `weave_axes()` helper. This helper is also used in `facet_grid()` when required.

thomasp85: see... I knew you could do it

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review Thomas!

PR #4471, opened 2021-05-07
allow fewer elements in named values vector in manual scales

### LINKED ISSUE #3451: scale_manual throws unexpected error when na.values given and insufficient values provided
In the help documentation for `scale_manual()`, the values portion states: 

> values    "a set of aesthetic values to map data values to. If this is a named vector, then the values will be matched based on the names. If unnamed, values will be matched in order (usually alphabetical) with the limits of the scale. Any data values that don't match will be given na.value."

Given this definition, I was surprised that an error was still thrown in the following situation: 

For discrete scales, regardless of whether the actual values match, if there are too few values provided, the "Insufficient Values" error is thrown, as in the test below:

https://github.com/tidyverse/ggplot2/blob/92d27779e210ce16f441eeec26926b8b045a47a8/tests/testthat/test-scale-manual.r#L38

You could provide a sufficient number of (non-matching) values and avoid the error.

```
df <- data.frame(x = 1, y = 1:3, z = factor(c("a","b", NA), exclude = NULL))
p <- qplot(x, y, data = df, colour = z)

# Case 1: Insufficient values error (as expected)
ggplot_build(p + scale_colour_manual(values = c("a" = "black")))

# Case 2: Insufficient values error (not as expected)
ggplot_build(p + scale_colour_manual(values = c("a" = "black"), na.value="grey"))

# Case 3: plot with "b" points given na.value (as expected)
ggplot_build(p + scale_colour_manual(values = c("a" = "black", "c" = "white"), na.value="grey"))

```

I was expecting an error if `na.value` was not specified (Case 1). 
I was NOT expecting an error if `na.value` WAS specified (Case 2).  

My thinking being that case 2 and case 3 are doing basically the same thing but Case 3 has some arbitrary extra value(s) added simply to match the number of values.  I'm not sure what was intended, but seems to me that if `na.value` is provided, the number of values given shouldn't really matter because "Any data values that don't match will be given na.value."  

Related to #2428

thomasp85:
Fix #3451 

This PR relaxes the requirement on the manual scale `values` vector and allow it to contain fewer elements than the data. Non-existing values are implicitly considered `NA` and mapped to the `na.value` value

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # ggplot2 (development version)
 
+* Manual scales now allow named vectors passed to `values` to contain fewer 
+  elements than existing in the data. Elements not present in values will be set
+  to `NA` (@thomasp85, #3451)
+
 * Add support for the BrailleR package for creating descriptions of the plot
   when rendered (@thomasp85, #4459)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -126,6 +126,9 @@
 
 * The scale arguments `limits`, `breaks`, `minor_breaks`, `labels`, `rescaler`
   and `oob` now accept purrr style lambda notation (@teunbrand, #4427).
+  
+* `as_labeller()` (and therefore also `labeller()`) now handles functions in
+  purrr-style lambda notation (@netique, #4188).
 
 # ggplot2 3.3.3
 This is a small patch release mainly intended to address changes in R and CRAN.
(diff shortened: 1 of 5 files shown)


paleolimbot (approved):

thomasp85:
What is the behaviour you'd want? This change means that unmatched values gets the na value (defaults to `NA` which means they won't get drawn)

thomasp85:
Why am I so bad at this today...

thomasp85:
Ah - forgot to push the fix, sorry

clauswilke:
I've looked at the code but I have no idea why the patch does what it does. If @paleolimbot thinks it's good I'm fine with that.

paleolimbot (approved):

PR #4475, opened 2021-05-11
Make sure `mapping` is not stateful

thomasp85:
This PR fixes the "statefulness" of the `mapping` field introduced in #4265 by aligning it with how we now treat `geom_params` and `stat_params`. In fact the state has existed since the introduction of `LayerSf` where the mapping would be modified to ensure the geometry was catched.

This clean-up leaves a couple of dangling issues, mainly the fact that `default_mapping` no longer needs to be passed around. It is kept as-is so we don't change the signature of the guide calculation and break extension packages. We can remove it entirely when we rewrite the guide functionality to ggproto


clauswilke (approved):
Ok, looks good to me.

I notice that there doesn't seem to be a unit test for the `inherit.aes` argument, so maybe add a very simple one.

Issue #4494, opened 2021-05-29
Are you sure that changing `coord_sf` the way it is implemented in v3.3.4-rc is a good idea?


MikkoVihtakari:
Referring to this problem: https://github.com/MikkoVihtakari/ggOceanMaps/issues/3

As far as I have understood, the way you implement `coord_sf` change in the developmental version will break every package and script relying on getting CRS from the `crs` argument because `coord_sf(default_crs = NULL)` will not work on old versions of ggplot2.

``` r
library(ggplot2)
packageVersion("ggplot2")
#> [1] '3.3.3'

ggplot(data.frame(lon = 60, lat = 70), aes(x = lon, y = lat)) + 
  geom_point() + coord_sf(crs = 4326, default_crs = NULL)
#> Error in coord_sf(crs = 4326, default_crs = NULL): unused argument (default_crs = NULL)
```

<sup>Created on 2021-05-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

What if you implemented this to the `crs` argument instead? `NULL` the new way and any integer with 4 numbers a defined CRS and the old way. As far as I see, that way could be backward compatible, no?

Or perhaps there is another trick to make old and new versions of ggplot2 work?

``` r
library(ggOceanMaps)
#> Loading required package: ggplot2
#> Loading required package: ggspatial
#> Setting data download folder to a temporary folder
#> /var/folders/fm/xtn13rlj7d3gp8twy7hh9prxtlq_m4/T//Rtmp3hrlwI. This
#> means that any downloaded map data need to be downloaded again when you
#> restart R. To avoid this problem, change the default path to a
#> permanent folder on your computer. Add following lines to your
#> .Rprofile file: {.ggOceanMapsenv <- new.env(); .ggOceanMapsenv$datapath
#> <- 'YourCustomPath'}. You can use usethis::edit_r_profile() to edit the
#> file.'~/Documents/ggOceanMapsLargeData'would make it in a writable
#> folder on most operating systems.
packageVersion("ggplot2")
#> [1] '3.3.3.9000'

X <- basemap_data(limits = 60)

system.time({
  p <- ggplot2::ggplot() + 
    ggspatial::layer_spatial(data = X$shapefiles$land, fill = "grey") +
    ggspatial::geom_spatial_path(data = X$map.grid$lat.limit.line, aes(x = lon, y = lat), crs = 4326) +
[...]

PR #4496, opened 2021-05-31
Make coord_sf() default use projected coords

clauswilke:
Set `default_crs = NULL`, as discussed in #4494.

### DIFF
--- a/tests/figs/deps.txt
+++ b/tests/figs/deps.txt
@@ -1,3 +1,3 @@
 - vdiffr-svg-engine: 1.0
 - vdiffr: 0.3.3
-- freetypeharfbuzz: 0.2.6
+- freetypeharfbuzz: 0.2.5
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,10 +1,12 @@
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - master
+#on:
+#  push:
+#    branches:
+#      - master
+#  pull_request:
+#    branches:
+#      - master
+
+on: [push, pull_request]
 
 name: R-CMD-check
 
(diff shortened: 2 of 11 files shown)


tests/figs/deps.txt
```diff
@@ -1,3 +1,3 @@
 - vdiffr-svg-engine: 1.0
 - vdiffr: 0.3.3
-- freetypeharfbuzz: 0.2.6
+- freetypeharfbuzz: 0.2.5
```
clauswilke, line 3: @thomasp85 Not entirely sure why this change happened. It doesn't seem to affect any of the reference images. If you think it matters, please let me know what I need to do to change this back. I updated to the latest vdiffr but that didn't seem to make a difference.

clauswilke:
I think this is done, but I'd like to do a bit more testing and review before committing to these changes.

MikkoVihtakari:
One comment I can see just by skimming the diff: just `default_crs = 4326` should be enough (the integer should be passed to st_crs internally) as with the `crs` argument in previous versions of ggplot2. Now it says "`default_crs = sf::st_crs(4326)`" in the documentation.

clauswilke:
Yes, `default_crs = 4326` is sufficient. Just trying to be a bit more explicit and precise in the official documentation. For example, the `datum` argument has always been `datum = sf::st_crs(4326)`.

MikkoVihtakari:
Yep, I saw that now. It's in Usage though. Minor details, but perhaps just write in Details that both `sf::st_crs(4326)` and 4326 are sufficient and that 4326 will be passed to `sf::st_crs`?

tests/figs/deps.txt
thomasp85: I think it is because you have an older version of freetypeharfbuzz installed and it records the version last used to validate the tests

thomasp85 (approved):

PR #4502, opened 2021-06-06
set.seed() before sample()

yutannihilation:
This code is heavily used in our examples. Shouldn't we set a seed to make this deterministic?

```r
dsub <- diamonds[sample(nrow(diamonds), 1000), ]
```


yutannihilation:
Thanks

PR #4503, opened 2021-06-08
Switch to vdiffr 1.0.0 snapshots

lionel-:
See https://vdiffr.r-lib.org/news/index.html#vdiffr-1-0-0-2021-06-08 (blog post upcoming).

Practically, the switch to testthat 3e means that figures are now stored in `tests/testthat/_snaps/` instead of `tests/figs/`. This causes several plots to hit the 100 character limits for relative paths inside a package. To work around this I have shortened several titles in the second commit.

The SVG engine update causes a few differences in snapshots. I have reviewed all of them and the changes seem fine:

- Points now look smaller. They actually have the expected size.
- Some textboxes have slightly different extents, which is expected.
- Colour gradients in legends now fit their boxes better.


yutannihilation (approved):
Thanks. Other than the actual SVG files are too many to review one by one (I believe you), looks good to me! Do you think we should include this in the upcoming version (v3.3.4) of ggplot2 as well?

lionel-:
> Do you think we should include this in the upcoming version (v3.3.4) of ggplot2 as well?

I believe there would be no harm in including it but there is no rush.

thomasp85:
The 3.3.4-rc is frozen at this point so we will just wait for the next (bigger) release

yutannihilation:
Okay, if it's not necessary to include this, let's wait for the next release (with migration to testthat 3e). Thanks for confirming!

PR #4518, opened 2021-06-17
Followup after v3.3.4 release

### LINKED ISSUE #4513: Change in ggsave behaviour
I used to "add" ggsave("something.png") to a ggplot object but this has stopped working after version 3.3.3. In case it's been decided that this "feature" wasn't a good idea and the removal was on purpose, please better document that change (even if the possibility to do + ggsave() wasn't documented before; I think many people used it because it might the pipes very elegant).

In short, this worked on 3.3.3:

```r
library(ggplot2)
data.frame(a = rnorm(100), b = rnorm(100)) |> 
  ggplot(aes(a, b)) +
  ggsave("temp.png")
```

But throws this error on 3.3.4:

```r
Error: Can't add `ggsave("temp.png")` to a ggplot object.

```

yutannihilation:
* Use the dev version
* Revert the tweaked settings for GitHub Actions CI
* Add a note that `+ ggsave()` no longer works, fix #4513

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,12 +1,10 @@
-#on:
-#  push:
-#    branches:
-#      - master
-#  pull_request:
-#    branches:
-#      - master
-
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
Thanks

yutannihilation:
Thanks for approving :)

PR #4528, opened 2021-06-23
Fix issue 4527

clauswilke:
Creating PR against master for now so the automated tests are run.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
   #4521)
 * Warn when grDevices specific arguments are passed to ragg devices (@thomasp85, 
   #4524)
+* Fix an issue where `coord_sf()` was reporting that it is non-linear
+  even when data is provided in projected coordinates (@clauswilke, #4527)
 
 # ggplot2 3.3.4
 This is a larger patch release fixing a huge number of bugs and introduces a 
--- a/R/coord-sf.R
+++ b/R/coord-sf.R
@@ -22,7 +22,7 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
 
     params <- list(
       crs = crs,
-      default_crs = self$default_crs %||% crs
+      default_crs = self$default_crs
     )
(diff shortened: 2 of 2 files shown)


clauswilke:
@thomasp85 Sorry, I don't quite know the right flow when working with a branch. I checked out v3.3.5-rc. Should I set up the PR against that branch also? Will CI run then?

thomasp85:
Test should run no matter what branch you want to merge into - the only difference I've noticed is that issues are not automatically closed if a PR is merged into something other than master

Are you confident that this change doesn't have any negative impact? Or should I run a revue-check?

clauswilke:
> Are you confident that this change doesn't have any negative impact?

I'm never confident about that. :-) A revdep-check wouldn't hurt.

The transformation function aborts if either the two crs's are the same or if one is `NULL`. So whether we explicitly set the default crs equal to the layer crs or set it to `NULL` shouldn't make any difference. I think early in writing this code I was thinking that `default_crs` would have to actually take on a value at some point during rendering, but that's actually more difficult to reason about than just keeping it `NULL` throughout.

https://github.com/tidyverse/ggplot2/blob/389b864657821e967ba50fbf5097b135c22ec0a4/R/coord-sf.R#L504-L510

Obviously if the `default_crs` parameter is set to some value none of this matters. Maybe I should catch that possibility in the `is_linear()` function also. Let me try this out and possibly make another commit.

thomasp85:
👍 I'll run a revdepcheck when you ping me

clauswilke:
Sorry, never mind, that requires a bunch of other changes that I don't want to make now. I think this PR is our best bet. If you have a bunch of sf plotting code that you could run manually to test that would be great. I have done so and haven't seen any issues, and the tests pass, but we know about these pesky corner cases. :-)

clauswilke:
Closing this PR so I can open one for the release branch.

PR #4560, opened 2021-07-22
`stat_bin()` accepts functions for `breaks`

aijordan:
Allowing `breaks` in `stat_bin()` to accept functions is useful in the context of grouping structures. Particularly, when data are mixed discrete and continuous, e.g., precipitation.

The changes also enable facet histograms for estimators with variable bin width, e.g, equal area of all blocks, or the "essential histogram" ([doi:10.1093/biomet/asz081](https://doi.org/10.1093/biomet/asz081)).

This PR is in the same spirit as #1890

### DIFF
--- a/R/stat-bin.r
+++ b/R/stat-bin.r
@@ -23,6 +23,12 @@
 #' @param breaks Alternatively, you can supply a numeric vector giving
 #'    the bin boundaries. Overrides `binwidth`, `bins`, `center`,
 #'    and `boundary`.
+#'
+#'    Can also be a function that returns a numeric vector of bin boundaries
+#'    calculated from unscaled x. Here, "unscaled x"
+#'    refers to the original x values in the data, before application of any
+#'    scale transformation. When specifying a function along with a grouping
+#'    structure, the function will be called once per group.
 #' @param closed One of `"right"` or `"left"` indicating whether right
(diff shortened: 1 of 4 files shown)


clauswilke:
Please open an issue before creating a PR, so we can discuss.

teunbrand:
Is there any change we could revive this PR? If not, I can fix the issue in a different PR.
I think the content of the PR looks good, but PR might need to be re-targetted from the 'master' branch to the 'main' branch and merge in latest changes.

aijordan:
I'd be happy to recreate the PR. I cannot reopen or change the base, it seems.

teunbrand:
Yeah that'd be great, thanks!
If you were to recreate it, I'd slightly adapt the wording so that it is clearer what the function input/output it, something like:
> Can also be a function that takes group-wise values as input and returns bin boundaries.  

I think scale transformation has already been applied once it arrives at break calculation, so that part can be omitted I think.

aijordan:
I believe the comment about scale transformation was essentially copied from the argument description for `binwidth`. I suppose it could or should be removed there as well!?

teunbrand:
I'll open an issue about the `binwidth` docs, they appear incorrect. You can leave it as-is for now, as it is unrelated to the changes proposed in this PR.

Issue #4094, opened 2020-06-26
geom_path has a default size of 0.5


2011:
I just started using rstudio a few weeks ago (and ggplot2 for just a couple of weeks), but if this doesn't count as a bug, it certainly represents an inconsistency. In geom_point for example, size = 0.5 reduces the size of the dots, but with geom_path, it produces the same line as omitting the size parameter, and geom_path, size = 1 produces a thicker line (I assume twice as thick).

```r
ggplot() + geom_path(aes(disp, mpg), mtcars, size = 1)
```

yutannihilation:
Default aes can vary depending on the type of Geom. It might not fit on every cases, but it is what it is. For example, you might argue points and line should have the same default, but it might be inconvenient when you try to draw both.

``` r
library(ggplot2)

ggplot(mtcars, aes(disp, mpg)) +
  geom_line()+
  geom_point()
```

![](https://i.imgur.com/dEHU2us.png)

``` r

# points are absorbed
ggplot(mtcars, aes(disp, mpg)) +
  geom_line(size = 1.5)+
  geom_point(size = 1.5, stroke = 0)
```

![](https://i.imgur.com/1Z5aytD.png)

<sup>Created on 2020-06-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

2011:
@yutannihilation I understand that different geoms might need different aes characteristics, but when a user omits a size parameter, he should get a the equivalent of "size = 1", so that adjustments can start from there. I have no objection to having "size = 1" produce the line that omitting the size parameter or using "size = 0.5" presently does, but the current situation (where the user does not even know the starting point when attempting to adjust the size), seems both not user friendly and just plain wrong.

yutannihilation:
I too think it's nice if there's some easy way to see the default values more easily. But it's not worth changing the default values, which changes the visual of the existing code without size specified.

2011:
Seeing the default values would help. I don't know anything about the defaults for other geoms (as I said, I just started using ggplot2 a couple of weeks ago).

PR #4563, opened 2021-07-22
Fix a typo in changelog (wrong issue number)

krassowski:
Just fixing a typo discovered when looking at this entry find out what was the reason for deprecation and whether there are some ramifications, so having a correct link might be helpful to others too (if you agree, please also consider updating it manually in the release notes on GitHub).

The correct issue is #4097 but #4094 was referenced in the changelog.


yutannihilation (approved):
Oh, true... Thanks for catching!

PR #4574, opened 2021-08-04
Prevent dropping data when drawing legend

teunbrand:
There is a very niche bug in the legend drawing code that only affects plotting when the extracted layer data contains a single column. I ran into this bug when I was experimenting with a geom that has a `fill` aesthetic but not an `alpha` aesthetic.


clauswilke (approved):

clauswilke:
Thanks! It's so frustrating that `drop = FALSE` isn't the default.

teunbrand:
Agreed! If/when ggplot2 switches to {vctrs}, the `vec_slice()` function would be great here.

PR #4543, opened 2021-07-05
Use the dev version of pkgdown

yutannihilation:
Fix #4540

Currently the CRAN version of pkgdown seems to have a bug on linking topics. @maelle adviced me to use the dev version of pkgdown (c.f. https://github.com/r-lib/pkgdown/issues/1733), and I confirmed it works as expected on my forked repo: https://yutannihilation.github.io/ggplot2/dev/reference/theme.html

PR #4583, opened 2021-08-12
Fix incorrect name in scale_discrete documentation

michaelgrund:
~~The filename of ``scale-discrete.r`` contains a hyphen right before the .r ending which results e.g. in broken references in several other functions.~~

As I misunderstood how the generation of the documentation works, here's now an adjusted PR which fixes an incorrect name in ``scale_discrete.Rd``.  

The wrong name results in broken links which currently appear when referencing to ```scale_x_discrete()``` e.g. from [scale_continous](https://ggplot2.tidyverse.org/reference/scale_continuous.html).


yutannihilation:
Thanks for this attempt! This is a bug on pkgdown's side. Should be fixed by https://github.com/tidyverse/ggplot2/pull/4543

PR #4585, opened 2021-08-16
update cheat sheet link

averiperny:
The ggplot2 cheat sheet file name changed so I updated the link here.


yutannihilation (approved):
Thank you!

Issue #4616, opened 2021-09-15
Issue with scale_color_manual


chris-prener:
I have just updated to the most recent `ggplot2` version (somewhat belatedly) and have noticed a change in how `scale_colour_manual` behaves. This seems along the same lines as #4511 but with a different issue. 

The reprex below generates a plot where the legend order follows the values of y for the most recent date. This works as expected plot 1 below, where `scale_colour_brewer` is used to define the palette. 

Until I upgraded `ggplot2`, I had the same result with `scale_colour_manual`. Now, however, the legend items are ordered alphabetically instead. This is reproduced in plot 2 below.

I can't find anything to suggest that this was an intended change - apologies if I have missed that. I really appreciate folks taking a look at this and helping sort out what is going on here!

``` r
# dependencies
suppressPackageStartupMessages({
  library(dplyr)
  library(forcats)
  library(ggplot2)
  library(readr)
  
  library(RColorBrewer)
})
  
# load data
metro_data <- read_csv("https://raw.githubusercontent.com/slu-openGIS/MO_HEALTH_Covid_Tracking/master/data/metro_all/metro_full.csv")
#> Rows: 4792 Columns: 14
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr   (2): short_name, full_name
#> dbl  (11): geoid, cases, case_rate, new_cases, case_avg, case_avg_rate, deat...
#> date  (1): report_date
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
# prep data
## subset
metro_data <- filter(metro_data, report_date >= "2020-03-10")

## create factors
metro_data <- mutate(metro_data, factor_var = fct_reorder2(short_name, report_date, case_rate))

# plot data
## define colors
pal <- brewer.pal(n = 8, name = "Set1")
cols <- c("Cape Girardeau" = pal[6], "Columbia" = pal[3], 
          "Jefferson City" = pal[4], "Joplin" = pal[7],
          "Kansas City" = pal[2], "Springfield" = pal[5], 
          "St. Joseph" = pal[8], "St. Louis" = pal[1])

## create plot 1 (legend ordered by y values)
ggplot() +
  geom_line(metro_data, mapping = aes(x = report_date, y = case_rate, color = factor_var), size = 2) +
  scale_colour_brewer(palette = "Set1")
```

[...]

Issue #4556, opened 2021-07-13
drop=TRUE does not drop unused values (after recent update)


Espinosa-Gongora:
Thank you for amazing packages. I am working on a series of reports, each report containing a different dataset and multiple plots, and I need colors to match between plots and reports. I am avoiding RColorBrewer (e.g. myColors <- brewer.pal(5,"Set1") ) because we have already published some of the first reports using other colours that we want to keep. 

This was my original good function:

scale_color_test <- function(...){
  ggplot2:::manual_scale(
    "color",
    values = setNames(c("grey50","gold", "blue2" ,"green3" ,  "red3"),
                      c("color_grey","color_gold","color_blue","color_green","color_red")),
    labels = setNames(c("color grey","color gold","color blue","color green","color red"),
                      c("color_grey","color_gold","color_blue","color_green","color_red")),
    drop=TRUE)
}

But this function recently stopped dropping unused values in the legends (exact same script run before and after a recent update).

Here is my reproducible example, and some of the things I have tried to fix this issue:

df.temp1<-data.frame(var1=factor(c("color_grey","color_gold","color_blue","color_green","color_red")),
                    var2=c(1,2,3,4,5),
                    var3=c(9,8,6,5,4),
                    var4=c("these colors","these colors","these colors","those colors","those colors"))

df.temp2<-data.frame(var1=factor(c("color_gold","color_blue","color_red")),
                    var2=c(2,3,5),
                    var3=c(8,6,4),
                    var4=c("these colors","these colors","those colors"))

# My old approach that used to work (second plot not dropping unused values in legend anymore):

scale_color_test <- function(...){
  ggplot2:::manual_scale(
    "color",
    values = setNames(c("grey50","gold", "blue2" ,"green3" ,  "red3"),
                      c("color_grey","color_gold","color_blue","color_green","color_red")),
    labels = setNames(c("color grey","color gold","color blue","color green","color red"),
                      c("color_grey","color_gold","color_blue","color_green","color_red")),
    drop=TRUE)
}

ggplot(df.temp1, aes(x=var2, y=var3,color=var1))+
  geom_point(shape=21,stroke=1,size=4)+
  scale_color_test()+
  facet_grid(rows = vars(var4))+
  theme_bw()
  
ggplot(df.temp2, aes(x=var2, y=var3,color=var1))+
  geom_point(shape=21,stroke=1,size=4)+
  scale_color_test()+
  facet_grid(rows = vars(var4))+
[...]

PR #4547, opened 2021-07-07
Add tests for values, limits, and breaks in manual scale

banfai:
Issue introduced in #4471, if there are extra named values in a manual scale then they will show up on the legend even with `drop = TRUE`.

Question: should these values be forced with `drop = FALSE`? (currently it's possible)

Issue #4545, opened 2021-07-07
Unused levels persist in legends from version 3.3.4


wjakethompson:
It appears a bug was introduced in version 3.3.4. When using `scale_fill_manual()`, unused levels persist in the legend. Here is some example data.

``` r
library(tidyverse)

set.seed(123)
dat <- tibble(var = paste("Group", 1:9),
              val = runif(9),
              grp = sample(LETTERS[1:4], 9, replace = TRUE))
```

When all levels are present, things work as expected:

``` r
dat %>%
  ggplot(aes(x = val, y = var)) +
  geom_col(aes(fill = grp)) +
  scale_fill_manual(values = c("A" = "red", "B" = "blue", "C" = "orange",
                               "D" = "black"))
```

![](https://i.imgur.com/vxhItX1.png)

But when `grp = "C"` is no longer in the data, that value still persists in the legend.

``` r
dat %>%
  filter(grp != "C") %>%
  ggplot(aes(x = val, y = var)) +
  geom_col(aes(fill = grp)) +
  scale_fill_manual(values = c("A" = "red", "B" = "blue", "C" = "orange",
                               "D" = "black"))
```

![](https://i.imgur.com/UwQLbrZ.png)

From [this thread on StackOverflow](https://stackoverflow.com/questions/68278895/suppress-unused-fill-levels-in-ggplot2-legend), it appears that this worked as expected in version 3.3.3. I have tested version 3.3.4 and the development version from GitHub, and the unused level persists in the legend for both.

<sup>Created on 2021-07-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.1.0 (2021-05-18)
#>  os       macOS Big Sur 10.16         
#>  system   x86_64, darwin17.0          
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       America/Chicago             
#>  date     2021-07-06                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                            
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 4.1.0)                    
[...]

PR #4619, opened 2021-09-15
Manual scale limits default to present names of values

teunbrand:
#### Why this PR?

In ggplot2 v3.3.4, #4471 was introduced to allow fewer items in `values` than auto-detected breaks.
This has had some consequences, most notably that an extra legend key was produced if a named `values` was longer than auto-detected `breaks` (#4511, #4534, #4556, #4545). One workaround is currently to use `limits = force`, which feels unneccessary (https://github.com/tidyverse/ggplot2/issues/4511#issuecomment-866185530). Additionally, the ordering of the legend keys was effected by the names in the `values` argument (#4616).

#### What does this PR do?

This PR aims to revert back to the old behaviour (in most part), while preserving the spirit of #4471. It does so by replacing a vector-based `limits` by a function-based `limits` that intersects the names of the values by the observed limits.

#### Foreseen problems

Both this PR and #4547 aim to do the same thing and therefore these PRs are in conflict. This PR solely affects manual scales, whereas all scales are affected in #4547.

#### Examples

``` r
library(ggplot2)

  geom_point(size = 2)

values <- c("f" = "dodgerblue", "r" = "tomato", "4" = "limegreen", 
            "dummy" = "black")

# Does not change default order (#4616), nor produces extra key (#4511)
p + scale_colour_manual(
    values = values
  )
```

``` r
# Can add extra key by setting limits
# The plot is the same as CRAN version without limits argument
p + scale_colour_manual(
  values = values,
  limits = names(values)
)
```

``` r
# Can set undefined values to na.value (spirit of #4471)
p + scale_colour_manual(
  values = values[1:2]
)
```

``` r
# Can set fewer keys by setting breaks (same as CRAN version)
p + scale_colour_manual(
  values = values,
  breaks = names(values)[1:2]
)
```

<sup>Created on 2021-09-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/scale-manual.r
```diff
@@ -132,8 +132,9 @@ manual_scale <- function(aesthetic, values = NULL, breaks = waiver(), ..., limit
     force(values)
   }
 
-  if (is.null(limits)) {
-    limits <- names(values)
+  if (is.null(limits) && !is.null(names(values))) {
+    force(values)
```
yutannihilation, line 136: Why is this line needed? Wasn't `values` forced in the last `if` section?

yutannihilation (commented):
Thanks, while I'm not fully confident yet, this pull request looks the right direction to me. Manual scales are somewhat special, so it's probably a good idea to treat them specially, rather than modifying the whole `Scale`.

R/scale-manual.r
teunbrand: You're right, it must have slipped my attention. Apologies!

yutannihilation: I see, thanks.

clauswilke:
I have to admit I'm quite confused by this. It's hard to keep track of the many different issues that were filed. I thought we had merged the `limits = force` solution. Is that not the case?

In any case, @teunbrand, could you add tests and a news item, just like in #4547? It would be good to have a test for every single problem that has shown up.

And also, #4471 was exclusive to manual scales, so a fix exclusive to manual scales would be preferred. This argues in favor of this PR over #4547.

yutannihilation:
> It's hard to keep track of the many different issues that were filed.

I too feel so...

> I thought we had merged the `limits = force` solution.

If I understand correctly, this pull request is based on the same idea. This comment (https://github.com/tidyverse/ggplot2/issues/4511#issuecomment-888392422) might help:

> In addition: perhaps, if the aim of #4471 was to allow fewer elements, then the line above could be changed to:

yutannihilation:
> In any case, @teunbrand, could you add tests and a news item, just like in #4547? It would be good to have a test for every single problem that has shown up.

I hope we can narrow #4547 to only the tests and merge it. Even if we'll choose this pull request, it's good if we can count these prior works on #4547 as a contribution.

clauswilke:
Ah, Ok. I have no concerns about this, except the order of merging. Can we merge the test before merging the fix?

yutannihilation:
Yes, let's merge the tests first.

banfai:
@clauswilke @yutannihilation 
I agree, this is a better solution. I have modified #4547 as requested to contain the tests only.

banfai:
@teunbrand
This PR should include an update to NEWS.md
It could be changed to something like this, if you agree to share credit 😉 
```
* Fixes issue with `scale_*_manual` introduced in #4471 where extra elements
  in `values` or their ordering changes the legend
  (@teunbrand, @banfai, #4511, #4534, #4556, #4545, #4616)
```

teunbrand:
@banfai Yes of course, I'm sorry for the conflict with your PR and the hassle that came with it!

yutannihilation:
@banfai 
Ah, thanks, I was going to ask to include you in NEWS otherwise :)

@teunbrand
Now that #4547 is merged, could you merge the current master for tests?

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # ggplot2 (development version)
 
+* `scale_*_manual()` no longer displays extra legend keys, or changes their 
+  order, when a named `values` argument has more items than the data. The 
+  previous behaviour can be replicated by using 
+  `scale_*_manual(values = vals, limits = names(vals))`. (@teunbrand, @banfai, 
+  #4511, #4534, #4556, #4545, #4616)
```
yutannihilation, line 7: While there are many issues filed, I think they can be summarised to these two ones.

```suggestion
  #4511, #4534)
```

R/scale-manual.r
```diff
@@ -132,8 +132,8 @@ manual_scale <- function(aesthetic, values = NULL, breaks = waiver(), ..., limit
     force(values)
   }
 
-  if (is.null(limits)) {
-    limits <- names(values)
+  if (is.null(limits) && !is.null(names(values))) {
+    limits <- function(x) {intersect(x, names(values))}
```
yutannihilation, line 136:
```suggestion
    limits <- function(x) intersect(x, names(values))
```

Tidyverse style guide says:

> `{` should be the last character on the line. 
(https://style.tidyverse.org/syntax.html#indenting)

So, it's probably recommended to write this as either

``` r
    limits <- function(x) intersect(x, names(values))
```

or

``` r
    limits <- function(x) {
      intersect(x, names(values))
    }
```

yutannihilation (commented):
Minor comments. Let me for a while to check if this fixes all the problems that are reported so far.

I'd also like to request to add a short comment on the `if` branch why `limits` needs to be a function.

yutannihilation:
Hmm, I have no idea what this error is...

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Error (test-scale-manual.r:119:3): limits and breaks (#4619) ────────────────
Error: Error: argument is missing, with no default
Error: Error: R CMD check found ERRORs
Execution halted
Backtrace:
    █
 1. └─ggplot2::scale_colour_manual(...) test-scale-manual.r:119:2
 2.   └─ggplot2:::manual_scale(aesthetics, values, breaks, ..., na.value = na.value)
 3.     └─ggplot2::discrete_scale(...)
 4.       └─ggplot2::ggproto(...)

[ FAIL 1 | WARN 0 | SKIP 66 | PASS 1619 ]
```

yutannihilation:
Ah, I should have caught that trailing comma, sorry.

teunbrand:
Alright this seems all in its intended state for now :) Thanks for the feedback @yutannihilation!

>  Let me for a while to check if this fixes all the problems that are reported so far.

That seems like a good plan to me, please let me know if you find something.

banfai:
@teunbrand This was a better solution, I'm glad we could reconcile the two PRs.
@yutannihilation Sorry about the trailing comma error, there used to be an `na.values` arg there and it didn't throw an error on my machine.

yutannihilation:
@banfai 
No worries, it seems something changed between R 3.4 and 3.5. It's what only the CI can find, so it's not your fault. I was a bit too lazy to check the details of the failures.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # ggplot2 (development version)
 
+* `scale_*_manual()` no longer displays extra legend keys, or changes their 
+  order, when a named `values` argument has more items than the data. The 
+  previous behaviour can be replicated by using 
+  `scale_*_manual(values = vals, limits = names(vals))`. (@teunbrand, @banfai, 
+  #4511, #4534)
```
yutannihilation, line 7: The word "previous behaviour" might be a bit confusing to those who want further previous behaviour, i.e. behaviour before v3.3.4. How about describing it a bit more explicitly something like "To display all the `values` on the legend, use `scale_*_manual(values = vals, limits = names(vals))`"? What do you think?

yutannihilation (approved):
I confirmed this solves all the related issues. Now this looks good to me!

c.f. just in case this is useful for someone, this is my quick summary of the issues.
https://gist.github.com/yutannihilation/3d0954fe2a183c08288807215e3c9705

NEWS.md
clauswilke: Agreed, "previous behavior" is confusing. Better to describe the specific behavior. I like the proposed sentence.

yutannihilation:
Perfect, thank you so much!

PR #4625, opened 2021-09-22
Move Hmisc from Suggests to Imports

dgromer:
Fixes https://github.com/hadley/ggplot2-book/issues/276 as proposed in https://github.com/hadley/ggplot2-book/pull/277#issuecomment-923920536


yutannihilation:
Commented there.

https://github.com/hadley/ggplot2-book/pull/277#issuecomment-924686240

Issue #4636, opened 2021-10-07
`annotation_logticks()` with secondary axis


benjaminrich:
Goal is to create a plot with 2 y-axes, linear on the left, and logarithmic on the right. This is simple enough, but it is not possible to get `annotation_logticks()` to produce tick marks that line up correctly in this case. It would require a that further transformation be applied inside `annotation_logticks()`. I have a working solution and will submit a PR.

Reprex:

```r
library(ggplot2)

df1 <- data.frame(x=0:10, y=seq(10, 0, len=11))
df2 <- data.frame(x=0:10, y=seq(0.1, 1000, len=11))

ylim1 <- c(0, 10)
ylim2 <- log10(c(0.1, 1000))

f <- function(y) (diff(ylim1)/diff(ylim2))*(y - min(ylim2)) + min(ylim1)
g <- function(y) (diff(ylim2)/diff(ylim1))*(y - min(ylim1)) + min(ylim2)


ggplot() +
  geom_col(data=df1, aes(x=x, y=y, color="linear scale"), alpha=0.2) +
  geom_line(data=df2, aes(x=x, y=f(log10(y)), color="log scale"), size=2) +
  scale_y_continuous("Linear Scale",
    sec.axis=sec_axis(~ g(.), name="Log Scale", labels=function(x) 10^x)) +
  annotation_logticks(sides="r", color="red") +
  scale_color_manual(breaks=c("linear scale", "log scale"), values=c("blue", "red")) +
  theme_classic() +
  theme(
    axis.title.y       = element_text(color="blue"),
    axis.text.y        = element_text(color="blue"),
    axis.ticks.y       = element_line(color="blue"),
    axis.title.y.right = element_text(color="red"),
    axis.text.y.right  = element_text(color="red"),
    axis.ticks.y.right = element_line(color="red"))
```
![image](https://user-images.githubusercontent.com/17572252/136449032-a10b9e91-e35a-4a36-b145-d9c04dee2b40.png)

The correct/desired output should be:
![image](https://user-images.githubusercontent.com/17572252/136449594-3906a20f-d122-4945-8289-b503fcb8bf4e.png)

hadley:
This secondary y-axis is not a simple transformation of the primary y-axis so is out of scope for ggplot2, sorry.

PR #4637, opened 2021-10-07
Add `trans` option in `annotation_logticks()`

benjaminrich:
This is intended to address #4636.

I could really use some help with this, as I've never contributed to ggplot2 before, and I'm not really sure of the best way to do this.

It's a simple enough change, should not break anything, and it does solve the problem described in #4636 if you add `trans= ~ f(.)` inside `annotation_logticks()`, like this:

```r
df1 <- data.frame(x=0:10, y=seq(10, 0, len=11))
df2 <- data.frame(x=0:10, y=seq(0.1, 1000, len=11))

ylim1 <- c(0, 10)
ylim2 <- log10(c(0.1, 1000))

f <- function(y) (diff(ylim1)/diff(ylim2))*(y - min(ylim2)) + min(ylim1)
g <- function(y) (diff(ylim2)/diff(ylim1))*(y - min(ylim1)) + min(ylim2)


ggplot() +
  geom_col(data=df1, aes(x=x, y=y, color="linear scale"), alpha=0.2) +
  geom_line(data=df2, aes(x=x, y=f(log10(y)), color="log scale"), size=2) +
  scale_y_continuous("Linear Scale",
    sec.axis=sec_axis(~ g(.), name="Log Scale", labels=function(x) 10^x)) +
  annotation_logticks(sides="r", color="red", trans= ~ f(.)) +
  scale_color_manual(breaks=c("linear scale", "log scale"), values=c("blue", "red")) +
  theme_classic() +
  theme(
    axis.title.y       = element_text(color="blue"),
    axis.text.y        = element_text(color="blue"),
    axis.ticks.y       = element_line(color="blue"),
    axis.title.y.right = element_text(color="red"),
    axis.text.y.right  = element_text(color="red"),
    axis.ticks.y.right = element_line(color="red"))
```
image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -266,7 +266,7 @@ Collate:
     'zxx.r'
     'zzz.r'
 VignetteBuilder: knitr
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 Roxygen: list(markdown = TRUE)
 Encoding: UTF-8
 Config/Needs/website:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Allow transformations inside `annotate_logticks()`, which is useful for
+  secondary axes (@benjaminrich, #4636)
+
 * `scale_*_manual()` no longer displays extra legend keys, or changes their 
   order, when a named `values` argument has more items than the data. To display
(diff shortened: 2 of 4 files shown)


hadley:
Thanks for working on this PR, but this extends secondary y-axes in a direction that I'm not generally comfortable with.

PR #4656, opened 2021-10-29
Update to newest CI

thomasp85:
This updates our CI scripts to the latest version, hopefully removing the need for continuous keeping up with best practices.


yutannihilation (approved):
Looks good!

The "polygon edge not found" error on macOS runner should be occasional one so I bet we can ignore it (while I think it's great if we'll find the cause eventually).

Issue #4391, opened 2021-03-25
Allow geom_sf() to accept line parameter arrow?


JakeRuss:
I see the discussion in #2826, that led to PR #2847, but arrow was not included in those additions. Was this intentional or just an oversight?

I think it would be a simple PR and I'll gladly submit it, unless there is a reason it was omitted?

JakeRuss:
I'll see what I can do. What's your timeline for this round of ggplot2 updates?

PR #4659, opened 2021-11-04
Add arrow to line params

JakeRuss:
My vague recollection when I opened #4391, was I thought adding arrow was basically adding it to the list. As you can see from this commit it doesn't work. 

I expected the following to work, but no. 
````
library(sf)
library(ggplot2)

coords <- tibble::tibble(lat = c(26.562855, 28.538336), 
                         lon = c(-81.949532, -81.379234)) %>%
  st_as_sf(coords = c("lon", "lat"),  crs = 4326, remove = FALSE) %>%
  st_coordinates() %>%
  st_linestring()

ggplot() + 
  geom_sf(data = coords, size = 2, 
          arrow = arrow(angle = 45, ends = "last", type = "open", length = unit(0.5, "inches")))
````

I'm not sure I understand enough to complete this task

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* `geom_sf()` now respects `arrow` parameter for lines (@jakeruss, #4659)
+
 * Updated documentation for `print.ggplot` to reflect that it returns
   the original plot, not the result of `ggplot_build()`. (@r2evans, #4390)
 
--- a/R/geom-sf.R
+++ b/R/geom-sf.R
@@ -128,14 +128,15 @@ GeomSf <- ggproto("GeomSf", Geom,
 
   draw_panel = function(data, panel_params, coord, legend = NULL,
                         lineend = "butt", linejoin = "round", linemitre = 10,
-                        na.rm = TRUE) {
+                        arrow = NULL, na.rm = TRUE) {
     if (!inherits(coord, "CoordSf")) {
       abort("geom_sf() must be used with coord_sf()")
     }
 
     # Need to refactor this to generate one grob per geometry type
     coord <- coord$transform(data, panel_params)
-    sf_grob(coord, lineend = lineend, linejoin = linejoin, linemitre = linemitre, na.rm = na.rm)
+    sf_grob(coord, lineend = lineend, linejoin = linejoin, linemitre = linemitre,
+            arrow = arrow, na.rm = na.rm)
   },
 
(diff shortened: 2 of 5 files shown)


clauswilke:
Arrows are not set in `gpar()`, they are set directly in the grob. See e.g. the documentation to `linesGrob()`. I suspect that if you add the arrow info here it will work:
https://github.com/tidyverse/ggplot2/blob/6d19950ec69e18dce119ae121059bae554ce421f/R/geom-sf.R#L216

JakeRuss:
Thank you @clauswilke, that does seem to have done the trick. My example above now works

clauswilke:
Maybe also add a visual test?

JakeRuss:
My apologies, but I'm not sure what that means. If you could point me to an example I'd gladly add it.

clauswilke:
See: https://vdiffr.r-lib.org/

Example:
https://github.com/tidyverse/ggplot2/blob/6e3ac8116195126d3cc843e50fca64bf6e4f43a4/tests/testthat/test-geom-sf.R#L177-L204

JakeRuss:
Greatly appreciated, I will add

JakeRuss:
@thomasp85 Ready to review at your schedule

clauswilke:
@JakeRuss It looks like you didn't commit the new reference images for the two visual tests.

JakeRuss:
@clauswilke Thank you for all of your pointers

clauswilke:
The tests are failing. Please take a look. Looks like you need to write `sf::st_coordinates()` instead of just `st_coordinates()`. There may be other, similar issues once you fix this one.

clauswilke:
`data_frame()` is a function defined by the ggplot2 codebase. It should not require a `tibble::` prefix. See also the other tests.

clauswilke:
The reason why some of the tests still fail is because the reference images differ between the version you committed and the version generated in the github action. You may have outdated versions of some libraries. Not clear which ones, but if not vdiffr then possibly sf.

JakeRuss:
Thanks, I will clean this up again today. I was running the testthat tests locally and it was complaining about data_frame, which I see now was because I didn't have it loaded.

clauswilke (approved):
Looks good to me.

@thomasp85 Any comments or concerns from your end?

thomasp85 (approved):

thomasp85:
Thanks @JakeRuss !

PR #4664, opened 2021-11-09
Consistent use of lineend and linejoin in geoms and keys

### LINKED ISSUE #4653: Inconsistent setting of lineend in draw_key_*
How lineend is set in the draw_key functions does not seem to be standardised and e.g. differs between `draw_key_polygon()` and `draw_key_path()`. In general this is not an issue when the linetype is solid as the ends of the lines are not visible, but once the line is non-solid the lineend will be applied to each segment in the line-pattern. This can result in some dramatic visual discrepancies between what is shown in different keys and in the plot itself.

Example of difference in look between plot and legend as well as difference between polygon and path key:

``` r
ggplot() + 
  geom_segment(
    aes(x = 0, y = 0:1, xend = 1, yend = 1:0, linetype = c('a', 'b')), 
    size = 1.5, 
    lineend = 'round'
  ) + 
  geom_polygon(
    aes(x = c(0.25, 0.75, 0.5), y = c(0.12, 0.12, 0.88), linetype = 'b'), 
    fill = NA, 
    size = 1.5, 
    colour = 'red'
  ) + 
  scale_linetype_manual(values = c('solid', '11'))
```

thomasp85:
This PR fixes #4653, but also goes a step further and makes the use of `lineend` and `linejoin` parameters more consistent across the board. 

Until now, `lineend` has only been possible to set in a few geom that has explicit ends in their lines, but this parameter affect the ends of stippled lines so should be exposed everywhere the `linetype` aesthetic is available.

Further, `linejoin` has also been unavailable in a wide range of geoms, meaning that the corners of e.g. box plots couldn't be controlled. This has now been fixed.

Across the board, the default `lineend` is now `"butt"`. For `linejoin` it is either `"mitre"` (for squar-ish geoms), or `"round"` (for smooth geoms).

All `draw_key_*` functions will now look for `linetype` and `lineend` in `params` if meaningful so that their display matches that shown in the plot.

**Note**: This PR removes the hack around rectangle joins we put in place cf https://github.com/tidyverse/ggplot2/issues/3037#issuecomment-457504667. This is to make behaviour consistent across geoms. I believe the existence of ragg and all the continued issues with the windows device means that we should not forever cripple our code base to appease it


clauswilke (approved):
Looks good to me. I don't see anything that causes me concerns. Though it's a lot of changes at once, so I may have missed something. Let's hope for the best. :-)

PR #4667, opened 2021-11-11
Don't allow stroke = NA to mess up point sizing

### LINKED ISSUE #4624: geom_point(): when stroke=NA, cannot change size of the point
I'm trying to generate a plot with points with no border displayed, just fill and to be able to change their size. However, once `stroke=NA`, I cannot change the size of the points. 

``` r
library(ggplot2)

df = data.frame(x=rnorm(100), y=rnorm(100))
ggplot(df, aes(x, y)) + geom_point(shape=21, stroke=NA, fill="black", size=5)
```

``` r
ggplot(df, aes(x, y)) + geom_point(shape=21, stroke=NA, fill="black", size=0.5)
```

<sup>Created on 2021-09-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

thomasp85:
Fix #4624 by treating ´NA` as `0` when calculating point size

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Setting `stroke` to `NA` in `geom_point()` will no longer impair the sizing of
+  the points (@thomasp85, #4624)
+
 * All geoms now have consistent exposure of linejoin and lineend parameters, and
   the guide keys will now respect these settings (@thomasp85, #4653)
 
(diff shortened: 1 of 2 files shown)


clauswilke (approved):

ZhangSR-kolar:
Hi, thanks for the improvement of the 'stroke' vs. 'size' issue. However, when I use the development version of ggplot2, though 'stroke = NA' won't change the 'size' aes in the plot region, the size legend was still marked and showed with the same size.
`ggplot(mtcars, aes(wt, mpg)) + geom_point(aes(size = qsec), alpha = .5, stroke = NA)`
image

yutannihilation:
Oh, good catch! Thanks.

PR #4670, opened 2021-11-12
follow tidyverse recycling rules for aesthetics

### LINKED ISSUE #4588: Misleading error message when plotting with empty data.frame
When plotting with an empty data (which may happen when filtering goes wrong), there is an error if constant aesthetics are mixed with column names, and the error message is a little confusing.

Example:
```r
> ggplot(head(mtcars, 0)) + geom_point(aes(x = hp, y = mpg)) # no error here, plot is empty
> ggplot(head(mtcars, 0)) + geom_point(aes(x = 3, y = 4))        # no error here, plot contains one point
> ggplot(head(mtcars, 0)) + geom_point(aes(x = 3, y = mpg))   # error!
 Fehler: Aesthetics must be either length 1 or the same as the data (1): y
```

Taken literally, the error message is wrong: Aesthetics `y` does have the same length as the data (0).

Expected behaviour: No error message, return an empty plot.

thomasp85:
Fix #4588 

This PR allows mix of 0-length and 1-length vectors (recycled to 0) in aesthetics... It both puts ggplot2 in line with our decided recycling rules and makes it less error-prone to set aesthetics to scalars as these values can now be mixed with all sizes of data (even 0)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Aesthetics of length 1 are now recycled to 0 if the length of the data is 0 
+  (@thomasp85, #4588)
+
 * Setting `stroke` to `NA` in `geom_point()` will no longer impair the sizing of
   the points (@thomasp85, #4624)
 
--- a/R/geom-.r
+++ b/R/geom-.r
@@ -208,7 +208,7 @@ NULL
 .stroke <- 96 / 25.4
 
 check_aesthetics <- function(x, n) {
-  ns <- vapply(x, length, numeric(1))
+  ns <- vapply(x, length, integer(1))
   good <- ns == 1L | ns == n
 
   if (all(good)) {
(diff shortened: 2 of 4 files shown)


hadley:
Add a small test?

jubilatious1:
I'm not sure I understand. A variable of length-zero assigned an aesthetic is effectively an "option" in R, is it not? 

`Aesthetics of length 1 are now recycled to 0 if the length of the data is 0 `

So zero-length data (effectively NAs) cause their calling Aesthetics to assume NA values as well?

hadley:
@jubilatious1 NA has length 1, not length 0. You can read about the tidyverse recycling rules at https://vctrs.r-lib.org/articles/type-size.html#size-1, but this PR is not the right place for discussion of them.

jubilatious1:
Thank you.

hadley (approved):

PR #4671, opened 2021-11-16
Pass on binwidth and height to geom

### LINKED ISSUE #4580: geom_hex ignores binwidth argument when sample size is small
`geom_hex` ignores the binwidth argument when there are only few data points:

```r
library(ggplot2)
n = 1
ggplot(tibble(x=rnorm(n), y=rnorm(n))) + 
  geom_hex(aes(x=x, y=y), binwidth=c(.1,.1)) + 
  xlim(-3, 3) + ylim(-3,3)
```

For small values of `n` (below 10) the bin widths are not 0.1. For larger `n` (try 100), the problem goes away and the bin width is 0.1 as specified.

thomasp85:
Fix #4580 

This PR makes `stat_bin_hex()` write `width` and `height` to the data which `geom_hex()` can pick up instead of deriving the resolution from data. This fixes issues with very sparse data where the bin resolution is not obvious from the bin centers

### DIFF
--- a/R/geom-hex.r
+++ b/R/geom-hex.r
@@ -66,6 +66,9 @@ GeomHex <- ggproto("GeomHex", Geom,
     } else {
       dx <- resolution(data$x, FALSE)
     }
+    # Adjust for difference in width and height of regular hexagon. 1.15 adjusts
+    # for the effect of the overlapping range in y-direction on the resolution
+    # calculation
     if (!is.null(data$height)) {
       dy <- data$height[1] /  sqrt(3) / 2
     } else {
--- a/R/geom-hex.r
+++ b/R/geom-hex.r
@@ -56,13 +56,31 @@ geom_hex <- function(mapping = NULL, data = NULL,
 GeomHex <- ggproto("GeomHex", Geom,
   draw_group = function(data, panel_params, coord, lineend = "butt",
                         linejoin = "mitre", linemitre = 10) {
+    if (empty(data)) {
+      return(zeroGrob())
+    }
     if (!inherits(coord, "CoordCartesian")) {
       abort("geom_hex() only works with Cartesian coordinates")
     }
+    # Extract binwidth and height from data if possible
+    if (!is.null(data$width)) {
+      data$xend <- data$x + data$width
+    }
+    if (!is.null(data$height)) {
+      data$yend <- data$y + data$height
+    }
 
     coords <- coord$transform(data, panel_params)
+
+    binwidth <- c(NA, NA)
+    if (!is.null(data$width)) {
+      binwidth[1] <- coords$xend[1] - coords$x[1]
+    }
+    if (!is.null(data$height)) {
+      binwidth[2] <- coords$yend[1] - coords$y[1]
+    }
     ggname("geom_hex", hexGrob(
-      coords$x, coords$y,
+      coords$x, coords$y, binwidth,
       gp = gpar(
         col = coords$colour,
         fill = alpha(coords$fill, coords$alpha),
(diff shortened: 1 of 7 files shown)


R/geom-hex.r
```diff
@@ -56,13 +56,31 @@ geom_hex <- function(mapping = NULL, data = NULL,
...
                         linejoin = "mitre", linemitre = 10) {
+    if (empty(data)) {
+      return(zeroGrob())
+    }
     if (!inherits(coord, "CoordCartesian")) {
       abort("geom_hex() only works with Cartesian coordinates")
     }
+    # Extract binwidth and height from data if possible
+    if (!is.null(data$width)) {
+      data$xend <- data$x + data$width
```
clauswilke, line 67: Just so I understand: This is basically a trick to get the width and the height transformed, correct? `xend` and `yend` serve no other purpose otherwise. Have you tried what happens in non-linear coordinates? It'll probably mostly work, but there may be some edge cases where weird stuff happens.

Maybe add a few comments in the code explaining what you're doing, so when the inevitable bug happens 10 years down the road we at least know what the intent was. :-)

thomasp85: you understand correctly... I'll add some comments for our future selfs

thomasp85:
I've decided to refactor `geom_hex()`, both to avoid the weird trick to get the bin width transformed, but more importantly to support non-linear coordinate systems which was an artificial limitation of the geom... @clauswilke can you give it a review again?

clauswilke (commented):
I think this is a good idea. Give me a few days to look this over carefully. On quick glance looks good though.

R/geom-hex.r
```diff
@@ -56,12 +56,33 @@ geom_hex <- function(mapping = NULL, data = NULL,
...
+    # Get hex sizes
+    if (!is.null(data$width)) {
+      dx <- data$width[1] / 2
+    } else {
+      dx <- resolution(data$x, FALSE)
+    }
+    if (!is.null(data$height)) {
+      dy <- data$height[1] /  sqrt(3) / 2
+    } else {
+      dy <- resolution(data$y, FALSE) / sqrt(3) / 2 * 1.15
```
clauswilke, line 72: Do we have an explanation for the magic numbers here (sqrt(3), 2, 1.15)? I realize they're just copied over from the old code, but some comment would be great.

Also, the factor 1.15 is missing in line 70, and this is different from the old code. Could you add a comment explaining the difference between line 70 and line 72?

Otherwise looks good.

clauswilke (approved):

R/geom-hex.r
thomasp85: I think it is simply a matter of calculating the relation between the height and width of a regular hexagon. The 1.15 is there because resolution in the y-direction is a different concept to the binwidth (bins are shifted so they overlap in the y-axis, affecting the resolution)... I can add some comment on this of course

jeffcraggy:
Not sure what is happening here, but these examples do not run correctly.

https://ggplot2.tidyverse.org/reference/geom_hex.html

> ggplot(diamonds, aes(carat, price)) +  geom_hex()

**Warning message:
Computation failed in `stat_binhex()`:**

yutannihilation:
@jeffcraggy 
Hi, thanks for the report. You need to install hexbin package. I filed a new issue: https://github.com/tidyverse/ggplot2/issues/4736

jeffcraggy:
I see. Thought it came with the tidyverse ggplot2 object. Anyways I used the contour heat map . But thank you, he bin is a great feature

Jeff

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #4672, opened 2021-11-16
Convert size = NA to 0 before rendering guide_legend

### LINKED ISSUE #4559: Setting `size = NA` on a geom causes error in legend generation
This is a bit of a weird corner case I think, but I've found (on CRAN and on the github version as of today) that setting `size = NA` on several geoms causes guide generation to fail and therefore the plot to fail to generate. 

I'll use this data:

```r
df = data.frame(
  x = c(1, 2),
  group = c("a","b")
)
```

First, a reprex with `geom_tile()` (the `fill` mapping is needed so that a guide is generated):

``` r
library(ggplot2)
df = data.frame(
  x = c(1, 2),
  group = c("a","b")
)
ggplot(df, aes(y = 0, x = x, width = 1, height = 1, fill = group)) +
  geom_tile(size = NA)
#> Error in grid.Call.graphics(C_setviewport, vp, TRUE): non-finite location and/or size for viewport
```
<sup>Created on 2021-07-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

I would expect behavior more like setting `color = NA`; i.e. geoms (or outlines of geoms) would not be drawn but the plot would still generate. I've tracked this down to *somewhere* in the guide generation code. If you turn the legend off, you get a plot:

``` r
ggplot(df, aes(y = 0, x = x, width = 1, height = 1, fill = group)) +
  geom_tile(size = NA, show.legend = FALSE)
```

Which is what I would expect. Similar problems occur in other geoms; e.g. `geom_point()`:

``` r
ggplot(df, aes(y = 0, x = x, color = group)) +
  geom_point(size = NA)
#> Warning: Removed 2 rows containing missing values (geom_point).
#> Error in grid.Call.graphics(C_setviewport, vp, TRUE): non-finite location and/or size for viewport
```

Here the warning is expected, but I think the error is not --- I would have expected output like what happens if you set `color = NA`; i.e. a blank plot:

``` r
ggplot(df, aes(y = 0, x = x, color = group)) +
  geom_point(color = NA)
#> Warning: Removed 2 rows containing missing values (geom_point).
```

I believe the problem may be related to this line: https://github.com/tidyverse/ggplot2/blob/13c07309e5d6798277d5d0c6f76ead88b6112e40/R/guide-legend.r#L385

As when `size = NA` this will result in `key_size_mat` being all `NA`s, and then key widths and heights downstream end up all `NA`s. However I don't know enough about the surrounding code/context to suggest an appropriate solution.

Thanks!

thomasp85:
Fix #4559 

This is a simple fix that converts all instances of `NA` in the size aesthetic to `0`

### DIFF
--- a/R/guide-legend.r
+++ b/R/guide-legend.r
@@ -278,6 +278,10 @@ guide_geom.legend <- function(guide, layers, default_mapping) {
     # override.aes in guide_legend manually changes the geom
     data <- modify_list(data, guide$override.aes)
 
+    if (!is.null(data$size)) {
+      data$size[is.na(data$size)] <- 0
+    }
+
     list(
       draw_key = layer$geom$draw_key,
       data = data,
@@ -383,6 +387,7 @@ guide_gengrob.legend <- function(guide, theme) {
   )
 
   key_size_mat <- do.call("cbind", lapply(guide$geoms, function(g) g$data$size / 10))
+
   if (nrow(key_size_mat) == 0 || ncol(key_size_mat) == 0) {
     key_size_mat <- matrix(0, ncol = 1, nrow = nbreak)
   }
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Setting `size = NA` will no longer cause `guide_legend()` to error 
+  (@thomasp85, #4559)
+
 * Setting `stroke` to `NA` in `geom_point()` will no longer impair the sizing of
   the points (@thomasp85, #4624)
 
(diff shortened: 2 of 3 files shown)


R/guide-legend.r
```diff
@@ -278,6 +278,10 @@ guide_geom.legend <- function(guide, layers, default_mapping) {
     # override.aes in guide_legend manually changes the geom
     data <- modify_list(data, guide$override.aes)
 
+    if (!is.null(data$size)) {
```
paleolimbot, line 281: I forget if `data` is a data.frame here, but if it is, I wonder of `"size" %in% names(data)` is a bit better (lest `data` ever become a tibble)

paleolimbot (approved):
Great!

R/guide-legend.r
thomasp85: We have this pattern all over the place so once we begin to accept tibbles inside the code we will have a lot of places to fix up

PR #4674, opened 2021-11-17
Remove filename from the call to custom devices if file is present in the formals

### LINKED ISSUE #4539: Regression: ggsave() fails if `device` is function without `...`
Since a9b3b17bd92a0526deb8626586ec44a4b3e714aa, if a `device` is passed to `ggsave()` and this device lacks a `...` argument (at least if it also uses a `file` argument instead of `filename`), it produces an "unused argument" error.

``` r
ggplot2::ggsave("test.emf", ggplot2::ggplot(), device = devEMF::emf)
#> Saving 7 x 5 in image
#> Error in (function (file = "Rplot.emf", width = 7, height = 7, bg = "transparent", : unused argument (filename = "test.emf")
```

<sup>Created on 2021-07-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

thomasp85:
Fix #4539 

This PR make it so that only file _or_ filename is passed on as argument to the custom device

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Better compatibility of custom device functions in `ggsave()` 
+  (@thomasp85, #4539)
+
 * Strip padding in `facet_grid()` is now only in effect if `strip.placement = "outside"`
   _and_ an axis is present between the strip and the panel (@thomasp85, #4610)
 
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Thanks, looks good!

PR #4676, opened 2021-11-18
Check if transformed limits in binning scales are valid before assigning them

### LINKED ISSUE #4510: scale_*_steps (presumably binned_scales) doesn't like 0.01 with trans="log10"
I would like to bin my colors using a log10 transformation, and it seems that if the number 0.01 is included in the data, the scaling breaks with a warning:
```
Warning message:
In self$trans$transform(limits) : NaNs produced
```
This causes the transformation to happen, but the legend doesn't show up.

Here is an example of what doesn't work:
```r
ggplot(data = data.frame(x = 1:5, 
                         y = c(0.01, 0.43, 0.46,
                               0.05, 0.17))) +
  geom_point(aes(x = x, y = y, color = y)) +
  scale_color_steps(trans = "log10")
```

Here are 3 examples that do work, which makes me suspect it is a bug:

```r
ggplot(data = data.frame(x = 1:5, 
                         y = c(0.0101, 0.43, 0.46,
                               0.05, 0.17))) +
  geom_point(aes(x = x, y = y, color = y)) +
  scale_color_steps(trans = "log10") 

ggplot(data = data.frame(x = 1:5, 
                         y = c(0.00999, 0.43, 0.46,
                               0.05, 0.17))) +
  geom_point(aes(x = x, y = y, color = y)) +
  scale_color_steps(trans = "log10")

ggplot(data = data.frame(x = 1:6, 
                         y = c(0.01, 0.43, 0.46,
                               0.05, 0.17, 0.09))) +
  geom_point(aes(x = x, y = y, color = y)) +
  scale_color_steps()
```
I can speculate on why this is happening, but I'm unclear if there's an easy way to eliminate this problem. 

One fix is to use "log1p" instead of "log10". It took awhile for me to pinpoint the issue and discover the solution. Even if a fix for 0.01 can't be included, it would be helpful to add some language in the binned_scale (or scale_*_steps) documentation.

thomasp85:
Fix #4510 

This PR fixes a rare bug when the scale derives new limits to make the terminal bins equal size and these limits ends up being invalid for the given transformation (e.g. `log10`)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Binning scales are now more resilient to calculated limits that ends up being
+  `NaN` after transformations (@thomasp85, #4510)
+
 * Strip padding in `facet_grid()` is now only in effect if `strip.placement = "outside"`
   _and_ an axis is present between the strip and the panel (@thomasp85, #4610)
 
--- a/R/scale-.r
+++ b/R/scale-.r
@@ -1083,11 +1083,12 @@ ScaleBinned <- ggproto("ScaleBinned", Scale,
             new_limits[1] <- breaks[1]
             breaks <- breaks[-1]
           }
-          limits <- new_limits
         } else {
           bin_size <- max(breaks[1] - limits[1], limits[2] - breaks[1])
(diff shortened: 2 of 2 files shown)


yutannihilation (approved):
Looks good.

PR #4682, opened 2021-11-24
Check for range == NULL - happens if no data has been added to plot

### LINKED ISSUE #4509: empty plot with secondary axis fails
Hi,

I randomly encountered a bug while working with the ggplot package. I was trying to create an empty plot with secondary (so other functions in my package can fill it up later). However, adding the secondary axis to an empty ggplot object throws an error if the object is called.

E.g.,:
```r
plot <- ggplot2::ggplot() 
plot <- plot + ggplot2::scale_y_continuous(
  name = "ylab", limits = c(0, 1), 
  sec.axis = ggplot2::sec_axis(~ ., name = "ylab2", breaks = c(0, 1), labels = c(0, 2)))
```
which throws this error when calling ``plot``
> Error in seq.default(range[1], range[2], length.out = self$detail) : 
>   'from' must be of length 1

However, adding a geom (either before or after adding the secondary axis) fixes the issue:
```r
plot + ggplot2::geom_point(data = data.frame(x = 0.5, y = 0.5), mapping = ggplot2::aes(x = x, y = y))
```

Best,
Frantisek

thomasp85:
Fix #4509 

This adds a simple check in secondary axes for whether there is any range available to make the monotonity check on

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Fixed a bug throwing errors when trying to render an empty plot with secondary
+  axes (@thomasp85, #4509)
+
 * Axes are now added correctly in `facet_wrap()` when `as.table = FALSE`
   (@thomasp85, #4553)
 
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good.

PR #4683, opened 2021-11-24
Add non_missing_aes to geom_tile

### LINKED ISSUE #4495: GeomTile should have a non_missing_aes with xmin, xmax, ymin, ymax
Similarly to `GeomBar` and `GeomCol` that use `GeomRect` to draw, I believe `GeomTile` too should  declare a `non_missing_aes` member for xmin, xmax, ymin, ymax variables.

For example in `GeomBar` I see this:
```
  # These aes columns are created by setup_data(). They need to be listed here so
  # that GeomRect$handle_na() properly removes any bars that fall outside the defined
  # limits, not just those for which x and y are outside the limits
  non_missing_aes = c("xmin", "xmax", "ymin", "ymax"),
```

I stumbled to this, using `geom_bin2d` (which uses `GeomTile`) where it may produce data with NA in any of xmin, xmax, ymin, ymax. Here is a reprex showing that result grob contains NAs.

``` r
library(ggplot2)
p <- ggplot(diamonds, aes(x, y)) + xlim(4, 10) + ylim(4, 10) + geom_bin2d()
g <- layer_grob(p, 1)[[1]]
#> Warning: Removed 478 rows containing non-finite values (stat_bin2d).
any(is.na(g$x)|is.na(g$y))
#> [1] TRUE
```

<sup>Created on 2021-05-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Adding the `non_missing_aes` fixes this.

thomasp85:
Fix #4495

Correctly declare min, xmax, ymin, and ymax as aes needing values

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # ggplot2 (development version)
 
 * `geom_tile()` now correctly recognises missing data in `xmin`, `xmax`, `ymin`,
-  and `ymax` (@thomasp85, #4495)
+  and `ymax` (@thomasp85 and @sigmapi, #4495)
 
 * Axes are now added correctly in `facet_wrap()` when `as.table = FALSE`
   (@thomasp85, #4553)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* `geom_tile()` now correctly recognises missing data in `xmin`, `xmax`, `ymin`,
+  and `ymax` (@thomasp85, #4495)
+
 * Axes are now added correctly in `facet_wrap()` when `as.table = FALSE`
   (@thomasp85, #4553)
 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* `geom_tile()` now correctly recognises missing data in `xmin`, `xmax`, `ymin`,
+  and `ymax` (@thomasp85, #4495)
```
yutannihilation, line 4: Shouldn't we credit @sigmapi as well for the fix?

yutannihilation (approved):
Looks good.

NEWS.md
thomasp85: yeah, good catch

thomasp85, line 4:
```suggestion
  and `ymax` (@thomasp85 and @sigmapi, #4495)
```

Issue #4681, opened 2021-11-24
Dataset licenses


bkmgit:
Typically datasets are released under different licenses than software. It may be helpful to add license information to the data in https://github.com/tidyverse/ggplot2/tree/main/data-raw

hadley:
Basically, my position is that under US copyright law, datasets are not copyrightable, so the software licenses don't apply anyway. But I don't think it's worth the time to clearly articulate or communicate this position and the status quo is fine for 99.999% of users.

PR #4686, opened 2021-11-27
License information

bkmgit:
There is no original source for the data. OpenML indicates it is in the public domain,https://www.openml.org/d/42225

Related issue https://github.com/tidyverse/ggplot2/issues/4681

### DIFF
--- a/R/data.R
+++ b/R/data.R
@@ -1,7 +1,8 @@
 #' Prices of over 50,000 round cut diamonds
 #'
 #' A dataset containing the prices and other attributes of almost 54,000
-#'  diamonds. The variables are as follows:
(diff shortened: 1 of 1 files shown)


clauswilke:
If there is some sort of source that can be pointed to for the statement that the dataset is in the public domain then it would be good to add that.

bkmgit:
@clauswilke Thanks.  Cannot find any other source of information on the origin of that data. The relevant commit https://github.com/tidyverse/ggplot2/commit/4c678917393c807090f8c1168ac8d6483390a529 was made 7 years ago by @hadley

hadley:
I don't think we can add anything here without being sure of the provenance. However, in the US data is not copyrightable.

PR #4697, opened 2021-12-14
Use pak in R-CMD-check.yaml

yutannihilation:
It seems pak-based workflow is ready to use now. Let's see if this works...

> > This is now possible with `?ignore` and `?ignore-before-r`: see some GHA oriented docs here-ish: https://github.com/r-lib/actions/tree/v2/setup-r-dependencies#ignoring-optional-dependencies-that-need-a-newer-r-version
https://github.com/r-lib/pak/issues/290#issuecomment-992400476

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -33,9 +33,9 @@ jobs:
           - {os: ubuntu-18.04,   r: '3.3',   vdiffr: false, xref: false}
 
     env:
-      # R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
+      R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
       # don't treat missing suggested packages as error
-      _R_CHECK_FORCE_SUGGESTS_: false
+      # _R_CHECK_FORCE_SUGGESTS_: false
       # Some packages might unavailable on the older versions, so let's ignore xref warnings
       _R_CHECK_RD_XREFS_: ${{ matrix.config.xref }}
       # Runs vdiffr test only on the latest version of R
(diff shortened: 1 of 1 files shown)


yutannihilation:
This should be ready to merge now.

.github/workflows/R-CMD-check.yaml
```diff
@@ -50,54 +42,46 @@ jobs:
...
-        if: runner.os == 'Linux'
-        run: |
-          while read -r cmd
-          do
-            eval sudo $cmd
-          done < <(Rscript -e 'writeLines(remotes::system_requirements("ubuntu", "18.04"))')
       - name: Install system dependencies on macOS
         if: runner.os == 'macOS'
         run: |
           # XQuartz is needed by vdiffr
```
thomasp85, line 55: Is this still needed? I _think_ XQuartz is provided by default now...

yutannihilation: Oh, really? Let's try removing this.

yutannihilation:
Hm, glue now requires R >= 3.4. Since we depend on glue very much, it might be a good time to consider dropping the support for R 3.3.

thomasp85:
Just drop it from the test suite - no need to add a hard dependency on 3.4 in the DESCRIPTION...

yutannihilation:
I see. Considering the users with R 3.3 still can install ggplot2 if they install some older version of glue, just dropping it from the test suite is enough.

thomasp85:
Yes, that is my thinking

yutannihilation:
Failures seems because of this problem. I'll rerun CI when the next version of sf gets released on CRAN.

https://github.com/r-spatial/sf/issues/1899

yutannihilation:
Closing in favor of  #4747

PR #4707, opened 2021-12-28
Duplicated aes warning with multiple modified aesthetics

teunbrand:
This is a simple fix for some warnings that are thrown when multiple modified aesthetics are used.
The warnings could be triggered with the reprex below. With this PR, these warnings won't be triggered.

``` r
# Reprexed before PR
library(ggplot2)
library(colorspace)

df <- data.frame(x = 1, y = 1, lab = "test")

ggplot(df, aes(x, y, label = lab)) +
  geom_label(
    aes(colour = stage(lab, after_scale = darken(colour, 0.3)),
        fill   = after_scale(lighten(colour, 0.5)))
  )
#> Warning: Duplicated aesthetics after name standardisation: NA
#> Warning: Duplicated aesthetics after name standardisation: NA
```

Effectively, the two chunks of code below take the names of the (modified) aesthetics twice.

https://github.com/tidyverse/ggplot2/blob/c89c265a57fd71f8a0288ce81037296aadc0a012/R/geom-.r#L146

https://github.com/tidyverse/ggplot2/blob/759c63c2fd9e00ba3322c1b74b227f63c98d2e06/R/aes.r#L168-L169

This surprisingly works well when there is only a single modified aesthetic, but not so much with multiple ones.

The following erroneous code (duplicated colour/color):
``` r
ggplot(df, aes(x, y, label = lab)) +
  geom_label(
    aes(colour = stage(lab, after_scale = darken(colour, 0.3)),
        color  = after_scale(lighten(color, 0.5)))
  )
```
Used to throw these warnings:
```
#> Warning: Duplicated aesthetics after name standardisation: colour
#> Warning: Duplicated aesthetics after name standardisation: NA
#> Warning: Duplicated aesthetics after name standardisation: NA
```
And now throws the following warnings:
```
#> Warning: Duplicated aesthetics after name standardisation: colour
#> Warning: Duplicated aesthetics after name standardisation: colour
#> Warning: Duplicated aesthetics after name standardisation: colour
```


hadley:
Could you please add a simple unit test that uses `expect_snapshot_warning()` to capture the warning messages?

teunbrand:
Sure, are you OK with bumping the {testthat} requirement to >=3.1.2 for this test?

hadley:
Yes, that's fine.

teunbrand:
Thanks, I also bumped rlang to >=1.0.0 due to a recently added use of `on_load()`.

tests/testthat/test-aes-calculated.r
```diff
@@ -49,3 +49,19 @@ test_that("make_labels() deprases mappings properly", {
   expect_identical(make_labels(aes(x = 1)), list(x = "x"))
   expect_identical(make_labels(aes(x = NULL)), list(x = "x"))
 })
+
+test_that("staged aesthetics warn appropriately for duplicated names", {
+  # Test should *not* report `NA` as the duplicated aes (#4707)
+  df <- data.frame(x = 1, y = 1, lab = "test")
+  expr <- substitute(
```
hadley, line 56: Since I don't think you're actually doing any substitution, `quote()` would be more appropriate here. But I think you could also just do `p <- ggplot() + ...` in the `expect_snapshot_warning()`.

tests/testthat/test-aes-calculated.r
```diff
@@ -49,3 +49,19 @@ test_that("make_labels() deprases mappings properly", {
...
+    ggplot(df, aes(x, y, label = lab)) +
+      geom_label(
+        aes(colour = stage(lab, after_scale = colour),
+            color  = after_scale(color))
+      )
+  )
+  # One warning in plot code due to evaluation of `aes()`
+  expect_snapshot_warning(p <- eval(expr))
+  # Two warnings in building due to `stage()`/`after_scale()`
```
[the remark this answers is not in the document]
teunbrand: Ah indeed! The second of those warnings only occurred when `ggplot_build(p)` was evaluated in the console, due to automatically printing the result. I've now disabled the legend, which caused that extra warning.

tests/testthat/test-aes-calculated.r
teunbrand: Good points, I wasn't aware of this, so thanks!

hadley (approved):
Thanks! The last thing is to add a bullet to the top of `NEWS.md`. It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

PR #4709, opened 2022-01-03
Add options to guide_colourstep

tomsaunders98:
The guide_coloursteps function does not pass on raster or nbin options to the guide_colourbar function which work correctly (with ragg) when passed on. I'm not entirely sure if there is a reason why they aren't passed on but I thought it made sense to include them as options. 

I also updated the associated documentation for guide_colourbar to include explainers of the nbin and raster params, (taken from guide-colorbar).


hadley:
These don't seem needed to be me because a binned colour scale is fundamentally discrete?

PR #4720, opened 2022-01-27
Corrects bins / binwidth override documentation

### LINKED ISSUE #4651: `geom_contour()` documentation states false precedence of bin and binwidth parameters
Right now the documentation for `geom_contour()` and friends states that the `bin` argument is overridden by the `binwidth` parameter. However, this seems not to be the case. 

``` r
library(ggplot2)


# Defining bins
v + geom_contour(bins = 5)
```

``` r
# Defining binwidth
v + geom_contour(binwidth = 0.001)
```

``` r
# Defining both. Per documentation, binwidth should 
# take precedence and this should be the same as the second plot
v + geom_contour(binwidth = 0.001, bins = 5)
```

<sup>Created on 2021-10-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

The issue lays on the `contour_breaks()` function, which computes the binwidth from the bins if bins is not NULL: 

https://github.com/tidyverse/ggplot2/blob/759c63c2fd9e00ba3322c1b74b227f63c98d2e06/R/stat-contour.r#L158-L170 

I believe that previous versions of `geom_contour()` did respect the documentation. Should the documentation be changed to reflect the new precedence or should the code be changed to reflect the documentation?

eliocamp:
Closes #4651

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,6 +53,10 @@
   
 * `geom_contour()` now accepts a function in the `breaks` argument (@eliocamp, #4652).
 
+* Updated documentation for `geom_contour()` to correctly reflect argument 
+precedence between `bins` and `binwidth`. (@eliocamp, #4651)
+
+
 # ggplot2 3.3.5
 This is a very small release focusing on fixing a couple of untenable issues 
 that surfaced with the 3.3.4 release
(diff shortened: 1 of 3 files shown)


hadley (approved):

hadley:
Thanks @eliocamp!

PR #4734, opened 2022-02-12
Fix misalignment in geom_dotplot when stackratio != 1 and stackdir != "up"

### LINKED ISSUE #4614: Dots misaligned in geom_dotplot
---

Hello!
I've ran into a problem where the dots of geom_dotplot are not aligned correctly. The problem was described [here](https://stackoverflow.com/questions/51406933/how-to-use-ggplot2s-geom-dotplot-with-symmetrically-spaced-and-separated-dots) years ago, but it seems to still be here (and I couldn't find any issue about it so I assume nobody ever filed one). The link also provides a possible solution but no one seems to have a clue if this solution affects anything elsewhere. 

```r
library(ggplot2)

data <- data.frame(x = rep(seq(2018, 2021, 1), 15), 
                   y = sample(seq(3, 5, .125), 60, 
                              replace = T))

ggplot(data, aes(factor(x), y)) +
  geom_dotplot(binaxis = "y", stackdir = "center", stackratio = 1)

ggplot(data, aes(factor(x), y)) +
  geom_dotplot(binaxis = "y", stackdir = "center", stackratio = 2) #points shifted to left

ggplot(data, aes(factor(x), y)) +
  geom_dotplot(binaxis = "y", stackdir = "center", stackratio = 0.5) #points shifted to right

```

mjskay:
This PR fixes #4614.

## The problem

For `geom_dotplot()` / the `dotstack` grob, the `(1 - x$stackratio) / 2` adjustment on these lines...

https://github.com/tidyverse/ggplot2/blob/c89c265a57fd71f8a0288ce81037296aadc0a012/R/grob-dotstack.r#L34-L42

...is there because otherwise when `stackratio != 1` and `stackdir = "up"`, the base of the first dot in a stack will no longer touch the baseline. This is because `stackratio` will expand (or contract) the dot stack away (or towards) the origin. 

However, currently this adjustment only works correctly for `stackdir = "up"`. For example, here is `stackdir = "down"` (for other examples, see #4614):

```r
ggplot(data.frame(x = c(rep(1, 3), rep(2, 2))), aes(x)) +
  geom_dotplot(binwidth = 0.5, alpha = 0.5, stackdir = "down", stackratio = 1.5) +
  coord_fixed() +
  ylim(-2, 2) +
  xlim(0, 6)
```
image

Notice how the top dot in each stack does not touch `y = 0` as it should.

## The solution

The solution is that the `(1 - x$stackratio) / 2` adjustment needs to depend on `stackdir`:

- For `"up"`, it should remain `(1 - x$stackratio) / 2`
- For `"down"`, it should be reversed: `-(1 - x$stackratio) / 2`
- For `"center"` and `"centerwhole"`, it is unnecessary, and should be `0`.

This PR implements that change, and includes test cases to ensure the adjustment is correct when `stackratio < 1` and when `stackratio > 1`.

### DIFF
--- a/tests/testthat/test-geom-dotplot.R
+++ b/tests/testthat/test-geom-dotplot.R
@@ -125,6 +125,34 @@ test_that("geom_dotplot draws correctly", {
     ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "centerwhole") + coord_flip()
   )
 
+  # Stacking methods with stackratio < 1
+  expect_doppelganger("stack up, stackratio = 0.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "up", stackratio = 0.5)
+  )
+  expect_doppelganger("stack down, stackratio = 0.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "down", stackratio = 0.5)
+  )
+  expect_doppelganger("stack center, stackratio = 0.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "center", stackratio = 0.5)
+  )
+  expect_doppelganger("stack centerwhole, stackratio = 0.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "centerwhole", stackratio = 0.5)
+  )
+
+  # Stacking methods with stackratio > 1
+  expect_doppelganger("stack up, stackratio = 1.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "up", stackratio = 1.5)
+  )
+  expect_doppelganger("stack down, stackratio = 1.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "down", stackratio = 1.5)
+  )
+  expect_doppelganger("stack center, stackratio = 1.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "center", stackratio = 1.5)
+  )
+  expect_doppelganger("stack centerwhole, stackratio = 1.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "centerwhole", stackratio = 1.5)
+  )
+
   # Binning along x, with groups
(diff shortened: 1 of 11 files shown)


tests/testthat/test-geom-dotplot.R
```diff
@@ -125,6 +125,34 @@ test_that("geom_dotplot draws correctly", {
     ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "centerwhole") + coord_flip()
   )
 
+  # Stacking methods with stackratio < 1
```
hadley, line 128: Can you please make this a separate test? (The existing huge test unfortunately isn't a good pattern to follow). And can you please also create the smallest possible dataset needed to demonstrate the problem?

tests/testthat/test-geom-dotplot.R
```diff
@@ -125,6 +125,34 @@ test_that("geom_dotplot draws correctly", {
...
+  expect_doppelganger("stack up, stackratio = 0.5",
+    ggplot(dat, aes(x)) + geom_dotplot(binwidth = .4, stackdir = "up", stackratio = 0.5)
+  )
+  expect_doppelganger("stack down, stackratio = 0.5",
```
hadley, line 132: I think one test with non-default stack ratio would be adequate. Visual tests are expensive to maintain so we really want to keep them as minimal as possible.

hadley (commented):
Thanks for working on this 😄. As well addressing the comments, can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

tests/testthat/test-geom-dotplot.R
mjskay: Sure, done --- I pulled out the four tests for `stackratio = 1.5` into a separate test with a single doppelganger on a simplified dataset that should be sensitive to problems with any of the four values of `stackdir`. Dunno if you would rather those be four separate small doppelgangers or one merged doppelganger as I have done.

tests/testthat/test-geom-dotplot.R
mjskay: Done.

mjskay:
Happy to, thanks for the feedback! I've added a NEWS entry and updated the tests based on your comments.

hadley (approved):
Visual test is perfect, thanks!

PR #4737, opened 2022-02-21
Use the full range of the palette for binned viridis

### LINKED ISSUE #4372: Scale_*_viridis_b not using full width of palette
When using scale_(fill/colour)_viridis_b, the colors don't go all the way to the end of the palette, i.e. very-yellow yellow).
Is it intended to do that? How to change this behavior if it is?

```r
library(ggplot2)

ggplot(data = ggplot2::diamonds) +
  geom_bar(aes(x = cut, fill = depth, group = depth)) +
  scale_fill_viridis_b(n.breaks = 4)
```
image

```r
scales::show_col(
  viridis::viridis_pal()(4)
)
```
Full palette (4 steps) 
image

gregleleu:
Resolves #4372 
Adding a parameter to use the full width of the scale in scale_*_viridis_b functions.


R/scale-viridis.r
```diff
@@ -123,15 +124,23 @@ scale_fill_viridis_c <- function(..., alpha = 1, begin = 0, end = 1,
 scale_colour_viridis_b <- function(..., alpha = 1, begin = 0, end = 1,
                                    direction = 1, option = "D", values = NULL,
                                    space = "Lab", na.value = "grey50",
-                                   guide = "coloursteps", aesthetics = "colour") {
-  binned_scale(
-    aesthetics,
-    "viridis_b",
-    gradient_n_pal(
+                                   guide = "coloursteps", aesthetics = "colour", pal_full_range = FALSE) {
+  if(pal_full_range) {
```
hadley, line 128: Why does this need to be an option? Wouldn't you always want to use the full range?

gregleleu: as per our discussion in #4372 with @clauswilke @yutannihilation: both ways of doing it could make sense

hadley: Can you please provide an argument as to why the current behaviour would ever be desired?

gregleleu: I can't say that I can. Maybe @clauswilke @yutannihilation?

hadley: @thomasp85 should issue a final ruling, but I'm pretty sure the original behaviour was a mistake, and we should just always use the full range. We'll need to flag this as a visual change in the NEWS.

thomasp85: Agree - we should just go with the approach in this PR. The original implementation was an error on my part

thomasp85:
@gregleleu do you want to finish this off, given the feedback above?

gregleleu:
@thomasp85 Sure, so removing the `pal_full_range` argument I had added and making this the new default behavior + mention in the NEWS, is that right?

thomasp85:
@gregleleu sorry for missing the question — Yes exactly. This PR will implement the one true behaviour

thomasp85 (approved):

thomasp85:
Thanks @gregleleu — as an advise for the future begin to develop PRs in their own separate branch instead of in main

thomasp85:
The first basically. You keep the main branch as a complete copy of the upstream main and only use it as base for new branches to make PRs from.

The usethis package has some helpful functions for this workflow. You basically use `usethis::pr_init(<branch name>)` to create a new branch for a PR, then upon the first `usethis::pr_push()` a PR is set up against the upstream main branch etc...

PR #4757, opened 2022-03-15
Orientation-aware key glyphs

### LINKED ISSUE #4732: Legend glyphs for geom_pointrange do not follow their plotted direction
*(This is super similar to this past issue https://github.com/tidyverse/ggplot2/issues/1389, but the addition of bidirectional geoms in v3.3.0 have changed the circumstances of that original issue)*

---
 
Since ggplot2 3.3.0, geoms like `geom_boxplot()` and `geom_pointrange()` work with either the x or the y aesthetic, thus eliminating the need for packages like {ggstance} for horizontal point ranges:

```r
library(ggplot2)

ggplot(df, aes(x = x, y = y, color = factor(x))) +
  geom_pointrange(aes(xmin = x - 1, xmax = x + 1))
```

Currently, however, the legend keys for point ranges show vertical lines regardless of the direction of the actual point range. There are ways to adjust the legend keys by either [modifying some of the underlying grid grobs](https://stackoverflow.com/a/41475858/120898), or by adding an empty `stat_summary()` layer in combination with `geom_point()` and `geom_line()` ([example](https://stackoverflow.com/a/40229390/120898)), but these are fairly hacky.

In a previous issue #1389, [the official recommendation from Hadley was to use **ggstance**](https://github.com/tidyverse/ggplot2/issues/1389#issuecomment-336525032), but now that `geom_pointrange()` can go an any direction now since v3.3.0, it would be neat if the legend glyphs matched the direction of the geoms for things like point ranges and box plots. I have no idea how feasible or difficult that would be though.

Thanks!

mjskay:
This PR closes #4732 by providing orientation-aware key glyphs for `geom_boxplot()`, `geom_crossbar()`, `geom_pointrange()`, and `geom_linerange()`.

For example, **before these changes**:

``` r
library(ggplot2)
library(patchwork)

df <- data.frame(
  middle = 1:2,
  lower = 0:1,
  upper = 2:3,
  min = -1:0,
  max = 3:4,
  group = c("a","b")
)

p <- ggplot(df, aes(
  x = middle,
  y = group,
  xmiddle = middle,
  xlower = lower,
  xupper = upper,
  xmin = min,
  xmax = max,
  color = group
))

(p + geom_boxplot(stat = "identity")) +
  (p + geom_crossbar()) +
  (p + geom_linerange()) +
  (p + geom_pointrange())
```

<sup>Created on 2022-03-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

**After these changes**:

Specifically, this PR:

- modifies `draw_key_boxplot()`, `draw_key_crossbar()`, and `draw_key_pointrange()` to be orientation-aware
- provides a new key glyph, `draw_key_linerange()`, which is an orientation-aware version of `draw_key_path()` (because `draw_key_path()` is used by other geoms whose legends should not be orientation-aware, I could not simply make `draw_key_path()` orientation aware).
- changes the key glyph of `geom_linerange()` to be `draw_key_linerange()` instead of `draw_key_path()`.
- adds tests for horizontal ~~and vertical~~ legends. ~~This might be too many graphical tests; I'm happy to reduce them --- perhaps by removing the vertical tests (the old behavior) and just leaving the horizontal ones.~~ I have condensed the visual tests down to just two for horizontal legends; I'm not sure there's a clear way to reduce them to fewer.

This PR **does not** add orientation-aware key glyphs to `geom_errorbar()` or `geom_errobarh()`, as these geoms consist of both horizontal and vertical lines, so it seemed reasonable to leave them as is. In fact, `geom_errorbarh()` already has the same key glyph as `geom_errorbar()` rather than one of the opposite orientation, corroborating my sense that the legends for these two geoms do not need to be orientation-aware.


hadley (approved):
Looks great, thanks @mjskay

Issue #4752, opened 2022-03-04
Support `rlang` dynamic dots 


mone27:
At the moment ggplot is using base `...`, and `rlang` dynamic dots are not supported [https://rlang.r-lib.org/reference/dyn-dots.html](https://rlang.r-lib.org/reference/dyn-dots.html).

I think this could be a good addition to ggplot as it add the support of:
 - splice arguments 
 -  name injection 
 - ignoring trailing commands. 
 At the same time it should not change the current features of ggplot.

The implantation should also be quite straightforward: replace `list(...)` with `list2(...)`

At the moment I can find that dynamic dots is supported only in [`guides`](https://github.com/tidyverse/ggplot2/blob/759c63c2fd9e00ba3322c1b74b227f63c98d2e06/R/guides-.r#L66).

If there is no reason in not having this feature I can try to make a PR for it.

hadley:
This seems like a good idea to me. A PR would be great just to see how much code this would affect.

PR #4764, opened 2022-03-18
Adds support of `rlang::list2` dynamic dots

mone27:
Migrates the dynamic dots management (`...`) from `base::list` to `rlang::list2`, see #4752

The functions modified are the one of the Public API that accepts `...` as arguments.

In total 89 functions have been update, they include:

- `geom_*`
- `stat_*`
- `lims`, `expand_limits`
- `aes_`, `aes_string`
- `coord_map`
- `annatation`, `annotation_map`, `annotation_logticks`

Here there is the script that I have used to find the functions that use `...` with `list` [https://gist.github.com/mone27/285f022b9efbb4bc244acb6622f3411e](https://gist.github.com/mone27/285f022b9efbb4bc244acb6622f3411e)

The PR should be complete, possible additional steps are:

- document the use of `rlang` `...` [https://rlang.r-lib.org/reference/dyn-dots.html#add-dynamic-dots-support-in-your-functions](https://rlang.r-lib.org/reference/dyn-dots.html#add-dynamic-dots-support-in-your-functions). However, in `ggplot` most of the `...` are actually not documented.
- add some test for the dynamic dots support


hadley:
This seems like a reasonable idea to me. It's probably fine to not worry about the documentation (although can you please add a news bullet) and even the tests are iffy, because what you really want to test is that every user facing `...` uses dynamic-dots, and I don't see an cheap/easy way to test that.

yutannihilation:
If the intention is to prohibit the use of `list` completely like we do on `data.frame`, we can add a test here:

https://github.com/tidyverse/ggplot2/blob/main/tests/testthat/test-prohibited-functions.R

(But, as this test is a bit tricky, we can leave it for later)

hadley (approved):

Issue #4783, opened 2022-03-31
Non-linear scale changes geom_density values


davidchall:
It is often useful to use `geom_density()` and `geom_function()` together, to compare data to a model. This is even an example in the `geom_function()` [man page](https://ggplot2.tidyverse.org/reference/geom_function.html#ref-examples). However, the density values output by `geom_density()` change when a logarithmic axis is used, making such comparisons difficult.

I suspect that `geom_density()` is performing the kernel density estimation upon the data _after_ the scale transformation has been applied. This will yield smoother curves in the logarithmic scale. However, it also reduces the distance between points, and this increases the density value.

If this is what's going on, perhaps we can find a way to apply an inverse transformation to the density values?

``` r
library(tidyverse)

p <- tibble(x = rnorm(1000, mean = 10)) %>%
  ggplot(aes(x)) +
  geom_density() +
  geom_function(fun = dnorm, args = list(mean = 10), color = "red")

p + scale_x_continuous(limits = c(1, NA))
```

![](https://i.imgur.com/khtrOzH.png)

``` r
p + scale_x_log10(limits = c(1, NA))
```

![](https://i.imgur.com/eztvayq.png)

<sup>Created on 2022-03-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

davidchall:
Related: https://github.com/r-lib/scales/issues/322

mjskay:
Right, density functions need to have the Jacobian correction applied when being transformed. In this case, given a density function `f(x)` (here `dnorm(x)`) and a scale transformation function `g(x)` (here `log(x, base = 10)`), you must multiply `f` by the absolute value of derivative of the inverse of `g` applied to the transformed `x` values:

![CodeCogsEqn (1)](https://user-images.githubusercontent.com/6345019/161464084-75314246-4bb5-4c5b-9294-e5068ad272b3.svg)

In this case *g*(*x*) is log<sub>10</sub>(*x*) and \[*g*<sup>-1</sup>\]'(*x*) is 10<sup>*x*</sup>log(10) so this simplifies to f(*x*)|*x*log(10)|. You can apply this correction manually:

``` r
library(tidyverse)

set.seed(1234)

tibble(x = rnorm(1000, mean = 10)) %>%
  ggplot(aes(x)) +
  geom_density(aes(x)) +
  geom_function(fun = \(x) dnorm(x, mean = 10) * abs(x * log(10)), color = "red") +
[...]

PR #4785, opened 2022-03-31
Rescale geom_density for non-linear scales

davidchall:
This is a proof-of-concept to address #4783. There might be some context I'm missing here, so I'm interested to hear feedback before proceeding any further.

``` r
library(tidyverse)

p <- tibble(x = rnorm(1000, mean = 10)) %>%
  ggplot(aes(x)) +
  geom_density() +
  geom_function(fun = dnorm, args = list(mean = 10), color = "red")

p + scale_x_continuous(limits = c(1, NA))
```

``` r
p + scale_x_log10(limits = c(1, NA))
```

``` r
p + scale_x_sqrt(limits = c(1, NA))
```

<sup>Created on 2022-03-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -133,10 +133,12 @@ compute_density <- function(x, w, from, to, trans, bw = "nrd0", adjust = 1,
     sum(y * x_diff, na.rm = TRUE)
   }
 
-  total_pre <- integrate_density(dens$x, dens$y)
-  total_post <- integrate_density(trans$inverse(dens$x), dens$y)
+  if (trans$name != "identity") {
+    area_pre <- integrate_density(dens$x, dens$y)
+    area_post <- integrate_density(trans$inverse(dens$x), dens$y)
 
-  dens$y <- dens$y * total_pre / total_post
+    dens$y <- dens$y * area_pre / area_post
+  }
 
   new_data_frame(list(
     x = dens$x,
--- a/R/stat-density.r
+++ b/R/stat-density.r
@@ -92,15 +92,17 @@ StatDensity <- ggproto("StatDensity", Stat,
       range <- scales[[flipped_names(flipped_aes)$x]]$dimension()
     }
 
+    trans <- scales[[flipped_names(flipped_aes)$x]]$trans
+
     density <- compute_density(data$x, data$weight, from = range[1],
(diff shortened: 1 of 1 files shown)


R/stat-density.r
```diff
@@ -125,6 +127,17 @@ compute_density <- function(x, w, from, to, bw = "nrd0", adjust = 1,
...
+    x_diff <- diff(x)
+    x_diff <- c(x_diff[1], x_diff)
+    sum(y * x_diff, na.rm = TRUE)
+  }
+
+  total_pre <- integrate_density(dens$x, dens$y)
+  total_post <- integrate_density(trans$inverse(dens$x), dens$y)
+
+  dens$y <- dens$y * total_pre / total_post
+
```
has2k1, line 140: A few observations

1. If `stats::density` guarantees that the area is functionally 1, then `total_pre` is not needed.
2. Instead of `integrate_density` why not hide all this diligence under a `rescale_density` function.
3. Rescale the density only if the `trans` is not an identity transformation.

davidchall: Thanks, Hassan. For now, I'm just hoping to understand if the ggplot authors think we should be rescaling the density. I don't expect this will be the final implementation.

Having said that, you've raised some excellent points. I'll address them below:

> If `stats::density` guarantees that the area is functionally 1, then `total_pre` is not needed.

Unfortunately, `stats::density()` doesn't guarantee this. In fact, the output is a linear interpolation of a fast Fourier transform, so it's very unlikely the area equals 1. Given this, we need to take a ratio of the areas before and after the transformation.

> Instead of `integrate_density` why not hide all this diligence under a `rescale_density` function.

When it comes to integrating the area under the curve, there are quite a few choices to make (e.g., use `stats::integrate()`, use the [trapezoidal rule](https://en.wikipedia.org/wiki/Trapezoidal_rule), etc). I've just done a very simple approximation (the [Riemann sum](https://en.wikipedia.org/wiki/Riemann_sum)), and I fully expect this will be replaced. So my 1integrate_density()` function is acting as a placeholder of sorts.

> Rescale the density only if the trans is not an identity transformation.

I totally agree on this. We shouldn't waste CPU time when there's no transformation to be performed. I'm not sure of the best way to identify this though, so I'd love feedback on how to achieve this.

PR #4791, opened 2022-04-01
Re-run with dev roxygen2

hadley:


hadley:
There are a few spurious diffs because I have dev scales installed, but overall it looks good, I think.

PR #4796, opened 2022-04-07
Update errors, warnings, and messages to cli

thomasp85:
This PR updates all user messaging to use the cli framework for proper formating. This PR is almost a direct translation from the old behaviour, but it does include a few refactors to make the code simpler where possible, using e.g. the auto collapsing feature of cli.

The changes has led to a few failing tests. These tests have been updated to use the new `expect_snapshot_*()` functions to better handle messages with formating.

This PR will be followed an additional PR looking into better error message chaining to let users know where errors occurred during rendering


R/aes-evaluation.r
```diff
@@ -121,7 +121,7 @@ is_calculated <- function(x) {
   } else if (is.pairlist(x)) {
     FALSE
   } else {
-    abort(glue("Unknown input: {class(x)[1]}"))
+    cli::cli_abort("Unknown input: {.cls {class(x)[1]}}")
```
hadley, line 124: Did you add new tests if the errors weren't covered by existing tests?

R/aes.r
```diff
@@ -103,7 +103,7 @@ new_aesthetic <- function(x, env = globalenv()) {
 }
 new_aes <- function(x, env = globalenv()) {
   if (!is.list(x)) {
-    abort("`x` must be a list")
+    cli::cli_abort("{.arg x} must be a {.cls list}")
```
hadley, line 106: I don't think we use `.cls` for simple types.

R/aes.r
```diff
@@ -398,7 +397,7 @@ warn_for_aes_extract_usage_expr <- function(x, data, env = emptyenv()) {
   if (is_call(x, "[[") || is_call(x, "$")) {
     if (extract_target_is_likely_data(x, data, env)) {
       good_usage <- alternative_aes_extract_usage(x)
-      warn(glue("Use of `{format(x)}` is discouraged. Use `{good_usage}` instead."))
+      cli::cli_warn("Use of {.code {format(x)}} is discouraged. Use {.code {good_usage}} instead.")
```
hadley, line 400:
```suggestion
      cli::cli_warn(c(
        "Use of {.code {format(x)}} is discouraged."
        i = "Use {.code {good_usage}} instead."
      ))
```

?

R/annotation-custom.r
```diff
@@ -71,7 +71,10 @@ GeomCustomAnn <- ggproto("GeomCustomAnn", Geom,
   draw_panel = function(data, panel_params, coord, grob, xmin, xmax,
                         ymin, ymax) {
     if (!inherits(coord, "CoordCartesian")) {
-      abort("annotation_custom only works with Cartesian coordinates")
+      cli::cli_abort(c(
+              "{.fn annotation_custom} only works with Cartesian coordinates",
+        "i" = "Only use {.fn coord_cartesian} to modify the coordinate system"
+      ))
```
hadley, line 77:
```suggestion
      cli::cli_abort(c(
        "{.fn annotation_custom} only works with Cartesian coordinates",
        "i" = "Only use {.fn coord_cartesian} to modify the coordinate system"
      ))
```

I'm not sure I understand what the `i` bullet means

R/annotation-map.r
```diff
@@ -59,13 +59,13 @@ NULL
...
   if (!is.data.frame(map)) {
-    abort("`map` must be a data.frame")
+    cli::cli_abort("{.arg map} must be a {.cls data.frame}")
   }
   if (!is.null(map$lat)) map$y <- map$lat
   if (!is.null(map$long)) map$x <- map$long
   if (!is.null(map$region)) map$id <- map$region
   if (!all(c("x", "y", "id") %in% names(map))) {
-    abort("`map`must have the columns `x`, `y`, and `id`")
+    cli::cli_abort("{.arg map} must have the columns {.col x}, {.col y}, and {.col id}")
```
hadley, line 68: I wonder if `{.col {c("x", "y", "id")}}` is worth it? Or maybe pull out into separate variable?

R/autoplot.r
```diff
@@ -15,10 +15,6 @@ autoplot <- function(object, ...) {
 
 #' @export
 autoplot.default <- function(object, ...) {
-  abort(glue(
-    "Objects of type ",
-    glue_collapse(class(object), "/"),
-    " not supported by autoplot."
-  ))
+  cli::cli_abort("Objects of type {.cls {class(object)}} is not supported by autoplot.")
```
hadley, line 18:
```suggestion
  cli::cli_abort("Objects of class {.cls {class(object)[[1]]}} are not supported by autoplot.")
```

R/axis-secondary.R
```diff
@@ -160,7 +160,7 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
   # Inherit settings from the primary axis/scale
   init = function(self, scale) {
     if (self$empty()) return()
-    if (!is.function(self$trans)) abort("transformation for secondary axes must be a function")
+    if (!is.function(self$trans)) cli::cli_abort("transformation for secondary axes must be a function")
```
hadley, line 163:
```suggestion
    if (!is.function(self$trans)) cli::cli_abort("Transformation for secondary axes must be a function")
```

Could we make this refer to the original argument name?

R/axis-secondary.R
```diff
@@ -188,7 +188,7 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
 
     # Test for monotonicity
     if (length(unique(sign(diff(full_range)))) != 1)
-      abort("transformation for secondary axes must be monotonic")
+      cli::cli_abort("transformation for secondary axes must be monotonic")
```
hadley, line 191:
```suggestion
      cli::cli_abort("Transformation for secondary axes must be monotonic")
```

R/bin.R
```diff
@@ -1,6 +1,6 @@
 bins <- function(breaks, closed = "right",
                  fuzz = 1e-08 * stats::median(diff(breaks))) {
-  if (!is.numeric(breaks)) abort("`breaks` must be a numeric vector")
+  if (!is.numeric(breaks)) cli::cli_abort("{.arg breaks} must be a {.cls numeric} vector")
```
hadley, line 3: Should we update the code style at the same time? (i.e. to put in own `{}` block)

R/bin.R
```diff
@@ -50,18 +50,18 @@ bin_breaks <- function(breaks, closed = c("right", "left")) {
...
 bin_breaks_width <- function(x_range, width = NULL, center = NULL,
                              boundary = NULL, closed = c("right", "left")) {
-  if (length(x_range) != 2) abort("`x_range` must have two elements")
+  if (length(x_range) != 2) cli::cli_abort("{.arg x_range} must have two elements")
 
   # if (length(x_range) == 0) {
   #   return(bin_params(numeric()))
   # }
-  if (!(is.numeric(width) && length(width) == 1)) abort("`width` must be a numeric scalar")
+  if (!(is.numeric(width) && length(width) == 1)) cli::cli_abort("{.arg width} must be a {.cls numeric} scalar")
```
hadley, line 58:
```suggestion
  if (!(is.numeric(width) && length(width) == 1)) cli::cli_abort("{.arg width} must be a number")
```

R/bin.R
```diff
@@ -87,7 +87,10 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
   max_x <- x_range[2] + (1 - 1e-08) * width
 
   if (isTRUE((max_x - origin) / width > 1e6)) {
-    abort("The number of histogram bins must be less than 1,000,000.\nDid you make `binwidth` too small?")
+    cli::cli_abort(c(
```
hadley, line 90: Indenting

R/bin.R
```diff
@@ -102,11 +105,11 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
 
 bin_breaks_bins <- function(x_range, bins = 30, center = NULL,
                             boundary = NULL, closed = c("right", "left")) {
-  if (length(x_range) != 2) abort("`x_range` must have two elements")
+  if (length(x_range) != 2) cli::cli_abort("{.arg x_range} must have two elements")
 
   bins <- as.integer(bins)
   if (bins < 1) {
-    abort("Need at least one bin.")
+    cli::cli_abort("Need at least one bin.")
```
hadley, line 112: `bin` must be 1 or greater?

R/compat-plyr.R
```diff
@@ -193,7 +193,7 @@ revalue <- function(x, replace) {
     lev[match(names(replace), lev)] <- replace
     levels(x) <- lev
   } else if (!is.null(x)) {
-    abort("x is not a factor or character vector")
+    cli::cli_abort("{.arg x} is not a {.cls factor} or {.cls character} vector")
```
hadley, line 196:
```suggestion
    cli::cli_abort("{.arg x} must be a factor or character vector")
```

R/compat-plyr.R
```diff
@@ -239,14 +239,14 @@ as.quoted <- function(x, env = parent.frame()) {
   } else if (is.call(x)) {
     as.list(x)[-1]
   } else {
-    abort("Only knows how to quote characters, calls, and formula")
+    cli::cli_abort("Only knows how to quote {.cls character}, {.cls call}, and {.cls formula}")
```
hadley, line 242:
```suggestion
    cli::cli_abort("Must be a character vector, call, or formula")
```

R/coord-.r
```diff
@@ -63,28 +63,28 @@ Coord <- ggproto("Coord",
 
   render_fg = function(panel_params, theme) element_render(theme, "panel.border"),
 
-  render_bg = function(panel_params, theme) {
-    abort("Not implemented")
+  render_bg = function(self, panel_params, theme) {
+    cli::cli_abort("{.fn {snake_class(self)}} has not implemented a {.fn render_bg} method")
```
hadley, line 67: Is inlining the function name into the error message needed, or does it now appear in the call?

hadley (commented):
Done up to `coord-.R`. This seems like an ok place to stop for now so you can handle the feedback so far.

hadley:
(Also needs redocumenting to fix the build errors)

R/aes.r
thomasp85: hmm... but if we have something like `{.cls {class(x)}}` and `x` is a list, it will get the formatting and thus be inconsistent if we didn't apply it fully for the hardcoded situations...

R/bin.R
thomasp85: I don't have hard feelings - I avoided it to make the PR minimal, but I'm fine going through it all one more time

R/axis-secondary.R
thomasp85: we could... but it is a function call into an argument which makes it a bit clumsy.. e.g. you would usually do `scale_x_continuous(..., sac.axis = sec_axis(trans = <x>, ...))`

thomasp85: since `trans` is also an argument to the scale function you'd have to qualify the naming and I feel the prose version above is less clumsy than `The trans argument in sec_axis() passed to scale_x_continuous() must be a function` or something like that

R/annotation-custom.r
thomasp85: it's meant to say that if you want to change aspects of the coordinate system you must do it through a `coord_cartesian()` call and not with `coord_whatever()`...

I adopted it from the warnings from `coord_sf()` where the user is more likely to add e.g. `coord_cartesian(expand = FALSE` or something like that, but I agree it makes less sense here... should I just remove it

R/annotation-map.r
thomasp85: no strong opinions... spelling it out makes it a bit clearer what the message will end up looking like...

R/bin.R
thomasp85: I think `bins` can be derived from other arguments than `bin` so we shouldn't target that argument name directly

R/bin.R
```diff
@@ -87,7 +87,10 @@ bin_breaks_width <- function(x_range, width = NULL, center = NULL,
   max_x <- x_range[2] + (1 - 1e-08) * width
 
   if (isTRUE((max_x - origin) / width > 1e6)) {
-    abort("The number of histogram bins must be less than 1,000,000.\nDid you make `binwidth` too small?")
+    cli::cli_abort(c(
+            "The number of histogram bins must be less than 1,000,000.",
```
thomasp85, line 91:
```suggestion
      "The number of histogram bins must be less than 1,000,000.",
```

R/coord-.r
thomasp85: it doesn't with this PR, and I'm still not sure whether it will happen for anything but the layers once I refactor it... so I made the choice to have the other component errors identify the name in the errors

R/bin.R
thomasp85: I've now done this

R/aes-evaluation.r
hadley: They don't have to be, but it does determine how closely I'll read the messages.

R/aes.r
hadley: I think that's ok because the type of message usually only comes up in some fallback, where most of the time `x` will be an S3 object.

R/bin.R
hadley: It looks like this code path is only hit when `breaks` and `binwidth` are NULL, so I think it's only determined by `bins`.

R/facet-wrap.r
```diff
@@ -503,7 +453,7 @@ wrap_dims <- function(n, nrow = NULL, ncol = NULL) {
     nrow <- ceiling(n / ncol)
   }
   if (nrow * ncol < n) {
-    abort("The given dimensions cannot hold all panels. Please increase `ncol` or `nrow`")
+    cli::cli_abort("The provided dimensions cannot hold all panels. Please increase {.arg ncol} and/or {.arg nrow}")
```
hadley, line 456: Maybe?

```R
cli::cli_abort(c(
  "Need {n} panels, but together{.arg nrow} and({ncol}) only provide {nrow * ncol}",
  i = "Please increase {.arg ncol} and/or {.arg nrow}"
))

File: R/geom-sf.R [hunk 18]
```diff
@@ -126,11 +126,14 @@ GeomSf <- ggproto("GeomSf", Geom,
...
-  draw_panel = function(data, panel_params, coord, legend = NULL,
+  draw_panel = function(self, data, panel_params, coord, legend = NULL,
                         lineend = "butt", linejoin = "round", linemitre = 10,
                         arrow = NULL, na.rm = TRUE) {
     if (!inherits(coord, "CoordSf")) {
-      abort("geom_sf() must be used with coord_sf()")
+      cli::cli_abort(c(
+        "{.fn {snake_class(self)}} can only be used together with an sf coordinate system",
+        "i" = "Please only modify the coordinate system using {.fn coord_sf}"
+      ))
```
hadley, line 136:
```suggestion
      cli::cli_abort("{.fn {snake_class(self)}} can only be used with {.fn coord_sf}"))
```

R/geom-sf.R
```diff
@@ -175,9 +178,7 @@ sf_grob <- function(x, lineend = "butt", linejoin = "round", linemitre = 10,
   remove[is_other] <- detect_missing(x, c(GeomPolygon$required_aes, GeomPolygon$non_missing_aes))[is_other]
   if (any(remove)) {
     if (!na.rm) {
-      warning_wrap(
-        "Removed ", sum(remove), " rows containing missing values (geom_sf)."
-      )
+      cli::cli_warn("Removed {sum(remove)} rows containing missing values (geom_sf)")
```
hadley, line 181:
```suggestion
      cli::cli_warn("Removed {sum(remove)} row{s?} containing missing values (geom_sf)")
```

hadley (commented):
Done to the end of `R/grob-dotstack.r`

R/aes.r
thomasp85: ok - I'll update

R/bin.R
thomasp85: you are right - I'll update

R/geom-.r
```diff
@@ -135,12 +135,11 @@ Geom <- ggproto("Geom",
...
       if (length(nondata_modified) > 0) {
-        msg <- glue(
-          "Modifiers must return valid values. Problematic aesthetic(s): ",
-          glue_collapse(vapply(nondata_modified, function(x) glue("{x} = {as_label(modifiers[[x]])}"), character(1)), ", ", last = " and "),
-          ". \nDid you map your mod in the wrong layer?"
-        )
-        abort(msg)
+        cli::cli_abort(c(
+          "Aesthetic modifiers returns invalid values",
+          "x" = "The following mappings are invalid: {.code {glue('{nondata_modified} = {as_label(modifiers[[nondata_modified]])}')}}",
```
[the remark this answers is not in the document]
thomasp85: I think my thinking was that it was unlikely to be more than 1 or 2 problems so having them in one line was ok... but I think you are right on second thought

R/geom-point.r
```diff
@@ -185,41 +185,18 @@ translate_shape_string <- function(shape_string) {
...
-
-    more_problems <- if (n_bad > 5) {
-      sprintf("\n* ... and %d more problem%s", n_bad - 5, ifelse(n_bad > 6, "s", ""))
-    } else {
-      ""
-    }
-
-    abort(glue("Shape names must be unambiguous:", collapsed_names, more_problems))
+    cli::cli_abort(c(
+      "{.arg shape_string} elements must be unambiguous",
```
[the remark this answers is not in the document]
thomasp85: you could also get this from `scale_shape_manual(values = c('foo', 'bar', 'baz'))`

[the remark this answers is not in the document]
thomasp85: but that doesn't change your point 🙈

R/facet-grid-.r
```diff
@@ -157,19 +157,19 @@ grid_as_facets_list <- function(rows, cols) {
...
     if (!is.null(cols)) {
-      msg <- "`rows` must be `NULL` or a `vars()` list if `cols` is a `vars()` list"
+      msg <- "{.arg rows} must be {.val NULL} or a {.fn vars} list if {.arg cols} is a {.fn vars} list"
       if(inherits(rows, "ggplot")) {
-        msg <- paste0(
-          msg, "\n",
-          "Did you use %>% instead of +?"
+        msg <- c(
+          msg,
+          "i" = "Did you use {.code %>%} instead of {.code +}?"
```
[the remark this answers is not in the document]
yutannihilation: Sorry, I didn't notice this pull request addresses the same one as #4809. May I merge #4809? Or, do you prefer to address it in this pull request? (As a side comment, the same change should also be applied to the line 156 of `R/layer.r`.)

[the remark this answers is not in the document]
thomasp85: let's do it in this PR so it' all aligned

thomasp85:
@Hadley all errors except for a few special concessions have now been covered with tests

R/limits.r
```diff
@@ -109,11 +109,11 @@ ylim <- function(...) {
 #' ggplot2:::limits(c("A", "b", "c"), "x")
 #' ggplot2:::limits(c("A", "b", "c"), "fill")
 #' ggplot2:::limits(as.Date(c("2008-01-01", "2009-01-01")), "x")
-limits <- function(lims, var) UseMethod("limits")
+limits <- function(lims, var, ...) UseMethod("limits")
```
hadley, line 112: I think you'd be better off making an explicit `call` argument since otherwise `...` will swallow any extra arguments.

R/plot-construction.r
```diff
@@ -51,7 +54,10 @@
   if      (is.theme(e1))  add_theme(e1, e2, e2name)
   else if (is.ggplot(e1)) add_ggplot(e1, e2, e2name)
   else if (is.ggproto(e1)) {
-    abort("Cannot add ggproto objects together. Did you forget to add this object to a ggplot object?")
+    cli::cli_abort(c(
+            "Cannot add {.cls ggproto} objects together",
+      "i" = "Did you forget to add this object to a {.cls ggplot} object?"
+    ))
```
hadley, line 60:
```suggestion
    cli::cli_abort(c(
      "Cannot add {.cls ggproto} objects together",
      "i" = "Did you forget to add this object to a {.cls ggplot} object?"
    ))
```

hadley (approved):
Looks good!

R/guide-bins.R
```diff
@@ -156,7 +156,7 @@ guide_train.bins <- function(guide, scale, aesthetic = NULL) {
     breaks <- strsplit(gsub("\\(|\\)|\\[|\\]", "", breaks), ",\\s?")
     breaks <- as.numeric(unlist(breaks))
     if (anyNA(breaks)) {
-      abort('Breaks not formatted correctly for a bin legend. Use `(<lower>, <upper>]` format to indicate bins')
+      cli::cli_abort('Breaks not formatted correctly for a bin legend. Use {.strong (<lower>, <upper>]} format to indicate bins')
```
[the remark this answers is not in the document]
thomasp85: I'm not on the fence about this... it seems to me that `.code` should be reserved for actual code examples - this is more of a formatting template

R/layer.r
```diff
@@ -199,7 +197,7 @@ Layer <- ggproto("Layer", NULL,
     } else if (is.function(self$data)) {
       data <- self$data(plot_data)
       if (!is.data.frame(data)) {
-        abort("Data function must return a data.frame")
+        cli::cli_abort("{.fn layer_data} must return a {.cls data.frame}")
```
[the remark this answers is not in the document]
thomasp85: yes - the places where it is needed

R/layer.r
```diff
@@ -68,12 +68,13 @@ layer <- function(geom = NULL, stat = NULL,
                   position = NULL, params = list(),
                   inherit.aes = TRUE, check.aes = TRUE, check.param = TRUE,
                   show.legend = NA, key_glyph = NULL, layer_class = Layer) {
+  call_env = caller_env()
```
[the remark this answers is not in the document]
thomasp85: It's just temporary for improving the errors in the layer constructor. When we eventually store it in the layer object it will just be the call stripped of the environment

R/layer.r
```diff
@@ -68,12 +68,13 @@ layer <- function(geom = NULL, stat = NULL,
...
   if (is.null(geom))
-    abort("Attempted to create layer with no geom.")
+    cli::cli_abort("Attempted to create layer with no geom.", call = call_env)
   if (is.null(stat))
-    abort("Attempted to create layer with no stat.")
+    cli::cli_abort("Attempted to create layer with no stat.", call = call_env)
```
thomasp85, line 75:
```suggestion
    cli::cli_abort("Can't create layer without a stat.", call = call_env)
```

R/layer.r
```diff
@@ -68,12 +68,13 @@ layer <- function(geom = NULL, stat = NULL,
...
   if (is.null(position))
-    abort("Attempted to create layer with no position.")
+    cli::cli_abort("Attempted to create layer with no position.", call = call_env)
```
thomasp85, line 77:
```suggestion
    cli::cli_abort("Can't create layer without a position.", call = call_env)
```

PR #4799, opened 2022-04-10
find_scale() works with scales from packages that are loaded but not attached

### LINKED ISSUE #4705: Support scale functions in packages not attached via scale_type()
This is from https://github.com/r-lib/pillar/pull/404 that implements `scale_type.pillar_num()` (returning `"num"`) and `scale_x_num()` and `scale_y_num()` . The scales become available only after the pillar package is attached, see below. This seems to be due to how scale lookup is performed in https://github.com/tidyverse/ggplot2/blob/c89c265a57fd71f8a0288ce81037296aadc0a012/R/scale-type.R#L13.

Should we allow `scale_type()` to return a reference to an environment where to (also) look for scale functions? Perhaps if our `scale_type()` returns `"pillar::num"`, or as an attribute to the value returned by `scale_type()` ?

``` r
data <- tibble::tibble(
  x = pillar::num((1:10) / 100, fixed_exponent = -3, notation = "eng"),
  y = pillar::num((1:10) / 100, scale = 100, label = "%"),
  z = pillar::num(10^(-5:4), notation = "si")
)
data
#> # A tibble: 10 × 3
#>         x     y     z
#>     <eng>     %  <si>
#>  1  10e-3     1   10µ
#>  2  20e-3     2  100µ
#>  3  30e-3     3    1m
#>  4  40e-3     4   10m
#>  5  50e-3     5  100m
#>  6  60e-3     6    1 
#>  7  70e-3     7   10 
#>  8  80e-3     8  100 
#>  9  90e-3     9    1k
#> 10 100e-3    10   10k

ggplot2::ggplot(data, ggplot2::aes(x, y)) +
  ggplot2::geom_point()
```

``` r
library(pillar)

ggplot2::ggplot(data, ggplot2::aes(x, y)) +
  ggplot2::geom_point()
```

<sup>Created on 2021-12-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

https://github.com/r-quantities/units/pull/294 is also affected, CC @Enchufa2.

krlmlr:
Closes #4705.

### DIFF
--- a/R/scale-type.R
+++ b/R/scale-type.R
@@ -7,10 +7,23 @@ find_scale <- function(aes, x, env = parent.frame()) {
   }
 
   type <- scale_type(x)
+  extra_env <- attr(type, "env")
   candidates <- paste("scale", aes, type, sep = "_")
 
+  get_extra_env <- function() {
+    if (!is.null(extra_env)) {
+      return(extra_env)
+    }
+    # Call only once
+    extra_env <<- all_namespaces()
+    return(extra_env)
+  }
+
   for (scale in candidates) {
-    scale_f <- find_global(scale, env, mode = "function")
+    scale_f <- find_global(
+      scale, env, mode = "function",
+      get_extra_env()
+    )
     if (!is.null(scale_f))
       return(scale_f())
   }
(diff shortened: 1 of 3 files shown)


thomasp85:
What is the performance repercussions of searching all loaded namespaces? There are many standard aesthetics for which scales do not exist, because they are used as-is so `find_scale()` is called quite a lot where it will end up looking through all loaded namespaces

krlmlr:
How many `find_scale()` are there in a usual plot? Should we avoid calling `find_scale()` for bare unclassed vectors?

Implementers can optimize by adding an `"env"` attribute to their `scale_type()` return value too.

thomasp85:
There will be one for every aesthetic not provided with an explicit scale in the plot construction... conserving this path for non-atomic aesthetics is probably a worthwhile addition...

Think about this, if we add it we should consider adding this approach to `check_subclass()` so that stats, geoms, and position can be looked up by name even in packages that are only loaded

teunbrand:
Closing this in favour of #6747

PR #4800, opened 2022-04-14
Upgrade rlang dependency to >= 1.0.0

### LINKED ISSUE #4797: Unexported (until v1.0.0) rlang functions used
The dev version of `ggplot2` uses `on_load()`/`run_on_load()` from `rlang`, but these functions are not exported for `rlang` versions < 1.0.0. I think the `ggplot2` DESCRIPTION dependency needs to be upgraded to `rlang` >= 1.0.0.

I noticed this after getting the following error when trying to install the dev version of `ggplot2` while only having `rlang` version 0.4.10:

```
* installing *source* package 'ggplot2' ...
** using staged installation
** R
** data
*** moving datasets to lazyload DB
** inst
** byte-compile and prepare package for lazy loading
Error in on_load({ : could not find function "on_load"
Error: unable to load R code in package 'ggplot2'
Execution halted
ERROR: lazy loading failed for package 'ggplot2'
Warning in install.packages :
  installation of package 'ggplot2' had non-zero exit status
```

Not sure what the reprex would be here other than installing specific versions of these packages with `devtools::install_version()`; can provide if needed.

Crosspost from [stackoverflow](https://stackoverflow.com/questions/71799966/should-ggplot2-dev-version-depend-on-rlang-1-0-0-due-to-exported-on-load-fun?noredirect=1#comment126884142_71799966).

billybarc:
Closes #4797

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,7 +40,7 @@ Imports:
     lifecycle,
     MASS,
     mgcv,
-    rlang (>= 0.4.10),
+    rlang (>= 1.0.0),
     scales (>= 0.5.0),
     stats,
     tibble,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Require rlang >= 1.0.0 (@billybarc, #4797)
+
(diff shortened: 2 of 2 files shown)


thomasp85 (approved):

thomasp85:
Thanks!

PR #4812, opened 2022-04-23
Try no RSPM to fix temporary errors on the Windows runner

yutannihilation:
(This pull request is not intended to be merged, just to test if the CI runs well even after the release of R 4.2. But we can merge this if we want the CI to be fixed immediately.)

It seems RSPM is not ready for R 4.2 for Windows.


yutannihilation:
By the way, do we really need the runners for each Rtools version? If so, should we add a runner for `r: '4.1'` to keep testing Rtools40?

https://github.com/tidyverse/ggplot2/blob/21df8dc88e9b736a136bc39362defe2943008767/.github/workflows/R-CMD-check.yaml#L28-L29

yutannihilation:
Seems it's already resolved now.

PR #4832, opened 2022-05-06
Documentation clarifying stat_density/after_stat n

### LINKED ISSUE #4476: Document computed variable `n`  in `stat_density()`.
Today I was surprised by the following. I wanted to compute proportional densities of two groups. This is doable by using `y = after_stat(count / nrow(df))`, but I had precomputed `n <- nrow(df)` and used it as below:

``` r
library(ggplot2)

df <- data.frame(
  x = c(rnorm(1000, -1), rnorm(100, 1)),
  group = rep(c("A", "B"), c(1000, 100))
)

n <- nrow(df) 

ggplot(df, aes(x)) +
  geom_density(aes(y = after_stat(count / n),
                   fill = group))
```

Both densities above integrate to 1, whereas I wanted the sum of density areas to be 1. I can fix this by using another symbol for the number of observations.

``` r
random_symbol <- n

ggplot(df, aes(x)) +
  geom_density(aes(y = after_stat(count / random_symbol),
                   fill = group))
```

<sup>Created on 2021-05-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I don't think the behaviour is problematic, but it would be nice if `n` was documented as a computed variable in `?stat_density`.

As a side note, the note in `?after_stat` reads:
> Evaluation after stat transformation will only have access to the variables calculated by the stat.

Though in the second plot, I can use `random_symbol` from the global environment.

thoolihan:
Fixes #4476 by clarifying that n is a computed value of stat_density() that can is not available as original mapping in after_stat. Also wording better specifies that after_stat masks original mapping, but not variables in parent scope(s).

### DIFF
--- a/R/aes-evaluation.r
+++ b/R/aes-evaluation.r
@@ -23,10 +23,11 @@
 #' `after_stat()` replaces the old approaches of using either `stat()` or
 #' surrounding the variable names with `..`.
 #'
-#' @note Evaluation after stat transformation will only have access to the
-#' variables calculated by the stat. Evaluation after scaling will only have
-#' access to the final aesthetics of the layer (including non-mapped, default
-#' aesthetics). The original layer data can only be accessed at the first stage.
+#' @note Evaluation after stat transformation will have access to the
+#' variables calculated by the stat, not the original mapped values. Evaluation
+#' after scaling will only have access to the final aesthetics of the layer
+#' (including non-mapped, default aesthetics). The original layer data can only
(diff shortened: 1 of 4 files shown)


thoolihan:
Could use a second pair of eyes. In particular, the documentation may now be redunant in R/aes-evaluation.r, but I'd like a second opinion

thomasp85 (approved):

PR #4844, opened 2022-05-14
Add set.seed() before runif(), rnorm(), and rpois()

yutannihilation:
I fixed the non-deterministic examples using `sample()` on #4502, but I found there are `runif()`, `rnorm()`, and `rpois()` as well.


thomasp85 (approved):
LGTM

Issue #4843, opened 2022-05-13
Example code stat_summary with fun.data does not return a result


JoFAM:
When I run the following code from the help page of `?stat_summary()`:

```{r}
d <- ggplot(mtcars, aes(cyl, mpg)) + geom_point()
d + stat_summary(fun.data = "mean_cl_boot", colour = "red", size = 2)
```
I don't see any summary stat displayed on the plot and I get following warning:

```
Warning message:
Computation failed in `stat_summary()`:
```

Session info:

R version 4.1.2 (2021-11-01)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19044)

Matrix products: default

locale:
[1] LC_COLLATE=English_United Kingdom.1252  LC_CTYPE=English_United Kingdom.1252   
[3] LC_MONETARY=English_United Kingdom.1252 LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] ggplot2_3.3.6 dplyr_1.0.8  

loaded via a namespace (and not attached):
 [1] rstudioapi_0.13  magrittr_2.0.2   tidyselect_1.1.1 munsell_0.5.0   
 [5] colorspace_2.0-2 R6_2.5.1         rlang_1.0.1      fansi_1.0.2     
 [9] tools_4.1.2      grid_4.1.2       gtable_0.3.0     utf8_1.2.2      
[13] cli_3.1.1        DBI_1.1.2        withr_2.4.3      ellipsis_0.3.2  
[17] digest_0.6.29    assertthat_0.2.1 tibble_3.1.6     lifecycle_1.0.1 
[21] crayon_1.4.2     farver_2.1.0     purrr_0.3.4      vctrs_0.3.8     
[25] glue_1.6.1       labeling_0.4.2   compiler_4.1.2   pillar_1.7.0    
[29] generics_0.1.2   scales_1.1.1     pkgconfig_2.0.3

JoFAM:
Never mind, it happens when the package Hmisc is not installed. Maybe that could be returned by the warning message :-)

Issue #4819, opened 2022-04-27
geom_hex() doesn't work


dax44:
After updating to r 4.2.0 geom_hex() doesn't work for me :(

```r
d <- ggplot(diamonds, aes(carat, price))
d + geom_hex()
```

```
Warning message:
Computation failed in `stat_binhex()`:
```

and in plots I have this
![image](https://user-images.githubusercontent.com/8738353/165468802-140836f3-423e-402d-831e-ac2ab4702f07.png)

yutannihilation:
You need to install hexbin package. Sorry that the error message is cryptic. See https://github.com/tidyverse/ggplot2/issues/4736.

PR #4845, opened 2022-05-14
Propagate errors from check_installed()

### LINKED ISSUE #4736: stat_binhex() fails to notify users that they need to install hexbin package
https://github.com/tidyverse/ggplot2/pull/4671#issuecomment-1046090308

I can reproduce this on my local without hexbin installed.

```
> library(ggplot2)
> d <- ggplot(diamonds, aes(carat, price))
> d + geom_hex()
Warning message:
Computation failed in `stat_binhex()`:

```

I think this is because `check_installed` is inside `compute_group`, where the error is ignored.

https://github.com/tidyverse/ggplot2/blob/c89c265a57fd71f8a0288ce81037296aadc0a012/R/stat-binhex.r#L52

yutannihilation:
Fix #4736

Ideally, it's nice if `check_install()` would interactively ask users to install the package, but it seems it doesn't work inside `tryCatch()`, which is used in `Stat$compute_layer()`.

``` r
> tryCatch(rlang::check_installed("ggfortify"), error = function(e) print(e))
<error/rlib_error_package_not_found>
Error:
! The package `ggfortify` is required.
```

As a workaround, this pull request propagates the error to the user instead of gracefully drawing a blank plot with a mysterious warning "Computation failed".


hadley:
I think it's probably worth waiting on @thomasp85's revamp of wrapped errors before doing anything here. We now have a much better mechanism for rethrowing errors while maintaining context (https://rlang.r-lib.org/reference/topic-error-chaining.html), but this should be applied globally to all existing `tryCatch()`s before attempting to resolve the specific package installation problem.

yutannihilation:
Sure, I'll wait (and learn about the feature in rlang).

thomasp85:
The cli update has now been merged into main so if you merge this into here we can finish it off

yutannihilation:
No idea why this fails...

```
> devtools::test(filter = "stat-summary")
ℹ Loading ggplot2
ℹ Testing ggplot2
✔ | F W S  OK | Context
✖ | 2       5 | stat-summary [0.8s]                                                                
───────────────────────────────────────────────────────────────────────────────────────────────────
Failure (test-stat-summary.R:95:3): stat_summary() errors when check_installed() fails (#4736)
`ggplot_build(p + stat_summary(fun.data = fun1))` threw an unexpected error.
Message: Computation failed in `stat_summary()`
Caused by error:
! usual error
Class:   rlang_warning/warning/condition
Backtrace:
  1. testthat::expect_warning(...)
       at test-stat-summary.R:95:2
 46. base::.handleSimpleError(`<fn>`, "usual error", base::quote(`<fn>`(df$y)))
 47. rlang h(simpleError(msg, call))
 48. handlers[[1L]](cnd)
 49. cli::cli_warn(...)
       at ggplot2/R/stat-.r:108:10

Failure (test-stat-summary.R:95:3): stat_summary() errors when check_installed() fails (#4736)
`expect_error(...)` did not throw the expected warning.
───────────────────────────────────────────────────────────────────────────────────────────────────

══ Results ════════════════════════════════════════════════════════════════════════════════════════
Duration: 0.8 s

[ FAIL 2 | WARN 0 | SKIP 0 | PASS 5 ]
```

R/stat-.r
```diff
@@ -97,10 +97,18 @@ Stat <- ggproto("Stat",
     args <- c(list(data = quote(data), scales = quote(scales)), params)
     dapply(data, "PANEL", function(data) {
       scales <- layout$get_scales(data$PANEL[1])
-      try_fetch(do.call(self$compute_panel, args), error = function(cnd) {
-        cli::cli_warn("Computation failed in {.fn {snake_class(self)}}", parent = cnd)
-        new_data_frame()
-      })
+      try_fetch(do.call(self$compute_panel, args),
```
hadley, line 100: Unless I'm missing something, I think this should be more like:

```R
try_fetch(do.call(self$compute_panel, args),
  rlib_error_package_not_found = function(cnd) {
    cli::cli_abort("Aborted computation in {.fn {snake_class(self)}}", parent = cnd)
  },
  error = function(cnd) {
    cli::cli_warn("Computation failed in {.fn {snake_class(self)}}", parent = cnd)
    new_data_frame()
  }
)
```

hadley, line 100: Unless I'm missing something, I think this should be more like:

```R
try_fetch(do.call(self$compute_panel, args),
  rlib_error_package_not_found = function(cnd) {
    cli::cli_abort("Aborted computation in {.fn {snake_class(self)}}", parent = cnd)
  },
  error = function(cnd) {
    cli::cli_warn("Computation failed in {.fn {snake_class(self)}}", parent = cnd)
    new_data_frame()
  }
)
```

hadley: Also note that the `in {.fn {snake_class(self)}}` is temporary unless we replace it with a proper `call`. Not sure if we want to do that now or later.

hadley: Also note that the `in {.fn {snake_class(self)}}` is temporary unless we replace it with a proper `call`. Not sure if we want to do that now or later.

yutannihilation: I thought so at first, but the rethrown error gets caught in the `error` case, so I had to do it all in `error` case. Are there any better way?

``` r
ignore_all_but_custom_error <- function(expr) {
  result <- rlang::try_fetch(
    expr,
    custom_error = function(cnd) {
      rlang::inform("custom_error detected")
      rlang::abort("rethrowing the error", parent = cnd)
    },
    error = function(cnd) {
      rlang::inform("ignoring the error")
    },
  )
  
  invisible(result)
}

# works as expected
ignore_all_but_custom_error(stop("foo"))
#> ignoring the error

# hmm...
ignore_all_but_custom_error(rlang::abort("foo", class = "custom_error"))
#> custom_error detected
#> ignoring the error
```

<sup>Created on 2022-05-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

yutannihilation: This looks not bad to me, but let me think. I'm a noob at cli...

```
> ggplot(diamonds, aes(carat, price)) + geom_hex()
Error in `fun()` at ggplot2/R/compat-plyr.R:375:4:
! Aborted computation in `stat_binhex()`
Caused by error in `f()` at ggplot2/R/ggproto.r:178:11:
! The package `hexxbin` is required for `stat_binhex()`
```

hadley: Hmmm, I suspect this is because of the way that `tryCatch()` adds the calling handlers behind the scenes. Hopefully @lionel- can suggest a fix.

lionel-: I think I'd structure the errors in this way, just to avoid multiple formulations of the same idea (an error occurred):

```
Error in `fun()` at ggplot2/R/compat-plyr.R:375:4:
Caused by error in `stat_binhex()`:
Caused by error in `f()` at ggplot2/R/ggproto.r:178:11:
! The package `hexxbin` is required for `stat_binhex()`
```

lionel-: Now I'm wondering if `check_installed()` should signal non-error conditions. This way it'd work in `tryCatch()`.

There are two steps currently:

1. Signal a classed condition (that currently inherits from `"error"`) with a restart.
2. Call `invokeRestart("abort")` if user selected no, bypassing everything on the stack including error handlers.

The second step is `tryCatch()`-proof but not the first step. If it were, I think no complications would be needed.

hadley: But doesn't the same problem arise with any classed condition? i.e. you can't use `tryCatch()` to handle a classed error _and_ all other errors.

lionel-: > But doesn't the same problem arise with any classed condition?

If we use a classed condition that doesn't inherit from `"error"` to communicate, `"error"` handlers won't be called when we signal. This way we avoid interactions with catch-all-errors handlers.

>  i.e. you can't use tryCatch() to handle a classed error and all other errors.

Sorry I'm not sure I understand what you mean.

hadley: I think we have to keep the error class as errors, otherwise they wouldn't get caught by a generic `error` argument? 

Isn't the whole point of the design of `tryCatch()` to allow you to deal with error subclasses individually or as a whole?

yutannihilation: > Caused by error in `stat_binhex()`:
> Caused by error in `f()` at ggplot2/R/ggproto.r:178:11:

I couldn't figure out how to structure like this. Can we add the additional "Caused by" like without providing the error message (the line starting with `!`)?

yutannihilation:
As the things discussed here are beyond my knowledge, I'm getting to feel I'm not the right person to address this... But, I'd like to include at least a temporary fix in the next release because it seems the current error message confuses users (https://github.com/tidyverse/ggplot2/pull/4671#issuecomment-1046090308, #4819, and #4843. I think we've never seen this kind of questions this frequently before the last release). Let me think what I can do here.

hadley:
We can easily ensure that the error message is correctly forwarded into the warning; Lionel and I just need to figure out the right way to handle the package not installed error. We should definitely get a fix into 3.4.

yutannihilation:
Thanks. Lionel commented as follows on my question on rlang's repo, so I was wondering if "the right way to handle the package not installed error" would require a new feature on `check_installed()`'s side.

> We could add a `fallback` argument to complement `action`.
https://github.com/r-lib/rlang/issues/1402#issuecomment-1128462863

hadley:
I don't think so; we need to get to the root cause of why `tryCatch()` doesn't work here, and resolve that.

yutannihilation:
I see.

R/stat-.r
lionel-: I was wondering whether check-installed really follows the error model or instead is more like an interrupt.

- Like interrupts, the current computation is aborted after user interaction, when the user selects "No". By contrast errors are triggered by the program rather than the user.

- Like interrupts, this is accomplished using `invokeRestart("abort")`, at which point nothing can prevent the jump (except an intercepting `"abort"` restart).

Originally I restarted to `"abort"` instead of throwing an error to improve the UI. When the user selects no, they are taken back to a prompt instead of seeing an error.

Also note that currently `tryCatch(error = )` is causing `check_installed()` to exit too soon, I think it's a bug. We added a restart mechanism in r-lib/rlang#1199 in which we signal the error condition instead of a bare condition. I think this is not correct since condition communication should always be done with bare conditions (or more precisely classed conditions that don't inherit from error, warning, or message) to avoid catch-all handlers.

We can fix the restart signal protocol but then we still need to think about what UI do we want to implement:

1. Should `check_installed()` prompt the user at all within `tryCatch(error = )`?

2. If we do prompt the user, and they select "No", should they consistently be taken back to the prompt even when within `tryCatch()`? In that case, this means we implement the interrupt model in interactive sessions. Otherwise, we should resignal the condition, this time inheriting from `"error"`.

Regarding (1), I think we should prompt the user in interactive sessions, since that is the documented purpose of `check_installed()`. Though I'm a bit concerned about users setting up a loop e.g. with `safely()` to run a long computation and then come back to their computer the next day to find the `check_installed()` prompt. Maybe we should detect that we are called within `source()` and in that case behave as if the session was non-interactive?

Regarding (2), I think the interrupt model makes the most sense to me. I'd be a bit surprised to select "No" and then see the computation continue.

hadley: Ok, that makes sense — that way the initial call is an interrupt and it only becomes an error if you select "No". I'm not too concerned with your `safely()` scenario, because I think it will be relatively unlikely that a computation half-way through the loop will require package installation, and generally you will have prototyped at least once before doing the loop. 

I agree that selecting "No" should take you to the top level, as otherwise in the code below you'd need to select "No" 10 times in order to exit the `map()`:

```R
f <- function(i) {
  # make a plot that uses hexbin
  check_installed("hexbin")
}
purrr::map(1:10, safely(f))
```

lionel-: > I agree that selecting "No" should take you to the top level,

hmm but then we agree that `check_installed()` is either a no-op (after a restart if user selects yes) or an interrupt (if no is selected), never an error, right?

lionel-: Great point regarding repeated prompts in loops by the way. This clearly shows why `check_installed()` should implement the interrupt model rather than the error model.

thomasp85:
@Hadley do we elect to close this and let the "real fix" happen in rlang?

lionel-:
With dev rlang we now have the behaviour described in @hadley's last message.

Thanks for looking into this issue @yutannihilation!

thomasp85:
Ok - closing for now

yutannihilation:
Oh, cool. Glad to see the issue I had no idea how to fix was resolved while I did nothing :) Thanks so much!

PR #4846, opened 2022-05-15
Add test for retransform argument

teunbrand:
As discussed in #4835, I added a test that should break when the retransform isn't applied correctly on computed aesthetics.


clauswilke (approved):

clauswilke:
This looks good to me. Would be good to have another set of eyes look at this, though, as I haven't done much ggplot development recently.

yutannihilation:
> I haven't done much ggplot development recently.

Same here :)

Looks good to me.

PR #4849, opened 2022-05-17
Fix labels/breaks/limits interactions in bin guides

### LINKED ISSUE #4831: labeled breaks for binned color scale malfunction when show.limits = T
---

When formatting labels for breaks in color scales, ggplot returns an error for the length of label vector if _show.limits = T_. This error appears even if the breaks and labels vectors are the same length. 

This may be because limits and breaks are passed to different vectors before compiling the final plot and therefore the user cannot easily write labels for limits and breaks together. An example is provided here: [https://stackoverflow.com/questions/71976832/custom-labels-for-limit-and-break-values-when-using-scale-color-binned-in-r-ggpl](url)

Is it possible to integrate limits and breaks in a more streamlined way to avoid the error message and not have to define labels with a custom function that circumvents the issue?

```r
library(ggplot2)

ggplot(mpg, aes(cty, hwy, color = year)) +
  geom_point() +
  scale_color_binned(limits = c(1999, 2008),
                    breaks = c(2000, 2002, 2004, 2006),
                    labels = c('2', '3', '4', '5'),
                    show.limits = F) # labels work

ggplot(mpg, aes(cty, hwy, color = year)) +
  geom_point() +
  scale_color_binned(limits = c(1999, 2008),
                    breaks = c(2000, 2002, 2004, 2006),
                    labels = c('2', '3', '4', '5'),
                    show.limits = T) # method fails when limits are activated

    # Error in f(): ! Breaks and labels are different lengths

```

thomasp85:
Fix #4831

This PR properly defines how breaks and labels behaves with and without `show.limits` set and in the presence of an overlap between limits and breaks.

* If a limit is also present in breaks it will be shown no matter the value of `show.limits`
* If `labels` is given as a vector of values then `show.limits` will be ignored with a warning


hadley (approved):
Does this need any documentation updates anywhere

Is the change to [guide-bins-can-show-limits.svg](https://github.com/tidyverse/ggplot2/pull/4849/files#diff-874fcab3a2f446e0d3ead3ff97daae029dbb98e88732e49c8c9efce1167f8f35) deliberate?

thomasp85:
I guess it would make sense to add a note in the docs

The change in the already present visual test is because this PR fixes a bug that was uncovered as well

PR #4853, opened 2022-05-20
Use rlang::hash instead of digest

### LINKED ISSUE #4458: Switch digest::digest() to rlang::hash()?
As of version 0.4.10, rlang has `hash()`. We previously discussed about the alternative of `digest::digest()` (c.f. https://github.com/tidyverse/ggplot2/issues/4247#issue-728905266), and this seems the one, while at the moment we don't have strong motivation to migrate. I don't want to include this in the next release, but let's revisit here when it's done. I think `rlang::hash()` is good in that

* it seems faster than `digest::digest()` for our usages, and
* we'll have less dependencies since we already have rlang in Imports.

Here's the benchmark:

``` r
library(base64enc)
library(rlang)
library(digest)

x <- list(
  "z",
  as.character(1:6),
  data.frame(
    colour = scales::viridis_pal()(300),
    value  = seq(1, 6, length.out = 300)
  ),
  "colourbar"
)

microbenchmark::microbenchmark(
  digest_w_colon  = digest::digest(x),
  digest_wo_colon = digest(x),
  digest_xxhash32 = digest::digest(x, "xxhash32"),
  digest_xxhash64 = digest::digest(x, "xxhash64"),
  serialize_ascii = serialize(x, NULL, TRUE),
  serialize_bin   = serialize(x, NULL, FALSE),
  rawToChar       = rawToChar(serialize(x, NULL, TRUE)),
  rlang_w_colon   = rlang::hash(x),
  rlang_wo_colon  = hash(x)
)
#> Unit: microseconds
#>             expr     min       lq      mean   median       uq      max neval
#>   digest_w_colon 113.855 123.9700 132.96479 128.3845 136.9935  224.025   100
#>  digest_wo_colon 101.118 112.5620 123.06313 116.9640 125.2750  197.081   100
#>  digest_xxhash32  85.048  96.9555 111.56161 101.8455 113.0255  699.039   100
#>  digest_xxhash64  86.641  97.4070 109.75329 103.4085 110.7890  223.573   100
#>  serialize_ascii 822.214 840.2070 872.77923 847.3240 867.1200 1448.936   100
#>    serialize_bin  47.652  54.0820  58.62599  56.6065  62.0685   95.946   100
#>        rawToChar 869.960 886.7930 909.63336 890.8435 900.6660 1288.549   100
#>    rlang_w_colon  45.479  52.5935  57.35674  54.9265  59.8355   98.420   100
#>   rlang_wo_colon  34.342  38.1960  41.78910  39.8310  42.0270  156.552   100
#>    cld
#>    c  
#>   bc  
#>   b   
#>   b   
#>     d 
#>  a    
#>      e
#>  a    
#>  a
```

<sup>Created on 2021-04-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

thomasp85:
Fix #4458 

This PR removes the digest dependencies in favour of `rlang::hash()`.

Since guides are sorted by their hash value and the hash value changes it can lead to changes in the automatic ordering of legends, as reflected in the change in visual tests. Since we make no promises on how they are sorted I think this is fine but should of course be communicated

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,6 @@ Depends:
     R (>= 3.3)
 Imports: 
     cli,
-    digest,
     glue,
     grDevices,
     grid,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Use `rlang::hash()` instead of `digest::digest()`. This update may lead to 
+  changes in the automatic sorting of legends (@thomasp85, #4458)
+
 * Fix various issues with how `labels`, `breaks`, `limits`, and `show.limits`
   interact in the different binning guides (@thomasp85, #4831)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 # ggplot2 (development version)
 
 * Use `rlang::hash()` instead of `digest::digest()`. This update may lead to 
-  changes in the automatic sorting of legends (@thomasp85, #4458)
+  changes in the automatic sorting of legends. In order to enforce a specific
+  legend order use the `order` argument in the guide. (@thomasp85, #4458)
 
 * Fix various issues with how `labels`, `breaks`, `limits`, and `show.limits`
   interact in the different binning guides (@thomasp85, #4831)
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -209,7 +209,7 @@ guide_train.bins <- function(guide, scale, aesthetic = NULL) {
   guide$key <- key
   guide$hash <- with(
     guide,
-    digest::digest(list(title, key$.label, direction, name))
+    hash(list(title, key$.label, direction, name))
   )
   guide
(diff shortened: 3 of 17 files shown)


yutannihilation:
Looks good, and I agree with you on this point,

> Since we make no promises on how they are sorted

however, IIUC, the legend order had been consistent for this 11 years (since the this first commit?: https://github.com/tidyverse/ggplot2/commit/0795c6f1e3f3d4d3b9afc0aaea176d3e7a24a7f0), which sounds long enough to make the users assume the behavior would never change. So, I think the NEWS item should have a bit more words (probably including how to determine the order manually by `guide_*(order = )`?), although this change isn't so "breaking"

thomasp85:
@yutannihilation I've improved the news bullet - it will be part of the "Breaking Changes" section in the news once we polish it prior to release

yutannihilation (approved):
Thanks, looks good!

PR #4856, opened 2022-05-23
Chained errors with contextual info in layers

thomasp85:
This PR improves the error reporting when errors occur during evaluations of a layer. It also adds an improvement to all errors in ggproto methods since the methods are now called by name so we avoid the useless `Error in f()` we get now.

The new look of layer errors is:

``` r
ggplot(economics_long, aes(date, value01, colour = date)) +
  geom_point() + 
  geom_line(linetype = 2)
#> Error in `geom_line(linetype = 2)`:
#> ! Problem converting geom to grob
#> ℹ Error occurred in the 2nd layer
#> Caused by error in `draw_panel()` at ggplot2/R/ggproto.r:179:18:
#> ! `geom_line()` can't have varying colour, size, and/or alpha along the line when
#>   linetype isn't solid
```

which is a mouthful but a huge improvement in terms of letting the user understand how their code is wrong

The remaining question is if this error chaining should be expanded to other ggproto objects in play during plot rendering. I feel it is less imperative than the layer errors because facets and coords are singular in the plot and they are thus less likely to cause confusion as to where the error occurred. I'm however open to expanding the `try_fetch` implementations in `ggplot_build()` to cover it all

### DIFF
--- a/R/ggproto.r
+++ b/R/ggproto.r
@@ -153,7 +153,7 @@ fetch_ggproto <- function(x, name) {
     return(res)
   }
 
-  make_proto_method(x, res)
+  make_proto_method(x, res, name)
 }
 
 #' @export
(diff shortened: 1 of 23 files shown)


hadley (approved):
This is awesome 😄

clauswilke:
This is great!

One minor comment: There is a bit of a naming inconsistency between "stat" and "statistic", sometimes you use one, sometimes the other. I think "stat" is used much more widely and so I would consider using it always.

There are two cases where you currently write out "statistic", I believe:
- "computing statistics"
- "mapping statistics to aesthetics"

I think it would be fine to write instead:
- "computing stats"
- "mapping stats to aesthetics"

(Don't feel strongly about this either way.)

clauswilke (approved):

PR #4857, opened 2022-05-25
use approx instead of nearest match when placing secondary ticks

### LINKED ISSUE #3576: Incorrect tick positions of the secondary axis

thomasp85:
Fix #3576
Close #3581 

This PR implements a very simple approach to doing proper interpolation to find the placement of the secondary ticks thus removing the small imprecisions that can occur when using secondary axes.


yutannihilation (approved):
Sorry, somehow I missed I was assigned as a reviewer here. Reading the description of #3581, I think I understand the problem, and this looks good to me.

clauswilke:
I'm not sure I fully understand everything involved, so I'll abstain from making a comment one way or another.

PR #4865, opened 2022-06-07
Update documentation for geom_map()

### LINKED ISSUE #3721: Mark geom_map() as superseded
I believe we are already in the mood of moving away from `geom_map()` as `geom_sf()` almost superseded it. But, AFAIK, this hasn't been discussed formally so far. Considering we keep developing many codes around `geom_sf()`, I think it's good to mark `geom_map()` soft-deprecated before we face difficulties about the different behaviors between them.

Here I file this issue to discuss what functionalities are missing from `geom_sf()` and its related functions, which might drive people to fall back to `geom_map()`. Any thoughts?

Edit: to be clear, I don't think we can remove `geom_map()` considering its wide spread usages (so I chose the word "soft-deprecated"). Still, I believe we can discourage the usage of `geom_map()` to prevent further confusions by improving `geom_sf()` or providing some alternatives.

clauswilke:
This fixes #3721. 

Marked as WIP as I want to give it one more careful review before I consider it final.

### DIFF
--- a/R/coord-map.r
+++ b/R/coord-map.r
@@ -1,5 +1,8 @@
 #' Map projections
 #'
+#' @description
+#' `r lifecycle::badge("superseded")`
+#'
 #' `coord_map()` projects a portion of the earth, which is approximately
 #' spherical, onto a flat 2D plane using any projection defined by the
 #' `mapproj` package. Map projections do not, in general, preserve straight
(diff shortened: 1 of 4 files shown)


clauswilke:
I've looked it over one more time. I think it's ready to go.

thomasp85 (approved):
LGTM

PR #4866, opened 2022-06-09
Check for dropped aesthetics

### LINKED ISSUE #3250: Unexpected failure of density plot without any error or warning

clauswilke:
Check for dropped aesthetics and issue a warning if aesthetics are dropped that shouldn't be. Fixes #3250.

As a consequence of this change, stats now need to advertise which aesthetics they are dropping, via the new `dropped_aes` field. If they don't, unit tests may fail because code now generates warnings that previously ran quietly.

Example of the new feature:

``` r
library(tidyverse)

ggplot(mtcars, aes(mpg, fill = am)) + geom_density(alpha = 0.4)
#> Warning: The following aesthetics were dropped during statistical transformation: fill
#> ℹ This can happen when ggplot fails to infer the correct grouping structure in
#>   the data.
#> ℹ Did you forget to specify a `group` aesthetic or to convert a numerical
#>   variable into a factor?
```

``` r
ggplot(mtcars, aes(mpg, fill = am, group = am)) + geom_density(alpha = 0.4)
```

<sup>Created on 2022-06-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


R/stat-density-2d.r
```diff
@@ -149,6 +153,8 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
     } else { # lines is the default
       contour_stat <- StatContour
```
thomasp85, line 154: I think you need to make a new object since you are modifying a field of it below and it will affect everything inheriting from `StatContour` afterwards
```suggestion
      contour_stat <- ggproto(NULL, StatContour)
```

R/stat-.r
```diff
@@ -122,7 +127,20 @@ Stat <- ggproto("Stat",
       )
     }, stats, groups, SIMPLIFY = FALSE)
 
-    rbind_dfs(stats)
+    data_new <- rbind_dfs(stats)
```
thomasp85, line 130: `rbind_dfs()` will be removed in favour of `vec_rbind()` soon, just to remember depending on whether this gets merged before or after

thomasp85 (changes requested):
Generally looks good - should we update the docs for the affected stats to say that these aesthetics are dropped? Seems like a nice thing to do

R/stat-density-2d.r
clauswilke: Yes, correct. It didn't matter in the past but it matters now. I'll fix it.

clauswilke:
Let me think about documenting the dropped aesthetics. In some cases it's just a temporary variable that gets dropped, so documenting that is not so useful. In other cases (e.g. `stat_density_2d()`) one of the input variables gets dropped, so documenting that makes sense.

thomasp85:
yeah - "thankfully" the stats aes documentation is not automatic so it is just a matter of cherry picking where to add it to the docs

R/stat-.r
clauswilke: I just saw your PR. I think I'll wait until you've merged it and then apply the update in my code.

thomasp85:
@clauswilke is it possible to get the docs updated soon? I plan to start the release process soonish to give the revdep maintainers the summer to fix potential issues

clauswilke:
@thomasp85 I think I can do it today. I was waiting for you to merge the vctrs PR, but that has happened, correct?

thomasp85:
Ah, sorry -- I missed that. Yes, vctrs has been merged in

clauswilke:
@thomasp85 I think I've addressed everything. I documented the stats where it seemed appropriate. In particular, I didn't document `stat_boxplot()` because I think it'd be more confusing if I did than if I didn't.

thomasp85 (approved):
lgtm

PR #4342, opened 2021-02-06
Very experimental use of vctrs

hadley:
Just a 15 minute experiment. I have no recollection of how the facet data structure works, so this is just banging rocks together.

Fixes #4175

PR #4868, opened 2022-06-10
Use vctrs internally

thomasp85:
Supersedes #4342 

This PR aims to bring the internals of ggplot2 into a state that is compliant with vctrs data types, as well as clean up the implementation different places. The aim is not to convert every base function that has a vctrs equivalent - only that which is necessary or beneficial.

Current state includes:
* removal of `rbind_dfs()` in favour of `vec_rbind()`
* Use `vec_cbind()` instead of `cbind()` in the places it makes sense
* Use `vec_interleave()` in our own `interleave()` function
* Convert `mapped_discrete` class into a vctrs based class
* Change all `do.call()` to `rlang::inject()` (not strictly vctrs related but in line with our move to splicing lists into calls)


thomasp85:
I've come to believe that it would be most beneficial to have this PR merged in with the current scope and only bring in more vctrs functions as we find need for it.

R/annotation-custom.r
```diff
@@ -73,7 +73,12 @@ GeomCustomAnn <- ggproto("GeomCustomAnn", Geom,
     if (!inherits(coord, "CoordCartesian")) {
       cli::cli_abort("{.fn annotation_custom} only works with {.fn coord_cartesian}")
     }
-    corners <- new_data_frame(list(x = c(xmin, xmax), y = c(ymin, ymax)), n = 2)
+    corners <- data_frame(
+      x = c(xmin, xmax),
+      y = c(ymin, ymax),
+      .size = 2,
+      .name_repair = "minimal"
```
hadley, line 80: You shouldn't need `.name_repair` here, right? Because none of the names are user supplied?

R/compat-plyr.R
```diff
@@ -390,7 +293,7 @@ dapply <- function(df, by, fun, ..., drop = TRUE) {
 
   ids <- id(grouping_cols, drop = drop)
   group_rows <- split_with_index(seq_len(nrow(df)), ids)
-  rbind_dfs(lapply(seq_along(group_rows), function(i) {
+  vec_rbind(!!!lapply(seq_along(group_rows), function(i) {
```
hadley, line 296: Pull the `lapply()` out into its own variable, since you're here?

R/coord-polar.r
```diff
@@ -212,23 +212,23 @@ CoordPolar <- ggproto("CoordPolar", Coord,
...
+        y = vec_interleave(0, 0.45 * cos(thetamin)) + 0.5,
         id.lengths = rep(2, length(thetamin)),
         default.units = "native"
       ),
 
       element_render(
         theme, majorr, name = "radius",
-        x = rep(rfine, each = length(thetafine)) * sin(thetafine) + 0.5,
-        y = rep(rfine, each = length(thetafine)) * cos(thetafine) + 0.5,
+        x = rep(rfine, each = length(thetafine)) * rep(sin(thetafine), length(rfine)) + 0.5,
```
hadley, line 230:
```suggestion
        x = vec_rep_each(rfine, length(thetafine)) * vec_rep(sin(thetafine), length(rfine)) + 0.5,
```

Not sure if that looks better or not.

R/facet-.r
```diff
@@ -591,7 +600,7 @@ combine_vars <- function(data, env = emptyenv(), vars = NULL, drop = TRUE) {
     if (drop) {
       new <- unique_combs(new)
     }
-    base <- unique(rbind(base, df.grid(old, new)))
+    base <- unique(vec_rbind(base, df.grid(old, new)))
```
hadley, line 603: Might be worth switching to `vec_unique()` here; I think it's quite a bit faster?

R/facet-grid-.r
```diff
@@ -271,7 +278,7 @@ FacetGrid <- ggproto("FacetGrid", Facet,
       facet_rep <- rep(1:nrow(to_add), each = nrow(data))
 
       data <- unrowname(data[data_rep, , drop = FALSE])
-      facet_vals <- unrowname(cbind(
+      facet_vals <- unrowname(vec_cbind(
```
hadley, line 281: Do you still need the `unrowname()`?

R/geom-errorbarh.r
```diff
@@ -51,7 +51,7 @@ geom_errorbarh <- function(mapping = NULL, data = NULL,
 #' @usage NULL
 #' @export
 GeomErrorbarh <- ggproto("GeomErrorbarh", Geom,
-  default_aes = aes(colour = "black", size = 0.5, linetype = 1, height = 0.5,
```
hadley, line 54: Should the `linewidth` stuff be here?

R/guide-colorbar.r
```diff
@@ -413,11 +418,7 @@ guide_gengrob.colorbar <- function(guide, theme) {
     # If any of the labels are quoted language objects, convert them
     # to expressions. Labels from formatter functions can return these
     if (any(vapply(label, is.call, logical(1)))) {
-      label <- lapply(label, function(l) {
-        if (is.call(l)) substitute(expression(x), list(x = l))
-        else l
-      })
-      label <- do.call(c, label)
+      label <- inject(expression(!!!label))
```
hadley, line 421: Maybe just `as.expression(label)`?

R/limits.r
```diff
@@ -184,10 +184,9 @@ limits.POSIXlt <- function(lims, var, call = caller_env()) {
 expand_limits <- function(...) {
   data <- list2(...)
   data_dfs <- vapply(data, is.data.frame, logical(1))
-  data <- do.call(c, c(list(data[!data_dfs]), data[data_dfs]))
+  data <- unlist(c(list(data[!data_dfs]), data[data_dfs]), recursive = FALSE)
```
hadley, line 187: I don't know if it's important, but I have no intuition for what this code does. Is it `inject(c(data[!data_dfs], !!!data[data_dfs])`?

R/limits.r
```diff
@@ -184,10 +184,9 @@ limits.POSIXlt <- function(lims, var, call = caller_env()) {
...
   n_rows <- max(vapply(data, length, integer(1)))
-  data <- lapply(data, rep, length.out = n_rows)
-  data <- new_data_frame(data)
+  data <- data_frame(!!!lapply(data, rep, length.out = n_rows), .name_repair = "minimal")
```
hadley, line 189: I feel like we get better tracebacks if we avoid doing complex computations in a `!!!`

R/scale-discrete-.r
```diff
@@ -141,35 +141,104 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+  # vctrs does not support inheriting from numeric base class
+  class(obj) <- c(class(obj), "numeric")
+  obj
+}
+mapped_discrete <- function(x = double()) {
+  if (is.null(x)) return(NULL)
+  if (is.array(x)) x <- as.vector(x)
+  new_mapped_discrete(vec_cast(x, double()))
+}
+is_mapped_discrete <- function(x) inherits(x, "ggplot2_mapped_discrete")
```
hadley, line 156: Would be worth getting @lionel- or @DavisVaughan to review the vctrs code.

R/scales-.r
```diff
@@ -84,7 +84,7 @@ scales_transform_df <- function(scales, df) {
 
   transformed <- unlist(lapply(scales$scales, function(s) s$transform_df(df = df)),
     recursive = FALSE)
-  new_data_frame(c(transformed, df[setdiff(names(df), names(transformed))]))
+  data_frame(!!!transformed, !!!df[setdiff(names(df), names(transformed))], .name_repair = "minimal")
```
hadley, line 87: Maybe 

```R
untransformed <- df[setdiff(names(df), names(transformed))
vec_cbind(as_data_frame(transformed), untransformed)
```
?

R/annotation-custom.r
DavisVaughan:
``` r
bench::mark(vctrs::data_frame(x=1,y=2),vctrs::data_frame(x=1,y=2,.name_repair = "minimal"))
#> # A tibble: 2 × 6
#>   expression                                                     min   median
#>   <bch:expr>                                                <bch:tm> <bch:tm>
#> 1 vctrs::data_frame(x = 1, y = 2)                            24.39µs  31.61µs
#> 2 vctrs::data_frame(x = 1, y = 2, .name_repair = "minimal")   5.15µs   5.89µs
#> # … with 3 more variables: `itr/sec` <dbl>, mem_alloc <bch:byt>, `gc/sec` <dbl>
```

<sup>Created on 2022-06-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/compat-plyr.R
```diff
@@ -374,13 +277,13 @@ dapply <- function(df, by, fun, ..., drop = TRUE) {
   apply_fun <- function(x) {
     res <- fun(x, ...)
     if (is.null(res)) return(res)
-    if (length(res) == 0) return(new_data_frame())
+    if (length(res) == 0) return(data_frame(.name_repair = "minimal"))
```
DavisVaughan, line 280: Using `vctrs::new_data_frame()` is probably fine here if you wanted to

R/facet-.r
```diff
@@ -257,7 +257,7 @@ df.grid <- function(a, b) {
     i_a = seq_len(nrow(a)),
     i_b = seq_len(nrow(b))
   )
-  unrowname(cbind(
+  unrowname(vec_cbind(
     a[indexes$i_a, , drop = FALSE],
     b[indexes$i_b, , drop = FALSE]
   ))
```
DavisVaughan, line 263: Might be better to unrowname the individual data frames here so `vec_cbind()` doesn't have to do any "work" to combine them since you are just going to remove them right after that

R/facet-.r
```diff
@@ -468,7 +468,14 @@ eval_facet <- function(facet, data, possible_columns = NULL) {
...
 layout_null <- function() {
   # PANEL needs to be a factor to be consistent with other facet types
-  new_data_frame(list(PANEL = factor(1), ROW = 1, COL = 1, SCALE_X = 1, SCALE_Y = 1))
+  data_frame(
+    PANEL = factor(1),
+    ROW = 1,
+    COL = 1,
+    SCALE_X = 1,
+    SCALE_Y = 1,
+    .name_repair = "minimal"
```
DavisVaughan, line 477: `.size = 1L` since you seem to be using `.size` elsewhere

R/facet-.r
DavisVaughan: For example

``` r
time <- nycflights13::flights[c("year", "month", "day")]

bench::mark(unique(time), vctrs::vec_unique(time))
#> # A tibble: 2 × 6
#>   expression                   min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>              <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unique(time)            401.23ms 428.22ms      2.34    9.22MB    15.2 
#> 2 vctrs::vec_unique(time)   3.64ms   4.73ms    164.      3.31MB     3.99
```

<sup>Created on 2022-06-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/guide-bins.R
```diff
@@ -178,7 +178,7 @@ guide_train.bins <- function(guide, scale, aesthetic = NULL) {
     limits <- all_breaks[c(1, length(all_breaks))]
     breaks <- all_breaks[-c(1, length(all_breaks))]
   }
-  key <- new_data_frame(setNames(list(c(scale$map(bin_at), NA)), aes_column_name))
+  key <- data_frame(c(scale$map(bin_at), NA), .name_repair = ~ aes_column_name)
```
DavisVaughan, line 181: What you had before with `vctrs::new_data_frame()` on a named list seems clearer to me than using name repair this way

R/guide-colorbar.r
```diff
@@ -214,7 +214,7 @@ guide_train.colorbar <- function(guide, scale, aesthetic = NULL) {
   if (length(breaks) == 0 || all(is.na(breaks)))
     return()
 
-  ticks <- new_data_frame(setNames(list(scale$map(breaks)), aesthetic %||% scale$aesthetics[1]))
+  ticks <- data_frame(scale$map(breaks), .name_repair = ~ aesthetic %||% scale$aesthetics[1])
```
DavisVaughan, line 217: Same comment about name_repair usage

R/limits.r
DavisVaughan: I guess this function doesn't do standard tidyverse recycling? i.e. this would be an error if it did

``` r
ggplot2:::expand_limits(x = 1:2, y = 1:4)$data
#>   x y
#> 1 1 1
#> 2 2 2
#> 3 1 3
#> 4 2 4
```

<sup>Created on 2022-06-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/scale-discrete-.r
```diff
@@ -141,35 +141,104 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+#' @export
+format.ggplot2_mapped_discrete <- function(x, ...) format(vec_data(x), ...)
```
DavisVaughan, line 158: I'm not sure if we are 100% confident in `vec_data()`. I haven't actually ever needed it in practice. Could you just use `unclass(x)` instead?

R/scale-discrete-.r
```diff
@@ -141,35 +141,104 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+#' @export
+vec_ptype2.ggplot2_mapped_discrete.factor <- function(x, y, ...) new_mapped_discrete()
+#' @export
+vec_ptype2.factor.ggplot2_mapped_discrete <- function(x, y, ...) new_mapped_discrete()
+#' @export
+vec_cast.ggplot2_mapped_discrete.ggplot2_mapped_discrete <- function(x, to, ...) x
+#' @export
+vec_cast.ggplot2_mapped_discrete.numeric <- function(x, to, ...) new_mapped_discrete(x)
+#' @export
+vec_cast.numeric.ggplot2_mapped_discrete <- function(x, to, ...) vec_data(x)
```
DavisVaughan, line 186: We tend to use this a lot vs `vec_data()` in cast methods like this one - if you decide to avoid using `vec_data()` it may be useful

```r
> vctrs:::unstructure
function (x) 
{
    attributes(x) <- NULL
    x
}
```

R/scale-discrete-.r
```diff
@@ -141,35 +141,104 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+#' @export
+vec_ptype2.character.ggplot2_mapped_discrete <- function(x, y, ...) character()
+#' @export
+vec_ptype2.ggplot2_mapped_discrete.factor <- function(x, y, ...) new_mapped_discrete()
+#' @export
+vec_ptype2.factor.ggplot2_mapped_discrete <- function(x, y, ...) new_mapped_discrete()
+#' @export
+vec_cast.ggplot2_mapped_discrete.ggplot2_mapped_discrete <- function(x, to, ...) x
+#' @export
+vec_cast.ggplot2_mapped_discrete.numeric <- function(x, to, ...) new_mapped_discrete(x)
```
DavisVaughan, line 184: Is there a reason you provide cast methods for numeric AND double/integer? It seems like the double/integer cast methods would be enough

R/scale-discrete-.r
```diff
@@ -141,35 +141,104 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+#' @export
+vec_cast.numeric.ggplot2_mapped_discrete <- function(x, to, ...) vec_data(x)
+#' @export
+vec_cast.ggplot2_mapped_discrete.integer <- function(x, to, ...) new_mapped_discrete(x)
```
DavisVaughan, line 188: I don't think this works? `x` is an integer vector but `new_mapped_discrete()` asserts that `x` must be a double vector, right?

R/scale-discrete-.r
```diff
@@ -141,35 +141,104 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
...
+#' @export
+vec_ptype2.ggplot2_mapped_discrete.integer <- function(x, y, ...) new_mapped_discrete()
+#' @export
+vec_ptype2.integer.ggplot2_mapped_discrete <- function(x, y, ...) new_mapped_discrete()
+#' @export
+vec_ptype2.ggplot2_mapped_discrete.character <- function(x, y, ...) character()
+#' @export
+vec_ptype2.character.ggplot2_mapped_discrete <- function(x, y, ...) character()
+#' @export
+vec_ptype2.ggplot2_mapped_discrete.factor <- function(x, y, ...) new_mapped_discrete()
```
DavisVaughan, line 178: Our type hierarchy is like:

```
character
   ^
   |
   |
factor
```

So it felt weird to me that:

```
<chr> + <mapped_discrete> = <chr>

# but

<fct> + <mapped_discrete> = <mapped_discrete>
```

In my head that only works if you consider a mapped_discrete as something between a character and a factor, is that right?

Side question - I'd just like to confirm that you really do want these ptype2 methods related to factors and characters, i.e. are you sure you want this to work?

```
vec_c("x", mapped_discrete(1))
```

If mapped_discrete is a numeric value, I don't feel like it should be combine-able with a character vector. `vec_c("x", 1)` fails in vctrs.

R/utilities-matrix.r
```diff
@@ -15,26 +15,10 @@ cunion <- function(a, b) {
...
-  # Replicate elements of length one up to correct length
-  singletons <- vapply(vectors, length, integer(1)) == 1L
-  vectors[singletons] <- lapply(vectors[singletons], rep, lengths)
-
-  # Interleave vectors
-  n <- lengths
-  p <- length(vectors)
-  interleave <- rep(1:n, each = p) + seq(0, p - 1) * n
-  unlist(vectors, recursive = FALSE)[interleave]
+  vec_interleave(...)
```
DavisVaughan, line 23: Nice!

DavisVaughan (commented):
I think my main comment is that I want to make sure you know about `vctrs::new_data_frame()`. You switch from your custom `new_data_frame()` over to `data_frame()` in a large number of places, but I wonder if you might have been happier just getting the benefits of `vctrs::new_data_frame()` instead.

In particular:
- In cases where you return an empty data frame, like `data_frame(.name_repair = "minimal")`, I think `new_data_frame()` is cleaner
- `new_data_frame()` doesn't do any name repair, so you don't pay that cost at all and there is no `.name_repair = "minimal"` to set. So I think you could have kept some of your old code that looked like `new_data_frame(list(...))` and been pretty happy

R/scale-discrete-.r
lionel-: oh do we? I don't remember this vctrs version but I see that it removes names dimensions and everything. Seeing it, I'd think of it as a purely internal tool that shouldn't be used by others. `rlang:::unstructure()` is a bit safer to use.

R/compat-plyr.R
lionel-: What does `.name_repair = "minimal"` do with no inputs? It's the same as `data_frame()` right?

lionel-:
@DavisVaughan Regarding the switch of `new_data_frame()` to `data_frame()`, I think it's the right move if there is no performance issue because the former can easily cause internal errors in the vctrs code. In general it seems that we should disable input checking and validation only in performance critical code (with comments describing why the assumptions are safe) or in very low level data creation code.

R/scale-discrete-.r
DavisVaughan: Oh yea the rlang one looks good

R/compat-plyr.R
DavisVaughan: @lionel- it is just comparatively very slow because of the name repair stuff

``` r
bench::mark(vctrs::data_frame(),vctrs::data_frame(.name_repair = "minimal"))
#> # A tibble: 2 × 6
#>   expression                                       min   median `itr/sec`
#>   <bch:expr>                                  <bch:tm> <bch:tm>     <dbl>
#> 1 vctrs::data_frame()                          20.64µs  22.15µs    40510.
#> 2 vctrs::data_frame(.name_repair = "minimal")   3.28µs   3.69µs   232047.
#> # … with 2 more variables: mem_alloc <bch:byt>, `gc/sec` <dbl>
```

<sup>Created on 2022-06-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/facet-.r
thomasp85: nice - is `vec_unique()` one of those vctrs functions where it makes sense to switch 100% to?

R/facet-grid-.r
thomasp85: I'm a bit unsure - I've kept it as a precaution but I'll investigate it a bit more

R/guide-colorbar.r
thomasp85: oh - nice 😅

thomasp85:
Thanks for the thorough review. Regarding the `data_frame()`choice over `new_data_frame()` this was chosen mainly because we have a lot of places where the data frame is constructed from scalars and vectors and `new_data_frame()` would then require it be wrapped in `df_list()`. While there are places where it may not be needed (e.g. for the empty data frame construction) It seems cleaner for the code base to use a single constructor function. It may make sense to create a thin wrapper around `data_frame(..., .name_repair = "minimal")` as suggested. I'm definitely open to that.

Is for the comments on the `ggplot2_mapped_discrete` class, the intent is really to make it behave like a numeric vector with all the quirks it may entail (for instance the weird coercion rules of factor and character). The only purpose of this class is to carry around information that this column was once a discrete data type.

R/scale-discrete-.r
thomasp85: see my main comment.

R/scale-discrete-.r
thomasp85: yes - thanks

R/scale-discrete-.r
thomasp85: All I can say is that they were added gradually to fix failing code... I can try to remove them one by one and see if it works

thomasp85: hum - it seems removing the `numeric` casts and ptype's didn't break anything

R/scale-discrete-.r
thomasp85: since mapped discrete are only constructed internally from bare numerics it should be fine to use `vec_data()`

R/coord-polar.r
thomasp85: hmm... I feel we should avoid using the vctrs functions here and there. AFAIK the `vec_rep*()` are actually a slightly slower than base rep so I'd steer away from this purely for looks

R/limits.r
thomasp85: @DavisVaughan this is not intended to be seen as data construction in any way - just a way to specify multiple limits for different aesthetics at the same time. Having it adhere to any recycling rules would not make much sense

R/limits.r
hadley: Ah, it's `tidy::unpack()`; you might just mention "unpack data frame columns" as a comment. I don't think there's a function for this in vctrs yet, but @DavisVaughan can confirm.

R/guide-bins.R
thomasp85: interesting - I find the other way much more readable as it avoids multiple nested calls

R/facet-grid-.r
thomasp85: turns out we do as base R reintroduces row names when indexing repeated rows 😞

thomasp85:
This PR prompted a review of the use of `unrowname()`. During this it occurred to me that it would make sense to remove row names at the point of injection i.e. in `layer_data()` in order to make sure that row names are not present internally.

This means that `layer_data()` no longer returns an unmodified version of the input data.frame hence the changed unit tests, but I feel this is a fair bargain... @clauswilke @yutannihilation @Hadley please raise your voice if this is contentious to you:-)

yutannihilation:
I think that's fine. The conclusion of a discussion on `layer_data()` (https://github.com/tidyverse/ggplot2/issues/3018#issuecomment-442482864) was that `layer_data()` is for debugging purposes and should return what's actually used internally. So, if we use unnamed data frames internally, we should just return it.

PR #4875, opened 2022-06-14
Example of changing legend glyph for geom_text

ewallace:
A common use-case is to want to change the legend glyph for geom_text from 

E.G. This stack overflow question has 48k views:
https://stackoverflow.com/questions/18337653/remove-a-from-legend-when-using-aesthetics-and-geom-text

This pull request adds to `geom_text` help a simple example of replacing the a with a dot / [unicode black circle](https://util.unicode.org/UnicodeJsps/character.jsp?a=25CF), using `guide_legend`.

```r 
p <- ggplot(mtcars, aes(wt, mpg, label = rownames(mtcars)))
# ... etc

# Customise legend, change legend label from a to a dot
p + geom_text(aes(colour = factor(cyl))) +
  scale_colour_hue("Cylinders",
                   guide = guide_legend(override.aes = aes(label = "●")))
```

### DIFF
--- a/R/geom-path.r
+++ b/R/geom-path.r
@@ -34,8 +34,9 @@
 #' @examples
 #' # geom_line() is suitable for time series
 #' ggplot(economics, aes(date, unemploy)) + geom_line()
+#' # separate by colour and use "timeseries" legend key glyph
 #' ggplot(economics_long, aes(date, value01, colour = variable)) +
-#'   geom_line()
+#'   geom_line(key_glyph = "timeseries")
 #'
 #' # You can get a timeseries that run vertically by setting the orientation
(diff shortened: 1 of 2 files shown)


clauswilke:
I think adding one example each to `geom_text()` and `geom_line()` would be good. Those are probably the geoms where people most likely want to change the key glyph. For `geom_line()` I'd demonstrate the timeseries glyph.

If you can integrate it into one of the existing examples, so we don't increase the computational requirement to run all the examples, that's even better. I just looked quickly at `geom_text()`. It has tons of examples, including some with legends. One of two of those can be repurposed I think.

ewallace:
I've now integrated key_glyph into 1 example each of geom_text and geom_line. Hopefully that's good to merge when tests pass?

Thanks for the help and sorry for the delay.

PR #4884, opened 2022-06-19
Add rename_size=TRUE to GeomSf, GeomCrossbar, GeomPointrange, and GeomSmooth

### LINKED ISSUE #4883: 'size' no longer affects linewidth in 'geom_sf()'
I'm observing that changing `size` in `geom_sf()` no longer affects linewidth:

``` r
library("ggplot2")
library("sf")
#> Linking to GEOS 3.9.1, GDAL 3.3.2, PROJ 7.2.1; sf_use_s2() is TRUE
indices <- which(nc$CNTY_ID >= 1820 & nc$CNTY_ID <= 1830)
nc <- nc[indices, ]

p <- ggplot(nc, aes(fill = NAME)) +
  theme_bw(15)

p + geom_sf(size = 3)
```

Changing `linewidth` does affect linewidth:

``` r
p + geom_sf(linewidth = 3)
```

This breaking change isn't documented in the NEWS.md file and according to the discussion in https://github.com/tidyverse/ggplot2/pull/3672 it doesn't seem like such a "breaking change" was intended until after proper deprecation?:

@yutannihilation wrote:

> Just curious, will this eventually change the behavior of size on Geoms that actually have both size and linewidth? I know there's no breaking change here, but https://github.com/tidyverse/ggplot2/issues/4872 made me think this might need some consideration to avoid ambiguity.
>
> For example, in the case of geom_pointrange(), it's not clear what to do when we want to increase only the size of the points. Does specifying only size achieve this? Or, will this increase both the size of the points and that of the lines? I feel the former behavior is right, but it would be a breaking change.

and @thomasp85 replied:

> At this point the ambiguous geoms will not differentiate but I think we would do that down the line after proper deprecation

yutannihilation:
Fix #4883 

``` r
library(ggplot2)

ns <- asNamespace("ggplot2")
geoms_names <- ls(ns, pattern = "Geom.*")
geoms <- lapply(geoms_names, \(x) get(x, ns))

# geoms that use linewidth but don't have rename_size = TRUE
idx <- vapply(geoms, \(x) {
  use_linewidth <- "linewidth" %in% x$aesthetics()
  dont_rename_size <- !isTRUE(x$rename_size)
  use_linewidth && dont_rename_size
}, logical(1L))

geoms_names[idx]
#> [1] "GeomCrossbar"   "GeomPointrange" "GeomSf"         "GeomSmooth"
```

<sup>Created on 2022-06-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/geom-crossbar.r
+++ b/R/geom-crossbar.r
@@ -107,5 +107,7 @@ GeomCrossbar <- ggproto("GeomCrossbar", Geom,
       GeomPolygon$draw_panel(box, panel_params, coord, lineend = lineend, linejoin = linejoin),
       GeomSegment$draw_panel(middle, panel_params, coord, lineend = lineend, linejoin = linejoin)
     )))
-  }
+  },
+
+  rename_size = TRUE
 )
--- a/R/geom-pointrange.r
+++ b/R/geom-pointrange.r
@@ -57,5 +57,7 @@ GeomPointrange <- ggproto("GeomPointrange", Geom,
         GeomPoint$draw_panel(transform(data, size = size * fatten), panel_params, coord)
       ))
     )
-  }
+  },
+
+  rename_size = TRUE
 )
--- a/R/geom-sf.R
+++ b/R/geom-sf.R
@@ -149,7 +149,9 @@ GeomSf <- ggproto("GeomSf", Geom,
     } else {
       draw_key_polygon(data, params, size)
     }
-  }
+  },
+
+  rename_size = TRUE
 )
 
 default_aesthetics <- function(type) {
--- a/R/geom-smooth.r
+++ b/R/geom-smooth.r
@@ -160,5 +160,7 @@ GeomSmooth <- ggproto("GeomSmooth", Geom,
   optional_aes = c("ymin", "ymax"),
 
   default_aes = aes(colour = "#3366FF", fill = "grey60", linewidth = 1,
-    linetype = 1, weight = 1, alpha = 0.4)
+    linetype = 1, weight = 1, alpha = 0.4),
+
+  rename_size = TRUE
 )


trevorld:
Are you sure you should do `rename_size=TRUE` for all those geoms?  In particular shouldn't `geom_sf()` use both `size` to control the size of any "points" and `linewidth` to control the width of any "lines"?

yutannihilation:
In my understanding, when `rename_size = TRUE`, `size` affects both on the point size and the line width. When not, `size` affects only the point size. Am I wrong?

> Are you sure you should do `rename_size=TRUE` for all those geoms?

That's my intention, but I'm not fully sure, so I'd welcome your comments :)

``` r
library(ggplot2)

pt <- sf::st_point(c(2, 4))
ln <- sf::st_linestring(rbind(c(0, 1), c(5, 4)))
d <- sf::st_sf(id = 1:2, geometry = sf::st_sfc(pt, ln))

# current main branch
ggplot(d) +
  geom_sf(size = 10)
```



``` r
#> ℹ Loading ggplot2

# this pull request
ggplot(d) +
  geom_sf(size = 10)
```



<sup>Created on 2022-06-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

trevorld:
You're right (I misunderstood exactly what all of what `rename_size=TRUE` does).

yutannihilation:
I see. There might be better word than "rename", but I have no idea atm

thomasp85 (approved):

thomasp85:
Thanks - those were oversights on my part

yutannihilation:
Thanks for confirming! I'm getting to feel we need a bit more consideration on `rename_size`. I'll file a new issue so let's discuss there.

PR #4888, opened 2022-06-23
increase dig.lab until breeaks are unique

### LINKED ISSUE #4874: Issue with geom_contour_filled when the z values are too close together
There seems to be a bug in the function `geom_contour_filled` where it fails in some simple cases.

``` r
library(tidyverse)

#Generate some data
data = crossing(x = 1:10, y = 1:10) %>% 
  mutate(z = (x+y)/4)

#Works great
ggplot(data) + geom_contour_filled(aes(x, y, z = z))
```

``` r
#Make a minor tweak and it fails
data2 = data %>% mutate(z = z + 100)
ggplot(data2) + geom_contour_filled(aes(x, y, z = z))
#> Warning: Computation failed in `stat_contour_filled()`:
#> factor level [4] is duplicated
```

Looking into it, I believe the problem is with the legend names (more specifically the function `pretty_isoband_levels`). If the z values are close, you can get different levels with the same names because it only uses 3 digits (eg. "(101, 101]" and again "(101, 101]"). 

<sup>Created on 2022-06-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

thomasp85:
Fix #4874 

This PR introduces a gradual increasing of the number of digits until all breaks are uniquely identifiable

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # ggplot2 (development version)
 
+* Fix a bug in `stat_contour_filled()` where break value differences below a 
+  certain number of digits would cause the computations to fail (@thomasp85, 
+  #4874)
+
 * `stage()` now properly refers to the values without scale transformations for
   the stage of `after_stat`. If your code requires the scaled version of the
   values for some reason, you have to apply the same transformation by yourself,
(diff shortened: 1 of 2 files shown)


clauswilke (approved):
Looks good to me.

Issue #4601, opened 2021-09-02
Feature request: nicer error message for coord_cartesian


petyaracz:
I would like `coord_cartesian` to return a more human-readable error when I mess up the syntax:

## example

```r
library(ggplot2)

ggplot(iris, aes(Petal.Length,Petal.Width)) + 
  geom_point() +
  coord_cartesian(xlim(2,6))

```

## observed behaviour

```
Error in rep(no, length.out = len) : 
  attempt to replicate an object of type 'environment'
In addition: Warning message:
In is.na(coord_limits) :
  is.na() applied to non-(list or vector) of type 'environment'
```

## expected behaviour

```
Error in coord_cartesian: did you mean "xlim = c()"?
```

hadley:
Reprex:

``` r
library(ggplot2)

ggplot(iris, aes(Petal.Length,Petal.Width)) + 
  geom_point() +
  coord_cartesian(xlim(2,6))
#> Warning in is.na(coord_limits): is.na() applied to non-(list or vector) of type
#> 'environment'
#> Error in rep(no, length.out = len): attempt to replicate an object of type 'environment'
```

<sup>Created on 2022-04-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Just needs a little type checking in `coord_cartesian()`.

teunbrand:
Considered completed by #4894.

PR #4894, opened 2022-06-29
Nicer error message in coord-* functions

92amartins:
The original issue is that `coord-*` functions don't validate `xlim` and `ylim` arguments leading to confusing error output when these functions are called with bad arguments. See https://github.com/tidyverse/ggplot2/issues/4601 and https://github.com/tidyverse/ggplot2/issues/5297.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Nicer error messages for xlim/ylim arguments in coord-* functions
+  (@92amartins, #4601, #5297).
+
 * `coord_sf()` now uses customisable guides provided in the scales or 
   `guides()` function (@teunbrand).
 
--- a/R/coord-.R
+++ b/R/coord-.R
@@ -218,3 +218,22 @@ render_axis <- function(panel_params, axis, scale, position, theme) {
     zeroGrob()
   }
 }
+
+# Utility function to check coord limits
+check_coord_limits <- function(
+    limits, arg = caller_arg(limits), call = caller_env()
+) {
+  if (is.null(limits)) {
+    return(invisible(NULL))
+  }
+  if (!obj_is_vector(limits) || length(limits) != 2) {
+    what <- "{.obj_type_friendly {limits}}"
+    if (is.vector(limits)) {
+      what <- paste0(what, " of length {length(limits)}")
+    }
+    cli::cli_abort(
+      paste0("{.arg {arg}} must be a vector of length 2, not ", what, "."),
+      call = call
+    )
+  }
+}
--- a/R/coord-cartesian-.R
+++ b/R/coord-cartesian-.R
@@ -61,6 +61,8 @@
 #' d + coord_cartesian(xlim = c(0, 1))
 coord_cartesian <- function(xlim = NULL, ylim = NULL, expand = TRUE,
                             default = FALSE, clip = "on") {
+  check_coord_limits(xlim)
+  check_coord_limits(ylim)
   ggproto(NULL, CoordCartesian,
     limits = list(x = xlim, y = ylim),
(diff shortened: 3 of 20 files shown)


R/coord-cartesian-.r
```diff
@@ -61,6 +61,11 @@
 #' d + coord_cartesian(xlim = c(0, 1))
 coord_cartesian <- function(xlim = NULL, ylim = NULL, expand = TRUE,
                             default = FALSE, clip = "on") {
+  if(is.environment(xlim))
+    cli::cli_abort('Did you mean "xlim = c()"?')
+  if(is.environment(ylim))
+    cli::cli_abort('Did you mean "ylim = c()"?')
+
```
thomasp85, line 68:
```suggestion
  if (is.environment(xlim)) {
    cli::cli_abort("Did you mean {.code xlim = c(...)}?")
  }
  if (is.environment(ylim)) {
    cli::cli_abort('Did you mean {.code ylim = c(...)}?')
  }
```

thomasp85 (changes requested):
Thanks. Can I get you to update to match the style guide (suggested change), as well as add a unit test for the two errors

92amartins:
> Thanks. Can I get you to update to match the style guide (suggested change), as well as add a unit test for the two errors

Thanks, for the review @thomasp85!

I applied your suggestion regarding style in f047b04a0a99adbb00eb47d803d87e4e0a856f43

And wrote a unit test for the arguments in eec525e380f3a2a39c65662307afb44b7eb3b132

Please let me know if further improvements are needed.

92amartins:
Good point, @teunbrand! I'll extend the checks for other functions in the coord family.

teunbrand:
If you intend to recycle the checks in other `coord_*()` functions, it might be worth it to write a check function for coord limits. Here is a rough draft for one:

``` r
library(rlang)

check_coord_limits <- function(
  limits, arg = caller_arg(limits), call = caller_env()
) {
  if (is.null(limits)) {
    return(invisible(NULL))
  }
  if (!is.vector(limits) || length(limits) != 2) {
    what <- "{.obj_type_friendly {limits}}"
    if (is.vector(limits)) {
      what <- paste0(what, " of length {length(limits)}")
    }
    cli::cli_abort(
      paste0("{.arg {arg}} must be a vector of length 2, not ", what, "."),
      call = call
    )
  }
  invisible()
}

# Erroneous input
xlim <- ggplot2::xlim(0, 2)
ylim <- 1:3

check_coord_limits(xlim)
#> Error:
#> ! `xlim` must be a vector of length 2, not a
#>   <ScaleContinuousPosition/ScaleContinuous/Scale/ggproto/gg> object.
#> Backtrace:
#>     ▆
#>  1. └─global check_coord_limits(xlim)
#>  2.   └─cli::cli_abort(...)
#>  3.     └─rlang::abort(...)
check_coord_limits(ylim)
#> Error:
#> ! `ylim` must be a vector of length 2, not an integer vector of length
#>   3.
#> Backtrace:
#>     ▆
#>  1. └─global check_coord_limits(ylim)
#>  2.   └─cli::cli_abort(...)
#>  3.     └─rlang::abort(...)
```

<sup>Created on 2023-04-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
As pointed out in #5297, the lengthy check I proposed earlier might be better replaced by `vec_check_size(xlim, 2)`.

teunbrand (approved):
Thanks for doing this: this looks good to me.

thomasp85 (approved):
LGTM

teunbrand:
Thanks so much @92amartins for incorporating the commentary and for the PR!

PR #4898, opened 2022-07-05
Update ggplot2-specs.Rmd

jdonland:
Change first "Font face" subheading to "Font family".


yutannihilation (approved):
Thanks for catching!

PR #4904, opened 2022-07-09
Remove duplicated CoordPolar$render_fg()

yutannihilation:
For unknown reason, `CoordPolar` has two `render_fg()`s. I just happened to find this, and found no clue on the context why this happened (it has had the two versions since 2015: https://github.com/tidyverse/ggplot2/commit/1487e0100515312b4ff2777163d09fdaf4f86655).

I think this is almost harmless, but this should be removed to avoid confusion. The code is already diverged a bit, but I believe the second is the one actually used (e.g. https://github.com/tidyverse/ggplot2/pull/4115 goes only to the second one), so I remove the first one.

### DIFF
--- a/R/coord-polar.r
+++ b/R/coord-polar.r
@@ -245,41 +245,6 @@ CoordPolar <- ggproto("CoordPolar", Coord,
 
     # Combine the two ends of the scale if they are close
     theta <- theta[!is.na(theta)]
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM - I agree the second one should be kept

yutannihilation:
Thanks.

PR #4913, opened 2022-07-23
Do not mention soft-deprecated aes_()

yutannihilation:
Since `aes_()` is already soft-deprecated, let's not mention it in the doc. This pull request also update the version of roxgen2 to 7.2.1.


thomasp85 (approved):
LGTM - I can't see how any of the failing tests are related to this PR

yutannihilation:
As this change a lot of docs, I'm a bit afraid this will cause some conflicts on other pull requests. But, I need to merge this sooner or later anyway, so let's do now...

PR #4914, opened 2022-07-23
Show proper error when both x and y are specified on stat_density() and stat_ecdf()

### LINKED ISSUE #4776: Confusing error message and documentation for `geom_density`
The documentation and - in the greatest part - error message for `geom_density` is confusing and should be corrected.

Documentation states that the 
```
geom_density() understands the following aesthetics (required aesthetics are in bold):
- x
- y
```
When providing eiter `x` or `y` easthetic, everything works as expected. If trying to map BOTH `x` and `y`, following occurs:

```r
ggplot(diamonds, aes(x = x, y = y)) +
   geom_density()
   
Error in `check_required_aesthetics()`:
! geom_density requires the following missing aesthetics: y
```
Whereas the `y` aesthetic is clearly provided. IMO it should lead the user to the `geom_density2d`.

yutannihilation:
Fix #4776

### DIFF
--- a/R/stat-bindot.r
+++ b/R/stat-bindot.r
@@ -8,16 +8,7 @@ StatBindot <- ggproto("StatBindot", Stat,
   default_aes = aes(y = after_stat(count)),
   dropped_aes = c("bin", "bincenter"), # these are temporary variables that are created and then removed by the stat
 
-  setup_params = function(self, data, params) {
-    has_x <- !(is.null(data$x) && is.null(params$x))
-    has_y <- !(is.null(data$y) && is.null(params$y))
-    if (!has_x && !has_y) {
-      cli::cli_abort("{.fn {snake_class(self)}} requires an {.field x} or {.field y} aesthetic.")
-    }
-    if (has_x && has_y) {
-      cli::cli_abort("{.fn {snake_class(self)}} must only have an {.field x} {.emph or} {.field y} aesthetic.")
-    }
-
+  setup_params = function(data, params) {
     if (is.null(params$binwidth)) {
       cli::cli_inform("Bin width defaults to 1/30 of the range of the data. Pick better value with {.arg binwidth}.")
     }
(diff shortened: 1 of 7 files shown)


thomasp85 (changes requested):
We are now doing unit tests for all errors so can I get you to add tests for this

yutannihilation:
Sure

yutannihilation:
It seems this existing code will be broken by this change. I'm still thinking, but this might mean the same case can be found with `stat_bin()` and `stat_count()`...?

``` r
library(ggplot2)

do_plot <- function() {
  ggplot(mtcars, aes(mpg, factor(cyl))) +
    stat_density(aes(fill = after_stat(density)), geom = "raster", position = "identity")
}

do_plot()
```



``` r

#> ℹ Loading ggplot2
do_plot()
#> Error in `stat_density(aes(fill = after_stat(density)), geom = "raster", position =
#> "identity")`:
#> ! Problem while computing stat.
#> Caused by error in `setup_params()` at ggplot2/R/ggproto.r:182:16:
#> ! `stat_density()` must only have an x or y aesthetic.
```



<sup>Created on 2022-07-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

clauswilke:
@yutannihilation The example you found that breaks is an important use case I think. Fundamentally this highlights the core problem, which is that we don't explicitly distinguish between aesthetics given to the stat and to the geom. We may have to accept the confusing error message.

yutannihilation:
Thanks. Yes, I'm almost giving up this pull request. Rather, my current concern is if `stat_bin()` and `stat_count()` might be also wrong. The below example won't work even if I remove "must only have an x or y" check, but I don't figure out if this should work like the example of `stat_density()` above or not...

``` r
ggplot(mtcars, aes(mpg, factor(cyl))) +
  stat_bin(aes(fill = after_stat(count)), geom = "raster", position = "identity")
```

yutannihilation:
As this pull request seems impossible, I'm giving up this. Sorry that I didn't come up with the usage.

Issue #4394, opened 2021-03-28
Add a warning in geom_bar when mapping of a continuous variable to the fill aesthetic is not possible?


svigneau:
Consider the following plot:

```r
demo <- diamonds[1:10, ] 
ggplot(data = demo, mapping = aes(x = cut, fill = price)) +
  geom_bar()
```

Among the 10 diamonds in the "demo" datasets, there is only one diamond with a "Fair" cut and one diamond with an "Ideal" cut. Each of these diamonds has a price and the corresponding bar is colored based on this price using a color scale, in agreement with the aes_colour_fill_alpha documentation. Bars for the other cuts, however, represent several diamonds which all have different prices, making such mapping impossible, and the corresponding bars are consequently plotted in the default gray color.

This behavior tends to be confusing to the user as there is no explanation for why the mapping of prices to the fill aesthetics happens for some bars but not others. This gets even more confusing when no mapping happens at all, as for instance with:

```r
ggplot(data = diamonds, mapping = aes(x = cut, fill = price)) +
  geom_bar()
```

I think usability could be improved by printing an informative warning message in situations where continuous variables cannot be mapped to the fill aesthetics, and by updating the aes_colour_fill_alpha documentation to indicate the expected behavior in such situations.

yutannihilation:
When the `fill` columns get dropped in all the groups, it's captured by this logic introduced in #4866. But, when there's any undropped one, it fails to capture.

https://github.com/tidyverse/ggplot2/blob/a979ffd26cdb456d54e2671c2eed16c65bc878b7/R/stat-.r#L135-L145

``` r
library(ggplot2)

d <- data.frame(
  id = c("a", "a", "b", "b", "c"),
  attr = 1:5
)

ggplot(d[1:4, ], aes(id, fill = attr)) +
  geom_bar()
#> Warning: The following aesthetics were dropped during statistical transformation: fill
#> ℹ This can happen when ggplot fails to infer the correct grouping structure in
#>   the data.
#> ℹ Did you forget to specify a `group` aesthetic or to convert a numerical
#>   variable into a factor?
```

![](https://i.imgur.com/YOoVi5B.png)

``` r

ggplot(d, aes(id, fill = attr)) +
  geom_bar()
```

![](https://i.imgur.com/S56M85h.png)

<sup>Created on 2022-07-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

clauswilke:
[...]

PR #4917, opened 2022-07-24
Drop non-constant aesthetics more thoroughly

yutannihilation:
Fix https://github.com/tidyverse/ggplot2/issues/4394

ggplot2 drops an aesthetic when the values are not constant within the group. Currently, this happens per group. Even when the dropping happens, if any of the groups still have the aesthetics, the result keeps the column because `vec_rbind()` kindly fills the mismatch with `NA`.

https://github.com/tidyverse/ggplot2/blob/10eba36a92f9a0f91a9a5a11b27f38fe7a2432ae/R/stat-.r#L133

This creates incomplete result like below. As discussed on #4394, this should drop the aesthetic instead of just showing a warning (which I tried in #4916).

``` r
library(ggplot2)

d <- data.frame(
  id = c("a", "a", "b", "b", "c"),
  attr = 1:5
)

ggplot(d, aes(id, fill = attr)) +
  geom_bar()
```

<sup>Created on 2022-07-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This pull request inlines `uniquecols()` to catch the dropped columns and record it to an empty env so that we can refer to it. I feel this is better rather than inferring from the results what columns are dropped, but I'm open to any better idea.

### DIFF
--- a/R/stat-.r
+++ b/R/stat-.r
@@ -120,13 +120,21 @@ Stat <- ggproto("Stat",
       self$compute_group(data = group, scales = scales, ...)
     })
 
+    # record dropped columns
+    dropped_columns <- new_environment()
     stats <- mapply(function(new, old) {
       if (empty(new)) return(data_frame0())
-      unique <- uniquecols(old)
-      missing <- !(names(unique) %in% names(new))
-      vec_cbind(
+
+      # ignore the columns that will be overwritten by `new`
+      old <- old[, !(names(old) %in% names(new)), drop = FALSE]
+
+      # drop columns that are not constant within group
+      unique_idx <- vapply(old, function(x) length(unique0(x)) == 1, logical(1L))
+      env_bind(dropped_columns, !!!set_names(names(old)[!unique_idx]))
+
+      result <- vec_cbind(
         new,
-        unique[rep(1, nrow(new)), missing,drop = FALSE]
+        old[rep(1, nrow(new)), unique_idx, drop = FALSE]
       )
     }, stats, groups, SIMPLIFY = FALSE)
 
@@ -135,7 +143,8 @@ Stat <- ggproto("Stat",
     # The above code will drop columns that are not constant within groups and not
     # carried over/recreated by the stat. This can produce unexpected results,
     # and hence we warn about it.
-    dropped <- base::setdiff(names(data), base::union(self$dropped_aes, names(data_new)))
+    dropped <- ls(dropped_columns)
+    dropped <- dropped[!dropped %in% self$dropped_aes]
     if (length(dropped) > 0) {
       cli::cli_warn(c(
         "The following aesthetics were dropped during statistical transformation: {.field {glue_collapse(dropped, sep = ', ')}}",
--- a/R/stat-.r
+++ b/R/stat-.r
@@ -152,7 +152,7 @@ Stat <- ggproto("Stat",
         "i" = "Did you forget to specify a {.code group} aesthetic or to convert a numerical variable into a factor?"
       ))
     }
-    data_new
+    data_new[, !names(data_new) %in% dropped, drop = FALSE]
   },
 
   compute_group = function(self, data, scales) {
--- a/R/stat-.r
+++ b/R/stat-.r
@@ -120,31 +120,50 @@ Stat <- ggproto("Stat",
       self$compute_group(data = group, scales = scales, ...)
     })
 
-    # record dropped columns
-    dropped <- character(0)
+    # Record columns that are not constant within groups. We will drop them later.
+    non_constant_columns <- character(0)
+
     stats <- mapply(function(new, old) {
+      # In this function,
+      #
+      #   - `new` is the computed result. All the variables will be picked.
+      #   - `old` is the original data. There are 3 types of variables:
+      #     1) If the variable is already included in `new`, it's ignored
+      #        because the values of `new` will be used.
+      #     2) If the variable is not included in `new` and the value is
+      #        constant within the group, it will be picked.
+      #     3) If the variable is not included in `new` and the value is not
+      #        constant within the group, it will be dropped. We need to record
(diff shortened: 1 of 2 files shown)


clauswilke:
@thomasp85 A few more days without proper internet for me. Should be able to look this over carefully early next week.

thomasp85:
@clauswilke no problem at all, was just going through open PRs and this seemed to belong in your review bucket given the issue discussion — no stress

thomasp85:
@clauswilke will you have time for this one within the next couple of weeks? Sorry to keep pinging you — if you are floored by other work just let me know

clauswilke:
@thomasp85 I'll look it over today.

yutannihilation:
Thanks. Honestly, I'm not sure if this is the best way to record the dropped columns. Any suggestion is welcome.

clauswilke:
@yutannihilation Yeah, that's why I've been stalling. I'm not entirely sure what to think. The process of making an environment, adding things to it, and then listing all the contents seems very awkward. Also difficult to reason about. Have you considered working with a simple vector of names and using set functions (`union()`) to combine sets of names?

yutannihilation:
I think the idea of using an env as a hashmap is not so uncommon. I came up with several ideas including yours, and I just couldn't figure out which one is most efficient (and didn't have time to benchmark them).

So, do you mean something like this?

``` r
    dropped_columns <- list()
    stats <- mapply(function(new, old) {
      ...

      dropped_columns <- append(dropped_columns, names(old)[!unique_idx])

      ...
    }, stats, groups, SIMPLIFY = FALSE)

    ...

    dropped <- Reduce(dropped_columns, union)
```

clauswilke:
Would this work?

```
dropped_columns <- character(0)
stats <- mapply(function(new, old) {
    ...
    dropped_columns <- c(dropped_columns, names(old)[!unique_idx])
    ...
}, stats, groups, SIMPLIFY = FALSE)
...
dropped <- unique(dropped_columns)
```
Either way, much simpler to read than the code using environments. For code that is not performance critical, I'd always go with the code that is easier to read than the one that (potentially) is a little faster.

clauswilke:
Or alternatively:

```
dropped <- character(0)
stats <- mapply(function(new, old) {
    ...
    dropped <- union(dropped, names(old)[!unique_idx])
    ...
}, stats, groups, SIMPLIFY = FALSE)
...
```

yutannihilation:
Thanks, makes sense. I'll use the first one.

Another one I attempted was returning a list rather than a data.frame (and I forgot to remove `result <-`...).

``` r
    stats <- mapply(function(new, old) {
      ...
      dropped <- union(dropped, names(old)[!unique_idx])
      
      result <- vec_cbind(
        new,
        old[rep(1, nrow(new)), unique_idx, drop = FALSE]
      )
      
      list(
        result = result,
        dropped = dropped
      )
    }, stats, groups, SIMPLIFY = FALSE)

    data_new <- vec_rbind(!!!lapply(stats, function(x) x[["result"]])

    dropped <- vec_unique(vctrs_c(!!!lapply(stats, function(x) x[["dropped"]])))
```

clauswilke:
@yutannihilation Thanks! I have two more questions:

1. What is `unique0()`?
2. What happens when different columns get dropped in different groups. How does the data frame get assembled again? Would it make sense to add a test for this case, to make sure things work as expected?

yutannihilation:
> What is `unique0()`?

It's defined here. After the vctrs integration (#4868), most of the code uses `unique0()`, so I followed it.

https://github.com/tidyverse/ggplot2/blob/c86ed7bc59060d94fb070fe1bb1574e8b6026ff8/R/utilities.r#L355

> Would it make sense to add a test for this case, to make sure things work as expected?

Sure. I'll add tests.

yutannihilation:
Added tests.

> 2\. What happens when different columns get dropped in different groups. How does the data frame get assembled again?

What do you mean by "what happens" here? All get dropped. Do you expect anything more? Sorry, I'm not sure if I understand the intention of this question.

clauswilke:
@yutannihilation I was just trying to understand how the code works. There's something that I'm still missing. You're creating separate data frames for each group, and then you're row-binding them all together. That's fine, but what happens when the data frames for different groups have different columns. Shouldn't row-binding fail in this case? But now looking it up, it seems like `vec_rbind()` can handle this scenario, so I guess that's why it works.

R/stat-.r
```diff
@@ -135,15 +143,16 @@ Stat <- ggproto("Stat",
...
+    dropped <- dropped[!dropped %in% self$dropped_aes]
     if (length(dropped) > 0) {
       cli::cli_warn(c(
         "The following aesthetics were dropped during statistical transformation: {.field {glue_collapse(dropped, sep = ', ')}}",
         "i" = "This can happen when ggplot fails to infer the correct grouping structure in the data.",
         "i" = "Did you forget to specify a {.code group} aesthetic or to convert a numerical variable into a factor?"
       ))
     }
-    data_new
+    data_new[, !names(data_new) %in% dropped, drop = FALSE]
```
clauswilke, line 155: I'm not sure the logic here is quite right. Shouldn't line 155 come before line 147? We want to drop all columns in `dropped`, not just the ones that are not in `self$dropped_aes`.

yutannihilation: Thinking this again, it seems you are right. Thanks for catching.

yutannihilation:
I added many comments and rewrote the code a bit. I hope this will make it a bit easier to understand the logic...

clauswilke (approved):
Thanks, looks great now!

yutannihilation:
Thanks so much for reviewing this complicated pull request! I hope I can do better next time...!

clauswilke:
I think you did great. Just because there's room for improvement doesn't mean the first iteration was bad.

Issue #4899, opened 2022-07-05
Simplify alignment for column geoms


wurli:
Currently the alignment of columns is always centre, which may not always be desired. E.g. in the following case, values of `date` always give the first of the month, but are used to indicate the whole month (as is fairly common practice):

```
library(ggplot2)
library(dplyr, warn.conflicts = FALSE)
library(lubridate, warn.conflicts = FALSE)

df <- tibble(
  month = as_date(c("2020-01-01", "2020-02-01", "2020-03-01")),
  value = 1:3
)

ggplot(df, aes(month, value)) +
  geom_col() +
  scale_x_date(date_labels = "%b %d")
```
![image](https://user-images.githubusercontent.com/17475731/177399041-7ff49399-2ad7-40e7-a15b-680235652a5c.png)

In this case an `align` argument to `geom_col()` would be really useful to align the columns with the first of each month. `align` could accept values `"centre"` (the default), `"right"` and `"left"`, which would be the option used here. The current alternatives are to use `position = position_nudge()`, which is fairly esoteric for such a simple task (and wouldn't always work that well, e.g. since February only has 28 days), or to instead use `geom_rect()`, which again seems much too complex for such a simple task.

If you agree that this sounds like a useful feature I'd be happy to submit a PR.

As always, thanks for the hard work on this beautiful package! 

(N.B, this example is a bit contrived due to the use of `scale_x_date()` but it's the simplest example I could think of)

yutannihilation:
Thanks for the suggestion, but this doesn't sound convincing to me. The problem of this example looks the width of each bar, rather than the alignment? (not sure if the intention of your example is to show daily values or monthly values)

``` r
library(ggplot2)
library(dplyr, warn.conflicts = FALSE)
library(lubridate, warn.conflicts = FALSE)

df <- tibble(
  month = as_date(c("2020-01-01", "2020-02-01", "2020-03-01")),
  value = 1:3
)

ggplot(df, aes(month, value)) +
  geom_col(width = 1) +
  scale_x_date(date_labels = "%b %d")
```

![](https://i.imgur.com/zfRXvQw.png)

<sup>Created on 2022-07-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

wurli:
[...]

Issue #3798, opened 2020-02-06
Remove GeomCol


thomasp85:
GeomCol is a direct duplicate of GeomBar. We should remove it and simply make `geom_col()` a GeomBar+StatIdentity call

teunbrand:
Should GeomCol go through some deprecation process? We could have `ggproto()` throw an error when you attempt to inherit from GeomCol.

teunbrand:
Closing this issue in favour of broader scope of #5962

PR #4921, opened 2022-07-26
Feat/issue 4899 simplify alignment for column geoms

wurli:
This PR adds a `just` argument to `geom_col()` and `geom_bar()` (#4899) allowing a user to more easily change the alignment of columns. In particular, it makes it easy to align the sides of columns with axis breaks in cases where the `resolution()` of the data is not trivial to calculate. 

* `just` is used rather than `hjust`. This has a downside that it breaks with the `hjust`/`vjust` pattern used elsewhere in ggplot2, however it seems preferable since `hjust` would necessitate `vjust` for horizontal columns, which would raise difficult questions about whether to modify other behaviour, e.g. of `has_flipped_aes()`. 
* There is a small improvement to the documentation for the `width` argument - _resolution of the data_ now reads as _`resolution()` of the data_ to more clearly hint at what's happening under the hood
* This PR does not (as yet) include the extra aesthetics `xmin` and `ymin` for `geom_col()`

Here's how to use the `just` argument:
``` r
devtools::load_all()
#> i Loading ggplot2

df <- data.frame(
  x = 1:3,
  y = 1:3
)

ggplot(df, aes(y, x)) +
  geom_col(just = 1)
```

<sup>Created on 2022-07-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


yutannihilation:
Thanks. Could you avoid using the Windows line ending? If you are using Windows, you can set it here (you might also need to review your Git settings as Git might convert the line endings on committing).

image

wurli:
> Thanks. Could you avoid using the Windows line ending? If you are using RStudio, you can set it here (you might also need to review your Git settings as Git might convert the line endings on committing).

Thanks, done now. I was puzzling about why the diff was showing all files as changed!

R/geom-bar.r
```diff
@@ -123,16 +129,18 @@ GeomBar <- ggproto("GeomBar", GeomRect,
...
 
-  extra_params = c("na.rm", "orientation"),
+  extra_params = c("just", "na.rm", "orientation"),
 
   setup_data = function(data, params) {
     data$flipped_aes <- params$flipped_aes
     data <- flip_data(data, params$flipped_aes)
     data$width <- data$width %||%
       params$width %||% (resolution(data$x, FALSE) * 0.9)
+    data$just <- data$just %||% params$just %||% 0.5
```
yutannihilation, line 139:
```suggestion
    data$just <- params$just %||% 0.5
```

In `width`'s case, `width` might be provided by the `Stat`; e.g. the bin widths of a histogram can vary so the computed data contains the `width`s. But, there's no Stat that returns `just`.

R/geom-bar.r
```diff
@@ -40,7 +40,11 @@
 #' automatically determines the orientation from the aesthetic mapping. In the
 #' rare event that this fails it can be given explicitly by setting `orientation`
 #' to either `"x"` or `"y"`. See the *Orientation* section for more detail.
-#' @param width Bar width. By default, set to 90% of the resolution of the data.
+#' @param just Adjustment for column placement. Set to `0.5` by default, meaning
+#'   that columns will be centered about axis breaks. Set to `0` or `1` to place
+#'   columns to the left/right of axis breaks.
+#' @param width Bar width. By default, set to 90% of the `resolution()` of the
```
yutannihilation, line 46: Good idea! How about using ``[`resolution()`]`` to link the document?

yutannihilation (changes requested):
Thanks, looks almost good. Random comments:

* I think we should add at least one example to show how `just` works briefly.
* When used with some positions (e.g. `position_dodge()`), non-default `just` might result in a bit weird look. I'm wondering if we can add some friendly note about this.

R/geom-bar.r
wurli: Makes sense, thanks!

R/geom-bar.r
wurli: Sounds good 👍

wurli:
> Thanks, looks almost good. Random comments:
> * I think we should add at least one example to show how just works briefly.
> * When used with some positions (e.g. `position_dodge()`), non-default just might result in a bit weird look. I'm wondering if we can add some friendly note about this.

Both of these are now complete. The friendly note is in the documentation for the `just` argument itself, but may be more appropriate in the example. Not a big deal either way though 😄 Thanks!

R/geom-col.r
```diff
@@ -42,16 +44,19 @@ GeomCol <- ggproto("GeomCol", GeomRect,
...
 
   setup_data = function(data, params) {
     data$flipped_aes <- params$flipped_aes
     data <- flip_data(data, params$flipped_aes)
     data$width <- data$width %||%
       params$width %||% (resolution(data$x, FALSE) * 0.9)
-    data <- transform(data,
-                      ymin = pmin(y, 0), ymax = pmax(y, 0),
-                      xmin = x - width / 2, xmax = x + width / 2, width = NULL
+    data$just <- data$just %||% params$just %||% 0.5
```
yutannihilation, line 54: Ah, sorry, I forgot to mention this `data$just` as well.

R/geom-bar.r
```diff
@@ -123,16 +138,18 @@ GeomBar <- ggproto("GeomBar", GeomRect,
...
     data$flipped_aes <- params$flipped_aes
     data <- flip_data(data, params$flipped_aes)
     data$width <- data$width %||%
       params$width %||% (resolution(data$x, FALSE) * 0.9)
+    data$just <- params$just %||% 0.5
     data <- transform(data,
       ymin = pmin(y, 0), ymax = pmax(y, 0),
-      xmin = x - width / 2, xmax = x + width / 2, width = NULL
+      xmin = x - width * (1 - just), xmax = x + width * just,
+      width = NULL, just = NULL
```
yutannihilation, line 152:
```suggestion
      width = NULL
```

This `width = NULL` is to remove `width` column (if exists) passed from the `Stat`. As the data doesn't have `just` column, we don't need this `just = NULL`.

R/geom-col.r
```diff
@@ -42,16 +44,19 @@ GeomCol <- ggproto("GeomCol", GeomRect,
...
+    data <- transform(
+      data,
+      ymin = pmin(y, 0), ymax = pmax(y, 0),
+      xmin = x - width * (1 - just), xmax = x + width * just,
+      width = NULL, just = NULL
```
yutannihilation, line 59: This `just = NULL` as well.

yutannihilation (commented):
Thanks! The friendly note looks good. Sorry, I failed to catch some more minor points. Also, could you resolve the conflict on NEWS.md?

R/geom-col.r
wurli: Sorry, I'm not quite sure what approach you're advocating... Currently we have this:
```
    data$just <- data$just %||% params$just %||% 0.5
    data <- transform(
      data,
      ymin = pmin(y, 0), ymax = pmax(y, 0),
      xmin = x - width * (1 - just), xmax = x + width * just,
      width = NULL, just = NULL
    )
```
Are you saying we shouldn't remove the `just` column? In that case we'd have this, but that seems a bit odd as `width` gets removed:
```
    data$just <- data$just %||% params$just %||% 0.5
    data <- transform(
      data,
      ymin = pmin(y, 0), ymax = pmax(y, 0),
      xmin = x - width * (1 - just), xmax = x + width * just,
      width = NULL
    )
```

Do you think the whole `setup_data` field for `geom_col()` and `geom_bar()` should have its own named function (e.g. `setup_col_data()`) to avoid code duplication btw? For my money the logic is complex enough that it'd be quite a bit safer to have it in one place.

yutannihilation: Really sorry, for some reason, I simply didn't read this line... You are right. Please forget my comment above.

```r
data$just <- data$just %||% params$just %||% 0.5
```

wurli: Haha, I did wonder! No worries at all 😄

R/geom-col.r
yutannihilation: Sorry for the confusion, but I think I was sane on this comment.

yutannihilation:
> Do you think the whole `setup_data` field for `geom_col()` and `geom_bar()` should have its own named function (e.g. `setup_col_data()`) to avoid code duplication btw?

I can't remember the reason why `GeomBar` and `GeomCol` are implemented separately, but, as now they are just duplicate. There's a plan to remove it (https://github.com/tidyverse/ggplot2/issues/3798), but, for now, could you just leave them duplicate? I think we should do one by one..

R/geom-col.r
wurli: Ah! Well spotted, it took me way to long to see what you were talking about there. I'll make the change.

yutannihilation (approved):
Thanks! No further comments from my side.

@thomasp85 
Could you also take a brief look when you have time?

thomasp85 (approved):

thomasp85:
LGTM

yutannihilation:
Thanks for reviewing!

PR #4922, opened 2022-07-26
check the input faceting names

### LINKED ISSUE #4138: facets fail with uninformative error for some column names

thomasp85:
Fix #4138 

Provide meaningful error when using reserved names in faceting specs. Down the line we may want to explore removing the constraints altogether but for now a better error is all we need

### DIFF
--- a/R/facet-.r
+++ b/R/facet-.r
@@ -486,6 +486,17 @@ check_layout <- function(x) {
   cli::cli_abort("Facet layout has a bad format. It must contain columns {.col PANEL}, {.col SCALE_X}, and {.col SCALE_Y}")
 }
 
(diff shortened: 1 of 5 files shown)


yutannihilation (approved):
Looks good.

PR #4933, opened 2022-08-07
fix dropping of inf values from jitter plot

### LINKED ISSUE #4790: geom_point() and geom_jitter(width = 0, height = 0) have different behaviors for zero values in scale_y_log10()
I'm using ggplot2 3.3.5 and found that `geom_point()` and `geom_jitter(height = 0, width = 0)` have different behaviors for dealing with zero values in log-transformed axes.  In particular, geom_point() represents zeros as dots lying on the x-axis whereas geom_jitter() does not plot the zeros at all.  For me, this is relevant, because sometimes I want the jitter width to be non-zero, but height to be zero, and I don't want my dots to disappear!  Right now I work around the issue by setting the zeros to some small value I choose to make the plot nice.

```
data = data.frame(x = c(0, 0, 1, 1, 2, 2), y = c(0, 0, 0, 1, 2, 3))
ggplot(data, aes(x = x, y = y)) + geom_point() + scale_y_log10()
ggplot(data, aes(x = x, y = y)) + geom_jitter(height = 0, width = 0) + scale_y_log10()
```
image

javlon:
Fix #4790 

This PR fixes a bug where infinity values were dropped in a jitter plot.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Fix a bug in `position_jitter()` where infinity values were dropped (@javlon,
+  #4790).
+
 * `geom_linerange()` now respects the `na.rm` argument (#4927, @thomasp85)
 
 * Improve the support for `guide_axis()` on `coord_trans()` (@yutannihilation, #3959)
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good! Thanks. Could you add a NEWS item on top of https://github.com/tidyverse/ggplot2/blob/main/NEWS.md ?

c.f. https://style.tidyverse.org/news.html

yutannihilation:
Perfect, thanks!

PR #4936, opened 2022-08-09
fix misplacing points in position_dodge2

### LINKED ISSUE #4327: point is misplaced if width = 0 is used in position_dodge2 with geom_point
If `position_dodge2(width = 0)` is used with the `position` argument in `geom_point()` (one?) is wrongly placed horizontally (i. e., wrong 'column' for discrete values). This only seems to be an issue if `width = 0`. 
If `position_dodge2()` is not intended to be used with `geom_point()` a warning would be helpful.

``` r
library(tidyverse)

df <- tribble(~dimension, ~group, ~value,
              "dim 1", "A", 0.6,
              "dim 1", "B", 0.7,
              "dim 2", "A", 0.4,
              "dim 2", "B", 0.5)

plot <- ggplot(data = df,
               mapping = aes(x = dimension,
                             y = value,
                             color = group))  

# point in wrong column (point of 'dim 1' 'B' appears in column of 'dim 2'
plot +  geom_point(
             position = position_dodge2(width = 0))
```

``` r

# correct column
plot + geom_point(
  position = position_dodge2(width = 0.5))
```

<sup>Created on 2021-01-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

javlon:
Fix #4327 

This PR fixes the bug that misplaced geom objects when they had zero width.

### DIFF
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -138,7 +138,7 @@ find_x_overlaps <- function(df) {
   overlaps[1] <- counter <- 1
 
   for (i in seq_asc(2, nrow(df))) {
-    if (is.na(df$xmin[i]) || is.na(df$xmax[i - 1]) || df$xmin[i] >= df$xmax[i - 1]) {
+    if (is.na(df$xmin[i]) || is.na(df$xmax[i - 1]) || df$xmin[i] > df$xmax[i - 1]) {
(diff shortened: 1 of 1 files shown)


yutannihilation (approved):
Looks good. Thanks!

PR #4944, opened 2022-08-13
fix alignment problem

### LINKED ISSUE #4350: alignment problem with preserve = "single" and preserve = "total"
This issue is a follow-up on https://github.com/tidyverse/ggplot2/issues/2712#issue-335176790.

The previously posted solution does not work when there are 3 grouping levels, in which case the geom_point and geom_boxplot layers no longer match up again.

Reprex:
```r
mtcars2 <- bind_rows(mtcars, data.frame(gear = c(4,5), mpg = c(22, 19), am = c(2, 2)))
ggplot(mtcars2, aes(factor(gear), mpg, fill = factor(am))) + 
   geom_boxplot(position = position_dodge2(0.75, preserve = 'single')) +
   geom_point(position = position_dodge(0.75, preserve = 'total'))
```

javlon:
Fix #4350 

This PR fixes the alignment problem. The problem is fixed by creating normalized fake width.

### DIFF
--- a/R/position-dodge2.r
+++ b/R/position-dodge2.r
@@ -139,8 +139,7 @@ find_x_overlaps <- function(df) {
   overlaps[1] <- counter <- 1
 
   for (i in seq_asc(2, nrow(df))) {
-    if (is.na(df$xmin[i]) || is.na(df$xmax[i - 1]) || df$xmin[i] > df$xmax[i - 1] ||
-      (df$xmin[i] == df$xmax[i - 1] && df$xmin[i] != df$xmax[i])) {
+    if (is.na(df$xmin[i]) || is.na(df$xmax[i - 1]) || df$xmin[i] >= df$xmax[i - 1]) {
       counter <- counter + 1
     }
     overlaps[i] <- counter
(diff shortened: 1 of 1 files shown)


thomasp85:
@karawoo would you be able to take a look at this?

also, @javlon have you fixed the issue pointed out by @smouksassi in #4350?

javlon:
> @karawoo would you be able to take a look at this?
> 
> also, @javlon have you fixed the issue pointed out by @smouksassi in #4350?

Sorry I haven't seen this issue. The only way to fix mentioned issue requires changes in the `position_dodge` algorithm. Changes include information about the `width` of the boxplot (or any other object) from `position_dodge2`. It means `position_dodge` should know information about other layers which might lead to more trouble in the future than the proposed solution. 

Above PR include changes only in the `position_dodge2` algorithm, and now it has the same algorithm for position selection as `position_doge`.

Yes, the distance between boxplots has changed slightly, but the overall width for each group is constant.

karawoo:
I should say @javlon that I really appreciate you working on this and your solution does look really good for basic boxes + points. It's just a tricky problem with a lot of edge cases, unfortunately 😕

javlon:
Thank you, @karawoo, for your clear explanation of the problem. I forgot about this kind of usage of boxplots.

PR #4945, opened 2022-08-14
revert pr for position_dodge2

javlon:
Reverted PR #4936 for #4327 
Added test for the bug with the previous solution


yutannihilation (approved):
Thanks! Sorry for troubling you...

PR #4950, opened 2022-08-20
Warn on the use of `..var..` and `stat(var)`

### LINKED ISSUE #3693: Warn on `stat()` and `..var..`-notation?
After #3534, now we have 3 different ways to access the calculated variables:

* `after_stat(var)`
* `stat(var)`
* `..var..`

Though it seems very difficult to remove other syntax than `after_stat()`, I think we should promote the most proper one a bit more eagerly to avoid confusion. Can we warn on the usage of `stat()` and `..var..`-notation?

yutannihilation:
Fix #3693

Considering the wide use of them, now I feel it's a bit difficult to deprecate them. Still, I think it's a good idea to issue a non-warning message at least because we want to advertise `after_stat()` more.

### DIFF
--- a/R/aes-evaluation.r
+++ b/R/aes-evaluation.r
@@ -112,8 +112,6 @@ is_calculated <- function(x, warn = FALSE) {
   } else if (is.symbol(x)) {
     res <- is_dotted_var(as.character(x))
     if (res && warn) {
-      # TODO: if we want to set always = TRUE, we probably need some mechanism
-      # to limit the warning only once per plot build
       lifecycle::deprecate_warn("3.4.0", I("The dot-dot notation (`..var..`)"), "after_stat()")
     }
     res
@@ -122,8 +120,6 @@ is_calculated <- function(x, warn = FALSE) {
   } else if (is.call(x)) {
     if (identical(x[[1]], quote(stat))) {
       if (warn) {
-        # TODO: if we want to set always = TRUE, we probably need some mechanism
-        # to limit the warning only once per plot build
         lifecycle::deprecate_warn("3.4.0", "stat()", "after_stat()")
       }
       TRUE
--- a/R/aes-evaluation.r
+++ b/R/aes-evaluation.r
@@ -92,9 +92,20 @@ is_dotted_var <- function(x) {
   grepl(match_calculated_aes, x)
 }
 
+# TODO: if we want to set the frequency "always", we probably need some
+# mechanism to generate a frequency ID that is unique per plot build. Otherwise,
+# users might face too many warnings.
+warn_old_calculated_aes <- function() {
+  # TODO: move to cli_warn()
+  cli::cli_inform(c(
+    "The dot-dot notation ({.var ..var..}) and {.fun stat} have been superseded",
+    "i" = "To access calculated variables, please use {.fun after_stat} instead"
+  ), .frequency = "regularly", .frequency_id = "ggplot2-old-calculated-aes")
+}
+
 # Determine if aesthetic is calculated
-is_calculated_aes <- function(aesthetics) {
-  vapply(aesthetics, is_calculated, logical(1), USE.NAMES = FALSE)
+is_calculated_aes <- function(aesthetics, warn = FALSE) {
+  vapply(aesthetics, is_calculated, warn = warn, logical(1), USE.NAMES = FALSE)
 }
 is_scaled_aes <- function(aesthetics) {
   vapply(aesthetics, is_scaled, logical(1), USE.NAMES = FALSE)
@@ -102,7 +113,7 @@ is_scaled_aes <- function(aesthetics) {
 is_staged_aes <- function(aesthetics) {
   vapply(aesthetics, is_staged, logical(1), USE.NAMES = FALSE)
 }
-is_calculated <- function(x) {
+is_calculated <- function(x, warn = FALSE) {
   if (is_call(get_expr(x), "after_stat")) {
     return(TRUE)
   }
@@ -110,14 +121,21 @@ is_calculated <- function(x) {
   if (is.atomic(x)) {
     FALSE
   } else if (is.symbol(x)) {
-    is_dotted_var(as.character(x))
+    res <- is_dotted_var(as.character(x))
+    if (res && warn) {
+      warn_old_calculated_aes()
+    }
+    res
   } else if (is_quosure(x)) {
-    is_calculated(quo_get_expr(x))
+    is_calculated(quo_get_expr(x), warn = warn)
   } else if (is.call(x)) {
     if (identical(x[[1]], quote(stat))) {
+      if (warn) {
+        warn_old_calculated_aes()
+      }
       TRUE
     } else {
-      any(vapply(x, is_calculated, logical(1)))
+      any(vapply(x, is_calculated, warn = warn, logical(1)))
     }
   } else if (is.pairlist(x)) {
     FALSE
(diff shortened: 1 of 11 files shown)


yutannihilation:
> why you are not using the `lifecycle::deprecate_soft()` functionality

The main reason is that I just copied the code from here. Actually, I'm not sure when to use `lifecycle::*` functions and when to use `cli::*` functions... But, in this case, I think "`lifecycle::deprecate_soft()`" is not suitable because `..var..` and `stat(var)` is not deprecated (`NEWS.md` says they are "superseded").

https://github.com/tidyverse/ggplot2/blob/7ac55dd359a391ec15682724f5a6997e6b57588a/R/layer.r#L132-L136

thomasp85:
Ah, I see... The reason why we don't use it in the example above is that the auto-generation of text doesn't really capture what is happening due to the special syntax of ggplot2

In general we will always use the lifecycle functions for deprecations. If they have been superseded I don't think we should issue any warning at all since that means they are still proper syntax

I guess the question then is: Do we want to formally deprecate them? If so we should use lifecycle

yutannihilation:
I see, thanks for the explanation!

> If they have been superseded I don't think we should issue any warning at all since that means they are still proper syntax

> I guess the question then is: Do we want to formally deprecate them? If so we should use lifecycle

Makes sense. In my opinion, we should deprecate them then. A superseded function or syntax is preserved because it might not be very easy for all users to replace it with the superseder one. But, in this case, all `after_stat(var)`, `..var..`, and `stat(var)` refer to the same thing with different names. I'll wait for Hadley's comment.

hadley:
I think it's probably ok to deprecate them; the warning is only once every 8 hours so it's not terribly annoying.

Note that in dev lifecycle you can use `I()` in `what` and `reason` to generate your own text for this sort of situation. lifecycle will be going to CRAN soon, so I think it's ok to temporarily depend on dev lifecycle to get access to this functionality.

yutannihilation:
Thanks! Then, let's move forward.

But, I found `lifecycle::deprecate_soft()` doesn't fit in this case because both `..var..` and `stat(var)` are not functions that are called directly, but are just expressions evaluated non-standardly. My guess is that this was also the true reason why you chose `cli_inform()` in `linewidth`'s case. Am I correct?

Let me raise one more question. If the intention is to mimic the behaviour of `lifecycle::deprecate_soft()`, shouldn't we use `cli::cli_warn()` instead of `cli::cli_inform()`? (I mean both in this pull request and in the `linewidth`'s case)

hadley:
See https://lifecycle.r-lib.org/dev/articles/communicate.html#anything-else

I’d also recommend using `deprecate_warn()` as soft deprecation will only affect package developers (which I don’t think is what you want).

thomasp85:
@yutannihilation you can easily use lifecycle for `deprecate_warn("3.4.0", "stat()", "after_stat()")`... it's a bit more involved to use it for `..var..` so there we should rely on dev lifecycle

yutannihilation:
@hadley 
Thanks, you are right. This deprecaution targets the end users. It makes sense to skip `_soft()` stage.

@thomasp85 
I see, thanks. I'll use `I()` for the dot-dot notation.

hadley:
If you don't mind, could you also update the line size deprecation to use the new lifecycle feature?

yutannihilation:
Sure!

yutannihilation:
``` r
#> ℹ Loading ggplot2

p <- ggplot(data.frame(), aes(..x.., stat(y)))
b <- ggplot_build(p)
#> Warning: The dot-dot notation (`..var..`) was deprecated in ggplot2 3.4.0.
#> Please use `after_stat()` instead.
#> Warning: `stat()` was deprecated in ggplot2 3.4.0.
#> Please use `after_stat()` instead.
```

<sup>Created on 2022-08-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/aes-evaluation.r
```diff
@@ -93,31 +93,42 @@ is_dotted_var <- function(x) {
...
     return(TRUE)
   }
   # Support of old recursive behaviour
   if (is.atomic(x)) {
     FALSE
   } else if (is.symbol(x)) {
-    is_dotted_var(as.character(x))
+    res <- is_dotted_var(as.character(x))
+    if (res && warn) {
+      # TODO: if we want to set always = TRUE, we probably need some mechanism
```
hadley, line 115: I think you can remove this comment; we're still some time away from setting `always = TRUE`, if we ever do so. This is a very old feature, so I think the deprecation needs to be quite mild.

yutannihilation: Sure.

tests/testthat/test-geom-sf.R
```diff
@@ -99,18 +99,18 @@ test_that("geom_sf() removes rows containing missing aes", {
     colour = c("red", NA)
   )
 
-  p <- ggplot(pts) + geom_sf()
+  p <- ggplot(pts)
   expect_warning(
-    expect_identical(grob_xy_length(p + aes(size = size)), c(1L, 1L)),
+    expect_identical(grob_xy_length(p + geom_sf(aes(size = size))), c(1L, 1L)),
```
yutannihilation, line 104: I needed a tweak here because of this issue: https://github.com/tidyverse/ggplot2/issues/4960

R/aes-evaluation.r
```diff
@@ -93,31 +93,38 @@ is_dotted_var <- function(x) {
...
     return(TRUE)
   }
   # Support of old recursive behaviour
   if (is.atomic(x)) {
     FALSE
   } else if (is.symbol(x)) {
-    is_dotted_var(as.character(x))
+    res <- is_dotted_var(as.character(x))
+    if (res && warn) {
+      lifecycle::deprecate_warn("3.4.0", I("The dot-dot notation (`..var..`)"), "after_stat()")
```
hadley, line 115: Could we interpolate the variable name in to `..{{var}}..` and `after_stat({var}})`? (assuming `var <- as.character(x))`)

hadley (commented):
Could you please add a snapshot test that shows both new warnings?

R/aes-evaluation.r
yutannihilation: Do you mean this is supposed to work...?

``` r
var <- "foo"
lifecycle::deprecate_warn("3.4.0", "a()", "b({{var}})")
#> Error in `as_string()`:
#> ! Can't convert a call to a string.

#> Backtrace:
#>     ▆
#>  1. └─lifecycle::deprecate_warn("3.4.0", "a()", "b({{var}})")
#>  2.   ├─id %||% msg
#>  3.   └─lifecycle:::lifecycle_message(...)
#>  4.     └─lifecycle:::spec(with, NULL, signaller)
#>  5.       └─lifecycle:::spec_arg(what$call, signaller = signaller)
#>  6.         └─rlang::as_string(node_car(arg))
#>  7.           └─rlang:::abort_coercion(x, "a string")
#>  8.             └─rlang::abort(msg, call = call)
```

<sup>Created on 2022-08-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley: Oh sorry, you’ll need to glue/paste yourself.

hadley (approved):
Looks great!

R/aes-evaluation.r
yutannihilation: I see. Thanks for the suggestion anyway!

yutannihilation:
Thanks!

@thomasp85 
Do you have further comments?

thomasp85:
LGTM

yutannihilation:
Thanks!

PR #4956, opened 2022-08-23
rename size -> linewidth in element_rect and element_line

### LINKED ISSUE #4953: Should element_rect() and element_line() also use `linewidth` instead of `size`?
#3672 introduced `linewidth` to Geoms. On the other hand, `element_rect()` and `element_line()` still uses `size`. Do we need to rename them for consistency?

Currently, this doesn't work.

``` r
library(ggplot2)

ggplot(mpg, aes(x = hwy, y = cty, color = drv)) +
  geom_point() +
  theme(
    legend.key = element_rect(linewidth = 3, colour = "red")
  )
#> Error in element_rect(linewidth = 3, colour = "red"): unused argument (linewidth = 3)
```

<sup>Created on 2022-08-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85:
Fix #4953 

This fixes the oversight in the size to linewidth change in the element functions.

I believe I've fixed all instances of calls to the `element_*()` functions but please review carefully

### DIFF
--- a/R/theme-defaults.r
+++ b/R/theme-defaults.r
@@ -117,12 +117,12 @@ theme_grey <- function(base_size = 11, base_family = "",
     # Elements in this first block aren't used directly, but are inherited
     # by others
     line =               element_line(
-                           colour = "black", size = base_line_size,
+                           colour = "black", linewidth = base_line_size,
                            linetype = 1, lineend = "butt"
                          ),
     rect =               element_rect(
                            fill = "white", colour = "black",
-                           size = base_rect_size, linetype = 1
+                           linewidth = base_rect_size, linetype = 1
                          ),
     text =               element_text(
                             family = base_family, face = "plain",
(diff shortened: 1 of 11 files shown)


thomasp85:
scratch that - my search didn't include tests and vignettes... will fix

yutannihilation:
This one also needs to be tweaked, but, as far as my regex-fu can detect ([result](https://gist.github.com/yutannihilation/c29fac781a6286f62d0bd2047bc6a5a1)), it looks good to me so far! I'll wait for the further fix on tests and vignettes.

https://github.com/tidyverse/ggplot2/blob/c86ed7bc59060d94fb070fe1bb1574e8b6026ff8/R/theme-elements.r#L46

thomasp85:
okay @yutannihilation this seems ready for review now

yutannihilation (approved):
Looks good to me!

PR #4959, opened 2022-08-23
Use linewidth more

yutannihilation:
During reviewing #4956, I found several examples still use `size` where it should be `linewidth`. This pull request mainly fixes it as well as

* Added `sf::` prefix to the examples in `annotation_map()`. These don't fail just because they are in dontrun.
* Fixed `annotation_map()`. It seems it was broken.
* Add a note that `pointrange()` also understands `size` aesthetics in actual.


thomasp85 (approved):
Thank you

PR #4961, opened 2022-08-24
Don't use I() to format the layer call

thomasp85:
@Hadley am I right in remembering that you ultimately didn't want the full call in the error message out of fear that it would be too long?

### DIFF
--- a/R/plot-build.r
+++ b/R/plot-build.r
@@ -421,7 +421,7 @@ by_layer <- function(f, layers, data, step = NULL) {
       out[[i]] <- f(l = layers[[i]], d = data[[i]])
     },
     error = function(cnd) {
-      cli::cli_abort(c("Problem while {step}.", "i" = "Error occurred in the {ordinal(i)} layer."), call = I(layers[[i]]$constructor), parent = cnd)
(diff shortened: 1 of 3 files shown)


hadley (approved):
Yeah, that sounds right. Probably worth including at test in the general vicinity? (i.e. I don't think we need one explicitly to test that long calls are truncated, but I'm surprised that no tests changed.)

thomasp85:
I was quite surprised as well. Just looked into it and the `layer_grob()` helper circumvents the whole new layer error setup so there is no tests that captures that... will add one or two tests for that.

BTW, just to be clear, the removal of `I()` makes it so that all calls are truncated down to only the function name, not just long ones. This is what you prefer, right?

hadley:
Removing `I()` means that rlang is responsible for truncating — currently it uses a pretty simple heuristic of only display the function name (except for various infix/special functions) but it might do more in the future. I'm happy for ggplot2 to defer to rlang here.

thomasp85:
Interestingly enough, the first line of the error message is not recorded in the snapshot (the `error in ...` line) which explains why we didn't see any change... however I've also added some formal unit tests in `test-layer.R`

hadley:
Hmmm, maybe we should be using `expect_snapshot(error = TRUE)`?

thomasp85:
If we now agree that rlang takes care of the first line it is ok to not have any checks in ggplot2 I think?

Issue #4886, opened 2022-06-20
About the name and the default value of `rename_aes`


yutannihilation:
@thomasp85 
Sorry for bothering you. Re-reading #3672, I found I don't have clear answers to the following points. Can we discuss here?

1. Why is the default value of `rename_size` `FALSE`? I'm afraid it would feel a bit inconsistent that the extension packages that don't catch up with this change would behave differently, while we are very careful not to introduce breaking changes on ggplot2 itself by setting `rename_size = TRUE` everywhere necessary. What about
    a. using `rename_size = TRUE` by default for the first release, or
    b. making the default value of `rename_size` as an option or an edition? In this case, the user can choose the behavior they prefer, and the developer can use `rename_size = FALSE` (or `TRUE`) if they think it's inappropriate to expose it as an option.
3. Can we have another word than "rename"? It seems "rename" sounds like the original one is removed (c.f. https://github.com/tidyverse/ggplot2/pull/4884#issuecomment-1159863428), like dplyr's `rename()`, while `rename_size = TRUE` actually keeps the `size`.

thomasp85:
1. Because setting `rename_size = TRUE` will trigger a deprecation warning if size is used. I don't think anything is gained by doing this any softer - at some point there will have to be a switch and it might as well happen now when the change is in the minds of the extension developers. I think history has shown that our extension developers are quite quick to catch up. Using an edition sets a bad precedence for any breaking change to warrant that.

2. I'm not married to the name, but I do think it communicates the intend of the setting quite well, i.e. it opts in to the renaming. The fact that the old aesthetic is kept is an implementation detail of sorts

yutannihilation:
Thanks. For point 2, I got it. Sorry for the noise.

For point 1, I think you are right, but let me think for a while before closing this issue. It seems the part of my confusion was because I didn't see any deprecation warnings when playing with `geom_sf()` or `geom_pointrange()`.

clauswilke:
@thomasp85 @yutannihilation Do we have a guide for extension developers telling them what exactly they have to do to ready their packages for this change?

yutannihilation:
I guess this is the guide, and probably this will go into the release blog post?

https://twitter.com/thomasp85/status/1537024892293599235

clauswilke:
[...]

PR #4964, opened 2022-08-25
Remove rename_size = TRUE for sf and pointrange geoms

thomasp85:
This PR partly reverses #4884 as the omission of `rename_size = TRUE` in geom_pointrange()` and `geom_sf()` was intentional since `size` remains a valid aesthetic.

### DIFF
--- a/R/geom-pointrange.r
+++ b/R/geom-pointrange.r
@@ -57,7 +57,5 @@ GeomPointrange <- ggproto("GeomPointrange", Geom,
         GeomPoint$draw_panel(transform(data, size = size * fatten), panel_params, coord)
       ))
     )
-  },
-
-  rename_size = TRUE
+  }
 )
--- a/R/geom-sf.R
+++ b/R/geom-sf.R
@@ -149,9 +149,7 @@ GeomSf <- ggproto("GeomSf", Geom,
     } else {
       draw_key_polygon(data, params, size)
     }
-  },
-
-  rename_size = TRUE
+  }
 )
 
 default_aesthetics <- function(type) {


yutannihilation:
Hmm, sorry, I think I'm still in confusion. Could you explain what's our strategy again?

I still feel `rename_size` is needed for these Geoms to avoid introducing any visual changes to existing code using `size`. The authors of such code can recognize the warning when they execute the same code with the next version of ggplot2 so that they can update their code. Removing `rename_size` means the plots are silently changed with no warnings. The downside of `rename_size = TRUE` is, however, users cannot specify the point size only; `size` will affect both point size and line width so they need to specify `linewidth` as well just to restore the line width to the default value.

So, in my understanding this is a trade off between preserving the old code and making the new code less verbose, and I rather support the former.

thomasp85:
Yes - this is a breaking change for these two geoms... I believe it is worth it - size and linewidth has to be decoupled at some point and I'd rather make it now where the introduction of linewidth is in fresh memory instead of later on

yutannihilation:
Hmm, let me review the past discussion. I think that's okay, but I don't remember if we agreed on that direction. Sorry for keeping you waiting.

thomasp85:
tagging in @hadley, just for completeness

hadley (approved):

yutannihilation:
As this got approved, please feel free to merge if you are in hurry.

yutannihilation:
@thomasp85 
Just for the record. Looking at https://github.com/tidyverse/ggplot2/issues/4883, which made us add `rename_size = TRUE` to these geoms, it seems you wrote

> At this point the ambiguous geoms will not differentiate but I think we would do that down the line after proper deprecation

in [the original PR](https://github.com/tidyverse/ggplot2/pull/3672#issuecomment-1154713022). So, I think this part of the description of this PR is not quite correct. I bet that **wasn't** intentional, considering we agreed to postpone the braking change "after proper deprecation".

> the omission of rename_size = TRUE in geom_pointrange() and geom_sf() was intentional 

After all, I support your decision because [the FAQ section of your blog post](https://www.tidyverse.org/blog/2022/08/ggplot2-3-4-0-size-to-linewidth/) was convincing, and I agree that we should do it now before forgetting the plan. But, I'd appreciate if you respect the past discussions a bit more...

Anyway, thanks for your hard work!

yutannihilation:
I apologize that [my comment above](https://github.com/tidyverse/ggplot2/pull/4964#issuecomment-1226854921) was wrong. I wrote

> The authors of such code can recognize the warning when they execute the same code with the next version of ggplot2

but, in reality, they never see warnings because `size` is a valid aesthetic. So, probably it wouldn't be that effective if we keep `rename_size = TRUE` for a while (It seems I already noticed it in https://github.com/tidyverse/ggplot2/issues/4886#issuecomment-1160481545). Sorry for the noise...

yutannihilation:
No worries. Changing course itself is not a problem. I just wanted more details to make sure this is the right move because this topic has been too complex for me to understand quickly.

smouksassi:
keep in mind geom_boxplot has fatten as well similar to crossbar
https://github.com/tidyverse/ggplot2/blob/ae446507437c5062ccaf8097de66b0b3f259da47/R/geom-boxplot.r#L280
not sure if we will be able to control both independently?

thomasp85:
@trevorld that is a good question. boxplot is a tricky one because size have been used for outlier sizing unless `outlier.size` is given and since it is given by default it is seldom used that way... So, `size` has predominantly been used to scale the stroke and it makes sense to convert it, but we can't remove size from `default_aes` since it remain a valid aesthetic...

`fatten` in boxplot/crossbar is orthogonal to all this and will remain in effect. `fatten` in pointrange becomes redundant with the switch to linewidth and will be deprecated with time

trevorld:
> boxplot is a tricky one because size have been used for outlier sizing unless outlier.size is given

Actually `size` only affects outlier size only if the user explicitly passes `outlier.size = NULL`.  `size` provides a backup if `outlier.size` is `NULL` but since `outlier.size` has a numeric default it is never `NULL` unless the user explicitly passes it a `NULL`.  IMO you should probably just remove `size` from `default_aes`.  I don't think many users are going to do `linewidth = 4, size = 4, outlier.size = NULL` instead of `linewidth = 4, outlier.size = 4`.

```r
p <- ggplot(mpg, aes(class, hwy))
p + geom_boxplot()
p + geom_boxplot(size=4) # outliers the same size as `size = 1`, `size = 2`, `size = 3` since we are using `outlier.size` default
p + geom_boxplot(size=4, outlier.size = NULL) # now outliers are bigger

```

PR #5001, opened 2022-09-29
Use theta-specific theme elements

willgearty:
The tick text around the polar plot previously always used the `axis.text.x` theme element, but if you change `theta` to `y`, you'll actually want to use the `axis.text.y` theme element.


willgearty:
I think a better/more intuitive way to handle this would be by having r and theta theme elements, so I'm closing this for now.

PR #5007, opened 2022-10-05
Placement of text elements on the plot

JobNmadu:
The placement of the text in the plot is slightly adjusted. The current placement is not very exact. c("bottom-left", "top-left", "bottom-right", "top-right", "center")


teunbrand (approved):
Thanks for catching this mistake!

Adding a little bit of context: this PR fixes an example in `geom_text()` where there is a mismatch between the text and their position.



Probably the checks failed because the CI was in a bad streak, and I can't rerun them now because they're too old. I've run the check with the proposed changes locally, and they run just fine. @thomasp85 could I merge this PR?

PR #5018, opened 2022-10-21
transparent theme

topepo:
I know... yet another theme. This one is best for presentations with non-standard background colors.


clauswilke:
For the various backgrounds, is there a reason you use rectangles with fill color transparent rather than `element_blank()`? It seems to me that removing those elements entirely would be slightly cleaner.

thomasp85 (commented):
@teunbrand can you look into updating this to the newest theme style (using ink and paper etc)

teunbrand:
This is now updated to latest theme changes.
I've moved this theme away from using fixed values like `"grey20"`.
Instead, it does a neat trick where it blends `ink` and `paper` like all the other themes, but the trick is that `paper = alpha(ink, 0)` thus mixing between a solid and transparent colour.

In the example below, that causes the gridlines to always be darker over the background and not lighter over the red parts.

``` r
devtools::load_all("~/packages/main/ggplot2/")
#> ℹ Loading ggplot2

  geom_point() +
  theme_transparent() +
  facet_wrap(~"Strip title")

# Make rectangle pattern
rectangles <- rectGrob(
  unit(c(0.25, 0.75), "cm"), unit(c(0.25, 0.75), "cm"), 
  width = unit(0.5, "cm"), height = unit(0.5, "cm"), 
  gp = gpar(fill = "red", col = NA)
)
pattern <- pattern(
  rectangles, 
  x = unit(0.5, "cm"), y = unit(0.5, "cm"),
  width = unit(1, "cm"), height = unit(1, "cm"), 
  extend = "repeat"
)

grid.newpage()
grid.draw(rectGrob(gp = gpar(fill = pattern, col = NA)))
grid.draw(ggplotGrob(p + theme_transparent()))
```



<sup>Created on 2024-11-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

topepo:
Thanks for getting this over the finish line!

thomasp85 (approved):
LGTM

teunbrand:
Thanks Max and Thomas!

Issue #4462, opened 2021-05-04
Add "end" parameter to coord_polar to allow non-complete polar plots


JohannesNE:
It would be useful to allow non-complete polar plots. E.g. half or quarter plots. This could be parameterized with `coord_polar(start = -pi/2, end = pi/2, direction = -1)` and `coord_polar(start = -pi/2, end = 0, direction -1)` respectively.

This would also allow a more intuitive placement of the radial axis.

Examples:
Half: 
![](https://i.stack.imgur.com/BwBG7.png)
Quarter:
![](https://i.stack.imgur.com/PDtWW.png)

PR #5028, opened 2022-11-03
Partial polar plots

teunbrand:
This is my take on #4462, that adds an `end` argument to `coord_polar()` to allow partial polar plots.

In brief:

* There are no visual changes if `end` is not used.
* It was straightforward to rescale to [start, end] instead of [0, 2&pi;] to get the partial polar coordinates.
* After that rescaling, I had to account for the bounding box of the circle sector.
* The bit I found more difficult were the radius axis placement. I've resolved this as detailed below.

Normally, the axis is placed in the upper part of the y-axis. If we have only the lower part of a circle, I had to reflect the break positions:

``` r
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(displ, hwy, colour = factor(cyl))) +
  geom_point()

p + coord_polar(start = 0.5 * pi, end = 1.5 * pi)
```

If 0 or &pi; are not in the [start, end] interval, I have placed the axis in the x-direction. (Below also reflected because left instead of right).

``` r
p + coord_polar(start = 1.25 * pi, end = 1.75 * pi)
```

If no integer multiple of 0.5&pi; is in that interval, I can't appropriately place an axis anywhere. In that case, I've thrown a message with instructions.

``` r
p + coord_polar(start = 0.1 * pi, end = 0.4 * pi)
#> Could not find appropriate placement for the radius axis.
#> ℹ A radius axis requires the [`start`-`end`] range to include one of: `c(0,
#>   0.5, 1, 1.5) * pi`.
```

<sup>Created on 2022-11-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

After the guides are ported over to the ggproto system, we might consider a more complete solution for axis placement than this.


thomasp85:
Is this still meaningful to get in or should we tackle it from anew after ggproto guides have been merged?

teunbrand:
I think the new ggproto guides don't interact with `coord_polar()` other than making `coord_polar()` skip doing proper guide stuff (as it wasn't using this), so these changes can safely co-exist. After the ggproto guides are merged, we can tackle bringing `coord_polar()` into the fold of guide-using coords, either in this PR or a new one. However, I'm pretty keen on the functionality that this PR adds.

thomasp85:
ok, I'll do a review next week

teunbrand:
Closing this PR in favour of #5334.

PR #5030, opened 2022-11-05
Improvements to `geom_label()`

### LINKED ISSUE #2785: geom_label() can't rotate text & gets descent heights wrong
I think `geom_label()` could use an overhaul. It has two significant limitations:

1. Labels can't be rotated (see e.g. https://github.com/tidyverse/ggplot2/pull/2782 and example)
2. Descent heights are ignored (see example)

``` r
library(ggplot2)
library(tibble)

data <- tibble(
  text = c('abc', 'def', 'gjq'),
  angle = c(0, 45, 90),
  x = c(1, 2, 3)
)

ggplot(data, aes(x, label = text, angle = angle)) +
  geom_label(aes(y = 1), label.padding = unit(0, "lines"), size = 10) +
  geom_label(aes(y = 2), size = 10) + xlim(0.5, 3.5) + ylim(0.5, 2.5)
```

Created on 2018-07-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

I have [proof-of-concept code](https://github.com/clauswilke/gridtext/blob/master/R/labels-grob.R) that addresses both issues:
``` r
library(gridtext)
library(grid)
library(tibble)

# example with rotated labels, descenders, and padding
label_data <- tibble(
  label = c("Descenders: pgqjy", "This is a label\nwith two lines", "Hello!"),
  x = unit(.4, "npc"),
  y = unit(c(.9, .5, .3), "npc"),
  box_hjust = 0,
  box_vjust = 0.5,
  hjust = 1,
  vjust = 1,
  color = "blue",
  fill = "azure1",
  fontsize = 10,
  fontfamily = "Comic Sans MS",
  angle = c(0, 45, -45),
  padding = list(mar(5, 5, 3, 5)),
  margin = list(mar(5, 5, 5, 5))
)

grid.newpage()
g <- labels_grob(label_data)
grid.draw(g)
```

``` r

# example with no padding and label frames all made the same size
label_data <- tibble(
  label = c("These", "are", "multiple", "labels", "gqpj"),
  x = unit(c(.1, .3, .5, .7, .9), "npc"),
  y = unit(0.5, "npc"),
  box_hjust = 0.5,
  box_vjust = 1,
  hjust = 0.5,
  vjust = 1,
  color = "black",
  angle = 0,
  fontsize = 10,
  margin = list(mar(5, 5, 5, 5))
)
grid.newpage()
g <- labels_grob(label_data, align_frames = TRUE)
grid.draw(g)
```

Created on 2018-07-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

Fundamentally, the problem is caused by very limited text handling capabilities directly in grid, so it's a fair amount of work to write code that does these things. This raises the issue of whether we should address these problems one-by-one in ggplot2 where they arise or rather write a better text-rendering library and then use that from ggplot2.

### LINKED ISSUE #4753: text rendered incorrrectly when size and justification parameters are supplied to geom_label()
I posted about this in [the RStudio Discourse forum](https://community.rstudio.com/t/strange-effect-of-justification-parameters-on-geom-label-rendering/131047?u=dromano), but didn't explicitly point out the effect of supplying the `size` parameter.  Have received no responses so am posting here.

The problem appears to be an interaction between the `size` parameter of `geom_label()` and how the justification parameter values are applied to the text itself, but is not apparent unless off-center justification is supplied.

``` r
library(tidyverse)

tibble() %>% 
  ggplot() +
  geom_label(aes(0, 0, label = 'text'), 
             hjust = 0,
             size = 20,
             )
```

<sup>Created on 2022-03-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

teunbrand:
This PR aims to fix #2785 and fix #4753.

Briefly:

* Viewports of individual labels rotate according to the `angle` aesthetic.
* Textboxes now take `grobDescent(text_grob)` into account.
* The viewport sets a `fontsize` parameter, that allows the typical `unit(0.25, "lines")` label padding to get appropriate text sizes.

The 2nd and 3rd points lead to visual changes, but I'd argue that these are appropriate.

The reprex provided in https://github.com/tidyverse/ggplot2/issues/2785#issue-345452842

``` r
#> ℹ Loading ggplot2

data <- data.frame(
  text = c('abc', 'def', 'gjq'),
  angle = c(0, 45, 90),
  x = c(1, 2, 3)
)

ggplot(data, aes(x, label = text, angle = angle)) +
  geom_label(aes(y = 1), label.padding = unit(0, "lines"), size = 10) +
  geom_label(aes(y = 2), size = 10) + 
  xlim(0.5, 3.5) + 
  ylim(0.5, 2.5)
```

The reprex from https://github.com/tidyverse/ggplot2/issues/4753#issuecomment-1066970150

``` r

ggplot() +
  annotate(
    "label",
    x = 0, 
    y = 0, 
    label = "text",
    hjust = c(0, 1),
    size = 20
  )
```

<sup>Created on 2022-11-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


teunbrand:
Hi @clauswilke, could I ask you to review this PR whenever you could find the time? I think you have the most experience with textboxes, so I'd welcome your suggestions.

clauswilke:
I think the problem you'll have is that if you have one label `abc` and one label `gqj` and place them side by side they will have different heights. You have to decide whether that's correct or not. It's probably better than the current situation.

I've always felt that text at a given font size should have a defined ascent and descent regardless of content (and in fact, that's how ascent and descent are defined in typography), but grid doesn't work that way if I recall correctly, and as your examples show.

clauswilke:
Notice how in a web browser these two labels have the same height: `ace` `Mqj`

Screenshot:
Screen Shot 2023-01-14 at 11 20 13 PM

teunbrand:
Thanks that is a keen observation I hadn't thought of myself, so that is useful! I guess I could use `ggplot2:::font_descent()` instead to calculate the descent. Do you think it is worth it to make a special case for text that doesn't have descenders in any of the labels?

thomasp85:
Just to add my own opinion, we should def use the same descent irrespective of the content of the text box, and `font_descent()` is the way to get it. I think `titleGrob()` already do a lot of the same work under the hood.

Not sure if supporting `margin()` in the padding is a feature that is worth the added complexity though, but no strong opinions there

teunbrand:
Thanks for the thoughts. I'll explore around a bit to see to what degree `titleGrob()`'s logic can be re-used here, in which case the added complexity for `margin()` might be ... marginal ...

teunbrand:
Some updates:
* Using `titleGrob()` for the labels simplifies the label drawing, as it also supplies appropriate grob widths/heights for the box.
* This made it also easy to implement `label.padding = margin(...)` input.
* I removed the `makeContent` method, because I don't think we need it. I've looked at the history of it, but I couldn't figure out what exactly should be calculated at grob drawing time. If there is a good reason for it, I can put it back, but if there isn't, we've reduced the complexity a bit.

Showing the `margin` argument and descent calculations:

``` r
#> ℹ Loading ggplot2

data <- data.frame(
  text = c('abc', 'def', 'gjq'),
  angle = c(0, 45, 90),
  x = c(1, 2, 3)
)

ggplot(data, aes(x, label = text, angle = angle)) +
  geom_label(aes(y = 1), label.padding = unit(0, "lines"), size = 10) +
  geom_label(
    aes(y = 2), size = 10,
    label.padding = margin(t = 20, r = 0, b = 5, l = 5)
  ) + 
  xlim(0.5, 3.5) + 
  ylim(0.5, 2.5)
```



<sup>Created on 2023-01-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/geom-label.R
```diff
@@ -121,32 +122,41 @@ labelGrob <- function(label, x = unit(0.5, "npc"), y = unit(0.5, "npc"),
...
+  box <- roundrectGrob(
+    x = x, y = y - 0.5 * descent,
+    width  = grobWidth(text),
+    height = grobHeight(text),
+    just   = c(hjust, vjust),
+    r = r, gp = rect.gp, name = "box"
   )
 
-  setChildren(x, gList(r, t))
+  gTree(children = gList(box, text), name = name, vp = vp)
```
thomasp85, line 161: Do you have a reason for this change. AFAIK, the `makeContent()` method expects you to return a modified version of the input, not a new object (i.e. use `setChildren()`). This is because grid keeps track of the actual objects under the hood

teunbrand: Yeah the reason is that we don't need `makeContent()` here, so I've included all necessary steps in the `labelGrob()` constructor. This simplifies the grob a lot, and the `makeContent()` method wasn't doing any calculations that rely on the device size or resolution.

thomasp85: Ah - I got confused by the diff... this is no longer a `makeContent` method

thomasp85 (approved):

thomasp85:
have you tested the impact on performance? Not that `geom_label()` was ever that performant.

When this is merged it would be good to at least inform ggrepel so they can test whether the new label grobs affect their package

R/geom-label.R
teunbrand: Yep that's right :)

teunbrand:
> have you tested the impact on performance?

It is good that you asked, because I just did, and printing a plot with 100 labels takes 2.5x longer with this PR (695ms vs 268ms).  Most of this extra time is spend doing viewport operations. I think I should spend some time trying to make this more efficient.

> inform ggrepel 

I will, but I ran their unit tests and they all pass with this PR. Plotting `geom_label_repel()` doesn't use `labelGrob()`, so this makes sense to me.

thomasp85:
I don't think `titleGrob()` has ever been subjected to a high degree of performance scrutiny since a plot only ever consist of a few titles. Seeing if this could be improved would be worthwhile

clauswilke:
If I remember correctly from writing gridtext, I felt that titleGrob was overly complex and convoluted. You don't need a second viewport just to add margins around text. You can just make a bigger box and put the text in the right location so there are margins. So there may be efficiency gains from a refactoring.

https://github.com/tidyverse/ggplot2/blob/f139f4019b2c9ce9cce55f81280bfdd50fd4bb6d/R/margins.R#L107-L156

teunbrand:
Yeah I did some testing and a lot of the performance loss is by using titleGrob and not necessarily the viewport that I use here to rotate the labels. Perhaps it makes more sense to put in another PR where we refactor `titleGrob()` instead of trying to optimise this implementation of `geom_label()`.

teunbrand:
Performance hit is now only 26% slower (100 labels, 344ms vs 273ms), which is an improvement over the 160% slower we had earlier. I find this an acceptable trade-off for the features it offers, so I'll merge this.

PR #5032, opened 2022-11-06
Extend palette contract: Palette capabilities

zeehio:
The goal of this pull request is to improve the performance of checking for missing values when mapping values to colours using a palette.

Some palettes may be able to guarantee that they never return missing values as colours. For those palettes it does not make sense that we check if there are missing values in their output.

In this pull request, I use (although I don't depend on it):
- https://github.com/r-lib/scales/pull/372

Basically, that pull request extends the palette contract by adding attributes to it. One of those attributes is `may_return_NA` which, if set to `FALSE`, then ggplot2 can assume the output of the palette has no missing values and skip the corresponding check.

The relevant part of this pull request is:

```r
# A specific palette can have as attribute "may_return_NA = FALSE"
# If it has such attribute, we will skip the ifelse(!is.na(scaled), ...)
pal_may_return_na <- ggproto_attr(self$palette, "may_return_NA", default = TRUE)
if (pal_may_return_na) {
  scaled <- ifelse(!is.na(scaled), scaled, self$na.value)
}
```

Related to:
- #4989

### DIFF
--- a/R/ggproto.r
+++ b/R/ggproto.r
@@ -351,3 +351,7 @@ format.ggproto_method <- function(x, ...) {
 
 # proto2 TODO: better way of getting formals for self$draw
 ggproto_formals <- function(x) formals(environment(x)$f)
+
+ggproto_attr <- function(x, which, default = NULL) {
+  attr(environment(x)$f, which = which, exact = TRUE) %||% default
+}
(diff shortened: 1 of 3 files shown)


teunbrand:
Closing this PR in favour of #6344. Thank you for exploring this problem, but I feel `anyNA()` is a more general check than relying on function attributes.

PR #5039, opened 2022-11-08
Correct spelling of 'aesthetic' in `check_linewidth()`

MattCowgill:
The word 'aesthetic' is misspelled in the deprecation message issued by `check_linewidth()`. This PR fixes the typo.

PR #5061, opened 2022-11-25
mini typo

heike:


yutannihilation:
Thanks for catching! But, there's already a pull request for fixing this: #5039

Issue #5054, opened 2022-11-17
 ggplot2::theme arguments are not updated starting from  3.3.2 version 


Polkas:
In the past all useable theme arguments were part of the ggplot2::theme arguments.
Starting from 3.3.2 there are new arguments strip.text.x.top, strip.text.x.bottom, strip.text.y.left, strip.text.y.right which are not included in the ggplot2::theme arguments.
This is not a problem as ggplot2::theme contains ..., but it is not consistent.
In my cases I use ` methods::formalArgs(ggplot2::theme)` to get possible arguments and validate my own function but it stop to work properly now. 
It seems fair as you are using ..., but why to list the arguments there at all. 
Now you are listing but 99% of all arguments.
I understand this was a form of a better documentation for you and the end users.
Thank you for your support.

Polkas:
More than that the new strip.text.x.top, strip.text.x.bottom, strip.text.y.left, strip.text.y.right arguments are not documented in the ggplot2::theme.

Polkas:
What do you think about this problem? I could create a PR and update ggplot2::theme.

Polkas:
I just create the appropriate PR for this task. I will be glad to get a response.

PR #5062, opened 2022-11-27
theme docs - strip.text.x.* family

Polkas:
closes https://github.com/tidyverse/ggplot2/issues/5054

- The new (>=3.3.2 version) strip.text.x.top, strip.text.x.bottom, strip.text.y.left, strip.text.y.right arguments are documented in the ggplot2::theme now.
- I added the additional example for that new feature.
- There was a surprising small difference in the one testthat snap file, so I had to update it (please check out the svg file diff).


yutannihilation:
Thanks for catching! Looks good, but could you exclude the update on the snapshot? I think it's unnecessary.

Polkas:
Please check out that after revert to the old snap svg file the test are failing, for some systems.
I will revert the revert:)

yutannihilation:
Hmm, so does it mean this change actually affects the behaviour of ggplot2? Sorry, then the things are a bit complex than I expected. We have to figure out what's happening here...

Polkas:
I will investigate it. Summary for current situation and the source of the test error.

https://github.com/tidyverse/ggplot2/blob/fc191a07accd55bbc0b8c3d34a82f3992a3e4e53/tests/testthat/test-facet-strips.r#L164-L168

The difference in svg is the same in two places `translate(46.59,403.41)` -> `translate(40.54,403.41)` in the text tag.

Polkas:
So I fixed another problem as new theme arguments was expected but not taken into account.
https://github.com/tidyverse/ggplot2/blob/fc191a07accd55bbc0b8c3d34a82f3992a3e4e53/R/facet-grid-.r#L321

from that place we ended in `ggplot2:::build_strip` and there we look for theme arguments which not exists in theme before, but of course now they are.

https://github.com/tidyverse/ggplot2/blob/63125db184b6e7d5516007574b5d15acbfd8f470/R/labeller.r#L503-L534

Polkas:
More precious proof.

```r
debug(ggplot2:::calc_element)
p <- ggplot(mtcars, aes(disp, drat)) + geom_point()
p + facet_grid(am ~ cyl, switch = "both")
# in debug mode
theme[["strip.text.y.left"]]
# is not null but a list , previously (before PR) it was
```

teunbrand:
The suspicious thing about the visual test, is that the text is rotated by -90 degrees instead of typical 90 (only for `theme_test()`).
After doing some digging, I found the difference comes from `ggplot2:::plot_theme()`. Using CRAN ggplot2:

``` r
library(ggplot2) # CRAN version

p <- ggplot(mtcars, aes(disp, drat)) + geom_point() + theme_test() + 
  facet_grid(am ~ cyl, switch = "both")
b <- ggplot_build(p)
old <- calc_element("strip.text.y.left", ggplot2:::plot_theme(b$plot))
```

Likewise with a small edit to mimic this PR:

``` r
devtools::load_all("~/packages/ggplot2/") # mimicks this PR
#> ℹ Loading ggplot2

p <- ggplot(mtcars, aes(disp, drat)) + geom_point() + theme_test() + 
  facet_grid(am ~ cyl, switch = "both")
b <- ggplot_build(p)
new <- calc_element("strip.text.y.left", ggplot2:::plot_theme(b$plot))
```

The difference is in the angles:

``` r
waldo::compare(old, new)
#> `old$angle`:  90
#> `new$angle`: -90
```

<sup>Created on 2022-11-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

In particular, these lines give different results, as `strip.text.y.left` is now no longer considered 'missing', so it doesn't get updated with a reasonable default (angle = 90).

https://github.com/tidyverse/ggplot2/blob/63125db184b6e7d5516007574b5d15acbfd8f470/R/theme.r#L444-L445

The simplest 'fix' would be to add `strip.text.y.left = element_text(angle = 90)` to `theme_test()`, as all other themes (expect for `theme_void()`) inherit from `theme_gray()` in some way, which has a sensible `strip.text.y.left` element.

yutannihilation:
Wow, thanks you two so much for investigating! The fix sounds good to me. I'll try to understated the problem.

yutannihilation:
> The simplest 'fix' would be to add `strip.text.y.left = element_text(angle = 90)` to `theme_test()`, as all other themes (except for `theme_void()`) inherit from `theme_gray()` in some way, which has a sensible `strip.text.y.left` element.

Sorry for the delay. I think this fix is good enough. `strip.text.y.left` was added to `theme_gray()` on https://github.com/tidyverse/ggplot2/pull/3683/files, but not to `theme_test()`. I think this was simply an oversight. I'm not sure if there can be some better structure for testing, but let's not change today.

@Polkas 
Could you add `strip.text.y.left = element_text(angle = 90)` after this line? It should fix the test.

https://github.com/tidyverse/ggplot2/blob/5ebe979d4e7c9333c36cda7d4104b9f8160c29b2/R/theme-defaults.r#L621

Polkas:
@yutannihilation, thanks for a clear recommendation. I just applied it, and the _snaps for a few plots have to be updated.

yutannihilation:
Thanks, but I think the change is not what @teunbrand and I recommended...? The suggestion was to add `strip.text.y.left`, not to modify `strip.text.y`. Sorry, I think I don't get your intention here.

Polkas:
I understand that `strip.text.y.left` and `strip.text.y.right` inherit from `strip.text.y`, this was the source of the confusion. Sorry It was late, I see it now clear we want to be direct here with strip.

Ps. I learn a new thing that git can overwrite chmod of the file, I had to fix it manually for one file here with `git update-index --chmod=+w tests/testthat/_snaps/geom-dotplot/stackratio-1-5.svg`

R/theme-defaults.r
```diff
@@ -618,7 +618,8 @@ theme_test <- function(base_size = 11, base_family = "",
                            margin = margin(0.8 * half_line, 0.8 * half_line, 0.8 * half_line, 0.8 * half_line)
                          ),
     strip.text.x =       NULL,
-    strip.text.y =       element_text(angle = -90),
+    strip.text.y = element_text(angle = -90),
+    strip.text.y.left = element_text(angle = 90),
```
yutannihilation, line 622: Thanks, almost perfect! Sorry for nitpicking, could you align these `element_text()` to the other lines?

Polkas: I understand the need for consistency; I just updated it.

yutannihilation:
I guess you mistakenly committed `tests/testthat/_snaps/theme/`?

Polkas:
> I guess you mistakenly committed `tests/testthat/_snaps/theme/`?



> I guess you mistakenly committed `tests/testthat/_snaps/theme/`?

I created another issue connected with the arm64 https://github.com/tidyverse/ggplot2/issues/5110. I will remove these files.

yutannihilation:
Oh, curious. Anyway, this pull request looks perfect now. Thanks!

yutannihilation (approved):

PR #5072, opened 2022-12-01
Make `lwd` an alias for `linewidth`

### LINKED ISSUE #5051: Using `lwd` parameter in geom_line() produces `size` warning
I recently updated ggplot and found I am getting the "Using `size` aesthetic for lines was deprecated in ggplot2 3.4.0." warning for my lines of code where I use the `lwd` parameter. Is `lwd` also being deprecated? I think a more clear/informative warning would be helpful, because it took me a while to find the source since I didn't explicitly use `size` anywhere. 

Simplified example that produces the warning:

```r
library(ggplot2)

   geom_line(lwd = 0.25)

```

teunbrand:
This PR attempts to fix #5051 based on @clauswilke's suggestion.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* The `lwd` alias now correctly replaced by `linewidth` instead of `size` 
+  (@teunbrand based on @clauswilke's suggestion #5051).
+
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good, thanks!

Issue #5066, opened 2022-11-29
scale_fill_binned throws error if all bins have the same value


sebneus:
I want to print a couple of ggplots in a group_walk loop. But one of the plots only returns values in one bin which resolves in an error. 

Here is an example:
``` r
library(ggplot2)

data.frame(x = c("a", "b"), 
           y = c("A", "B")) |> 
  ggplot() + 
  geom_bin_2d(aes(x = x, y = y)) + 
  scale_fill_binned()
#> Error in cut.default(x, breaks, labels = FALSE, include.lowest = TRUE, : invalid number of intervals
```

<sup>Created on 2022-11-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

If there is a bin with a at least two different numbers (here 1 _and_ 2), the plot succeeds:

``` r
library(ggplot2)

data.frame(x = c("a", "b", "b"), 
           y = c("A", "B", "B")) |> 
  ggplot() + 
  geom_bin_2d(aes(x = x, y = y)) + 
  scale_fill_binned()
```

![](https://i.imgur.com/AGNYzSS.png)

<sup>Created on 2022-11-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

If I remove the `scale_fill_binned()` part the chart is drawn, but with no color steps. Similar (but different) error is thrown if I use `guides(fill = "colorsteps")`. Using `geom_tile(aes(x, y, fill = after_stat(count)), stat = "bin2d")` shows the same behaviour.

Regards
SebNeu

jan-glx:
Unfortuneatly @teunbrand 's fix does not quite `cut` 😛 it for `scale_x_binned`. See also this example:



``` r
library(ggplot2)
ggplot(data.frame(x=1), aes(x=x)) + geom_bar() + scale_x_binned(limits = c(0,1))
```

![](https://i.imgur.com/UUZ4l4u.png)

``` r
ggplot(data.frame(x=1), aes(x=x)) + geom_bar() + scale_x_binned()
#> Error in cut.default(x, all_breaks, labels = FALSE, include.lowest = TRUE, : invalid number of intervals
```

![](https://i.imgur.com/peVtEQ9.png)

<sup>Created on 2023-08-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
[...]

PR #5074, opened 2022-12-01
zero-range limits in binned scale

teunbrand:
This PR is intended as a *partial* fix to #5066. In brief, the `map()` method circumvents `cut()` when there aren't enough breaks for that function. It resolves the error in #5066, but does *not* resolve the `guide_coloursteps()` issue (because that guide is in reconstruction). In choosing a palette-value for zero-range limits, it now behaves similar to `scale_fill_continuous()`.

``` r
#> ℹ Loading ggplot2
library(patchwork)

a <- data.frame(x = c("a", "b"), 
           y = c("A", "B")) |> 
  ggplot() + 
  geom_bin_2d(aes(x = x, y = y))

b <- a + scale_fill_binned()

a + b
```

<sup>Created on 2022-12-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


thomasp85 (approved):
LGTM

PR #5078, opened 2022-12-03
Reign in overly ambitious `arrow()` in `geom_path()`

### LINKED ISSUE #4935: aes(color) causes arrow() to become overly enthusiastic
In this repex only one arrow should be drawn at (x = 1, y = 1). However, it appears something in the drawing stack gets confused by the color aesthetic and starts putting an arrow on every segment.
```
data = tibble(x = c(0, 0.5, 1), y = x, color = x)
ggplot(data) +
  geom_path(aes(x = x, y = y, color = color), arrow = arrow(ends = "last"))
```
This happens with `geom_line()` as well. In real world use the results I get are poor―even in the small hello world I was doing when I first encountered this issue 1800 extra arrowheads get drawn, which turns out to be both impressively slow to render (AGG backend with alpha) and almost unreadably cluttered.

@hadley closed #2280 because this type of intermediate arrow drawing was too difficult to implement but, apparently, it's become more possible than anticipated. This makes me wonder if perhaps `arrow(ends = { "last", "first", "both })` should start additionally differentiating something like `{ "segment_last", "segment_first", "segment_both" }` in order to give callers reasonable control over what's drawn. @yutannihilation suggested a commit in this direction in in #4133.

A couple workarounds for this issue are given in #4133 and it looks like at one point there might have been an attempt to address the issue from within the ggplot codebase.

- @yutannihilation's approach of suppressing unwanted arrows with an equivalent of `arrow(length = unit(if_else(<clause>, 0.25, 0), "line"))` is elegant. But, unfortunately, nothing I've tried lead to correctly placed arrows in my actual use case. So I can't figure out how to turn it into a viable workaround.
- I can get acceptable results at the hello world level using an equivalent of @JTC-R's doubled `geom_path()`. However, overdrawing composes poorly with alpha aesthetics and it'd be nice to have some way of getting variable arrowhead color and fill while maintaining only one arrow per group.
- There also appears to be at least one documentation error since [`arrow.fill`](https://ggplot2.tidyverse.org/reference/geom_segment.html) is rejected as both an unknown aesthetic and as an unknown parameter.

Additionally, I noticed `NA` doesn't work like it usually would to shut off arrow drawing. Instead, an arrow fragment is drawn extending far off the plotting area.
```
data = tibble(x = c(0, 0.5, 1), y = x, color = x)
ggplot(data) +
  geom_path(aes(x = x, y = y, color = color), arrow = arrow(length = unit(c(NA_real_, 5), "mm"), ends = "last"))
```

teunbrand:
This PR aims to fix #4935.

Visually, it brings the `!constant` --> `segmentsGrob()` code branch in line with to the `polylineGrob()` branch.

The `polylineGrob()` branch:

``` r
#> ℹ Loading ggplot2

data <- tibble::tibble(
  t = seq(5, -1, length.out = 100) * pi,
  x = sin(t) * 1:100,
  y = cos(t) * 1:100,
  group = 0:99 %/% 20
)

p <- ggplot(data, aes(x = x, y = y, group = group))

p + geom_path(arrow = arrow())
```

The `segmentsGrob()` branch:

``` r
p + geom_path(aes(colour = t), arrow = arrow())
```

Mixed use of arrow ends:

``` r
p + geom_path(
  aes(colour = t), 
  arrow = arrow(ends = c("last", "both", "first", "both", "last"))
)
```

<sup>Created on 2022-12-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


thomasp85 (approved):
LGTM - I've wanted to do this forever so happy to see it done...

PR #5081, opened 2022-12-03
coord_munch can convert closed shapes to closed form

### LINKED ISSUE #3271: coord_munch should convert closed shapes to closed form
Currently `coord_munch()` only munches the line segments between the start and end, that is, it doesn't automatically munch the closing of polygons, leading to weird straight edges when e.g. plotting polygons with `coord_polar()`... `geom_rect()` circumvents this by converting the rectangle to a closed form polygon, but such an operation is not done in e.g. `geom_polygon()`. In any way, this should be the responsibility of `coord_munch()` and not the individual geoms, as the polygon representation in grid is strictly open-form.

The easiest fix is probably to add a switch to `coord_munch()` to differentiate between open and closed shapes, and then have closed shape geoms identify themselves as such to the muncher

teunbrand:
This PR aims to fix #3271.

In brief:
* It repeats the first row in a group to after the last row in a group.
* Due to this, the `rect_to_poly()` function in `geom_rect()` made a redundant edge. While amending this, I found a more efficient way to polygonise rectangles, which is 2x faster for n = 2 rectangles, but 200x faster for n = 1000 rectangles.
* A few visual tests changed due to closing the shape, but I manually checked them and they're invisible changes.

A visual comparison:
Before:
``` r
library(ggplot2)

offset <- function(value, offset) as.vector(outer(value, offset, `+`))

df <- data.frame(
  x = offset(c(-0.5, 0, 1, 1.5, 1, 0, 0, 0, 1, 1), c(0, 2)),
  y = offset(c(0.5, 0, 0, 0.5, 1, 1, 0.25, 0.75, 0.75, 0.25), c(0, 2)),
  group = rep(1:2, c(10, 10)),
  subgroup = rep(1:4, c(6, 4, 6, 4))
)

ggplot(df, aes(x, y, group = group, fill = factor(group))) +
  geom_polygon(aes(subgroup = subgroup)) +
  coord_polar()
```

After:
``` r
#> ℹ Loading ggplot2

ggplot(df, aes(x, y, group = group, fill = factor(group))) +
  geom_polygon(aes(subgroup = subgroup)) +
  coord_polar()
```

<sup>Created on 2022-12-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/coord-munch.r
```diff
@@ -9,11 +9,17 @@
 #'   All other variables are duplicated as needed.
 #' @param range Panel range specification.
 #' @param segment_length Target segment length
+#' @param make_closed Whether data should be converted to closed polygons before
+#'   munching.
 #' @keywords internal
 #' @export
-coord_munch <- function(coord, data, range, segment_length = 0.01) {
+coord_munch <- function(coord, data, range, segment_length = 0.01, make_closed = FALSE) {
```
thomasp85, line 16: `make_closed` is probably not the right name here... the coordinates are closed per definition so something like `is_closed()` is better

R/coord-munch.r
```diff
@@ -9,11 +9,17 @@
...
   if (coord$is_linear()) return(coord$transform(data, range))
 
+  if (make_closed) {
+    data <- close_poly(data)
+  }
+
```
thomasp85, line 22: if the shape is closed then the last coordinate (which would be a replicate of the first) should be removed before returning the new data

R/coord-munch.r
teunbrand: So should I just flip the logic around and use `is_closed = TRUE` as default and then `geom_polygon()` uses `is_closed = FALSE`?

R/coord-munch.r
teunbrand: I'm not sure I'm following along correctly. The aim of this PR is to add an option to add the last coordinate as a replicate of the first (i.e. close a polygon), not opening up polygons.

R/coord-munch.r
thomasp85: No, `is_closed` describes whether the coordinates should be considered closed, which would be true for polygons... so it is simply a change in argument name and nothing more

R/coord-munch.r
thomasp85: Yes, we need to add the last coordinate so that the closing segment can be munched as well. But after the munching has been done we should remove the last coordinate again so that it is not replicated in the boundary

thomasp85: This should happen lower down in the function of course (after the `munch_data()` call), so maybe my comment was misplaced

teunbrand:
TODO: remove last point after munching

teunbrand:
The snapshots of maps have revert to their old state after removing the closing point.
However, rectangles in non-linear coordinates lose a closing point. Linejoins on rectangles look like they should.

thomasp85:
Yes - rectangles always had one too many so this change is considered a fix

thomasp85 (approved):
LGTM, but if you could provide a test that shows how this fixes rendering of closed paths it would be great

PR #5086, opened 2022-12-05
Switch to scales::Range classes

### LINKED ISSUE #2710: Range functions should be exported
I think the functions in https://github.com/tidyverse/ggplot2/blob/master/R/range.r should be exported. They are needed by anybody who wants to write their own scale functions from scratch.

Possibly they would fit better into the scales package, but since they depend on ggproto there's a dependency problem.

teunbrand:
This PR aims to fix #2710.

In brief, it replaces the unexported Range ggproto classes with the scales::Range R6 classes. This makes it slightly easier for extension developers to use the Range systems. Extenders have been able to do this for a while, but hopefully this serves as an 'endorsement' of using scales' Range classes.


teunbrand:
Yeah good point. It seems like a good idea to search github for the use of `ggplot2:::continuous_range()`/`ggplot2:::discrete_range()` and inform people that they might consider to switch to the scales::Range classes.

teunbrand:
Moreover, this move would be a little bit too risky for an upcoming patch release, so I think it'd be best to wait for after that has happened. I'll convert this to a draft until that happens, so it isn't accidentally merged.

yutannihilation:
I agree with you. Thanks, let's wait.

teunbrand:
Hu @yutannihilation, with ggplot2 3.4.1 having landed on CRAN, do you think this is OK to merge now?

yutannihilation:
Oh, thanks for the notice. I think so, but maybe we can merge #5171 first? We'll anyway need to resolve conflict on NEWS.md there, so I think either is fine.

yutannihilation:
Now that #5171 got merged, let's merge this.

teunbrand:
Thanks Hiroaki!

PR #5091, opened 2022-12-06
Write out units in full in `ggsave` error

### LINKED ISSUE #5048: ggsave inches not pixels error difficult to parse
If you give dimensions to `ggsave()` in pixels, you get the following error:

```
! Dimensions exceed 50 inches (`height`
  and `width` are specified in in not pixels).
```

The second `in` is italicized, but it still took me a few times of reading it over to realize it means "inches". I think it would be better if it just said "inches" there so that confusion can be avoided.

``` r
library(ggplot2)

  geom_point()
```

``` r

ggsave(filename = "mtcars.png", width = 1600, height = 900)
#> Error in `ggsave()`:
#> ! Dimensions exceed 50 inches (`height` and `width` are specified in in
#>   not pixels).
#> ℹ If you're sure you want a plot that big, use `limitsize = FALSE`.

#> Backtrace:
#>     ▆
#>  1. └─ggplot2::ggsave(filename = "mtcars.png", width = 1600, height = 900)
#>  2.   └─ggplot2:::plot_dim(...)
#>  3.     └─cli::cli_abort(...)
#>  4.       └─rlang::abort(...) at cli/R/rlang.R:43:2
```

<sup>Created on 2022-11-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5048.

In brief, all units are written out in full in the error message (not just inches). Also, there was a possibility that the user could see a confusing error like this:

> "Dimensions exceed 50 inches (height and width are specified in pixels not pixels)."

Which is now being avoided by omitting the 'not pixels' ending of the error message.

### DIFF
--- a/R/save.r
+++ b/R/save.r
@@ -147,6 +147,13 @@ plot_dim <- function(dim = c(NA, NA), scale = 1, units = "in",
   }
 
   if (limitsize && any(dim >= 50)) {
+    units <- switch(
+      units,
+      "in" = "inches",
+      "cm" = "centimeters",
+      "mm" = "millimeters",
+      "px" = "pixels"
+    )
     cli::cli_abort(c(
       "Dimensions exceed 50 inches ({.arg height} and {.arg width} are specified in {.emph {units}} not pixels).",
       "i" = "If you're sure you want a plot that big, use {.code limitsize = FALSE}.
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good!

PR #5098, opened 2022-12-08
Reset geom/stat defaults

### LINKED ISSUE #4993: Reset original defaults with `update_geom_defaults()`
Sometimes, as I'm interactively building a plot, I change geom default aesthetics with `update_geom_defaults()`, then change my mind and want to put them back. Currently, I don't see a way to do that other than by knowing what the original defaults were or restarting my R session. 

It would be really helpful to be able to do something like this to reset the `size` default to 1.5:

```r
update_geom_defaults("geom", list(size = NULL))
```

And/or like this to reset all aesthetics to their original defaults:

```r
update_geom_defaults("geom", NULL)
```

teunbrand:
This PR aims to fix #4993.

In brief, it creates caches for geom/stat defaults that are populated the first time a default is changed with `update_{geom/stat}_defaults()`. Then it uses that cache to reset the original defaults when `new = NULL`. In addition, the functions invisibly return the previous defaults.

I could not find an appropriate unit test for `update_stat_defaults()`. The examples also don't include `update_stat_defaults()` and I'm having a hard time imagining a use case for that function.

### DIFF
--- a/R/geom-defaults.r
+++ b/R/geom-defaults.r
@@ -1,28 +1,69 @@
+geom_defaults_cache <- new.env(parent = emptyenv())
+stat_defaults_cache <- new.env(parent = emptyenv())
+
 #' Modify geom/stat aesthetic defaults for future plots
 #'
 #' @param stat,geom Name of geom/stat to modify (like `"point"` or
 #'   `"bin"`), or a Geom/Stat object (like `GeomPoint` or
 #'   `StatBin`).
-#' @param new Named list of aesthetics.
+#' @param new Named list of aesthetics. Alternatively, `NULL` to reset the
+#'   defaults.
 #' @keywords internal
 #' @export
 #' @examples
 #' update_geom_defaults("point", list(colour = "darkblue"))
 #' ggplot(mtcars, aes(mpg, wt)) + geom_point()
-#' update_geom_defaults("point", list(colour = "black"))
+#' # Reset defaults by using `new = NULL`
+#' update_geom_defaults("point", NULL)
 #' @rdname update_defaults
 update_geom_defaults <- function(geom, new) {
-  g <- check_subclass(geom, "Geom", env = parent.frame())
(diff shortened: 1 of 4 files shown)


teunbrand:
I realised that the `test-function-args.r` test accidentally picked up on the cache as a function to test arguments for. Renaming the cache appears to resolve that problem.

teunbrand:
This is just a note that #5146 conflicts with this PR and either should update according to which one gets merged first.

R/geom-defaults.r
```diff
@@ -29,27 +31,63 @@
...
-  g <- check_subclass(geom, "Geom", env = parent.frame())
-  old <- g$default_aes
-  new <- rename_aes(new)
-  new_names_order <- unique(c(names(old), names(new)))
-  new <- defaults(new, old)[new_names_order]
-  g$default_aes[names(new)] <- new
-  invisible()
+  if (is.null(new)) {
+
+    vec_assert(geom, character(), 1)
```
thomasp85, line 40: Shouldn't you use the check infrastructure here?

R/geom-defaults.r
```diff
@@ -29,27 +31,63 @@
...
+    old <- cache_geom_defaults[[geom]]
+    if (!is.null(old)) {
+      new <- update_geom_defaults(geom, old)
+      env_unbind(cache_geom_defaults, geom)
```
thomasp85, line 44: Is there a purpose for removing the cache here? Can't we just keep it there if it is needed in the future

R/geom-defaults.r
teunbrand: Good point, also forgot the case where `geom = GeomPoint` as ggproto case. Fixed now.

R/geom-defaults.r
teunbrand: Other than a habit of cleaning up after itself, it doesn't really have any function. I've kept it now.

Issue #5113, opened 2022-12-20
Faulty y-axis for plots based on stat_ecdf with scale_y_log10


m-pilarski:
When I'm log-scaling the y-axis of CDF-plots (`stat_ecdf`), the y-values are drawn between 1 and 10 instead of 0 and 1. The y-values of the data extracted from the plot-object (`ggplot_build(...)$data[[1]]$y`) are ranging from 0 to 1 instead of -1 to 0.  When I'm log-scaling the x-axis of CDF-Plots or the x- or y-axis of other plots (e.g. `geom_line`), everything is drawn as expected.

``` r
library(ggplot2)

df <- data.frame(var=1:10/10)

plot_ecdf_ident <- ggplot(df, aes(x=var)) + stat_ecdf()
plot_ecdf_log10_y <- plot_ecdf_ident + scale_y_log10()
plot_ecdf_log10_x <- plot_ecdf_ident + scale_x_log10()

plot_line_ident <- ggplot(df, aes(x=var, y=var)) + geom_line()
plot_line_log10_y <- plot_line_ident + scale_y_log10()
plot_line_log10_x <- plot_line_ident + scale_x_log10()

patchwork::wrap_plots(
  plot_ecdf_ident, plot_ecdf_log10_y, plot_ecdf_log10_x, 
  plot_line_ident, plot_line_log10_y, plot_line_log10_x, 
  nrow=2
)
```

![](https://i.imgur.com/wVDBU8h.png)

``` r

range(ggplot_build(plot_ecdf_log10_y)$data[[1]]$y)
#> [1] 0 1
range(ggplot_build(plot_line_log10_y)$data[[1]]$y)
#> [1] -1  0

packageVersion("ggplot2")
#> [1] '3.4.0'
packageVersion("scales")
#> [1] '1.2.1'
```

<sup>Created on 2022-12-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

m-pilarski:
I found out that the faulty log-scaling has nothing to do with the default `pad=TRUE` in `stat_ecdf`.  I also found out that the transformation woks, when I'm feeding the the data extracted from the untransformed plot to a new plot with `geom_step`.

``` r
library(ggplot2)

`%>%` <- magrittr::`%>%`

df <- data.frame(var=1:10/10)

plot_ecdf_ident_nopad_1 <- ggplot(df, aes(x=var)) + stat_ecdf(pad=FALSE)

[...]

PR #5115, opened 2022-12-22
`stat_ecdf()` responds to scale transform

### LINKED ISSUE #5112: Bug in ggplot2 3.4.0 using scales::probability_trans
Prompted by this question on StackOverflow, https://stackoverflow.com/questions/74809918/y-axis-scaling-to-normal-distribution there seems to be a bug in how ggplot2 3.4.0 does axis transformations. The example that works as expected in version 3.3.5, producing a nice straight line, is 

```r
data.frame(v = rnorm(1000)) %>% 
    ggplot(aes(x = v)) + 
    stat_ecdf() + 
    scale_y_continuous(trans=scales::probability_trans("norm"))
```

Compare that to the same code in 3.4.0, which gives an S-shaped curve and also has the y axis starting at 0.5. 

In this case, the transformation seems to be using `pnorm` and `qnorm` the wrong way round in adjusting the scale and the axis. 

I tried a few other transformations, and several of them look odd, so I think this is an example of something deeper in the axis transformation routine. 

I'm using R 4.2.2 on Windows 11.

teunbrand:
This PR aims to fix #5112 and #5113.

Because this stat used to have `y = after_stat(y)`, it computed a column named `y`. This, of course, is recognised as a position aesthetic, but that also meant that it was back-transformed in `scales_backtransform_df()` (whereas it shouldn't be). The fix proposed herein is to simply rename the computed aesthetic to `ecdf` such that it is no longer automatically backtransformed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Renamed computed aesthetic in `stat_ecdf()` to `ecdf`, to prevent incorrect
+  scale transformations (@teunbrand, #5113 and #5112).
 * Fixed a regression in `geom_hex()` where aesthetics were replicated across 
   bins (@thomasp85, #5037 and #5044)
 * Fixed spurious warning when `weight` aesthetic was used in `stat_smooth()` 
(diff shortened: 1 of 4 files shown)


thomasp85 (approved):
LGTM - this is in theory a breaking change but I suspect no one will notice

teunbrand:
Right, yes good observation

PR #5117, opened 2022-12-23
Computed variable and delayed evaluation docs

### LINKED ISSUE #4951: Computed variable docs should suggest `after_stat()`
e.g. `geom_freqpoly()` lists `count`, `density`, etc. Should mention that you need `after_stat()` somewhere.

teunbrand:
This PR aims to fix #4951

Briefly, in 'computed variables' sections there is now a pointer to the `?after_stat` page, and all variable names are wrapped in `after_stat()`. In addition, the `?after_stat` page has been restructured to more clearly articulate what to expect for each stage, as well as sprinkling around some extra, slightly more complicated examples.

### DIFF
--- a/R/aes-evaluation.r
+++ b/R/aes-evaluation.r
@@ -1,12 +1,19 @@
 #' Control aesthetic evaluation
 #'
+#' @description
 #' Most [aesthetics][aes()] are mapped from variables found in the data.
 #' Sometimes, however, you want to delay the mapping until later in the
 #' rendering process. ggplot2 has three stages of the data that you can map
 #' aesthetics from, and three functions to control at which stage aesthetics
 #' should be evaluated.
 #'
-#' @usage # These functions can be used inside the `aes()` function
+#' @description
+#' `after_stat()` replaces the old approaches of using either `stat()`, e.g.
+#' `stat(density)`, or surrounding the variable names with `..`, e.g.
+#' `..density..`.
+#'
+#' @usage
+#' # These functions can be used inside the `aes()` function
 #' # used as the `mapping` argument in layers, for example:
 #' # geom_density(mapping = aes(y = after_stat(scaled)))
 #'
@@ -80,9 +87,9 @@
 #' ```
 #'
 #' ## Complex staging
-#'   If you want to map the same aesthetic multiple times, e.g. map `x` to a
-#'   data column for the stat, but remap it for the geom, you can use the
-#'   `stage()` function to collect multiple mappings.
+#' If you want to map the same aesthetic multiple times, e.g. map `x` to a
+#' data column for the stat, but remap it for the geom, you can use the
+#' `stage()` function to collect multiple mappings.
 #'
 #' ```r
 #' # Use stage to modify the scaled fill
@@ -102,10 +109,6 @@
 #'     fun.data = ~ round(data.frame(mean = mean(.x), sd = sd(.x)), 2)
 #'   )
 #' ```
-#' @note
-#' `after_stat()` replaces the old approaches of using either `stat()`, e.g.
-#' `stat(density)`, or surrounding the variable names with `..`, e.g.
-#' `..density..`.
 #' @rdname aes_eval
 #' @name aes_eval
 #'
--- a/R/aes-evaluation.r
+++ b/R/aes-evaluation.r
@@ -90,16 +90,15 @@
 #' ggplot(mpg, aes(class, hwy)) +
 #'   geom_boxplot(aes(fill = stage(class, after_scale = alpha(fill, 0.4))))
 #'
-#' # Using data for computing summary, but placing label elsewhere, as well as
-#' # using the computed 'y' as part of the label.
+#' # Using data for computing summary, but placing label elsewhere.
+#' # Also, we're making our own computed variable to use for the label.
 #' ggplot(mpg, aes(class, displ)) +
 #'   geom_violin() +
 #'   stat_summary(
 #'     aes(y = stage(displ, after_stat = 8),
-#'         label = after_stat(paste(y, ymax, sep = " ± "))),
+#'         label = after_stat(paste(mean, "±", sd))),
 #'     geom = "text",
-#'     fun = ~ round(mean(.x), 2),
-#'     fun.max = ~ round(sd(.x), 2)
+#'     fun.data = ~ round(data.frame(mean = mean(.x), sd = sd(.x)), 2)
 #'   )
 #' ```
 #'
--- a/R/aes-evaluation.r
+++ b/R/aes-evaluation.r
@@ -6,7 +6,8 @@
 #' three functions to control at which stage aesthetics should be evaluated.
 #'
 #' @usage # These functions can be used inside the `aes()` function
-#' # used as the `mapping` argument in layers.
+#' # used as the `mapping` argument in layers, for example:
+#' # geom_density(mapping = aes(y = after_stat(scaled)))
 #'
 #' @param x <[`data-masking`][rlang::topic-data-mask]> An aesthetic expression
 #'   using variables calculated by the stat (`after_stat()`) or layer aesthetics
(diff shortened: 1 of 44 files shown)


R/aes-evaluation.r
```diff
@@ -2,39 +2,106 @@
...
+#' # Using data for computing summary, but placing label elsewhere, as well as
+#' # using the computed 'y' as part of the label.
+#' ggplot(mpg, aes(class, displ)) +
+#'   geom_violin() +
+#'   stat_summary(
+#'     aes(y = stage(displ, after_stat = 8),
+#'         label = after_stat(paste(y, ymax, sep = " ± "))),
+#'     geom = "text",
+#'     fun = ~ round(mean(.x), 2),
+#'     fun.max = ~ round(sd(.x), 2)
```
yjunechoe, line 102: Small nitpicking only because of my love/hate relationship with the `fun.*` arguments in `stat_summary()`:

May I suggest just passing a function to `fun.data` returning a dataframe with the desired variables-as-columns, as opposed to supplying functions to `fun` and `fun.max` that each return vectors?

So those 2 lines can be replaced with:

```r
# possibly `\(val) ...` so the argument is not confused with the x-positional aesthetic
fun.data = \(x) round(data.frame(mean = mean(x), sd = sd(x)), 2)
```

Then, the columns from `fun.data` output can be referenced in the `after_stat()` for `label` like this, which IMO is more transparent:

```r
after_stat(paste(mean, sd, sep = " ± "))
```

And it also produces a more straightforward layer data:

```r
p_new <- ggplot(mpg, aes(class, displ)) +
   geom_violin() +
   stat_summary(
     aes(y = stage(displ, after_stat = 8),
         label = after_stat(paste(mean, sd, sep = " ± "))),
     geom = "text",
     fun.data = \(x) round(data.frame(mean = mean(x), sd = sd(x)), 2)
   )
layer_data(last_plot(), 2)[,1:7]
#>   y       label x group mean   sd PANEL
#> 1 8 6.16 ± 0.53 1     1 6.16 0.53     1
#> 2 8 2.33 ± 0.45 2     2 2.33 0.45     1
#> 3 8 2.92 ± 0.72 3     3 2.92 0.72     1
#> 4 8 3.39 ± 0.45 4     4 3.39 0.45     1
#> 5 8 4.42 ± 0.83 5     5 4.42 0.83     1
#> 6 8  2.66 ± 1.1 6     6 2.66 1.10     1
#> 7 8 4.46 ± 1.07 7     7 4.46 1.07     1
```

Without overriding `fun.data` (or doing an ugly `fun.min = ~ NULL`), you get the vestigial `ymin` column that's there to support the default pointrange geom. Plus, the mean that's calculated from `fun` gets mapped to `y` and then is immediately overriden by `stage(after_stat = 8)` which may be surprising (in case one wanted to use it to map `size` in the after-scale of something):

```r
p_old <- ggplot(mpg, aes(class, displ)) +
   geom_violin() +
   stat_summary(
     aes(y = stage(displ, after_stat = 8),
         label = after_stat(paste(y, ymax, sep = " ± "))),
     geom = "text",
     fun = ~ round(mean(.x), 2),
     fun.max = ~ round(sd(.x), 2)
   )
layer_data(last_plot(), 2)[,1:7]
#>   y       label x group ymin ymax PANEL
#> 1 8 6.16 ± 0.53 1     1   NA 0.53     1
#> 2 8 2.33 ± 0.45 2     2   NA 0.45     1
#> 3 8 2.92 ± 0.72 3     3   NA 0.72     1
#> 4 8 3.39 ± 0.45 4     4   NA 0.45     1
#> 5 8 4.42 ± 0.83 5     5   NA 0.83     1
#> 6 8  2.66 ± 1.1 6     6   NA 1.10     1
#> 7 8 4.46 ± 1.07 7     7   NA 1.07     1
```

Lastly and relatedly, because sd is held in the `ymax` column in the old example, it gets used later in the retraining of the y-scales, so if you plot `p_old` you see it actually gets an expansion down to around 0 in the y-axis because `ymax` is around that value. All of this mess because `stat_summary()` appears as if it's geom-agnostic when it kinda isn't 😅

I may be missing obvious problems with my approach and this is possibly off-label usage of `fun.data`, but as long as `fun.data` dataframe doesn't touch standard aesthetics (like overriding/duplicating the `x` column or hard-coding `colour` before the scale sees it), this should be ok?

(And of course thanks for this great addition to the docs!)

teunbrand: Thanks June, I love the suggestion. It makes it much clearer about what is going on! Especially the y-axis expansion I hadn't spotted, so thanks :) The only way I'll deviate from your suggestion is that the example should work in older R versions than R4.1.0, so I'll stick to the rlang lambda syntax instead of base R lambda syntax for now.

yutannihilation:
Looks awesome!

@hadley 
Could you take a look when you have time? Since #4951 is what you proposed, probably you are the best one to review this.

R/aes-evaluation.r
```diff
@@ -2,39 +2,105 @@
...
+#' @note
 #' `after_stat()` replaces the old approaches of using either `stat()` or
 #' surrounding the variable names with `..`.
 #'
-#' @note Evaluation after stat transformation will have access to the
-#' variables calculated by the stat, not the original mapped values. Evaluation
-#' after scaling will only have access to the final aesthetics of the layer
-#' (including non-mapped, default aesthetics). The original layer data can only
-#' be accessed at the first stage.
+#' @section Staging:
```
hadley, line 25:
```suggestion
#' #Staging
```

For consistency

R/aes-evaluation.r
```diff
@@ -2,39 +2,105 @@
...
-#' aesthetic mapping should be postponed until after stat transformation.
-#' Similarly, you should use `after_scale()` to flag evaluation of mapping for
-#' after data has been scaled. If you want to map the same aesthetic multiple
-#' times, e.g. map `x` to a data column for the stat, but remap it for the geom,
-#' you can use the `stage()` function to collect multiple mappings.
+#' ggplot2 has three stages of the data that you can map aesthetics from, and
+#' three functions to control at which stage aesthetics should be evaluated.
 #'
+#' @usage # These functions can be used inside the `aes()` function
+#' # used as the `mapping` argument in layers.
```
hadley, line 9: Since you're add a fake usage, maybe also include some example `aes()` calls in comments?

R/aes-evaluation.r
```diff
@@ -2,39 +2,105 @@
...
-#' variables calculated by the stat, not the original mapped values. Evaluation
-#' after scaling will only have access to the final aesthetics of the layer
-#' (including non-mapped, default aesthetics). The original layer data can only
-#' be accessed at the first stage.
+#' @section Staging:
+#' Below follows an overview of the three stages of evaluation and how aesthetic
+#' evaluation can be controlled.
+#'
+#' ## Stage 1: direct input
+#'   The default is to map at the beginning, using the layer data provided by
```
hadley, line 30: These shouldn't be indented

R/aes-evaluation.r
```diff
@@ -2,39 +2,105 @@
...
+#' ggplot(faithful, aes(x = waiting)) +
+#'   geom_histogram()
+#'
+#' # Choosing a different computed variable to display, matching up the
+#' # histogram with the density plot
+#' ggplot(faithful, aes(x = waiting)) +
+#'   geom_histogram(aes(y = after_stat(density))) +
+#'   geom_density()
+#' ```
+#'
```
hadley, line 64: Might be worth mentioning deprecated `..density..` style.

R/aes-evaluation.r
```diff
@@ -2,39 +2,105 @@
...
+#' Below follows an overview of the three stages of evaluation and how aesthetic
+#' evaluation can be controlled.
+#'
+#' ## Stage 1: direct input
+#'   The default is to map at the beginning, using the layer data provided by
+#'   the user. If you want to map directly from the layer data you should not do
+#'   anything special. This is the only stage where the original layer data can
+#'   be accessed.
+#'
+#' ```{r direct_input, eval = FALSE}
```
hadley, line 35: If you're not going to eval these,  ```` ```r ```` is simpler.

R/aes-evaluation.r
```diff
@@ -55,6 +121,44 @@
 #' # Use stage to modify the scaled fill
 #' ggplot(mpg, aes(class, hwy)) +
 #'   geom_boxplot(aes(fill = stage(class, after_scale = alpha(fill, 0.4))))
+#'
+#' # Making a proportional stacked density plot
+#' ggplot(mpg, aes(cty)) +
+#'   geom_density(
+#'     aes(colour = factor(cyl),
```
hadley, line 128: Needs to follow https://style.tidyverse.org/syntax.html#long-lines. You might want to use https://github.com/lionel-/codegrip to make the transformations easier. (Although it probably doesn't work inside roxygen2 comments yet)

R/geom-dotplot.r
```diff
@@ -18,16 +18,19 @@
 #'
 #' @eval rd_aesthetics("geom", "dotplot")
 #' @section Computed variables:
+#' These are calculated by the 'stat' part of layers and can be accessed with
+#' [delayed evaluation][aes_eval].
 #' \describe{
```
hadley, line 23: Should we just switch these to regular bulleted lists? Is the output of a `\describe` that much better?

hadley (changes requested):
Looks good! I added a few comments specifically about the style.

R/geom-dotplot.r
teunbrand: I think there are downsides to both `\describe` and to `\itemize`. `\describe` is slightly too spacious for my taste, whereas `\itemize` looks messy because the description follows the item name without any visual cue that they are separate. I think to have best of both words, we could use `\itemize` with `\cr` to separate the item name from description, which would then look like this:

* `after_stat(density)`
  density estimate

* `after_stat(count)`
  density * number of points - useful for stacked density plots.

R/aes-evaluation.r
teunbrand: I've more explicitly put this in the `@note` now.

teunbrand:
Thanks for your comments Hadley! I've adapted at most places, and I'll go write an `rd_computed_vars()` helper function to format the section better.

teunbrand:
Alright if we would disagree about the list method (describe vs itemize), it should now be relatively easy to adjust the `rd_compute_vars()` wrapper instead of manually going through all of the stat functions :)

R/utilities-help.r
```diff
@@ -49,3 +49,57 @@ rd_orientation <- function() {
...
+rd_computed_vars <- function(..., .details = "", .skip_intro = FALSE) {
+  args  <- list(...)
+  items <- names(args)
+  descr <- unname(args)
+
+  # Format preamble
+  header <- "@section Computed variables: "
+  intro  <- paste0(
+    "These are calculated by the 'stat' part of layers and can be accessed ",
+    "with \\link[=aes_eval]{delayed evaluation}. "
```
hadley, line 89: Why not use markdown here?

R/utilities-help.r
```diff
@@ -49,3 +49,57 @@ rd_orientation <- function() {
...
+
+  # Format items
+  fmt_items <- gsub(",", ")}, \\code{after_stat(", items, fixed = TRUE)
+  fmt_items <- gsub("|", ")} \\emph{or} \\code{after_stat(",
+                    fmt_items, fixed = TRUE)
+  fmt_items <- paste0("\\item{\\code{after_stat(", fmt_items, ")}")
+
+  # Compose item-list
+  fmt_descr <- paste0(gsub("\n", "", descr), "}")
+  fmt_list  <- paste(fmt_items, fmt_descr, sep = "\\cr ")
```
hadley, line 102:
```suggestion
  fmt_list  <- paste(fmt_items, fmt_descr, sep = ": ")
```
?

hadley (approved):

hadley:
Ok, that looks good to me!

PR #5119, opened 2022-12-24
Weighted eCDF

### LINKED ISSUE #5058: Add weights to stat_ecdf
Currently, `stat_ecdf` does not support specifying a `weight` aesthetic to calculate and plot weighted ECDFs. There are a number of third-party implementations that have been made, e.g. [ggecdf](https://malcolmbarrett.github.io/ggecdf/reference/geom_ecdf.html), [stat_ecdf_weighted](https://github.com/NicolasWoloszko/stat_ecdf_weighted), [stat_ewcdf](https://finnlindgren.github.io/StatCompLab/reference/stat_ewcdf.html). However, as ggplot2 continues to develop and evolve, it would be useful to have this as a fully supported feature to ensure it continues being available.

teunbrand:
This PR aims to fix #5058.

Briefly, it adds an optional `weight` aesthetic to `stat_ecdf()`. If a weight is present, we calculate the ecdf in a different way, wherein each observations is weighted by the amount of the observation's weight relative to the sum of all weights in the group.


thomasp85:
I have no experience with ecdfs so I can't really comment on the correctness of the weighted implementation. @clauswilke or @yutannihilation do you feel confident in reviewing this?

yutannihilation:
I too am not familiar with eCDF, sorry...

clauswilke:
The calculation looks correct on first glance. I might want to read it through a little more carefully before signing off on it, but fundamentally it's very simple. And eCDF is simply a cumulative sum of the ordered values, divided by the total sum. To make this weighted, you multiply each value by a weight before you sum.

I have one concern though: I don't particularly like using a built-in function to calculate eCDF and a custom function to calculate weCDF. Why not use the same function for both and set the weights to 1 if not provided?

teunbrand:
> I have one concern though: I don't particularly like using a built-in function to calculate eCDF and a custom function to calculate weCDF. Why not use the same function for both and set the weights to 1 if not provided?

Fair point, I don't think there is a particular reason I did it this way. Setting the weights to 1 should indeed give identical output.

R/stat-ecdf.R
```diff
@@ -100,7 +121,11 @@ StatEcdf <- ggproto("StatEcdf", Stat,
     if (pad) {
       x <- c(-Inf, x, Inf)
     }
-    data_ecdf <- ecdf(data$x)(x)
+    if (is.null(data$weight)) {
+      data_ecdf <- ecdf(data$x)(x)
+    } else {
+      data_ecdf <- wecdf(data$x, data$weight)(x)
+    }
```
thomasp85, line 128: We should just use `wecdf()` as per the discussion

thomasp85 (approved):
LGTM apart from the one comment

PR #5122, opened 2022-12-27
Fix typo for `compute_panel_` to remove the trailing underscore

billdenney:
I think that this is a typo in the documentation.  If not, then can the difference between `compute_panel` and `compute_panel_` please be clarified?


teunbrand (approved):
Looks good to me. I don't think there is a `compute_panel_()` method anywhere.

PR #5129, opened 2023-01-04
Add `drop` argument to `stat_ydensity()`

### LINKED ISSUE #5002: Dropped cases are taken into account for the dodging of geom_boxplot but not geom_violin
Dropped cases are taken into account for dropped groups are still taken into account for the dodging of geom_boxplot but not geom_violin

The issue is detailed with reproducible code, a diagnosis, and solution from a commenter on StackOverflow. It appears there may be an inconsistency between geom_violinn and geom_boxplot.

https://stackoverflow.com/questions/73900036/boxplot-and-violin-plot-misaligned-in-ggplot2-for-only-one-level-of-the-x-axis/73900343#73900343

teunbrand:
This PR aims to fix #5002.

Briefly, it adds a `drop = TRUE/FALSE` argument to `stat_ydensity()`, so that groups with less than 2 observations can still be considered for position adjustment purposes (i.e. dodging).

I have some doubts whether the default for `drop` should be `TRUE` or `FALSE`, but I've opted for preserving current behaviour by default.

As a visual example:

``` r
#> ℹ Loading ggplot2

df <- data.frame(
  x = factor(rep(1:2, each = 4)),
  y = rep(1:2, 4),
  g = rep(c("A", "A", "B", 'B'), 2)
)

dodge <- position_dodge(0.9)
p <- ggplot(head(df, -1), mapping = aes(x, y, fill = g)) +
  geom_boxplot(position = dodge)

p + geom_violin(alpha = 0.5, position = dodge, drop = TRUE)
#> Warning: Groups with fewer than two datapoints have been dropped.
#> ℹ Set `drop = FALSE` to consider such groups for position adjustment purposes.
```

``` r

p + geom_violin(alpha = 0.5, position = dodge, drop = FALSE)
#> Warning: Cannot compute density for groups with fewer than two datapoints.
```

<sup>Created on 2023-01-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


thomasp85 (approved):
LGTM... I think we should preserve the current behaviour as the code does now

PR #5131, opened 2023-01-05
Edits to ggplot man page

kevinxperese:
* changed "component" to "components"
* used "pattern" instead of "way" and "method"
* broke each pattern description in to separate paragraph
* used more explicit names in example code
* added more explicit comments in example code
* removed error bars in last pattern of example code (and removed standard deviations from the summary stats calculated)

### DIFF
--- a/man/ggplot.Rd
+++ b/man/ggplot.Rd
@@ -26,64 +26,64 @@ set of plot aesthetics intended to be common throughout all
 subsequent layers unless specifically overridden.
 }
 \details{
-\code{ggplot()} is used to construct the initial plot object,
-and is almost always followed by \code{+} to add component to the
-plot. There are three common ways to invoke \code{ggplot()}:
+\code{ggplot()} is used to construct the initial plot object
+and is almost always followed by a plus sign (\code{+}) to add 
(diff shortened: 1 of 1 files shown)


teunbrand:
Hi there,

In general, I think these proposed changes make the example code clearer.
However, I think you might have edited man page itself, which is autogenerated from roxygen comment blocks.
This is indicated on the first two line of every doc page, e.g.:
https://github.com/tidyverse/ggplot2/blob/d85eb6255ee39c3538a8a03e179fcb97dc594193/man/ggplot.Rd#L1-L2

The documents are generated with [{roxygen2}](https://roxygen2.r-lib.org/), so if you want to change the documentation, you'd need the change the text behind roxygen comments (the `#'` indicators) in the file where the function is defined:

https://github.com/tidyverse/ggplot2/blob/d85eb6255ee39c3538a8a03e179fcb97dc594193/R/plot.r#L1-L75

As a small side note for the future, it would be helpful for maintainers if you'd indicate *why* you are proposing changes. In this case that is to 'improve the documentation of the `ggplot()` function' for example.

teunbrand:
I'm closing this in favour of #5132

PR #5132, opened 2023-01-05
Improve clarity of docstring for ggplot()

kevinxperese:
* changed "component" to "components"
* used "pattern" instead of "way" and "method"
* broke each pattern description in to separate paragraph used more explicit names in example code
* used more explicit data frame and variable names in example code
* added more explicit comments in example code
* removed error bars in last pattern of example code (and removed standard deviations from the summary stats calculated)

### DIFF
--- a/R/plot.r
+++ b/R/plot.r
@@ -8,10 +8,11 @@
 #' `ggplot()` is used to construct the initial plot object,
 #' and is almost always followed by a plus sign (`+`) to add
 #' components to the plot.
+#'
 #' There are three common patterns used to invoke `ggplot()`:
 #'
-#' * `ggplot(df, aes(x, y, other aesthetics))`
-#' * `ggplot(df)`
+#' * `ggplot(data = df, mapping = aes(x, y, other aesthetics))`
+#' * `ggplot(data = df)`
 #' * `ggplot()`
 #'
 #' The first pattern is recommended if all layers use the same
@@ -22,12 +23,17 @@
 #' The second pattern specifies the default data frame to use
 #' for the plot, but no aesthetics are defined up front. This
 #' is useful when one data frame is used predominantly for the
-#' plot, but the aesthetics may vary from one layer to another. 
+#' plot, but the aesthetics vary from one layer to another. 
 #'
 #' The third pattern initializes a skeleton `ggplot` object, which
 #' is fleshed out as layers are added. This is useful when
 #' multiple data frames are used to produce different layers, as
 #' is often the case in complex graphics.
+#'
+#' The `data =` and `mapping =` specifications in the arguments are optional
+#' (and are often omitted in practice), so long as the data and the mapping
+#' values are passed into the function in the right order. In the examples
+#' below, however, they are left in place for clarity.
 #'
 #' @param data Default dataset to use for plot. If not already a data.frame,
 #'   will be converted to one by [fortify()]. If not specified,
@@ -40,39 +46,55 @@
 #' @export
 #' @examples
 #' # Create a data frame with some sample data, then create a data frame
-#' # containing the mean for each group in the sample data.
+#' # containing the mean value for each group in the sample data.
 #' set.seed(1)
+#' 
 #' sample_df <- data.frame(
 #'   group = factor(rep(letters[1:3], each = 10)),
 #'   value = rnorm(30)
 #' )
-#' group_means_df <- do.call(rbind, lapply(split(sample_df, sample_df$group), function(d) {
-#'   data.frame(group_mean = mean(d$value), group = d$group)
-#' }))
+#' 
+#' group_means_df <- setNames(
+#'   aggregate(value ~ group, sample_df, mean),
+#'   c("group", "group_mean")
+#' )
+#' 
+#' # The following three code blocks create the same graphic, each using one
+#' # of the three patterns specified above. In each graphic, the group means
+#' # data frame is used to plot larger red points on top of the sample data.
 #' 
 #' # Pattern 1
-#' # The group means data frame is used to plot larger red points on top
-#' # of the sample data. Note that we don't need to supply `data =` or `mapping =`
-#' # in each layer because the arguments are passed into ggplot() in the default
-#' # positions.
-#' ggplot(sample_df, aes(x = group, y = value)) +
+#' # Both the `data` and the `mapping` arguments are passed into the `ggplot()`
+#' # call. Note that we don't need to supply `mapping` or `data` arguments in
+#' # the first `geom_point()` layer because those arguments get passed along
+#' # from the `ggplot()` call.
+#' ggplot(data = sample_df, mapping = aes(x = group, y = value)) +
 #'   geom_point() +
-#'   geom_point(group_means_df, aes(y = group_mean), colour = 'red', size = 3)
+#'   geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
+#'              colour = 'red', size = 3
+#'     )
 #' 
 #' # Pattern 2
-#' # Same plot as above, declaring only the data frame in ggplot().
-#' # Note how the x and y aesthetics must now be declared in
-#' # each geom_point() layer.
-#' ggplot(sample_df) +
-#'   geom_point(aes(x = group, y = value)) +
-#'   geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
+#' # Same plot as above, passing in only the `data` argument in the `ggplot()`
+#' # call. Note how the `mapping` arguments must now be passed in for each
+#' # `geom_point()` layer because there are no `mapping` arguments passed
+#' # along from the `ggplot()` call.
+#' ggplot(data = sample_df) +
+#'   geom_point(mapping = aes(x = group, y = value)) +
+#'   geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
+#'              colour = 'red', size = 3
+#'   )
 #' 
 #' # Pattern 3
-#' # Alternatively, we can fully specify the plot in each layer. This
-#' # can be particularly useful when working with complex, multi-dataset graphics.
+#' # Same plot as above, passing in neither the `data` or `mapping` arguments
+#' # to the `ggplot()` call. Both those arguments must then be fully specified
+#' # in each layer. This pattern can be particularly useful when creating more
+#' # complex graphics.
 #' ggplot() +
-#'  geom_point(sample_df, aes(x = group, y = value)) +
-#'  geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
+#'   geom_point(mapping = aes(x = group, y = value), data = sample_df) +
+#'   geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
+#'              colour = 'red', size = 3
+#'   )
 ggplot <- function(data = NULL, mapping = aes(), ...,
                    environment = parent.frame()) {
   UseMethod("ggplot")
--- a/R/plot.r
+++ b/R/plot.r
@@ -57,7 +57,7 @@
 #' # positions.
 #' ggplot(sample_df, aes(x = group, y = value)) +
 #'   geom_point() +
-#'   geom_point(group_means_df, aes(y = group_mean), colour = 'red', size = 3)
+#'   geom_point(group_means_df, aes(y = group_mean, colour = 'red', size = 3))
 #' 
 #' # Pattern 2
 #' # Same plot as above, declaring only the data frame in ggplot().
@@ -65,15 +65,14 @@
 #' # each geom_point() layer.
 #' ggplot(sample_df) +
 #'   geom_point(aes(x = group, y = value)) +
-#'   geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
+#'   geom_point(group_means_df, aes(x = group, y = group_mean, colour = 'red', size = 3))
 #' 
 #' # Pattern 3
 #' # Alternatively, we can fully specify the plot in each layer. This
 #' # can be particularly useful when working with complex, multi-dataset graphics.
 #' ggplot() +
 #'  geom_point(sample_df, aes(x = group, y = value)) +
-#'  geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
-#' }
+#'  geom_point(group_means_df, aes(x = group, y = group_mean, colour = 'red', size = 3))
 ggplot <- function(data = NULL, mapping = aes(), ...,
                    environment = parent.frame()) {
   UseMethod("ggplot")
--- a/R/plot.r
+++ b/R/plot.r
@@ -57,7 +57,7 @@
 #' # positions.
 #' ggplot(sample_df, aes(x = group, y = value)) +
 #'   geom_point() +
-#'   geom_point(group_means_df, aes(y = group_mean, colour = 'red', size = 3))
+#'   geom_point(group_means_df, aes(y = group_mean), colour = 'red', size = 3)
 #' 
 #' # Pattern 2
 #' # Same plot as above, declaring only the data frame in ggplot().
@@ -65,14 +65,14 @@
 #' # each geom_point() layer.
 #' ggplot(sample_df) +
 #'   geom_point(aes(x = group, y = value)) +
-#'   geom_point(group_means_df, aes(x = group, y = group_mean, colour = 'red', size = 3))
+#'   geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
 #' 
 #' # Pattern 3
 #' # Alternatively, we can fully specify the plot in each layer. This
 #' # can be particularly useful when working with complex, multi-dataset graphics.
 #' ggplot() +
 #'  geom_point(sample_df, aes(x = group, y = value)) +
-#'  geom_point(group_means_df, aes(x = group, y = group_mean, colour = 'red', size = 3))
+#'  geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
 ggplot <- function(data = NULL, mapping = aes(), ...,
                    environment = parent.frame()) {
   UseMethod("ggplot")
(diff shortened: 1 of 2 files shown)


R/plot.r
```diff
@@ -36,43 +39,40 @@
...
+#' # Pattern 1
+#' # The group means data frame is used to plot larger red points on top
+#' # of the sample data. Note that we don't need to supply `data =` or `mapping =`
+#' # in each layer because the arguments are passed into ggplot() in the default
+#' # positions.
+#' ggplot(sample_df, aes(x = group, y = value)) +
 #'   geom_point() +
-#'   geom_point(data = ds, aes(y = mean), colour = 'red', size = 3)
-#'
+#'   geom_point(group_means_df, aes(y = group_mean), colour = 'red', size = 3)
```
teunbrand, line 60: This example currently doesn't work. The first argument is the mapping, the second one is the data.
Probably best to be explicit and give `data = group_means_df`. My suggestion:
```r
  geom_point(
    aes(y = group_mean),
    data = group_means_df, 
    colour = 'red', size = 3
  )
```

R/plot.r
```diff
@@ -36,43 +39,40 @@
...
-#' ggplot(df) +
-#'   geom_point(aes(gp, y)) +
-#'   geom_point(data = ds, aes(gp, mean), colour = 'red', size = 3)
-#'
-#' # Alternatively we can fully specify the plot in each layer. This
-#' # is not useful here, but can be more clear when working with complex
-#' # mult-dataset graphics
+#' ggplot(sample_df) +
+#'   geom_point(aes(x = group, y = value)) +
+#'   geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
```
teunbrand, line 68: This has the same problem as the example before. My suggestion:

```r
  geom_point(
    aes(x = group, y = group_mean),
    data = group_means_df, 
    colour = "red", size = 3
  )
```

R/plot.r
```diff
@@ -36,43 +39,40 @@
...
-#'   geom_point(data = df, aes(gp, y)) +
-#'   geom_point(data = ds, aes(gp, mean), colour = 'red', size = 3) +
-#'   geom_errorbar(
-#'     data = ds,
-#'     aes(gp, mean, ymin = mean - sd, ymax = mean + sd),
-#'     colour = 'red',
-#'     width = 0.4
-#'   )
+#'  geom_point(sample_df, aes(x = group, y = value)) +
+#'  geom_point(group_means_df, aes(x = group, y = group_mean), colour = 'red', size = 3)
```
teunbrand, line 75: Here the `data` argument is in the wrong position for both layers.
My suggestion:

``` r
  geom_point(aes(x = group, y = value), data = sample_df) +
  geom_point(
    aes(x = group, y = group_mean),
    data = group_means_df, 
    colour = "red", size = 3
  )
```

R/plot.r
```diff
@@ -36,43 +39,40 @@
...
+#' # can be particularly useful when working with complex, multi-dataset graphics.
 #' ggplot() +
-#'   geom_point(data = df, aes(gp, y)) +
-#'   geom_point(data = ds, aes(gp, mean), colour = 'red', size = 3) +
-#'   geom_errorbar(
-#'     data = ds,
-#'     aes(gp, mean, ymin = mean - sd, ymax = mean + sd),
-#'     colour = 'red',
-#'     width = 0.4
-#'   )
```
teunbrand, line 75: It is not very clear to me why the errorbar was taken out of this example.

teunbrand (changes requested):
Thanks, I appreciate the effort!
However, there are some issues with the examples for which I've left suggestions. I apologise if they're nit-picky. If you have any questions, please feel free to ask.

R/plot.r
```diff
@@ -36,43 +39,40 @@
...
-#'   y = rnorm(30)
+#' sample_df <- data.frame(
+#'   group = factor(rep(letters[1:3], each = 10)),
+#'   value = rnorm(30)
 #' )
-#' ds <- do.call(rbind, lapply(split(df, df$gp), function(d) {
-#'   data.frame(mean = mean(d$y), sd = sd(d$y), gp = d$gp)
+#' group_means_df <- do.call(rbind, lapply(split(sample_df, sample_df$group), function(d) {
+#'   data.frame(group_mean = mean(d$value), group = d$group)
 #' }))
```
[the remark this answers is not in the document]
kevinxperese: Will change this, thanks.

R/plot.r
kevinxperese: My bad. I was confused by the comment:
```Note that we don't need to supply `data` or `mapping` in each layer because the defaults from ggplot() are used.```, and further confused by the ordering of the `data` and `mapping` arguments being reversed in the `ggplot()` and `geom_point()` functions. And I should have tested my changes!

Does this comment work:
```Note that we don't need to supply `mapping` or `data` arguments in the first geom_point() layer because those arguments are passed along from the arguments supplied in the ggplot() call.```

Also, how do you feel about having explicit `mapping =` keywords in the appropriate places for the `ggplot()` and `geom_point()` calls? I know both `mapping =` and `data =` are often left out in practice, but as a person learning these patterns for the first time (who I think are the most likely readers of this documentation), it might be helpful to have them in there. (At least, they're helpful to me.)

File: R/plot.r (hunk 4 above)
Comment (kevinxperese) [reply R/plot.r:75]: It felt to me that the point of the examples was to show three different patterns for constructing the same graphic, and I thought the the error bars made that more confusing than illuminating. 

Would a comment above the three examples like this be helpful to further clarify the intent?

    # The following three code blocks create the same graphic, each using a different
    # pattern. In each graphic, the group means data frame is used to plot larger red
    # points on top of the sample data.
    
Or perhaps that wasn't the over-arching point. I can also now see that the error bars are meant to be an example of a "complex, multi-dataset graphic." But maybe showing the simpler versions of the three patterns would be easier for new learners?

File: R/plot.r (hunk 1 above)
Comment (teunbrand) [reply R/plot.r:60]: > Also, how do you feel about having explicit mapping = keywords

Yeah I think that is a good idea. Especially when it is the first time the argument is shown in a new position.

> To help clarify, does this comment work?

I think the word 'arguments' is perhaps used 1 too many times, but I like the clarification in general.
Perhaps something like this?

```
Note that we don't need to supply `mapping` or `data` arguments in the first geom_point() layer
 because those arguments are passed along from the main `ggplot()` call`
```

File: R/plot.r (hunk 4 above)
Comment (teunbrand) [reply R/plot.r:75]: I don't think this comment is necessary because if people are executing the examples, it should be relatively straightforward to see how the plot relates to their input data. 

> Or perhaps that wasn't the over-arching point. I can also now see that the error bars are meant to be an example of a "complex, multi-dataset graphic." But maybe showing the simpler versions of the three patterns would be easier for new learners?

Yeah I like the clarity of the 3 patterns in these proposed changes and I agree that the patterns should be consistent. Perhaps a fourth example can be serviceable here to show how one could mix and match patterns, which may look like this:

```r
# When composing multi-dataset plots, it may be convenient to set the most
# frequently used dataset, along with shared aesthetics, in the main `ggplot()`
# function.
ggplot(group_means_df, aes(x = group)) +
 geom_point(aes(y = value), data = sample_df) +
 geom_point(aes(y = group_mean), colour = 'red', size = 3) +
 geom_errorbar(aes(ymin = group_mean - sd, ymax = group_mean + sd),
               colour = "red", width = 0.4)
```
(assuming that `sd` is also calculated)

R/plot.r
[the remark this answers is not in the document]
kevinxperese: Actually, any issues with simplifying the code to create the `group_means_df` to this:

    group_means_df <- sample_df %>%
      group_by(group) %>%
      summarise(group_mean = mean(value))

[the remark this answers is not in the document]
teunbrand: I agree that is more readable and simpler (and exactly why dplyr is convenient), though ggplot2 doesn't import magrittr or dplyr, so we can't use these in examples. The base R pipe is also too new, because it doesn't work on some older R versions which ggplot2 still supports (R 3.5.3). If we don't care about compute an `sd` column, we could also use:

``` r
group_means_df <- setNames(
  aggregate(value ~ group, sample_df, mean), 
  c("group", "group_mean")
)
```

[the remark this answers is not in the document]
kevinxperese: Makes sense, and I like your simplified version of the code.
My vote (which I realize carries a fairly low weight) would be to omit the `sd` calculation and the fourth example plot.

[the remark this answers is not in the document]
teunbrand: I'm fine with omitting the 4th example/sd calculation. My comment earlier about the error bars was inquiring, not prescribing.

R/plot.r
```diff
@@ -36,42 +45,55 @@
...
-#' # Note how the x and y aesthetics must now be declared in
-#' # each geom_point() layer.
-#' ggplot(df) +
-#'   geom_point(aes(gp, y)) +
-#'   geom_point(data = ds, aes(gp, mean), colour = 'red', size = 3)
-#'
-#' # Alternatively we can fully specify the plot in each layer. This
-#' # is not useful here, but can be more clear when working with complex
-#' # mult-dataset graphics
+#'   geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
```
teunbrand, line 73: I think it is a feature of the pattern to omit `x = group` here,  because that aesthetic is inherited from the `ggplot()` call.

R/plot.r
```diff
@@ -36,42 +45,55 @@
...
+#'              colour = 'red', size = 3
+#'     )
```
teunbrand, line 75: I apologise profusely for being nit-picky about style here, but might I be able to convince you to indent the second point layer as follows according to the style guide's direction for [long lines](https://style.tidyverse.org/syntax.html#long-lines)? I think `mapping` and `data` can go on the same line because they're related. Similarly, `colour = 'red', size = 3` are both static aesthetics.

```r
  geom_point(
    mapping = ..., data = ...,
    colour = ..., size = ...
  )
```
This applies to the examples in pattern 2 and 3 as well.

kevinxperese: I did give this some thought.

While I'm all for style guides, I do think that an 80 character line length is too arbitrary of a criteria to be strictly enforced.

That said, I kept the line as it currently is for 2 reasons:
1. The line is 77 char without the leading "#' ", and exactly 80 with, 
2. I felt that the parallelism across the layers and across the 3 patterns made it much easier to see the differences in the patterns.

That is, I think this:
```
ggplot(data = sample_df) +
  geom_point(mapping = aes(x = group, y = value)) +
  geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
             colour = 'red', size = 3
  )
```
is easier to read than this:
```
ggplot(data = sample_df) +
  geom_point(mapping = aes(x = group, y = value)) +
  geom_point(
    mapping = aes(x = group, y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )
```

And this:
```
ggplot() +
  geom_point(mapping = aes(x = group, y = value), data = sample_df) +
  geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
             colour = 'red', size = 3
  )
```
is easier to read than this:
```
ggplot() +
  geom_point(
    mapping = aes(x = group, y = value), data = sample_df
  ) +
  geom_point(
    mapping = aes(x = group, y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )
```

If those don't sway you, though, I'll indent them as you suggest.

R/plot.r
kevinxperese: I see your point, but I was putting a lot of emphasis on the parallelism across the examples. 
Could we add a comment to the effect that the `x = group` could also be omitted because the it would be passed along from the `ggplot()` call?

```
# Pattern 1
# Both the `data` and the `mapping` arguments are passed into the `ggplot()`
# call. Note that we don't need to supply `mapping` or `data` arguments in
# the first `geom_point()` layer because those arguments get passed along
# from the `ggplot()` call. Note also that the `x = group` argument in the 
# second `geom_point()` could also be omitted because it would also be 
# passed along from the `ggplot() call.
ggplot(data = sample_df, mapping = aes(x = group, y = value)) +
  geom_point() +
  geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
             colour = 'red', size = 3
  )
```

R/plot.r
teunbrand: I sympathise with your arguments: I too feel that a style guide is too restrictive at times and do agree that the parallelism has merit. The reason I suggested it is not because the style guide recommends it (though it is easy to point to), which has some value but shouldn't be followed blindly. Instead, I suggested it because it would be more internally consistent with various other examples in other doc pages, which I do feel is somewhat important. Perhaps we could get the opinion of a third person to decide for us. 

@thomasp85
TL;DR: do you have any opinion stylewise on the following two alternatives, or a suggestion of your own?

```r
#            Nice alignment of arguments
ggplot() + # V         V   V          V
  geom_point(mapping = aes(x = group, y = value), data = sample_df) +
  geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
             colour = 'red', size = 3
  )
```

versus

```r
# Dogmatically consistent
ggplot() +
  geom_point(mapping = aes(x = group, y = value), data = sample_df) +
  geom_point(
    mapping = aes(x = group, y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )
```

R/plot.r
teunbrand: I think it might be better to omit `x = group` here, and then note in the comment that we're allowed to do so because it has a global mapping.

kevinxperese: How about this:

    # Pattern 1
    # Both the `data` and the `mapping` arguments are passed into the `ggplot()`
    # call. Note that we don't need to supply `mapping` or `data` arguments in
    # the first `geom_point()` layer because those arguments get passed along
    # from the `ggplot()` call.
    ggplot(data = sample_df, mapping = aes(x = group, y = value)) +
      geom_point()
    
    # Pattern 2
    # Same plot as above, passing in only the `data` argument in the `ggplot()`
    # call. Note how we must now pass the `mapping` argument in for the
    # `geom_point()` layer because it is no longer passed along from the 
    # `ggplot()` call.
    ggplot(data = sample_df) +
      geom_point(mapping = aes(x = group, y = value))
    
    # Pattern 3
    # Same plot as above, but with larger red points on top of the sample data
    # to show the mean values. Here, we are not passing either the `data` or
    # `mapping` arguments into the `ggplot()` call. We must then fully specify 
    # those arguments in each layer. This pattern can be particularly useful
    # when creating more complex, multi-data frame graphics.
    ggplot() +
      geom_point(mapping = aes(x = group, y = value), data = sample_df) +
      geom_point(mapping = aes(x = group, y = group_mean), data = group_means_df,
                 colour = 'red', size = 3
      )

This removes the second layer from the first two patterns, which maybe helps focus reader on the differences in the patterns. The third pattern then adds in the additional layer with the red points for the mean values, which is also in the direction of showing the pattern being used for a "more complex, multi-data frame" graphic.  

The outstanding issue, then (I think) is whether the second layer in Pattern 3 should have the code wrap or have a line length pushing the 80 char limit but show better alignment across the arguments.

R/plot.r
thomasp85: The argument for consistency in style across examples weigh the heaviest IMO

R/plot.r
teunbrand: I can see how that simplifies things, but maybe a bit too much as the body text in the details sections talks about using pattern 1 for adding a layer using data from another data frame. I think it is important to show this. Let me suggest the following:

```r
# Pattern 1
# Both the `data` and the `mapping` arguments are passed into the `ggplot()`
# call. The`mapping` or `data` arguments are omitted in the first `geom_point()` 
# layer because those arguments get passed along from the `ggplot()` call. 
# Note that the the second point layer re-uses the `x = group` mapping through
# that mechanism, but overrides the y-aesthetic.
ggplot(data = sample_df, mapping = aes(x = group, y = value)) +
  geom_point() +
  geom_point(
    mapping = aes(y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )

# Pattern 2
# Same plot as above, passing in only the `data` argument in the `ggplot()`
# call. The `mapping` arguments are now required for each
# `geom_point()` layer because there is no global `mapping` argument passed
# along from the `ggplot()` call.
ggplot(data = sample_df) +
  geom_point(mapping = aes(x = group, y = value)) +
  geom_point(
    mapping = aes(x = group, y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )

# Pattern 3
# Same plot as above, passing in neither the `data` or `mapping` arguments
# to the `ggplot()` call. Both those arguments must then be fully specified
# in each layer. While not useful here, this pattern can be useful when 
# creating complex graphics with multiple datasets.
ggplot() +
  geom_point(mapping = aes(x = group, y = value), data = sample_df) +
  geom_point(
    mapping = aes(x = group, y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )
```

teunbrand:
Come to think of it, there is also a 4th pattern that is useful in some circumstances. That is to put a partial mapping in `ggplot()` and complete it in the layers, e.g.:

```r
ggplot(mapping = aes(x = group)) +
  geom_point(mapping = aes(y = value), data = sample_df) +
  geom_point(
    mapping = aes(y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )
```

R/plot.r
kevinxperese: Thanks @thomasp85 and @teunbrand! 

My most recent commit wraps those long lines, and has some edits in the comment blocks for each of the three patterns to attempt to use consistent phrasing across the examples.

kevinxperese:
> I'm not convinced it should be in the docs though.

I think I agree.
The description for `ggplot()` has:

> There are three common patterns used to invoke `ggplot()`

I don't know what the threshold is for "common," though. (My brain went down the rabbit hole of wanting to actually attempt to quantify the distribution of these four patterns!)

teunbrand (approved):
Thanks for all your efforts is clarifying these patterns!

I think there is GitHub magic to automatically propagate these changes to the manual pages, but in case that fails: could I ask you to run `devtools::document()`?

/document

kevinxperese:
Thanks so much, @teunbrand, for iterating on this with me -- I very much appreciate ALL your work on this project!!!

teunbrand:
No thank *you* for contributing: I think the patterns are much clearer now, so that will be helpful for people who're trying to learn ggplot2!

PR #5139, opened 2023-01-07
Try coerce ordered in `vec_rbind0()`

### LINKED ISSUE #5109: Can't combine ordered factors in a facet_grid
When I use two ordered factors in a `facet_grid`, with `as.table = FALSE` in an attempt to get the plot to arrange the facets in a conventional increasing left-to-right, bottom-to-top grid, I get a warning and an error 
```
Error:
! Can't combine `..1$B` <factor<0715c>> and `..2$B` <ordered<08921>>.
Warning message:
Combining variables of class <ordered> and <ordered> was deprecated in ggplot2 3.4.0.
ℹ Please ensure your variables are compatible before plotting (location: `join_keys()`)
This warning is displayed once every 8 hours.
Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated. ```
```
Here's a reprex
```r
library(ggplot2)
library(tidyr)

data <- data.frame(
  x = rnorm(100), y = rnorm(100),
  A = factor(sample(c("tall", "grande", "venti"), 100, replace = TRUE), 
             levels = c("tall", "grande", "venti"), ordered = TRUE),
  B = factor(sample(c("L", "M", "S"), 100, replace = TRUE),
             levels = c("S", "M", "L"), ordered = TRUE))

ggplot(data) + 
  geom_point(aes(x = x, y = y)) +
  facet_grid(B ~ A, as.table = TRUE)

# FAILS
ggplot(data) + 
  geom_point(aes(x = x, y = y)) +
  facet_grid(B ~ A, as.table = FALSE)
```

This seems like a regression in this particular context, since it gives a lot less control over the appearance of the arrangement of the grid of small multiples, even when I am 'doing the right thing' and ordering my factors to get specific effects. Perhaps it was introduced to fix some other problem, but I can't see what it accomplishes in this context.

The error message seems to be from the function `utilities::with_ordered_restart` here 

https://github.com/tidyverse/ggplot2/blob/63125db184b6e7d5516007574b5d15acbfd8f470/R/utilities.r#L664

 but it's not clear to me why it should be treated as an error in what is surely a common use-case.

Possibly related to #4180

teunbrand:
This PR aims to fix #5109

Briefly, it tries to cast two ordered factors to plain factors (which can be combined in vctrs). This used to throw an error when in facetted plots, you set two ordered factors as facetting variables. This now becomes a warning.

I'd really appreciate a second pair of eyes here, as I'm still somewhat confused about the order of events in trying to restart upon errors. The thing that makes me a bit uncomfortable, is that with this PR, we're only throwing warnings on `vec_ptype2()` errors, and no longer on `vec_cast()` errors (this should just throw the usual vctrs error, I think). I don't know if we're missing an important case when the warning should be thrown in `vec_cast()`.

To quickly recap, you weren't allowed to combine two ordered factors with incompatible levels, even though we should be able to. You also aren't allowed to combine factors and dates, which should be the case.

``` r
# For reprex purposes we want to show all warnings

# Legal combination
ggplot2:::vec_rbind0(
  data.frame(a = factor(c("A", "B"), ordered = TRUE)), 
  data.frame(a = factor(c("B", "C"), ordered = TRUE))
)
#> Warning: Combining variables of class <ordered> and <ordered> was deprecated in ggplot2
#> 3.4.0.
#> ℹ Please ensure your variables are compatible before plotting (location: )
#> Error:
#> ! Can't combine `..1$a` <factor<a022a>> and `..2$a` <ordered<fe399>>.

# Illegal combination
ggplot2:::vec_rbind0(
  data.frame(a = factor(c("A", "B"), ordered = TRUE)), 
  data.frame(a = Sys.Date() + 0:1)
)
#> Warning: Combining variables of class <ordered> and <Date> was deprecated in ggplot2
#> 3.4.0.
#> ℹ Please ensure your variables are compatible before plotting (location: )
#> Error in `ggplot2:::vec_rbind0()`:
#> ! Can't combine `..1$a` <factor<a022a>> and `..2$a` <date>.
```

With this PR, you no longer get an error by combining two incompatible ordered factors, but you do get a warning and the correct result. Combining dates and factors is still prohibited.

``` r
#> ℹ Loading ggplot2

# Legal combination
vec_rbind0(
  data_frame0(a = factor(c("A", "B"), ordered = TRUE)), 
  data_frame0(a = factor(c("B", "C"), ordered = TRUE))
)
#> Warning: Combining variables of class <ordered> and <ordered> was deprecated in ggplot2
#> 3.4.0.
#> ℹ Please ensure your variables are compatible before plotting (location: )
#>   a
#> 1 A
#> 2 B
#> 3 B
#> 4 C

# Illegal combination
vec_rbind0(
  data_frame0(a = factor(c("A", "B"), ordered = TRUE)), 
  data_frame0(a = Sys.Date() + 0:1)
)
#> Warning: Combining variables of class <ordered> and <Date> was deprecated in ggplot2
#> 3.4.0.
#> ℹ Please ensure your variables are compatible before plotting (location: )
#> Error in `vec_rbind0()`:
#> ! Can't combine `..1$a` <factor<a022a>> and `..2$a` <date>.
```

<sup>Created on 2023-01-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/utilities.r
```diff
@@ -668,11 +669,13 @@ with_ordered_restart <- function(expr, .call) {
         ")"
       )
 
-      deprecate_soft0(
-        "3.4.0",
-        I(msg),
-        details = desc
-      )
+      if (inherits(cnd, "vctrs_error_ptype2")) {
```
thomasp85, line 672: Can you explain about the reasoning for this guard? Why is it suddenly necessary?

teunbrand: This was more out of practicality than necessity. Without this, combining two ordered factors would throw 2 warnings, once during `vec_ptype2()` and once during `vec_cast()`. I don't know if there is a situation where `vec_cast()` should warn whereas `vec_ptype2()` shouldn't. See also here: https://github.com/tidyverse/ggplot2/issues/5109#issuecomment-1365831369

thomasp85: hmm... maybe keep track of whether the input is two ordered vectors and only in that case have the guard...

Maybe it is overdoing it but since I can't figure out the repercussions I'd rather be on the safe side

teunbrand: I too feel uncomfortable about this guard. I think the order of events is as follows:

1. `vec_ptype2(x, y)` searches for a common class to coerce `x` and `y` to. Let's call that `w`.
2. `vec_cast(x, w)` coerces `x` to `w`.
3. `vec_cast(y, w)` coerces `y` to `w`

So if `x` and `y` are ordered factors, `w` will be a factor. The case where we only have two ordered factors is step (1). So, step (2) and (3) will still emit a warning about converting ordered to factors. The ideal situation would be to emit a warning at (1) and then silence (2) and (3). However, the 'track whether x and y are both ordered factors'-rule can't be applied in (2) and (3) because `w` is a regular factor.

I'll puzzle around a bit more to see what might be done, but it'll probably become more complicated than just tracking ordered factors.

teunbrand:
Alright, should be safer now, though it has become a little bit more complex than I'd hoped for.

thomasp85:
hmm... I must admit I'm now closer to just accepting a double warning rather than complicating the restart handler in this way - it feels like dangerous over-engineering for a situation that arrises so infrequently. Thoughts?

teunbrand:
Yeah I agree with you that this might be over-complicated for little practical purposes other than displaying a slightly more comprehensible set of warnings. Also, these are lifecycle warnings, so they're only displayed once every 8 hours under default settings. The important bit is turning the original error into a warning so that the plot in #5109 at least renders (which is an easy fix).

If the warnings become a pertinent problem, which seems unlikely, we could always decide to fix these at some later point.

thomasp85:
Agreed - let's roll back to simply avoiding the error

teunbrand:
Alright, I've reverted everything but the error avoidance

thomasp85 (approved):
LGTM

thomasp85:
But please add a unit test and NEWS bullet

PR #5141, opened 2023-01-09
Update geom_dotplot.Rd

jdonland:
Correct duplicated "x" in descriptions of x and y in Computed variables section.


teunbrand:
Hi Jesse, I'm sorry for taking such a while to get back to you. Ideally, changes in documentation should be made in the `.R` file, which are then propagated to `.Rd` files by the {roxygen2} package. You might have encountered the following in the file you've edited:

https://github.com/tidyverse/ggplot2/blob/7f32d55fc7cc377cd5fab833c9ba0cf7b5dad256/man/geom_dotplot.Rd#L1-L2

Can I ask you to make the change in the `.R` file instead and regenerate the documentation from these?

jdonland:
Surely it will be quicker for someone who already has roxygen set up to do this.

PR #5143, opened 2023-01-10
Named shapes "dot" and "blank"

### LINKED ISSUE #5088: Allow point shapes to be mixtures of names and symbols
Point shapes can be specified by supplying numerical codes. For example:

```r
# Works, but shape is cryptic:
ggplot(tibble(x = c('.', 'circle'), 
              shape = c(46, 19))) + 
  geom_point(aes(x, x, shape = shape)) + 
  scale_shape_identity() 
```

It would nice to specify the shape using a name as follows:
```r
# Uses "c" as shape for second point:
ggplot(tibble(x = c('.','circle'))) + 
  geom_point(aes(x, x, shape = x)) + 
  scale_shape_identity() 

# Generates error:
ggplot(tibble(x = c('circle','.'))) + 
  geom_point(aes(x, x, shape = x)) + 
  scale_shape_identity()
```

However, the first snippet treats "circle" as "c" (and hence uses the letter "c" as the plot shape) while the second snippet generates an error.

The problem is that the plotting code (https://github.com/tidyverse/ggplot2/blob/main/R/geom-point.r#L146) wants the entire vector of shapes to be either single latter symbols or longer labels, as opposed to a mixture. Furthermore, it decides which is the case based on the first item in the vector of shapes. Would it be possible to look up shape names in `pch_table` for those names with `nchars() > 1` and convert the rest to their ascii code? (I suppose there may be issues with non-ascii but single letter shapes.)

teunbrand:
This PR implements the suggestion in https://github.com/tidyverse/ggplot2/issues/5088#issuecomment-1364554410 to fix #5088.

Briefly, it adds `"dot"` and `"blank"` as valid names for shapes in `geom_point()` that get translated to `shape = "."` and `shape = NA` respectively. The specs vignette was also updated to reflect this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Added `"dot"` and `"blank"` to list of named shapes, representing 
+  `shape = "."` and `shape = NA` respectively (@teunbrand, #5088).
 * Using two ordered factors as facetting variables in 
   `facet_grid(..., as.table = FALSE)` now throws a warning instead of an
   error (@teunbrand, #5109).
--- a/R/geom-point.r
+++ b/R/geom-point.r
@@ -176,7 +176,9 @@ translate_shape_string <- function(shape_string) {
     "square filled"         = 22,
     "diamond filled"        = 23,
     "triangle filled"       = 24,
-    "triangle down filled"  = 25
+    "triangle down filled"  = 25,
+    "dot"                   = 46,
+    "blank"                 = NA
   )
 
   shape_match <- charmatch(shape_string, names(pch_table))
--- a/vignettes/ggplot2-specs.Rmd
+++ b/vignettes/ggplot2-specs.Rmd
@@ -172,13 +172,13 @@ Shapes take five types of values:
       "diamond", paste("diamond", c("open", "filled", "plus")),
       "triangle", paste("triangle", c("open", "filled", "square")),
       paste("triangle down", c("open", "filled")),
-      "plus", "cross", "asterisk"
+      "plus", "cross", "asterisk", "dot", "blank"
     )
     
     shapes <- data.frame(
       shape_names = shape_names,
-      x = c(1:7, 1:6, 1:3, 5, 1:3, 6, 2:3, 1:3),
-      y = -rep(1:6, c(7, 6, 4, 4, 2, 3))
+      x = c(1:7, 1:6, 1:3, 5, 1:3, 6, 2:3, 1:5),
+      y = -rep(1:6, c(7, 6, 4, 4, 2, 5))
     )
     
     ggplot(shapes, aes(x, y)) +


yutannihilation:
I personally against adding `dot`. Since `"."` doesn't support `size`, I think this is outside of "grammar of graphics." 

``` r
library(ggplot2)

d <- data.frame(x = 1:3)

ggplot(d, aes(x, size = 3^x)) +
  geom_point(aes(y = "circle"), shape = "circle") +
  geom_point(aes(y = "."), shape = ".")
```



<sup>Created on 2023-02-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

For `blank`, while there's a blank linetype (`0`) for `lty`, there's no blank point type for `pch`. Maybe we should just recommend using `NA` directly?

teunbrand:
Thanks for your thoughts!

> I personally against adding dot. Since "." doesn't support size, I think this is outside of "grammar of graphics."

I agree it doesn't really play well with the grammar of graphics. However, there currently aren't any prohibitions on specifying `shape = "."`, so the question then becomes whether it should be recognised as a named shape or not.

> For blank, while there's a blank linetype (0) for lty, there's no blank point type for pch. Maybe we should just recommend using NA directly?

The documentation of `?points` mentions that `NA_integer` is reserved for plotting no symbol. Using `NA` directly drops the data in the `GeomPoint$handle_na` method with a warning. My thought is that if you go through the trouble of specifying `shape = "blank"`, your intent is to hide the symbols and thus don't need the warning. I think the warning is mostly there to alert the user against unintentional dropping of data.

Aside from that, you cannot currently mix valid names with NAs (backtraces omitted):

``` r
ggplot2:::translate_shape_string(c("circle", NA_character_))
#> Error in `ggplot2:::translate_shape_string()`:
#> ! Shape aesthetic contains invalid value: NA

# This following is particularly uninformative
ggplot2:::translate_shape_string(c(NA_character_, "circle"))
#> Error in if (nchar(shape_string[1]) <= 1) {: missing value where TRUE/FALSE needed
```

<sup>Created on 2023-02-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

yutannihilation:
A bit off-topic, but I think ggplot2 should have prohibited any use of these single characters as a shape. We can use arbitrary letter here, but I feel this isn't a "point" geom anymore. That said, even so, I guess `shape = 46` should remain valid, so I have no idea what's the right API would be. I'd like to wait for other maintainers' comments.

``` r
library(ggplot2)

  geom_point(shape = "龍", size = 30, alpha = 0.3)
```



<sup>Created on 2023-02-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

> The documentation of ?points mentions that `NA_integer` is reserved for plotting no symbol.

Yeah, that's the point, but I think

> you cannot currently mix valid names with NAs

is not right. From the users' perspective, `NA`s can be passed and get removed before `translate_shape_string()`. Honestly, I don't use `linetype = "blank"` etc., so I might be biased.

``` r
library(ggplot2)

d <- data.frame(
  id = c("non-missing", "missing"),
  shape = c("circle", NA)
)

ggplot(d, aes(id, y = 0L, shape = shape)) +
  geom_point(size = 30) +
  scale_shape_identity()
#> Warning: Removed 1 rows containing missing values (`geom_point()`).
```



<sup>Created on 2023-02-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85:
I too am against adding "dot"... I think a pixel-sized spot has value but not as part of any scaling etc so I think its main use is when setting shape directly rather than mapping it - and in these instances `shape = "."` is just as informative IMO

teunbrand:
Right, but all the arguments made for "dot", also apply to "blank", so I think it'd be better to close this PR then, no?

thomasp85:
yeah, I think you can close it

PR #5144, opened 2023-01-10
ScalesList methods

### LINKED ISSUE #1310: ScalesList-related functions should be moved into ggproto object
There are some functions that operate on `ScalesList` objects. They should be turned into methods.

teunbrand:
This PR aims to fix #1310.

In brief, ye olde issue suggested to convert the `scales_*()` functions to ggproto methods of the `ScalesList` object, which is what this PR does.

I think it has 2 benefits:

* The `ScalesList$scales` don't need to be passed around, making such calls a little bit more terse.
* For consistency purposes this is nice, as the upcoming ggproto'isation of the guides follows a similar pattern.


thomasp85:
blast from the past :-)

thomasp85 (approved):
LGTM - can you confirm that no implementations have been changed (the diff is pretty messed up so I don't fully trust my eye)

teunbrand:
Other than some style things, the only thing I changed is `has_default_transform()` check that was in the `$transform_df()` method and applied it to the `$backtransform_df()` method as well. 

I don't know if it is needed, but it would make the two methods more parallel. (as a reminder, you added this in 101c3a66144e92f924a0dc26b4bafd95c230c370). Happy to revert if you didn't apply it to de backtransform for good reasons.

thomasp85:
that was most likely an oversight by me... I also didn't add any unit tests for it by the looks of it - could I get you to add such a test to make sure we don't end there again

teunbrand:
I'm not quite sure what prompted the `has_default_transform()` check. I've guessed it would be custom `trans_new()` or `Scale$transform()`, but I can't trigger the code to fail if I leave out the check, so I'm unsure what to test for.

teunbrand:
I'm merging this. If we want to add a test for the `has_default_transform()` logic, could you open up a new issue with a little bit more information?

PR #5146, opened 2023-01-11
Closes #5145

### LINKED ISSUE #5145: `update_geom_defaults()` should class results
Using `update_geom_defaults()` and `update_stat_defaults()` don't class the updated list, so it prints differently. Example below. Will submit simple PR to fix presently.

``` r
library("ggplot2")

GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `shape`  -> 19
#> * `colour` -> "black"
#> * `size`   -> 1.5
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5

str(GeomPoint$default_aes)
#> List of 6
#>  $ shape : num 19
#>  $ colour: chr "black"
#>  $ size  : num 1.5
#>  $ fill  : logi NA
#>  $ alpha : logi NA
#>  $ stroke: num 0.5
#>  - attr(*, "class")= chr "uneval"

update_geom_defaults("point", list(size = 4))

GeomPoint$default_aes
#> $size
#> [1] 4
#> 
#> $shape
#> [1] 19
#> 
#> $colour
#> [1] "black"
#> 
#> $fill
#> [1] NA
#> 
#> $alpha
#> [1] NA
#> 
#> $stroke
#> [1] 0.5

class(GeomPoint$default_aes) <- "uneval"
GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `size`   -> 4
#> * `shape`  -> 19
#> * `colour` -> "black"
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5
```

<sup>Created on 2023-01-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

dkahle:


R/geom-defaults.r
```diff
@@ -15,6 +15,7 @@ update_geom_defaults <- function(geom, new) {
   g <- check_subclass(geom, "Geom", env = parent.frame())
   old <- g$default_aes
   g$default_aes <- defaults(rename_aes(new), old)
+  class(g$default_aes) <- "uneval"
```
teunbrand, line 18: Perhaps the following is more foolproof if, for whatever reason, somebody creates a geom where the defaults don't follow conventions.
```r
g$default_aes[] <- defaults(rename_aes(new), old)
```
This runs the result through `ggplot2:::new_aes()` whenever `g$default_aes` is a proper mapping with the uneval class. You could then leave out the manual class assignment.

This also applies to the stat variant.

teunbrand (changes requested):
Thanks for preparing a PR!

I've left a suggestion that I think will simplify the issue even further.

R/geom-defaults.r
teunbrand: I like the idea of preserving the order is nice. However, I think there are some niche but legitimate use-cases for defining aesthetics outside the already available aesthetics. Using the `[names(old)]` would ensure that such aesthetics are dropped, so I would be more comfortable if you could revert that particular change.

dkahle: The problem is that the other strategy doesn't actually work because it mixes aesthetics:

``` r
library("ggplot2") # using g$default_aes[] <- defaults(rename_aes(new), old)

GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `shape`  -> 19
#> * `colour` -> "black"
#> * `size`   -> 1.5
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5

update_geom_defaults("point", list(color = "red"))

GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `shape`  -> "red"
#> * `colour` -> 19
#> * `size`   -> 1.5
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5

ggplot(mtcars, aes(disp, mpg)) + geom_point()
#> Error in `geom_point()`:
#> Caused by error in `translate_shape_string()` at ggplot2/R/geom-point.r:122:6:
#> ! Shape aesthetic contains invalid value: "red"
```

<sup>Created on 2023-01-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

For defining aesthetics outside the already available aesthetics, I think the `g$default_aes[] <- ` will bug then regardless of how we do the naming, right?, because there will be more new default aesthetics than the old list holds. Does that mean back to the `class() <- "uneval"`? (Possibly after reordering, and appending new aesthetics?)

teunbrand: Good catch! Yeah the uneval class method could work, but that also wouldn't preserve the order I think. We might also make the renaming of `new` a separate step, then we could use those to preserve order. Something like the following:

``` r
# Path might differ on your machine, but this is to use all the
# internals
#> ℹ Loading ggplot2

update_geom_defaults2 <- function(geom, new) {
  g <- check_subclass(geom, "Geom", env = parent.frame())
  old <- g$default_aes
  new <- rename_aes(new)
  new <- defaults(new, old)[union(names(old), names(new))]
  g$default_aes[names(new)] <- new
  invisible()
}

GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `shape`  -> 19
#> * `colour` -> "black"
#> * `size`   -> 1.5
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5

update_geom_defaults2("point", list(color = "red"))

GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `shape`  -> 19
#> * `colour` -> "red"
#> * `size`   -> 1.5
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5

update_geom_defaults2("point", list(foo = "bar"))

GeomPoint$default_aes
#> Aesthetic mapping: 
#> * `shape`  -> 19
#> * `colour` -> "red"
#> * `size`   -> 1.5
#> * `fill`   -> NA
#> * `alpha`  -> NA
#> * `stroke` -> 0.5
#> * `foo`    -> "bar"
```

<sup>Created on 2023-01-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand: I'm fine with that as well, but I think the difference in readability is pretty small.

teunbrand:
Would you mind a few small unit test to `tests/testthat/test-geom-.R` that tests for the behaviour you care about (class, ordering)? It is mostly so that we don't accidentally break that behaviour in the future.

dkahle:
Sure, give that a look. I also added a few examples to the docs, but I know time is at a premium for the graphics part. If we need to take that out, no objection.

R/geom-defaults.r
```diff
@@ -7,14 +7,42 @@
...
 #' ggplot(mtcars, aes(mpg, wt)) + geom_point()
+#'
+#' # reset default
 #' update_geom_defaults("point", list(colour = "black"))
+#'
+#' # note: adding default aesthetics is possible
+#' # update_geom_defaults("point", list(foo = "bar"))
+#' # GeomPoint$default_aes
+#'
+#'
```
teunbrand, line 25: I think it is fine to omit this, this usage isn't advertised as it is easy to get yourself in trouble if you don't know what you're doing and end up breaking a geom.

R/geom-defaults.r
```diff
@@ -7,14 +7,42 @@
...
+#' # updating a stat's default aesthetic settings
+#' # example: change stat_bin()'s default y-axis to the density scale
+#' StatBin$default_aes
+#' update_stat_defaults("bin", list(y = expr(after_stat(density))))
+#' StatBin$default_aes
+#' ggplot(data.frame(x = rnorm(1e3)), aes(x)) +
+#'   geom_histogram() +
+#'   geom_function(fun = dnorm, color = "red")
```
teunbrand, line 33: I like this example as I've always wondered myself what `update_stat_defaults()` might be useful for. I was surprised this doesn't work on the CRAN version, so that is an extra bonus. If you want to get rid of the `expr()` you could also just use `update_stat_defaults("bin", aes(y = after_stat(density)))`, which might be somewhat more explicit.

teunbrand:
Thanks, the update looks great! I only have two minor comments. After that I think this PR is good to go

R/geom-defaults.r
dkahle: That doesn't quite work:
``` r
# without expr()
update_stat_defaults("bin", list(y = after_stat(density)))
StatBin$default_aes
#> Aesthetic mapping: 
#> * `x`      -> `after_stat(count)`
#> * `y`      -> `function (x, ...) 
#> UseMethod("density")`
#> * `weight` -> 1

StatBin$default_aes |> str()
#> List of 3
#>  $ x     : language ~after_stat(count)
#>   ..- attr(*, ".Environment")=<environment: namespace:ggplot2> 
#>  $ y     :function (x, ...)  
#>  $ weight: num 1
#>  - attr(*, "class")= chr "uneval"

ggplot(data.frame(x = rnorm(1e3)), aes(x)) +
  geom_histogram() +
  geom_function(fun = dnorm, color = "red")
#> Error in `strip_dots()` at ggplot2/R/aes-evaluation.r:330:4:
#> ! Unknown input: <function>
```

Interestingly, the environment of the promise is wrong with my `expr()` version, but it works:
``` r
# with expr()
update_stat_defaults("bin", list(y = expr(after_stat(density))))
StatBin$default_aes
#> Aesthetic mapping: 
#> * `x`      -> `after_stat(count)`
#> * `y`      -> `after_stat(density)`
#> * `weight` -> 1

StatBin$default_aes |> str()
#> List of 3
#>  $ x     : language ~after_stat(count)
#>   ..- attr(*, ".Environment")=<environment: namespace:ggplot2> 
#>  $ y     : language ~after_stat(density)
#>   ..- attr(*, ".Environment")=<environment: R_GlobalEnv> 
#>  $ weight: num 1
#>  - attr(*, "class")= chr "uneval"

ggplot(data.frame(x = rnorm(1e3)), aes(x)) +
  geom_histogram() +
  geom_function(fun = dnorm, color = "red")
```




The environment attached to the promise seems not to matter. Example:
``` r
density <- 0
after_stat(density) # evaluated in .GlobalEnv
#> [1] 0

ggplot(data.frame(x = rnorm(1e3)), aes(x)) +
  geom_histogram() +
  geom_function(fun = dnorm, color = "red")
```



My takeaway:
1. Just writing `after_stat(density)`, without `expr()`, evaluates it in `.GlobalEnv`. If it's not set, `density` defaults to `stats::density`, the function itself, which of course breaks.
2. With `expr()` passes the expression along, but with the wrong environment. (It's dynamically using the environment in which the assignment is made.) But clearly **ggplot2** ignores that environment, so it works. 

Thoughts?

teunbrand: Sorry I have failed to attract attention to the salient bit: instead of using `list()`, one can use `aes()` to capture the expression with the environment, and that should work fine as input for the updater.

dkahle: Oh, wow! Yep, totally missed that part! Pushing now...

teunbrand (approved):
Thanks so much for contributing, looks great now!

dkahle:
Thanks @teunbrand !

PR #5150, opened 2023-01-14
Update `coord_flip()` docs

### LINKED ISSUE #5130: Updates to `coord_flip()` docs
I think we want to mention how to flip individual geoms easily, and say that `coord_flip()` is no longer needed for most purposes. (I'm not sure we should go as far as superseding it, but maybe?)

teunbrand:
This PR aims to fix #5130.

In brief, I tried to make it clear that swapping aesthetics is the preferred method.

* The description mentions easily swapping x/y aesthetics 
* Every example is preceded by the same plot using swapped aesthetics. 

I was unsure what goes into a decision to supersede a function, so I defaulted to not tag it with a superseded badge.

### DIFF
--- a/R/coord-flip.r
+++ b/R/coord-flip.r
@@ -1,30 +1,44 @@
 #' Cartesian coordinates with x and y flipped
 #'
-#' Flip cartesian coordinates so that horizontal becomes vertical, and
-#' vertical, horizontal. This is primarily useful for converting geoms and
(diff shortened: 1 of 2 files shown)


thomasp85:
I think it would make sense to mark it as superseded - thoughts @hadley ?

hadley:
Yeah, that’s fine with me.

PR #5158, opened 2023-01-24
Peformance: vec_unique_count

teunbrand:
I've noticed that `length(unique0(x))` is a common pattern in ggplot2 code. However, up to vectors with 1000 members, `vctrs::vec_unique_count()` seems to be faster (they perform about equal thereafter). For small vectors, the memory footprint is also better and about equal for larger vectors. The improvement is small, but the pattern is common enough that I thought this was warranted to save a few microseconds per plot.

A benchmark:

``` r
#> ℹ Loading ggplot2

lengths <- 2^(0:24)

bms <- lapply(lengths, function(n) {
  x <- sample(LETTERS, n, TRUE)
  bench::mark(
    length(unique0(x)),
    vec_unique_count(x), 
    iterations = 50
  )
})

df <- do.call(rbind, bms)
df$n <- rep(lengths, each = 2)

ggplot(df, aes(n, as.numeric(median), colour = as.character(expression))) +
  geom_line() +
  scale_y_log10() +
  scale_x_log10() +
  labs(
    x = "Length of vector",
    y = "Seconds",
    colour = "Pattern"
  )
```

<sup>Created on 2023-01-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/axis-secondary.R
+++ b/R/axis-secondary.R
@@ -193,7 +193,7 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
     full_range <- self$transform_range(old_range)
 
     # Test for monotonicity
-    if (length(unique0(sign(diff(full_range)))) != 1)
+    if (!is_unique(sign(diff(full_range))))
       cli::cli_abort("Transformation for secondary axes must be monotonic")
   },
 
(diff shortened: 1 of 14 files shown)


yutannihilation (approved):
Looks good.

teunbrand:
Thanks for the review!

PR #3491, opened 2019-08-18
Let geom_sf() remove rows when shape, size or colour is NA

yutannihilation:
Fix #3483

Currently, `GeomSf` doesn't have `non_missing_aes`, but it should be set to avoid errors reported on #3483. I chose `shape`, `size` and `colour` here because

* As `GeomSf` is a mix of points, lines, and polygons, it should use the combination of `non_missing_aes` of `GeomPoint`, `GeomLine` and `GeomPolygon`.
    * `GeomPoint` uses `shape`, `size` and `colour`
    * `GeomLine` and `GeomPolygon` don't have their `non_missing_aes`.
* For lines and polygons, `GeomSf` removes rows (e.g. the size is `NA`) that `GeomLine` and `GeomPolygon` won't. But, the result is the same; lines and polygons whose sizes are all `NA` won't be drawn. So, I think it's no problem to remove the whole rows.

``` r
devtools::load_all("~/repo/R/ggplot2")
#> Loading ggplot2

pts <- sf::st_sf(
  geometry = sf::st_sfc(sf::st_point(0:1), sf::st_point(1:2)),
  size = c(1, NA),
  shape = c("a", NA),
  colour = c("red", NA)
)

ggplot(pts) + geom_sf(aes(size = size))
#> Warning: Removed 1 rows containing missing values (geom_sf).
```

``` r
ggplot(pts) + geom_sf(aes(shape = shape))
#> Warning: Removed 1 rows containing missing values (geom_sf).
```

``` r

# default colour scale maps a colour even to a NA, so identity scale is needed to see if NA is removed
ggplot(pts) + geom_sf(aes(colour = colour)) + scale_colour_identity()
#> Warning: Removed 1 rows containing missing values (geom_sf).
```

<sup>Created on 2019-08-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #5162, opened 2023-01-26
`geom_text()` drops rows with `angle = NA`

### LINKED ISSUE #2757: geom_text throws error when angle = NA
When using `geom_text()` if the angle parameter is set using a variable whose value can be `NA` then `geom_text()` throws an error (`Error in validDetails.text(x) : invalid 'rot' value`) when it encounters the NA value.  Desired behavior would be to handle this gracefully and simply NOT plot the NA value (as would happen with other aesthetics). 

```r
library(ggplot2)
df <- data.frame(x=seq(1,10,1), y=seq(1,10,1), angle=c(seq(0,80,10),NA))

# this works:
ggplot(df[1:9,], aes(x=x, y=y)) + geom_text(label="-", aes(angle=angle))

# this breaks:
ggplot(df, aes(x=x, y=y)) + geom_text(label="-", aes(angle=angle))
# with error:
# Error in validDetails.text(x) : invalid 'rot' value
# also a confusing error because the user did not pass a 'rot' parameter.
# that 'rot' == 'rotation' == 'angle' may not be entirely clear to users.

```

image

teunbrand:
This PR aims to fix #2757.

Briefly, it adds the angle to the non-missing aesthetics, causing the `GeomText$handle_na()` method to drop these observations with missing angles.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -59,7 +59,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          cache-version: 2
+          cache-version: 3
           extra-packages: >
             any::rcmdcheck,
             Hmisc=?ignore-before-r=3.6.0,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* `geom_text()` drops observations where `angle = NA` instead of throwing an
+  error (@teunbrand, #2757).
 * Using two ordered factors as facetting variables in 
   `facet_grid(..., as.table = FALSE)` now throws a warning instead of an
   error (@teunbrand, #5109).
--- a/R/geom-text.r
+++ b/R/geom-text.r
@@ -198,6 +198,8 @@ geom_text <- function(mapping = NULL, data = NULL,
 GeomText <- ggproto("GeomText", Geom,
   required_aes = c("x", "y", "label"),
 
+  non_missing_aes = "angle",
+
   default_aes = aes(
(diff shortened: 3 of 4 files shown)


teunbrand:
Tests fail on R devel due to update to the graphics engine and its interaction with vdiffr.

From the [R-devel/NEWS](https://developer.r-project.org/blosxom.cgi/R-devel/NEWS/2023/01/26) site:

> Thu, 26 Jan 2023
CHANGES IN R-devel GRAPHICS
The graphics engine version, ‘R_GE_version’, has been bumped to ‘16’ and so packages that provide graphics devices should be reinstalled.

> The ‘grDevices’ and ‘grid’ packages have new functions for rendering typeset glyphs, primarily: ‘grDevices::glyphInfo()’ and ‘grid::grid.glyph()’.

 > Rendering of typeset glyphs is only supported so far on the Cairo-based graphics devices and on the ‘pdf()’ device.

yutannihilation:
Good catch. Could you bump the cache-version to clear the packages compiled with an outdated version of R?

https://github.com/tidyverse/ggplot2/blob/15cb2fbc1de1c906fe5264d3cef85c2a9338f375/.github/workflows/R-CMD-check.yaml#L62

teunbrand:
Thanks! That seems to have solved the issue.

yutannihilation (approved):
Looks good! When I did a similar fix https://github.com/tidyverse/ggplot2/pull/3491 on `geom_sf()`, I accidentally broke some valid use cases, but I bet there's no such a downside in `geom_text()`'s case.

teunbrand:
Thanks! I think we should be safe too; it would always throw an error and now it doesn't.

Issue #4020, opened 2020-05-22
Justifying the legend with respect to the full plot area rather than the panel


LiRogers:
Now that the plot titles and captions can be aligned with respect to the full plot area, it would be great if the same was true for the legend. At the moment, if the legend is positioned at the top of the plot below the title it is aligned to the plot panel, making it misaligned with the title:

``` r
library(ggplot2)

ggplot(data = diamonds, mapping = aes(x = carat, y = price, color = cut)) +
  geom_point() +
  labs(title = "A plot title") +
  theme(plot.title.position = "plot",
        legend.position = "top",
        legend.justification = "left")
```

![](https://i.imgur.com/WGxMzJw.png)

<sup>Created on 2020-05-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
  
Experimenting with negative values in `legend.justification = c(x,y)` can get the legend to align with the title, but the value is dependent on the size of the y-axis labels

clauswilke:
Yes, that's a good point. Marking this as TODO for ggplot2 3.4.

teunbrand:
Possibly related, #3989 also proposes more legend alignment options

thomasp85:
Agreed that this would make sense

malcalakovalski:
Would it be possible to do the same for plot.tag.position? If the tag position is set to "topleft" it's also misaligned with the title.

```r 
library(ggplot2)

ggplot(data = diamonds, mapping = aes(x = carat, y = price, color = cut)) +
  geom_point() +
  labs(title = "A plot title",
       tag = 'Figure 1') +
  theme(plot.title.position = "plot",
        plot.tag.position = 'topleft')
```



![diamonds_tag](https://user-images.githubusercontent.com/56268747/128263217-a71ecc47-7f65-4c0a-b0be-3fcbe8d0adb0.png)

clauswilke:
@malcalakovalski Use `plot.tag.position = c(0, 1)`:

```r
library(ggplot2)

ggplot(data = diamonds, mapping = aes(x = carat, y = price, color = cut)) +
  geom_point() +
  labs(
    title = "A plot title",
    tag = "Figure 1"
  ) +
  theme(
    plot.title.position = "plot",
    plot.tag.position = c(0, 1),
    plot.tag = element_text(hjust = 0, vjust = 1)
  )

```
![Uploading Screen Shot 2021-08-04 at 7.08.59 PM.png…]()

[...]

PR #5167, opened 2023-01-30
Tags at panels

### LINKED ISSUE #4297: tag position inside plot borders
Some peer-reviewed journals (e.g. Oecologia) require figure panel labels to be _inside_ of plot borders (although they seem to enforce these requirements inconsistently).  It would be nice if there were an option like `theme(plot.tag.position = "topleftinner")` to place a label at the upper left corner of the inside of plot borders.  Currently, I think I have to kludge this by tweaking `theme(plot.tag = element_text(vjust = , hjust = )`, but it's inconsistent, so I have to do this tweaking for every panel individually.  Maybe there's an easier way that I don't know about?

Related to: https://github.com/thomasp85/patchwork/issues/222

``` r
library(ggplot2)
p1 <- ggplot(mtcars) + geom_point(aes(mpg, disp)) + labs(tag = "a)")
p1 + theme(plot.tag.position = c(0,1),
           plot.tag = element_text(vjust = 1.5, hjust = -2.85))
```

``` r
# left of "a" just touching plot border

p3 <- ggplot(mtcars) + geom_smooth(aes(disp, qsec)) + labs(tag = "b)")
p3 + theme(plot.tag.position = c(0,1),
           plot.tag = element_text(vjust = 1.5, hjust = -2.85))
```

``` r
# "b)" is slightly further in with same theme()
```

<sup>Created on 2020-12-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

teunbrand:
This PR aims to fix #4297.

Briefly, it adds the option to use `plot.tag.position = c({position}, "panel")` and draw the plot tag in the panel. It should also work with the patchwork package.

``` r
#> ℹ Loading ggplot2

  geom_point()

p +
  labs(tag = "A)") +
  theme(plot.tag.position = c("panel", "topright"))
```

``` r

library(patchwork)

(p + p) + plot_annotation(tag_levels = 'A') &
  theme(plot.tag.position = "panel")
```

<sup>Created on 2023-01-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Some side-notes:

* I separated out the tag logic from the already enormous `ggplot_gtable.ggplot_built()` method. I refactored a little bit along the way.
* Instead of adding yet another theme argument, I expanded the use of the already existing one. I think a lot in the theme, related to positions and alignments, is named confusingly, so I didn't want to pile on this already dizzying naming scheme (also nagged about this [here](https://github.com/tidyverse/ggplot2/issues/4020#issuecomment-1386109142)).
* I once again ran into [this bug](https://github.com/tidyverse/ggplot2/pull/4879#issuecomment-1369235237), so I fixed it in the same way.


R/plot-build.r
```diff
@@ -432,3 +359,88 @@ by_layer <- function(f, layers, data, step = NULL) {
   )
   out
 }
+
+# Add the tag element to the gtable
+table_add_tag <- function(table, label, theme) {
+  # Initialise the tag margins
+  table <- gtable_add_padding(table, unit(0, "pt"))
```
teunbrand, line 366: Patchwork for some reason assumes this padding is present regardless of whether it was needed, so I left this in.

R/plot-build.r
```diff
@@ -432,3 +359,88 @@ by_layer <- function(f, layers, data, step = NULL) {
...
+    )
+    place <- find_panel(table)
+  } else {
+    n_col <- ncol(table)
+    n_row <- nrow(table)
+    # Actually fill margin with relevant units
+    if (top)    table$heights <- unit.c(height, table$heights[-1])
+    if (left)   table$widths  <- unit.c(width,  table$widths[-1])
+    if (right)  table$widths  <- unit.c(table$widths[-n_col],  width)
+    if (bottom) table$heights <- unit.c(table$heights[-n_row], height)
```
teunbrand, line 432: I think this pattern was some workaround for grid units not handling subassignment in R3.2 and earlier. Can we subassign units now?

teunbrand:
Should merge #5175 before this. Then update this PR to drop changes introduced in [5dd6c2a](https://github.com/tidyverse/ggplot2/pull/5167/commits/5dd6c2a1ee5edc324c73980fb0995f6cb551a0c3).

R/plot-build.r
thomasp85: yes, thanks... We should generally try to be predictable wrt the gtable structure that ggplot2 produces as it makes aligning multiple plots easier for patchwork

teunbrand: That makes good sense to me :)

thomasp85:
With this PR there will be 3 different placement strategies for tags. Either in the margin (adding space for it), on top of the full plot area (not adding space), or in the panel area (not adding space). Do all three have a name in this new API?

teunbrand:
I also found that if you use the margin placement, `hjust/vjust` is ignored. This is isn't as much as issue if you use the plot placement, as the two numbers you give essentially determine the `hjust/vjust`.

To illustrate:

``` r
library(ggplot2)

  geom_point() +
  labs(tag = "Foobar")
```

In the plot below, I'd expect the tag to be in the top left, with a left margin for the plot, but not a top margin for the plot.

```
p + theme(plot.tag.position = "left",
          plot.tag = element_text(vjust = 1))
```



Likewise, here I would have expected the text to be at the top right, with a plot margin on top but not on the right.

``` r
p + theme(plot.tag.position = "top",
          plot.tag = element_text(hjust = 1))
```



<sup>Created on 2023-03-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85:
I like the "margin" name and think we should include that for completeness...

As for alignment, it works "correctly" right now in that it is perfectly aligned to the textbook that closely wraps the tag text. It is not given that the alignment is relative to the panel size.

I can see the argument for having it as you describe because the alignment is right now useless but I don't think it is obviously a bug as it is

NEWS.md
```diff
@@ -19,6 +19,8 @@
   `panel.grid.major`'s linewidth setting (@teunbrand, #5179)
 * The `datetime_scale()` scale constructor is now exported for use in extension
   packages (@teunbrand, #4701).
+* The `plot.tag.position` in `theme()` can now accept `"panel"` to set the tag
+  relative to the panel area instead of the plot area (#4297).
```
thomasp85, line 23: Can you update the bullet to include the full set of new possibilities

thomasp85:
Can we somehow make this new alignment work with numeric positions..?

teunbrand:
We could in theory, but the way it is setup right now it isn't straightforward, as we cannot pass a mixed type character/numeric to the `plot.tag.position` setting. There is currently *some* flexibility of placement, in that you can set the hjust/vjust if only one of the directions is fixed:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  labs(tag = "Foobar")

p + theme(plot.tag.position = c("panel", "top"),
          plot.tag = element_text(hjust = 0.75))
```



<sup>Created on 2023-04-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I think we might be able to extend this by not having the position default to 'topleft' for the 'panel' setting.

thomasp85:
yeah, I know the limitations of the current setup. I was thinking we should maybe considering a different setup such as providing an additional theme element (e.g. `plot.tag.location`) that holds the `margin`/`plot`/`panel` info... this would also bring it somewhat more in line with other location styles

teunbrand:
We could make that work, but how should a numeric value interact with `plot.tag.location = "margin"`?

thomasp85:
yeah, that's a good question... 

But I think having that not work (and documented) is better than the alternative

teunbrand:
I'm just throwing an error now with numeric `plot.tag.position` and `plot.tag.location = "margin"`.

thomasp85 (approved):
LGTM

PR #5174, opened 2023-02-02
Export `datetime_scale()`

### LINKED ISSUE #4701: Export `datetime_scale`
`datetime_scale()` is not currently exported, but may be required when writing extensions to ggplot.  E.g. Writing new sclaes for custom geoms etc.

All other scales are exported. i.e. `continuous_scale()`, `discrete_scale()` and `binned_scale()`

Could `datetime_scale` also be exported?

teunbrand:
This PR aims to fix #4701.

Briefly, it exports the formerly internal `datetime_scale()` for use in extension packages. Parameter documentation is largely inherited from other functions, but the `trans` argument suggests date/time transforms.


R/scale-date.r
```diff
@@ -275,7 +275,16 @@ scale_y_time <- function(name = waiver(),
...
+#' Date/time scale constructor
+#'
+#' @inheritParams scale_x_datetime
+#' @inheritParams continuous_scale
+#' @param trans For date/time scales, the name of a date/time transformation or
+#'   the object itself. Built-in transformations include "hms", "date" and "time".
+#' @inheritDotParams continuous_scale -aesthetics -trans -palette -breaks -minor_breaks -labels -guide
+#'
+#' @export
+#' @keywords internal
```
teunbrand, line 287: I might be a fool, but I don't understand the pkgdown error:

> ! All topics must be included in reference index
✖ Missing topics: datetime_scale
ℹ Either add to _pkgdown.yml or use `@keywords internal`

Whereas adding the keywords is exactly what I've done

karawoo: Is it possible you ran `document()` before adding `@keywords internal` to the roxygen comments? It doesn't look like it made it through to the .Rd file.

teunbrand: Whoops, my bad: a fool it is then! Sharp eye, thanks for catching this!

thomasp85 (approved):
LGTM

PR #5188, opened 2023-02-14
`remove_missing()` considers characters as finite

### LINKED ISSUE #4284: remove_missings(..., finite = TRUE) removes everything if there is a character column
In base-R, character values are not "finite" (i.e. `is.finite("a")` is `FALSE`).  Therefore, if `remove_missings()` with `finite = TRUE` encounters a character column, it throws away everything. Thus, the first of the following two lines raises a warning and returns an empty data.frame, while the second line preserves the input data.frame:
```r
remove_missing(data.frame(foo = "bar", stringsAsFactors = FALSE), finite = TRUE)
remove_missing(data.frame(foo = "bar", stringsAsFactors = TRUE), finite = TRUE)
```

For me, this led to the following problem that I found extremely difficult to debug: I had a custom `stat` with `required_aes = c("x", "colour")`.  If the colour-aesthetic was mapped to a factor column, everything worked as expected - non-NA-values of factor variables are finite.  If the colour-aesthetic was mapped to a character column, the layer was dropped.

Minimal example:
```r
StatXMean <- ggplot2::ggproto(
  "StatXMean", Stat,
  compute_group = function(data, scales, params, na.rm = FALSE) {
    dplyr::summarize(dplyr::group_by(data, colour),
                     xintercept = mean(x, na.rm = TRUE))
  },  required_aes = c("x", "colour")
)

stat_x_mean <- function(mapping = NULL, data = NULL, geom = "vline",
                        position = "identity", na.rm = FALSE, show.legend = NA,
                        inherit.aes = TRUE, ...) {
  ggplot2::layer(
    stat = StatXMean, data = data, mapping = mapping, geom = geom,
    position = position, show.legend = show.legend, inherit.aes = inherit.aes,
    params = list(na.rm = na.rm, ...),
  )
}

  geom_point() + stat_x_mean()

ggplot(mpg, aes(displ, hwy, colour = as.factor(drv))) +
  geom_point() + stat_x_mean()
```

teunbrand:
This PR aims to fix #4284.

Briefly, `remove_missing()` now considers character vectors as finite (unless they're `NA`).
This makes it easier to have required aesthetics that are not numeric in custom Stats subclasses.


R/utilities.r
```diff
@@ -90,7 +90,7 @@ remove_missing <- function(df, na.rm = FALSE, vars = names(df), name = "",
   missing <- detect_missing(df, vars, finite)
 
   if (any(missing)) {
-    df <- df[!missing, ]
+    df <- df[!missing, , drop = FALSE]
```
teunbrand, line 93: This shouldn't occur in wild plotting because there should always be multiple columns. However, since it came up in testing, I thought to change this too.

thomasp85: Agreed - I think we should aim to make the code as clear and type stable as possible

thomasp85 (approved):
LGTM

teunbrand:
Thanks for another review Thomas!

PR #5199, opened 2023-02-21
use tempfiles for plots

MichaelChirico:
I think it's not strictly necessary for CRAN, but it is consistent with the rest of the suite.

(discovered because we run tests without local write permissions, this is the only test in the suite with this issue)

Also moved one test out of the `withr()` environment because it's independent of it, I think it's for the best.

### DIFF
--- a/tests/testthat/test-ggsave.R
+++ b/tests/testthat/test-ggsave.R
@@ -57,12 +57,11 @@ test_that("ggsave can handle blank background", {
 })
 
 test_that("ggsave warns about empty or multiple filenames", {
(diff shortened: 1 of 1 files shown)


yutannihilation:
Thanks for catching. How about replacing `withr::with_file()` with `withr::with_tempfile()`?

MichaelChirico:
Oh, nice, I didn't realize `with_tempfile()` could accept a vector of inputs like that. Done.

yutannihilation:
Yeah, I too just noticed it today... Thanks, I'll merge this after the CI pass.

yutannihilation (approved):
Thanks!

PR #5202, opened 2023-02-22
Friendlier error when supplying extension-less files

MichaelChirico:
The error when trying `ggsave(p, tempfile())` is head-cocking:

```
Error in `ggsave(filenames, plot)`: Unknown graphics device ""
```

This should error should help fix the issue faster.


thomasp85 (approved):
Thanks!

PR #5203, opened 2023-02-23
Description and comment typos fixed

hub-shale:
Updates text in geom_text.Rd (reference page) for grammatical errors and style.


teunbrand:
Hi there,

It seems you have edited the documentation page directly. ggplot2 generates its documentation with the roxygen package, which means that the source file must be edited instead of the documentation page. As the instructions at the top of the documentation page read:

https://github.com/tidyverse/ggplot2/blob/77bab75a8375be0718cbe10dff90004b37dae798/man/geom_text.Rd#L1-L2

You would need to edit `geom-text.R` or `geom-label.R` file directly. To update the documentation page, you can then run `devtools::document()`.

PR #5210, opened 2023-02-27
Performance: use `lengths()`

teunbrand:
A common pattern in the codebase was to use `vapply(x, length, integer(1)` to get the length of every member in a list. That can be replaced by the consistently faster (and more readable) `lengths()` function. Overall, I don't think it makes a ton of difference, but may shave the occasional microsecond off a plot.

``` r
library(ggplot2)

lengths <- 2^(0:20)

set.seed(42)
lens <- sample(0:1000, max(lengths), replace = TRUE)

my_list <- lapply(lens, rnorm)

bms <- lapply(lengths, function(n) {
  x <- my_list[seq_len(n)]
  bench::mark(
    lengths(x),
    vapply(x, length, integer(1)), 
    iterations = 50
  )
})

df <- do.call(rbind, bms)
df$n <- rep(lengths, each = 2)

ggplot(df, aes(n, as.numeric(median), colour = as.character(expression))) +
  geom_line() +
  scale_y_log10() +
  scale_x_log10() +
  labs(
    x = "Length of list",
    y = "Seconds",
    colour = "Pattern"
  )
```

<sup>Created on 2023-02-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/annotation.r
+++ b/R/annotation.r
@@ -56,7 +56,7 @@ annotate <- function(geom, x = NULL, y = NULL, xmin = NULL, xmax = NULL,
   aesthetics <- c(position, list(...))
 
   # Check that all aesthetic have compatible lengths
-  lengths <- vapply(aesthetics, length, integer(1))
+  lengths <- lengths(aesthetics)
   n <- unique0(lengths)
 
   # if there is more than one unique length, ignore constants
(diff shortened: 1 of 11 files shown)


yutannihilation (approved):
Looks good!

teunbrand:
Thanks for the review!

PR #5223, opened 2023-03-07
replace `rlang::flatten_if()`

teunbrand:
The upcoming rlang 1.1.0 deprecates `flatten()` and variants, see https://github.com/r-lib/rlang/blob/main/NEWS.md.

The recommended replacement is `purrr::list_flatten()`, but it seems wild to me that ggplot2 would take on an additional dependency for this 1 occasion that `flatten_if()` is used.

This PR mimics the logic of `purrr::list_flatten()` to replace `rlang::flatten_if()` without taking on {purrr} as a dependency.


thomasp85 (approved):
LGTM - good catch

PR #5231, opened 2023-03-13
Fix `makeContext.dotstackGrob()` method inconsistency

teunbrand:
The `makeContext()` function in {grid} has only one argument: `x`.
The proposed change will remove the argument `recording` from `makeContext.dotstackGrob`, which didn't seem to be used anyway. 
This brings the method in line with the generic.

Tested on r-devel/4.3.0 on Windows

### DIFF
--- a/R/grob-dotstack.R
+++ b/R/grob-dotstack.R
@@ -25,7 +25,7 @@ dotstackGrob <- function(
 is_npc <- function(x) isTRUE(grepl('^[^+^-^\\*]*[^s]npc$', as.character(x)))
 
 #' @export
-makeContext.dotstackGrob <- function(x, recording = TRUE) {
+makeContext.dotstackGrob <- function(x) {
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM - have you tested this on R-devel?

teunbrand:
Yes, I downloaded yesterday's R-devel and tested locally on my Windows machine that this PR would fix the new warning that popped up in CRAN checks. You can also see that the 'ubuntu-latest (devel)' passes in this PR, but not in other recent ones.

The 'oldrel-2/3' failures are unrelated to this PR, I think.

thomasp85:
Great - I'm always a bit weary of relying on the devel cli checks to catch recent changes in r-devel so best to confirm locally as you did.

The other failures are indeed unrelated and due to Hmisc moving to using the `|>` pipe which are not available in old versions of R

Issue #5230, opened 2023-03-13
Review {Hmisc} dependency


teunbrand:
This issue is mostly prompted by recent CI failures due to the Hmisc package switching to use a few native pipes (more details at https://github.com/harrelfe/Hmisc/issues/159, a PR for a fix might be underway). Since tidyverse packages are expected to support non-native-pipe versions of R (< 4.1.0) up to 2025, it might be best to explore whether ggplot2 can drop Hmisc from the 'Suggests' list.

In addition to the reason above, this paper https://arxiv.org/abs/2208.11674 describes {Hmisc} as a 'heavy parent' package (and off-topic: ggplot2 as a 'hub' package). It might be good for R package ecosystem health if ggplot2 did not depend on {Hmisc} if it is a heavy parent. Moreover, it seems also wise to prevent a similar issue as with the {isoband} package down the line.

As far as I can tell, the {Hmisc} package is needed for the `mean_cl_boot()`,  `mean_cl_normal()`, `mean_sdl()` and `median_hilow()` functions over here: https://github.com/tidyverse/ggplot2/blob/04a5ef274e912bee76180154b25d8bca0206feb1/R/stat-summary.R#L248-L273

It might not be terribly inconvenient to offer a few homegrown alternatives for these 4 functions. Any thoughts?

yutannihilation:
Sorry for noticing this trouble late.

If the intent is just to pass the CI, I think we don't need to desperate to drop Hmisc. In my understanding, the tidyverse's "4 previous versions" rule doesn't mean it maintains all the functionalities including the ones that need the Suggest dependencies. In fact, ggplot2 already gave up some packages on the older versions.

https://github.com/tidyverse/ggplot2/blob/d6d0523fc51dfb7ad67c093900269b0e7701d26a/.github/workflows/R-CMD-check.yaml#L66-L67

But, considering we have repeatedly faced such issues caused by the different support policies, it might be a good idea to implement these functions within ggplot2.

teunbrand:
> If the intent is just to pass the CI

It would be nice if we could pass the CI, but I think that is a secondary issue. I think the future-proofing and ecosystem bits are of some importance though. 

This probably isn't limited to just {Hmisc}. There are some other suggested packages that I cannot find any (or very limited) code references to (some are mentioned in docs though). Examples of these are {munsell}, {RColorBrewer}, {lattice}, {rpart}, {rgeos}.

[...]

PR #5242, opened 2023-03-21
Remove rgeos from the dependency

yutannihilation:
(See https://github.com/tidyverse/ggplot2/issues/5230#issuecomment-1475217544)

rgoes is used nowhere. I couldn't find how it was used. We can find the name "rgoes" only in this NEWS item of ggplot2 v0.9.2. My guess is that `SpatialPolygonsDataFrame` was living in rgeos at that time (now in sp)?

> * Dependency on `gpclib` removed, and `fortify.SpatialPolygonsDataFrame` will
  now use `rgeos` if available - this is particularly useful if you're not
  able to use the non-free `gpclib`.

It seems rgeos already became unused in 2015, so anyway I believe it's safe to remove rgoes now.

https://github.com/tidyverse/ggplot2/pull/1098

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,7 +63,6 @@ Suggests:
     quantreg,
     ragg,
     RColorBrewer,
-    rgeos,
     rmarkdown,
     rpart,
     sf (>= 0.7-3),


thomasp85:
This seems timely as rgeos is being retired some time during 2023 according to its DESCRIPTION file

PR #5247, opened 2023-03-23
`mapped_discrete` has resolution of 1

### LINKED ISSUE #5211: different widths in bar plot when using undropped factor levels
I found a potential problem with the widths of bars using `geom_bar` when using factor levels, and keeping unused levels with `drop = F` argument in `scale_x_discrete`. This can be overcome by setting the `width` in aesthetic but thought the variable nature of widths in plotting might be an issue.

``` r
library(ggplot2)

df <- data.frame(
  x = factor(c('A', 'D'), levels = LETTERS[1:5])
)

ggplot(df, aes(x)) +
  geom_bar() +
  scale_x_discrete(drop = F)
```

``` r

df <- data.frame(
  x = factor(c('B', 'D'), levels = LETTERS[1:5])
)

ggplot(df, aes(x)) +
  geom_bar() +
  scale_x_discrete(drop = F)
```

<sup>Created on 2023-02-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5211.

It does so by having `resolution()` consider the `mapped_discrete` class as it does integers, where the resolution is set to 1.
I've added some tests so that it becomes easier to see if `resolution()` breaks in the future.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* To improve `width` calculation in bar plots with empty factor levels, 
+  `resolution()` considers `mapped_discrete` values as having resolution 1 
+  (@teunbrand, #5211)
 * `coord_flip()` has been marked as superseded. The recommended alternative is
   to swap the `x` and `y` aesthetic and/or using the `orientation` argument in
   a layer (@teunbrand, #5130).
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #5254, opened 2023-03-30
Weighted density

### LINKED ISSUE #4176: In `stat_density()`, the calculated `count` aesthetic should take weights into account
The weight should sum to one, ignoring color or fill aesthetics. That was the behavior before as can be seen here: https://stackoverflow.com/questions/20342494/density-of-each-group-of-weighted-geom-density-sum-to-one
In the way that geom_density is now there is no way to make the sum of both areas be equal to one.

Example:
```r
library(ggplot2)
library(ggplot2movies) 
m <- ggplot()
m + geom_density(data = movies, aes(rating, weight = votes, colour = factor(Action)))
```
The plot above is equal to 

```r
m + geom_density(data = movies[movies$Action == 0, ], aes(rating, weight = votes/sum(votes)), fill=NA, colour="black") +
    geom_density(data = movies[movies$Action == 1, ], aes(rating, weight = votes/sum(votes)), fill=NA, colour="blue")
```
I would expect a much smaller area for Action == 1 in the first plot, as less than 10% of the movies are action movies.

teunbrand:
This PR aims to fix #4176.

It adds the `wdensity` computed variable, which is the density times the sum of weights.
If weights are absent, this reduces to the same as the `count` computed variable.

An example:

``` r
#> ℹ Loading ggplot2

set.seed(42)
df <- data.frame(x = rnorm(100), y = abs(rnorm(100)), type = c(1, 5))

p <- ggplot(df, aes(x, colour = factor(type)))

# Weights are pointless here because they are by group
p + geom_density(aes(weight = type, y = after_stat(count)))
```

``` r
# Less pointless weights
p + geom_density(aes(weight = type, y = after_stat(wdensity)))
```

<sup>Created on 2023-03-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85:
I'm fine with this, but looping in @clauswilke as he seemed to have more opinions on design in the issue thread

R/stat-density.R
```diff
@@ -23,6 +23,8 @@
 #' @eval rd_computed_vars(
 #'  density  = "density estimate.",
 #'  count    = "density * number of points - useful for stacked density plots.",
+#'  wdensity = "density * sum of weights. In absence of weights, the same as
+#'  `count`",
```
clauswilke, line 27: Probably better to add a period at the end of this sentence as is the case for all the other descriptions.

R/stat-density.R
```diff
@@ -112,17 +114,19 @@ StatDensity <- ggproto("StatDensity", Stat,
...
                             kernel = "gaussian", n = 512,
                             bounds = c(-Inf, Inf)) {
-  nx <- length(x)
+  nx <- w_sum <- length(x)
   if (is.null(w)) {
     w <- rep(1 / nx, nx)
   } else {
-    w <- w / sum(w)
+    w_sum <- sum(w)
+    w <- w / w_sum
```
clauswilke, line 122: The function `fit_data_to_bounds()` checks for division by zero but here you're not doing that. I think in weird corner cases with zero weights this could give you different results depending on whether a point is inside the bounds or not.

clauswilke (commented):
I have two relatively minor comments (see code annotations) and one bigger one. The bigger one is: Have you checked how the weights interact with the density reflection when there are bounds imposed? Is there anything that needs to be done or are things behaving as expected. I'm not sure I remember what the density reflection does, so it may well be that everything is correct as is, just wanted to be certain.

R/stat-density.R
teunbrand: Ah yes, thanks for catching!

R/stat-density.R
teunbrand: I've been trying to come up with examples that should be valid, but aren't, but I'm not having an easy time of it. If you censor a non-zero weight and are left with all zero weights, the weighted density should reflect that and show a flat line. Currently, it does this, but also passes on a warning from `stats::density()`: 

``` r
#> ℹ Loading ggplot2
df <- data.frame(x = 1:3)

ggplot(df, aes(x)) +
  geom_density(aes(weight = c(1, 0, 0), y = after_stat(wdensity)),
               bounds = c(1.1, Inf))
#> Warning: Some data points are outside of `bounds`. Removing them.
#> Warning in density.default(x, weights = w, bw = bw, adjust = adjust, kernel =
#> kernel, : sum(weights) != 1 -- will not get true density
```



<sup>Created on 2023-04-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'm wondering if there is something to be improved here.

clauswilke: Ok. As far as I'm concerned you can handle it however you see fit, I just wanted to call out the inconsistency.

thomasp85 (approved):
LGTM

PR #5256, opened 2023-04-01
Warn when using constant aesthetics

### LINKED ISSUE #5253: Surprising behaviour when `aes()` is used to map all aesthetics to constant values
When all mappings in a geom are to constant values, whole-plot default data should not affect the number of grobs generated.
Current behaviour of making as many copies as rows there are in data is surprising/counterintuitive as demonstrated by https://stackoverflow.com/questions/75881173/

In the example below, there is no mapping to variables in `data` in layer 2 of either `p1` or `p2`, however, when rendered, `p1` includes as many grobs for layer 2, as there are rows in data. In `p2` only one grob is created, resulting in a much smaller SVG file.

``` r
library(ggplot2)

p1 <- ggplot(data = mpg, aes(class, hwy)) + 
  geom_boxplot() + 
  geom_text(aes(label = "***", x = 5.5, y = 41), size = 6)

str(p1$layers[[2]]$data)
#>  list()
#>  - attr(*, "class")= chr "waiver"
str(p1$layers[[2]]$mapping)
#> List of 3
#>  $ x    : num 5.5
#>  $ y    : num 41
#>  $ label: chr "***"
#>  - attr(*, "class")= chr "uneval"

p2 <- ggplot(data = mpg, aes(class, hwy)) + 
  geom_boxplot() + 
  geom_text(data = data.frame(), aes(label = "***", x = 5.5, y = 41), size = 6)

str(p2$layers[[2]]$data)
#> 'data.frame':    0 obs. of  0 variables
str(p1$layers[[2]]$mapping)
#> List of 3
#>  $ x    : num 5.5
#>  $ y    : num 41
#>  $ label: chr "***"
#>  - attr(*, "class")= chr "uneval"

object.size(cowplot::as_grob(p1))
#> 814272 bytes
object.size(cowplot::as_grob(p2))
#> 794984 bytes

svglite::svglite("p1.svg")
print(p1)
dev.off()
#> png 
#>   2

file.size("p1.svg")
#> [1] 50646

svglite::svglite("p2.svg")
print(p2)
dev.off()
#> png 
#>   2

file.size("p2.svg")
#> [1] 12667

# the diference can be seen starting at line 94 in the .svg files.
```

The following line appears once in p2.svg, and repeated 234 times in p1.svg.

```
<text x='515.78' y='82.03' text-anchor='middle' style='font-size: 17.07px; font-family: "Arial";' textLength='19.92px' lengthAdjust='spacingAndGlyphs'>***</text>
```
<sup>Created on 2023-03-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5253.

Briefly, it adds a warning when using an all-constant mapping in a layer:

``` r
#> ℹ Loading ggplot2

p <- ggplot(mtcars, aes(x = 1)) +
  geom_point(aes(y = 2))
b <- ggplot_build(p)
#> Warning: All aesthetics have length 1, but the data has 32 rows in `geom_point()`.
#> ℹ Did you mean to use `annotate()`?
```

It does take into account inherited aesthetics, so no warning is issued here:

``` r
p <- ggplot(mtcars, aes(x = seq_along(mpg))) +
  geom_point(aes(y = 2))
b <- ggplot_build(p)
```

There is an ambiguous case where I was unsure a warning should be emitted. If a layer inherits an aesthetic that it does not use, should the warning been thrown for the aesthetics it does use?

``` r
# To warn or not to warn?
p <- ggplot(mtcars, aes(label = cyl)) +
  geom_point(aes(x = 1, y = 1))
b <- ggplot_build(p)
```

<sup>Created on 2023-04-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The `geom_abline()`/`geom_vline()`/`geom_hline()` are excluded from this check. Extenders can exclude their geoms from this check by using the same exclusion flag (`check_constant_aes = FALSE` in the geom's ggproto fields).

### DIFF
--- a/R/layer.R
+++ b/R/layer.R
@@ -304,10 +304,9 @@ Layer <- ggproto("Layer", NULL,
     if ((self$geom$check_constant_aes %||% TRUE)
         && length(aes_n) > 0 && all(aes_n == 1) && n > 1) {
       cli::cli_warn(c(
-        paste0("All aesthetics have length 1, but the data has {n} rows ",
-               "in {.fn {self$constructor[1]}}."),
+        "All aesthetics have length 1, but the data has {n} rows.",
         i = "Did you mean to use {.fn annotate}?"
-      ))
+      ), call = self$constructor)
     }
     check_aesthetics(evaled, n)
 
--- a/R/layer.R
+++ b/R/layer.R
@@ -301,9 +301,11 @@ Layer <- ggproto("Layer", NULL,
         n <- if (min(aes_n) == 0) 0L else max(aes_n)
       }
     }
-    if (length(aes_n) > 1 && all(aes_n == 1) && n > 1) {
+    if ((self$geom$check_constant_aes %||% TRUE)
+        && length(aes_n) > 0 && all(aes_n == 1) && n > 1) {
       cli::cli_warn(c(
-        "All aesthetics have length 1, but the data has {n} rows.",
+        paste0("All aesthetics have length 1, but the data has {n} rows ",
+               "in {.fn {self$constructor[1]}}."),
         i = "Did you mean to use {.fn annotate}?"
       ))
     }
--- a/R/layer.R
+++ b/R/layer.R
@@ -292,14 +292,20 @@ Layer <- ggproto("Layer", NULL,
     }
 
     n <- nrow(data)
+    aes_n <- lengths(evaled)
     if (n == 0) {
       # No data, so look at longest evaluated aesthetic
       if (length(evaled) == 0) {
         n <- 0
       } else {
-        aes_n <- lengths(evaled)
         n <- if (min(aes_n) == 0) 0L else max(aes_n)
       }
+    }
+    if (length(aes_n) > 1 && all(aes_n == 1) && n > 1) {
+      cli::cli_warn(c(
+        "All aesthetics have length 1, but the data has {n} rows.",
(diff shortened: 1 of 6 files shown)


R/layer.R
```diff
@@ -292,15 +292,23 @@ Layer <- ggproto("Layer", NULL,
...
       } else {
-        aes_n <- lengths(evaled)
         n <- if (min(aes_n) == 0) 0L else max(aes_n)
       }
     }
+    if ((self$geom$check_constant_aes %||% TRUE)
+        && length(aes_n) > 0 && all(aes_n == 1) && n > 1) {
+      cli::cli_warn(c(
+        paste0("All aesthetics have length 1, but the data has {n} rows ",
+               "in {.fn {self$constructor[1]}}."),
```
teunbrand, line 308: Not sure if `self$constructor[1]` is a very safe pattern to report the layer

thomasp85: See https://github.com/tidyverse/ggplot2/blob/d7f22413efea3dd2a7c9effff05d4b2aa2c2d300/R/plot-build.R#L430

We prefer to pass this in as the surrounding call, rather than name it in the error text

teunbrand: Thanks that makes much more sense!

PR #5260, opened 2023-04-05
Text size switch

### LINKED ISSUE #3799: Add fontsize aesthetic to geom_text
A common source of confusion is the scaling of text in layers and how to match sizing between that and theme element text. This is mainly due to `size` being interpreted as `cm` to make it in line with the sizing of other geoms.

Would it make sense to introduce an alternative aesthetic to geom_text that was interpreted as 'pts' instead which, if given, would take precedence?

Maybe it shouldn't even be an aesthetic, but a switch, I dunno... But this comes up often enough that we should discuss how to address it...

teunbrand:
This PR aims to fix #3799.

Briefly, it adds a `size_unit` parameter to `geom_text/label()` that allows one to use `"pt"` to have the size interpreted as points.

Example:

``` r
#> ℹ Loading ggplot2

ggplot() +
  annotate("text",  x = -1, y = 0, label = "Abc", size = 10, size_unit = "mm") +
  annotate("text",  x = 1,  y = 0, label = "Def", size = 10, size_unit = "pt") +
  annotate("label", x = -1, y = 1, label = "Ghi", size = 10, size_unit = "mm") +
  annotate("label", x = 1,  y = 1, label = "Jkl", size = 10, size_unit = "pt") +
  xlim(-2, 2) + ylim(-1, 2)
```

<sup>Created on 2023-04-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (changes requested):
As garish as it is, arguments in ggplot2 are "." separated instead of "_" separated

thomasp85:
I'm trying to figure out if mm and pt are too limiting since grid supports a multitude of units... maybe it is not so big a deal

teunbrand:
I can see some utility for absolute units, but I'd have no idea what to expect for `1null` or `1npc` text sizes.

thomasp85:
yeah, I'm only thinking of adding the other absolute sizes, not the relative ones

teunbrand:
I've added centimetres, inches and picas in addition to the points and millimetres. Other absolute units are 'bigpts', 'dida', 'cicero' and 'scaledpts', but I've never seen anybody use these in the wild, so I didn't add them. Let me know if you'd like to add these.

PR #5218, opened 2023-03-06
Allow 'weight' to be dropped in stats

teunbrand:
This PR aims to fix #5216.

Briefly, in several stats, warnings were emitted when the 'weight' aesthetic was used. However, the weight aesthetic is *meant* to be dropped, so we register them as such. 

In essence, the problem is the same as in #5053, so the proposed remedy is the same as in #5071, but applied to more stats.

PR #5071, opened 2022-12-01
Drop `weight` aesthetic quietly in `stat_smooth()`

teunbrand:
This PR attempts to fix #5053 based on @clauswilke's suggestion.

PR #5267, opened 2023-04-12
Silence additional weight warnings

### LINKED ISSUE #5263: `geom_violin()` warns that `weight` aesthetic dropped during statistical transformation
Similar to https://github.com/tidyverse/ggplot2/issues/5053, `geom_violin()` also warns that weight is dropped, although it appears to be used as expected.

Reprex: (note that the two `geom_violin()` calls produce different outlines)

```r
library(ggplot2)

  geom_point(aes(size = cyl)) +
  geom_violin(aes(weight = cyl)) +
  geom_violin()
```

teunbrand:
This PR aims to fix #5263.

It continuous the saga of silencing dropped weight warnings, like #5071 and #5218 before it.

I also found `geom_dotplot()` to use a `weight` aesthetic, but it reports it doesn't and throws a warning about dropped weights when present. I fixed this by promoting it to a visible aesthetic and silencing the warning. In the example below, only the bin width warnings persist.

``` r
library(ggplot2)

p <- ggplot(mtcars, aes(x = mpg))
p1 <- p + geom_dotplot(aes(weight = cyl))
#> Warning in geom_dotplot(aes(weight = cyl)): Ignoring unknown aesthetics: weight
p2 <- p + geom_dotplot()
patchwork::wrap_plots(p1, p2)
#> Bin width defaults to 1/30 of the range of the data. Pick better value with
#> `binwidth`.
#> Warning: The following aesthetics were dropped during statistical transformation: weight
#> ℹ This can happen when ggplot fails to infer the correct grouping structure in
#>   the data.
#> ℹ Did you forget to specify a `group` aesthetic or to convert a numerical
#>   variable into a factor?
#> Bin width defaults to 1/30 of the range of the data. Pick better value with
#> `binwidth`.
```

<sup>Created on 2023-04-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/geom-dotplot.R
+++ b/R/geom-dotplot.R
@@ -188,7 +188,8 @@ GeomDotplot <- ggproto("GeomDotplot", Geom,
   required_aes = c("x", "y"),
   non_missing_aes = c("size", "shape"),
 
-  default_aes = aes(colour = "black", fill = "black", alpha = NA, stroke = 1, linetype = "solid"),
+  default_aes = aes(colour = "black", fill = "black", alpha = NA,
+                    stroke = 1, linetype = "solid", weight = 1),
 
   setup_data = function(data, params) {
     data$width <- data$width %||%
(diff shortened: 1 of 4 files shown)


thomasp85 (approved):
LGTM

PR #5269, opened 2023-04-13
Update missing value warning

### LINKED ISSUE #5250: Incorrect warning from geom about missing values when scale sets limits, unless data is filtered first
---

I found a problem with a misleading warning from geom_xxx when a scale is truncated with the limits argument

I expected no warnings.

I suspect this is related to or the same problem as https://github.com/tidyverse/ggplot2/issues/5118 but because the discussion there is histogram-specific I have opened a new issue.

```r

library(tidyverse)

d <- tibble (yr = rep(2000:2020, 2),
             animal = rep(c("cat", "dog"), each = 21),
             value = rnorm(42))

d |>
  ggplot(aes(x = yr, y = value, colour = animal)) +
  geom_line() +
  scale_x_continuous(limits = c(2011, 2020))
#> Warning: Removed 22 rows containing missing values (`geom_line()`).
```

``` r

# also happens with other geoms:
d |>
  ggplot(aes(x = yr, y = value, colour = animal)) +
  geom_point() +
  scale_x_continuous(limits = c(2011, 2020))
#> Warning: Removed 22 rows containing missing values (`geom_point()`).
```

``` r

# Warning goes away if data is explicitly filtered to match the scale limits:
d |>
  filter(yr >= 2011) |>
  ggplot(aes(x = yr, y = value, colour = animal)) +
  geom_line() +
  scale_x_continuous(limits = c(2011, 2020))
```

<sup>Created on 2023-03-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
```

### PR DESCRIPTION
**Pull request by teunbrand**
This PR aims to fix #5250.

Briefly, it updates the warning message that is omitted by `Geom$handle_na()` and their kin to read:
```r
#> Warning message:
#> Removed 10 rows containing missing or excluded values (`stat_boxplot()`). 
```
Where the 'or excluded' is new. This is to better reflect that values might have been excluded by the out-of-bounds functions, in addition to being properly missing.

### DIFF
--- a/R/geom-path.R
+++ b/R/geom-path.R
@@ -140,7 +140,10 @@ GeomPath <- ggproto("GeomPath", Geom,
     data <- data[kept, ]
 
     if (!all(kept) && !params$na.rm) {
-      cli::cli_warn("Removed {sum(!kept)} row{?s} containing missing values ({.fn {snake_class(self)}}).")
+      cli::cli_warn(paste0(
+        "Removed {sum(!kept)} row{?s} containing missing ",
+        "or excluded values ({.fn {snake_class(self)}})."
+      ))
     }
 
     data
--- a/R/geom-path.R
+++ b/R/geom-path.R
@@ -141,8 +141,8 @@ GeomPath <- ggproto("GeomPath", Geom,
 
     if (!all(kept) && !params$na.rm) {
       cli::cli_warn(paste0(
-        "Removed {sum(!kept)} row{?s} containing missing ",
-        "or excluded values ({.fn {snake_class(self)}})."
+        "Removed {sum(!kept)} row{?s} containing missing values or values ",
+        "outside the scale range ({.fn {snake_class(self)}})."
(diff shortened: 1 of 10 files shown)


thomasp85:
Is "excluded" the right word here? We want to mainly make people aware of the possibility that the scale is removing them so I wonder if we can make that even more clear...

Maybe something along these lines? @clauswilke and @batpigandme do you have any input on how to make it read clearly?

```r
#> Warning message:
#> Removed 10 rows containing missing values or values outside the scale range (`stat_boxplot()`). 
```

batpigandme:
I like your suggestion, @thomasp85, since "excluded" could be a bit ambiguous.

clauswilke:
Agreed. The latest version looks good.

teunbrand:
Thanks for the input folks. I've updated the message as suggested.

thomasp85 (approved):
LGTM

PR #5270, opened 2023-04-13
Export `translate_shape_string()`

### LINKED ISSUE #5191: export `translate_shape_string()`
Function `translate_shape_string()` or a replacement of it is needed in packages extending 'ggplot2' with geoms that accept the shape aesthetic, such as my package 'ggpp'. I have copied the definition of `translate_shape_string()` from 'ggplot2' into 'ggpp' but this could lead to inconsistencies in the future, for example in error handling and messages.

It would be convenient if 'ggplot2' would export this function, although the workaround of redefining it when needed is obviously fully functional.

teunbrand:
This PR aims to fix #5191.

Briefly, it documents and exports the `translate_shape_string()`, to make it easier to construct extensions of `geom_point()`.


thomasp85 (approved):
LGTM

PR #5273, opened 2023-04-17
Refactor `titleGrob()`

teunbrand:
This PR aims to fix a performance bottleneck for #5030 as identified in https://github.com/tidyverse/ggplot2/pull/5030#issuecomment-1509597781.

Briefly, it introduces the following changes:

1. The `titleGrob` grobs no longer have viewports.
2. Instead of viewports, the position of the `textGrob` is adjusted to get the same effect.
3. The `titleGrob()` constructor now does the job of what previously was done by `title_spec()` and `add_margins()`.
4. The `assemble_strip()` no longer needs to awkwardly hassle with all these viewports.
5. A bunch of functions, like `title_spec()`, `add_margins()`, `margin_width()`, `margin_height()` have become obsolete. I've removed them, but we might expect this to bite us during revdepchecks because they're missing.

Benchmark below shows a ~4 fold reduction in drawing time for 574 titleGrobs.

``` r
library(ggplot2)
library(grid)
library(scales)

x <- rescale(as.numeric(economics$date))
y <- rescale(as.numeric(economics$unemploy))
lab <- as.character(economics$pop)

make_grob <- function(fun, x, y, label, margin = margin()) {
  do.call(
    grobTree,
    Map(
      fun,
      x = x, y = y, label = label, hjust = 0.5, vjust = 0.5,
      MoreArgs = list(margin_x = TRUE, margin_y = TRUE, margin = margin)
    )
  )
}

grob_old <- make_grob(
  ggplot2:::titleGrob, x = x, y = y, label = lab, 
  margin = margin(5, 5, 5, 5)
)

#> ℹ Loading ggplot2

grob_new <- make_grob(
  titleGrob, x = x, y = y, label = lab, 
  margin = margin(5, 5, 5, 5)
)

render <- function(grob) {
  file <- tempfile(fileext = ".png")
  ragg::agg_png(file)
  grid.newpage(); grid.draw(grob)
  dev.off()
  unlink(file)
}

bench::mark(
  render(grob_old),
  render(grob_new),
  min_iterations = 10
)
#> # A tibble: 2 × 6
#>   expression            min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>       <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 render(grob_old)    575ms    585ms      1.67    3.67MB     5.02
#> 2 render(grob_new)    126ms    135ms      7.50  132.94KB     5.25
```

<sup>Created on 2023-04-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


tests/testthat/_snaps/guides/align-facet-labels-facets-horizontal.svg
```diff
@@ -18,7 +18,7 @@
   </clipPath>
 </defs>
 <g clip-path='url(#cpMC4wMHw3MjAuMDB8MC4wMHw1NzYuMDA=)'>
-<rect x='-0.000000000000064' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
+<rect x='0.00' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
```
teunbrand, line 21: All these visual tests needed to be updated because there are slight rounding errors that are now rounded differently. Nothing looks different to the human eye though.

tests/testthat/_snaps/theme/custom-strip-elements-can-render.svg
```diff
@@ -18,7 +18,7 @@
   </clipPath>
 </defs>
 <g clip-path='url(#cpMC4wMHw3MjAuMDB8MC4wMHw1NzYuMDA=)'>
-<rect x='0.000000000000064' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
+<rect x='0.00' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
```
teunbrand, line 21: I don't think this snapshot is snapshotting what it should. I think a `registerS3method()` call is missing from the test, because it renders rectangles interactively, but you don't see this back in the snapshot.

R/margins.R
```diff
@@ -66,145 +64,80 @@ title_spec <- function(label, x, y, hjust, vjust, angle, gp = gpar(),
...
-      gp = gp
-    )
-    child_vp <- viewport(layout.pos.row = 2, layout.pos.col = 2)
-  } else if (margin_x) {
-    widths <- unit.c(margin[4], width, margin[2])
-    vp <- viewport(layout = grid.layout(1, 3, widths = widths), gp = gp)
-    child_vp <- viewport(layout.pos.col = 2)
+  # Calculate new x/width
+  if (margin_x) {
+    new_width  <- unit.c(margin[4], width, margin[2])
```
thomasp85, line 92: another double space :-)
```suggestion
    new_width <- unit.c(margin[4], width, margin[2])
```

R/margins.R
```diff
@@ -66,145 +64,80 @@ title_spec <- function(label, x, y, hjust, vjust, angle, gp = gpar(),
...
     margin <- margin(0, 0, 0, 0)
   }
+  margin_x <- isTRUE(margin_x)
+  margin_y <- isTRUE(margin_y)
 
-  if (margin_x && margin_y) {
-    widths <- unit.c(margin[4], width, margin[2])
-    heights <- unit.c(margin[1], height, margin[3])
+  # Initialise new values for position and dimensions
+  new_x <- new_y <- new_width <- new_height <- NULL
```
thomasp85, line 88: We don't really use this pattern for initialising multiple values - better to have them at their own lines

R/margins.R
teunbrand: Thanks for catching these; they are 'scars' of code alignment that I reversed later :')

PR #5278, opened 2023-04-19
Fix custom strip test

teunbrand:
It recently came to my attention in https://github.com/tidyverse/ggplot2/pull/5273#discussion_r1169222970 that a visual test wasn't testing what it should.

Briefly, the test is supposed to test whether custom `element_*()` classes can be drawn in strips. However, the `element_grob.element_test()` method wasn't recognised in the `expect_doppelganger()` test. This PR fixes the test by registering the method.


thomasp85 (approved):
Good catch

PR #5279, opened 2023-04-19
Scales don't squish to zero-range domains

### LINKED ISSUE #5271: Reciprocal transformation to continuous scale drops axis
R version 4.2.3
{ggplot2} version 3.4.2

In replicating some of the examples from [section 10.1.7](https://ggplot2-book.org/scale-position.html#scale-transformation) in "ggplot2: Elegant Graphics for Data Analysis" I found that the following code produces a figure without a guide on the y axis (see image below) despite the same code rendering a y axis in the book.

```r
library(ggplot2)
  geom_point() + 
  scale_y_continuous(trans = "reciprocal")
```

image

Other transformations passed to `scale_y_continuous()` (e.g., "log10", "sqrt") work as expected. The book uses an older version of {ggplot2} (3.3.6), so I'm wondering if this behavior might be a bug in the latest version.

teunbrand:
This PR aims to fix #5271.

Briefly, `reciprocal_trans()`'s domains are [-Inf, Inf]. The reciprocal of this domain is [0,0]. The limits get squished to [0,0]. We cannot compute breaks on this. Hence, in this PR, we refuse to squish limits to zero-range domains. 
Moreover, it adds a series of tests for computing breaks with almost all transformd (boxcox is broken though), so we might see better when we might break a particular transfomation.

### DIFF
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -648,7 +648,7 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     domain <- suppressWarnings(self$trans$transform(self$trans$domain))
     domain <- sort(domain)
     # To avoid NaN causing issues. NaN are dropped by the sort()
-    if (length(domain) == 2) {
+    if (length(domain) == 2 && !zero_range(domain)) {
       limits <- oob_squish(limits, domain)
     }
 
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #5281, opened 2023-04-22
Silence new `density()` warning

teunbrand:
This PR aims to prevent a newly introduced warning.

From the R4.3.0 [NEWS](https://cran.r-project.org/bin/windows/base/NEWS.R-4.3.0.html) bullets:

> density(x, weights = *) now warns if automatic bandwidth selection happens without using weights; new optional warnWbw may suppress the warning. Prompted by Christoph Dalitz' [PR#18490](https://bugs.r-project.org/show_bug.cgi?id=18490) and its discussants.

This affects ggplot2 when it uses weights:

``` r
library(ggplot2)

cutoff <- min(mtcars$mpg)

p <- ggplot(mtcars, aes(mpg, weight = cyl)) +
  stat_density()
invisible(ggplotGrob(p))
#> Warning in density.default(x, weights = w, bw = bw, adjust = adjust, kernel =
#> kernel, : Selecting bandwidth *not* using 'weights'
```

<sup>Created on 2023-04-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Because of backwards compatibility with older R versions, we cannot use the `warnWbw` argument.  This PR prevents this warning by precomputing the `bw` argument of `density()` and noting in the docs that weights are ignored for bandwidth calculation (as they always have been, it is just stated now).

### DIFF
--- a/R/stat-density.R
+++ b/R/stat-density.R
@@ -226,13 +226,13 @@ precompute_bw = function(x, bw = "nrd0") {
     bw <- arg_match0(bw, c("nrd0", "nrd", "ucv", "bcv", "sj", "sj-ste", "sj-dpi"))
     bw <- switch(
       to_lower_ascii(bw),
-      nrd0 = bw.nrd0(x),
-      nrd  = bw.nrd(x),
-      ucv  = bw.ucv(x),
-      bcv  = bw.bcv(x),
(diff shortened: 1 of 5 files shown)


thomasp85 (approved):
LGTM

PR #5283, opened 2023-04-24
Legend spacing

### LINKED ISSUE #4352: guide_legend ignores legend.spacing.(x/y) depending on byrow
Hello there,

I apologise in advance if I submit issues too frequently; I do really appreciate all the good work you put into ggplot2.
By default, a legend appears to ignore the `legend.spacing.y` theme setting.

``` r
library(ggplot2)

p <- ggplot(mtcars, aes(mpg, fill = as.factor(cyl))) +
  geom_density() +
  theme(legend.spacing.y = unit(0.5, "cm"))
p
```

However, when `byrow = TRUE` in the guide, the `legend.spacing.y` is applied as expected.

``` r
p + guides(fill = guide_legend(byrow = TRUE))
```

<sup>Created on 2021-02-22 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

For horizontal legend layouts, `legend.spacing.x` is ignored when `byrow = TRUE`, which is opposite to the vertical case. 

``` r
ggplot(mtcars, aes(mpg, fill = as.factor(cyl))) +
  geom_density() +
  theme(legend.spacing.x = unit(0.5, "cm"),
        legend.position = "bottom") +
  guides(fill = guide_legend(label.position = "top",
                             byrow = TRUE))
```

I was surprised by this as I would have expected the `legend.spacing` to be applied regardless of the layout order of the keys, but I'm likely not aware of the considerations going in to this. I believe the [code](https://github.com/tidyverse/ggplot2/blob/master/R/guide-legend.r#L448-L578) controlling the legend spacing and gaps sometimes `interleave()`s the label/key sizes with spacings, and sometimes doesn't, conditional on the `byrow` argument. 

In an earlier comment (https://github.com/tidyverse/ggplot2/issues/3587#issuecomment-546700387), it was proposed to get rid of the legend spacing, but I think there is a use-case for the spacing for the keys (the text margins maybe less so). Would it be a good idea to always interleave the spacing regardless of the `byrow` argument?

### LINKED ISSUE #4977: `legend.spacing.{x/y}` is not inherited in `guide_legend()`.
I was working on some code related to legends, and found out that the legend guide doesn't do inheritance of the `legend.spacing.{x/y}` theme elements from the parent `legend.spacing` element.

When I set a plot with `legend.spacing = *some largish number*`, I expect the spacing between the legend keys to increase relative to the default. Note that in the plot below, we can dispense with the `theme()` call and get exactly the same plot.

``` r
library(ggplot2)

p <- ggplot(mpg, aes(displ, cty)) +
  geom_point(aes(colour = factor(cyl))) +
  guides(colour = guide_legend(byrow = TRUE)) # we need byrow here, see #4352

p + theme(legend.spacing = unit(1, "cm"))
```

Instead, I would have expected this:

``` r
p + theme(legend.spacing.y = unit(1, "cm"))
```

Contrary to many issues posted about theme inheritance, the child element is truly empty in the default theme.

``` r
theme_get()$legend.spacing.y
#> NULL
```

<sup>Created on 2022-09-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This is not a hard fix, and I'm happy to take it along in #4879, but it would be a visual change.\* Would the maintainers agree to adjust this?

\* It would be a visual change because currently, `legend.spacing.{x/y}` is typically `NULL` in the default theme, which sets the spacing relative to the title text in the lines below (which would no longer occur if we adjust this):

https://github.com/tidyverse/ggplot2/blob/a58b48c961cb391b8646bf072b6620a0c9f3d999/R/guide-legend.r#L337-L338

teunbrand:
This PR aims to fix #4352 and fix #4977 and entails a visual change.

To recap the issue: whether the `legend.spacing.{x/y}` was adhered to between keys was strangely dependent on the `guide_legend(byrow = ...)` setting. This made setting vertical spacing very unintuitive. With guides having been recently rewritten, there is no better time to justify visual changes.

This PR currently does three things:
* The `legend.spacing.{x/y}` inherits from the `legend.spacing` theme setting.
* The effect of `byrow` on spacing is nullified.
* It discriminates the text-to-something spacing, from key spacing.

There are a few things I'd like some input on:
* Should we expose the text-to-something spacing to users, as a `legend.spacing.text` theme setting? Note that for finetuning text-to-something spacing, the `margin` argument is available.
* Under defaults, the proposed change is aggressively noticeable. Should we change some theme settings to come closer to the previous plots?

This PR breaks a bunch of visual tests, so I've not committed these changes yet in anticipation of further changes.

A visual example of the new spacing:

``` r
#> ℹ Loading ggplot2

  guides(colour = guide_legend(ncol = 2))
p
```

Removing the spacing brings the plot closer to previous plots.

``` r
p + theme(legend.spacing.y = unit(0, "pt"))
```

<sup>Created on 2023-04-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/guide-legend.R
+++ b/R/guide-legend.R
@@ -435,12 +435,13 @@ GuideLegend <- ggproto(
     }
 
     # Spacing
-    gap <- title$size %||% elements$theme.title$size %||%
+    tgap <- title$size %||% elements$theme.title$size %||%
       elements$text$size %||% 11
-    gap <- unit(gap * 0.5, "pt")
-    # Should maybe be elements$spacing.{x/y} instead of the theme's spacing?
-    elements$hgap <- width_cm( theme$legend.spacing.x %||% gap)
-    elements$vgap <- height_cm(theme$legend.spacing.y %||% gap)
+    tgap <- unit(tgap * 0.5, "pt")
+    elements$hgap <- width_cm( elements$spacing.x %||% tgap)
+    elements$vgap <- height_cm(elements$spacing.y %||% tgap)
+    elements$tgap <- convertUnit(tgap, "cm", valueOnly = TRUE)
+
     elements$padding <- convertUnit(
       elements$margin %||% margin(),
       "cm", valueOnly = TRUE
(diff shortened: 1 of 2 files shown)


thomasp85 (changes requested):
We should definitely not introduce such a major visual change, so while the intent of the PR is good, we should bring the default behaviour back in line with the current.

As for exposing the "text" gap I'm in general against because it seems like a difficult thing to teach users, but if you can make a motivated example of where it may be needed please show it

teunbrand:
Closing this in favour of #5456

Issue #3951, opened 2020-04-18
Suggestion: `rel`-class can inherit value from `unit`-class theme elements


teunbrand:
I know `rel` objects are not really supported as theme elements on their own, but since the `size` argument in `element_line/rect/text()` can take a `rel` object and inherit the (modified) value from the parent, it doesn't seem like too far a stretch to have child `rel` theme elements.

Suppose you'd like your legend keys to be half as tall as usual, it would be neat if you could do the following:

``` r
library(ggplot2)

ggplot(iris, aes(Sepal.Length, Sepal.Width)) +
  geom_point(aes(colour = Species)) +
  theme(legend.key.height = rel(0.5))
#> Error: Theme element `legend.key.height` must be an `unit` object.
```
<sup>Created on 2020-04-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think it would require changes at these locations.

* This check would have to allow `rel` classes for `unit` elements

https://github.com/tidyverse/ggplot2/blob/7e6d1259237d9a2b48b8156e62ada2c47c411f31/R/theme.r#L530-L533

* This might need an extra logical branch for `rel`-`rel` and `rel`-`unit` elements  

https://github.com/tidyverse/ggplot2/blob/7e6d1259237d9a2b48b8156e62ada2c47c411f31/R/theme.r#L665-L668

* Don't allow `rel` objects as 'root'-nodes in the element tree. I think the most appropriate spot to do this would be in `validate_element()`.

https://github.com/tidyverse/ggplot2/blob/40f67518a8d0ee8d7b6a1d1851c206656864fac8/R/theme-elements.r#L508

Thanks for considering!

thomasp85:
I don't see any problem in allowing this, though I can not say when we will add it

PR #5287, opened 2023-04-26
Minor ticks

### LINKED ISSUE #4387: New feature request: ticks for minor breaks
On some graphs, in particular when grids are not drawn, it would be relevant to add ticks for minor breaks.

 Similar to `annotation_logticks()`, it could be relevant to have a function `annotation_minorticks()` to draw such minor ticks, or something similar.

teunbrand:
This PR aims to fix #4387.

Briefly, it adds 3 arguments to `guide_axis()`:
* `minor.ticks` which takes an `element_{line/blank}` or `TRUE/FALSE` to draw minor ticks or not.
* `minor.length` to control the length of the minor ticks, relative to the theme setting.
* `major.length` to control the length of the major ticks, relative to the theme setting.

Less briefly:
I wasn't particularly keen to add the whole parade of `axis.ticks.minor.x.top` and `axis.ticks.length.minor.x.top` etc. theme settings for the minor axes. Mostly because I was being lazy, but also the theme arguments are busy enough as it is. The current options provide the same flexibility though.

I found out that the axis ticks were anchored incorrectly, which you of course don't see if you have one tick size. As a result of fixing this, the x/y xend/yend order in the colourbar/bin guides' ticks are swapped, which is the cause of all the changes in the visual tests (the alternative was to do this in the axis, which would affect *every* plot). Note that this affects the svgs, but not what is drawn to screen.

Here is a bit of a contrived example to show the minor ticks, and that they could also work for (fake) discrete axes, if only they had `get_breaks_minor()` implementations. Turning the major ticks off gives the discrete labels nice 'fences' between their areas on the plot.

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(match(class, sort(unique(class))), displ, group = class)) +
  geom_boxplot() +
  # Fake discrete axis
  scale_x_continuous(
    breaks = 1:length(unique(mpg$class)),
    labels = sort(unique(mpg$class)),
    guide = guide_axis(minor.ticks = TRUE, major.length = 0)
  ) +
  guides(y = guide_axis(minor.ticks = element_line(colour = "red"))) +
  theme(axis.ticks.length = unit(8, "pt"))
```


thomasp85:
I would very much suggest we move all these visual settings to the theming. It is where they belong even if the arguments feel a bit much. `axis.ticks.minor` would default to `element_blank()` and the rest inherits from that

We are kinda stuck with theming being applicable from both the guide constructor and theme with some of the old guides but there is no reason to repeat old mistakes :-)

teunbrand:
Yeah I can see why they should live in the theme. The only problem I could foresee is that at the time the information is taken from the scale (in `Coord$train_panel_guides()`), it is unknown whether minor ticks should be drawn. We could of course always extract the minor positions, but I'm not sure (yet) what the performance implications of that are.

teunbrand:
Giving this a second thought, it is not immediately clear to me what the inheritance scheme should be.

For the axis ticks do we do it like this:
```
# Pros: can set all minor ticks on/off in one go
# Cons: most of the time, you want your minor ticks to be styled like your major ticks.

             axis.ticks
            /          \
axis.ticks.x           axis.ticks.minor
     |                       |
axis.ticks.x.top       axis.ticks.minor.x
                             |
                       axis.ticks.minor.x.top
```
or like this:
```
# Pros: directly inherit style from major ticks
# Cons: cannot set all minor ticks at once

axis.ticks
    |
axis.ticks.x
    |
axis.ticks.x.top
    |
axis.ticks.x.top.minor
```
For the tick lengths, I think the following makes more sense, since there isn't any style to inherit:
```
                  axis.ticks.length
                  /               \
axis.ticks.length.x           axis.ticks.length.minor
        |                            |
axis.ticks.length.x.top       axis.ticks.length.minor.x
                                     |
                              axis.ticks.length.minor.x.top
```

teunbrand:
The minor ticks currently need to be set `TRUE`/`FALSE` at the guide level, but styling comes from the theme.

Couple thoughts:

* Ideally, the minor tick length should be a fraction of the major tick length. We'd need to implement #3951 to do this properly, so that the minor tick lengths can inherit from their major tick equivalents.

* A neat thing we could do, is to derive from the scale whether the user had supplied minor breaks, and if they did, draw minor breaks. That'd be a visual change though.

* We might let the minor tick setting be purely derived from the theme, but we'd need so provide the theme through the `coord$setup_panel_guides()` method.

thomasp85:
Did you ever look into the performance implications of always extracting minor tick positions? I have a feeling it is neigh zero...

Also, should we look into #3951 before merging this?

teunbrand:
> performance implications of always extracting minor tick positions

Currently, extracting minor positions too is about 2x as expensive as just major positions: it goes from ~1ms to ~2ms for guide training.

> Also, should we look into https://github.com/tidyverse/ggplot2/issues/3951 before merging this?

Yeah let's try

teunbrand:
I was surprised to learn that apparently ggplot2's theme system works perfectly fine when a `el_def()` has multiple parents, despite having no elements with multiple parents.

In any case, now the leaf `axis.minor.ticks.length.{aes}.{position}` nodes have `axis.minor.ticks.length.{aes}` as one parent, and `axis.ticks.length.{aes}.{position}` (the major equalent) as the other parent. This allows us to inherit a `<rel>` through the first parent, and resolve this as a proper `<unit>` through the second parent. I'm pretty happy about this.

Small demo

``` r
#> ℹ Loading ggplot2

minor_axis <- guide_axis(minor.ticks = TRUE)

  geom_point() +
  guides(x = minor_axis, x.sec = minor_axis,
         y = minor_axis, y.sec = minor_axis) +
  theme(
    axis.ticks.length = unit(1, "cm"),
    axis.ticks.length.y = rel(0.5),
    axis.minor.ticks.length = rel(0.5),
    axis.minor.ticks.length.x.top = unit(1, "cm"),
    axis.minor.ticks.length.y.right = rel(4)
  )
```



<sup>Created on 2023-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/guide-axis.R
```diff
@@ -100,9 +103,26 @@ GuideAxis <- ggproto(
...
+    major <- Guide$extract_key(scale, aesthetic, ...)
+    if (!minor.ticks) {
+      return(major)
+    }
+    if (!is.null(major)) {
+      major$.type <- "major"
+    }
+    minor <- setdiff(scale$get_breaks_minor(), major$.value)
+    new_scale <- ggproto(NULL, scale, breaks = minor, get_labels = .no_labels)
+    minor <- Guide$extract_key(new_scale, aesthetic, ...)
```
thomasp85, line 121: Is building up a whole new guide object really the best way to go about this? (asking in earnest). Wouldn't the above suggest the need for an additional method in `Guide`, or a switch in `Guide$extract_key()`

R/guide-axis.R
```diff
@@ -280,7 +291,33 @@ GuideAxis <- ggproto(
...
+      vec_slice(key, (key$.type %||% "major") == "major"),
+      elements, params, position,
+      elements$major_length
+    )
+
+    if (!params$minor.ticks || inherits(elements$minor, "element_blank")) {
+      return(major)
+    }
+
+    elements$ticks <- elements$minor
```
thomasp85, line 306: Same as the above with the new object. I think it is a code smell if it is necessary to rename/overwrite inputs and call it multiple times. Can you see a way out of this?

R/guide-axis.R
```diff
@@ -314,9 +351,21 @@ GuideAxis <- ggproto(
...
+      range <- range(
+        range, convertUnit(elements$minor_length, "cm", valueOnly = TRUE)
+      )
+    }
+
+    length <- unit(range[2], "cm")
+    spacer <- max(unit(0, "pt"), unit(-1 * diff(range), "cm"))
+
+    # Text
+    labels <- do.call(unit.c, lapply(grobs$label, measure))
```
thomasp85, line 368: Use `inject()` - and if possible keep the computation (`lapply(...)`) out of the `inject` call to avoid complicated debugging situations

thomasp85 (changes requested):

R/guide-axis.R
teunbrand: This doesn't build a new guide, it uses the parental `extract_key()` method.
It's equivalent to `ggproto_parent(Guide, self)$extract_key()`, except guides are stateless, so we don't need that mechanism.
The copy-and-modify of the scale is a bit redundant here, but we might need it later for #5410. Anyway I've worked around t hat for now.

R/guide-axis.R
teunbrand: Yeah `Guide$build_ticks()` can now also accept a single element as its `elements` argument, so we don't need this renaming thing.

R/guide-axis.R
teunbrand: Found an even simpler method :)

thomasp85 (approved):
LGTM

PR #5289, opened 2023-04-26
Axis caps

### LINKED ISSUE #4907: Natively Cap the Axis Lines, as featured on the lemon package (feature)
---- Feature Request:  Natively Cap the Axis Lines ----

It would be amazing to have the option to cap the axis lines, such as featured on the `lemon` package. The `lemon` package, however, requires to remove the ggplot2 default axis lines so that new ones can be drawn:
>  In order to manipulate the axis lines, they must be drawn. Modify the theme so the panel.border is not drawn (it will be on top of the axis lines), and have the axis lines drawn.

Therefore, it would be nice being able to cap the axis lines natively (without getting rid of the original axis lines).

Please find below a reprex of an example chart with capped axis lines:

```r
# one can install the `lemon` package from CRAN.
# install.packages('lemon')
library(tidyverse)
library(lemon)

set.seed(2)
letters_data <- purrr::map_dfr(
  letters,
  ~ data.frame(
    idx = 1:400,
    value = cumsum(runif(400, -1, 1)),
    type = .,
    flag = sample(c(TRUE, FALSE), size = 400, replace = TRUE),
    stringsAsFactors = FALSE
  )
)

letters_data %>% 
ggplot() +
  geom_line(aes(idx, value, colour = type)) +
  coord_capped_cart(bottom = "both", 
                    left = "both",
                    ylim = c(-50,25)) +
  theme_classic() + 
  theme(legend.position="none",
        axis.ticks.length = unit(.25, "cm"), 
        axis.line = element_line()) + 
  coord_capped_cart(bottom=capped_horizontal(), 
                    xlim = c(0, 400),
                    ylim = c(-40, 20),
                    left=capped_vertical(capped = "both"))
```
image

teunbrand:
This PR aims to fix #4907.

Briefly, it adds a `cap` argument to `guide_axis()` that truncates the axis line to the breaks. There are 4 different options, presented below. Also `TRUE/FALSE` are shortcuts for `"both"/"none"`.

I've thought about allowing user-defined coordinates as the capping points, either by function or input on data values, but that'd complicate the implementation quite a bit. I also thought that this is probably simple enough not to separate out to a separate guide.

``` r
#> ℹ Loading ggplot2

ggplot(mtcars, aes(hp, disp)) +
  geom_point() +
  theme(axis.line = element_line()) +
  guides(
    x = guide_axis(cap = "both"),
    y = guide_axis(cap = "upper"),
    y.sec = guide_axis(cap = "lower"),
    x.sec = guide_axis(cap = "none")
  )
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,8 @@
     * More informative error for mismatched 
      `direction`/`theme(legend.direction = ...)` arguments (#4364, #4930).
     * `guide_coloursteps()` and `guide_bins()` sort breaks (#5152).
+    * `guide_axis()` gains a `cap` argument that can be used to trim the
+      axis line to extreme breaks (#4907).
 
 * `geom_label()` now uses the `angle` aesthetic (@teunbrand, #2785)
 * 'lines' units in `geom_label()`, often used in the `label.padding` argument, 
--- a/R/guide-axis.R
+++ b/R/guide-axis.R
@@ -14,6 +14,11 @@
 #' @param n.dodge The number of rows (for vertical axes) or columns (for
 #'   horizontal axes) that should be used to render the labels. This is
 #'   useful for displaying labels that would otherwise overlap.
+#' @param cap A `character` to cut the axis line back to the last breaks. Can
+#'   be `"none"` (default) to draw the axis line along the whole panel, or
+#'   `"upper"` and `"lower"` to draw the axis to the upper or lower break, or
+#'   `"both"` to only draw the line in between the most extreme breaks. `TRUE`
+#'   and `FALSE` are shorthand for `"both"` and `"none"` respectively.
 #' @param order A positive `integer` of length 1 that specifies the order of
 #'   this guide among multiple guides. This controls in which order guides are
 #'   merged if there are multiple guides for the same position. If 0 (default),
--- a/R/guide-axis.R
+++ b/R/guide-axis.R
@@ -239,10 +239,17 @@ GuideAxis <- ggproto(
 
   # The decor in the axis guide is the axis line
   build_decor = function(decor, grobs, elements, params) {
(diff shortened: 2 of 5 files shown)


grantmcdermott:
+1ing as a user who has been looking forward to this capability for a while :-)

OTOH am I correct in thinking that, even with this PR and https://github.com/tidyverse/ggplot2/pull/4879, there still isn't a way to enable this capping behaviour globally? E.g. You still can't add guides as part of `theme_set()`, right?

teunbrand:
That's right, I haven't added this as a theme option, but it might be something to consider. You can of course do the following:

```r
capped_axes <- guides(
    x = guide_axis(cap = "both"),
    y = guide_axis(cap = "upper"),
    y.sec = guide_axis(cap = "lower"),
    x.sec = guide_axis(cap = "none")
  )
  
p1 <- ggplot(...) + ... + capped_axes
P2 <- ggplot(...) + ... + capped_axes
# etc.
```

grantmcdermott:
Thanks, your suggestion is effectively what I do now (with your very nice ggh4x package). But I do think there’s additional value in make it “themeable”. Adding my support to your consideration.

teunbrand:
I think it might make more sense to fold in #4271 first before adding this as a theme setting, that way you could set default guides in the theme.

grantmcdermott:
SGTM, thanks Teun.

thomasp85 (approved):
LGTM

teunbrand:
The way guides are implemented, that is not an easy feature to incorporate.
The capping is calculated before axes get access to theme settings, so at that point it is too late to truncate.
At the point axes get access to theme settings, it no longer has access to the scales/coord to rescale or deform the outermost positions.

PR #5290, opened 2023-04-26
Ignoring `AsIs` objects

### LINKED ISSUE #5142: Circumventing scales
To recap the discussion of #4821, passing (unmapped) aesthetics with length > 1 outside `aes()` is not ideal. However, it is currently necessary to bypass mapping values to a scale. Hence, there seems to be a need to pass values parallel to the data that circumvent scales, that isn't passed outside `aes()`.

This brings us to a discussion about what we can do to more formally support circumventing scales. In https://github.com/tidyverse/ggplot2/issues/4821#issuecomment-1371178925, I suggested that scales should ignore `AsIs` objects / `I()` input.

The current state is that the scale type for `AsIs` objects default to `scale_*_identity()`. In simple situations, the `AsIs` class already does what we want: we don't want to map it to a scale, so we use the identity scale to keep the values verbatim. The problem with this situation arises when another mapping, in another layer, *does* require a non-identity scale. There are no solutions for this problem in vanilla ggplot2.

Because there is a need to pass verbatim values, doing so outside `aes()` is discouraged, and the identity scale solution has unacceptable drawbacks, I think it might make sense to have scales explicitly ignore `AsIs` objects.

Here are the upsides:

* The semantic meaning of `I()` would change very little. It still means 'pass these values verbatim'.
* They are proper aesthetics, so data parallelism would be ensured (#4821) because they're evaluated during `Layer$compute_aesthetics()` and not `Geom$use_defaults()`.
* No need for `scale_*_identity()` to get involved, so no clashes with other scales.

The drawbacks are:

* Need to rewrite scale methods to ignore `AsIs` objects on purpose, which might have some small overhead.
* There might be a rare case in which people redundantly specificy e.g. `aes(size = I(value))` and use `scale_size_identity()` simultaneously in order to construct a legend. While I'm not certain it poses a problem, but we might discourage this by having `scale_*_identity()` explicitly warn when it sees `AsIs` objects. 

Ambiguities:

* There is no `scale_{x/y}_identity()`, so what should happen if people use `aes(x = I(var))`? Should they be interpreted as `npc` units between 0-1, just like after coord transformation? Should they just behave as regular positions?

I'd be happy to discuss, and, if people see some merit in this, prepare a PR to explore implementation.

teunbrand:
This PR aims to fix #5142.

Briefly, all scales ignore `AsIs` objects instead of invoking an identity scale.

Contrary to my expectation, I only had to edit `ScalesList` methods (and their position equivalents) and no individual scales, so no deep rewrite was needed.
While in #5142 I expressed worry about how position aesthetics would play with this, I've grown to love ignored position aesthetics. No more `annotation_custom()` or awkward manual transformations if you want to make relative annotations!

``` r
#> ℹ Loading ggplot2

df <- data.frame(t = seq(0, 2*pi, length.out = 100))

ggplot(mpg) +
  geom_point(aes(displ, hwy)) +
  geom_path(
    data = df, 
    aes(x = I(cos(t) / 2.1 + 0.5), 
        y = I(sin(t) / 2.1 + 0.5),
        colour = I(2))
  ) +
  annotate(
    "text", x = I(0.5), y = I(0.5),
    label = "I'm in the middle",
    colour = 4, size = 10
  )
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # ggplot2 (development version)
 
+* Plot scales now ignore `AsIs` objects constructed with `I(x)`, instead of
+  invoking the identity scale. This allows these columns to co-exist with other
+  layers that need a non-identity scale for the same aesthetic. Also, it makes
+  it easy to specify relative positions (@teunbrand, #5142).
+
 * `geom_text()` and `geom_label()` gained a `size.unit` parameter that set the 
   text size to millimetres, points, centimetres, inches or picas 
   (@teunbrand, #3799).
--- a/R/facet-.R
+++ b/R/facet-.R
@@ -102,16 +102,18 @@ Facet <- ggproto("Facet", NULL,
     # loop over each layer, training x and y scales in turn
     for (layer_data in data) {
       match_id <- match(layer_data$PANEL, layout$PANEL)
+      names <- names(layer_data)
+      names <- names[!vapply(layer_data, inherits, what = "AsIs", logical(1))]
 
       if (!is.null(x_scales)) {
-        x_vars <- intersect(x_scales[[1]]$aesthetics, names(layer_data))
+        x_vars <- intersect(x_scales[[1]]$aesthetics, names)
         SCALE_X <- layout$SCALE_X[match_id]
 
         scale_apply(layer_data, x_vars, "train", SCALE_X, x_scales)
       }
 
       if (!is.null(y_scales)) {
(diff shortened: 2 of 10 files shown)


thomasp85:
As discussed we should investigate if we could instead using the same mechanics as `after_scale()` for this functionality

teunbrand:
Closing this in favour of #5477

Issue #3959, opened 2020-04-22
coord_trans/polar() ignore position guides


teunbrand:
While I was trying to making a variant of `guide_axis()` that displays tick marks at minor breaks (see for example [SO](https://stackoverflow.com/questions/61087231/ggplot2-add-minor-tick-marks-outside-plotting-area-without-turning-clip-off/)), I stumbled on a possible bug in `coord_trans()` and `coord_polar()`, which appear to ignore position guides.

Standard cartesian coordinates seem to work just fine.

``` r
library(ggplot2)

p <- ggplot(mpg, aes(hwy, cty)) +
  geom_point() +
  guides(x = guide_axis("custom axis"))

p + coord_cartesian()
```

![](https://i.imgur.com/5gUBrmR.png)

Polar coordinates don't render the correct axis title.

``` r

p + coord_polar()
```

![](https://i.imgur.com/s0k9jAX.png)

Transformed coordinates also don't render the correct axis title.

``` r

p + coord_trans()
```

![](https://i.imgur.com/owVJZH9.png)

<sup>Created on 2020-04-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The same pattern holds if I specify `scale_x_continuous(guide = guide_axis("custom axis))` or use `guides(x.sec = guide_axis("custom axis"))`.

yutannihilation:
Thanks, confirmed.

Currently, the `title` of `guide` is referenced only by `CoordCartesian$label()` via `panel_guide_label()`:

https://github.com/tidyverse/ggplot2/blob/d3d47be055055ebf328aeed0bebb990a635c49ab/R/coord-cartesian-.r#L154-L173

https://github.com/tidyverse/ggplot2/blob/d3d47be055055ebf328aeed0bebb990a635c49ab/R/coord-cartesian-.r#L217-L220

The other `Coord` classes use the inherited `Coord$label()`:

https://github.com/tidyverse/ggplot2/blob/d3d47be055055ebf328aeed0bebb990a635c49ab/R/coord-.r#L62

But, it seems we cannot simply transplant the `label()`. `panel_params` doesn't contain `guides` by default, so we need `setup_panel_guides()` to add it.

https://github.com/tidyverse/ggplot2/blob/d3d47be055055ebf328aeed0bebb990a635c49ab/R/coord-cartesian-.r#L134

Besides, `guides` needs to have `position`, otherwise `guide_for_position(guides, position)` returns `NULL`.

[...]

PR #5293, opened 2023-05-01
Guides for `coord_sf()`

teunbrand:
This PR aims to fix a problem akin to #3959, but for `coord_sf()`.

Briefly, `coord_sf()` was doing its own guide thing, ignoring any user-provided guides. 
Now, it renders the guides through the guides system.

Less briefly:
To make this work, I had to cast the graticule positions into a viewscale that the guides can use for training.
I was a bit surprised that axis placement is tightly controlled in `coord_sf()` itself, in that `scale_x_continuous(position = "top")` is ignored and just adding `guides(x.sec = "axis")` doesn't just set a top axis. 
Maybe it would be good to allow some flexibility here, but I didn't change that behaviour in this PR.
I'm new-ish to {sf}, so if anybody knows of hairy edge cases I mightn't have thought of, please let me know.

To illustrate:

``` r
#> ℹ Loading ggplot2

ggplot(nc) +
  geom_sf() +
  coord_sf(crs = 3347, label_graticule = c("NSWE")) + # all labels
  guides(
    x = guide_none("guide_none() passes on their title"),
    y = guide_axis("angles work", angle = 45),
    x.sec = guide_axis("dodging works", n.dodge = 2),
    y.sec = guide_axis(angle = -45, n.dodge = 2)
  )
```

<sup>Created on 2023-05-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85:
@clauswilke you have spend the most time with sf - do you have the bandwidth to look at this?

clauswilke:
@thomasp85 Yes, this touches mostly code that I wrote, so I guess I'll have to look at it carefully. I don't remember much of it, but I'll take a look.

@teunbrand Two quick items from my end:

1. Are we using `Map()` in ggplot? I thought we're generally using apply functions.
2. It looks to me like a bunch of comments have disappeared in porting the code. I would encourage you to comment a bit more generously. This is confusing code, and without comments nobody will understand what it does five years into the future.

clauswilke:
To be clear: My code already didn't have that many comments. The new code has even fewer. I think the goal should be more comments than my original code had.

teunbrand:
> Are we using Map() in ggplot? I thought we're generally using apply functions.

Under the hood, `Map()` uses the apply-family of functions but doesn't simplify output. Should there be a reason not to use it?

> I think the goal should be more comments than my original code had.

Yeah this is a fair point and had some problem deciphering which parts were doing what as well.

clauswilke:
Regarding `Map()`, I'll leave it to @thomasp85 to make this decision. As a general principle, since we're writing ggplot2 using somewhat obscure (for people coming from the tidyverse) base-R functionality, it makes sense to use a limited set of functions to keep the code easier to read. At least I always thought that we're doing this. I could be wrong.

R/coord-sf.R
```diff
@@ -235,6 +235,17 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
...
     graticule <- self$fixup_graticule_labels(graticule, scale_x, scale_y, params)
 
+    # Convert graticule to viewscales for axis guides
+    viewscales <- Map(
+      view_scales_from_graticule,
+      scale = list(x = scale_x, y = scale_y, x.sec = scale_x, y.sec = scale_y),
+      aesthetic = c("x", "y", "x.sec", "y.sec"),
+      label = self$label_axes[c("bottom", "left", "top", "right")],
+      MoreArgs = list(graticule = graticule, bbox = bbox,
+                      label_graticule = self$label_graticule)
```
clauswilke, line 245: Just FYI: I think the formatting here is not tidyverse style.

R/coord-sf.R
```diff
@@ -716,3 +575,141 @@ parse_axes_labeling <- function(x) {
...
+  # Determine what columns in the graticule contain the starts and ends of the
+  # axis direction and the orthogonal direction.
+  axis_start <- paste0(axis, "_start")
+  axis_end   <- paste0(axis, "_end")
+  orth_start <- paste0(orth, "_start")
+  orth_end   <- paste0(orth, "_end")
+
+  # Find the start and endpoints in the graticule that are in close proximity
+  # to the axis position to generate 'accepted' starts and ends. Close proximity
+  # here is defined as within 1% of the scale range of the *orthogonal* scale.
```
clauswilke, line 632: I think you're using .1%, not 1%.

clauswilke (approved):
I've reviewed it and had two more minor comments. Overall this seems fine to me, though I will say I don't fully understand the guide system. In the long run, this will likely be more maintainable, though, so moving forward in this direction is good.

teunbrand:
Thanks for the review Claus! Yeah I suppose the guide system will take some getting used to, but I'd be happy to try to answer any questions you might have.

R/coord-sf.R
teunbrand: Good catch.

R/coord-sf.R
teunbrand: Whoops, yes you're right.

PR #5296, opened 2023-05-02
Improve error message when `plot.title.position` is not correctly mentioned.

olivroy:
Before this PR

```r
ggplot() +theme(plot.title.position = "Null")
Error in `ggplot_gtable()`:
! `plot.title.position` should be either
  "\"panel\"" or "\"plot\"".
```
With this PR

```r
ggplot() +theme(plot.title.position = "Null")
Error in `ggplot_gtable()` 
! `plot.title.position` should be either "panel" or "plot".
```

I wonder how we could better improve this message, so it becomes:

```r
ggplot() +theme(plot.title.position = "Null")
Error in `theme()` 
! `plot.title.position` should be either "panel" or "plot".
```

But I don't know how to do that.

Edit: I added suggestion to use `rlang::arg_match()` to get rid of the `if ()`, but this doesn't solve the error call.

### DIFF
--- a/R/plot-build.R
+++ b/R/plot-build.R
@@ -276,16 +276,19 @@ ggplot_gtable.ggplot_built <- function(data) {
   # positioning of caption is governed by plot.caption.position
   #   "panel" means align to the panel(s)
   #   "plot" means align to the entire plot (except margins and tag)
-  title_pos <- theme$plot.title.position %||% "panel"
+  title_pos <- arg_match0(
+    theme$plot.title.position %||% "panel",
+    c("panel", "plot"),
+    arg_nm = "plot.title.position",
+    error_call = expr(theme())
+  )
 
-  if (!(title_pos %in% c("panel", "plot"))) {
-    cli::cli_abort("{.var plot.title.position} should be either {.val panel} or {.val plot}.")
-  }
-
-  caption_pos <- theme$plot.caption.position %||% "panel"
-  if (!(caption_pos %in% c("panel", "plot"))) {
-    cli::cli_abort("{.var plot.caption.position} should be either {.val panel} or {.val plot}.")
-  }
+  caption_pos <- arg_match0(
+    theme$plot.caption.position %||% "panel",
+    values = c("panel", "plot"),
+    arg_nm = "plot.caption.position",
+    error_call = expr(theme())
+  )
 
   pans <- plot_table$layout[grepl("^panel", plot_table$layout$name), , drop = FALSE]
   if (title_pos == "panel") {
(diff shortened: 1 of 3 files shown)


olivroy (commented):
This could be replaced with 

```r
arg_match(title_pos, c("panel", "plot"), error_arg = "plot.title.position")

**Comment by olivroy** [comment]
Hi,

Thanks for the suggestion.

Would you like me to update my PR and snapshots with this instead?

```r
# new
arg_match0(
  title_pos, c("panel", "plot"), 
  arg_nm = "plot.title.position", 
  error_call = expr(theme())
)
  # delete
 if (!(title_pos %in% c("panel", "plot"))) {
    cli::cli_abort("{.var plot.title.position} should be either {.val panel} or {.val plot}.")
}
```

teunbrand:
That'd be great, thanks! `plot.caption.position` might also benefit from this treatment.

olivroy:
Just a note, 

globally, the Error in ... is not tested.

```r 
 ggplot(mtcars) + geom_point(aes(disp, mpg)) + theme(plot.title.position = "test")
Error in `theme()`:
! `plot.title.position` must be one of "panel" or "plot", not
  "test".
```
It is the new error message. but I tried another thing for example with plot.background.

```r
ggplot() + theme(plot.background = "w")
Error in `mapply()` 
! The `plot.background` theme element must be a <element_rect>
  object.
```
This is not very informative!

teunbrand:
Yes I think ggplot2 could do a little bit better in reporting where the error has originated sometimes. To be fair, ```! The `plot.background` theme element must be a <element_rect> object.``` is a good error message, I think. The bad part is the ```Error in `mapply()``` bit. However, I think that a witch hunt for this ambiguous origin reporting is perhaps better for a separate PR, so I wouldn't worry too much about this for this PR. If you think this is important, you could open an issue about it so that we can discuss what should be done.

teunbrand (approved):
Looks good to me

teunbrand:
I think the failing test on R devel is unrelated to this PR.

From the [R news](https://cran.r-project.org/doc/manuals/r-devel/NEWS.html) section:

> density() more consistently computes grid values for the FFT based convolution, following Robert Schlicht's analysis and proposal in [PR#18337](https://bugs.r-project.org/show_bug.cgi?id=18337), correcting density values typically by a factor of about 0.999. Optional old.coords=TRUE provides back compatibility.

Which causes `geom_violin()` (and presumably other density based stats) to change across versions. I don't know if there is a CI solution for this, but for now I think it's OK to merge this PR.

@olivroy Thanks for the work you've put into this PR!

Issue #3997, opened 2020-05-15
Integrating new grid features in ggplot2


thomasp85:
Hi Team

The graphic engine and grid is getting some love lately and we will look into a future where gradient + image fill and non-rectangular clipping is available on the devices that supports it. I think both have interest from the ggplot2 side.

While we are not there yet, @pmur002 has a working implementation and I think we may just as well begin discussing to what extend and how we may use this in ggplot2. I think the most interesting (seeing as ggpattern exist) possibility is gradient fill. This will most definitely be a challenge to integrate into the scale system because gradients depend on multiple values. Should the scaling happen from discrete values to a set of predefined gradients? or should each stop be mapped to different values. Lots of things to discuss!

Non-rectangular clipping also opens up a lot of possibilities, some not having a clear parallel in the current API (e.g. clipping groups or layers to certain areas)

@clauswilke @karawoo @Hadley @yutannihilation @paleolimbot please join the discussion 🙂

paleolimbot:
The geologists are going to be so excited! My quick unqualified comment would be that a vectorized "fill spec" (probably generated by `ScalePattern$map()`) might do the trick. How that interacts with the colour and fill aesthetics seems tricky.

clauswilke:
I think allowing `fill` to vary along x or y in polygons would be very useful, and would probably be something people would expect. As seen here: https://wilkelab.org/ggridges/reference/geom_ridgeline_gradient.html

pmur002:
If you want to play along, this is all happening in an svn branch [https://svn.r-project.org/R/branches/R-defs/], though that is VERY unstable, both in the sense of being a segfault magnet and in the sense of changing all the time.  I am also documenting progress [https://www.stat.auckland.ac.nz/~paul/Reports/r-defs-proposal/defs-proposal.html], though that also needs updating already.

clauswilke:
Paul, this is great! Do you take feature requests? One feature that would be really useful and probably isn't that much work (famous last words!) is some sort of link or anchor. Obviously it'd have to be ignored in raster devices, but it'd be super useful for svg or pdf output.

pmur002:
[...]

PR #5299, opened 2023-05-06
Barebones support for `<GridPattern>` fills.

teunbrand:
This PR sprang forth from #3997.

Briefly, it attempts to provide some barebones support for {grid}'s new pattern features.

Less briefly:
Essentially, the only piece of ggplot2's internals that hampered using grid's patterns was the `alpha()` function that has restrictions on what it considers a colour. To lift this restriction, there is now a `fill_alpha()` function exported (for extension devs), that can set alpha on any of {grid}'s 3 pattern classes.

What can you do with this PR:
* Set `geom_rect(fill = <(list of) GridPattern>)` outside the `aes()`.
* Set `geom_rect(aes(fill = something)) + scale_fill_manual(values = <list of GridPattern>)`.

What this PR doesn't do:
* Provide scale infrastructure for mapping patterns.
* Provide additional aesthetics for patterns.

I'm also pinging @trevorld and @coolbutuseless because they might have some opinions about this PR and experience with patterns.

A small demo:

``` r
library(grid)
#> ℹ Loading ggplot2

# Borrowed straight from the README of https://github.com/clauswilke/ggtextures
images <- c(
  "http://www.hypergridbusiness.com/wp-content/uploads/2012/12/rocks2-256.jpg",
  "http://www.hypergridbusiness.com/wp-content/uploads/2012/12/stone2-256.jpg",
  "http://www.hypergridbusiness.com/wp-content/uploads/2012/12/siding1-256.jpg"
)

patterns <- lapply(images, function(url) {
  img <- grid::rasterGrob(
    magick::image_read(url),
    width  = unit(2, "cm"),
    height = unit(2, "cm")
  )
  grid::pattern(img, width = img$width, height = img$height, extend = "repeat",
                group = FALSE)
})

ggplot(mpg, aes(class)) +
  geom_bar(aes(fill = factor(drv)), colour = 'black') +
  scale_fill_manual(values = patterns)
```

<sup>Created on 2023-05-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/utilities-patterns.R
```diff
@@ -0,0 +1,102 @@
...
+# Similar to grid:::is.pattern
+is_pattern <- function(x) {
+  inherits(x, "GridPattern")
+}
+
+# Function that applies alpha to <GridPattern> objects.
+# For linear or radial gradients, this is as simple as modifying their `colours`
+# slot with an alpha.
+# For tiled patterns, we attach an alpha mask in the grobs' viewport.
+pattern_alpha <- function(x, alpha) {
```
teunbrand, line 69: Should this be an S3 generic + methods?

R/utilities-patterns.R
```diff
@@ -0,0 +1,102 @@
...
+    mask <- as.mask(rectGrob(gp = gpar(fill = alpha("black", alpha[1]))))
+    if (is.null(grob$vp)) {
+      grob$vp <- viewport(mask = mask)
+    } else {
+      grob$vp$mask <- mask
+    }
+    # Re-attach new function environment
+    new_env <- new.env(parent = environment(x$f))
+    env_bind(new_env, grob = grob)
+    environment(x$f) <- new_env
```
teunbrand, line 99: This 'feels' dirty, but I can't see how to apply an alpha to tiled patterns otherwise.

trevorld: An alternative may be instead of extracting the `grob` you extract the `gp` and update the alpha in that `gpar()` object...  If you do extract the `grob` you could also do `grob = grobTree(grop, gp = gpar(alpha = ...`

If you are going to alpha mask it is a good idea to make sure the color choice works if the alpha mask is mis-interpreted as a luminance mask (i.e. the popular program `inkscape` does such a mis-interpretation).  Maybe "black" works but maybe "white" is a better choice.

R/utilities-patterns.R
```diff
@@ -0,0 +1,105 @@
...
+    return(alpha(fill, alpha))
+  }
+  if (any(vapply(fill, is_pattern, logical(1)))) {
+    if (utils::packageVersion("grid") < "4.2") {
+      # Pattern fills were introduced in R 4.1.0, but *vectorised* patterns
+      # were only introduced in R 4.2.0.
+      cli::cli_abort(
+        "Using patterns in {.pkg ggplot2} requires at least R version 4.2.0."
+      )
+    }
```
trevorld, line 37: Would it make sense to throw a warning if the active graphics device doesn't support these patterns?

Something like the equivalent of `gridpattern::guess_has_R4.1_features(c("gradients", "patterns"))`.  In the future this should simply be `all(c("LinearGradient", "RadialGradient", "TilingPattern") %in% dev.capabilities()$patterns)` once `{ragg}` and `{svglite}` are updated to pass this info to `dev.capabilities()`.

R/utilities-patterns.R
teunbrand: I think the 'dirty' part is rummaging through function environments to make edits. Setting `gpar(alpha = ...)` works for some grobs, but unfortunately not for others, like `rasterGrob()`. Using "white" is a good suggestion, I hadn't quite considered this choice.

R/utilities-patterns.R
teunbrand: Thanks, yeah that is a good suggestion. I've had a look at `guess_has_R4.1_features()` and kudos to you for being so careful to your users to properly check the device's capabilities! I've no issue just rejecting ragg/svglite until they report their capabilities properly though.

R/utilities-patterns.R
trevorld: If using alpha masking for "clipping" white is definitely better than black.  Not sure for non-1 alpha values.  Perhaps instead of "white" it could be that `rgb(alpha, alpha, alpha)` is better?  https://github.com/r-lib/svglite/issues/146

teunbrand: Just tried this, and it sort of makes sense for inkscape, but it doesn't quite correspond to what one would see with e.g. png exports. I'm a bit worried that with the `rgb(alpha, alpha, alpha)` approach we'd be fooling users into thinking their plot has rendered correctly in inkscape (which, as you've pointed out, it doesn't because it uses luminance masks).

trevorld: * Here is some documentation saying that the alpha value for a luminance mask should be equal to the luminance value times the alpha value so `alpha("white", alpha[1])` should theoretically work (i.e. alpha = 1 * alpha): https://css-tricks.com/almanac/properties/m/mask-mode/#aa-luminance-masks
* Playing around this seems to work with my version of `inkscape` (and it seems that `eog` was improved to properly interprets alpha masks as alpha masks).

teunbrand: For rasterGrob patterns, we could perhaps just multiply the transparency of the colours directly. The `rgb(alpha, alpha, alpha)` is probably the most sensible choice, as the luminance of that grayscale is not exactly but roughly equal to the input values.

trevorld: > For rasterGrob patterns, we could perhaps just multiply the transparency of the colours directly.

> For rasterGrob patterns, we could perhaps just multiply the transparency of the colours directly.

This is what `{gridpattern}` / `{ggpattern}` basically does for "array" style patterns.

> The rgb(alpha, alpha, alpha) is probably the most sensible choice, as the luminance of that grayscale is not exactly but roughly equal to the input values.

Are you sure `rgb(alpha, alpha, alpha)` works better than `alpha("white", alpha[1])` if the luminance mask value transparency is the luminosity value times the alpha value?  `alpha("white", alpha[1])` should have an equivalent alpha mask value and luminance mask value (alpha = 1 * alpha) whereas `rgb(alpha, alpha, alpha)` does not?

R/utilities-patterns.R
thomasp85: Adding this info was stalled as implementation of the new composition operators became more complex than expected. I would advice to special case them since these are our (Posits) packages, they do provide the required functionality, and we advise their use throughout

R/utilities-patterns.R
teunbrand: @thomasp85 ggplot2 doesn't import {farver} but for alpha multiplication, we'd need something like `farver::multiply_channel(x, "alpha", 0.5)`. Would it make sense to wrap this functionality in the {scales} package, that already imports farver?

R/utilities-patterns.R
trevorld: > I would advice to special case them since these are our (Posits) packages, they do provide the required functionality, and we advise their use throughout

@thomasp85 what is the recommended way to "special case" whether the active graphics device is `svglite::svglite()` (supports alpha masks and gradients) instead of `vdiffr:::svglite()` (does not support alpha masks and gradients) since for both graphics devices `names(dev.cur())` is `"devSVG"` and `grDevices::dev.capabilities()$patterns` is `NA`.

trevorld: > it would be really nice to have a function(s) exported from ggplot2 that checks for support for different graphics engine features

@mjskay I've refactored `gridpattern::guess_has_R4.1_features()` so it is now also available as a "standalone" file.
  If you don't want to import `{gridpattern}` you may copy its source under the permissive [Unlicense](https://unlicense.org) into your own R package
  by either using `usethis::use_standalone("trevorld/gridpattern", "standalone-guess_has_R4.1_features.R")`
  or simply copying `standalone-guess_has_R4.1_features.R` into your `R` directory and 
  adding `grDevices` and `utils` to the `Imports` of your `DESCRIPTION` file.

Although not perfect it goes beyond using `grDevices::dev.capabilities()` and can guess R 4.1 graphic feature support in most supported base R graphic devices in R version 4.1 as well as in `{ragg}` in R versions 4.1+ and sometimes (but not always) in `{svglite}`.

Known limitations: will conservatively but incorrectly guess `FALSE` for `svglite::svglite()` if called within `vdiffr::expect_doppelganger()` and for `png(type = "cairo")` in R version 4.1 on Windows.

p.s. In case you weren't aware `gridpattern::patternGrob(pattern = "gradient")` can provide linear gradient support (via a rasterized fallback) on graphics devices that don't support the R 4.1 gradient pattern feature including R versions before R 4.1.

teunbrand:
Alright, time for a small update.

* I've switched alpha masks for tiled patterns to use white as the base colour.
* It now uses the new `check_device()` in appropriate places.

Remaining issues:
* For tile patterns based on `rasterGrob()`, we currently do not modify the colours directly. Doing so would require `farver::multiply_channel()`, and {farver} is not a direct dependency (it is via {scales}).
* Should we make `pattern_alpha()` an S3 generic + methods?

Some examples for all pattern options:

Linear gradients work as expected.
``` r
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(drv))

grad <- linearGradient(rainbow(10))
p + geom_bar(fill = grad, alpha = 0.5)
```



Radial gradients work as expected.
``` r
grad <- radialGradient(rainbow(10))
p + geom_bar(fill = grad, alpha = 0.5)
```



Tiled patterns work as expected.
``` r
tile <- pattern(
  circleGrob(r = unit(1, "mm"), gp = gpar(fill = "black")), 
  extend = "repeat", width = unit(3, "mm"), height = unit(3, "mm")
)
p + geom_bar(fill = tile, alpha = 0.5)
```



The {gridpattern} patterns work if wrapped in `grid::pattern()`.
``` r
pattern <- pattern(gridpattern::grid.pattern_circle(
  density = 0.5, grid = "hex_circle", fill = c("blue", "yellow", "red"), 
  draw = FALSE
))
p + geom_bar(fill = pattern, alpha = 0.5)
```



Example of error message when device doesn't support feature:
``` r
png(type = "windows")
p + geom_bar(fill = pattern, alpha = 0.5)
#> Error in `geom_bar()`:
#> Caused by error in `fill_alpha()` at ggplot2/R/geom-rect.R:54:6:
#> ! Unable to check the capabilities of the png device.
dev.off()
```

<sup>Created on 2023-10-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

brunomioto:
> Alright, time for a small update.
> 
> * I've switched alpha masks for tiled patterns to use white as the base colour.
> * It now uses the new `check_device()` in appropriate places.
> 
> Remaining issues:
> * For tile patterns based on `rasterGrob()`, we currently do not modify the colours directly. Doing so would require `farver::multiply_channel()`, and {farver} is not a direct dependency (it is via {scales}).
> * Should we make `pattern_alpha()` an S3 generic + methods?
> 
> Some examples for all pattern options:
> 
> Linear gradients work as expected.
> ``` r
> devtools::load_all("~/packages/ggplot2")
> #> ℹ Loading ggplot2
> #> Warning: package 'testthat' was built under R version 4.3.1
> 
> p <- ggplot(mpg, aes(drv))
> 
> grad <- linearGradient(rainbow(10))
> p + geom_bar(fill = grad, alpha = 0.5)
> ```
> 
> 
> 
> Radial gradients work as expected.
> ``` r
> grad <- radialGradient(rainbow(10))
> p + geom_bar(fill = grad, alpha = 0.5)
> ```
> 
> 
> 
> Tiled patterns work as expected.
> ``` r
> tile <- pattern(
>   circleGrob(r = unit(1, "mm"), gp = gpar(fill = "black")), 
>   extend = "repeat", width = unit(3, "mm"), height = unit(3, "mm")
> )
> p + geom_bar(fill = tile, alpha = 0.5)
> ```
> 
> 
> 
> The {gridpattern} patterns work if wrapped in `grid::pattern()`.
> ``` r
> pattern <- pattern(gridpattern::grid.pattern_circle(
>   density = 0.5, grid = "hex_circle", fill = c("blue", "yellow", "red"), 
>   draw = FALSE
> ))
> p + geom_bar(fill = pattern, alpha = 0.5)
> ```
> 
> 
> 
> Example of error message when device doesn't support feature:
> ``` r
> png(type = "windows")
> p + geom_bar(fill = pattern, alpha = 0.5)
> #> Error in `geom_bar()`:
> #> ! Problem while converting geom to grob.
> #> ℹ Error occurred in the 1st layer.
> #> Caused by error in `fill_alpha()` at ggplot2/R/geom-rect.R:54:6:
> #> ! Unable to check the capabilities of the png device.
> dev.off()
> ```
> 
> <sup>Created on 2023-10-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
> 

Awesome update @teunbrand !

It's possible to manage the angle of linear gradients?

teunbrand:
> It's possible to manage the angle of linear gradients?

Yes, but that all happens in `grid::linearGradient()`, specifically by tweaking the `x1`,`x2`,`y1` and `y2` arguments, but that isn't something that ggplot2 should worry about.

trevorld:
Looks good!

I observe that `fill_alpha()` throws an error if you give it a length-one list with a color vector of length greater than one:

```r
> fill_alpha(list(c("blue", "red")), 0.5)
Error in `fill_alpha()`:
! fill must be a valid colour or list of <GridPattern> objects.
```

Future extension authors perhaps may want it to support color vectors greater than one?  [A future release of `{ggpattern}`](https://github.com/coolbutuseless/ggpattern/issues/100) will probably allow the `pattern_fill` aesthetic to be lists of color vectors greater than one although in this case the alpha adjustment may be handed off to `{gridpattern}`.

teunbrand:
Ah right, perhaps the error message should be clearer, but you can give it a bare vector of colours, they don't need to be in a list.

```r
fill_alpha(c("blue", "red"), 0.5)
> [1] "#0000FF80" "#FF000080"
```

trevorld:
> If you have a ragged list of plain colours (to compose your own patterns), you might as well stick to `regular scales::alpha()`. Unless you're thinking about nesting patterns in patterns?

1. `scales::alpha()` currently throws an error if you give it a length-one list with a color vector of length strictly greater than one.
2. The immediate goal is to allow `{ggpattern}` users to directly compose multi-color patterns in `{ggplot2}` without messing with custom patterns but I think I should allow nesting patterns in patterns in the future.
3.  I'm going to want this lower level stuff to happen in `{gridpattern}` which I don't want to depend on `{ggplot2}` so I'm probably not going to want to directly call `ggplot2::fill_pattern()` but I may want to copy and paste a fork of the functions (and update the DESCRIPTION/LICENSE appropriately).

teunbrand:
Alright I have given it some more consideration, but the most important thing to me is that `fill_alpha()` delivers output that is valid input for `gpar(fill = ...)`. A `<list>` of colours is not valid input for `gpar()`:

``` r
grid::gpar(fill = c("red", "blue"))
#> $fill
#> [1] "red"  "blue"
grid::gpar(fill = list("red", "blue"))
#> Error in validGP(list(...)): 'fill' gpar list components must all be patterns
grid::gpar(fill = list(list("red", "green"), "blue"))
#> Error in validGP(list(...)): 'fill' gpar list components must all be patterns
```

This PR *is* allowing mixed inputs of `list(<pattern>, <character_colour>)` for user convenience, in which case the character colour is converted to a pattern. However, geoms are free to slice the input data to their whims, which means some `Geom$draw_group()` method (or legend keys), can have `list(<character_colour>)` as input to `fill_alpha()`. To deliver compatible output with `gpar()`,  `fill_alpha()` should `unlist()` such cases. In other words, `fill_alpha()` is sacrificing type stability to minimise fuss with `gpar()`. If we can convince the {grid} package to support mixed input, it would be no problem  to handle ragged lists\* here.

\* I'm using the term 'ragged list' to mean a list where `lengths(my_list)` can be any sequence of positive integers (and 0).

Now, aside from the problem that ragged lists of colours aren't valid input for `gpar()`, the need to `unlist()` lists of colours might give unexpected results. For example, if your 'ragged' list actually turns out to be a flat list where `all(lengths(x) == 1)`, then you get a plain character vector because we need valid `gpar()` input. However if your ragged list is truly ragged, you would get a list back where `all(lengths(input) == lengths(output))` (if `fill_alpha()` was adjusted to add support for such cases). This would require some defensive programming on the extension developers part to deal with the type instability.

Lastly, and this is more of a minor point, if `fill_alpha()` were to handle a ragged list of `colour`s, then it should also support a ragged list of `alpha`s and it gets tedious to throw the appropriate errors if these are misaligned.

Thus, my advice for ragged lists is to handle these in your package according to your needs. ggplot2 is MIT licenced so you're free to adapt `fill_alpha()` for your purposes if you want to support ragged lists.

R/geom-tile.R
```diff
@@ -5,7 +5,8 @@
 #' corners (`xmin`, `xmax`, `ymin` and `ymax`), while
 #' `geom_tile()` uses the center of the tile and its size (`x`,
 #' `y`, `width`, `height`). `geom_raster()` is a high
-#' performance special case for when all the tiles are the same size.
+#' performance special case for when all the tiles are the same size, and no
+#' pattern fills are applied.
```
thomasp85, line 9: Should `geom_raster()` throw an error when encountering a pattern fill?

R/utilities-patterns.R
```diff
@@ -0,0 +1,100 @@
...
+#'   # Pattern input
+#'   fill_alpha(list(grid::linearGradient()), 0.5)
+#' }
+fill_alpha <- function(fill, alpha) {
+  if (!is.list(fill)) {
+    # Happy path for no patterns
+    return(alpha(fill, alpha))
+  }
+  if (is_pattern(fill) || any(vapply(fill, is_pattern, logical(1)))) {
+    check_device("patterns", action = "abort")
```
thomasp85, line 31: I'm not sure this should error, honestly. The function is not called at render time so it is entirely feasible and ok to call `ggplotGrob()` while a device with no support is active, then render the grab later or save it or something

R/utilities-patterns.R
thomasp85: yeah, it should :-)

R/geom-tile.R
teunbrand: It will because `alpha()` doesn't accept patterns, but I presume you mean an informative error :)

R/utilities-patterns.R
thomasp85: Yes, throwing a warning is a practical middle-ground... What happens with check_device if no device is open? I'd like for it to not force one open

R/geom-tile.R
thomasp85: yup

R/utilities-patterns.R
teunbrand: You'd get the following warning:
```r
#> Warning message:
#> Unable to check the capabilities of the null device device. 
```

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review Thomas!

PR #3649, opened 2019-12-01
WIP: show.key, a new parameter of guide_legend() to fix #3648

microly:
fix #3648 

This is only a pr draft to show my solution to #3648.
The documemtation is lack, and the codes may need improve.
Any comment is welcome!
I will add documemtation and anything needed later.

PR #5302, opened 2023-05-09
Smarter keys

### LINKED ISSUE #3648: Implement smart merging of guides so inappropriate key glyphs are omitted
Currently, if two layers have non-overlapping values in an aesthetic and different key glyphs, both glyphs are drawn for all data values, as seen here:

``` r
library(ggplot2)

  geom_point(aes(color = "points")) +
  geom_smooth(aes(color = "regression line"))
```

This is suboptimal, because the line glyph is not sensible for the "points" entry and the point glyph is not sensible for the "regression line" entry. The workaround currently is to use `override.aes()`, but it is cumbersome and requires deep knowledge about the default settings for various key glyphs:

``` r

  geom_point(aes(color = "points")) +
  geom_smooth(aes(color = "regression line")) +
  guides(
    color = guide_legend(
      override.aes = list(
        shape = c(19, NA),
        linetype = c(NA, 1),
        fill = c(NA, "grey60")
      )
    )
  )
```

There should be an option, and maybe made the default, to produce the second outcome whenever appropriate.

teunbrand:
This PR aims to fix #3648. This PR is mutually incompatible with #3649.

Briefly, it automatically checks whether the key value is actually in the data and doesn't draw keys when it isn't. This contrasts with the approach taken in #3649, where the user is given more control.

Less briefly: it performs this check when `show.legend = NA`, which is the default. When `show.legend = TRUE`, is draws the keys regardless of presence in the layer data. To get back the old behaviour and you have multiple legends, you'd have to be specific to include a layer in the keys of a particular legend, for example `show.legend = c(colour = TRUE)`.

A small demo from https://github.com/tidyverse/ggplot2/issues/3648#issue-530627481. Note how there is no point displayed for the 'regression line' key, and no line for the 'points' key.

``` r
#> ℹ Loading ggplot2

  geom_point(aes(color = "points")) +
  geom_smooth(aes(color = "regression line"))
#> `geom_smooth()` using method = 'loess' and formula = 'y ~ x'
```

<sup>Created on 2023-05-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/plot-build.R
+++ b/R/plot-build.R
@@ -89,6 +89,7 @@ ggplot_build.ggplot <- function(plot) {
   if (npscales$n() > 0) {
     lapply(data, npscales$train_df)
     data <- lapply(data, npscales$map_df)
+    plot$key_data <- npscales$key_data(data)
   }
 
   # Fill in defaults etc.
@@ -178,7 +179,8 @@ ggplot_gtable.ggplot_built <- function(data) {
   }
 
   legend_box <- plot$guides$build(
-    plot$scales, plot$layers, plot$mapping, position, theme, plot$labels
+    plot$scales, plot$layers, plot$mapping, position, theme, plot$labels,
+    plot$key_data
   )
 
   if (is.zero(legend_box)) {
--- a/R/scales-.R
+++ b/R/scales-.R
@@ -81,6 +81,37 @@ ScalesList <- ggproto("ScalesList", NULL,
     data_frame0(!!!mapped, df[setdiff(names(df), names(mapped))])
   },
 
+  key_data = function(self, data) {
+
+    scales <- self$scales
+
+    lapply(scales, function(s) {
+      if (!s$is_discrete()) {
+        return(NULL)
+      }
+      n <- s$n.breaks.cache %||% sum(!is.na(s$limits %|W|% s$get_limits()))
+      if (n < 1) {
+        return(NULL)
+      }
+      pal <- s$palette.cache %||% s$palette(n)
+      pal <- c(pal, s$na.value)
+      aes <- s$aesthetics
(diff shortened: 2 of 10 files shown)


R/plot-build.R
```diff
@@ -89,6 +89,7 @@ ggplot_build.ggplot <- function(plot) {
   if (npscales$n() > 0) {
     lapply(data, npscales$train_df)
     data <- lapply(data, npscales$map_df)
+    plot$key_data <- npscales$key_data(data)
```
teunbrand, line 92: Checking the data has to happen before the defaults are filled in and `after_scale()` modifications are applied. Otherwise the data might be lost. I gave the scales list a method to check the data.

R/scales-.R
```diff
@@ -81,6 +81,37 @@ ScalesList <- ggproto("ScalesList", NULL,
...
+        }
+        present <- vapply(aes, function(a) {
+          vec_in(pal, d[[a]])
+        }, logical(length(pal)))
+        if (length(dim(present)) > 1) {
+          present <- rowSums(present) > 0
+        }
+        present
+      }, logical(length(pal)))
+      list(aesthetics = aes, data = data_frame0(pal = pal, member = out))
```
teunbrand, line 111: At this point, `member` is a `length(pal)` * `length(data)` logical matrix tracking whether a palette value was seen.

teunbrand: Also `aes` can be a vector, e.g. `c("colour", "fill")`. The inclusion/exclusion rule in the guide only needs to match one of the aesthetics, not all to be included.

teunbrand:
Closing this PR in favour of #5502

PR #5304, opened 2023-05-10
Only call `zero_range()` and `as.numeric()` in transformed space

DavisVaughan:
In case the `limits` in data space don't support conversion to numeric

---

I've been working on scale support for {clock} types in https://github.com/r-lib/clock/pull/345. These include scales for things like `year_month_day<month>`, a "year-month" style date vector, and `year_week_day<week>`, a "year-week" style date vector.

I've got pretty much everything working, but there is a show stopping issue with `ScaleContinuous$get_breaks()` that I've hopefully fixed here.

It looks like `scales::zero_range()` is being called on `as.numeric(limits)`, where `limits` is in the original _data_ space. This is problematic for clock, because calendar types like `year_month_day` and `year_week_day` do not support `as.numeric()`, so the limits can't be converted directly. However, they _do_ support `scale$trans$transform()` to convert from a clock type in data space to a numeric value in transformed space.

IIUC, I think that the intention of `scales::zero_range()` is that it should only be called on objects in _transformed_ space anyways. Looking at other uses of it in ggplot2:
- Most importantly, in `ScaleContinuous$get_minor_breaks()` it is called on `limits` when they are in transformed space, so this PR basically just makes `get_breaks()` consistent with that. https://github.com/tidyverse/ggplot2/blob/aa2ddbe95b521359cebc19cf1e7463ad08ed7ebe/R/scale-.R#L692
- In `get_breaks()` just above where this change happens it is called on `domain` which is in transformed space https://github.com/tidyverse/ggplot2/blob/aa2ddbe95b521359cebc19cf1e7463ad08ed7ebe/R/scale-.R#L651
- In `bin_breaks_bins()` it is called on `x_range` which comes through as a `$dimension()` call, and the dimension is always returned in transformed space https://github.com/tidyverse/ggplot2/blob/aa2ddbe95b521359cebc19cf1e7463ad08ed7ebe/R/bin.R#L112

Hopefully this is enough evidence to support this change.

I couldn't think of a good test for this, so if anyone has a better idea, feel free to take over.


DavisVaughan:
@teunbrand if you could please take over I would really appreciate it! It would probably be much faster for you at this point.

teunbrand:
Thanks @DavisVaughan for the PR!

Issue #5307, opened 2023-05-13
typos


spaette:
Similarly
Tightly
accommodate
accepted
aesthetic
aesthetics
appearance
aesthetic
available
changes
compatibility
computed
constructed
continuous
correspond
correctly
default
deparses
discrete
distinct
existing
existence
expansion
features
fractional

```
$ grep -nr Simiarly ggplot2
ggplot2/R/coord-cartesian-.R:46:#' # Simiarly, we can use expand = FALSE to turn off expansion with the
ggplot2/man/coord_cartesian.Rd:63:# Simiarly, we can use expand = FALSE to turn off expansion with the
$ grep -nr Tighly ggplot2
ggplot2/NEWS.md:2262:* Tighly linked geoms and stats (e.g. `geom_boxplot()` and `stat_boxplot()`)
$ grep -nr accomodate ggplot2
ggplot2/vignettes/extending-ggplot2.Rmd:363:    *min*, and *max* are intentionally renamed to accomodate base graphical
$ grep -nr accpeted ggplot2
ggplot2/R/axis-secondary.R:99:  # sec_axis() historically accpeted two-sided formula, so be permissive.
$ grep -nr aesthethic ggplot2
ggplot2/_pkgdown.yml:35:    aesthethic mappings, specified by `aes()`. You then add layers, scales,
$ grep -nr aesthietcs ggplot2
ggplot2/tests/testthat/test-layer.R:24:test_that("unknown aesthietcs create warning", {
$ grep -nr apperance ggplot2
ggplot2/NEWS:1980:  over plot apperance
$ grep -nr asthetic ggplot2
ggplot2/tests/testthat/test-layer.R:35:test_that("unknown NULL asthetic doesn't create warning (#1909)", {
$ grep -nr avialable ggplot2
ggplot2/R/autolayer.R:18:  cli::cli_abort("No autolayer method avialable for {.cls {class(object)[1]}} objects")
ggplot2/tests/testthat/_snaps/autolayer.md:3:    No autolayer method avialable for <character> objects
$ grep -nr chages ggplot2
ggplot2/NEWS.md:784:* Fixes unit tests for sf graticule labels caused by chages to sf
$ grep -nr compatability ggplot2
ggplot2/R/guide-colorsteps.R:16:#'   (default), no tick marks are drawn. For backward compatability, can also
ggplot2/R/guide-.R:3:#' A constructor function for guides, which performs some standard compatability
ggplot2/NEWS:1767:  compatability
[...]

PR #5313, opened 2023-05-21
Fix typos part III

teunbrand:
This PR continues #5307, fixing typos in doc strings that needed to be re-documented.

### DIFF
--- a/R/coord-cartesian-.R
+++ b/R/coord-cartesian-.R
@@ -43,7 +43,7 @@
 #' # limits. You can set the limits precisely by setting expand = FALSE
 #' p + coord_cartesian(xlim = c(325, 500), expand = FALSE)
 #'
-#' # Simiarly, we can use expand = FALSE to turn off expansion with the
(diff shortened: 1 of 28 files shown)


spaette:
typo `exisiting` appeared twice in this file and also in man/bidirection.Rd

```
$ ed -s ggplot2/R/utilities.R <<<'431,435p'
#' - `main_is_orthogonal`: This argument controls how the existence of only a `x`
#'   or `y` aesthetic is understood. If `TRUE` then the exisiting aesthetic
#'   would be then secondary axis. This behaviour is present in [stat_ydensity()]
#'   and [stat_boxplot()]. If `FALSE` then the exisiting aesthetic is the main
#'   axis as seen in e.g. [stat_bin()], [geom_count()], and [stat_density()].
$ 
```

teunbrand:
Thanks for catching this!

thomasp85 (approved):
LGTM

PR #5316, opened 2023-05-30
Fix partial matching: grob$label -> grob$labels

TimTaylor:


teunbrand (approved):
Oops yes, this was my bad. Thanks for catching!

Issue #5244, opened 2023-03-22
Upcoming maptools retirement


yutannihilation:
A year ago, R-spatial announced rgdal, rgeos, and maptools will retire by the end of 2023.

https://r-spatial.org//r/2022/04/12/evolution.html

ggplot2 doesn't use rgdal, and just removed rgeos in #5242. However, maptools is still used in `fortify.SpatialPolygonsDataFrame()` (and the examples section of the document).

https://github.com/tidyverse/ggplot2/blob/1d1f79560c23241999529a92178a913769b8d28e/R/fortify-spatial.R#L35

We need to decide how to deal with this. I think we have several options:

1. Replace `maptools::unionSpatialPolygons()` with sf functions (convert to an sf object, transform it and convert back to an sp object)
2. Remove the functionality related to `maptools::unionSpatialPolygons()`
3. Drop `fortify()` methods for sp objects altogether

yutannihilation:
https://geocompx.org/post/2023/rgdal-retirement/#solutions-for-package-developers

hadley:
I think we can drop it. We should schedule this for the next release.

yutannihilation:
Thanks, let's drop it.

Note that the reason I couldn't decide is that I was afraid this assumption

> everyone has moved to sf by now

might not be really true. For example, I read a book about spatial statistics using R, which was published in 2022, and it still uses rgoes and rgdal in several places although it primarily uses sf. I thought this probably means, even if the internal implementations were already replaced with sf, these longstanding packages might still generate the result as sp (I'm not familiar with spatial statistics, so I'm not sure).

But, converting an sp object to sf should be easy, and the sp ecosystem is about to retire anyway. So, it's probably time to drop it.

hadley:
Alternatively, if ggplot2 still has a sf dependency, we could switch to `sf::st_union()` instead.

thomasp85:
Agreed - leaning heavily into dropping it

Ludecan:
Hey there. I'm currently maintaining a very old spatial data codebase that heavily used `sp` for spatial data manipulation and `ggplot2` for plotting. As a part of it I did both plots of point and raster data while plotting using a background shapefile like this:
![redesDisponibles](https://github.com/tidyverse/ggplot2/assets/17169248/2e82ef5e-6dd1-409f-aca8-a694f35a1d74)

[...]

PR #5327, opened 2023-06-17
Remove maptools and deprecate fortify() for sp objects

yutannihilation:
A part of #5244

While we lean to agree on dropping these `fortify()` for sp objects altogether, there might be several options. At the safest, we should remove maptools, because otherwise ggplot2 might be kicked out from CRAN in the worst-case scenario. So, this pull request does

1. remove maptools
2. deprecate `fortify()` for sp objects in general


teunbrand (approved):
LGTM, but would it make sense to be a little bit more explicit why people need to migrate to {sf}?
For example: "The maptools package {is/will be} retired: please migrate to sf."

teunbrand:
Right, I might've skimmed over the nuance here. In any case it was merely a suggestion, I think this PR is fine as-is.

teunbrand:
@yutannihilation There were no reverse dependency issues with retiring maptools and it would be nice to include this PR in the hotfix for v3.4.4, as maptools retires in October. If you'd like, I'm happy to merge this in to the v3.4.4-rc and tweak the version number. If you prefer changing the target branch and changing the version number in deprecation messages yourself, that is great too.

yutannihilation:
> If you prefer changing the target branch and changing the version number in deprecation messages yourself, that is great too.

I'm fine with either, but, in my understanding, this needs to be merged into the main branch anyway, no matter whether this is merged into the RC branch or not, right?

thomasp85:
The RC branch will be merged into main after release

thomasp85:
Ah, you already merged in main. Yeah, then merging this PR and cherry picking the commit is probably the right way

thomasp85:
hmm - can you try to change the target of the PR and see what it will change... I don't think it will bring in all the prior changes, but maybe I'm wrong

teunbrand:
I'm not very wise in the ways of git-fu but isn't there a way to rebase a branch to 'play out' the commits over a different branch?

yutannihilation:
Oops, I'm sorry. I actually tried to do this

> I'm not very wise in the ways of git-fu but isn't there a way to rebase a branch to 'play out' the commits over a different branch?

but accidentally pushed into `v3.4.4-rc` directly...

https://github.com/tidyverse/ggplot2/commits/v3.4.4-rc

If this looks good to you, you can use this as it is. If not, please recreate the branch. So sorry for the trouble.

teunbrand:
> If this looks good to you, you can use this as it is.

Yeah I think it is probably fine, thanks so much for resolving this!

yutannihilation:
@teunbrand 
I see. Thanks for confirming! I'll be more careful next time... Then, I'm closing this pull request in favor of the RC branch.

@thomasp85 
> hmm - can you try to change the target of the PR and see what it will change... I don't think it will bring in all the prior changes, but maybe I'm wrong

I too didn't know what actually would happen, so I created a pull request on my repo, here's how it looks like:

https://github.com/yutannihilation/ggplot2/pull/10

Issue #5323, opened 2023-06-09
ggsave() doesn't recognize .tif extension and device = agg_tiff() or tiff() is ignored


twest820:
Attempted hello world, failed badly (.png and .jpg are fine as usual).
```r
ggsave("foo.tif")  # try to write a TIFF using the most commonly file extension
Error in `ggsave()`:
! Unknown graphics device "tif"

ggsave("foo.tif", device = tiff(compression = "lzw")) # try specifying the graphics device explicitly
Error in `ggsave()`:
! Unknown graphics device "tif"

ggsave("foo.tif", device = ragg::agg_tiff(compression = "lzw"))
Error in `ggsave()`:
! Unknown graphics device "tif"

ggsave("foo.tiff", device = tiff(compression = "lzw")) # produces an uncompressed TIFF file

ggsave("foo.tiff", device = ragg::agg_tiff(compression = "lzw")) # also produces an uncompressed TIFF file and additionally doesn't honor (or at least warn about overriding) agg_tiff()'s other settings
```
From replies in #3525 it appears `tiff(compression = "lzw")` might have worked in 2019 and reverted since.

RStudio 2023.06.0 with AGG backend, R 4.2.3, ggplot 3.4.2.

teunbrand:
Sorry, maybe I'm misunderstanding, but the request here is that `"foo.tif"` be recognised as the same device as `"foo.tiff"`?
I don't think `device` accepts a function call, so you'd have to pass `compression = "lzw"` to the `...` argument instead of as part of the device call.

twest820:
> I don't think device accepts a function call

Hi, I think the concept you're looking for is a constructor. That's a function call which returns, in this case, the relevant device object which should be used in writing the file. The use of a device constructor here follows the documentation (and quite a few examples).

> the request here is that "foo.tif" be recognised as the same device as "foo.tiff"?

That would address one of the two issues here, yes.

teunbrand:
But the device functions aren't constructors and return `NULL`.  I'm afraid I can't find the documentation and examples you mention in neither ggplot2's docs or the ggplot2 book.

I think this is the recommended syntax, which is in working order even if you use `.tif` instead of `.tiff`:

```
ggsave("foo.tif", device = ragg::agg_tiff, compression = "lzw")
````

twest820:
It's in the [`ggsave()`](https://ggplot2.tidyverse.org/reference/ggsave.html) documentation:

> device
[...]

PR #5329, opened 2023-06-17
Use `.tif` as synonym to `.tiff`

teunbrand:
This PR aims to fix part of #5323.

Just as `.jpg` is synonymous to `.jpeg`, so is `.tif` synonymous to `.tiff`.
This updates the device list to recognise `.tif` extensions in filenames.


yutannihilation (approved):
Looks good.

Btw, is anyone looking at the visual test failures on R-devel? I checked at the release note of R-devel, but couldn't find anything that is apparently relevant.

yutannihilation:
Oh, I see. Thanks for the details!

PR #5331, opened 2023-06-19
Fix documentation of `stat` parameter

fkohrt:
`stat` accepts e.g. `ggplot2::StatBin` (class `Stat`), but not `ggplot2::GeomBar` (class `Geom`).


teunbrand:
Thanks for preparing a PR for this. Unfortunately, this particular point is also addressed in #5127, so these PRs are conflicting but both solve the issue.

teunbrand:
I do appreciate the effort going into this PR and want to thank you for that, but I think I'll close this one in favour #5127.

Issue #5332, opened 2023-06-20
Feature request: graphics device capabilities checker


teunbrand:
This feature request follows up on discussion in https://github.com/tidyverse/ggplot2/pull/5299#discussion_r1186764214. 

To summarise: if ggplot2 is considering taking up newer graphics features, such as patterns, we'd need to check if the device is capable of using these features. Several extension developers also expressed interest in having ggplot2 export such a checker.

@trevorld has released such a checker [here](https://github.com/trevorld/gridpattern/blob/main/R/standalone-guess_has_R4.1_features.R) that is free to copy and/or adapt due to the 'unlicence' licence.

There are a few hurdles that makes implementing this checker more complicated:
* {svglite} and {ragg} don't report several features that *are* implemented in `dev.capabilities()`.
* IIRC {vdiffr} uses a variant of {svglite} that uses the same device name, but does not support patterns for stability reasons.

PR #5334, opened 2023-06-25
TLC for polar coordinates: `coord_radial()`

### LINKED ISSUE #3959: coord_trans/polar() ignore position guides

### LINKED ISSUE #4462: Add "end" parameter to coord_polar to allow non-complete polar plots

### LINKED ISSUE #4815: coord_polar _ remove extra grid line
The issue has been discussed on https://stackoverflow.com/questions/72005238/scale-y-axis-on-a-polar-plot-in-ggplot/72005779#72005779 from which I copy the exchange:

When plotting the following,

``` r
library(ggplot2)
library(tibble)

foo <- tibble(theta = c(0, 10, 20), r = c(10, 30, 90))

ggplot(foo) +
  geom_point(aes(x = theta, y = r)) +
  coord_polar() +
  scale_y_continuous(limits = c(0, 90), expand = c(0, 0), breaks = 90) +
  scale_x_continuous(limits = c(0, 360), expand = c(0, 0), breaks = c(0, 90, 180, 270)) +
  theme(
    panel.grid.minor = element_blank(),
    panel.grid.major.y = element_line(colour = "red")
  )
```

the Y-axis (= radius) expands to 100 instead of 90. In addition there are two circles corresponding to major grid elements, one at 90 as required and one at 100. I added the "theme" statement to demonstrate that both correspond to major grid lines.

A workaround proposed by user Allan Cameron was

``` r
ggplot(foo) +
  geom_point(aes(x = theta, y = r)) +
  coord_polar() +
  scale_y_continuous(limits = c(0, 90), expand = c(0, 0), breaks = 90) +
  scale_x_continuous(limits = c(0, 360), expand = c(0, 0), breaks = c(0, 90, 180, 270)) +
  theme(
    panel.grid.minor = element_blank(),
    panel.grid.major.y = element_line(colour = c("red", NA))
  )
```

... which works, but is hardly satisfactory ! Quoting A.C. :
"There is no outer axis line in coord_polar, and a grid line seems to be used in its place, without respecting axis.line theme elements."

### LINKED ISSUE #5059: Feature request: `coord_polar()` argument to transform `angle` aesthetic.
Hi everyone,

I've quite frequently encountered people on stackoverflow having trouble aligning text to plots in polar coordinates.
I think it would be relatively straightforward to propagate the `theta` aesthetic to text angles, so I was wondering whether it might be considered to give a `transform_angle = FALSE/TRUE` argument to `coord_polar()` so that we could transform a plot that looks like this:

``` r
library(ggplot2)

df <- data.frame(
  x = 1:5,
  label = c("cat", "dog", "banana", "apple", "coffee")
)

p <- ggplot(df, aes(x, 1, label = label)) +
  scale_x_continuous(limits = c(1, 6)) +
  coord_polar()

p + geom_text()
```

Into one that looks like this:

``` r
p + geom_text(angle = seq(360, 0, length.out = 6)[-6])
```

<sup>Created on 2022-11-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Perhaps it could even be detected when text has an angle between 90&deg; and 270&deg; to add an extra 180&deg; to improve readability. I'd be willing to take a stab at a PR if this is something that is deemed fitting for ggplot2.

teunbrand:
This PR fixes #4815, fixes #5059, fixes #3959, and fixes #4462

Briefly, this PR introduces a new coord system `coord_polar2()` with expanded options (name of function and arguments are up for discussion).

Less briefly, this PR started out with a simpler goal of having `coord_polar()` use the axis system. Realising that this couldn't be done without breaking backward compatibility, this has instead become `coord_polar2()`. This PR then started accumulating additional features and here we are today. I think it might be best to explain this PR with a visual walkthrough.

If we use `coord_polar2()` plainly, we see it resembles `coord_polar()` with the following differences:

* Theta is represented by a proper position axis. Figuring out text placement here was the hardest part, and still isn't perfect. The default is to display text of the theta axis horizontally.
* The `panel.background` theme setting is applied to the circle that forms the background. You can still use `panel.border = element_rect(fill = NA)` in the theme if you need a frame for the plot.
* Radius axis line starts and stops at the actual radius limits and doesn't cover the entire left side of the rectangular panel.
* The coords has a `expand = TRUE` default argument so that the behaviour of scale expansion is more in line with `coord_cartesian()`  than with `coord_polar()`.

``` r
#> ℹ Loading ggplot2

  geom_point() +

p + coord_polar2()
```

One feature is that we can also place the radius axis *inside* the circle. Admittedly, this looks a little bit awkward because it intersects with the theta axis, which is why this isn't the default when you have a full circle.

``` r
p + coord_polar2(r_axis_inside = TRUE)
```

The next feature is that requested in #4462, namely to have partial polar coordinates. The following things are of note:
* The `start` and `end` arguments to control which sector of a circle the plot occupies.
* Since we don't have a full circle, the radius axis is placed inside and not at the rectangular bounding box. You can use `coord_polar2(r_axis_inside = FALSE)` to prevent this.
* The radius axis snaps to the start position. While the axis itself is rotated to reflect this, the labels remain horizontal.

``` r
p + coord_polar2(start = -0.25 * pi, end = 0.25 * pi)
```

* If you swap direction, the radius axis follows the swap.

``` r
p + coord_polar2(start = -0.25 * pi, end = 0.25 * pi, direction = -1)
```

Then the reason I started this PR: #3959. A few comments:

* There is a new `guide_axis_theta()` that can be used for the theta axis.
* Radius axes cannot be the new theta guide, which is what prompts the warning below. The inverse is also true, i.e. `guides(theta = "axis")` also throws a warning.
* If the `angle` argument is set, either in a radius axis or theta axis, this is interpreted as a *relative* angle. Note that the theta text follows the curvature and that the radius text are projected from the tick marks.

``` r
p + coord_polar2(start = -0.25 * pi, end = 0.25 * pi) +
  guides(
    r     = guide_axis_theta(),
    r.sec = guide_axis(angle = 0),
    theta = guide_axis_theta(angle = 0)
  )
#> Warning: `guide_axis_theta()` cannot be used for r.
#> ℹ Use one of x, y, or theta instead.
```

I then took a little bit of a liberty and decided on my own it would be fun to also be able to set a donut hole in polar coordinates. This is mostly just a convenience that doesn't force you to fiddle with the scale limits or expansions to not have all shapes disappear in a single point at the center. Also, it is the only circumstance I could ever imagine that you'd need a secondary theta axis.

``` r
p + coord_polar2(donut = 0.5) +
  guides(theta.sec = "axis_theta")
```

Lastly, I also implemented #5059, since getting the (relative) angle right in text geoms in polar coordinates is a bit of pain. Text between 90 and 270 degrees is flipped for readability reasons.

``` r
df <- data.frame(
  x = LETTERS[1:5], lab = c("cat", "farm", "banana", "airplane", "baker")
)

ggplot(df, aes(x, label = lab)) +
  geom_text(aes(y = "0 degrees"),  angle = 0)  +
  geom_text(aes(y = "90 degrees"), angle = 90) +
  coord_polar2(rotate_angle = TRUE)
```

<sup>Created on 2023-06-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

A few closing remarks:
* `guide_axis_theta()` also works reasonably well with cartesian coordinates.
* `guide_axis_theta()` does not respond to `hjust/vjust` settings, mostly because text placement had given me enough headaches at this point and the theme's defaults translate horribly to the theta axis.
* The `theta` positions uses `x.bottom` theme settings, `theta.sec` uses `x.top`, `r`/`r.sec` use `y.left`/`y.right` depending on the `direction` argument. In theory, we could change this, but it would involve some hairy adaptations to `guide_axis()` as well.
* Clipping currently isn't ideal, as it applies to the rectangular bounding box. We could use R4.1+ features to clip to the curved bounding boxes, but we'd need to implement #5332 for that.


R/guide-axis-theta.R
```diff
@@ -0,0 +1,332 @@
...
+    position = position,
+    name = "axis",
+    super = GuideAxisTheta
+  )
+}
+
+GuideAxisTheta <- ggproto(
+  "GuideAxisTheta", GuideAxis,
+
+  # TODO: delete if minor ticks PR (#5287) gets merged
```
teunbrand, line 81: This may be simplified once #5287 is merged.

R/guide-axis.R
```diff
@@ -41,7 +44,7 @@
 #'
 #' # can also be used to add a duplicate guide
 #' p + guides(x = guide_axis(n.dodge = 2), y.sec = guide_axis())
-guide_axis <- function(title = waiver(), check.overlap = FALSE, angle = NULL,
+guide_axis <- function(title = waiver(), check.overlap = FALSE, angle = waiver(),
```
teunbrand, line 47: The default is now `waiver()` because `coord_polar2()` may overrule angle for display purposes, but `NULL` still uses the theme's angle.

R/guide-axis.R
```diff
@@ -509,41 +518,42 @@ axis_label_priority_between <- function(x, y) {
...
 #'
 axis_label_element_overrides <- function(axis_position, angle = NULL) {
-  if (is.null(angle)) {
+
+  if (is.null(angle) || is.waive(angle)) {
     return(element_text(angle = NULL, hjust = NULL, vjust = NULL))
   }
 
-  # it is not worth the effort to align upside-down labels properly
-  check_number_decimal(angle, min = -90, max = 90)
```
teunbrand, line 517: I very much did need some alignments for upside-down labels, so I have expanded this function to include this.

R/layout.R
```diff
@@ -257,11 +257,13 @@ Layout <- ggproto("Layout", NULL,
         guides <- c("x", "x.sec")
       }
       params    <- self$panel_params[[1]]$guides$get_params(guides)
-      primary   <- params[[1]]$title %|W|% primary
-      secondary <- params[[2]]$title %|W|% secondary
-      position  <- params[[1]]$position %||% scale$position
-      if (position != scale$position) {
-        order <- rev(order)
+      if (!is.null(params)) {
```
teunbrand, line 260: Layout doesn't know about theta/r guides, so had to avoid taking labels from `NULL` params here that are returned when the x/y guide doesn't exist.

teunbrand:
Also worth pointing out that interactions with e.g. {ggtext} aren't that horrible:

``` r
#> ℹ Loading ggplot2
library(ggtext)
labels <- c(
  setosa = "<img src='https://upload.wikimedia.org/wikipedia/commons/thumb/8/86/Iris_setosa.JPG/180px-Iris_setosa.JPG'
    width='100' /><br>*I. setosa*",
  virginica = "<img src='https://upload.wikimedia.org/wikipedia/commons/thumb/3/38/Iris_virginica_-_NRCS.jpg/320px-Iris_virginica_-_NRCS.jpg'
    width='100' /><br>*I. virginica*",
  versicolor = "<img src='https://upload.wikimedia.org/wikipedia/commons/thumb/2/27/20140427Iris_versicolor1.jpg/320px-20140427Iris_versicolor1.jpg'
    width='100' /><br>*I. versicolor*"
)

ggplot(iris, aes(Species, Sepal.Width)) +
  geom_boxplot() +
  scale_x_discrete(
    name = NULL,
    labels = labels
  ) +
  coord_polar2(r_axis_inside = TRUE) +
  theme(
    axis.text.x = element_markdown(color = "black", size = 11),
    axis.title.y = element_blank(),
    plot.margin = margin(5,5,70,5)
  )
```



<sup>Created on 2023-06-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Note that you couldn't do this with regular `coord_polar()`.

R/coord-polar2.R
```diff
@@ -0,0 +1,580 @@
...
+        theme, majorr, name = "radius",
+        x = rescale(rep(r_fine, each = length(theta_fine)) *
+          rep(sin(theta_fine), length(r_fine)) + 0.5, from = bbox$x),
+        y = rescale(rep(r_fine, each = length(theta_fine)) *
+          rep(cos(theta_fine), length(r_fine)) + 0.5, from = bbox$y),
+        id.lengths = rep(length(theta_fine), length(r_fine)),
+        default.units = "native"
+      )
+    ))
+  },
```
thomasp85, line 326: Can you point me to where the background gets shaped like the plotting area (circle, partial circle, donut, etc) - I can't seem to find it in here

R/guide-axis-theta.R
```diff
@@ -0,0 +1,332 @@
...
+    # Position angle in radians
+    theta <- key$theta
+
+    # Offset distance to displace text away from outer circle line
+    xoffset <- elements$offset * sin(theta)
+    yoffset <- elements$offset * cos(theta)
+
+    # Note that element_grob expects 1 angle for *all* labels, so we're
+    # rendering one grob per label to propagate angle properly
+    do.call(grobTree, Map(
```
thomasp85, line 234: We have moved away from `do.call()` in favour of `rlang::inject()` when possible

R/guide-axis-theta.R
```diff
@@ -0,0 +1,332 @@
...
+    return(invisible())
+  },
+
+  arrange_layout = function(key, sizes, params) {
+    return(invisible())
+  },
+
+  assemble_drawing = function(grobs, layout, sizes, params, elements) {
+    if (params$position %in% c("theta", "theta.sec")) {
+      return(do.call(grobTree, grobs))
```
thomasp85, line 279: Same as above

R/guide-axis-theta.R
```diff
@@ -0,0 +1,332 @@
...
+      )
+      vp <- viewport(
+        x = unit(as.numeric(params$position == "left"), "npc"),
+        height = unit(1, "npc"), width = width,
+        just = opposite_position(params$position)
+      )
+    }
+
+    absoluteGrob(
+      do.call(gList, grobs),
```
thomasp85, line 306: same as above do.call comment

R/coord-polar2.R
teunbrand: Yeah it is in L294-L298. Essentially, we make an infinitely sized rectangle that we let the coord munching reshape to what it should be.

thomasp85:
Name suggestion: `coord_radial()`

teunbrand:
Latest changes:

- Renamed `coord_polar2()` to `coord_radial()`.
- Use minor ticks plumbing from #5287, so custom `extract_key` method and params became redundant.
- `coord_polar()` and `coord_radial()` now share the rescale helper functions

Should be ready for review :)

teunbrand:
I couldn't really use `Map()` with old R (3.6.3) units to render labels, so instead I just vectorised `rotate_just()`. This allows `element_text()` to have multiple angles, thereby eliminating the need for `Map()`ing the labels.

thomasp85 (approved):
I approve this, but please fix the small style stuff I commented on

R/guide-axis-theta.R
```diff
@@ -0,0 +1,276 @@
...
+    params <- GuideAxis$transform(params, coord, panel_params)
+
+    key <- params$key
+    n <- nrow(key)
+
+    params$theme_aes <- coord$theta %||% params$aesthetic
+
+    if (!("theta" %in% names(key))) {
+      # We likely have a linear coord, so we match the text angles to
+      # standard axes.
```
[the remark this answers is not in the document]
teunbrand: It will draw a very normal looking axis with slightly inferior options for text justification.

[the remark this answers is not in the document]
teunbrand:
``` r
#> ℹ Loading ggplot2

  geom_point() +
  guides(x = "axis_theta", y = "axis_theta") +
```



<sup>Created on 2023-11-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/guide-axis-theta.R
```diff
@@ -0,0 +1,276 @@
...
+    minor <- theta_tickmarks(
+      vec_slice(key, (key$.type %||% "major") == "minor"),
+      elements$minor, elements$minor_length
+    )
+
+    grobTree(major, minor, name = "ticks")
+  },
+
+  measure_grobs = function(grobs, params, elements) {
+    return(invisible())
```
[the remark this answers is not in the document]
teunbrand: Thanks for catching this!

teunbrand:
Thanks for the review Thomas!

PR #5335, opened 2023-06-25
Avoid theme partial string matching

### LINKED ISSUE #2724: R partial string matching can have unexpected results when theme elements are missing
This is related to #2723, but really a distinct issue that I wanted to record separately. If a theme element is missing then R may decide to return a different one whose name partially matches. This can have unexpected results.

``` r
library(ggplot2)
t <- theme_void()
t$legend.box
#> NULL

t <- theme_void() + theme(legend.box.background = element_rect(fill = "gray80"))
t$legend.box
#> List of 5
#>  $ fill         : chr "gray80"
#>  $ colour       : NULL
#>  $ size         : NULL
#>  $ linetype     : NULL
#>  $ inherit.blank: logi FALSE
#>  - attr(*, "class")= chr [1:2] "element_rect" "element"
```

Created on 2018-06-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

teunbrand:
This PR aims to fix #2724.

Briefly, it it replaces most instances of `theme$setting` with `calc_element("setting", theme)`.
A particularly gnarly case was `guide_legend()` that overrides inheritance, but I've tried to use double-bracket indexing in such cases. Also updated vignette with this, hopefully encouraging people to use `calc_element()` instead.


R/facet-grid-.R
```diff
@@ -362,9 +362,9 @@ FacetGrid <- ggproto("FacetGrid", Facet,
     panel_table$layout$name <- paste0('panel-', rep(seq_len(nrow), ncol), '-', rep(seq_len(ncol), each = nrow))
 
     panel_table <- gtable_add_col_space(panel_table,
-      theme$panel.spacing.x %||% theme$panel.spacing)
+                                        calc_element("panel.spacing.x", theme))
```
thomasp85, line 365: we don't really use this indentation style in the codebase

thomasp85 (commented):
`calc_element()` is a pretty meaty function to substitute in "just" to avoid partial matching... I'm a bit concerned about the impact on performance of this

teunbrand:
Closing this in favour of #5522

PR #5340, opened 2023-06-28
Fix typo

### LINKED ISSUE #5339: Documentation typo: `ymin` mentioned twice
The documentation currently says:

https://github.com/tidyverse/ggplot2/blob/23508b321357502d014e52b8118ed1b0f9c5c164/R/position-stack.R#L259

But probably the second field has to be `ymax`?

fkohrt:
Add `ymax` instead of `ymin` twice

Closes #5339


teunbrand (approved):
Looks good to me, thanks!

PR #5343, opened 2023-07-04
Better scale messages

### LINKED ISSUE #1312: scale_name property should be removed from Scale objects

### LINKED ISSUE #4185: awkward behavior from scale_fill_viridis_c(trans = "log")
Three interrelated issues:
1. Log transformation of zero values for a continuous fill produces a warning about infinite values in a discrete y axis. This warning could be more clearly worded since the fill is neither discrete nor a y axis. There also appears to be differentiation of NaN from -Inf, leading to a cryptic NaN warning from negative fill values.
2. There doesn't appear to be any way to tell ggplot these fill values are expected in order to suppress the warnings. The closest I can see is something like  `scale_fill_viridis_c(na.value = "transparent", trans = "log")`, which plots fine since apparently -Inf, NaN, and NA are treated interchangeably but is orthogonal to warning generation.
3. The resulting labels on the fill colorbar include 8+ decimal places, which is a curiously verbose default, though `format_format(digits = n)` provides a simple mitigation.

```r
library(ggplot2)
library(magrittr)
library(scales)
library(tidyverse)

data = crossing(x = seq(0, 1, length.out = 100), y = seq(0, 1, length.out = 100))
data %<>% mutate(value = x - y)
ggplot(data) + geom_raster(aes(x = x, y = y, fill = value)) + scale_fill_viridis_c(trans = "log")

Warning messages:
1: In self$trans$transform(x) : NaNs produced
2: Transformation introduced infinite values in discrete y-axis 

# change for more concise labeling
ggplot(data) + geom_raster(aes(x = x, y = y, fill = value)) + scale_fill_viridis_c(labels = format_format(digits = 2), trans = "log")
```

Observed on R 4.0.2 with tidyverse 1.3.0 (ggplot 3.3.2). I suspect this might also occur with other scale_fill functions.
ggplot scale_fill_viridis_c

### LINKED ISSUE #4258: Suggestion: Point out which continuous/discrete value was wrongly supplied to which discrete/continuous scale
When working with very complex plots, with many different values, mappings, and scales, it often happens that I mix something up and accidentally supply a continuous value to a discrete scale, or vice versa. In this case, while ggplot does tell me which way the mix-up occurred, it does not tell me *which value* was specifically wrongly supplied to *which scale*. Instead, I have to figure this out myself, often by frustrating trial-and-error if it is not immediately obvious from the code.

I would love it if the error message included:

- the name of the scale which a continuous/discrete value was wrongly supplied to,
- the aesthetic which is mapped to this scale,
- and the value which is mapped to this aesthetic.
- Possibly even more useful information that is available internally, that I am not aware of?

Here is some reproducible code to illustrate my problem:

```r
library(tidyverse)

set.seed(0)

df = tibble(col1 = rnorm(6) * 100,
            col2 = factor(c(100, 100, 101, 101, 102, 102)),
            col3 = rnorm(6),
            col4 = factor(c(1, 2, 1, 3, 5, 2)))

df %>% ggplot(mapping = aes(x = col1, y = col2, fill = col3, colour = col4)) +
  geom_point(pch = 21, size = 10, stroke = 3) +
  scale_fill_viridis_d(option = "viridis") +
  scale_colour_viridis_d(option = "magma")
```

When I run this, I get the dreaded:

```
Error: Continuous value supplied to discrete scale
```

However, finding out which scale would be changed—in this case, to be continuous—requires reading and understanding the code. (Of course, in this case, it is the `fill` scale, but can you see this at first glance?) It is even harder to do if:

- the plot is assembled at distant code locations; for example when a "barebones" plot only with geoms is constructed, and then "finished off" with scales/labs/etc. further down in the script; especially if the same plot can be "finished off" in multiple different ways
- the same aesthetic has multiple values mapped to multiple scales (using something like `ggnewscale`)
- there are scales involved where it is not immediately obvious whether they are discrete or continuous, particularly if they are supplied by another external module
- binned scales are involved
- you are new to `ggplot2` and don't even know what the error message is trying to tell you in the first place...
- and so on

Now my simple suggestion would be to amend the error message, so it reads for example:

```
Error: Continuous value (col3) supplied to discrete scale ("viridis") for aesthetic "fill"
```

I'm sure the exact wording can be improved; this is just an idea. I don't know how much effort this would be to implement, but it would certainly improve my quality of life a lot, and perhaps even that of others.

teunbrand:
This PR aims to fix #4185, fix #4258 and fix #1312.

Briefly, this PR sets the `call` field of a `Scale` ggproto class to a more informative call, and uses that call in the messaging.

Less brief: errors thrown in scales led to some confusion about *which* scale was throwing the error. Getting the appropriate call for every scale helps in throwing more informative messages. In addition, this PR deprecates the `scale_name` argument to scale constructors, as the `call` takes over its function. This is in opposition to suggested approach here: https://github.com/tidyverse/ggplot2/issues/1312#issuecomment-503166739. Some selected examples follow.

Previously, the below would throw:

> Error: Continuous value supplied to discrete scale

Note that now it mentions the scale, and which values are unexpected.

``` r
#> ℹ Loading ggplot2

# See #4258
ggplot() + 
  geom_point(aes(x = "A", y = 2)) +
  scale_x_continuous()
#> ! Discrete values supplied to continuous scale
#> ℹ Example values: "A"
```

Previously the following would throw the *very* awkward:

> #> Warning in log(x, base): NaNs produced
#> Warning: Transformation introduced infinite values in discrete y-axis

Now it mentions the *correct* scale and the transformation.

``` r
# See #4185
ggplot() +
  geom_point(aes(1, 1, colour = -1)) +
  scale_colour_viridis_c(trans = "log10")
#> Warning in log(x, base): NaNs produced
#> Warning in scale_colour_viridis_c(trans = "log10"): 
#> log-10 transformation introduced infinite values.
```

I also passed down some calls to several checkers, which previously simply didn't mention the scale.

``` r
scale_alpha_continuous(breaks = 1:2, labels = "A")
#> Error in `scale_alpha_continuous()`:
#> ! `breaks` and `labels` must have the same length
```

<sup>Created on 2023-07-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Lastly, I also have an alternate reason for this PR. If we'd ever want to solve https://github.com/tidyverse/ggplot2/issues/4269, we'd need to know which parameters we're allowed to override. For example, if the user provides `limits` in a proper scale, but not `breaks`, we should be allowed to override the `breaks` (and *maybe* `limits`). In any case, we'd need to figure out what arguments were user supplied. This is now a lot easier:

``` r
scale <- scale_colour_discrete(breaks = c("A", "B", "C"), name = "My Scale")
call_args_names(scale$call)
#> [1] "breaks" "name"
```


R/scale-colour.R
```diff
@@ -78,13 +78,18 @@
...
 
   if (is.function(type)) {
-    check_scale_type(type(...), "scale_colour_continuous", "colour")
+    if (!any(c("...", "call") %in% fn_fmls_names(type))) {
+      args$call <- NULL
+    }
+    check_scale_type(exec(type, !!!args), "scale_colour_continuous", "colour")
   } else if (identical(type, "gradient")) {
-    scale_colour_gradient(...)
+    exec(scale_colour_gradient, !!!args)
```
thomasp85, line 90: while not major I think we should use inject over exec for reasons I can't quite formulate

R/scale-colour.R
```diff
@@ -78,13 +78,18 @@
...
   } else if (identical(type, "viridis")) {
-    scale_colour_viridis_c(...)
+    exec(scale_colour_viridis_c, !!!args)
```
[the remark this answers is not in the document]
thomasp85: and there are more below - not commenting on them

R/scale-colour.R
teunbrand: Yeah that was my intuition as well, but that ran into some problems I'm not skilled enough to debug properly.

If I change this to `inject(scale_colour_gradient(!!!args))` I get the following error: 
`Error: evaluation nested too deeply: infinite recursion / options(expressions=)?`.

So unless you have very strong feelings about this, I'd prefer to keep the `exec` variant.

R/scale-.R
```diff
@@ -401,9 +416,8 @@ Scale <- ggproto("Scale", NULL,
 
   call = NULL,
   aesthetics = aes(),
-  scale_name = NULL,
   palette = function() {
-    cli::cli_abort("Not implemented")
+    cli::cli_abort("Not implemented.")
```
thomasp85, line 420: Not hugely important but shouldn't these reference `self$call` as well

R/scale-.R
```diff
@@ -419,7 +433,7 @@ Scale <- ggproto("Scale", NULL,
 
 
   is_discrete = function() {
-    cli::cli_abort("Not implemented")
+    cli::cli_abort("Not implemented.")
```
thomasp85, line 436: same as above

thomasp85 (approved):
Apart from the few comments LGTM

teunbrand:
I'm merging this despite Thomas' comments (to be fair, 'not hugely important' eerily looks like 'good enough').
The messages that currently lack a reported call are mostly developer-facing, and users should never see these.
Combined with a mild reluctance on my part to change the function signature: I'm happy with this as-is.

R/scale-continuous.R
```diff
@@ -105,12 +110,17 @@ scale_y_continuous <- function(name = waiver(), breaks = waiver(),
                                na.value = NA_real_, trans = "identity",
                                guide = waiver(), position = "left",
                                sec.axis = waiver()) {
+  call <- caller_call()
+  if (is.null(call) || !startsWith(as.character(call[[1]]), "scale_")) {
```
averissimo, line 114: I think this change created a problem when calling `scale_y_continous` within another package function via the double colon.

`as.character(call[[1]])` contents is of length 3 and the if clause fails

```R
# [1] "::"         "tern"       "g_lineplot"
```

averissimo: My guess is that `caller_call` is catching the parent call instead of the `scale_*` name

PR #5351, opened 2023-07-17
Update `ggsave()`'s docs

### LINKED ISSUE #5341: ggsave does not respect ggplot2::unit in width and height arguments, defaults to inches
I believe I've found a problem with the `ggsave` function. When denoting width and height using the `unit` function imported from the `grid` package. `ggsave` accepts the arguments without problem but actually takes the numeric value to mean inches, instead of the unit denoted.

Using the `units =` argument fixes this, so perhaps the only changes are warnings and changing the documentation. Let me know if you agree, and I could file a PR if needed.

``` r
library(ggplot2)

ggplot(mpg, aes(x = cty, y = hwy)) +
  geom_point()
```

``` r

ggsave("test.png", 
       width = unit(297, "mm"), 
       height = unit(210, "mm"))
#> Error in `ggsave()`:
#> ! Dimensions exceed 50 inches (`height` and `width` are specified in
#>   inches not pixels).
#> ℹ If you're sure you want a plot that big, use `limitsize = FALSE`.
#> Backtrace:
#>     ▆
#>  1. └─ggplot2::ggsave("test.png", width = unit(297, "mm"), height = unit(210, "mm"))
#>  2.   └─ggplot2:::plot_dim(...)
#>  3.     └─cli::cli_abort(...)
#>  4.       └─rlang::abort(...)
```

<sup>Created on 2023-06-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5341.

Briefly, the old argument description was somewhat suggestive that `ggsave()`'s width and height arguments should be supplied as `grid::unit()` objects. This PR separates the description of width/height from the unit description to be more clear about the expected input.

In addition, it was suggested in https://github.com/tidyverse/ggplot2/issues/5323#issuecomment-1615228420 that the docs could clarify that the device may be guessed from the filename extension and that there was no example of passing device arguments through `...`. This is also included.

### DIFF
--- a/R/save.R
+++ b/R/save.R
@@ -26,13 +26,16 @@
 #' @param plot Plot to save, defaults to last plot displayed.
 #' @param device Device to use. Can either be a device function
 #'   (e.g. [png]), or one of "eps", "ps", "tex" (pictex),
-#'   "pdf", "jpeg", "tiff", "png", "bmp", "svg" or "wmf" (windows only).
+#'   "pdf", "jpeg", "tiff", "png", "bmp", "svg" or "wmf" (windows only). If
+#'   `NULL` (default), the device is guessed based on the `filename` extension.
 #' @param path Path of the directory to save plot to: `path` and `filename`
 #'   are combined to create the fully qualified file name. Defaults to the
 #'   working directory.
 #' @param scale Multiplicative scaling factor.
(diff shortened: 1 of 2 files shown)


yutannihilation (approved):
Looks good to me.

teunbrand:
Thanks for the review!

Issue #1344, opened 2015-09-17
facet_grid() with scales="free" option won't free y-axis actually


yutannihilation:
When I use `scales = "free"` option in ggplot2, only X-axis is freed. Is this the intended result?

```r
library(ggplot2)
mt <- ggplot(mtcars, aes(mpg, wt, colour = factor(cyl))) + geom_point()
mt + facet_grid(. ~ cyl, scales = "free")
```

![plot](https://cloud.githubusercontent.com/assets/1978793/9948172/df9b3fc0-5ddb-11e5-8c1d-5dfc9ac7147e.png)

The option works fine with `facet_wrap()`; Both X-axis and Y-axis is freed as shown bellow. I expect this behaviour 

```r
mt + facet_wrap(~ cyl, scales = "free")
```

![plot2](https://cloud.githubusercontent.com/assets/1978793/9948184/02051a4a-5ddc-11e5-97f3-7fe8698ae53a.png)

yutannihilation:
Oh, sorry, I overlooked this issue...

https://github.com/hadley/ggplot2/issues/1152

Now I guess this is your design decision, so I close this issue.

Issue #1152, opened 2015-07-02
Is there a way to allow scales in facet_grid to be completely "free"?


kendonB:
Suppose we have an n x m grid.

Currently:
* ``scales = free_x`` allows for m scales along the bottom, and 1 common scale for all rows.
* ``scales = free_y`` allows for n scales along the side, and 1 common scale for all columns.
* ``scales = free`` allows for n scales along the side, and m scales along the bottom.

Is there a way to have all scales free? i.e. this would result in n*m independent scales, one for each subplot?

hadley:
Nope, sorry. If you want that, you can use `facet_wrap`.

PR #5354, opened 2023-07-17
Free scales for `coord_polar()`

### LINKED ISSUE #2815: Awkward interaction between faceting, polar coordinates, and free scales
The concept of free scales under faceting doesn't quite work the way I would think it should when we're dealing with polar coordinates.

If we make set of facets with simple polar bar plots, we get partial pies because by default the scales are the same across facets:
``` r
library(ggplot2)

ggplot(mpg, aes(x = 1, fill = drv)) + 
  geom_bar() +
  coord_polar(theta = "y") +
  facet_wrap(~manufacturer)
```

Trying to make the scales free, so that each facet has its own theta scale, doesn't work, because `coord_polar()` needs a fixed aspect ratio and hence `coord_polar()$is_free()` returns `FALSE`:
``` r

ggplot(mpg, aes(x = 1, fill = drv)) + 
  geom_bar() +
  coord_polar(theta = "y") +
  facet_wrap(~manufacturer, scales = "free_y")
#> Error: coord_polar doesn't support free scales
```

However, we can make a version of `coord_polar()` that returns `TRUE` and it works just fine and makes the pie charts as expected. But then the aspect ratio is wrong:
``` r
cp <- coord_polar(theta = "y")
cp$is_free <- function() TRUE

ggplot(mpg, aes(x = 1, fill = drv)) + 
  geom_bar() +
  cp +
  facet_wrap(~manufacturer, scales = "free")
```

It seems to me that there are two separate concepts that are being muddled together, the aspect ratio of the facets and the limits of the scales. For most coordinate systems these two concepts are linked one-to-one, but for some (like `coord_polar()`) they are not.

teunbrand:
This PR aims to fix #2815.

Briefly, it implements the suggestion to have `CoordPolar$is_free()` return `TRUE`. 

To deal with the aspect ratio, the requirement for fixed scales to have  `Coord`s affect the aspect ratio is dropped. I have so far failed to come up with examples where this may lead to awkward situations, so I hope I'm avoiding Chesterton's Fence here. This requirement was originally implemented in cd11888, but `coord_equal()`'s behaviour appears unaffected by this PR.

Also, a complication arose in https://github.com/tidyverse/ggplot2/issues/2815#issuecomment-698932620.  With this PR, `coord_equal()` can also have free scales in `facet_grid()`, but these are of course locked to the column for x-axes and to the rows for y-axes (see also #1344 and #1152, which might be a topic revisiting separately from this PR). Because `theta` can be either axis, it has no special relation with either horizontal or vertical directions, which makes this a bit awkward.


thomasp85:
is this tangentially related enough to the milestone for inclusion?

teunbrand:
Might be nice, but should double check if this holds for `coord_radial()` as well

teunbrand:
Yeah lets include this in the milestone, why not?

thomasp85:
is it ready for review?

teunbrand:
Yes, I'll request it

thomasp85 (approved):
LGTM

teunbrand:
Thanks!

PR #5357, opened 2023-07-19
Binned limits and reverse transform

### LINKED ISSUE #5355: Binned scales do not works with `NA` limits and cannot be reversed.
I apologise for posting two problems in the same issue.

Binned scales cannot be reversed, as reported in https://github.com/r-lib/scales/issues/389.
I expected these scales to be reversible.

``` r
library(ggplot2)

p <- ggplot(mtcars) + geom_bar(aes(mpg))

p + scale_x_binned(trans = "reverse")
```

Moreover, binned scales don't follow the convention where you can set one of the limits to `NA`  to take the 'natural' limit of the data.
I expect this scale to start at 10 and not throw errors.

``` r
p + scale_x_binned(limits = c(10, NA))
#> Error in if (zero_range(range)) zero_width else diff(range): missing value where TRUE/FALSE needed
```

<sup>Created on 2023-07-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5355.

Briefly, it now accepts `NA` in the limits to fill in the 'natural' limits, by using the continuous scale's `get_limits()`. Also, it tracks if limits have been supplied in reverse order and handles breaks correspondingly.

Small demo:

``` r
#> ℹ Loading ggplot2
library(patchwork)

p <- ggplot(mtcars) + geom_bar(aes(mpg))

p1 <- p + scale_x_binned()
p2 <- p + scale_x_binned(
  trans = "reverse", 
  right = FALSE # Just to be symmetrical with p1
) 

p1 | p2
```

<sup>Created on 2023-07-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
Thanks for the review!

PR #5358, opened 2023-07-19
Resolve R devel visual tests

teunbrand:
This PR aims to fix an issue with the visual tests on R devel.

Briefly, R [changed](https://developer.r-project.org/blosxom.cgi/R-devel/NEWS/2023/05/03#n2023-05-03) the `density()` calculation, which means that visual tests of e.g. `geom_violin()` and `geom_density()` fail on R devel. The way this is resolved is by conditionally overriding the default density method to use the backward compatibility argument `old.coord`.


yutannihilation (approved):
Looks awesome!

teunbrand:
Thanks for the review!

PR #5361, opened 2023-07-24
Passing parameters to `trans` argument

### LINKED ISSUE #5359: `scale_*_continuous` help does not explain how to use built-in transformation objects that require other arguments
The help for scale_*_continuous suggests that you can use a character name for the built-in transformation objects.

image

But it does not explain how to do this when the transformation object requires other arguments.

An example is "modulus" which requires a p argument

``` r
library(tidyverse)
  pressure |>
    ggplot() +
    geom_point(aes(x = temperature, , 
                   y = pressure)) +
    scale_y_continuous(trans = "modulus")
#> Error in match.fun(f)(): argument "p" is missing, with no default
```

<sup>Created on 2023-07-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5359.

The scales documentation now mentions how to pass parameters to transformations fed into the `trans` argument.

### DIFF
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -73,9 +73,10 @@
 #'
 #'   A transformation object bundles together a transform, its inverse,
 #'   and methods for generating breaks and labels. Transformation objects
-#'   are defined in the scales package, and are called `<name>_trans` (e.g.,
-#'   [scales::boxcox_trans()]). You can create your own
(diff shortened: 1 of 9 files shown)


thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review!

PR #5364, opened 2023-07-25
fix: obtain expected quantity values from manual scale

liang09255:
This PR fixes an error in the code and adds corresponding unit tests.

When using `manual_scale` to create a `discrete_scale`, the `palette` should be a function that can return the specified number of `values` when called with a single integer argument.

Here is the parameter description excerpt from `scale-.R`:

>@param palette A palette function that when called with a single integer argument (the number of levels in the scale) returns the values that they should take (e.g., [scales::hue_pal()]).

In the original `manual_scale` method, it simply returned all the `values`. However, when `values` are not named, we only need to return the corresponding first 'n' values. Therefore, the modified code includes a condition to handle this case.

### DIFF
--- a/R/scale-manual.R
+++ b/R/scale-manual.R
@@ -180,7 +180,7 @@ manual_scale <- function(aesthetic, values = NULL, breaks = waiver(), ...,
     if (n > length(values)) {
       cli::cli_abort("Insufficient values in manual scale. {n} needed but only {length(values)} provided.")
     }
-    values
+    unname(values[seq_len(n)])
   }
   discrete_scale(aesthetic, "manual", pal, breaks = breaks, limits = limits, ...)
 }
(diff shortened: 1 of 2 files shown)


teunbrand:
The case where a palette returns more values than requested by `n` is handled internally in `ScaleDiscrete$map()`.  This also handles other cases for e.g. custom palettes:

``` r
library(ggplot2)

  geom_point() +
  discrete_scale(
    "colour", "colour",
    palette = function(n) rainbow(10) # more than the 4 needed for `factor(cyl)`
  )
```



<sup>Created on 2023-07-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

What I'm not understanding here is why this should be fixed if this case is dealt with already.

Issue #5216, opened 2023-03-05
geom_bin_2d: warning message when using weight without group


szimmer:
I found a problem with geom_bin_2d

I expected to be able to use a weight variable without a group variable but get a warning when I do this. The weight does seem to be used appropriately as the result is different when not using the weight.

Here is the code to reproduce the bug:

``` r
library(survey)
#> Loading required package: grid
#> Loading required package: Matrix
#> Loading required package: survival
#> 
#> Attaching package: 'survey'
#> The following object is masked from 'package:graphics':
#> 
#>     dotchart
library(ggplot2)

data(api)

ggplot(aes(x=api99, y=api00, weight=pw), data=apiclus2) +
  geom_bin_2d() 
#> Warning: The following aesthetics were dropped during statistical transformation: weight
#> ℹ This can happen when ggplot fails to infer the correct grouping structure in
#>   the data.
#> ℹ Did you forget to specify a `group` aesthetic or to convert a numerical
#>   variable into a factor?
```

![](https://i.imgur.com/QGg5YVG.png)

``` r

ggplot(aes(x=api99, y=api00), data=apiclus2) +
  geom_bin_2d() 
```

![](https://i.imgur.com/st76C6F.png)

<sup>Created on 2023-03-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Standard output and standard error
</summary>

``` sh
-- nothing to show --
```

</details>
<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.2.2 (2022-10-31 ucrt)
#>  os       Windows 10 x64 (build 22621)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  English_United States.utf8
#>  ctype    English_United States.utf8
#>  tz       America/New_York
#>  date     2023-03-05
#>  pandoc   2.19.2 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.2.1)
[...]

PR #5372, opened 2023-07-27
Boxplot staples

teunbrand:
This PR aims to implement one of the features mentioned in #5216, namely: staples for boxplots.

Briefly, it adds a new argument `staplewidth` argument to `geom_boxplot()` that allows one to draw staples at the end of whiskers.

A few additional notes:
* Staples are only drawn when `staplewidth != 0` and `0` is the default, so existing plots won't be affected.
* The `staplewidth` is *relative* to the width of the box.
* Staples are styled in the same way as whiskers.

Some demonstrations:

Here you can see that the `staplewidth` is relative, as it adapts to the size of variable-width boxes.

``` r
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(drv, displ))

p + geom_boxplot(staplewidth = 1, varwidth = TRUE)
```

Here you can see that it works as one would expect with a trick to make half-boxplots:

``` r
p + geom_boxplot(aes(xmax = after_scale(x)), staplewidth = 0.5)
```

With negative `staplewidth` you can flip the staples in half-boxplots.

``` r
p + geom_boxplot(aes(xmax = after_scale(x)), staplewidth = -0.5)
```

<sup>Created on 2023-07-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

Issue #2004, opened 2017-01-18
Customizing legend key for geom_text and geom_label


alberthkcheng:
Currently, the legend key is fixed to "a" because of the following code in legend-draw.r:

```
draw_key_text <- function(data, params, size) {
  textGrob("a", 0.5, 0.5,....
```

What i need to do is, when my label is showing numbers only, i would like to have a number, e.g. "1" on the legend key instead of "a".

I tried to use the following but it does not look beautiful (in terms of code clarity):
```
draw_key_text <- function(data, params, size) {
  textGrob(params$legend.key, 0.5, 0.5,....
}

geom_text <- function(....
                      key.legend = "a",..){
.... layer(....params = list(...key.legend = key.legend,...))
}
```

Is there any suggestion?

hadley:
This seems like a good idea, but unfortunately it's not high enough on my priority list that I will implement it myself. However, I'd be happy to review a PR. The key challenge will be finding the right way to get the information to the legend - I suspect that the `guide_legend()` function would be the right place to set it.

PR #5389, opened 2023-08-12
add function `scale_label_identity` function

Yunuuuu:
Through these commits, we are enabled to effortlessly exhibit legend labeling using the `guide` parameter within the `scale_label_identity` function, aligning it with analogous aesthetic settings for other geoms. Furthermore, the assignment of the label key can be readily accomplished through the intrinsic data within the `draw_key_` function (https://github.com/tidyverse/ggplot2/issues/2004). This bring `scale_label_identity` function and catch following concers:
1. label as a geom aesthetic should appear in the scaled data from `draw_key_text` function.  As we can see with following code
```
library(ggplot2)
# Create sample data
data <- data.frame(
    x = c(1, 2, 3, 4, 5),
    y = c(1, 2, 3, 4, 5),
    label = c("A", "B", "C", "D", "E")
)
# Create the plot with geom_text
ggplot(data, aes(x, y)) +
    geom_text(aes(label = label, color = label),
        key_glyph = function(data, params, size) {
            browser()
        }
    )
```
image

2. users are inclined to modify the parameter show.legend to exhibit textual aesthetics (only to discover that it cannot fulfill their intended purpose), while the help document means the default value of `show.legend` shoule have include the mapping aesthetic:
```
show.legend: logical. Should this layer be included in the legends? NA, the default, includes if any aesthetics are mapped. FALSE never includes, and TRUE always includes. It can also be a named logical vector to finely select the aesthetics to display.
```

3. The hard-coded label in the `draw_key_text` function appears superfluous.


R/legend-draw.R
```diff
@@ -236,7 +236,6 @@ draw_key_smooth <- function(data, params, size) {
 #' @export
 #' @rdname draw_key
 draw_key_text <- function(data, params, size) {
-  if(is.null(data$label)) data$label <- "a"
```
teunbrand, line 239: I'm a little bit worried with this line.
I fear that the following code wouldn't draw labels in the legend:

```r
  geom_text(aes(
    label = rownames(mtcars), 
    colour = factor(cyl)
  ))
```

Could you confirm and amend if necessary?

teunbrand:
Overall I'm a little bit cautious about the need for such function, which might have been fruitful to discuss in an issue before posting a PR. Labels typically aren't mapped to a scale and introducing `scale_label_identity()` might prompt the expectation of various other label scales. I don't think the identity scale is a problem, but the other ones are tricky.

Yunuuuu:
Thanks for your comment, I wanted bring it into ggplot2 package for these issues: https://github.com/tidyverse/ggplot2/issues/2004, https://stackoverflow.com/questions/49965758/change-geom-texts-default-a-legend-to-label-string-itself?noredirect=1&lq=1, https://stackoverflow.com/questions/18337653/remove-a-from-legend-when-using-aesthetics-and-geom-text?noredirect=1&lq=1,  https://stackoverflow.com/questions/59091627/add-new-legend-for-geom-text-with-text-labels-as-legend-key, without scale_label function, the data in `draw_key_text` function will not contain the label informations, which has prompted users to modify the internal `grob` object or `GEOM` object which  shouldn't be touched in usual.

image

Yunuuuu:
I have noticed that the existence of the `scale_label` function leads to the incorporation of the `label` data within the `draw_key_text` function. Evidently, many users, including myself, have encountered difficulties with the hard-coded "a" in the textual legend. I have conducted a search on Stack Overflow and found several questions that are similar to this issue.

Yunuuuu:
Perhaps it might be advisable not to alter the `draw_key_text` function. Instead, could we explore the possibility of integrating the `scale_label_identify` function? Alternatively, are there any superior approaches to address the hardcoded "a" issue?

teunbrand:
I'll admit that the current way of dealing with this is not ideal. I'd say the recommended route without this PR would be to use the `override.aes` argument of the guide and manually relabelling it.

``` r
library(ggplot2)
#> Warning: package 'ggplot2' was built under R version 4.3.1

data <- data.frame(
  x = c(1, 2, 3, 4, 5),
  y = c(1, 2, 3, 4, 5),
  label = c("A", "B", "C", "D", "E")
)

ggplot(data, aes(x, y)) +
  geom_text(aes(label = label, color = label)) +
  guides(colour = guide_legend(override.aes = list(label = LETTERS[1:5])))
```



You could still have an identity scale, but you'd need to set the guide for it to work as expected.

``` r
scale_label_identity <- function(..., guide = "none") {
  sc <- discrete_scale("label", "identity", force, ...,
                       guide = guide,
                       super = ScaleDiscreteIdentity
  )
  
  sc
}

ggplot(data, aes(x, y)) +
  geom_text(aes(label = label, color = label)) +
  scale_label_identity(guide = "legend")
```



<sup>Created on 2023-08-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

> It has come to my attention that the current draw_key_text function is expected to include the scaled label information as outlined in the documentation

There is a subtle nuance in what is 'mapped' and what is 'scaled'. The `label` aesthetic is typically mapped but not scaled.

Yunuuuu:
Thank you for your explanations. I now realize the differences between "map" and "scale". Given this understanding, it might not be advisable to include such a function in this context. For me, It'll be okay to close the pull request.

yutannihilation:
As it seems closing this pull request is the conclusion here, I'm closing now.

PR #5403, opened 2023-08-30
`<rel>` and `<unit>` theme inheritance

### LINKED ISSUE #3951: Suggestion: `rel`-class can inherit value from `unit`-class theme elements

teunbrand:
This PR aims to fix #3951.

Essentially, anywhere one can specify a `unit()` in the theme for an element that is *not* a root element, one can now use `rel()` to inherit from the parent `unit` object with modification.

In the example below:
* Legend keys are 2x as long as they're wide
* Top x-axis ticks are 2x as long as default
* Bottom x-axis ticks are 4x as long as default, because inheritance passes through 2 `<rel>` objects.

``` r
#> ℹ Loading ggplot2

  geom_point() +
  guides(x.sec = "axis") +
  theme(
    legend.key.height = rel(2),
    axis.ticks.length.x = rel(2),
    axis.ticks.length.x.bottom = rel(2)
  )
```

<sup>Created on 2023-08-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
As a note: it also works on `margin()`s but I don't think any theme margin settings inherit something.

``` r
#> ℹ Loading ggplot2
combine_elements(rel(2), margin(5,5,5,5))
#> [1] 10points 10points 10points 10points
```

<sup>Created on 2023-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/theme.R
```diff
@@ -724,6 +724,19 @@ combine_elements <- function(e1, e2) {
...
+    if (is.rel(e2)) {
+      return(rel(unclass(e1) * unclass(e2)))
+    }
+    # If e2 is a unit, return modified unit
+    if (is.unit(e2)) {
+      return(unclass(e1) * e2)
+    }
+    return(e1)
+  }
+
```
thomasp85, line 739: We might want to build in a check for whether `e2` is numeric and apply rel to that, just for completeness

thomasp85 (approved):
Minor thing but otherwise good to go

R/theme.R
teunbrand: I was today years old when I learned:

``` r
library(grid)
is.numeric(unit(1, "npc"))
#> [1] TRUE
is.numeric(unit(1, "npc") + unit(1, "cm"))
#> [1] FALSE
```

<sup>Created on 2023-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #5409, opened 2023-09-06
Simplify scales$add_default

### LINKED ISSUE #4873: stage() should show a bit friendlier warning when used without after_stat and after_scale

### LINKED ISSUE #5408: Using stage with non-scaled aesthetics
Somewhat contrived example but able to solve real issues in gganimate:

```r
data <- data.frame(test=1:3, x = 0, y = 0)
ggplot(data, aes(x=x, y=y)) +
    geom_text(aes(label=stage(test, after_scale = format(label, digits = 3))))
# Error in `geom_text()`:
# ! Problem while mapping stat to aesthetics.
# ℹ Error occurred in the 1st layer.
# Caused by error in `stage()`:
# ! object 'test' not found
```

This surprisingly happens while looking for a default scale for `label` in `layer$map_statistic()`

thomasp85:
Fix #5408 and fix #4873

This PR fixes some corner case issues around the use of `stage()`. However, it also just cleans up the `add_default()` methods of `Scales`. Originally this method did a second evaluation of newly added aesthetics for the sake of deducing their type, but that information was already created in the calling function as part of the standard aesthetic evaluation. So, instead, the method is now passed in the evaled aesthetics rather than original data+aesthetics and saves a number of reevaluations


R/aes-evaluation.R
```diff
@@ -313,7 +313,7 @@ strip_stage <- function(expr) {
   } else if (is_call(uq_expr, "stage")) {
     # Prefer stat mapping if present, otherwise original mapping (fallback to
     # scale mapping) but there should always be two arguments to stage()
-    uq_expr$after_stat %||% uq_expr$start %||% (if (is.null(uq_expr$after_scale)) uq_expr[[3]]) %||% uq_expr[[2]]
+    uq_expr$after_stat %||% uq_expr$start %||% uq_expr[[1]] %||% (if (is.null(uq_expr$after_scale)) uq_expr[[3]]) %||% uq_expr[[2]]
```
teunbrand, line 316: I'm a bit unsure about the assumption that `uq_expr[[1]]` is the `start` argument.
Wouldn't it be safer to use `rlang::call_match(uq_expr, stage)` first?
I'm guessing that might also make the part after `if` redundant too.

teunbrand (approved):
LGTM, but small comment you might like to consider.
Does this happen to also fix #4873?
CI on devel is failing (or probably will fail at time of writing) because {vdiffr} has trouble installing, but you might've been aware of that.

R/aes-evaluation.R
thomasp85: good call

teunbrand:
Yes the one line looks much clearer now. Do you wish to merge this before new merge conflicts arise due to updates to the news file?

teunbrand:
Yes the one line looks much clearer now. Do you wish to merge this before new merge conflicts arise due to updates to the news file?

PR #5410, opened 2023-09-07
POC: secondary axis functionality in `guide_axis()`

teunbrand:
This PR is a proof of concept for putting the functionality of secondary axes in `guide_axis()`.

The long term goal may be to lift over responsibilities for secondary axis from the scale system to the guide system.
In this PR, `guide_axis()` gains the arguments `breaks`, `labels`, `minor.breaks` and `trans` that would be needed to have secondary axis functionality. This makes `guide_axis()` a lot more flexibile, as these arguments can be used to override what the scale passes down. Under the hood, the whole transforming values system gets captured in a new `<trans>` with the scales package. Some demos below.

This is just to show that it can work similarly to using `sec_axis()`.

``` r
#> ℹ Loading ggplot2

p <- ggplot(pressure, aes(temperature, pressure)) + geom_line() +
  labs(x = "Temperature (\u00b0C)")

p + guides(x.sec = guide_axis(
  title = "Temperature (\u00b0F)",
  trans = ~.x * 9/5 + 32
))
```

Because we capture the transformation in a `<trans>` object, we can also provide one directly. There is a little bit of awkwardness in this, as the forward `transform()` translates secondary data to primary range and `inverse()` goes from primary to secondary, which is opposite to the intuition of how `trans` works in `sec_axis()`.

``` r
my_trans <- trans_new(
  "temperature_trans",
  transform = \(x) (x - 32) * 5/9, # Fahrenheid to Celsius
  inverse   = \(x) x * 9/5 + 32    # Celsius to Fahrenheid
)

p + guides(x.sec  = guide_axis("Temperature (\u00b0F)", trans = my_trans))
```

Show that one can override the default breaks and labels:

``` r
p + guides(x.sec = guide_axis(
  breaks = breaks_width(50),
  labels = \(x) paste0(x, "\u00b0C"))
)
```

This PR also makes it possible to have secondary axes for discrete scales with custom labels.

``` r
ggplot(mpg, aes(drv, displ)) +
  geom_boxplot() +
  guides(x.sec = guide_axis(labels = c("Foo", "Bar", "Baz")))
```

<sup>Created on 2023-09-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Currently the minor breaks aren't implemented yet because #5287 provides some extra infrastructure for this.


teunbrand:
This still needs some more thought around scales that already have non-identity transformations.

teunbrand:
Now works better with transformed scales:

``` r
#> ℹ Loading ggplot2
library(patchwork)

p <- ggplot(pressure, aes(temperature, pressure)) + geom_line() +
  labs(x = "Temperature (\u00b0C)")

p1 <- p + scale_y_log10(sec.axis = sec_axis(sqrt))
p2 <- p + scale_y_log10() + guides(y.sec = guide_axis(trans = sqrt))
p1 | p2
```



Though it isn't identical w.r.t labels yet.

``` r
p1 <- p + scale_y_continuous(sec.axis = sec_axis(log10))
p2 <- p + guides(y.sec = guide_axis(trans = log10))
p1 | p2
#> Warning in self$trans(range): NaNs produced
#> Warning in fun(origin_seq): NaNs produced
```



<sup>Created on 2023-09-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
Closing this in favour of #5620

PR #5413, opened 2023-09-08
Correct length() usage in geom-boxplot

MichaelChirico:
Testing out the new `lintr::length_test_linter()` (https://github.com/r-lib/lintr/pull/2124) came across this. I think the code does as intended "by accident", but it's better to use the "correct" code instead and not rely on some coercions producing the right answer incidentally.


teunbrand (approved):
This subtle mistake has been present for ~15 years since 11279f6ee8f7238d5ec0a89d2e872da2cfb468ab. 
Looks good to me, thanks for the PR!

PR #5418, opened 2023-09-13
POC: `ignore()` for aesthetic evaluation.

teunbrand:
This PR explores https://github.com/tidyverse/ggplot2/pull/5290#issuecomment-1698603368.

While #5290 tackles the same issue on the data side of things, this PR aims to use mechanics similar to `after_scale()`.
Briefly, this PR introduces `ignore()` as an aesthetic evaluation modifier, which then causes scales to skip doing anything with or calculating anything from these modified aesthetics.

Here are some notes:
* The information on which aesthetics are ignored are stored in `Layer$computed_mapping`. Scales have plenty of opportunities to read/write from/to `data` independent of the layer. This caused an issue on how to transfer that information to the scales.
* The solution in this PR is to rename the ignored columns any time data interacts with scales. Scales match column names based on their aesthetics, so this effectively lets these ignored aesthetics be ignored. Any time the layer wants to compute anything on the data (stat, position, geom etc.), the columns are renamed back to their original names.
* That renaming strategy kind-of works until we arrive at coord transformation, which is typically under control of a `Layer$geom$draw_*()` function, which needs the original names. The 'solution' in this PR is to temporarily redefine the definition of position aesthetics in the `Layer$draw_geom()` method.
* I took some additional care in `annotate()` to preserve `ignore()`'d variables.

At a first glance, results are pretty similar to #5290.

``` r
devtools::load_all("~/packages/ggplot2") # This PR
#> ℹ Loading ggplot2

df <- data.frame(t = seq(0, 2*pi, length.out = 100))

ggplot(mpg) +
  geom_point(aes(displ, hwy, colour = factor(cyl))) +
  geom_path(
    data = df, 
    aes(x = ignore(cos(t) / 2.1 + 0.5), 
        y = ignore(sin(t) / 2.1 + 0.5),
        colour = ignore(2))
  ) +
  annotate(
    "text", x = ignore(0.5), y = ignore(0.5),
    label = "I'm in the middle",
    colour = 4, size = 10
  )
```

<sup>Created on 2023-09-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

However, this is only the case for very straightforward layers that do not reparametrize anything. For example, if you'd have to translate x/width parameters to xmin/xmax parameters, this approach doesn't keep track of what should be ignored.

``` r
devtools::load_all("~/packages/ggplot2") # This PR
#> ℹ Loading ggplot2
df <- data.frame(x = c(0.3, 0.5, 0.7), y = c(1, 2, 3))

ggplot(df, aes(ignore(x), y)) +
  geom_col() +
  xlim(0, 10)
```

<sup>Created on 2023-09-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Whereas #5290's approach bakes the ignoring part into the data, which begets the intended effect for these bars.

``` r
devtools::load_all("~/packages/ggplot2") # PR #5290
df <- data.frame(x = c(0.3, 0.5, 0.7), y = c(1, 2, 3))

ggplot(df, aes(I(x), y)) +
  geom_col() +
  xlim(0, 10)
```

<sup>Created on 2023-09-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

At this point, I think #5290 has more benefits than this approach.

### DIFF
--- a/R/aes-evaluation.R
+++ b/R/aes-evaluation.R
@@ -204,6 +204,12 @@ stage_scaled <- function(start = NULL, after_stat = NULL, after_scale = NULL) {
   after_scale
 }
 
+#' @rdname aes_eval
+#' @export
+ignore <- function(x) {
+  x
+}
+
 # Regex to determine if an identifier refers to a calculated aesthetic
 match_calculated_aes <- "^\\.\\.([a-zA-Z._]+)\\.\\.$"
 
@@ -221,6 +227,10 @@ is_scaled_aes <- function(aesthetics) {
 is_staged_aes <- function(aesthetics) {
   vapply(aesthetics, is_staged, logical(1), USE.NAMES = FALSE)
 }
+is_ignored_aes <- function(aesthetics) {
+  vapply(aesthetics, is_ignored, logical(1), USE.NAMES = FALSE)
+}
+
 is_calculated <- function(x, warn = FALSE) {
   if (is_call(get_expr(x), "after_stat")) {
     return(TRUE)
@@ -262,6 +272,9 @@ is_scaled <- function(x) {
 }
 is_staged <- function(x) {
   is_call(get_expr(x), "stage")
+}
+is_ignored <- function(x) {
+  is_call(get_expr(x), "ignore")
 }
 
 # Strip dots from expressions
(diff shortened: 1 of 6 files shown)


thomasp85:
Remind me what part of #5290 that compelled us to explore this direction?

teunbrand:
Well the #5290 touches a lot of places to make exceptions for `AsIs` objects that we thought might be simplified by using the `stage()/after_stat()/after_scale()` approach. I generally think it was good to explore, but I'd still prefer the approach in #5290. However, we might take the renaming approach from this PR and apply it to `AsIs` columns instead of `ignore()`'d aesthetics and the footprint of the PR might be a lot smaller and work well.

thomasp85:
Ok - let's discuss next meeting

teunbrand:
Closing this in favour of #5477

PR #5420, opened 2023-09-14
Don't attach Suggests package in test

MichaelChirico:
Doing so makes tests less hermetic since `library()` has permanent side effects (of course, `loadNamespace()` does too, but generally less serious).


teunbrand (approved):
Thanks for the PR! This looks good to me

PR #5426, opened 2023-09-18
Delete utilies-table.R

teunbrand:
The functions defined in this file aren't use anywhere, aren't exported, and the last meaningful change was ~9 years ago.
Deleting this file will lower future upkeep costs a bit.


thomasp85 (approved):
My best guess is it is vestigal from when stable functionality was part of ggplot2

PR #5430, opened 2023-09-20
Subtheme functions

### LINKED ISSUE #5301: Feature Request: Syntactic sugar for theme
I have two ideas for making themes slightly easier to use.

### Element shortcuts

Often, you simply want to switch a theme element on or off.
For example if you want to put axis lines on and legend titles off, you can use:

```r
  geom_point() +
  theme(
    legend.title = element_blank(),
    axis.line    = element_line()
  )
```

It would be neat if we could write:

```r
  geom_point() +
  theme(
    legend.title = "blank",
    axis.line    = "line"
  )
```

### Subtheme

Another idea for theme specification, is that we could have subtheme functions. So instead of writing this:

```r

  geom_point() +
  theme(
    axis.line.x.bottom  = element_line(colour = "red"),
    axis.ticks.x.bottom = element_line(colour = "red"),
    axis.text.x.bottom  = element_text(colour = "red")
  )
```

We could write this:

```r
  geom_point() +
  theme_axis_bottom(
    line  = element_line(colour = "red"),
    ticks = element_line(colour = "red"),
    text  = element_text(colour = "red")
  )
```
Which would then simply feed forward those arguments as the `axis.{part}.x.bottom` arguments to the `theme()`. One could have similar subthemes for the legend, the panel etc. It might also encourage a more ordered approach of theme specification, instead of the wild west that is currently theme declarations.

teunbrand:
This PR aims to fix #5301.

Briefly, it adds functions that allow specifying parts of a theme with shorter arguments.

A few considerations:
*  I tried making a function factory for these subtheme functions, to automatically populate the formals as the argument names are very regex friendly. However, roxygen2 wouldn't recognise the formals in the `@usage` section.
* Should these function be named `subtheme_*()` as to avoid confusion with e.g. `theme_light()` which is conceptually doing a different thing?


thomasp85 (commented):
I do think we need a different prefix than `theme_`, but conceptually I think it should be shorter since the whole point of these functions are to be shortcuts... maybe just `th_`? Or maybe that is too unpronouncable...

thomasp85 (commented):
I think I would probably prefer `theme_sub_*()` then (I know that is longer even). I like that all theme functions have the same (partial) prefix

```
p + 
  theme_minimal() + 
  theme_sub_axis(...)
```

looks more organised than

```
p + 
  theme_minimal() + 
  subtheme_axis(...)
```

Issue #4599, opened 2021-08-31
Improve behaviour when `labeller()` does not find a match in lookup table


stragu:
Currently, `ggplot2::labeller()` silently labels the facets with "NA" if the user supplies unnamed character vectors.

``` r
library(ggplot2)
ggplot(ToothGrowth, aes(y = len)) + 
  geom_boxplot() +
  facet_wrap(vars(supp), labeller = labeller(supp = c("orange juice yum", "vitamin C yuck")))
```

![](https://i.imgur.com/hwFq0rw.png)

If there is a typo in the name (or if there just isn't a new label defined in the vector), the same thing happens: "NA" is given as a title.

``` r
library(ggplot2)
ggplot(ToothGrowth, aes(y = len)) + 
  geom_boxplot() +
  facet_wrap(vars(supp), labeller = labeller(supp = c("OJ" = "orange juice yum", "vC" = "vitamin C yuck")))
```

![](https://i.imgur.com/8mY9fdy.png)

I am not sure what would be most appropriate, but I would have expected to get something like this:

```
Warning: labeller() could not match one or more of the original labels. The facets were titled "NA".
ℹ: did you forget to provide a named character vector, or are some labels missing from the lookup table?
```

The function allows using lookup tables that include _superfluous_ data, which is probably good for flexibility, but I think that naming a facet with "NA" should throw a warning, or at least a message, when the is _not enough_ data.

Or should the default behaviour change to using the original label if no match is found, allowing the use of partial lookup tables? I can see people wanting to only cherry pick a couple of labels to replace.

Using: 
* ggplot2 3.3.5
* R 4.1.1

PR #5432, opened 2023-09-21
Improve labeller() behavior for lookup tables

92amartins:
This is an attempt to fix the issue https://github.com/tidyverse/ggplot2/issues/4599

### DIFF
--- a/tests/testthat/test-labellers.R
+++ b/tests/testthat/test-labellers.R
@@ -20,3 +20,15 @@ test_that("labeller function catches overlap in names", {
     )
   expect_snapshot_error(ggplotGrob(p))
 })
+
+test_that("labeller handles badly specified labels from lookup tables", {
+  df <- data_frame0(am = c(0, 1))
+  labs <- labeller(am = c("0" = "Automatic", "11" = "Manual"))
+  expect_equal(labs(df), list(am = c("Automatic", "1")))
+})
+
+test_that("labeller allows cherry-pick some labels", {
+  df <- data_frame0(am = c(0, 1))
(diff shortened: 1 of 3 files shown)


tests/testthat/test-labellers.R
```diff
@@ -20,3 +20,31 @@ test_that("labeller function catches overlap in names", {
     )
   expect_snapshot_error(ggplotGrob(p))
 })
+
+# Visual tests ------------------------------------------------------------
+
+test_that("labeller handles unspecified labels from lookup tables", {
```
teunbrand, line 26: It might be a bit overkill to do a visual test for a textual change. 
The following is a bit more lightweight test for the behaviour this PR fixes.

```r
df <- data_frame0(am = c(0, 1))

labs <- labeller(am = c("0" = "Automatic", "11" = "Manual"))
expect_equal(labs(df), list(am = "Automatic", "1"))

labs <- labeller(am = c("0" = "Automatic"))
expect_equal(labs(df), list(am = "Automatic", "1"))
```

teunbrand (changes requested):
Thanks for the PR, it looks great! The only change I'd propose is to make the tests a little more direct.

tests/testthat/test-labellers.R
92amartins: Great suggestion! My first instinct was to test in that manner, but I didn't now how to retrieve the _actual_ applied labels to make an `expected_equal` test!

I'll update the tests! Thanks :)

teunbrand (approved):
Great stuff, thanks for your contribution!

Issue #5437, opened 2023-09-26
Release ggplot2 3.4.4


teunbrand:
Prepare for release:

* [x] `git pull`
* [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_ggplot2.html)
* [x] [Polish NEWS](https://style.tidyverse.org/news.html#news-release)
* [x] `devtools::build_readme()`
* [x] `urlchecker::url_check()`
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [x] `devtools::check_win_devel()`
* [x] `rhub::check_for_cran()`
* [x] `revdepcheck::cloud_check()`
* [x] Update `cran-comments.md`
* [x] `git push`

Submit to CRAN:

* [x] `usethis::use_version('patch')`
* [x] `devtools::submit_cran()`
* [x] Approve email

Wait for CRAN...

* [x] Accepted :tada:
* [x] Fix #5478 by merging #5479
* [x] `git push`
* [x] `usethis::use_github_release()`
* [x] `usethis::use_dev_version()`
* [x] `git push`

teunbrand:
For context, in R-devel `is.atomic(NULL)` is no longer `FALSE`.
This hotfix should patch behaviour where ggplot2 relies on that behaviour.

yutannihilation:
@olivroy 
In my understanding, ggplot2 removed rgoes long ago (before version 1.0). Do you see any problem?

https://github.com/search?q=repo%3Atidyverse%2Fggplot2%20rgeos&type=code

teunbrand:
I have no problem cherrypicking 5484b1df349b193c8eaba5c4b07e0755a49dbf0c for the release.

yutannihilation:
Sorry, it was just that I was confused. Thanks for catching.

teunbrand:
Before merging 3.4.4 into main, we should fix #5478 in the news

matthewjnield:
I have submitted #5479 to fix #5478.

PR #5438, opened 2023-09-26
V3.4.4 revdepcheck

teunbrand:
This updates the v3.4.4-test branch with cherry-picked commits from @yutannihilation's #5327 and intended changes for #5437.
The purpose is to run a reverse dependency check to see if we can reasonably fold-in the maptools retirement for the next hotfix.


R/fortify-spatial.R
```diff
@@ -9,32 +9,31 @@
...
-#'    proj4string = CRS("+proj=longlat +datum=NAD27"))
-#'  nc1_df <- fortify(nc1)
-#' }
 NULL
 
 #' @rdname fortify.sp
 #' @export
 #' @method fortify SpatialPolygonsDataFrame
 fortify.SpatialPolygonsDataFrame <- function(model, data, region = NULL, ...) {
+  deprecate_warn0("3.4.3",
```
yutannihilation, line 18: Thanks for handling this! If the version is going to be 3.4.4, maybe we should reflect here?

teunbrand: Yes, for sure. This branch is just for testing; if we merge the maptools retirement into the RC branch we'll put in the proper version number.

yutannihilation: I see. Thanks for the details!

PR #5440, opened 2023-09-27
Move CRS responsibility from GuideAxis to CoordSf

teunbrand:
This PR doesn't fix any issue, but it amends #5293.

I have become mildly frustrated by the lines below in the axis guide:

https://github.com/tidyverse/ggplot2/blob/69e74305386094cb1fc0b10495d18492cec5b012/R/guide-axis.R#L138-L141

Which I believe should be the coord's responsibility, not the axis' responsibility. This PR moves that ressponsibility from the axis to the coord.

### DIFF
--- a/R/coord-sf.R
+++ b/R/coord-sf.R
@@ -261,7 +261,7 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
       graticule = graticule,
       crs = params$crs,
       default_crs = params$default_crs,
-      !!viewscales
+      !!!viewscales
     )
   },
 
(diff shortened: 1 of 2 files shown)


R/coord-sf.R
```diff
@@ -255,19 +255,24 @@ CoordSf <- ggproto("CoordSf", CoordCartesian,
...
 
-    list(
+    list2(
       x_range = x_range,
       y_range = y_range,
       graticule = graticule,
       crs = params$crs,
       default_crs = params$default_crs,
-      viewscales = viewscales
+      !!viewscales
```
teunbrand, line 264: This makes `CoordSf$setup_panel_guides()` redundant and can rely on parental `Coord$setup_panel_guides()`.

thomasp85 (approved):
LGTM

PR #5442, opened 2023-09-27
`coord_sf()` scale function breaks

### LINKED ISSUE #4622: `geom_sf()`: add support for `scale_x_continuous(n.breaks = )`
Currently this argument does not have an effect when using `geom_sf()`.

``` r
library("ggplot2")
ggplot(nc) +
  geom_sf(aes(fill = AREA)) +
  scale_x_continuous(n.breaks = 2)
```

<sup>Created on 2021-09-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

teunbrand:
This PR aims to fix #4622.

Briefly, it adds three conventions of scales that work in other coordinate systems. It allows:
* `breaks = function()` to take long/lat limits and compute breaks.
* `breaks = NULL` to remove axes and gridlines.
* `n.breaks = x` to set a desired number of breaks.

Demonstrations:

``` r
#> ℹ Loading ggplot2

p <- ggplot(nc) +
  geom_sf()

p + scale_x_continuous(breaks = breaks_width(1))
```

``` r
p + scale_x_continuous(breaks = NULL)
```

``` r
p + scale_x_continuous(n.breaks = 9)
```

I have to note that it doesn't fix the case where out-of-bounds breaks get censored. This remains an error:

``` r
p + scale_x_continuous(breaks = -c(75, 78, 80, 82, 85),
                       labels = LETTERS[1:5])
#> Error in `fixup_graticule_labels()` at ggplot2/R/ggproto.R:180:16:
#> ! Breaks and labels along x direction are different lengths
```

<sup>Created on 2023-09-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/coord-sf.R
```diff
@@ -582,6 +591,50 @@ parse_axes_labeling <- function(x) {
...
+      }
+      # Convert bounding box to long/lat coordinates
+      bbox <- sf::st_as_sfc(sf::st_bbox(bbox, crs = crs))
+      bbox <- sf::st_bbox(sf::st_transform(bbox, 4326))
+      bbox <- as.numeric(bbox)
+
+      # If any bbox is NA the transformation has probably failed.
+      # (.e.g from IGH to long/lat). In this case, just provide full long/lat.
+      bbox[is.na(bbox)] <- c(-180, -90, 180, 90)[is.na(bbox)]
+    }
```
teunbrand, line 623: I'd really like some more sf-experienced eyes on this bit

thomasp85: @clauswilke you are perhaps the one best fitting the "sf-experienced" moniker :-)

clauswilke: I haven't looked carefully yet and I'm not entirely sure I even understand what this code is meant to do, but one thing I notice right away is hard-coded CRS values (line 617) and limits (line 622). Should this respect the choice of `default_crs` in `coord_sf()`? Also, I'm worried about just transforming a bounding box. This can give all sorts of unexpected results for weird coordinate systems. That's why I added different limits methods implemented in `calc_limits_bbox()`. Should that be used here, and the choice of the limit method respected (see argument `lims_method` in `coord_sf()`)? Finally, I added the function `sf_transform_xy()` to transform coordinates will imposing various checks. Is that useful here? (The answer to several of my questions may well be "no", just wanted to be sure I raised them so we've thought it all through carefully.)

thomasp85:
can this be included in the next release or are there unresolved questions?

teunbrand:
I think this a step in the right direction and shouldn't make anything worse, so I'd like to include this. However, these might be some typical quirks with regards to projection, like if you want to place ticks at longitude -180/180 it might be that the scale limits are calculated as -179.99/179.99 and the ticks aren't drawn.

thomasp85 (approved):
LGTM. As said, it is a step in the right direction even though there may be issues to iron out still

PR #5443, opened 2023-09-27
Scale call check

### LINKED ISSUE #5436: `scale_y_continuous()` fails if inside a function called via double colon `pkg::fun(...)`
I found a problem with recent PR #5343 when `scale_y_continuous` was called inside a package function via double colon

I expected the function not to throw an error

Please note that the problem is not with `{tern}` itself, as the call without double colon succeeds

```r
install.package("tern")
remotes::install_github("tidyverse/ggplot2")

library(tern)

# example from tern::g_lineplot function
library(nestcolor)

adsl <- tern_ex_adsl
adlb <- tern_ex_adlb %>% dplyr::filter(ANL01FL == "Y", PARAMCD == "ALT", AVISIT != "SCREENING")
adlb$AVISIT <- droplevels(adlb$AVISIT)
adlb <- dplyr::mutate(adlb, AVISIT = forcats::fct_reorder(AVISIT, AVISITN, min))
g_lineplot(adlb, adsl, subtitle = "Laboratory Test:")

# This call falls on if clause `... && !startsWith(as.character(call[[1]]), "scale_")` as it has length = 3
tern::g_lineplot(adlb, adsl, subtitle = "Laboratory Test:")
```

teunbrand:
This PR aims to fix #5436.

Briefly, instead of checking if the calling function *is* a `scale_`-prefixed function, it checks if any component of the calling function is a `scale_`-prefixed function. Added some test for other tricky call situations as well.

### DIFF
--- a/R/scale-continuous.R
+++ b/R/scale-continuous.R
@@ -85,7 +85,7 @@ scale_x_continuous <- function(name = waiver(), breaks = waiver(),
                                guide = waiver(), position = "bottom",
                                sec.axis = waiver()) {
   call <- caller_call()
-  if (is.null(call) || !startsWith(as.character(call[[1]]), "scale_")) {
+  if (is.null(call) || !any(startsWith(as.character(call[[1]]), "scale_"))) {
     call <- current_call()
   }
   sc <- continuous_scale(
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5446, opened 2023-10-01
Add a minor note about the difference of the unit of size to FAQ

yutannihilation:
I think it's a common case that the user wants to customize both the size of theme and that of geoms. So, probably this FAQ should have some mention about the difference of the unit.

https://ggplot2.tidyverse.org/articles/faq-customising.html#fonts

### DIFF
--- a/vignettes/articles/faq-customising.Rmd
+++ b/vignettes/articles/faq-customising.Rmd
@@ -505,7 +505,7 @@ GeomLabel$default_aes$size
 
 You can change the size using the `size` argument in `geom_text()` for a single plot. If you want to use the same updated size, you can set this with `update_geom_defaults()`, e.g. `update_geom_defaults("text", list(size = 6))`.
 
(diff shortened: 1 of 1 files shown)


teunbrand:
It might also be good to mention hat we've added a `size.unit` argument to `geom_text()` to switch the behaviour in #5260.

thomasp85:
@yutannihilation can you brush this up so we can include it in the next release

yutannihilation:
Thanks, but I'm yet to figure out how I can fit @teunbrand's comment into the FAQ; while `size.unit` is a variable tip, I think the FAQ talks about how to change the default size, and `size.unit` is not what the user can update by `update_geom_defaults()` (am I correct here?). Also, `size.unit` cannot change the fact that `GeomLabel$default_aes$size` is stored as mm.  Maybe it's less confusing not to mention `size.unit` here? Any ideas?

teunbrand:
> am I correct here?

Yeah it is a parameter, not an aesthetic. 

> Any ideas?

Maybe mention here that 'the default `size` is in **mm**' and add to the vignette it links to that there is a metric switch?

yutannihilation:
Thanks. How about this one?

teunbrand (approved):
Yes I think this looks good :)

yutannihilation:
Thanks!

PR #5449, opened 2023-10-02
Hotfix 3.4.4

teunbrand:
This PR contains the changes needed for #5437.
Mostly, it just replaces `is.atomic(x)` by `is.atomic(x) || is.null(x)`. The one exception is in `check_breaks_labels()` where NULL can be matched with an atomic vector without problems.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,7 @@
-# ggplot2 (development version)
+# ggplot2 3.4.4
+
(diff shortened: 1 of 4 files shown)


thomasp85 (approved):
Small thing but putting `is.null()` before the rest is preferable I think, both because that check is much easier, but also because it is the most basic value type...

Issue #1274, opened 2015-08-19
stat_summary_* API cleanup


hadley:
* [x] Rename `stat_bin2d()` and `stat_binhex()` to match `stat_summary_2d` and `stat_summary_hex`
* [x] Also need binning version of `stat_summary()`. Maybe `stat_summary_bin()` ?
* [x] `stat_summary_2d()` and `stat_summary_hex()` need `binwidth` argument

PR #5451, opened 2023-10-03
Removes usage of stat_bin2d

eliocamp:
`stat_bin2d()` was renamed to `stat_bin_2d()` in 3c19a96a70c4c2595640fa5d8f7b39c81b72905b for consistency with other functions (#1274). Traces of this functions are still in the documentation. This PR removes them. 

A small issue is that the documentation still reads:

`stat_bin2d() understands the following aesthetics (required aesthetics are in bold):`

I think this is an issue with the name of the stat being `StatBin2d`, which is then "snakeyfied" to `stat_bin2d`. The fix is probably not trivial and I don't know if it's worth it.

### DIFF
--- a/R/coord-cartesian-.R
+++ b/R/coord-cartesian-.R
@@ -49,7 +49,7 @@
 #'
 #' # You can see the same thing with this 2d histogram
 #' d <- ggplot(diamonds, aes(carat, price)) +
-#'   stat_bin2d(bins = 25, colour = "white")
+#'   stat_bin_2d(bins = 25, colour = "white")
 #' d
 #'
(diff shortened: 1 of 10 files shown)


teunbrand:
Thanks for the PR Elio!

> A small issue is that the documentation still reads:

I think this might be simply `"bin2d"` to `"bin_2d"` in the line below:

https://github.com/tidyverse/ggplot2/blob/a560c990d147845a5038ba720378b7000f4846c1/R/geom-bin2d.R#L8

eliocamp:
Oh, excellent. Latest commit changes this.

teunbrand (approved):
Looks great, thank you so much!

PR #5453, opened 2023-10-04
Cherrypick rgeos removal from description

teunbrand:
Because {rgeos} is going to be retired and we want to include the spatial retirements into the release.


thomasp85 (approved):
LGTM - is this really the only place it is referenced

teunbrand:
Yes, it was a 'ghost' suggested package, because no code actually used it.

Issue #5365, opened 2023-07-26
`geom_label` doesn't allow for varying line types


FDylanT:
I would like to be able to manipulate the line type of the box produced by `geom_label` or `annotate("label")` (e.g. from a solid to a dashed line). At present, `linetype` isn't an accepted parameter in this `geom`, and produces an "ignoring unknown parameters" warning.

teunbrand:
Is the goal here to have it as a mappable aesthetic or as a fixed parameter to pass to `geom_label()`?

aphalo:
Implementing this as a mapped aesthetic is dead simple. I can make a pull request if you like.

teunbrand:
We'd be happy to review a PR implementing this. The non-obvious bit that comes to mind is that `draw_key_label()` needs to reflect the linetype as well.

aphalo:
o.k., I will give it a try next week. Thanks for the hint!

PR #5454, opened 2023-10-04
More aesthetics for `geom_label()`

steveharoz:
# Updates to `geom_label()`'s aesthetics and parameters

1. The legend keys for `geom_label()` now have the same rounded rectangle shape and aesthetics as the labels.  (See questions below)

**Old:** image     **New:** image

2. Deprecated `label.size`. Use `linewidth` instead, which is more consistent with the rest of ggplot. ~~If `linewidth` is not set, it'll fall back to `label.size`, so nothing should break.~~ Update: setting up a soft deprecation fallback was too messy when going from an argument to an aesthetic, so I fully deprecated `label.size`.

3. New aesthetic: `linetype` for the label border. Default: "solid". Addresses #5365

4. New parameter: `border_colour` is the color of the label's border and is decoupled from the text color. If the value is `NULL` or not set, it will fall back to the text color, which is how it worked previously. Hopefully, that will avoid any breaking changes.

5. I added three examples to the docs. The mtcars graphs are too crowded to show off all the new aesthetics, so I made a dummy dataset for the latter two:

```r
geom_label(
  aes(fill = factor(cyl), linetype = qsec < 15),
  border_colour = "black", color = "white", linewidth = 1) +
scale_linetype_manual(values=c("solid", "blank"), limits = TRUE, labels = "1/4 mi < 15s", name = NULL)
```
image

```r
# Add aesthetics to the border for geom_label
data.frame(x = 1:10, y = 1:10) |>
ggplot() +
  geom_label(aes(
    label=month.abb[x],
    x=x,
    y=y,
    color = factor(x%%3),
    linewidth = x%%2,
    linetype = factor(x%%3)),
    fill = NA) +
scale_linewidth(range=c(0.5, 1.5)) +
scale_linetype_manual(values=c("solid", "blank", "dotted"))
```
image

```r

# Override the border color
data.frame(x = 1:10, y = 1:10) |>
ggplot() +
  geom_label(aes(
    label=month.abb[x],
    x=x,
    y=y,
    color = factor(x%%3),
    linewidth=x%%2,
    linetype=factor(x%%3)),
    border_color = "red",
    fill=NA) +
scale_linewidth(range=c(0.5, 1.5)) +
scale_linetype_manual(values=c("solid", "blank", "dotted"))
```
image

# Questions:
1. ~~Does the deprecated messaging look right for label.size?~~
1. There is a bug where the key_glyph is cropped at the bottom. Can anyone help with that?
1. ~~Are the aliases border_colour/border_color dealt with appropriately?~~

### DIFF
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -2,7 +2,7 @@
 #' @rdname geom_text
 #' @param label.padding Amount of padding around label. Defaults to 0.25 lines.
 #' @param label.r Radius of rounded corners. Defaults to 0.15 lines.
-#' @param label.size `r lifecycle::badge("deprecated")` Please use `linewidth` to set the size of the border.
+#' @param label.size `r lifecycle::badge("deprecated")` Please use `linewidth` to set the width of the border.
 #' @param border.colour Colour of the label's border. If `NULL`, it will fall back to the text colour.
 #' @param border.color An alias for `border.colour`.
 geom_label <- function(mapping = NULL, data = NULL,
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -4,6 +4,7 @@
 #' @param label.r Radius of rounded corners. Defaults to 0.15 lines.
 #' @param label.size `r lifecycle::badge("deprecated")` Please use `linewidth` to set the size of the border.
 #' @param border_colour Colour of the label's border. If `NULL`, it will fall back to the text colour.
+#' @param border_color An alias for `border_colour`.
 geom_label <- function(mapping = NULL, data = NULL,
                        stat = "identity", position = "identity",
                        ...,
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -3,6 +3,7 @@
 #' @param label.padding Amount of padding around label. Defaults to 0.25 lines.
 #' @param label.r Radius of rounded corners. Defaults to 0.15 lines.
 #' @param label.size `r lifecycle::badge("deprecated")` Please use `linewidth`. Size of label border, in mm.
+#' @param border_colour Colour of the label's border. If `NULL`, it will fall back to the text colour.
 geom_label <- function(mapping = NULL, data = NULL,
                        stat = "identity", position = "identity",
                        ...,
@@ -13,9 +14,11 @@ geom_label <- function(mapping = NULL, data = NULL,
                        label.r = unit(0.15, "lines"),
                        label.size = 0.25,
                        size.unit = "mm",
+                       border_colour = NULL,
                        na.rm = FALSE,
                        show.legend = NA,
-                       inherit.aes = TRUE) {
+                       inherit.aes = TRUE,
+                       border_color = border_colour) {
   if (!missing(nudge_x) || !missing(nudge_y)) {
     if (!missing(position)) {
       cli::cli_abort(c(
@@ -45,6 +48,7 @@ geom_label <- function(mapping = NULL, data = NULL,
       label.r = label.r,
       label.size = label.size,
       size.unit = size.unit,
+      border_colour = border_colour %||% border_color,
       na.rm = na.rm,
       ...
     )
@@ -71,8 +75,7 @@ GeomLabel <- ggproto("GeomLabel", Geom,
                         label.r = unit(0.15, "lines"),
                         label.size = 0.25,
                         size.unit = "mm",
-                        border_colour = NULL,
-                        border_color = border_colour) {
+                        border_colour = NULL) {
     lab <- data$label
     if (parse) {
       lab <- parse_safe(as.character(lab))
@@ -108,7 +111,7 @@ GeomLabel <- ggproto("GeomLabel", Geom,
           lineheight = row$lineheight
         ),
         rect.gp = gpar(
-          col = ifelse(row$linewidth == 0, NA, border_colour %||% border_color %||% row$colour),
+          col = ifelse(row$linewidth == 0, NA, border_colour %||% row$colour),
           fill = alpha(row$fill, row$alpha),
           lty = row$linetype,
           lwd = (row$linewidth %||% label.size) * .pt
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -62,7 +62,7 @@ GeomLabel <- ggproto("GeomLabel", Geom,
   default_aes = aes(
     colour = "black", fill = "white", size = 3.88, angle = 0,
     hjust = 0.5, vjust = 0.5, alpha = NA, family = "", fontface = 1,
-    lineheight = 1.2, linetype = "solid", linewidth = 0.25
+    lineheight = 1.2, linetype = 1, linewidth = 0.25
   ),
 
   draw_panel = function(self, data, panel_params, coord, parse = FALSE,
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -27,8 +27,8 @@ geom_label <- function(mapping = NULL, data = NULL,
     position <- position_nudge(nudge_x, nudge_y)
   }
 
-  if (!missing(label.size)) {
-    message("`label.size` is deprecated. Please use `linewidth` in the future.")
+  if (lifecycle::is_present(label.size)) {
+    deprecate_warn0("3.5.0", "geom_label(label.size)", "geom_label(linewidth)")
   }
 
   layer(
--- a/R/geom-label.R
+++ b/R/geom-label.R
@@ -3,8 +3,8 @@
 #' @param label.padding Amount of padding around label. Defaults to 0.25 lines.
 #' @param label.r Radius of rounded corners. Defaults to 0.15 lines.
 #' @param label.size `r lifecycle::badge("deprecated")` Please use `linewidth` to set the size of the border.
-#' @param border_colour Colour of the label's border. If `NULL`, it will fall back to the text colour.
-#' @param border_color An alias for `border_colour`.
+#' @param border.colour Colour of the label's border. If `NULL`, it will fall back to the text colour.
+#' @param border.color An alias for `border.colour`.
 geom_label <- function(mapping = NULL, data = NULL,
                        stat = "identity", position = "identity",
                        ...,
@@ -15,11 +15,11 @@ geom_label <- function(mapping = NULL, data = NULL,
                        label.r = unit(0.15, "lines"),
                        label.size = deprecated(),
                        size.unit = "mm",
-                       border_colour = NULL,
+                       border.colour = NULL,
                        na.rm = FALSE,
                        show.legend = NA,
                        inherit.aes = TRUE,
-                       border_color = border_colour) {
+                       border.color = border.colour) {
   if (!missing(nudge_x) || !missing(nudge_y)) {
     if (!missing(position)) {
       cli::cli_abort(c(
@@ -48,7 +48,7 @@ geom_label <- function(mapping = NULL, data = NULL,
       label.padding = label.padding,
       label.r = label.r,
       size.unit = size.unit,
-      border_colour = border_colour %||% border_color,
+      border.colour = border.colour %||% border.color,
       na.rm = na.rm,
       ...
     )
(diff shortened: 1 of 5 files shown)


R/geom-label.R
```diff
@@ -27,6 +27,10 @@ geom_label <- function(mapping = NULL, data = NULL,
     position <- position_nudge(nudge_x, nudge_y)
   }
 
+  if (!missing(label.size)) {
+    message("`label.size` is deprecated. Please use `linewidth` in the future.")
+  }
```
teunbrand, line 32: ggplot2 uses the {lifecycle} package to communicate deprecation messages. We'd have to set `label.size = deprecated()` in the formals to have it show up in the usage section of the docs and then use the messaging below:

```suggestion
  if (lifecycle::is_present(label.size)) {
    deprecate_warn0("3.5.0", "geom_label(label.size)", "geom_label(linewidth)")
  }
```

We could also use soft-deprecation with `deprecate_soft0()` here, but then we'd need a fallback mechanism to feed `label.size` into a `linewidth` parameter.

R/geom-label.R
```diff
@@ -58,15 +62,17 @@ GeomLabel <- ggproto("GeomLabel", Geom,
   default_aes = aes(
     colour = "black", fill = "white", size = 3.88, angle = 0,
     hjust = 0.5, vjust = 0.5, alpha = NA, family = "", fontface = 1,
-    lineheight = 1.2
+    lineheight = 1.2, linetype = "solid", linewidth = 0.25
```
teunbrand, line 65:
```suggestion
    lineheight = 1.2, linetype = 1, linewidth = 0.25
```

R/geom-label.R
```diff
@@ -58,15 +62,17 @@ GeomLabel <- ggproto("GeomLabel", Geom,
...
 
   draw_panel = function(self, data, panel_params, coord, parse = FALSE,
                         na.rm = FALSE,
                         label.padding = unit(0.25, "lines"),
                         label.r = unit(0.15, "lines"),
                         label.size = 0.25,
-                        size.unit = "mm") {
+                        size.unit = "mm",
+                        border_colour = NULL,
+                        border_color = border_colour) {
```
teunbrand, line 75: Would it make sense to add these to the formals of `geom_label()` as well? It might also be best to standardise to UK spelling immediately in `geom_label()`. In addition, I think ggplot2 has settled at using dot.case instead of snake_case for most arguments.

R/geom-label.R
```diff
@@ -58,15 +62,17 @@ GeomLabel <- ggproto("GeomLabel", Geom,
...
     hjust = 0.5, vjust = 0.5, alpha = NA, family = "", fontface = 1,
-    lineheight = 1.2
+    lineheight = 1.2, linetype = "solid", linewidth = 0.25
   ),
 
   draw_panel = function(self, data, panel_params, coord, parse = FALSE,
                         na.rm = FALSE,
                         label.padding = unit(0.25, "lines"),
                         label.r = unit(0.15, "lines"),
                         label.size = 0.25,
```
teunbrand, line 72: If `label.size` becomes deprecated, it might be best to leave them out of the belly of `GeomLabel` as well.

R/geom-label.R
```diff
@@ -102,9 +108,10 @@ GeomLabel <- ggproto("GeomLabel", Geom,
           lineheight = row$lineheight
         ),
         rect.gp = gpar(
-          col = if (isTRUE(all.equal(label.size, 0))) NA else row$colour,
+          col = ifelse(row$linewidth == 0, NA, border_colour %||% border_color %||% row$colour),
           fill = alpha(row$fill, row$alpha),
-          lwd = label.size * .pt
+          lty = row$linetype,
+          lwd = (row$linewidth %||% label.size) * .pt
```
teunbrand, line 114: Might also omit `label.size` if deprecated.

R/legend-draw.R
```diff
@@ -252,9 +252,31 @@ draw_key_text <- function(data, params, size) {
...
-  grobTree(
-    draw_key_rect(data, list()),
-    draw_key_text(data, list())
+  if(is.null(data$label))         data$label <- "a"
+  if(is.null(data$label.r))       data$label.r <- unit(0.1, "snpc")
+  if(is.null(params$label.padding)) params$label.padding <- unit(0.25, "lines")
+  if(length(params$label.padding) == 1) params$label.padding[2:4] <- params$label.padding
+
+  labelGrob(
+    label = data$label,
```
teunbrand, line 261: Might skip all the `if (is.null(foo$bar)) foo$bar <- "baz"` above if you use `bar = foo$bar %||% "baz"` in setting the arguments of `labelGrob()`.

teunbrand (changes requested):
Thanks for the PR, Steve!

I like the intent of the PR and the comments I've made are mostly about polishing and conventions. I don't think ggplot2 has settled on how paired aesthetics (`colour`/`border_colour`) should be implemented yet, but I can bring it up with Thomas soon.

R/geom-label.R
steveharoz: Thanks. I couldn't figure out how deprecated messages worked, so I'll trust you on how best to go about it.

R/legend-draw.R
steveharoz: Yeah, makes sense. Done in an upcoming commit.

R/geom-label.R
steveharoz: added in [3002125](https://github.com/tidyverse/ggplot2/pull/5454/commits/30021259dc7e61ede0494d237e91523e8cf78f0d)

As for dot.case vs. snake_case, it's not consistent, so I'm not sure. `nudge_x` and `nudge_y` are arguments that use snake_case in that same function. And the function name itself is `geom_label`. I'll lean towards underscore unless it's clearly settled. If so, happy to change it :)

R/geom-label.R
steveharoz: Update: I've fully deprecated `label.size`. It now only exists in `geom_label()` to give a deprecation warning.

steveharoz:
I've updated the comment at the beginning. See the two remaining questions.

teunbrand:
> After this PR, some work will be required in ggrepel to make geom_label_repel() understand that?

The ggrepel package has always been free to do whatever it wants with labels, so there'd be no need to wait on this PR for that.

> steveharoz requested a review from teunbrand 20 hours ago

I have to probe Thomas' wisdom on some points, I'll get back on this on tuesday/wednesday.

R/geom-label.R
thomasp85: argument names _should_ use dot.case...

thomasp85: Also, if you can think up a name that doesn't have variation between UK/US spelling it would be preferable because we really have gripes with the decision to support both. While we can't backtrack on that decision we can attempt to limit new entrants

steveharoz: 1. I updated to dot.case. 

2. Yeah, I see how the redundant variables are a headache. I couldn't come up with a good alternative to term for colour/color, so I'm copying what the rest of ggplot does. Open to alternatives!

R/geom-text.R
```diff
@@ -89,6 +89,18 @@
...
+#' p + geom_label(aes(color = factor(cyl)))
+#'
+#' # Alternatively, border.color can have a static value
+#' p + geom_label(aes(color = factor(cyl)), border.color = "black")
+#'
+#' # Use linetype and linewidth aesthetics to add highlights
+#' p + geom_label(
+#'   aes(fill = factor(cyl), linetype = qsec < 15),
+#'   border.colour = "black", color = "white", linewidth = 1) +
+#' scale_linetype_manual(values=c("solid", "blank"), limits = TRUE, labels = "1/4 mi < 15s", name = NULL)
```
teunbrand, line 102: Some indentation and simplification because having clear examples is more important than having pretty examples :)
```suggestion
#' p + 
#'  geom_label(
#'    aes(fill = factor(cyl), linetype = qsec < 15),
#'    border.colour = "black", colour = "white"
#'  ) +
#'  scale_linetype_manual(values = c("solid", "blank"), limits = TRUE)
```

teunbrand (commented):
Hi Steve,

I had a chat with Thomas and we agree that PR's direction is great.
I've left some minor formatting comments.
About the legend clipping, I think legends should be slightly smarter about determining the size of the keys, so I'll try to fix that in a seperate PR first before we can merge this one.

R/geom-text.R
steveharoz: Yeah, agreed!

teunbrand:
If #5465 gets merged, we can fix the key glyph cropping issue.

teunbrand:
Yeah that is exactly what Thomas and I were discussing, but we both thought making `border.colour` a proper aesthetic opens the floodgates for a new level of complexity that is difficult to understand and teach people. Here, as @bwiernik points out in the thread you linked, one could override a scale's `aesthetics` argument and that will work out fine. However, in #5423 there would be `whisker_linetype` aesthetics and linetype scales don't have an `aesthetics` argument. In fact, I'm planning to roll back the proper aesthetics in #5423 to fixed parameters because of this.

bwiernik:
Personally, I find fixed parameters extremely frustrating because very quickly collaborators or audiences for plots that use them start to ask for data-dependent customization of them (eg, making the linetype for the whiskers match the linetype for the box or vary depending on whether they cross a threshold).

It would seem that a better solution in the case of #5423 would be to add the `aesthetics` argument to linetype scales

smouksassi:
I agree that customization is key and having a coherent look across geoms would make plots much more readable i.e. linetype of a grouped boxplot follow the same logic as other mapped variable.  but yeah this exponentiates possibilities and complexities!
(I still liked when I could have the point linetype follow some aesthetics mapping in splus but alas in grid stroke of a point cannot have a different linetype...)

teunbrand:
Off topic:
@smouksassi You still can, but it is device dependent. With {ragg}:
``` r
library(grid)
grid.points(pch = 21, size = unit(2, "char"), gp = gpar(lty = 3), default.units = "npc")
```



<sup>Created on 2023-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
Here is my proposal for the label key drawing function:

```r
draw_key_label <- function(data, params, size) {
  if(is.null(params$label.padding)) params$label.padding <- unit(0.25, "lines")
  if(length(params$label.padding) == 1) params$label.padding[2:4] <- params$label.padding
  
  grob <- labelGrob(
    label = data$label %||% "a",
    x = 0.5,
    y = 0.5,
    padding = params$label.padding,
    r = data$label.r %||% unit(0.1, "snpc"),
    angle = data$angle %||% 0,
    text.gp = gpar(
      col = alpha(data$colour %||% "white", data$alpha),
      fontfamily = data$family %||% "",
      fontface = data$fontface %||% 1,
      fontsize = (data$size %||% 3.88) * .pt
    ),
    rect.gp = gpar(
      col = alpha(params$border.colour %||% params$border.color %||% data$colour %||% "black", data$alpha),
      fill = alpha(data$fill %||% "white", data$alpha),
      lty = data$linetype,
      lwd = len0_null(data$linewidth * .pt)
    ),
    vp = NULL
  )
  # Linewidth as cm, twice because it is on two sides of textbox
  lwd <- (data$linewidth %||% params$label.size) * .pt * 0.2
  # Get width/height of text. Margins are baked in.
  width  <- 
    convertWidth(widthDetails(grob$children[[2]]),  "cm", valueOnly = TRUE)
  height <-
    convertHeight(heightDetails(grob$children[[2]]), "cm", valueOnly = TRUE)
  attr(grob, "width")  <- width  + lwd
  attr(grob, "height") <- height + lwd
  grob
}
```

olivroy:
Is there any chance this will be merged for this release? or will this get looked at after release?

teunbrand:
@olivroy This PR is currently set as a draft (so we cannot merge it) and the release process has been kicked off.
While ideally this PR would have been flagged in the 3.5.0 milestone, it is currently not *too* late, as we haven't started the reverse dependency check yet (see https://github.com/tidyverse/ggplot2/issues/5588). However, we will start that soon, so I would expect it will not be included in 3.5.0.

steveharoz:
Been busy with other things. But I'll try to work on it this weekend.

teunbrand:
There is no rush, I'm reasonably sure that the upcoming release would have to be patched afterwards so we can include it in that release. If someone needs these features *right this instant*, they can extend `geom_label()` if they want.

teunbrand:
Hi @steveharoz, are you still interested in finishing this PR? If you're busy with other things that is fine too, and we could take over from here.

teunbrand:
Closing this in favour of #6307. Thank you @steveharoz for pioneering the approach!

Issue #3587, opened 2019-10-27
Remove space for legend title if it doesn't have a title


steveharoz:
If a guide has a label, you can make it flush with the edge of a plot:

```r
ggplot(PlantGrowth) + 
  aes(x=group, y=weight, fill=group) +
  geom_boxplot() + 
  theme(
    legend.position="top", 
    legend.justification = "left", 
    legend.margin = margin(0, 0, 0, 0))
```

![image](https://user-images.githubusercontent.com/2257540/67627750-35ad5900-f817-11e9-9a46-5d73e8f97ccb.png)

But if you set the label to NULL, a there is a gap or misalignment that doesn't seem to be removable:

```r
ggplot(PlantGrowth) + 
  aes(x=group, y=weight, fill=group) +
  geom_boxplot() + 
  labs(fill = NULL) + # removes legend title
  theme(
    legend.position="top", 
    legend.justification = "left", 
    legend.margin = margin(0, 0, 0, 0))
```

![image](https://user-images.githubusercontent.com/2257540/67627807-30044300-f818-11e9-8360-72e4b7e774b4.png)

clauswilke:
It looks like a unit of `legend.spacing.x` is inserted even if the name is missing. 
``` r
library(ggplot2)

ggplot(PlantGrowth) + 
  aes(x=group, y=weight, fill=group) +
  geom_boxplot() + 
  labs(fill = NULL) + # removes legend title
  theme(
    legend.position="top", 
    legend.justification = "left", 
    legend.margin = margin(0, 0, 0, 0),
    legend.spacing.x = unit(0, "pt")
  )
```

![](https://i.imgur.com/xGlS9Du.png)

You can work around this issue by placing some spaces before and after the labels, so that setting `legend.spacing.x` to zero doesn't cause problems.

``` r

ggplot(PlantGrowth) + 
  aes(x=group, y=weight, fill=group) +
  geom_boxplot() + 
  scale_fill_hue(
    name = NULL,
    labels = function(x) paste0(" ", x, "   ")
  ) +
  theme(
    legend.position="top", 
    legend.justification = "left", 
    legend.margin = margin(0, 0, 0, 0),
    legend.spacing.x = unit(0, "pt")
  )
```

![](https://i.imgur.com/Inc8aQd.png)

The legend drawing code is a bit of a mess, so I don't see an immediate approach to fixing this more generally.

steveharoz:
Thanks for the workaround!

[...]

PR #5456, opened 2023-10-05
Use text margin to space apart key-labels and guide-title

### LINKED ISSUE #5455: Bug: rotated text in `guide_colourbar()`
In current main branch, when using rotated legend text with `vjust = 1`, the text starts overlapping the colourbar itself.
In the example below, I expected the left-most 'C' to start at the right of the colourbar.

``` r
#> ℹ Loading ggplot2

  scale_colour_continuous(
    breaks = c(10, 20, 30),
    labels = c("A", "B\nB", "C\nC\nC")
  ) +
  theme(
    legend.text = element_text(angle = 270, vjust = 1)
  )
```

<sup>Created on 2023-10-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR is a first step at refactoring #5283 to be less aggressive in terms of visual changes.
Inadvertently, it also fixes #5455.

image

Briefly, this PR detaches the spacing from keys to labels and from titles to guides from the `legend.spacing.{x/y}` setting. Instead, the legend text/title's `margin` argument controls this particular spacing. It technically introduces a visual change, but keeps the defaults close to what they are.

Some details; this idea was coined by Claus in https://github.com/tidyverse/ggplot2/issues/3587#issuecomment-546700387. The gist of this PR is that it looks if user declared an explicit margin in the theme or guide, and if they didn't, it sets a default margin that is visually equivalent to current defaults. Since it now omits `legend.spacing.{x/y}` between key and label, this means `legend.spacing.{x/y}` is solely dedicated to the spacing between key-label pairs. In contrast to #5283, this wouldn't need to introduce a new text spacing concept, as it is captured in the margins.

There are two visual changes in the tests:
* One related to fixing #5455.
* One related to guide-to-guide spacing, which I don't quite understand at the moment.

### DIFF
--- a/R/guide-legend.R
+++ b/R/guide-legend.R
@@ -532,8 +532,8 @@ GuideLegend <- ggproto(
     hgap <- elements$hgap %||% 0
     widths <- switch(
       params$label.position,
-      "left"   = list(label_widths, hgap, widths, hgap),
-      "right"  = list(widths, hgap, label_widths, hgap),
+      "left"   = list(label_widths, widths, hgap),
+      "right"  = list(widths, label_widths, hgap),
       list(pmax(label_widths, widths), hgap * (!byrow))
     )
     widths  <- head(vec_interleave(!!!widths),  -1)
@@ -541,8 +541,8 @@ GuideLegend <- ggproto(
     vgap <- elements$vgap %||% 0
     heights <- switch(
       params$label.position,
-      "top"    = list(label_heights, vgap, heights, vgap),
-      "bottom" = list(heights, vgap, label_heights, vgap),
+      "top"    = list(label_heights, heights, vgap),
+      "bottom" = list(heights, label_heights, vgap),
       list(pmax(label_heights, heights), vgap * (byrow))
     )
     heights <- head(vec_interleave(!!!heights), -1)
@@ -554,14 +554,14 @@ GuideLegend <- ggproto(
     # Combine title with rest of the sizes based on its position
     widths <- switch(
       params$title.position,
-      "left"  = c(title_width, hgap, widths),
-      "right" = c(widths, hgap, title_width),
+      "left"  = c(title_width, widths),
+      "right" = c(widths, title_width),
       c(widths, max(0, title_width - sum(widths)))
     )
     heights <- switch(
       params$title.position,
-      "top"    = c(title_height, vgap, heights),
-      "bottom" = c(heights, vgap, title_height),
+      "top"    = c(title_height, heights),
+      "bottom" = c(heights, title_height),
       c(heights, max(0, title_height - sum(heights)))
     )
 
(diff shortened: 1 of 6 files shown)


clauswilke:
I think this is a great idea! If you can do this without major regressions that's even better.

teunbrand:
> I think this is a great idea! 

Well it was originally yours to begin with!
If you have some gnarly situations in mind, I'd be happy to explore if this works. Most visual tests are fine though, but they don't deviate from the defaults too much.

clauswilke:
I don't have any specific tests in mind, I'm just thinking about the many ways I've worked around the previous behavior and those workarounds may now produce weird results. But I don't have any specific example at hand. It's mostly stuff I did when I wrote my book, which is now already almost five years ago.

thomasp85:
I like this route, but we need to figure out the cause for the guide spacing regression and find a solution

tests/testthat/test-guides.R
```diff
@@ -601,10 +601,10 @@ test_that("guides title and text are positioned correctly", {
       scale_fill_continuous(name = "the\ncontinuous\ncolorscale")
   )
   expect_doppelganger("vertical gap of 1cm between guide title and guide",
-    p + theme(legend.spacing.y = grid::unit(1, "cm"))
+    p + theme(legend.title = element_text(margin = margin(b = 1, unit = "cm")))
```
teunbrand, line 604: I figured out why the spacing between guides has changed. Here, the legend spacing is default and no longer 1 cm. So, it is actually doing what it is supposed to do and this isn't a regression.

teunbrand:
The guide-to-guide spacing thing prompts the question though whether there should a separate theme element to control key-to-key spacing versus guide-to-guide spacing.

clauswilke:
> The guide-to-guide spacing thing prompts the question though whether there should a separate theme element to control key-to-key spacing versus guide-to-guide spacing.

Probably. I've always felt the customization options for legends were insufficient and the same values were inappropriately being reused over and over.

teunbrand:
I'm leaning towards adding `key.spacing{.x/.y}` arguments in `guide_legend()` to control the spacing  between keys. That way, `legend.spacing{.x/.y}` can be purely for spacing between guides.  The other guides are implemented as single keys, so that setting would be irrelevant for them.

teunbrand:
I considered implementing the `key.spacing` in another PR, but I felt it was so closely related, it might best fit in this PR.

So now we can set the spacing between keys using the `key.spacing` arguments in `guide_legend()`. No other guides use this spacing, so it wouldn't make too much sense to make this a `theme()` setting. Default behaviour of the `key.spacing` matches current behaviour, with the exception that `byrow` has no influence on the spacing (as it shouldn't).

``` r
#> ℹ Loading ggplot2

  guides(colour = guide_legend(
    ncol = 2, key.spacing.y = 1, key.spacing.x = 2
  ))
```



<sup>Created on 2023-11-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

An additional benefit is that `legend.spacing` is now 'pure' in that it *only* controls the spacing between legends and not within legends.

This figure might be a better explainer of what this PR does:

image

teunbrand:
Now also fixes the bin spacing issue:

``` r
#> ℹ Loading ggplot2

  geom_point(aes(size = cty)) +
  scale_size_continuous(
    breaks = c(10, 20, 30),
    labels = c("A", "B\nB", "C\nC\nC"),
    guide  = "bins"
  ) +
  theme(
    legend.text = element_text(angle = 270, vjust = 1)
  )
```



<sup>Created on 2023-11-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/guide-legend.R
```diff
@@ -436,13 +455,32 @@ GuideLegend <- ggproto(
...
+        !inherits(elements$text, "element_blank")) {
+      i <- match(params$label.position, .trbl[c(3, 4, 1, 2)])
+      elements$text$margin[i] <- elements$text$margin[i] + gap
+    }
+    if (is.null(params$title.theme$margin %||% theme$legend.title$margin) &&
+        !inherits(elements$title, "element_blank")) {
+      i <- match(params$title.position, .trbl[c(3, 4, 1, 2)])
+      elements$title$margin[i] <- elements$title$margin[i] + gap
+    }
+
```
thomasp85, line 483: Can you add a comment on what is going on here? Just for future us

thomasp85 (approved):
Apart from the small comment LGTM

R/guide-legend.R
teunbrand: Sure thing

PR #5465, opened 2023-10-10
Custom key glyph sizes

teunbrand:
This PR aims to fix a roadblock in #5454 by providing a mechanism by which keys can set custom sizes.

The problem is that it is heuristically okay, but sometimes inconvenient, for the legend key size to be determined by only the `linewidth` and `size` column in the data. For example, consider the following case when displaying arrows, where arrowhead size exceeds the allotted size on the legend:

``` r
#> ℹ Loading ggplot2

df <- data.frame(x = 0, xend = 1, y = 1:3, group = LETTERS[1:3])

p <- ggplot(df, aes(x, y, xend = xend, yend = y, colour = group))
p + geom_segment(arrow = arrow(length = unit(1, "cm")))
```

With this PR, key drawing functions can set their size by attaching a `"width"` and/or `"height"` attribute to the produced grob (in cm).
For example, for arrows we might sketch the following key drawing function, which incorporates the dimensions of the arrowhead in these attributes. (Note that this isn't quite *perfect*, since the glyphs go from 0.1npc to 0.9npc, but it illustrates the point)

``` r
draw_key_arrow <- function(data, params, size) {
  grob <- draw_key_path(data, params, size)
  if (!is.null(params$arrow)) {
    angle  <- params$arrow$angle / 180 * pi
    length <- grid::convertUnit(params$arrow$length, unitTo = "cm", valueOnly = TRUE)
    attr(grob, "width")  <- (cos(angle) * length) + 0.1
    attr(grob, "height") <- (sin(angle) * length * 2) + 0.1
  }
  grob
}
```

Now we can have less funky legend keys for the arrows:

``` r
p + geom_segment(arrow = arrow(length = unit(1, "cm"), angle = 15),
                 key_glyph = draw_key_arrow)
```

``` r
p + geom_segment(arrow = arrow(length = unit(1, "cm"), angle = 60),
                 key_glyph = draw_key_arrow)
```

<sup>Created on 2023-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
Maybe instead of this being encoded in attributes, it would be better to use the `viewport` argument?

teunbrand:
> Maybe instead of this being encoded in attributes, it would be better to use the viewport argument

I experimented with this, but it is more of a hassle than as a solution. The main issue I'm encountering is that when setting a viewport on a key that doesn't have one based on the `size` or `linewidth`, it applies in both x- and y-directions,

thomasp85 (approved):
LGTM

PR #5473, opened 2023-10-11
Stacked axes

teunbrand:
This PR is a proof-of-concept for an axis guide that stacks other axis guides.

For context: I don't really think stacked axes are all too useful for data visualisation except maybe to display two different units of the same measurement on an axis (degree celsius/fahrenheit, metres/inches etc.). Certainly in its current form, this PR cannot even do that (but that might be amended by #5410). Arguably, that functionality shouldn't live in ggplot2 but in extension packages.

However, the display of multiple units is *not* the goal of this axis guide. The goal of this axis guide is to incorporate flexibility in axis composition into ggplot2 itself. That way, the following should be possible:

* A package can develop an axis annotation guide. With stacked axes, this annotation guide needn't worry about rendering the typical line, ticks and text and can just focus on providing the annotations.
* Package A can develop a 'typical' axis and package B can develop an 'annotation' axis and they can just be composed by users as they see fit.

To give a more concrete example for this, one could view the guide displayed [here](https://github.com/teunbrand/gguidance#ranges) as a regular axis + annotation for ranges.

Here is some nightmare fuel to show that it works:

``` r
#> ℹ Loading ggplot2

axis1 <- guide_axis(angle = 45)
axis2 <- guide_axis(cap = "both")
axis3 <- guide_axis(n.dodge = 2)
stack <- guide_axis_stack(axis1, axis2, axis3)

  geom_point() +
  guides(
    x = stack, x.sec = stack,
    y = stack, y.sec = stack
  ) +
```

<sup>Created on 2023-10-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/guide-axis-stack.R
```diff
@@ -0,0 +1,191 @@
...
+    remove  <- which(!available)
+    removed <- vapply(axes[remove], snake_class, character(1))
+    axes[remove] <- NULL
+    cli::cli_warn(c(paste0(
+      "{.fn guide_axis_stack} cannot use the following guide{?s}: ",
+      "{.and {.fn {removed}}}."
+    ), i = "Guides need to handle {.field x} and {.field y} aesthetics."))
+  }
+
+  params <- lapply(axes, `[[`, name = "params")
```
teunbrand, line 62: Keep track of #5474

teunbrand:
Check compatible with `coord_radial()`

teunbrand:
Now also compatible with `coord_radial()` and `guide_axis_theta()`.

``` r
#> ℹ Loading ggplot2

p <- ggplot(mtcars, aes(hp, disp)) +
  geom_point() +
  coord_radial(start = 0.25 * pi, end = 1.75 * pi, donut = 0.5) +
  guides(
    theta = guide_axis_stack("axis_theta", guide_axis_theta(cap = "both")),
    theta.sec = guide_axis_stack("axis_theta", guide_axis_theta(cap = "both")),
    r = guide_axis_stack("axis", guide_axis(cap = "both")),
    r.sec = guide_axis_stack("axis", guide_axis(cap = "both"))
  ) +
p
```



Where the stacked ring appears depends on the size of the labels. It may go outside the bounding box of the panel and start to overlap with axis titles. I don't think there is a good solution for this, as the dimensions of the ring depend both on the panel size in relative units and the size of the axis in absolute units. However, people can always adjust margins manually if they insist.

``` r
p + scale_x_continuous(
  breaks = c(100, 200, 300), 
  labels = rep("long label", 3)
)
```



<sup>Created on 2023-11-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Also some weirdness may occur if the donut hole is too small to fit all `theta.sec` axes, but this should be very niche and I don't expect people to run into this on a regular basis.

``` r
#> ℹ Loading ggplot2

p <- ggplot(mtcars, aes(hp, disp)) +
  geom_point() +
  coord_radial(start = 0.25 * pi, end = 1.75 * pi, donut = 0.1) +
  guides(
    theta.sec = guide_axis_stack("axis_theta", "axis_theta", "axis_theta")
  ) +
p
```



<sup>Created on 2023-11-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
Should we add a way to individually control the theme of every axis in the stack?

thomasp85 (approved):
It's a bit annoying that the radial axis needs to have so much extra logic for the off-chance that it is stacked, but I see no way around it. And at least it is not mingled in with all the other code

PR #5475, opened 2023-10-11
Replace `sf_rescale01_x()` with `scales::rescale()`

teunbrand:
This PR fixes a personal pet peeve of mine.

First, let me start by apologising for the tone of this PR.
The two functions in the title are in principle doing the same operation, but `scales::rescale()` supports more data types than just numeric.
Every time I look at `coord_sf()`'s code, I have to remind myself that they're really doing the same thing despite being named differently. If this PR gets merged, this saves me 5 seconds of my life every time I'm dealing with `coord_sf()`.

<details>

Ulterior motive is also that I might want to feed not-bare-numeric data to `coord_sf()` at some point with a custom `rescale` method.

Typical coords use the `Scale$rescaler()` method. However, as the `geometry` column cannot be transformed by `rescale()`, we should not use the `Scale$rescaler()` method. If we did, annotations such as graticules and axis positions can become out of sync when a user has somehow engineered a non-standard `rescaler` in a position scale.

</details>


thomasp85:
Can you see any indication of the need for the old function?

teunbrand:
I don't think `sf_rescale01_x()` is needed. It was introduced in 379df62a65181bb14f426873abd436907c04ae71, I think as a pointwise equivalent to `sf_rescale01()`, which in turn is just `scales::rescale()`.

thomasp85 (approved):
LGTM

PR #5476, opened 2023-10-11
Simplify `check_device()` for {ragg}/{svglite}

### LINKED ISSUE #5470: Simplify `check_device()`
Both {ragg} and {svglite} now support `dev.capabilities()`:

* [svglite news](https://github.com/r-lib/svglite/blob/23506f71cc3181b6fb3882bbff8bff59d1ca008b/NEWS.md?plain=1#L4)
* [ragg news](https://github.com/r-lib/ragg/blob/9cb92fd2d7b8621878cc8b97a9507491caf03b8d/NEWS.md?plain=1#L6)

Which means we wouldn't need to special-case these devices in `check_device()`.
We should simplify the manual checks for these devices and instead just suggest to upgrade the packages.

teunbrand:
This PR amends #5350 and fixes #5470.

Briefly, the latest versions of {ragg}/{svglite} now support the `dev.capabilities()` method of querying their capabilities.
Therefore, we don't need to manually encode into the check what we know about the packages.
Instead, the check for {ragg}/{svglite} asks the user to install these versions and then resumes as normal.

Showing that it works:

``` r
#> ℹ Loading ggplot2
packageVersion("ragg")
#> [1] '1.2.6'

ragg::agg_png()

check_device("patterns")
#> [1] TRUE
check_device("compositing")
#> Warning: The agg_png device does not support compositing.
#> [1] FALSE

invisible(dev.off())
packageVersion("svglite")
#> [1] '2.1.2'

svglite::svglite()

check_device("patterns")
#> [1] TRUE
check_device("compositing")
#> Warning: The devSVG device does not support compositing.
#> [1] FALSE
```

<sup>Created on 2023-10-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/utilities-checks.R
```diff
@@ -247,6 +247,20 @@ check_device = function(feature, action = "warn", op = NULL, maybe = FALSE,
     dev_name <- names(dev_cur)
   }
 
+  # {ragg} and {svglite} report capabilities, but need specific version
+  if (dev_name %in% c("agg_jpeg", "agg_ppm", "agg_png", "agg_tiff")) {
+    check_installed(
+      "ragg", version = "1.2.6",
+      reason = paste0("checking device support for ", feat_name, ".")
```
olivroy, line 254:
```suggestion
      reason = paste0(for checking device support for ", feat_name, ".")
```

rlamg write pkg is required, so either to check or for checking

R/utilities-checks.R
```diff
@@ -247,6 +247,20 @@ check_device = function(feature, action = "warn", op = NULL, maybe = FALSE,
...
+    )
+  }
+  if (dev_name == "devSVG") {
+    check_installed(
+      "svglite", version = "2.1.2",
+      reason = paste0("checking device support for ", feat_name, ".")
```
olivroy, line 260:
```suggestion
      reason = paste0("for checking device support for ", feat_name, ".")
```

R/utilities-checks.R
teunbrand: Ah yes, good catch, thanks!

thomasp85 (approved):
LGTM

PR #5477, opened 2023-10-12
Ignoring `AsIs` objects (again)

### LINKED ISSUE #5142: Circumventing scales

teunbrand:
This PR attempts to fix #5142 and replaces #5290 and #5418.

Briefly, instead of assigning `scale_*_identity()` to `<AsIs>` objects, it ignores these in scale operations.

Benefits of this are:
  * You can safely do `aes(colour = I(...))` in one layers and do `aes(colour = proper_column)` in another layer, without having a clash of colour scales.
  * Using `x/y = I(...)` effectively makes the position aesthetic in an relative unit (`"npc"` units), which can be very convenient for annotations and such.

The mechanism this PR employs is to hide/unhide `<AsIs>` column at strategic moments during plot building.
'Ignoring' packs `<AsIs>` columns into one `.ignored` data.frame column.
'Exposing' unpacks the `.ignored` data.frame column into separate columns again.

A quick comparison with the other PRs:

Compared to #5290:

* No need for scales to make a formal exemption for `<AsIs>` columns; plot building takes care of this.
* For that reason, this PR has a much smaller footprint and is less invasive.

Compared to #5418:

* Don't need bother with mappings to derive which columns should be ignored.
* Because the `<AsIs>` class is 'infective', the mechanism automatically propagates through reparameterization, which is desirable (e.g. from x/width to xmin/xmax).

A quick benchmark to show that hiding/exposing scales well with dataset size (nrow = 32, 234 and 53.940 for mtcars, mpg and diamonds respectively). All in all, I think this PR wouldn't even add 1ms to plot building in 90% of plots.

``` r
#> ℹ Loading ggplot2

data <- list(
  transform(mtcars, cyl = I(cyl), disp  = I(disp)),
  transform(mpg,    cyl = I(cyl), displ = I(displ)),
  transform(diamonds, color = I(color), depth = I(depth))
)

bench::mark(
  "mtcars"   = ignore_data(data[[1]]),
  "mpg"      = ignore_data(data[[2]]),
  "diamonds" = ignore_data(data[[3]]),
  "all"      = ignore_data(data), 
  check = FALSE
)
#> # A tibble: 4 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 mtcars       38.7µs   41.3µs    22758.    3.84KB     4.55
#> 2 mpg          38.1µs   40.6µs    23882.        0B     4.78
#> 3 diamonds     39.5µs   42.2µs    23048.        0B     4.61
#> 4 all         107.9µs    114µs     8594.      304B     4.08

ignored <- ignore_data(data)

bench::mark(
  "mtcars"   = expose_data(ignored[[1]]),
  "mpg"      = expose_data(ignored[[2]]),
  "diamonds" = expose_data(ignored[[3]]),
  "all"      = expose_data(ignored),
  check = FALSE
)
#> # A tibble: 4 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 mtcars       11.2µs   12.4µs    77564.        0B     7.76
#> 2 mpg          11.3µs   12.2µs    76183.        0B     0   
#> 3 diamonds     11.2µs   12.4µs    78313.        0B     7.83
#> 4 all          25.6µs   27.9µs    34930.        0B     3.49
```

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -598,6 +598,47 @@ is_bang <- function(x) {
   is_call(x, "!", n = 1)
 }
 
+# Puts all columns with 'AsIs' type in a '.ignore' column.
+ignore_data <- function(data) {
+  if (!is_bare_list(data)) {
+    data <- list(data)
+  }
+  lapply(data, function(df) {
+    is_asis <- vapply(df, inherits, logical(1), what = "AsIs")
+    if (!any(is_asis)) {
+      return(df)
+    }
+    df <- unclass(df)
+    # We trust that 'df' is a valid data.frame with equal length columns etc,
+    # so we can use the more performant `new_data_frame()`
+    new_data_frame(c(
+      df[!is_asis],
+      list(.ignored = new_data_frame(df[is_asis]))
+    ))
+  })
+}
+
+# Restores all columns packed into the '.ignored' column.
+expose_data <- function(data) {
+  if (!is_bare_list(data)) {
+    data <- list(data)
+  }
+  lapply(data, function(df) {
+    is_ignored <- which(names(df) == ".ignored")
+    if (length(is_ignored) == 0) {
+      return(df)
+    }
+    df <- unclass(df)
+    new_data_frame(c(df[-is_ignored], df[[is_ignored[1]]]))
+  })
+}
+
+#' @export
+#' @method rescale AsIs
+rescale.AsIs <- function(x, to, from, ...) {
+  x
+}
+
 is_triple_bang <- function(x) {
   if (!is_bang(x)) {
     return(FALSE)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # ggplot2 (development version)
 
+* Plot scales now ignore `AsIs` objects constructed with `I(x)`, instead of
+  invoking the identity scale. This allows these columns to co-exist with other
+  layers that need a non-identity scale for the same aesthetic. Also, it makes
+  it easy to specify relative positions (@teunbrand, #5142).
+
 * Legend titles no longer take up space if they've been removed by setting 
   `legend.title = element_blank()` (@teunbrand, #3587).
 
--- a/R/plot-build.R
+++ b/R/plot-build.R
@@ -51,6 +51,7 @@ ggplot_build.ggplot <- function(plot) {
 
   # Compute aesthetics to produce data with generalised variable names
   data <- by_layer(function(l, d) l$compute_aesthetics(d, plot), layers, data, "computing aesthetics")
+  data <- ignore_data(data)
 
   # Transform all scales
   data <- lapply(data, scales$transform_df)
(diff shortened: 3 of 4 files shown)


R/utilities.R
```diff
@@ -598,6 +598,47 @@ is_bang <- function(x) {
...
+    df <- unclass(df)
+    new_data_frame(c(df[-is_ignored], df[[is_ignored[1]]]))
+  })
+}
+
+#' @export
+#' @method rescale AsIs
+rescale.AsIs <- function(x, to, from, ...) {
+  x
+}
```
teunbrand, line 640: Ideally this would live in {scales} because it is where the `rescale` generic lives.

teunbrand: Also one reason I'd like #5475 to be merged is so that this PR would work on `coord_sf()`.

teunbrand:
TODO: Add `.` prefix to functions

R/utilities.R
teunbrand: https://github.com/r-lib/scales/pull/404 was merged, so we might have to increment required scales version once that is released

thomasp85:
Can you update this for review?

teunbrand:
I've updated this PR with the latest suggestions

thomasp85 (approved):
LGTM

PR #5483, opened 2023-10-13
Move guide building to `ggplot_build()`

teunbrand:
This PR aims to take a step towards #3648.

Briefly, we might want to expose the plot `data` to the guides, so that the guides can observe the data they're supposed to represent. For this to happen, we cannot use the data available in `ggplot_gtable()` because the `after_scale()` transformations have been applied already and the original data is no longer suitable. Hence, we would need to move guide building to a point before `after_scale()` transformations. That last part is what this PR does and it is important to note that it does *not* do the data exposure (yet).

In order to achieve this, the following has to be taken into account:
* The `Guides$draw()` step needs to remain in `ggplot_gtable()` because it needs access to the theme.
* All earlier `Guides` methods need to be independent of the theme.

This means that guides don't know about the position or direction of the guide until the `Guide$draw()` step. That has the following implications:
* For the most part, this simply required validating directions/positions at the draw method instead of the train method.
* The hash used for checking uniqueness of guides can't incorporate the direction or position anymore, so some random order of guides have changed. I've fixed the order in some unit tests to prevent similar issues in the future.

More granular details in the comments below:


R/guide-bins.R
```diff
@@ -298,15 +298,15 @@ GuideBins <- ggproto(
...
     params$direction <- arg_match0(
-      params$direction %||% direction,
+      params$direction,
       c("horizontal", "vertical"), arg_nm = "direction"
     )
-    if (params$direction == "vertical") {
-      key$.value <- 1 - key$.value
-    }
-
-    params$key <- key
```
teunbrand, line 309: `guide_bins()` was the only guide that actually did something meaningful upon observing `direction` rather than just validating it.

R/guide-old.R
```diff
@@ -90,7 +90,7 @@ GuideOld <- ggproto(
   train = function(self, params, scale, aesthetic = NULL,
                    title = waiver(), direction = NULL) {
     params$title <- scale$make_title(params$title %|W|% scale$name %|W|% title)
-    params$direction <- params$direction %||% direction
+    params$direction <- params$direction %||% direction %||% "vertical"
```
teunbrand, line 93: For old guides, their methods may still expect `direction` to be available during training, which it is not anymore.
To preserve *most* of their functionality, we're defaulting their direction to vertical, which is probably 80% of plots.

R/plot-build.R
```diff
@@ -84,11 +84,18 @@ ggplot_build.ggplot <- function(plot) {
...
 
-  # Train and map non-position scales
+  # Hand off position guides to layout
+  layout$setup_panel_guides(plot$guides, plot$layers)
+
+  # Train and map non-position scales and guides
   npscales <- scales$non_position_scales()
   if (npscales$n() > 0) {
     lapply(data, npscales$train_df)
+    plot$guides <- plot$guides$build(npscales, plot$layers, plot$labels)
```
teunbrand, line 94: So at this point, it is still unclear whether guides should see mapped or unmapped data.
I've opted for unmapped data here, but if it should be mapped, than we need to swap this line with the line below.

teunbrand:
I've now added the plumbing to make sure that the plot's data reaches the guide building stage where guides interact with layers (at the `Guide$get_layer_key` method).

For this to make slightly more sense, I've added a `process_layers` method to guides, which only job currently is to see if a layer should be included and pass on the correct layers to `get_layer_key()`.

As a small demonstration that the plumbing works, here is a quick and dirty guide that reports which keys are actually represented in layers.

``` r
#> ℹ Loading ggplot2

orig <- guide_legend()
new <- ggproto(NULL, orig, get_layer_key = function(params, layers, data) {
  x <- params$key$.value %in% data[[1]]$colour
  print(setNames(x, params$key$.value))
  GuideLegend$get_layer_key(params, layers, data = data)
})

  geom_point(aes(colour = as.factor(cyl))) +
  scale_colour_discrete(limits = factor(4:8)) +
  guides(colour = new)

b <- ggplot_build(p)
#>     4     5     6     7     8 
#>  TRUE FALSE  TRUE FALSE  TRUE
```

<sup>Created on 2023-10-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM - Heads up @yutannihilation, @clauswilke, @hadley, @karawoo, @paleolimbot - this alters the build order which is generally sacred but we do this for good reasons.

It opens the door for much richer guides together with the guide extension possibilities that @teunbrand has developed

PR #5486, opened 2023-10-16
Element subclass inheritance

### LINKED ISSUE #5457: Should theme inheritance prefer the 'richer' element?
Sometimes, one thinks that one has set a theme element, but it turns out some downstream theme element overrides your settings.
In the example below, I want the markdown element to display the x-axis title in bold. However, it doesn't because `axis.title.x` already is a defined theme element in the default theme, which overrules the `element_markdown()`.

``` r
library(ggplot2)

  geom_point() +
  labs(x = "**displacement**")

p + theme(axis.title = ggtext::element_markdown())
```

Of course, there are plenty of ways to display the axis title in bold, such as setting a more appropriate theme element.

``` r
p + theme(axis.title.x = ggtext::element_markdown())
```

Since `<element_markdown>` is a subclass of `<element_text>`, I think it would be appropriate that the calculation below returns an `<element_markdown>` object.

``` r
calc_element(
  "axis.text.x.top",
  theme_get() + theme(axis.text = ggtext::element_markdown(colour = "red"))
)
#> List of 11
#>  $ family       : chr ""
#>  $ face         : chr "plain"
#>  $ colour       : chr "red"
#>  $ size         : num 8.8
#>  $ hjust        : num 0.5
#>  $ vjust        : num 0
#>  $ angle        : num 0
#>  $ lineheight   : num 0.9
#>  $ margin       : 'margin' num [1:4] 0points 0points 2.2points 0points
#>   ..- attr(*, "unit")= int 8
#>  $ debug        : logi FALSE
#>  $ inherit.blank: logi TRUE
#>  - attr(*, "class")= chr [1:2] "element_text" "element"
```

<sup>Created on 2023-10-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5457

Briefly, it checks if parent elements are a subclass of child elements and if they are, fill parent element with child values instead of completing missing/NULL child values with parent values.

Reprex from #5457:

``` r
#> ℹ Loading ggplot2
  geom_point() +
  labs(x = "**displacement**")

p + theme(axis.title = ggtext::element_markdown())
```

<sup>Created on 2023-10-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* In the theme element hierarchy, parent elements that are a strict subclass
+  of child elements now confer their subclass upon the children (#5457).
+
 * Legend titles no longer take up space if they've been removed by setting 
   `legend.title = element_blank()` (@teunbrand, #3587).
 
--- a/R/theme.R
+++ b/R/theme.R
@@ -767,7 +767,19 @@ combine_elements <- function(e1, e2) {
     e1$linewidth <- e2$linewidth * unclass(e1$linewidth)
   }
 
+  # If e2 is 'richer' than e1, fill e2 with e1 parameters
+  if (is.subclass(e2, e1)) {
+    new <- defaults(e1, e2)
+    e2[names(new)] <- new
+    return(e2)
+  }
+
   e1
+}
+
+is.subclass <- function(x, y) {
+  inheritance <- inherits(x, class(y), which = TRUE)
+  !any(inheritance == 0) && length(setdiff(class(x), class(y))) > 0
 }
 
 #' Reports whether x is a theme object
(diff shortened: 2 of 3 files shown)


thomasp85:
We agree this only has implications for when users actively use an extension package with a new element and overwrites an element in the theme tree, right?

teunbrand:
Yes this shouldn't have any implications elsewhere. It is just so that sublasses propagate through theme inheritance.

To use an example from `theme_grey()`; let's say a user sets a custom `axis.text = element_special_text(...)`, where `element_special_text()` produces an `element_text` subclass. Because `theme_grey()` has `axis.text.x = element_text(...)`, `axis.text.x.bottom` would *not* acquire the special subclass, whereas with this PR, it does acquire the special subclass.

teunbrand:
That was fast, thanks Thomas!

PR #5488, opened 2023-10-17
Arbitrary positions for guides

teunbrand:
When a plot has multiple guides, we're currently confined to exactly 1 side of the plot where guides are placed.
This PR adds a `position` argument to guides, that can overrule the side set in the theme.

Some details:

* The `Guides$build()` method calls `Guides$assemble()` for every position, delivering 5 guide-boxes (left, right, top, bottom and manual), instead of just once for one position.
* The code for adding guides to the plot's gtable has been isolated in a separate function. More so for my own sanity than necessity.
* If there are guides, the plot's gtable will get appropriate column/rows for *all* positions, regardless of whether there are guides for any specific position. (i.e. if all legends are placed on the left, the gtable will also get rows for bottom/top).
* The point above is mostly to make it easier for {patchwork} to align gtables, though this PR breaks patchwork at the moment.

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(displ, hwy, colour = cty, shape = factor(cyl), fill = class)) +
  geom_point() +
  scale_shape_manual(values = 21:25) +
  guides(
    colour = guide_colourbar(position = "left"),
    shape  = guide_legend(position = "bottom", direction = "horizontal")
  )
```

There are still some details to hammer out, namely that the guide direction should follow a custom `position` is set. Notice in the plot below that the colourbar has also changed direction, though it probably shouldn't have. Another thing to hammer out is when a guide declares `position = "manual"`, but the theme doesn't carry values for manual postions.

``` r
last_plot() + theme(legend.position = "top")
```

<sup>Created on 2023-10-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
A few updates.

1. Had to slightly change approach due to #5483. `Guides$draw()` has more responsibilities and `Guides$package_box()` wraps the legend box creation previously part of `Guides$assemble()`.
2. This made it easier to set the position-wise defaults, so that in the example above the colourbar isn't horizontal when placed on the left. See next example:

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(displ, hwy, colour = cty, shape = factor(cyl), fill = class)) +
  geom_point() +
  scale_shape_manual(values = 21:25) +
  guides(
    colour = guide_colourbar(position = "left"),
    shape  = guide_legend(position = "bottom", direction = "horizontal")
  ) +
```



3. I've also separated out the coordinates from `legend.position` into a new theme element that purely controls the positioning of inside legends. Providing a `<numeric>` to `legend.position` is now deprecated and should use `"inside"` to set the default placement to inside the plot.

``` r
last_plot() + theme(legend.position = "inside", 
                    legend.position.inside = c(0.8, 0.8))
```



<sup>Created on 2023-11-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
Legends can now be justified per position. Also seen below: justification with respect to full plot as in #4020.
Had to make some decision about how full-plot justification would work and legends would occupy same corner. I decided to give top/bottom legends priority as one might want to align these with the title.

``` r
#> ℹ Loading ggplot2

ggplot(mtcars) + 
  aes(
    disp, mpg, 
    colour = hp, 
    fill   = factor(gear),
    shape  = factor(cyl),
    size   = drat,
    alpha = wt
  ) +
  geom_point() +
  guides(
    shape  = guide_legend(position = "top"),
    colour = guide_colourbar(position = "bottom"),
    size   = guide_legend(position = "left"),
    alpha  = guide_legend(position = "right"),
    fill   = guide_legend(position = "inside", override.aes = list(shape = 21))
  ) +
  theme(
    legend.justification.top    = "left",
    legend.justification.bottom = c(1, 0),
    legend.justification.left   = c(0, 1),
    legend.justification.right  = "bottom",
    legend.justification.inside = c(0.75, 0.75),
    legend.location = "plot"
  )
```



<sup>Created on 2023-11-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
Reminder: check patchwork PR

teunbrand:
Add rows/cols all the time

teunbrand:
Now always includes rows/columns for every position and a `<zeroGrob>` when there are no legends for a position.

R/plot-build.R
```diff
@@ -481,3 +398,132 @@ table_add_tag <- function(table, label, theme) {
...
+    margin <- margin(1 - yjust, 1 - xjust, yjust, xjust, unit = "null")
+
+    vp <- viewport(
+      x = xjust, y = yjust, just = just,
+      height = .subset2(heights, position),
+      width  = .subset2(widths,  position)
+    )
+    box <- editGrob(box, vp = vp)
+    gtable_add_padding(box, margin)
+  })
```
teunbrand, line 457: For reasons I can't quite articulate, I feel like this code should be part of the guide-box building code instead of legend placement code.

R/guides-.R
```diff
@@ -458,49 +453,96 @@ Guides <- ggproto(
...
+      self$package_box,
+      MoreArgs = list(theme = theme)
+    )
+  },
+
+  # Render the guides into grobs
+  draw = function(self, theme, default_position = "right",
+                  params = self$params, guides = self$guides) {
+    positions <- vapply(
+      params, function(p) p$position[1] %||% default_position, character(1)
```
thomasp85, line 506: move the function argument to new line

R/guides-.R
```diff
@@ -458,49 +453,96 @@ Guides <- ggproto(
...
+    )
+    positions <- factor(positions, levels = c(.trbl, "inside"))
+
+    theme$legend.key.width  <- theme$legend.key.width  %||% theme$legend.key.size
+    theme$legend.key.height <- theme$legend.key.height %||% theme$legend.key.size
```
thomasp85, line 511: Are there any reason why `theme$legend.key.width` and `theme$legend.key.height` doesn't inherit from `theme$legend.key.size` and we use the theme calc infrastructure for this?

R/guides-.R
```diff
@@ -458,49 +453,96 @@ Guides <- ggproto(
...
+      MoreArgs = list(theme = theme)
+    )
+  },
+
+  # Render the guides into grobs
+  draw = function(self, theme, default_position = "right",
+                  params = self$params, guides = self$guides) {
+    positions <- vapply(
+      params, function(p) p$position[1] %||% default_position, character(1)
+    )
```
thomasp85, line 507: Is this really necessary since the `Guide$draw()` method already takes it's own position over the one supplied

R/plot-build.R
thomasp85: agreed - this feels like it belongs in `package_box()`

R/plot-build.R
```diff
@@ -481,3 +398,132 @@ table_add_tag <- function(table, label, theme) {
...
+        y = position[2],
+        just = just,
+        height = heights$inside,
+        width  = widths$inside
+      )
+    )
+  }
+
+  legends[.trbl] <- lapply(.trbl, function(position) {
+    box <- .subset2(legends, position)
```
thomasp85, line 437: Is `legends` a weird class that doesn't respond correctly to `[[`-indexing? If not we should use that

R/plot-build.R
```diff
@@ -481,3 +398,132 @@ table_add_tag <- function(table, label, theme) {
...
+  table
+}
+
+plot_extent <- function(table) {
+  layout <- table$layout
+  data_frame0(
+    t = min(.subset2(layout, "t")),
+    r = max(.subset2(layout, "r")),
+    b = max(.subset2(layout, "b")),
+    l = min(.subset2(layout, "l")),
```
thomasp85, line 526: Same with this use of `.subset2()`. If there is a good reason, fine. Otherwise use standard indexing

R/theme.R
```diff
@@ -455,6 +469,14 @@ theme <- function(line,
     }
     elements$legend.text.align <- NULL
   }
+  if (is.numeric(.subset2(elements, "legend.position"))) {
```
thomasp85, line 472: same again

R/guides-.R
teunbrand: Yeah might as well update this. The reason is historical accident I think.

R/guides-.R
teunbrand: This isn't for `Guide$draw()` method, but for the `split(grobs, position)` line after drawing the grobs.

R/theme.R
```diff
@@ -455,6 +469,14 @@ theme <- function(line,
...
     elements$legend.text.align <- NULL
   }
+  if (is.numeric(elements[["legend.position"]])) {
+    deprecate_soft0(
+      "3.5.0", I("A numeric `legend.position` argument in `theme()`"),
+      "theme(legend.position.inside)"
+    )
+    elements$legend.position.inside <- elements$legend.position
+    elements$legend.position <- "inside"
+  }
```
thomasp85, line 479: Can you elaborate on this change? I agree in principle but with the historic bagage I think we need a clear reason for deprecating a mode of a theme element

teunbrand: Mostly because there are several places where `legend.position` is used to make decisions about directions or justifications and it is a pain to add an `if (is.numeric(legend.position)` block every time. The numeric position itself is only used in one place relatively late in the guide building process, so converting it from numeric to `"inside"` once at the beginning is not an option. It just seemed cleaner to separate it.

teunbrand: Sorry I just remembered a clearer reason why I wanted to separate numeric position from position options. If `position = "inside"` at the guide level but at the theme level `legend.position` is not `"inside"` (or numeric), there wouldn't be a mechanism to control the placement of guides inside the panel (other than defaulting to the justification).

thomasp85: oh, so `position` can't be a numeric inside the `guide_*()` specification, only in the theme. I guess that makes sense that we want all "inside" guides in the same spot

teunbrand: Yes that's right, there is just one guide box for the inside position. I guess we *could* allow numeric positions in guide functions, but I wasn't prepared to deal with that chaos quite yet.

teunbrand:
Thanks for the review Thomas!

PR #5490, opened 2023-10-21
fix typo in doc

jan-glx:
In `annotate`'s doc:
"You can use these geoms **directory** for annotations."
 should read 
"You can use these geoms **directly**   for annotations."


teunbrand (approved):
Great, thanks!

PR #5492, opened 2023-10-24
`create.dir` argument for `ggsave()`

### LINKED ISSUE #5489: ggsave creating new directories when saving jpg/tiff/png devices
Related feature request #3021  was closed as not desired. ggsave however, indeed creates new directories when saving. 

This happens only with png, tiff, and jpeg devices
```
## all of the below would create a directory that does not exist
ggsave("test/test.png", plot = ggplot())
ggsave("test/test.jpg", plot = ggplot())
ggsave("test/test.tiff", plot = ggplot())
```
Does not happen with eps, ps, pdf, bmp, svg (error message). 
```
## All of the below give an error message 
ggsave("test/test.eps", plot = ggplot())
ggsave("test/test.ps", plot = ggplot())
ggsave("test/test.bmp", plot = ggplot())
ggsave("test/test.svg", plot = ggplot())
ggsave("test/test.pdf", plot = ggplot())
```

P.S. behaviour originally described in [this stack overflow thread](https://stackoverflow.com/q/77330346/7941188)

teunbrand:
This PR aims to fix #5489.

Briefly, it adds a `create.dir` argument to `ggsave()`, which controls whether a new directory can be created.

Example:
``` r
#> ℹ Loading ggplot2

p <- ggplot()

ggsave("foobar/test.png", plot = p)
#> Error in `ggsave()`:
#> ! Cannot find directory 'foobar'.
#> ℹ Please supply an existing directory or use `create.dir = TRUE`.

ggsave("foobar/test.png", plot = p, create.dir = TRUE)
#> ✔ Created directory: 'foobar'.
#> Saving 7 x 5 in image
```

<sup>Created on 2023-10-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Because the default `create.dir = FALSE` might be pain for interactive use, it prompts a small menu choice in interactive mode (which I don't know how to reprex)

```r
ggsave("foobar/test.png", plot = p)
#> Cannot find directory foobar.
#> ℹ Would you like to create a new directory?
#>
#> 1: Yes
#> 2: No
#>
#> Selection: 1
#> ✔ Created directory: foobar.
#> Saving 7 x 7 in image
```

### DIFF
--- a/R/save.R
+++ b/R/save.R
@@ -43,6 +43,10 @@
 #'   specifying dimensions in pixels.
 #' @param bg Background colour. If `NULL`, uses the `plot.background` fill value
 #'   from the plot theme.
+#' @param create.dir Whether to create new directories if a non-existing
+#'   directory is specified in the `filename` or `path` (`TRUE`) or return an
+#'   error (`FALSE`, default). If `FALSE` and run in an interactive session,
+#'   a prompt will appear asking to create a new directory when necessary.
 #' @param ... Other arguments passed on to the graphics device function,
 #'   as specified by `device`.
 #' @export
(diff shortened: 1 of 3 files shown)


R/save.R
```diff
@@ -119,6 +112,56 @@ ggsave <- function(filename, plot = last_plot(),
...
+                       call = caller_env()) {
+
+  if (length(filename) > 1 && is.character(filename)) {
+    cli::cli_warn(c(
+      "{.arg filename} must have length 1, not {length(filename)}.",
+      "!" = "Only the first, {.file {filename[1]}}, will be used."
+    ), call = call)
+    filename <- filename[1]
+  }
+  check_string(filename, allow_empty = FALSE)
```
olivroy, line 125:
```suggestion
  check_string(filename, allow_empty = FALSE, call = call)
```

R/save.R
```diff
@@ -119,6 +112,56 @@ ggsave <- function(filename, plot = last_plot(),
...
+
+  check_string(path, allow_empty = FALSE, allow_null = TRUE)
```
olivroy, line 127:
```suggestion
  check_string(path, allow_empty = FALSE, allow_null = TRUE, call = call)
```

R/save.R
```diff
@@ -119,6 +112,56 @@ ggsave <- function(filename, plot = last_plot(),
...
+  } else {
+    path <- dirname(filename)
+  }
+
+  # Happy path: target file is in valid directory
+  if (dir.exists(path)) {
+    return(filename)
+  }
+
+  check_bool(create.dir)
```
olivroy, line 139:
```suggestion
  check_bool(create.dir, call = call)
```

olivroy (commented):
I think that it would be better to pass down the error call to the type checks as well!

teunbrand:
Thanks for these suggestions olivroy!

thomasp85 (approved):
LGTM - maybe add a check for whether `create.dir` is `NA`

PR #5495, opened 2023-10-26
`guide_coloursteps()` always fits rectangles

### LINKED ISSUE #5481: Faint lines through colour steps guide segments when `even.steps = FALSE`
This reprex actually comes from the current ggplot2 3.4.4 docs for `guide_colorsteps`, where it is visible:

```r
df <- expand.grid(X1 = 1:10, X2 = 1:10)
df$value <- df$X1 * df$X2

  p <- ggplot(df, aes(X1, X2)) + geom_tile(aes(fill = value))
```

```r
# By default each bin in the guide is the same size irrespectively of how
# their sizes relate in data space
p + scale_fill_binned(breaks = c(10, 25, 50))
```

image

```r
# This can be changed with the `even.steps` argument
p + scale_fill_binned(
  breaks = c(10, 25, 50),
  guide = guide_coloursteps(even.steps = FALSE))
```

image

When I run this code on ggplot 3.4.4, I can't see the faint lines at low-resolution outputs, but if I run, ggsave at, say, `width = 1920, height = 1080, units = 'px'`, or with SVG output, they're definitely visible. Mapping `colour` to the same variable as `fill` hides the lines from the tiles themselves but not from the colour bar (unless there's a way to merge the fill and colour guides rather than hiding one?).

teunbrand:
This PR aims to fix #5481.

Briefly, when `even.steps = FALSE`, `guide_coloursteps()` used to defer to `guide_colourbar()`'s method of making a bar, which resulted in 300 rectangles by default. This PR gives `GuideColoursteps` its own `build_decor()` method that fits the rectangles to the breaks/limits, alleviating anti-aliasing artefacts in the case described above.


R/guide-colorsteps.R
```diff
@@ -158,25 +154,73 @@ GuideColoursteps <- ggproto(
...
+      "vertical"   = c("right", "left")
+    )
+    params$label.position <- params$label.position %||% valid_label_pos[1]
+    if (!params$label.position %in% valid_label_pos) {
+      cli::cli_abort(paste0(
+        "When {.arg direction} is {.val {params$direction}}, ",
+        "{.arg label.position} must be one of {.or {.val {valid_label_pos}}}, ",
+        "not {.val {params$label.position}}."
+      ))
+    }
```
teunbrand, line 192: This is all ported because we don't depend on `GuideColourbar$extract_params()` anymore.
This portion interacts with #5483, so should keep an eye on this.

thomasp85 (approved):
LGTM

PR #5381, opened 2023-08-07
Update documentation and example in scale-brewer.R to reflect scale_*_distiller's non-intuitive "direction"

ahcyip:
Capturing in documentation that scale_*_distiller has a non-intuitive default direction = -1, and that to reverse, user needs to specify direction = 1

Source:
https://github.com/tidyverse/ggplot2/issues/1439
https://github.com/tidyverse/ggplot2/issues/4444

PR #5499, opened 2023-10-27
Put linebreak in example of `scale_brewer`

teunbrand:
This PR fixes a NOTE in the CI:

```
❯ checking Rd line widths ... NOTE
  Rd file 'scale_brewer.Rd':
    \examples lines wider than 100 characters:
       # the order of colour can be reversed, but with scale_*_distiller(), the default direction = -1, so to reverse, use direction = 1.
  
  These lines will be truncated in the PDF manual.
```

This example was last updated in #5381


thomasp85 (approved):
LGTM

PR #5500, opened 2023-10-28
Axis guide for logarithmic ticks

### LINKED ISSUE #5325: Add option to reverse annotation_logticks
This is a renewal of issue #3822 which was closed due to lack of reprex and detail.

Sometimes a log scale needs to be shown in reverse. For example, when creating a volcano plot, the y-axis shows a -log10 representation of the p-value.

I would like to be able to add log ticks to the axis and have them correspond to the axis. Right now, `annotation_logticks()` only works in one direction and not in reverse. An option `reverse = 1` would be very useful.

```r
library(ggplot2)
df <- read.table(header = TRUE, text = "
                        p      ratio
                      1  0.28801033  3.7712035
                      2  0.48073187  0.8304503
                      3  0.06113270  0.7417715
                      4  0.96118760  1.7041385
                      5  0.07617357  0.7598174
                      6  0.15079531  0.7283150
                      7  0.29093358  0.8806807
                      8  0.12198144  0.5412664
                      9  0.18261693  1.4875495
                      10 0.36864248  1.0625448
                      11 0.04297883 72.6147534
                      12 0.01377004 32.0841697")

df$minus_log10_pvalue <- -log10(df$p)
df$log2_fc <- log2(df$ratio) 

ggplot(df, aes(x = log2_fc, y = minus_log10_pvalue)) +
  geom_point() +
  scale_x_continuous(limits = c(-max(abs(df$log2_fc)),max(abs(df$log2_fc)))) + 
  scale_y_continuous(limits = c(0,3), labels = function(i) 10^-i) +
  annotation_logticks(sides = "l", outside = TRUE) +
  coord_cartesian(clip = "off") +
  labs(x = "Fold change (log2)", y = "P-value") +
  theme_classic(9)
```

The output shows the nine intervening ticks between the major ticks being bunched towards the smaller rather than larger value. The same result is seen using the original data and the appropriate `trans` arguments for `scale_y_continuous()`

```r
library(scales)
ggplot(df, aes(x = log2_fc, y = p)) +
  geom_point() +
  scale_x_continuous(limits = c(-max(abs(df$log2_fc)),max(abs(df$log2_fc)))) + 
  scale_y_continuous(trans = c("log10", "reverse")) +
  annotation_logticks(sides = "l", outside = TRUE) +
  coord_cartesian(clip = "off") +
  labs(x = "Fold change (log2)", y = "P-value") +
  theme_classic(9)
```

teunbrand:
This PR aims to fix #5325, in particular through https://github.com/tidyverse/ggplot2/issues/5325#issuecomment-1595664134.

Briefly, it adds a new axis guide, `guide_axis_logticks()` that draws ticks at log10-spaced positions.

In comparison to `annotation_logticks()`:
* **Only** base 10 ticks. None of these base 2 or base *e* ticks. Even `?annotation_logticks` says: "These tick marks probably make sense only for base 10.", which I agree with.
* Proper axis, meaning label placement works out naturally and don't have to set `clip = "off"` when outside. Of course, can do equivalent of `annotation_logticks(outside = FALSE)` by setting negative tick lengths.
* Takes style from theme
* Has access to scale, so no need to guess whether user has transformed through scale or coord. User only needs to intervene when manually transformed.

In comparison with `guide_axis()`:
* Exact same `key` is used for the labels. The ticks get their own 'shadow'-key.
* Ticks all use `axis.ticks.{position}.{aes}` theme setting, so **no tick** uses the `axis.minor.ticks.{position}.{aes}` setting.

Why WIP?
* Should we be able to control the style of the minor/5 ticks and smallest/1 ticks seperately from the major/10 ticks? If so, do we handle that in the guide or make new theme settings?
* Should we mark `annotation_logticks()` as superseded?
* Once/if #5410 gets merged, we may think about providing a custom `breaks` and/or `labels` argument.

A bunch of demonstrations; ticks are placed in non-transformed, original data ranges. The label placement shows that `guide_axis_logticks()` does not intervene with that part of the guide and takes these verbatim from the scale.

``` r
#> ℹ Loading ggplot2

# Tiny wrapper for demonstrating scale/guide interactions
demo <- function(limits, trans = "identity", guide = "axis_logticks", ...) {
  scales::demo_continuous(limits, trans = trans, guide = guide, ...)
}

demo(c(0.1, 100))
#> scale_x_continuous(trans = trans, guide = guide)
```

Of course, these ticks start to make more sense with a log transformation:

``` r
demo(c(0.1, 100), trans = "log10")
#> scale_x_continuous(trans = trans, guide = guide)
```

Tick lengths are set in the guide. If not a `<unit>` they are interpreted as relative to the theme's tick length setting.

``` r
demo(
  c(0.1, 100), trans = "log10",
  guide = guide_axis_logticks(long = 5, mid = 2, short = 1)
)
#> scale_x_continuous(trans = trans, guide = guide)
```

By default, the ticks are placed 1 order of magnitude outside the limits. To clip these to the actual limits rather than the expanded limits, you can set `expanded = FALSE`.

``` r
demo(
  c(0.1, 100), trans = "log10", expand = expansion(add = 0.5),
  guide = guide_axis_logticks(expanded = FALSE)
)
#> scale_x_continuous(trans = trans, guide = guide, expand = expansion(add = 0.5))
```

When a user has manually transformed their data, the scale doesn't know about this so the ticks would be off. To intervene, one can set `prescale_base` to calculate appropriate ticks.

``` r
demo(log10(c(0.1, 100)), guide = guide_axis_logticks(prescale_base = 10))
#> scale_x_continuous(trans = trans, guide = guide)
```

Because ticks are placed in original data-space, this works naturally with compound transformations, like the reverse log10 transform.

``` r
demo(c(0.1, 100), trans = c("log10", "reverse"))
#> scale_x_continuous(trans = trans, guide = guide)
```

When the data include 0 or negative numbers, the ticks are mirrored around 0. 
The reason for this is as follows: 

1. It works out really naturally with pseudo-log (and alike) transformations.
1. A true log transform cannot include 0 or negative numbers, so we're not harming these transformations by mirroring.

``` r
demo(c(-100, 100), trans = "pseudo_log")
#> scale_x_continuous(trans = trans, guide = guide)
```

Of course, some decision then has to be made where near 0 one should stop drawing ticks, as otherwise one would end up with an infinite amount of ticks. The default is `0.1` but the user can control this too.

``` r
demo(
  c(-100, 100), trans = "pseudo_log", 
  guide = guide_axis_logticks(negative_small = 1)
)
#> scale_x_continuous(trans = trans, guide = guide)
```

Just to show that I'm not cheating this mirroring by seeing if the scale name is `"pseudo-log"`, we can do the same with custom transformations:

``` r
demo(c(-100, 100), trans = trans_new("", asinh, sinh))
#> scale_x_continuous(trans = trans, guide = guide)
```

<sup>Created on 2023-10-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
Now has customisation options for short ticks. Long ticks inherit from `axis.ticks`, middle ticks inherit from `axis.minor.ticks` and smallest ticks are set in `guide_axis_logticks(short_theme)` and inherit from the minor ticks. Still unsure if there wouldn't be a better argument name than `short_theme`.

In example below, you can see that the smallest ticks can be turned off on the x-axis, and the inheritance on the y-axis.

``` r
#> ℹ Loading ggplot2

ggplot(msleep, aes(bodywt, brainwt)) +
  geom_point(na.rm = TRUE) +
  scale_x_log10(guide = guide_axis_logticks(
    short = 2,
    short_theme = element_blank()
  )) +
  scale_y_log10(guide = guide_axis_logticks(
    short_theme = element_line(linewidth = 2)
  )) +
  theme(
    axis.minor.ticks.y.left = element_line(colour = "red")
  )
```



<sup>Created on 2023-11-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review Thomas!

PR #5502, opened 2023-10-30
Smarter keys (attempt 2)

### LINKED ISSUE #3648: Implement smart merging of guides so inappropriate key glyphs are omitted

teunbrand:
This PR aims to fix #3648 and supersedes #5302. It is incompatible with #3649.

Briefly, in this PR, legend guides check whether a value is present in layer data and if not, omits that key for that layer.

Less briefly, the default `layer(show.key = NA)` used to mean 'if layer has mapping with aesthetic, include it in legend, otherwise don't'. That remains true, but `show.key = NA` now also checks on a per-value basis whether that value is present in the layer data. To include or omit all values, regardless of whether they are present, one can use `layer(show.key = TRUE/FALSE)`, like before. This means that the previous behaviour can easily be reinstated by users. Note that this check only applies to discrete values, continuous values are *always* included (if the layer has the relevant mapping).

Compared to #5302, this PR makes use of the plumbing introduced in #5483, making this a much less invasive PR.

There is one visual test that changed for the better.

A demonstration that the layers now only draw the relevant keys:

``` r
#> ℹ Loading ggplot2

ggplot(data.frame(x = 1:2), aes(x, x)) +
  geom_point(aes(colour = "point", alpha = "point")) +
  geom_line(aes(colour = "line", alpha = "line"))
#> Warning: Using alpha for a discrete variable is not advised.
```

As `show.legend` can be a named logical vector, this can still be used to finetune inclusion or omission of keys. Below, we're omitting points for the alpha legend, but include all line keys for the alpha legend. This makes no sense for data visualisation, but it shows that the behaviour makes sense.

``` r
ggplot(data.frame(x = 1:2), aes(x, x)) +
  geom_point(
    aes(colour = "point", alpha = "point"),
    show.legend = c("colour" = NA, alpha = FALSE)
  ) +
  geom_line(
    aes(colour = "line", alpha = "line"),
    show.legend = c("colour" = NA, alpha = TRUE)
  )
#> Warning: Using alpha for a discrete variable is not advised.
```

<sup>Created on 2023-10-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review Thomas!

PR #5504, opened 2023-10-31
Use `scales::label_*` over scales::*format()`

olivroy:
Also added a period at the end of the `check_installed()` messages for consistency. https://style.tidyverse.org/error-messages.html


teunbrand (approved):
All these changes look good to me, thanks for the contribution!

Issue #5326, opened 2023-06-17
sf_transform_xy() warns a lot


yutannihilation:
I see this warning when I run tests, and I found `sf_transform_xy()` emits this.

```
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
```

I'm yet to figure out what's happening here, but it seems this is the cause: https://github.com/r-spatial/sf/issues/2166 Considering the test doesn't fail, I think this is not a serious problem at least at the moment. But let me report here anyway. 

Reprex is below. But, since this seems somehow related to the version of PROJ, you might not be able to reproduce this on your machine (I'm on Windows).

``` r
polygon <- sf::st_sfc(
  sf::st_polygon(list(matrix(c(-80, -76, -76, -80, -80, 35, 35, 40, 40, 35), ncol = 2))),
  crs = 4326 # basic long-lat crs
)
polygon <- sf::st_transform(polygon, crs = 3347)

points <- data.frame(
  x = c(-80, -80, -76, -76),
  y = c(35, 40, 35, 40)
)

points_trans <- ggplot2:::sf_transform_xy(points, 3347, 4326)
```

<sup>Created on 2023-06-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Standard output and standard error
</summary>

``` sh
✖ Install the styler package in order to use `style = TRUE`.
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
pipeline: Pipeline: Bad step definition: inv (File not found or invalid)
hgridshift: could not find required grid(s).
[...]

PR #5507, opened 2023-11-03
Silence tests

teunbrand:
Some tests are emitting messages that are mildly distracting when looking at the output of `devtools::test()`.
In this PR, I've silenced the messages that I found distracting, either by providing the defaults or suppressing messages that are thrown along with errors. In addition, I found that the `check_device()` tests were creating spurious files.

What messages are *not* fixed by this PR:
* #5326
* Sometimes `dev.capabilities()` used in `check_device()` throws a 'stack inbalance' message. For some reason, this only appears to occur when run in tests, but not in console, so I'm a bit puzzled still.

### DIFF
--- a/tests/testthat/test-fortify.R
+++ b/tests/testthat/test-fortify.R
@@ -1,5 +1,8 @@
 test_that("spatial polygons have correct ordering", {
-  skip_if_not_installed("sp")
+  suppressPackageStartupMessages({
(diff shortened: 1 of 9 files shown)


olivroy (approved):
I think this is important. It would be great if this was merged! Especially for these forgotten snapshot updates by someone ;)

Issue #5348, opened 2023-07-11
Feature request: `(sub)theme` argument for `guide_legend()` and friends


teunbrand:
Of the 21 arguments that `guide_legend()` has, more than half (11) are proxies for theme settings. 

<details>

* `title.theme` proxies `theme(legend.title)`
* `title.hjust` proxies `theme(legend.title = element_text(hjust))`
* `title.hjust` proxies `theme(legend.title = element_text(vjust))`
* `label` proxies `theme(legend.text = element_text/blank())`
* `label.theme` proxies `theme(legend.text = element_text())`
* `label.hjust` proxies `theme(legend.text = element_text(hjust))`
* `label.vjust` proxies `theme(legend.text = element_text(vjust))`
* `keywidth`/`barwidth` proxies `theme(legend.key.width)`
* `keyheight`/`barheight` proxies `theme(legend.key.height)`
* `direction` proxies `theme(legend.direction)`.
* `default.unit` is just syntactic sugar for the `keywidth` and `keyheight` arguments.

</details>

These are great for customising one legend in particular. However, it is a bit of a burden on the code to have this messy, manual layer of inheritance underneath the regular theme's inheritance. In addition, there are also theme settings that apply to all legends at once and cannot be customised per legend, even though you might want to.

<details>

You can set the following in `theme()` but not at the legend level:

* `legend.background`
* `legend.margin`
* `legend.spacing(.{x/y})`
* `legend.key`

<details>

As an aside, there is also the inverse case where styling options are provided in the `guide_*()` function but not in the theme:

* `guide_colourbar(frame, frame.colour, frame.linewidth, frame.linetype, ticks, ticks.colour, ticks.linewidth, ticks.length)`
* same for `guide_coloursteps()`
* `guide_bins(axis.colour, axis.linewidth, axis.arrow, ticks, ticks.length)`

</details>

</details>

Adding these "missing" theme settings to guides makes their formals even more complicated than they already are. 

So I'm proposing an alternative: get rid of all (or most of) these theme arguments in legends, and instead add one `subtheme` (or just `theme`) argument to guides that takes a (incomplete) theme object. This then:

1. Adds the flexibility of using the "missing" theme arguments on a per-guide basis, hassle free.
1. Reduces the complexity of arguments to `guide_*()` functions.
1. Makes for better maintainable code, as one could simply do `actual_theme <- plot_theme + subtheme` once instead of meticulously curating inheritance.

[...]

Issue #4814, opened 2022-04-26
Legend mis-alignement for parsing greek letters using scale::parse_format


Sibojang9:
In the first figure, the right-hand side legend is correctly aligned with the regular ggplot code. However, the greek characters were mis-aligned when `labels=parse_format()` is used to parse "microgram" in the second figure, as indicated by the gap (white space) next to the legend "1 ug". The erratic alignment behavior sounds like a bug to me.

 ``` r
library(ggplot2)
library(scales)
data <- data.frame(x=c(1,2),y=c(1,2),c= c("10000~mu*g","1~mu*g"))
ggplot(data,aes(x=x,y=y,color=c)) + geom_point()
```

![](https://i.imgur.com/cWnuk6s.png)

``` r
ggplot(data,aes(x=x,y=y,color=c)) + geom_point()+ scale_colour_discrete(labels=parse_format()) 
```

![](https://i.imgur.com/n0w1iGh.png)

<sup>Created on 2022-04-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

thomasp85:
While poorly documented, the default justification for expressions in labels has always been right-aligned. This can of course be overwritten in the guide. I can't comment on the reason for this decision as it predates my time with ggplot2 (meaning it is ancient 🙂). This also means that it is probably too late to change as others may likely rely on the default behaviour...

Sibojang9:
This is a typical sunk cost fallacy.

thomasp85:
Well, it's not really a bug so there is nothing to fix later. It was a poorly documented behaviour that is now properly documented.

Sibojang9:
Do you have the link to the docu?

thomasp85:
https://github.com/tidyverse/ggplot2/blob/eb0010e0ede79690a70e4487e07037c307ffd5bc/R/guide-legend.r#L34-L36

Sibojang9:
Thank you so much for sharing.

PR #5512, opened 2023-11-08
WIP: Cleanup legend code

teunbrand:
The legend code is lettered with all sorts of fussing about theme/element settings. This PR aims to to limit that clutter and make the reasoning about guide/theme interaction much clearer.

Essentially, it implements #5348 internally and passes around an `internal_theme` object that will get merged with the plot's theme. This has the following advantages:

* Reduces clutter
* `Guide$setup_elements()` becomes the *pure* place to break theme inheritance without overriding guide-level user input.
* `Guide$override_elements()` becomes the place to override guide-level user input.

It made the concession that expression vectors are no longer right-aligned by default (see #4814), as it seemed just very out-of-place to me. Please let me know if this should be preserved.

Current blockers:
* Forgot to implement #5486 for `merge_element()`
* #5455 gives a changed visual test that is changed and should probably be adressed before merging this.


R/theme.R
```diff
@@ -775,12 +774,12 @@ combine_elements <- function(e1, e2) {
...
   # Calculate relative sizes
   if (is.rel(e1$size)) {
-    e1$size <- e2$size * unclass(e1$size)
+    e1$size <- (e2$size %||% rel(1)) * unclass(e1$size)
   }
 
   # Calculate relative linewidth
   if (is.rel(e1$linewidth)) {
-    e1$linewidth <- e2$linewidth * unclass(e1$linewidth)
+    e1$linewidth <- (e2$linewidth %||% rel(1)) * unclass(e1$linewidth)
```
teunbrand, line 782: If `e2` has `NULL` size/linewidth, this used to result in `numeric(0)` as size/linewidth. {grid} cannot deal with 0-length gpar settings.

teunbrand:
OK, so the main idea here is that we can abstract away most of the theme-guide interaction fuss, by just having the guide carry an internal theme that can be added to the plot's theme.

However, this is currently also fussy due to the following behaviours of `add_theme()`.`

1. it happily merges text over blank, even if `inherit.blank = TRUE`.

``` r
#> ℹ Loading ggplot2

add_theme(
  theme(text = element_blank()), # plot theme
  theme(text = element_text(inherit.blank = TRUE)) # guide theme
)$text |> class()
#> [1] "element_text" "element"
```

2. It lets a proper element be overruled by `NULL` elements

``` r
add_theme(
  theme(text = element_text()), # plot theme
  theme(text = NULL) # guide theme
)$text
#> NULL
```

3. We cannot merge simpler elements into richer elements.

``` r
add_theme(
  theme(text = ggtext::element_markdown()), # plot theme
  theme(text = element_text(hjust = 0.5)) # guide theme
)
#> Error:
#> ! Problem merging the `text` theme element
#> Caused by error in `merge_element()` at ggplot2/R/theme.R:540:7:
#> ! Only elements of the same class can be merged
```

<sup>Created on 2023-11-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

For now, I've worked around this by making a custom theme merging function, but maybe there are smarter solutions that I couldn't come up with.

teunbrand:
Closing this in favour of #5554

PR #5515, opened 2023-11-09
Stretching legends

### LINKED ISSUE #4896: Feature request - Align horizontal legend width and plotting area
I need to align a horizontal legend with the square defined by the x and y-axis, as in the following example. This is partially related to #3989. 

```r
library(patchwork); library(dplyr)

g <- ggplot(mtcars, aes(mpg, wt, color = as.factor(am)))+
  geom_point()+
  guides(color = "none")

l <- ggplot(mtcars %>% distinct(am), 
            aes(as.factor(am), 1,  label = am)) +
  geom_point(aes(color = as.factor(am))) +
  geom_text(hjust = -1) +
  guides(color = "none") +
  theme(axis.text = element_blank(),
        axis.title = element_blank(),
        panel.grid = element_blank(),
        axis.ticks = element_blank())

l + g + plot_layout(heights = c(1,20))
```

image

My idea is to be able to do this for any plot, ideally with pure ggplot2, and that the legend width is preserved when I use `ggsave().` In addition, I've used different options inside `theme()` with the idea of doing this by using `NPC`

```
library(ggplot2)

g <- ggplot(mtcars) +
  geom_col(aes(x = cyl, y = mpg, fill = as.factor(am))) +
  theme(
    legend.position = "top",
    legend.justification = "center",
    legend.background = element_rect(fill = c("#f2f0f2"), colour = NA),
    legend.margin = margin(t = 0, r = 1, b = 0, l = 0, unit = "npc"),
    complete = T
  ) +
  theme(plot.background = element_rect(fill = "white", colour = NA))

g
```

teunbrand:
This PR aims to fix #4896.

Briefly it allows `keywidth`/`keyheight` and `barwidth`/`barheight` to be 'null' units, which fills the legends to the available space.

In somewhat more detail, 'null' in grid units are typically (to my understanding) interpreted as 'calculate all other units first, then divide the remaining space among the null units'. In this PR, I propagate that behaviour for legend key sizes. This works in several steps:

1. Guides as calculated as before, it is just in the `Guide$assemble_drawing()` stage that the key sizes are replaced with the provided 'null' units if provided by the user in the `guide_*()` function (i.e., we don't propagate 'null' units set in the theme).
2. During the guide box assembly, it is checked if any legend's size contains a 'null' unit and if so, interpret the null units and translate these to 'npc' units. This step is necessary because the interpretation in grid of 'null' unit nested in some 'sum' unit is different than that in a plain simple unit.
3. During `ggplot_gtable()`, the legend size is set to 1npc if 'sum' units are encountered in the legend size. These 'sum' units should only appear when a 'null' had  been set somewhere, because everything else in legends is casted to centimeters. This aligns the stretched legends to the panel.

I didn't want to mess with the guts of `<unit>` objects too much, so at several places I just delay taking the sum of units to look into their types. I had to improvise a `unitType` function for R <4.0, which works, but might not be great. In any case, in 2024 the minimum supported version of R should be R 4.0, so this might not matter for much longer.

I've marked this option as 'experimental' in the documentation, just to prepare users for the possibility that there might be unexpected behaviour that I cannot foresee at this point.

On to examples: if we want to implement #4896, we want to might want to start like this:

``` r
#> ℹ Loading ggplot2

ggplot(mtcars, aes(mpg, wt, colour = as.factor(am))) +
  geom_point() +
  guides(colour = guide_legend(
    keywidth = unit(1, "null"), 
    label.position = "bottom")
  ) +
```

This isn't quite what the result should be, so we'd have to set the legend margins to 0 and remove the title to get full alignment.

``` r
last_plot() + theme(
  legend.title = element_blank(),
  legend.margin = margin()
)
```

This shows a stretched colourbar:

``` r
  geom_point(aes(colour = drat)) +
  guides(colour = guide_colourbar(barheight = unit(1, "null")))
```

This shows a stretched legend:

``` r
  geom_point(aes(shape = factor(cyl))) +
  guides(shape = guide_legend(keyheight = unit(1, "null")))
```

The behaviour when you combine an absolute legend and relative legend follows the 'null' heuristics described above: first absolute units are honoured, then remainder is divided among null units.

``` r
  geom_point(aes(shape = factor(cyl), colour = drat)) +
  guides(shape = guide_legend(keyheight = unit(1, "null")))
```

When you just have 1 legend, the value of the 'null' unit doesn't matter, but it can come in handy when combining legends and you want to give them a different size.

``` r
  geom_point(aes(shape = factor(cyl), colour = factor(vs))) +
  guides(shape  = guide_legend(keyheight = unit(1, "null")),
         colour = guide_legend(keyheight = unit(0.5, "null"))) +
  theme(legend.key = element_rect(colour = "black"))
```

It was not really clear to me what should happen when the opposite size (that isn't parallel to the panel side) is set as a null unit, since there isn't any 'available space' to stretch into. With the current code, the 'available space' is 0, so the guide is shrunk. Note that this exactly the same as it works in CRAN ggplot2, so this isn't a regression.

``` r
  geom_point(aes(colour = drat)) +
  guides(colour = guide_colourbar(barwidth = unit(1, "null")))
```

<sup>Created on 2023-11-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/backports.R
```diff
@@ -17,6 +17,26 @@ if (getRversion() < "3.3") {
...
+  }
+  rep("", length(x)) # we're only interested in simple units for now
+}
+
+on_load({
+  if ("unitType" %in% getNamespaceExports("grid")) {
+    unitType <- grid::unitType
+  }
+})
+
```
teunbrand, line 39: If this gets merged in 2024 or later, we may drop this bit.

R/guide-colorbar.R
```diff
@@ -460,29 +459,29 @@ GuideColourbar <- ggproto(
...
-        height <- elements$key.height
+        width  <- 1 / nrow(decor)
+        height <- 1
         x <- (seq(nrow(decor)) - 1) * width
         y <- 0
       } else {
-        width  <- elements$key.width
-        height <- elements$key.height / nrow(decor)
+        width  <- 1
+        height <- 1 / nrow(decor)
```
teunbrand, line 476: Setting these sizes to npcs is benight because the actual size is set in `Guide$measure_grobs()`.

R/guide-legend.R
```diff
@@ -648,11 +647,19 @@ GuideLegend <- ggproto(
...
-      widths  = unit(c(sizes$padding[4], sizes$widths, sizes$padding[2]), "cm"),
-      heights = unit(c(sizes$padding[1], sizes$heights, sizes$padding[3]), "cm")
-    )
+    heights <- unit(c(sizes$padding[1], sizes$heights, sizes$padding[3]), "cm")
+    if (is.unit(params$keyheight) && unitType(params$keyheight) == "null") {
+      i <- unique(layout$layout$key_row)
+      heights[i] <- params$keyheight
+    }
+
+    gt <- gtable(widths = widths, heights = heights)
```
teunbrand, line 662: This is step 1

R/guides-.R
```diff
@@ -523,13 +523,16 @@ Guides <- ggproto(
                         height = heightDetails(grobs[[i]]))
         )
       }
+      spacing <- convertWidth(theme$legend.spacing.x, "cm")
+      widths  <- redistribute_null_units(widths, spacing, margin, "width")
```
teunbrand, line 527: This is where step 2 happens

R/plot-build.R
```diff
@@ -189,8 +189,17 @@ ggplot_gtable.ggplot_built <- function(data) {
...
+    if (any(unitType(legend_box$widths) == "sum")) {
+      legend_width  <- unit(1, "npc")
+    } else {
+      legend_width <- gtable_width(legend_box)
+    }
+    if (any(unitType(legend_box$heights) == "sum")) {
+      legend_height <- unit(1, "npc")
+    } else {
+      legend_height <- gtable_height(legend_box)
+    }
```
teunbrand, line 202: This is step 3

thomasp85: Maybe we could make this a bit more explicit? Can we attach an attribute or something to the unit vector when we do a stretch, and then check for that instead?

teunbrand:
What I previously called step 2 and 3 are now all part of `Guides$package_box()`, so there isn't really a need to discern these steps anymore.

R/backports.R
teunbrand: Nevermind, R 4.4.0 is scheduled for spring, so we should support the old unit system until then.

thomasp85 (approved):
LGTM

teunbrand:
I sneaked in a little fix related to an interaction with #5570 I ran into, but I'll merge anyway because it I'm pretty sure the fix can't harm anything.

PR #5518, opened 2023-11-13
Transform `midpoint` argument in relevant scales

### LINKED ISSUE #3198: midpoint is not transformed in scale_color_gradient2
When using `scale_color_gradient2` with `trans = "log10"`, I was expecting the `midpoint` to be transformed as well. The documentation does not specify how `midpoint` behaves under transformation.

# Reproducible example

## Without transformation:

```r
library(ggplot2)
df <- data.frame(x = c(1, 10, 100), y = c(1, 10, 100), color = c(1, 10, 100))
color_midpoint <- 10
ggplot(df) +
  geom_point(aes(x = x, y = y, color = color)) +
  scale_color_gradient2(midpoint = color_midpoint, low = "red", mid = "green", high = "blue")
```
imatge

## With transformation
```r
ggplot(df) +
  geom_point(aes(x = x, y = y, color = color)) +
  scale_color_gradient2(midpoint = log10(color_midpoint), low = "red", 
                        mid = "green", high = "blue", trans = "log10") +
  coord_trans(x = "log10", y = "log10")
```
imatge

I was expecting to use `midpoint = color_midpoint` so the `trans = "log10"` would apply to both the color aesthetic *and* the midpoint. However I had to apply manually `midpoint = log10(color_midpoint)`.

The main advantage of letting the `trans` transform the `midpoint` is that if I change the transformation then the `midpoint` keeps being valid (and consistent with the aesthetic units), while now I have to manually transform the `midpoint` for every transformation I use.

Is is possible to change the behaviour of `midpoint` so it affected by `trans`? If not, is it possible to document this behaviour?

Thanks for your time and splendid work

teunbrand:
This PR aims to fix #3198.

Briefly, it uses the scale transformation in `mid_rescaler()` to transform the `midpoint` argument. This makes it unnecessary to manually transform that value.

Some examples; standard use of midpoint without a transformation:
``` r
#> ℹ Loading ggplot2

df <- data.frame(x = c(1, 10, 100))

p <- ggplot(df, aes(x, x, colour = x)) +
  geom_point()
p + scale_colour_gradient2(midpoint = 10)
```

Midpoint is transformed appropriately:
``` r
p + scale_colour_gradient2(midpoint = 10, trans = "log10")
```

Warning when midpoint is invalid:
``` r
p + scale_colour_gradient2(midpoint = 0, trans = "log10")
#> Warning in scale_colour_gradient2(midpoint = 0, trans = "log10"): log-10
#> transformation introduced infinite values in `midpoint`.
```

<sup>Created on 2023-11-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/scale-.R
```diff
@@ -583,7 +583,7 @@ check_breaks_labels <- function(breaks, labels, call = NULL) {
 
 default_transform <- function(self, x) {
   new_x <- self$trans$transform(x)
-  check_transformation(x, new_x, self$trans$name, self$call)
+  check_transformation(x, new_x, self$trans$name, call = self$call)
```
thomasp85, line 586: needs to be updated to the news `trans` renaming

thomasp85 (approved):
LGTM once updated to new `trans` name

R/scale-.R
teunbrand: Done

PR #5519, opened 2023-11-13
`position_stack()` does not remove missing data

### LINKED ISSUE #3532: na.rm have no effect in geom_col with position_stack
Usually `na.rm = TRUE` will hide the warnings of removing NAs. However when using `position_stack` with `geom_col`, the warning cannot be turned off.

```r
df <- data.frame(x = 1:20, value = c(rep(NA, 10), rep(2, 10)))
# this don't have warnings
ggplot(df, aes(x, value)) +
  geom_point(na.rm = TRUE)

# this have warnings
ggplot(df, aes(x, value)) +
  geom_col(na.rm = TRUE)
# Warning message:
# Removed 10 rows containing missing values (position_stack). 
```

It seemed that the other usage of `remove_missing` in geom_* take the na.rm parameter, but `position_stack` doesn't take it as parameter thus no way to hide the warning. Maybe `position_stack` was supposed to take cleaned data from `geom_col` too.

```r
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

ggplot2_3.2.1
```

teunbrand:
This PR aims to fix #3532.

Instead of removing missing data, `position_stack()` sets all the positions of the missing data as missing. As missing data gets treated as having a height of 0, this shouldn't affect the stack calculations. However, the missing data now gets handled by the geom instead of the position, so the layer's `na.rm` setting kicks in.


tests/testthat/test-geom-col.R
```diff
@@ -7,8 +7,8 @@ test_that("geom_col removes columns with parts outside the plot limits", {
     ggplotGrob(p + ylim(0.5, 4)),
     "Removed 3 rows containing missing values or values outside the scale range"
   )
-  expect_warning( # warning created at build stage
-    ggplot_build(p + ylim(0, 2.5)),
+  expect_warning( # warning created at render stage
+    ggplotGrob(p + ylim(0, 2.5)),
```
teunbrand, line 11: Because geom's NA handling happens after building

tests/testthat/test-scales.R
```diff
@@ -127,7 +131,7 @@ test_that("oob affects position values", {
 
   # Bars depend on limits and oob
   expect_equal(low_censor[[1]]$y, c(0.2, 1))
-  expect_equal(mid_censor[[1]]$y, c(0.5))
+  expect_equal(mid_censor[[1]]$y, numeric(0))
```
teunbrand, line 134: I think previous test wasn't representative of what happens in a plot.
Because all bars have ymin = 0, and get censored at `c(3, 7)` limits, none of the bars are retained.

thomasp85 (approved):
Please add NEWS bullet

PR #5520, opened 2023-11-13
`stat_count()` respects uniqueness of `x`

### LINKED ISSUE #4609: stat_count gives cryptic error when used on a column of doubles
Run this trivial code (csv is attached):
```
library(tidyverse)

data <- read_csv("data.csv")

data %>%
  ggplot(aes(x=Tenure)) +
  geom_bar()
```

You get this warning:
```
Warning message:
Computation failed in `stat_count()`:
Elements must equal the number of rows or 1 
```

This is the resulting plot:
image

The data:
[data.csv](https://github.com/tidyverse/ggplot2/files/7131494/data.csv)

Expected behavior: it should "just work". All the data in this tibble is just doubles. I reviewed the `geom_bar` documentation, and I see no contraindications for this working.

If I have done something wrong here, then this becomes a feature request for a useful error message or improved documentation.

teunbrand:
This PR aims to fix #4609.

Briefly, `as.factor(x)` and `vec_unique(x)` have different definitions of what constitutes a unique value, because the `as.factor(x)` casts `x` as a character, losing any small digits in doubles. This PR uses `rowsum()` instead of `tapply()` to count, which preserves the uniqueness as it doesn't cast the group to a factor (also `rowsum()` is a bit faster).

Reprex from #4609:

``` r
#> ℹ Loading ggplot2
df <- data.frame(x = rep(c(1, 2), 5) + rep(c(0, -2.220446e-16), c(4, 1)))

ggplot(df, aes(x)) + geom_bar()
#> Warning: `position_stack()` requires non-overlapping x intervals
```

<sup>Created on 2023-11-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Of course, `position_stack()` not unreasonably complains about overlapping intervals, but this is a better hint at what might be going on that the current error.


thomasp85 (approved):
LGTM

thomasp85:
Please add NEWS bullet

PR #5529, opened 2023-11-20
Fix colourbar's `ticks.length` argument not working

teunbrand:
I became aware that the `ticks.length` argument of `guide_colourbar()` wasn't working as intended, so this PR fixes that problem.
I've including this argument in an already existing test so that we might know if this breaks again in the future.

Small demonstration of setting the tick length:

``` r
#> ℹ Loading ggplot2

  guides(colour = guide_colourbar(ticks.length = unit(0.5, "npc")))
```

<sup>Created on 2023-11-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/guide-colorbar.R
+++ b/R/guide-colorbar.R
@@ -439,14 +439,11 @@ GuideColourbar <- ggproto(
       "horizontal" = c("bottom", "top"),
       "vertical"   = c("right", "left")
     )
-    elements$ticks_length <- rep(elements$ticks_length, length.out = 2)
-    elem1 <- elem2 <- elements
-    elem1$ticks_length <- elements$ticks_length[2]
-    elem2$ticks_length <- elements$ticks_length[1]
+    ticks_length <- rep(elements$ticks_length, length.out = 2)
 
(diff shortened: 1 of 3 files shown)


thomasp85:
What issue does this solve? Can you elaborate since you don't link to an issue?

teunbrand:
Apologies for not filing an issue first, which I should have.

The issue is that the `ticks.length` argument was ignored. Here is the same reprex on the current main branch, where you can see that the ticks don't follow the length set in the `ticks.length` argument and retain the default `unit(0.2, "npc")` tick length:

``` r
library(ggplot2)
packageVersion("ggplot2") # current main
#> [1] '3.4.4.9000'

  guides(colour = guide_colourbar(ticks.length = unit(0.5, "npc")))
```



<sup>Created on 2023-11-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review!

PR #5534, opened 2023-11-22
Fixes partial matching on guides

eliocamp:
I caught this testing ggnewscale. This is using the latest dev.

``` r
library(ggplot2)
options(warnPartialMatchDollar = TRUE)
  geom_point(aes(colour = factor(year)), size = 5) 
#> Warning in grobs$label: partial match of 'label' to 'labels'

#> Warning in grobs$label: partial match of 'label' to 'labels'
```

I think this is the correct change.


teunbrand (approved):
Nice catch, thanks @eliocamp!

PR #5537, opened 2023-11-22
POC: double dispatch for `ggplot_add()`

### LINKED ISSUE #3986: Missing generic `+` method
(cc @clauswilke  https://github.com/tidyverse/ggplot2/pull/3818#issuecomment-623055651)

## Context

Let's say `foo` is a class defined by a package other than ggplot2 (Ex: `patchwork` or `GGally`). `foo` objects also inherit from `gg`. `bar` objects do not inherit from `gg`

You can currently perform `gg_obj + foo_obj` as both objects inherit from `gg` and would produce the same `+.gg` method from Opt's double dispatch.

Currently, to control how the right-hand-side (`foo` objects) are added, we can define `ggplot_add.foo`. 

## Problem

However, currently we can not customize how left-hand-side objects (ex: `foo_obj + gg_obj`) are added to `gg` objects.  

Ex: left hand side structures are not ggplot2 objects and will not work with ggplot2 internal methods
```r
x <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
foo_obj <- structure(
 list(x, x), 
  class = c("foo", "gg")
)
# can not customize how the theme is added
foo_obj + theme_bw()
```

Ex: conflicting `+` methods make addition impossible
```r
x <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
bar_obj <- structure(
 list(x, x), 
  class = c("bar") # does not inherit from 'gg'
)
`+.bar` <- function(e1, e2) {
  # only add to second plot
  e1[[2]] <- e1[[2]] + e2
  e1
}
# Conflicting `+` methods
bar_obj + theme_bw()
#> Error in bar_obj + theme_bw() : non-numeric argument to binary operator
#> In addition: Warning message:
#> Incompatible methods ("+.bar", "+.gg") for "+" 
```

### Real life situation

* To get around this in `GGally`, I overwrite the `+.gg` method.  Unfortunately, this produces a warning and does not dispatch to the next available plus method.
I am open to other options, but I don't know how to produce one that would play well with other packages.

* I would like to be able to add a scale to a `ggmatrix` object in appropriate places only (Ex: `ggmatrix_obj + gg_scale_obj`). With the current methods, I don't know how I can do this cleanly.

## Goal

Define a generic add method (ex: `add_gg(e1, e2)`).  This generic add method should be called within `+.gg` to avoid double dispatch confusion involving `+`.    

`ggplot2`'s `+.gg` is not commutative. So I believe this situation has merit. 

Using the implementation in https://github.com/tidyverse/ggplot2/pull/3818 , I would be able to define how a theme object is added to a custom object. 
``` r
library(ggplot2)
x <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
foo_obj <- structure(
 list(x, x), 
  class = c("foo", "gg")
)
`add_gg.foo` <- function(e1, e2) {
  # only add to second plot
  e1[[2]] <- e1[[2]] + e2
  e1
}
print.foo <- function(x, ...) {
  gridExtra::grid.arrange(x[[1]], x[[2]], ncol = 2)
}

foo_obj + theme_bw()
```

### LINKED ISSUE #5536: `ggplot_add.list()` uses the `%+%` function instead of the `+` generic.
`ggplot_add.list()` adds each element of the list using `%+%` instead of  `+` here:

https://github.com/tidyverse/ggplot2/blob/ca47270fc504193dcf3a29b89408f59df767668c/R/plot-construction.R#L156-L161

The idea is that `ggplot() + list(e1, e2, e3)` should be the same as `ggplot() + e1 + e2 + e3`, but this is not true if someone (😇) wanted to create a gg subclass with a custom `+`  method. 

I'm not sure why that code uses `%+%` instead of `+`. I modified the code and run tests both ways and everything passes. This logic has been since  https://github.com/tidyverse/ggplot2/commit/3c16f9c84bd5d37beaead9e50407700ad9e0a6df from 2017

Would it be possible to change `ggplot_add.list()` to use the `+` generic? Barring that, would it be possible to make `%+%` an exported generic (same as `+`)?

teunbrand:
This PR explores suggestions from #3818 to use double dispatch to `+` objects to ggplot subclasses. The ultimate goal is to fix #3986 and fix #5536.

Double dispatch is implemented for layers only at the moment, but we can expand that once we agree on the approach, which I'm still happy to change.
It currently borrows from the {vctrs} style of S3 double dispatch, though we could consider using {S7} if we're willing to take on an extra dependency.

Open questions:
* @hadley is this implementation of double dispatch appropriate?
* Should we use double dispatch for *all* currently supported classes? If not, which ones shouldn't use double dispatch?
* Should we go for S7 generics?

Small demo that a custom method is invoked:
``` r
#> ℹ Loading ggplot2

# Making a ggplot subclass
p <- ggplot(mpg, aes(displ, cty))
class(p) <- union("foo", class(p))

# Adding custom method for adding layers
ggplot_add.Layer.foo <- function(object, plot, object_name) {
  print("I'm adding a layer to foo")
  ggplot_add.Layer.default(object, plot, object_name)
}

# Should print text as well as the plot
p + geom_point()
#> [1] "I'm adding a layer to foo"
```

<sup>Created on 2023-11-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


hadley:
I don't think it's a good idea to implement custom double dispatch in ggplot2, and it would instead be better to use S7 (which would require switching the ggplot2 base class to a S7, but that in theory should be easy).

teunbrand:
Thanks for the advice, Hadley! I'll experiment around with S7 for a bit then.

teunbrand:
Because a ggplot object as S3 class might be baked into many extension assumptions, I haven't converted it to an S7 class (yet).
The double dispatch of S7 is sufficient to make the mechanism work, but without being able to call `S7::super()` on an S3 object, the extended methods might get slightly awkward.

``` r
#> ℹ Loading ggplot2
library(S7)

# Making a ggplot subclass
p <- ggplot(mpg, aes(displ, cty))
class(p) <- union("foo", class(p))

# Adding custom method for adding layers
method(ggplot_add, list(object = new_S3_class("Layer"), plot = new_S3_class("foo"))) <-
  function(object, plot, object_name) {
    print("I'm adding a layer to foo")
    
    # This is the awkward part:
    method(ggplot_add, list(object = new_S3_class("Layer"), plot = class_ggplot))(
      object, plot, object_name
    )
  }

# Should print text as well as the plot
p + geom_point()
#> [1] "I'm adding a layer to foo"
```



<sup>Created on 2023-11-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
You should be able to create an S7 class with exactly the same S3 class as previously (since S7 is designed to be backward compatible with S3); i.e. you should be able to make the without affecting extensions (unless they use `NextMethod()` which we should think about at the S7 level too).

R/ggplot2-package.R
```diff
@@ -5,6 +5,7 @@
 #' @import scales grid gtable rlang vctrs
 #' @importFrom glue glue glue_collapse
 #' @importFrom lifecycle deprecated
+#' @importFrom S7 class_any class_list method method<- new_generic new_S3_class S7_dispatch
```
hadley, line 8: I'd be tempted to minimise the imports to make it most clear where S7 is being used (at least for now).

R/plot-construction.R
```diff
@@ -87,49 +87,80 @@ add_ggplot <- function(p, object, objectname) {
...
+  dispatch_args = c("object", "plot"),
+  fun = function(object, plot, object_name) S7_dispatch()
+)
+
+class_ggplot <- new_S3_class("ggplot")
+
+method(
+  ggplot_add,
+  list(object = class_any, plot = class_ggplot)
+) <- function(object, plot, object_name) {
```
hadley, line 101: I think it's easier to read on one line:

```suggestion
method(ggplot_add, list(class_any, class_ggplot)) <- function(object, plot, object_name) {
```

R/plot-construction.R
```diff
@@ -87,49 +87,80 @@ add_ggplot <- function(p, object, objectname) {
...
   cli::cli_abort("Can't add {.var {object_name}} to a {.cls ggplot} object.")
 }
-#' @export
-ggplot_add.NULL <- function(object, plot, object_name) {
+
+method(
+  ggplot_add,
+  list(object = new_S3_class("NULL"), plot = class_ggplot)
```
hadley, line 107: You can just use `NULL` here

R/plot-construction.R
```diff
@@ -87,49 +87,80 @@ add_ggplot <- function(p, object, objectname) {
...
+) <- function(object, plot, object_name) {
   plot
 }
-#' @export
-ggplot_add.data.frame <- function(object, plot, object_name) {
+
+method(
+  ggplot_add,
+  list(object = new_S3_class("data.frame"), plot = class_ggplot)
```
hadley, line 114: `S7::class_data.frame`

R/plot-construction.R
```diff
@@ -87,49 +87,80 @@ add_ggplot <- function(p, object, objectname) {
...
+) <- function(object, plot, object_name) {
   plot$data <- object
   plot
 }
-#' @export
-ggplot_add.function <- function(object, plot, object_name) {
+
+method(
+  ggplot_add,
+  list(object = new_S3_class("function"), plot = class_ggplot)
```
hadley, line 122: `S7::class_function`

R/plot-construction.R
```diff
@@ -87,49 +87,80 @@ add_ggplot <- function(p, object, objectname) {
...
+    "Can't add {.var {object_name}} to a {.cls ggplot} object",
     "i" = "Did you forget to add parentheses, as in {.fn {object_name}}?"
   ))
 }
-#' @export
-ggplot_add.theme <- function(object, plot, object_name) {
+
+method(
+  ggplot_add,
+  list(object = new_S3_class("theme"), plot = class_ggplot)
```
hadley, line 132: I'd define all the custom S3 classes in one place.

R/plot-construction.R
```diff
@@ -87,49 +87,80 @@ add_ggplot <- function(p, object, objectname) {
...
 #' @export
-ggplot_add <- function(object, plot, object_name) {
-  UseMethod("ggplot_add")
-}
-#' @export
-ggplot_add.default <- function(object, plot, object_name) {
+ggplot_add <- new_generic(
+  "ggplot_add",
+  dispatch_args = c("object", "plot"),
+  fun = function(object, plot, object_name) S7_dispatch()
```
hadley, line 93: I'd be tempted to leave `object_name` as an optional argument for the methods (i.e. just rely on the default constructor which will also include `...`, which I'd now consider to be best practice in a generic. OTOH it might be better to make that change in a separate PR, as I'd probably now call this `error_arg`, and I'd expect a matching `error_call` argument in order to apply our current error best practices.

hadley:
Thanks for working on this! I don't know if S7 is ready for use in such a high profile package, but I think it's extremely worthwhile to try it out so we can make changes to S7 if needed.

teunbrand:
Thanks for pointing out all these improvements Hadley! 

>  i.e. you should be able to make the without affecting extensions

I had a shallow attempt at implementing a ggplot class, but the issue I kept running into is that the rest of the codebase always extracts information with `$` instead of `@`. We can of course fix all this for ggplot2 itself, but it would also mean all extensions that extract information directly (mostly those that reimplement `ggplot_build()` or `ggplot_gtable()`) would be affected. 

> (unless they use NextMethod() which we should think about at the S7 level too)

This would be a neat mechanism to have, or more broadly a 'dispatch on the next s3 class'-mechanism.

> I don't know if S7 is ready for use in such a high profile package,

I think it might be good to start using S7 for this double dispatch issue, and leave converting S3 to S7 classes for another day when the need arises.

R/plot-construction.R
teunbrand: This makes sense to me, but some lines became quite long, so I declared the function on a new line. I don't think the tidyverse style guide has explicit guidance on long LHS assignments.

R/plot-construction.R
teunbrand: Fair point, I'll leave this comment unresolved to not forget about this for now.

hadley:
Could you add a `$` method?

R/plot.R
```diff
@@ -101,49 +104,67 @@
 #'     mapping = aes(x = group, y = group_mean), data = group_means_df,
 #'     colour = 'red', size = 3
 #'   )
-ggplot <- function(data = NULL, mapping = aes(), ...,
```
hadley, line 104: Hmmm, this a weird generic. If I was to write it today, I'd probably call it `as_ggplot`, since it's really a coercion function not a constructor. So it might be better to keep the form of this as is, and instead call the S7 constructor inside of `ggplot.default`.  To avoid name clashes, could call it `S7_ggplot` or similar? Does that make sense?

teunbrand:
I've converted this PR to a draft to indicate that it is 'on hold' untill S7 is stable enough.

teunbrand:
Closing this in favour of #6364

PR #5548, opened 2023-11-29
Gradient as colourbar

### LINKED ISSUE #5547: Feature request: colourbar as gradients
The `guide_colourbar(raster = TRUE/FALSE)` argument currently controls whether the bar is drawn as a raster image or series of rectangles. I'd like to propose a 3rd option which uses `grid::linearGradient()` to render the bar. We already have `check_device()` to see if the device is able to render it.

Because `raster` is a boolean, it would probably [need to be deprecated](https://design.tidyverse.org/boolean-strategies.html#how-do-you-remediate-past-mistakes) and a new argument `display = "raster"/"rectangles"/"gradient"` (name up for discussion) could take over that function.

teunbrand:
This PR aims to fix #5547.

Briefly, it adds a new argument to `guide_colourbar()`: `display`, which can be set to `"gradient'` to display the bar as a gradient. 

In some more detail; since the `raster` argument was implemented as a boolean, I'd thought it best to deprecate it and use `display` as an enumeration instead. Possible values for `display` are `"raster"`, which is the same as current `raster = TRUE`, `"rectangles"`, equivalent to `raster = FALSE` and `"gradient"`, the new option. 

When `display = "gradient"`, the `nbins` argument is used for the number of colours set in the gradient, rather than actual bins. As such, we don't need to account for the half bin-size offset for the label placement. The `nbins` number can be a lot less in gradient displays compared to raster/rectangle displays. As such, `nbins` defaults to 15 for gradients.

Prior to R 4.1.0, {grid} hadn't implemented `linearGradient()`, so for older R versions it throws an error saying that `linearGradient` is unavailabe in the R version.

Small demo:

``` r
#> ℹ Loading ggplot2

  guides(colour = guide_colourbar(display = "gradient"))
```

### DIFF
--- a/R/guide-colorbar.R
+++ b/R/guide-colorbar.R
@@ -447,8 +447,7 @@ GuideColourbar <- ggproto(
   },
 
   build_decor = function(decor, grobs, elements, params) {
-
-    if (params$raster) {
+    if (params$display == "raster") {
       image <- switch(
         params$direction,
         "horizontal" = t(decor$colour),
@@ -462,7 +461,7 @@ GuideColourbar <- ggproto(
         gp = gpar(col = NA),
         interpolate = TRUE
       )
-    } else{
+    } else if (params$display == "rectangles") {
       if (params$direction == "horizontal") {
         width  <- elements$key.width / nrow(decor)
         height <- elements$key.height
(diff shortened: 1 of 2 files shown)


teunbrand:
TODO: double check how transformations are affected

R/guide-colorbar.R
```diff
@@ -133,8 +141,9 @@ guide_colourbar <- function(
   # bar
   barwidth = NULL,
   barheight = NULL,
-  nbin = 300,
-  raster = TRUE,
+  nbin = NULL,
+  display = "raster",
+  raster = lifecycle::deprecated(),
```
thomasp85, line 146: You can remove `lifecycle::` from here

R/guide-colorbar.R
```diff
@@ -158,6 +167,14 @@ guide_colourbar <- function(
   available_aes = c("colour", "color", "fill"),
   ...
 ) {
+  if (lifecycle::is_present(raster)) {
+    deprecate_soft0("3.5.0", "guide_colourbar(raster)", "guide_colourbar(display)")
+    check_bool(raster)
+    display <- if (raster) "raster" else "rectangles"
+  }
+  display <- match.arg(display, c("raster", "rectangles", "gradient"))
```
thomasp85, line 175: please use rlang `arg_match0`

thomasp85 (changes requested):

R/guide-colorbar.R
teunbrand: Indeed, thanks!

R/guide-colorbar.R
teunbrand: Good catch

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review!

PR #5550, opened 2023-11-29
Colourbar transparency

### LINKED ISSUE #5085: Adding `override.aes()` to `guide_colorbar()`
It seems like `guide_colorbar()` doesn't support `override.aes()` like its other `guide_*()` counterparts. As an example:

```r
library(tidyverse)
mtcars %>% 
  ggplot(aes(x = mpg, y = disp)) +
  geom_point(aes(color=disp), alpha = 0.5) +
  # Below line has no effect because guide_colorbar doesn't support override.aes
  guides(color = guide_colorbar(override.aes = list(alpha = 0.5)))

# Using guide_legend works, but converts to incorrect binned legend instead of colorbar legend
mtcars %>% 
  ggplot(aes(x = mpg, y = disp)) +
  geom_point(aes(color=disp), alpha = 1) +
  guides(color = guide_legend(override.aes = list(alpha = 0.5)))
```

I would like to adjust the colorbar legend to also have `alpha = 0.5` for consistency but I can't seem to find a good way to do this without using `override.aes()`. I think having `override.aes()` available in `guide_colorbar` would be useful in situations like this to maintain consistency across the plot. 

As another example use case, I often make choropleth maps with continuous colorbar legends using `geom_sf` and often fiddle with the `alpha` transparency of some layers to make the maps more readable . It would be nice to have the colorbar legend reflect these changes. Totally understand if this is very complex, but just thought I'd ask!

teunbrand:
This PR aims to fix #5085.

It adds an `alpha` argument to `guide_colourbar()` and `guide_coloursteps()` that control the transparency of the bar.
The alpha is applied in the `extract_decor()` stage, on only affects the display of the bar, not the colour values encoded in the keys (which is fine).

Small demo:

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(displ, hwy, colour = cty, fill = year)) +
  geom_point(shape = 21) +
  guides(
    colour = guide_colourbar(alpha = 0.7),
    fill   = guide_coloursteps(alpha = 0.3)
  )
```


thomasp85 (approved):
LGTM

PR #5551, opened 2023-11-29
Key background inherits panel background

### LINKED ISSUE #5549: Should `legend.key` inherit from `panel.background`?
One thing I often forget when I change the panel background is to also change the legend key accordingly.
I'd like to propose that the following code:

``` r
library(ggplot2)

  geom_point(aes(shape = factor(cyl))) +
  theme(
    panel.background = element_rect(fill = "cornsilk")
  )
```

Would produce the same plot as this:
``` r
  geom_point(aes(shape = factor(cyl))) +
  theme(
    panel.background = element_rect(fill = "cornsilk"),
    legend.key = element_rect(fill = "cornsilk")
  )
```

It would be a convenience if the legend key inherited from the panel background, to keep these in sync automatically.
To make this work as intended, the themes should probably omit the legend key definitions. These currently have a slightly different fill (`"grey95"`) than the background (`"grey92"`), but I think the difference would be barely noticible to a human eye.

teunbrand:
This PR aims to fix #5549.

Briefly, it sets `legend.key` to inherit from `panel.background` and eliminates the `legend.key` element from the built-in themes.
This causes the key backgrounds to automatically adhere to the panel background style.

``` r
#> ℹ Loading ggplot2

  geom_point(aes(shape = factor(cyl))) +
  theme(
    panel.background = element_rect(fill = "cornsilk")
  )
```

For many built-in themes, this was already achieved by encoding the `legend.key` element and `panel.background` seperately in the same style. With this PR, this doesn't need to happen seperately anymore.

However, because the default `theme_gray()` had a slightly lighter `legend.key` than `panel.background`, there is a noticible (but not very obvious) visual change in this theme. See attached example where the left part is the current default and the right part is the default with this PR:

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # ggplot2 (development version)
 
+* The `legend.key` theme element is set to inherit from the `panel.background`
+  theme element. The default themes no longer set the `legend.key` element.
+  This causes a visual change with the default `theme_gray()` (#5549).
+
 * Lines where `linewidth = NA` are now dropped in `geom_sf()` (#5204).
 
 * New `guide_axis_logticks()` can be used to draw logarithmic tick marks as
--- a/R/theme-defaults.R
+++ b/R/theme-defaults.R
@@ -172,7 +172,7 @@ theme_grey <- function(base_size = 11, base_family = "",
     legend.spacing.x =    NULL,
     legend.spacing.y =    NULL,
     legend.margin =      margin(half_line, half_line, half_line, half_line),
-    legend.key =         element_rect(fill = "grey95", colour = NA),
+    legend.key =         NULL,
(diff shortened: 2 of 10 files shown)


clauswilke (approved):
Surprisingly large patch for such a small change. Great idea and looks good to me.

Now looking at the old and new outputs side by side, I'd say the old one (with slightly lighter background in the legend) was wrong. You're coloring a smaller surface area, so if at all it should be darker not lighter. But matching is perfectly fine.

teunbrand:
Thanks for the review and kind words Claus! I agree with your assessment about surface area vs intensity, which would be easier to implement if we could do something like `element_rect(colour = ~colorspace::darken(.x, 0.2))` to adjust settings on the fly.

PR #5554, opened 2023-12-01
Refactor guide styling

### LINKED ISSUE #5348: Feature request: `(sub)theme` argument for `guide_legend()` and friends

### LINKED ISSUE #5539: Feature request: `legend.key.spacing` argument(s) added to `theme`
The new reworked legend spacing will allow the spacing between legend key/label elements to be controlled with `guide_legend(key.spacing = ...)`

It'd be great if you could allow this to be modified through the theme argument as well.

Otherwise, users cannot incorporate their desired legend key spacing up in a theme function.  They have to instead use the guides each time they make a plot to adjust it how they want. 

Thanks, and great work with the ggplot2 development!

### LINKED ISSUE #5540: Feature request: `legend.title.position` argument added to `theme`
Would the title.position argument from the `guide_legend`/`guide_colourbar`/`guide_coloursteps` functionality be able to be added as an argument (or arguments) to theme?

I always like the all my legends to have the title on top, so it'd be super helpful if I could do this via the theme

### LINKED ISSUE #5541: Feature request: `legend.byrow` argument added to `theme`
I think people tend to have a preference for one way or another for all of their plots, so it'd be great if you could add it into your theme once

teunbrand:
This PR aims to fix #5348 and also fixes #5539, fixes #5540 and fixes #5541 and implements https://github.com/tidyverse/ggplot2/pull/5473#issuecomment-1827661999.

Briefly, it is quite a beefy PR that adds a `theme` argument to all `guide_*()` functions and adds guide style elements to `theme()`.

Less briefly; I could not find a satisfactory way to simplify guide styling without breaking some backward compatibility. I attempted  this in #5512, but that PR added as much complexity as it took away. The crux of the PR is this:

* The `theme()` function has new arguments for styling guides, like `legend.frame`, `legend.ticks` etc.
* `guide_*()` functions no longer have style arguments, like `frame`, `ticks`, etc.
* To style individual guides, one can now use `guide_*(theme = theme(...))`

This allows one to change the look of an individual guide with, for example, a preset theme:

``` r
#> ℹ Loading ggplot2

  geom_point(aes(fill = cyl, colour = cty), shape = 21)

p + guides(fill = guide_legend(theme = theme_dark()))
```

In addition, all guide style options are now settable from the plot's theme:

``` r
red_frame <- theme(legend.frame = element_rect(colour = "red"))

p + red_frame
```

While retaining flexibility in individual guides.

``` r
p + guides(colour = guide_colourbar(theme = red_frame))
```

The 'old' options still work, but are passed through the `...` argument and are undocumented.

``` r
p + guides(colour = guide_colourbar(frame.colour = "red"))
```

Also, the 'old' options are soft-deprecated so developers will start seeing warnings about them:

``` r
withr::with_options(
  list(lifecycle_verbosity = "error"),
  guide_colourbar(frame.colour = "red")
)
#> Error:
#> ! The `frame.colour` argument of `guide_colourbar()` is deprecated as of
#>   ggplot2 3.5.0.
#> ℹ Please use the `theme` argument instead.
```

<sup>Created on 2023-12-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Some additional notes:

* `theme()` is already widely used to style non-data elements of the plot, so the way to style guides should be familiar.
* The PR greatly reduced [argument clutter](https://design.tidyverse.org/argument-clutter.html) in guides.
* There is now harmony in which style element can be set from the guide and which element from the theme.
* The inheritance logic is now clear: it is `(plot_theme %||% guide_defaults) + guide_theme` (with `%||%` applied element-wise, and `+` meaning `add_theme()`). This should allow for easier maintenance of the guides, as well as make it easier to extend guides.
* Some legend styling options don't have plot-level defaults as different guides had different defaults that I preserved. For example, `guide_colourbar()` has white ticks by default, `guide_bins()` has black ticks by default and `guide_coloursteps` has hidden ticks by default.
* We lose the 'expressions are right-aligned by default' heuristic. This is only a 'con' if you rely on this behaviour, it is less of a con if you, like me and others (#4814) are surprised by this behaviour. Removing this heuristic is a net positive I think, as it is now far more straightforward to derive the horizontal justification for legend labels.
* The `legend.key.spacing` is now set to the half the theme's `base_size`, which already is the default in 99% of legends except in the 1% where the legend doesn't have a title. Previously, it was half the title size, and if that was missing half the label size and if that was missing 5.5 points. This resulted in 1 visual snapshot without a legend title changing slightly. The additional text margins introduced in #5456 also derive from this setting for symmetry reasons.
* A downside is that it is somewhat more verbose to set guide-level themes, but I think it is worth it overall.
* If we ever want to get rid of guide code gnarlyness, now is *the* time to do it.


R/guide-.R
```diff
@@ -274,6 +279,7 @@ Guide <- ggproto(
   # Converts the `elements` field to proper elements to be accepted by
   # `element_grob()`. String-interpolates aesthetic/position dependent elements.
   setup_elements = function(params, elements, theme) {
```
teunbrand, line 281: This is crucial step

R/utilities.R
```diff
@@ -710,6 +710,17 @@ vec_rbind0 <- function(..., .error_call = current_env(), .call = caller_env()) {
   )
 }
 
+replace_null <- function(list, ..., env = caller_env()) {
+  # Collect dots without evaluating
+  dots <- match.call(replace_null, expand.dots = FALSE)$`...`
```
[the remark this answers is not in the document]
teunbrand: There is `rlang::call_match()`, but it calls `match.call()` under the hood. I had been assured this was an [idiomatic way](https://stackoverflow.com/a/13355505/11374827) of doing this.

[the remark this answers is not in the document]
thomasp85: `rlang::enexprs(...)` does it all in one fell swoop

R/utilities.R
```diff
@@ -773,9 +773,17 @@ vec_rbind0 <- function(..., .error_call = current_env(), .call = caller_env()) {
...
 
+# This function is used to vectorise the following pattern:
+#
+# list$name1 <- list$name1 %||% value
+# list$name2 <- list$name2 %||% value
+#
+# and express this pattern as:
+#
+# replace_null(list, name1 = value, name2 = value)
 replace_null <- function(list, ..., env = caller_env()) {
```
thomasp85, line 784: Maybe `list` is not the best argument name as it masks a very common/basic function name

thomasp85 (changes requested):

R/utilities.R
teunbrand: Agreed

thomasp85:
[a14493b](https://github.com/tidyverse/ggplot2/pull/5554/commits/a14493b0ff70ec8802deba1c01aeb72ad41b0717) is about the guide stretching stuff, right?

teunbrand:
And long title placement, yes

thomasp85 (approved):
LGTM

teunbrand:
Thanks Thomas!

PR #5559, opened 2023-12-03
`ggsave()` forces `plot` only after setting the current device

krlmlr:
Some adjustments to the plot object might require measurements, hence an active device.

## With this PR

``` r
library(ggplot2)

print_active_device <- function(p) {
  print(dev.cur())
  p
}

plot <-
  ggplot(data = data.frame(x = 1:3, y = 1:3), aes(x = x, y = y)) +
  geom_point()

ggsave(tempfile(fileext = ".pdf"), print_active_device(plot))
#> Saving 7 x 5 in image
#> pdf 
#>   3
```

<sup>Created on 2023-12-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## Current main

``` r
library(ggplot2)

print_active_device <- function(p) {
  print(dev.cur())
  p
}

plot <-
  ggplot(data = data.frame(x = 1:3, y = 1:3), aes(x = x, y = y)) +
  geom_point()

ggsave(tempfile(fileext = ".pdf"), print_active_device(plot))
#> Saving 7 x 5 in image
#> quartz_off_screen 
#>                 2
```

<sup>Created on 2023-12-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
Hi Kirill, I'm not quite sure I'm familiar with the context in which text must be measured on the intended device, would you mind expanding a bit on this? There are also two notes I have with this PR: (1) wouldn't this mismatch the device background with the plot's background fill colour? And (2) if reasonably possible it seems desirable that the `on.exit()` code is declared as soon as the device is activate. In theory, now something could go wrong in `plot_theme()` and the `on.exit()` code isn't executed.

krlmlr:
Thanks, good catch. Fixed the `on.exit()` call.

In the code I'm dealing with, `print_active_device()` calls `ggplotGrob()` and applies some transformation on the resulting object. Printing that object is the same as printing the plot, but `ggplotGrob()` needs an active device.

Sketch:

```r
print_active_device <- function(p) {
  # Measurement may happen here
  g <- ggplotGrob(p)
  my_transform_grob(g)
}
```

Is there a better way to do this? Hooks registered on the ggplot2 object, maybe?

thomasp85:
@krlmlr Do you still want this included after the discussion?

krlmlr:
It's a catch-22 that can't be resolved because `bg` is used in the `dev()` call.

PR #5566, opened 2023-12-04
Rename `trans` arguments to `transform`

### LINKED ISSUE #5558: Should the `trans` argument in `scale_*_continuous` etc be renamed to `transform`?
Given that the scales functions will be changing from `scales::*_trans` to `scales::transform_*`

teunbrand:
This PR aims to fix #5558.

Briefly, we follow in {scales}'s footsteps by renaming the `trans` argument to `transform`.
This mostly affects scale functions and the secondary axis functionality.


R/scale-.R
```diff
@@ -94,13 +96,18 @@ continuous_scale <- function(aesthetics, scale_name = deprecated(), palette, nam
...
                              call = caller_call(),
                              super = ScaleContinuous) {
   call <- call %||% current_call()
   if (lifecycle::is_present(scale_name)) {
     deprecate_soft0("3.5.0", "continuous_scale(scale_name)")
   }
+  if (lifecycle::is_present(trans)) {
+    deprecate_soft0("3.5.0", "continuous_scale(trans)", "continuous_scale(transform)")
+    transform <- trans
+  }
```
teunbrand, line 110: This is the crux of the PR. If `trans` is used, we assume they meant `transform` instead.

teunbrand:
Yes good call, thanks for catching that!

thomasp85:
yeah, that was kind of the cases we were worried about - let's keep the field name as is

R/scale-.R
```diff
@@ -668,17 +682,17 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     if (self$is_empty()) {
       return(numeric())
     }
-
+    transformer <- self$transformer
```
teunbrand, line 685: In addition to soft-deprecating the `trans` field, we may also change these types of statements to `transformer <- self$transformer %||% self$trans`.

thomasp85: yeah, that is a good idea.

I'm somewhat disliking `transformer` (I know I brought up that name). as it is so close to `transform`... Maybe something like `transform_obj` or something similar

teunbrand: I think `transformation` might also be good. Do you have a strong preference?

thomasp85: no that's good as well

R/axis-secondary.R
```diff
@@ -236,8 +236,8 @@ AxisSecondary <- ggproto("AxisSecondary", NULL,
 
     # patch for date and datetime scales just to maintain functionality
     # works only for linear secondary transforms that respect the time or date transform
-    if (transformer$name %in% c("date", "time")) {
-      temp_scale <- self$create_scale(new_range, trans = transformer)
+    if (transformation$name %in% c("date", "time")) {
+      temp_scale <- self$create_scale(new_range, trans = transformation)
```
thomasp85, line 240: We should probably change the name of the `trans` argument here as well, no?

teunbrand: Yeah might as well do that too

thomasp85:
Just to be sure, have you tested this branch against a scale extension that uses `trans`?

teunbrand:
Yeah I tested it against an extension and `trans` gets funneled to `transform` when used with a proper constructor. I don't think the {tsibble} example I linked to earlier works as it should. Because `scale_x_date()` is used to construct the ggproto object that is inherited from, the `transformation` field is populated and `transformation %||% trans` thus returns the `transformation` field.

teunbrand:
Should now be better backwards compatible:

``` r
#> ℹ Loading ggplot2

ggplot(economics[1:20,], aes(date, unemploy)) +
  geom_line() +
  tsibble::scale_x_yearquarter()
#> Warning: Scale$trans was deprecated in ggplot2 3.5.0.
#> ℹ Please use Scale$transformation instead.
#> This warning is displayed once every 8 hours.
#> generated.
```



<sup>Created on 2023-12-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review Thomas!

PR #5567, opened 2023-12-06
`check_device()`: Early exit for null device

teunbrand:
This PR aims to implement https://github.com/tidyverse/ggplot2/pull/5299#issuecomment-1842747716.

Briefly, if there is no device open, the 'device' is the null device. When that is the device name, `maybe` is returned early.


thomasp85 (approved):
LGTM

PR #5569, opened 2023-12-07
Change minor breaks interface

### LINKED ISSUE #3583: More flexibility for minor_breaks

teunbrand:
This PR aims to fix #3583.

Briefly, it reimplements #3591 and ensures major breaks are not censored before they arrive at the minor breaks calculation. This allows minor break functions to use out-of-bounds breaks to extrapolate towards the limits.

Major breaks are now censored at two other other locations:
* The view scales censor breaks to ensure that coords receive proper panel grid coordinates.
* The guides now censor breaks.

At first glance, it may seem redudant to censor both in the view scales and in the guides. For axis guides, that is certainly the case. However, non-position guides do not receive view scales, but the original scales, so these would need censoring.

[Originally](https://github.com/tidyverse/ggplot2/pull/3591#issuecomment-1782769789), I thought that this would be much more invasive, but it turns out we need just these two places to apply censoring.

To demonstrate, we can define a function that uses the major breaks in minor break calculations. Notice that the minor gridlines go outside the 20-40 breaks interval, because the original break calculation includes 10 and 50.

``` r
#> ℹ Loading ggplot2

minor_between <- function(n = 5) {
  force(n)
  function(limits, breaks) {
    minor  <- approx(
      x = seq_along(breaks),
      y = breaks,
      xout = seq(1, length(breaks), by = 1 / (n + 1))
    )$y
    oob_discard(minor, limits)
  }
}

  geom_point() +
  scale_y_continuous(minor_breaks = minor_between(3))
```

There are some caveats of course. In the plot below, the x-axis does not have a 10^-3 break, so the gridlines leading up to 10^-2 are missing. Also for a log axis, it doesn't do a great job when breaks are missing. This is illustrated below for x = 10^0, but it may also happen that the default log breaks go by steps of 100x instead of 10x when the range is large.

``` r
ggplot(msleep, aes(bodywt, brainwt)) +
  geom_point(na.rm = TRUE) +
  scale_y_log10(
    minor_breaks = minor_between(9)
  ) +
  scale_x_log10(
    breaks = 10^setdiff(-2:4, 0),
    minor_breaks = minor_between(9)
  ) +
  coord_equal()
```

<sup>Created on 2023-12-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

teunbrand:
Thanks for the review!

PR #5573, opened 2023-12-08
Add Teun as author

thomasp85:
Hi Team

Teun has been instrumental in many of the great things that has happened with ggplot2 in the last year or so. The next feature release is predominantly his work. I see it only fitting to include him as author together with the rest of us. Please approve this PR to show your support or discuss below

There will be a 3 weeks deadline beyond which silence is considered consent

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,6 +18,8 @@ Authors@R: c(
            comment = c(ORCID = "0000-0002-3385-7233")),
     person("Dewey", "Dunnington", role = "aut",
            comment = c(ORCID = "0000-0002-9415-4582")),
+    person("Teun", "van den Brand", role = "aut",
+           comment = c(ORCID = "0000-0002-9335-7468"))
     person("Posit, PBC", role = c("cph", "fnd"))
   )
 Description: A system for 'declaratively' creating graphics, based on "The
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ Authors@R: c(
     person("Dewey", "Dunnington", role = "aut",
            comment = c(ORCID = "0000-0002-9415-4582")),
     person("Teun", "van den Brand", role = "aut",
-           comment = c(ORCID = "0000-0002-9335-7468"))
+           comment = c(ORCID = "0000-0002-9335-7468")),
     person("Posit, PBC", role = c("cph", "fnd"))
   )
 Description: A system for 'declaratively' creating graphics, based on "The
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -44,6 +44,7 @@ The core developers of ggplot2 are:
 * [Kara Woo](https://github.com/karawoo)
 * [Hiroaki Yutani](https://github.com/yutannihilation)
 * [Dewey Dunnington](https://github.com/paleolimbot)
+* [Teun van den Brand](https://github.com/teunbrand)
 
 All core developers are bound by the [code of conduct](CODE_OF_CONDUCT.md).
 


DESCRIPTION
```diff
@@ -18,6 +18,8 @@ Authors@R: c(
            comment = c(ORCID = "0000-0002-3385-7233")),
     person("Dewey", "Dunnington", role = "aut",
            comment = c(ORCID = "0000-0002-9415-4582")),
+    person("Teun", "van den Brand", role = "aut",
+           comment = c(ORCID = "0000-0002-9335-7468"))
```
teunbrand, line 22:
```suggestion
           comment = c(ORCID = "0000-0002-9335-7468")),
```

yutannihilation (changes requested):
Definitely agree, too! Thank you so much Teun for your great contribution. I didn't expect the development of ggplot2 would make such a wonderful progress in the recent months!

@thomasp85 @hadley 
I have one request to you two before this gets merged. Could you update [GOVERNANCE.md](https://github.com/tidyverse/ggplot2/blob/main/GOVERNANCE.md) to reflect the current status? I don't think it describes what the governance is in actual. For example,

* The current "project lead" is Thomas? I think Thomas has the permission to push to the main branch directly.
* In my understanding, Teun is not yet a "core developer", but he already has the same permission to approve pull requests and close issues. I don't think the document describes in what case the permission is granted to non-"core developer".
* ~~Not all the "core developers" are invited in this approval process.~~ (Since this issue is public and the other core developers can react if they have some different opinion, I now feel this point is fine, sorry)

To be clear, it's not that I'm blaming the violation of what is currently defined as the rule. I'm totally fine with it if it's comfortable to you. I think this divergence probably means the current rule is a bit too formal to follow. So, it's time to loosen the rule. Probably, just deleting `GOVERNANCE.md` is fine.

thomasp85:
Thanks for raising this @yutannihilation - I agree we have been too lax compared to what we have laid out in governance

hadley:
I'd say that I am still the project lead and Teun is a core developer (i.e. this PR should add him to that bullet list too)

paleolimbot (approved):
+10!

It's been a joy to see all the issues I touched years ago slowly but surely get closed over the last few months. Cheers!

clauswilke (approved):
Happy to support this PR!

yutannihilation:
@hadley 
Thanks. Then, I think the document should

* define a new role for what Thomas currently is. Apparently, he has more power than the normal "core developers". If it's not the "project lead", I think a new role is needed.
* describe that the project lead can add a "core developer" without getting confirmation from other "core developers". (IMHO, it would be cleaner to define a new role that is not a "core developer" but is temporarily granted the same permission as "core developer" at the "project lead"'s direction. But, probably this is just a matter of preference)

hadley:
Why do you think Thomas has more power than other devs?

thomasp85:
@yutannihilation as to the second point I made an error there, giving Teun GitHub permissions way before we started the Core Developer inclusion. That's on me and won't happen again. I think the process as described in the governance file is as it should and I should have paid more attention to it

thomasp85:
I don't think anyone takes your comments in any kind of bad tone - I think it is good that we keep ourself up to the standards that we put forth ❤️. 

I think going forward if a new intern/contractor were to be had, we would either start the process earlier (if they had a proven track record already) or wait giving them GitHub permission until such a track record was established and then start the right process

thomasp85:
I'm admin by side effect through my job, my permissions on this project specifically is the same as everyone else

yutannihilation:
> I don't think anyone takes your comments in any kind of bad tone - I think it is good that we keep ourself up to the standards that we put forth ❤️.

Thanks!

After some thinking, I'm now convinced it's the best to stick with the current [GOVERNANCE.md](https://github.com/tidyverse/ggplot2/blob/main/GOVERNANCE.md) if following the process is not inconvenient to you. While I still feel there's some room for update, the document is the result of considerable amount of efforts (https://github.com/tidyverse/ggplot2/pull/2959) and I should respect it.

As my concern is mostly resolved, I'm approving. Thanks for your clarification.

yutannihilation (approved):

teunbrand:
Thanks everyone for the kind words and @yutannihilation vigilance on the governance document! ❤️

yutannihilation:
Congratulations!

Issue #5077, opened 2022-12-02
create_layout as a S3 generic


jtlandis:
I am curious if the developers of `ggplot2` would be open to making a `create_layout` S3 generic, that takes one parameter `plot`.

It may look something like the following 

```r
create_layout <- function(plot) UseMethod("create_layout", plot)
create_layout.ggplot <- function(plot) {
    ggproto(NULL, Layout, facet = plot$facet, plot$coordinates)
}
```

The main motivation is to allow for users to rely on `ggplot2`'s own `plot_build` method without having to rewrite the entire method just to change the Layout's functionality slightly. I see this was briefly up for debate before. As a result of #2527, only the `Layer` ggproto class was exported, but `create_layout` was not.

A specific use case I would like to build is to create a Layout that does not assume all panels have the same scales type. Many places in the code, `ggplot2` assumes all y positional scales share the same aesthetics so you see the following `self$panel_scales_y[[1]]$aesthetics` whereas my implementation may look something like `unique(unlist(lapply(self$panel_scales_y, function(s) s$aesthetics)))`

jtlandis:
Would the developers consider adding this in the next release of `ggplot2`?

It feels like a minor change to implement.

teunbrand:
I'm not opposed to this. Do you feel like setting up a PR?

teunbrand:
I can consider this issue fixed by #5576.

PR #5575, opened 2023-12-08
`create_layout` as an S3 generic

jtlandis:
Creates a pull request for #5077

gist: `create_layout` as an S3 generic will help package developers create their own Layout without having to write `ggplot_build` method.

Some testthat files needed to be changed to pass all checks.


jtlandis:
commits 11e420b and 9bdf9c0 should not have been included in this request. If the developers would prefer, I can make changes from a clean start.

PR #5562, opened 2023-12-04
Size of text/label keys

teunbrand:
This PR aims to fix #5561.

Briefly, it uses the mechanism introduced in #5465 to communicate the proper width and height of the grobs.
As a detail, since the labelGrob uses the rotation through the viewport, size has to be calculated after manual rotation.

Reprex from #5561:

``` r
devtools::load_all("~/packages/ggplot2")
#> ℹ Loading ggplot2

ggplot(mpg, aes(displ, hwy, label = class, colour = class)) +
  geom_text(angle = 45) +
  discrete_scale(
    "label", palette = identity_pal(), super = ScaleDiscreteIdentity
  )
```

``` r

ggplot(mpg, aes(displ, hwy, label = class, colour = class)) +
  geom_label(angle = -45) +
  discrete_scale(
    "label", palette = identity_pal(), super = ScaleDiscreteIdentity
  )
```

<sup>Created on 2023-12-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #5587, opened 2023-12-15
Set defaults in text based keys

teunbrand:
This PR fixes a bug introduced in #5562.

Briefly, I forgot to set proper defaults for some values that would make `draw_key_text()` and `draw_key_label()` incompatible with non-text geoms. In this PR, these defaults are set, so you can mix the text key with a point layer, for example.

``` r
#> ℹ Loading ggplot2

  geom_point(key_glyph = "text")
```

``` r
  geom_point(key_glyph = "label")
```

<sup>Created on 2023-12-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/legend-draw.R
+++ b/R/legend-draw.R
@@ -258,7 +258,10 @@ draw_key_smooth <- function(data, params, size) {
 #' @export
 #' @rdname draw_key
 draw_key_text <- function(data, params, size) {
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

teunbrand:
Thanks Thomas!

PR #5594, opened 2023-12-18
Fix note

teunbrand:
This PR to the RC branch aims to fix the following note:

> ❯ checking for portable file names ... NOTE
  Found the following non-portable file path:
    ggplot2/tests/testthat/_snaps/guides/white-to-red-semitransparent-colorbar-long-thick-black-ticks-green-frame.svg  
  Tarballs are only required to store paths of up to 100 bytes and cannot
  store those of more than 256 bytes, with restrictions including to 100
  bytes for the final component.

It does so by swapping the long title added by vdiffr to the subtitle and giving a shorter description as title.


thomasp85 (approved):
LGTM

teunbrand:
Thanks!

PR #5595, opened 2023-12-18
Ingest plain patterns as lists in `layer()`

### LINKED ISSUE #5593: Bug: patterns in legend keys
When providing a plain pattern as a fixed 'fill' aesthetic, the fill is not transferred to the legend key fill.

``` r
library(scales)
library(ggplot2)
packageVersion("ggplot2")
#> [1] '3.4.4.9000'

grad <- grid::linearGradient(pal_viridis()(10))

p <- ggplot(mpg, aes(factor(cyl), colour = "fixed"))

p + geom_bar(fill = grad)
```

If we instead provide the pattern as a list, it works fine.

``` r
p + geom_bar(fill = list(grad))
```

<sup>Created on 2023-12-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The culprit seems to be the following line:

https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/R/guide-legend.R#L233

Where only params of length 1 are included in the legend. Since a pattern is a list, it does not have length 1.

teunbrand:
This PR to the RC branch aims to fix #5593.

Briefly, instead of coercing a plain pattern to a list in `Geom$use_defaults()`, it does so in the `layer()` function.
This has two consequences:
1.  Other internals, like `GuideLegend$get_layer_key()` or `Geom$use_defaults()` don't have to deal with plain patterns.
2. If a pattern is provided to another aesthetic than 'fill', it is also coerced to a list, opening up 'fill'-variant aesthetics for patterns. For example, a package might use a 'fill_box' aesthetic for a geom.

Reprex from #5593:

``` r
#> ℹ Loading ggplot2

grad <- grid::linearGradient(pal_viridis()(10))

ggplot(mpg, aes(factor(cyl), colour = "fixed")) +
  geom_bar(fill = grad)
```

<sup>Created on 2023-12-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/geom-.R
+++ b/R/geom-.R
@@ -126,9 +126,6 @@ Geom <- ggproto("Geom",
       deprecate_soft0("3.4.0", I("Using the `size` aesthetic in this geom"), I("`linewidth` in the `default_aes` field and elsewhere"))
       default_aes$linewidth <- default_aes$size
     }
-    if (is_pattern(params$fill)) {
-      params$fill <- list(params$fill)
-    }
 
     # Fill in missing aesthetics with their defaults
     missing_aes <- setdiff(names(default_aes), names(data))
--- a/R/layer.R
+++ b/R/layer.R
@@ -124,6 +124,12 @@ layer <- function(geom = NULL, stat = NULL,
 
   all <- c(geom$parameters(TRUE), stat$parameters(TRUE), geom$aesthetics())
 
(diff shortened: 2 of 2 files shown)


thomasp85 (approved):
LGTM

teunbrand:
Thanks!

PR #5598, opened 2023-12-18
Bugfix `guide_axis_theta()` label size

teunbrand:
This PR aims to fix a bug in GuideAxisTheta extensibility.

Briefly: measuring the grob directly makes more sense than measuring the grob's children when a theta axis is at a class left/right/top/bottom position.

Less briefly: at the time I wrote these lines of code, the labels were a grobTree with individual labels packed into its children. Since then, the label construction has been vectorised in d7e9a2dae2e77f6545a7de6edaaf005587bd599f, and I forgot to adjust the axis measurement code. The problem I ran into is that the code expects that `$children` is not NULL, whereas they might be. In addition, this brings the `guide_axis_theta()` behaviour closer to `guide_axis()` for non-polar coordinates in terms of spacing.

### DIFF
--- a/R/guide-axis-theta.R
+++ b/R/guide-axis-theta.R
@@ -327,7 +327,7 @@ GuideAxisTheta <- ggproto(
     if (params$position %in% c("top", "bottom")) {
       height <- sum(
         elements$offset,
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5607, opened 2023-12-21
Missing `Scale$get_transformation()`

teunbrand:
This PR to the RC aims to fix a bug uncovered affecting 1 package.

The bug occurs in:
https://github.com/tidyverse/ggplot2/blob/rc/3.5.0/revdep/problems.md#newly-broken-219

Briefly, I forgot a few places to use `scale$get_transformation()` instead of directly getting `scale$transformation`.
Also, in the rare case that a scale has a non-default transformation we should only backtransform if there is an actual inverse function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,103 +1,80 @@
 # ggplot2 (development version)
 
-* The `get_guide_data()` function can be used to extract position and label
-  information from the plot (#5004).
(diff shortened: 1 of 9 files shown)


thomasp85 (approved):
LGTM

teunbrand:
Thanks Thomas

PR #5620, opened 2024-01-02
Discrete secondary axes

### LINKED ISSUE #3171: support sec.axis for scale_discrete
`scale_y_discrete()` and `scale_x_discrete()` currently appear not to support the `sec.axis` argument of their `scale_continuous()` brethren.

`sec.axis` would be nice for discrete scales as well, such as when you have various ways to refer to factor levels (short and long) and the like.

teunbrand:
This PR aims to fix #3171.

Briefly, it adds a `sec.axis` option for `scale_{x/y}_discrete()`. 

For context:
The secondary axis system is not ideal. We have tried to move that functionality to guides in #5410, but it becomes too complicated. Because it doesn't seem viable to handle this functionality in guides, I think it is time to implement secondary axes for discrete scales.

In some more detail:
It currently only accepts identity transformations, as it just doesn't make sense to me why one would transform a discrete axis.
This also means that `dup_axis()` should be preferred to `sec_axis()` for discrete scales.
Breaks can be given as numeric that can take arbitrary positions, or as discrete breaks that can take one of the limits values.
For discrete scales, breaks are mapped early on in `AxisSecondary$break_info()`.

An example plot with relabelling the x-axis and numeric positions in the y-axis:
``` r
#> ℹ Loading ggplot2

data <- data.frame(x = c("A", "B", "C"), y = c("D", "E", "F"))
ggplot(data, aes(x, y)) +
  geom_point() +
  scale_x_discrete(sec.axis = dup_axis(
    labels = c("foo", "bar", "baz"),
    name = "Secondary x-axis"
  )) +
  scale_y_discrete(sec.axis = dup_axis(
    breaks = c(1.5, 2.5), labels = c("grault", "garply"),
    name = "Secondary y-axis"
  ))
```

<sup>Created on 2024-01-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/axis-secondary.R
```diff
@@ -130,6 +130,11 @@ is.sec_axis <- function(x) {
 
 set_sec_axis <- function(sec.axis, scale) {
   if (!is.waive(sec.axis)) {
+    if (scale$is_discrete()) {
+      if (!identical(.subset2(sec.axis, "transform"), identity)) {
```
teunbrand, line 134: The `.subset2()` prevents wrapping the function into a ggproto method.

llrs:
Hi, many thanks for implementing this!
I see in preparation the  next ggplot2 release, is it going to make it in time?

thomasp85 (approved):
LGTM. Please add news bullet

PR #5624, opened 2024-01-03
Fix discrete map bug

### LINKED ISSUE #5623: `Error: Must request at least one colour from a hue palette` when plotting `NA` values
When using the `color` or `fill` aesthetics on a variable with only `NA` values, this produces the following error:
`Error: Must request at least one colour from a hue palette.`

The obvious workaround is to replace the NA values with a different value, such as "none". However, I feel like the default grey color for `NA` should get assigned even when plotting only `NA` values, rather than forcing the user to assign a different color (or circle around and reassign grey for the no-longer-NA values). For me, this would be the expected behavior and helpful in such cases as plotting a large number of graphs in a loop (e.g. when making a markdown report), where some data subsets have only of `NA` values in the `color`/`fill` variable, while others don't.

This behavior might possibly be related to this issue (even though another solution was found there): #5555

Reprex:
```r
data.frame(x = 1:10, y = 11:20, color = "NA") %>%  # this works as expected
    ggplot(aes(x = x, y = y, color = color)) +
    geom_line()

data.frame(x = 1:10, y = 11:20, color = NA) %>%  # this throws an error
    ggplot(aes(x = x, y = y, color = color)) +
    geom_line()
```
Note that the same can be observed for the `fill` aesthetic. `geom_line()` is merely there for illustrative purposes, as the error occurs in `ggplot()`

teunbrand:
This PR aims to fix #5623.

Briefly, when there are no non-NA discrete limits, we return the scale's `na.value`.

Reprex from #5623:

``` r
#> ℹ Loading ggplot2

data.frame(x = 1:10, y = 11:20, color = NA) |>
  ggplot(aes(x = x, y = y, color = color)) +
  geom_line()
```

<sup>Created on 2024-01-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5626, opened 2024-01-04
Revert renaming `Scale$trans` to `Scale$transformation`

teunbrand:
This PR to the RC aims to fix some reverse dependency problems.

To briefly describe the problem: people are using `foo <- Scale$trans` all over their code. We have removed the `Scale$trans` field in https://github.com/tidyverse/ggplot2/pull/5566, in favour of the `Scale$transformation` field, which broke all that code. This PR reverses the renaming of the field specifically.

I think the `Scale$get_transformation()` getter is still useful, so I left it in. It is mostly useful for view scales where the transformation is hidden in the `ViewScale$scale$trans`.


teunbrand:
TODO: use `get_transformation()` consistently

thomasp85:
I think you need to start from scratch because you've confused yourself about what to rename where. There is a method `transform()` which is not touched and you shouldn't try to rename the. 

I think basically it is only the field that should be renamed, the `get_transformation()` method should be updated for that and that would be it

thomasp85 (changes requested):
See previous comment

teunbrand:
I'm not sure I'm following along here. The following bit has me confused:

> There is a method transform() which is not touched and you shouldn't try to rename the.

I am not touching the `Scale$transform()` method in this PR, just the secondary axis' `transform` field, which recieves the user-provided transformation function. Perhaps the confusing thing is that in #5566 I renamed the secondary axis field to `transform` instead of `transformation`?

R/stat-function.R
```diff
@@ -75,7 +75,7 @@ StatFunction <- ggproto("StatFunction", Stat,
     y_out <- inject(fun(x_trans, !!!args))
     if (!is.null(scales$y) && !scales$y$is_discrete()) {
       # For continuous scales, need to apply transform
-      y_out <- scales$y$get_transformation()$transform(y_out)
+      y_out <- scales$y$transform(y_out)
```
thomasp85, line 78: This seems wrong still - rest is ok

thomasp85 (approved):
LGTM

teunbrand:
Thanks for the reviews Thomas!

PR #5627, opened 2024-01-04
`geom_raster()` fallback for non-Cartesian coords

### LINKED ISSUE #5503: Should `geom_raster()` fall back  to `geom_rect()` under non-linear coordinates?
The following code fails:

``` r
library(ggplot2)
#> Warning: package 'ggplot2' was built under R version 4.3.1

ggplot(faithfuld, aes(eruptions, waiting, fill = density)) +
  geom_raster() +
  coord_polar()
#> Error in `geom_raster()`:
#> Caused by error in `draw_panel()`:
#> ! `geom_raster()` only works with `coord_cartesian()`
```

<sup>Created on 2023-10-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

While it shouldn't be too far-fetched to use `geom_rect()`'s drawing method to render this under non-linear coordinate systems. We could throw an informative message saying that we cannot render the data as a raster.

teunbrand:
This PR aims to fix #5503.

Briefly, `geom_raster()` now falls back to drawing as `geom_rect()` when coordinates are non-linear.
A small outline in the same colour as the `fill` aesthetic is added to prevent anti-aliasing.

``` r
#> ℹ Loading ggplot2

ggplot(faithfuld, aes(eruptions, waiting, fill = density)) +
  geom_raster() +
  coord_polar()
#> `geom_raster()` only works with `coord_cartesian()`.
#> ℹ Falling back to drawing as `geom_rect()`.
```

<sup>Created on 2024-01-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5629, opened 2024-01-04
Set `vjust = 1` for `axis.title.y.right`.

teunbrand:
This cold PR aims to fix an informally discussed issue.

Briefly, the current `axis.title.y.right` rotation is -90° and has `vjust = 0`. This puts it in the position closest to the plot panel.
Meanwhile, `axis.title.y.left` rotation = 90° and has `vjust = 1`, placing it the furthest away from the plot panel.
To harmonise their behaviour relative to the panel and place both axes the furthest away, this PR gives `axis.title.y.right` also `vjust = 1`.


thomasp85 (approved):
LGTM

PR #5630, opened 2024-01-04
adhere to strip placement even if the axis is empty

thomasp85:
As discussed internally, the order of strips and axes does not follow the `strip.placement` setting for empty axes. This is fixed with this PR


teunbrand (approved):
I think this fix looks good, but it might be worth streamlining these pieces of code later down the road somewhere.

PR #5631, opened 2024-01-05
New `spacing` and `margins` root theme elements

### LINKED ISSUE #5622: Feature request: spacing and margins as 'root' theme elements
Similar to how `theme(line)` is the root element for all other line elements, we might make `spacing` and `margin` root elements that all other spacings/margins inherit from.

I think this has two benefits:

* It would then become easy to make spacings and margins wider or narrower on the fly, instead of having to use e.g. `theme_gray(base_size)` to simulaneously scale all margins and spacings. The `base_size` argument also scales other stuff, like text, that you may not be interested in scaling.
* Developers can declare specialty theme elements that scale along with the base theme in `register_theme_elements()`. The current solution for this problem is to inherit a spacing from an unrelated theme element and pray it doesn't interact inconveniently.

teunbrand:
This PR aims to fix #5622.

Briefly, this PR adds `spacing` and `margins` theme elements that act as the 'root' element for other spacing and margin settings in the plot. For example, the `legend.box.spacing` inherits from `spacing` and `legend.margin`  inherits from `margins`. The default looks of themes are preserved.

As a detail, functions like `theme_gray()` already *sort of* applied the same logic as this PR applies via the `half_line` variable. This PR implicitly ensures that the `half_line` variable also gets stored in the `spacing` element.

Regretfully, I could not find an elegant way to have text elements inherit from the `margins` element, so these remain as-is.

I also made the `legend.ticks.length` inherit from `legend.key.size`. Currently `legend.ticks.lenght` was set by default to `"npc"` units, but that might have no clear meaning depending on the type of legend. Inheriting from the key size preserves the look but is more clear what the length will be without knowing the internal detail that the 'npc' units relate to the dimensions of a particular cell in the gtable.

Some examples.

``` r
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(displ, hwy, colour = factor(cyl))) +
  geom_point() +
  facet_grid(~ drv)

p + theme(spacing = unit(1, "cm"))
```

``` r
p + theme(margins = margin(1, 1, 1, 1, "cm"))
```

<sup>Created on 2024-01-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5633, opened 2024-01-05
Drop empty groups from bar width calculation

### LINKED ISSUE #5632: Warning in `geom_col()` when panel has 0 rows
This is a slightly contrived example, but the bar's width calculation can be chatty when there is a factor level in the `PANEL` that has 0 rows in the data.

``` r
library(ggplot2)

df <- mpg
df$drv <- factor(df$drv, c("4", "f", "r"))

ggplot(subset(df, drv != "4"), aes(cyl, cty)) +
  geom_col() +
  facet_wrap(~ drv, drop = FALSE)
#> Warning in min(d[d > tolerance]): no non-missing arguments to min; returning
#> Inf
```

<sup>Created on 2024-01-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5632.

As the title suggests, this PR adds a `drop = TRUE` to a `split()` that could return a list with 0-length elements in it. With `drop = TRUE`, those 0-length elements are absent and the calculation proceeds without warnings.

### DIFF
--- a/R/geom-bar.R
+++ b/R/geom-bar.R
@@ -145,7 +145,7 @@ GeomBar <- ggproto("GeomBar", GeomRect,
     data <- flip_data(data, params$flipped_aes)
     data$width <- data$width %||%
       params$width %||% (min(vapply(
-        split(data$x, data$PANEL),
+        split(data$x, data$PANEL, drop = TRUE),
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

PR #5634, opened 2024-01-08
Fill potentially missing arguments in facets

teunbrand:
This PR to the RC aims to fix an issue uncovered during reverse dependency checks.

Briefly, we introduced `axes` and `axis.labels` arguments in `facet_grid()` and `facet_wrap()`.
Extensions likely do not have these arguments forwarded to the facet's parameters and may be missing, causing errors in the panel drawing stage.
This PR fills these parameters when they are absent, so that extensions continue to work.


thomasp85 (approved):
LGTM

PR #5635, opened 2024-01-08
Put deprecated `trans` last in `dup_axis()`

teunbrand:
This PR to the RC aims to fix a reverse dependency check issue with `dup_axis()`.

Briefly: we renamed `trans` to `transform`, but kept the deprecated `trans` argument as the second argument.
Folks use position-based argument matching, so that the `trans` argument gets populated with what should go into the `name` argument. This PR restores the original order by putting the deprecated argument last.


thomasp85 (approved):
LGTM

PR #5642, opened 2024-01-15
Avoid bounded density discontinuity

### LINKED ISSUE #5641: Bounded densities have discontinuities due to incomplete estimation of tails in the reflection method
I believe the current implementation of the reflection method for bounded densities has a minor bug which causes discontinuities in the density estimate.

For example, see this plot of a density estimate on uniform quantiles:

```r
x = ppoints(100)
p = ggplot(data.frame(x), aes(x)) + 
    geom_density(bounds = c(0,1))
p
```
image

The discontinuity is visible, but small. If you zoom in:

```r
p + coord_cartesian(ylim = c(.99, 1.01))
```
image

I believe this is happening because the tails are not being estimated to their full length, but only to `3*bw` beyond the boundary, so when those tails are reflected back there is a discontinuity at `3*bw` inside the boundary when the density of the reflected tail drops to exactly 0.

One solution to this is to estimate the tail density out to `diff(range(x))` beyond the boundary. This is what ggdist's implementation of the reflection method does, so it does not have the discontinuity:

```r
plot(ggdist::density_bounded(x, bounds = c(0,1)), ylim = c(0.99, 1.01))
```
image

Looking at the implementation in ggplot2, I'm not sure the best way to suggest fixing this issue since by the time `reflect_density()` is called the unbounded estimator has already been run. You'd probably at least have to change the bounds used by the unbounded estimator prior to the call to `reflect_density()`.

teunbrand:
This PR aims to fix #5641.

Briefly, the way bounded densities were estimated may result in discontinuities due to incomplete estimation of the tails.
As densities are 'mirrored' across finite bounds, in this PR we ensure that the 'mirror image' is the same size as the original by multiplying the `to`/`from` range in density estimation. We also increase `n` to not lose any detail in the bounded density.

The reprex from #5641 with this PR now shows a flatter line in the uniform density:

``` r
#> ℹ Loading ggplot2

x <- ppoints(100)
p <- ggplot(data.frame(x), aes(x)) + 
  geom_density(bounds = c(0,1))
p
```

``` r
p + coord_cartesian(ylim = c(.99, 1.01))
```

<sup>Created on 2024-01-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


teunbrand:
@mjskay Could I ask you if you think this implementation is correct?

mjskay:
Yeah this looks right to me!

thomasp85 (approved):
LGTM

PR #5644, opened 2024-01-15
Do not munch data too much

teunbrand:

Briefly, for reasons beyond my understanding, `coord_munch()` might receive a very large segment, let's say `1e20` for `x` or `y`. It might then attempt to break up this large segment into many tiny pieces. You'd then get `failed to allocate vector of 7Gb` types of errors. Setting an upper limit on the number of pieces prevents this unreasonable demand on the machine's memory.


thomasp85 (approved):
LGTM - this is a baffling error though

PR #5645, opened 2024-01-15
Drop `zeroGrob()` guides

teunbrand:

Briefly, in some unnatural circumstances you can create a guide that survives all steps but decides to draw a `zeroGrob()`.
`Guides$package_box()` cannot estimate the gtable size of these guides.

``` r
library(ggplot2)

ggplot(mpg[1:2, ], aes(displ, cty, colour = c(1, 4))) +
  geom_point() +
  scale_colour_continuous(
    limits = c(2.4, 2.6), breaks = c(1, 4),
    guide = "legend"
  )
#> Error in upgradeUnit.default(x): Not a unit object
```

<sup>Created on 2024-01-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The solution in this PR is to drop empty guides at the `Guides$draw()` step. The reason we might not want to censor guides as soon as we detect the key is empty, is because some (extension) guides might have genuine reasons for not using a key.

### DIFF
--- a/R/guides-.R
+++ b/R/guides-.R
@@ -547,7 +547,8 @@ Guides <- ggproto(
         direction = directions[i], params = params[[i]]
       )
     }
-    split(grobs, positions)
+    keep <- !vapply(grobs, is.zero, logical(1))
+    split(grobs[keep], positions[keep])
   },
 
   package_box = function(grobs, position, theme) {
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5648, opened 2024-01-16
Simplify legend layout logic

teunbrand:
This PR to the RC aims to make a simplification to `guide_legend()`.

Briefly, this PR adds a `legend_add_title()` helper that is used in `guide_legend()` and `guide_custom()`. This has two advantages:

* We don't have to maintain two separate instances of title logic.
* We can simplify the layout code in `guide_legend()`, as we no longer have to do bookkeeping and adjustments to the key/label positions based on the title.

I figured if there was ever a time to simplify the legend layout logic, without having to deal with backward compatibility issues, now is that time.


R/guide-legend.R
```diff
@@ -739,6 +665,69 @@ position_margin <- function(position, margin = margin(), gap = unit(0, "pt")) {
...
+      return(gtable)
+    }
+    extra_height <- unit((c(-1, 1) * just$vjust + c(1, 0)) * extra_height, "cm")
+    gtable <- gtable_add_rows(gtable, extra_height[1], pos =  0)
+    gtable <- gtable_add_rows(gtable, extra_height[2], pos = -1)
+  }
+
+  gtable
+}
+
```
thomasp85, line 730: Is there a particular reason why we don't make this a method of the Guide class?

teunbrand: Not a strong reason. A weak reason is that `Guide$assemble_drawing()` doesn't have a `self` to call the method with, and it needs to happen before adding the margin padding. That said, there is no reason why it may not have a `self` though.

thomasp85: Yeah, I think we should add it as a method. One of the nice side effects of compartmentalise these utility functions in the base ggproto class is that extensions have access to them without us needing to pollute the namespace

teunbrand: You make a convincing argument, I've added it to the `Guide` class.

R/guide-legend.R
```diff
@@ -521,97 +489,40 @@ GuideLegend <- ggproto(
...
   },
 
-  assemble_drawing = function(grobs, layout, sizes, params, elements) {
-    widths <- unit(c(sizes$padding[4], sizes$widths, sizes$padding[2]), "cm")
+  assemble_drawing = function(self, grobs, layout, sizes, params, elements) {
+
+    widths <- unit(sizes$widths, "cm")
     if (isTRUE(elements$stretch_x)) {
-      widths[unique(layout$layout$key_col)] <- elements$key_width
+      widths[unique(layout$key_col)] <- elements$key_width
```
thomasp85, line 516: let's update all these to `unique0()` while we are here

thomasp85 (approved):
LGTM

PR #5652, opened 2024-01-19
Change legend tick length inheritance

teunbrand:
This PR to the RC aims to fix a mild annoyance.

Briefly, instead of guides declaring their own `legend.ticks.length` with a default of `unit(0.2, "npc")`, we now inherit from the `legend.key.size` with a default of `rel(0.2)`. For all guides in ggplot2 this is effectively the same. This has two benefits:

* Guides don't need to bookkeep these defaults.
* While npc units means something in the context of colour bars and bins guides, a guide extension might have a different viewport in which to display ticks and 'npc' units may be awkward and inconsistent. This PR avoids this.


thomasp85 (approved):
LGTM

PR #5662, opened 2024-01-25
Restore `legend.byrow`

### LINKED ISSUE #5661: legend.byrow = TRUE doesn't seem to be working
Hello

I'm using the 3.4.4900 version of ggplot2 and the `legend.byrow` argument doesn't seem to be working as expected. I added the argument in both `theme` and `guides` but it makes no difference if you add it in only one of them or both (or none I guess).

```R
df <- data.frame(x = 1:20, y = 1:20, color = letters[1:20])
ggplot(df, aes(x, y)) +
    geom_point(aes(colour = color)) +
    theme(
        legend.byrow = TRUE
    ) +
    guides(col = guide_legend(nrow = 8, byrow = TRUE))
```
Expected behaviour
guide_legend-10
Actual behaviour
image

teunbrand:
This PR to the RC aims to fix #5661.

Briefly the `legend.byrow` setting wasn't working because the code assumed it was in `params` whereas we swapped it to `elements` later. 

This made me realise that `Guide$arrange_layout()` needed access to the `elements` (instead of sneakily passing it through `size` ), so this PR also adds `elements` as an argument. 

I was also surprised that there was no test for the `legend.byrow` setting, so I added one.

### DIFF
--- a/R/guide-.R
+++ b/R/guide-.R
@@ -329,7 +329,7 @@ Guide <- ggproto(
 
     # Arrange and assemble grobs
     sizes  <- self$measure_grobs(grobs, params, elems)
-    layout <- self$arrange_layout(key, sizes, params)
+    layout <- self$arrange_layout(key, sizes, params, elems)
     self$assemble_drawing(grobs, layout, sizes, params, elems)
   },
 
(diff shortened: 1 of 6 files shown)


thomasp85 (approved):
LGTM

PR #5663, opened 2024-01-25
Balance geom label whitespace

teunbrand:
This PR to the RC aims to fix an internally discussed issue.

Briefly, in the RC we always add the descent height at the bottom of the label, which is not always visually pleasing. In this PR, descent is also added to the top and descent is ignored when the label padding exceeds the descent.

I think a visual demonstration will perhaps explain it best.
In the current ggplot 3.4.4, the issue was that the descent was not applied at all, leaving some characters, like 'j' partially hanging outside the textbox with 0 padding.

``` r
library(ggplot2)
library(patchwork)

df <- data.frame(x = 1, y = 1, hjust = c(0, 1, 0, 1), vjust = c(0, 0, 1, 1))

p <- ggplot(df) +
  aes(
    x = x, y = y, 
    hjust = hjust, vjust = vjust,
    label = paste0("hjust = ", hjust, ", vjust = ", vjust)
  )

p1 <- p + geom_label(label.padding = unit(0, "pt")) + 
  labs(title = "padding = 0pt")
p2 <- p + geom_label(label.padding = unit(5, "pt")) + 
  labs(title = "padding = 5pt")
p1 / p2
```

<sup>Created on 2024-01-25 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The current development version mitigates this issue by always adding the descent to the bottom of the label (by means of #5030).
There are two issues with this:

* It looks imbalanced to have plenty of whitespace at the bottom and not at the top.
* The application of `vjust` is awkward, having the descent poke out beneath the anchor point when `vjust = 0`.

Same code with current RC branch:

This PR applies the descent only if the padding is insufficient, and simultaneously at the top and the bottom of the label.
This fixes the balance issue mentioned before. While at it, I also fixed the `vjust` issue.

Same code with this PR:


thomasp85 (approved):
LGTM

PR #5668, opened 2024-01-30
Only use finite breaks for computing fuzz

### LINKED ISSUE #5665: Histogram with breaks that uses Inf does not works
Hi, I was trying to use ```breaks``` from ```geom_histogram```, when we want to represent data, in order to show all of it, we can fill or group them, that is the point of breaks, it sets the limits between the bins and then we sum the samples there, well if we try something like ```c(-Inf, 2, 6, Inf)``` to get the bins -Inf/2, 2/6, 6/Inf it fails:

```
data <- data.frame(x = c(rep(1, 10), 5))

ggplot2::ggplot(data, ggplot2::aes(.data$x)) +
ggplot2::ggtitle("Distribucion de error") +
ggplot2::geom_histogram(
  ggplot2::aes(y = ggplot2::after_stat(count * 100 / sum(count))),
  breaks = c(-Inf, 2, 6, Inf),
  closed = "left"
)
Warning message:
Computation failed in `stat_bin()`
Caused by error in `cut.default()`:
! 'breaks' are not unique
```

Initially my idea was check if the all the ```1``` are inside the break -Inf/2, but sadly does not works.

<details>
sessionInfo()
R version 4.3.2 (2023-10-31)
Running under: Gentoo Linux

Matrix products: default
BLAS:   /usr/lib64/libblas.so.3.11.0 
LAPACK: /usr/lib64/liblapack.so.3.11.0

locale:
 [1] LC_CTYPE=es_CL.utf8       LC_NUMERIC=C             
 [3] LC_TIME=es_CL.utf8        LC_COLLATE=es_CL.utf8    
 [5] LC_MONETARY=es_CL.utf8    LC_MESSAGES=es_CL.utf8   
 [7] LC_PAPER=es_CL.utf8       LC_NAME=C                
 [9] LC_ADDRESS=C              LC_TELEPHONE=C           
[11] LC_MEASUREMENT=es_CL.utf8 LC_IDENTIFICATION=C      

time zone: Chile/Continental
tzcode source: system (glibc)

attached base packages:

other attached packages:
[1] wrapr_2.1.0

 [1] gtable_0.3.4       lwgeom_0.2-13      dplyr_1.1.4        compiler_4.3.2    
 [5] crayon_1.5.2       tidyselect_1.2.0   Rcpp_1.0.12        tidyr_1.3.0       
 [9] scales_1.3.0       ggplot2_3.4.4      R6_2.5.1           sfheaders_0.4.3   
[13] labeling_0.4.3     generics_0.1.3     igraph_1.6.0       classInt_0.4-10   
[17] s2_1.1.6           sf_1.0-16          tibble_3.2.1       units_0.8-5       
[21] munsell_0.5.0      DBI_1.2.1          pillar_1.9.0       rlang_1.1.2       
[25] utf8_1.2.4         fs_1.6.3           cli_3.6.2          withr_2.5.2       
[29] magrittr_2.0.3     class_7.3-22       wk_0.9.1           grid_4.3.2        
[33] sfnetworks_0.6.3   lifecycle_1.0.4    vctrs_0.6.5        KernSmooth_2.23-22
[37] proxy_0.4-27       glue_1.6.2         farver_2.1.1       tidygraph_1.2.3   
[41] fansi_1.0.6        e1071_1.7-14       colorspace_2.1-0   purrr_1.0.2       
[45] tools_4.3.2        pkgconfig_2.0.3 
</details>

Thx!

teunbrand:
This PR aims to fix #5665.

Briefly, when histogram `breaks` are short and contain non-finite values, the median difference between breaks can become `Inf`.
In this PR, we only use finite breaks to compute the median difference for the fuzz-factor.

Reprex from #5665

``` r
#> ℹ Loading ggplot2

data <- data.frame(x = c(rep(1, 10), 5))

ggplot(data, aes(x)) +
  geom_histogram(
    breaks = c(-Inf, 2, 6, Inf),
    closed = "left"
  )
#> (`geom_bar()`).
```

<sup>Created on 2024-01-30 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


R/bin.R
```diff
@@ -1,6 +1,7 @@
 bins <- function(breaks, closed = "right",
-                 fuzz = 1e-08 * stats::median(diff(breaks))) {
+                 fuzz = NULL) {
   check_numeric(breaks)
+  fuzz <- fuzz %||% 1e-08 * stats::median(diff(breaks[is.finite(breaks)]))
```
thomasp85, line 4: what happens if there is only one finite break?

teunbrand: Currently, it throws an error. I guess a not unreasonable approach would be to set the fuzz to a very small number?

thomasp85: yeah, I'd be fine with that, but we should not throw error at least

thomasp85 (approved):
LGTM

PR #5675, opened 2024-02-02
Update `scale_alpha()` examples

### LINKED ISSUE #2865: Range for continuous alpha
The default range for `scale_alpha_continuous()` goes from 0.1 to 1. I thought I'd create an issue to discuss the rationale behind a minimum that isn't 0.

This image via @clauswilke shows that the current minimum of 0.1 can misleadingly leave a lightly shaded region where values should be 0.
image

For discrete scales, I understand not including 0, as you might hide a category. But for continuous scales, why not start at 0?

teunbrand:
This PR aims to fix #2865.

Rendered examples:

``` r
library(ggplot2)
  geom_point(aes(alpha = year))

# The default range of 0.1-1.0 leaves all data visible
p
```

``` r

# Include 0 in the range to make data invisible
p + scale_alpha(range = c(0, 1))
```

``` r

# Changing the title
p + scale_alpha("cylinders")
```

<sup>Created on 2024-02-02 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/scale-alpha.R
+++ b/R/scale-alpha.R
@@ -21,9 +21,14 @@
 #' p <- ggplot(mpg, aes(displ, hwy)) +
 #'   geom_point(aes(alpha = year))
 #'
+#' # The default range of 0.1-1.0 leaves all data visible
 #' p
+#'
+#' # Include 0 in the range to make data invisible
+#' p + scale_alpha(range = c(0, 1))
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5676, opened 2024-02-02
Deduplicate `geom_function()` examples

### LINKED ISSUE #3181: Review duplicated examples about multiple notations
Some functions accept multiple ways of specification. It's useful to show that the multiple notations can generate the same results, but it's not necessary to execute all the examples. In order to reduce the check time, we need to restructure these examples (e.g. commenting out the less promoted notations)

* anonymous functions (`function(x)` vs `~`)
    * `stat_function()`
* facet specs (`vars()` vs `~` vs character)
    * `facet_wrap()`
    * `facet_grid()`

For anonymous functions, I think we should wait for ggplot2 to get support for `~` notation in all the places that accept functions.

teunbrand:
This PR aims to fix #3181.

Briefly, it comments out various options of providing the function to `geom_function()`. I didn't find the duplicated examples for facets mentioned in #3181, so I think removing the duplicates for `geom_functino()` suffices.

### DIFF
--- a/R/geom-function.R
+++ b/R/geom-function.R
@@ -42,14 +42,14 @@
 #'   geom_function(aes(colour = "t, df = 1"), fun = dt, args = list(df = 1))
 #'
 #' # Using a custom anonymous function
-#' base + geom_function(fun = function(x) 0.5*exp(-abs(x)))
-#'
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5677, opened 2024-02-02
Clarify that limits are given after expansion

### LINKED ISSUE #3445: Clarify documentation of breaks function in scale_continuous
This is a documentation clarification request (or perhaps it is a bug report). The documentation for `scale_continuous` says that the `breaks` argument can take "a function that takes the limits as input and returns breaks as output". With some experimentation, I determined that the input is the limits after expansion and _not_ what gets passed to the `limits` argument. If this is the intended behavior (reasonable given that having empty bands with no breaks around the data looks weird), it should probably be clear. My attempt to learn how to contribute on GitHub for the last hour did not get far, but here is my recommended wording: "A function that takes the limits (after expansion as specified by expand) as input and returns breaks as output"

To make it obvious what I mean, 

``` r
library(tidyverse)

# Data
dt = tibble(
  x = 1:6,
  y = rnorm(x)
)

# Function that seems to "take the limits as input and return breaks as output"
BreakFunc = function(lim) seq(lim[1], lim[2], 2)

# I expect the plot to have horizontal axis breaks at seq(0, 7, 2) but it does not
ggplot(dt, aes(x=x, y=y))+
  geom_point() +
  scale_x_continuous(
    limits = c(0, 7), 
    breaks = BreakFunc)
```

``` r
#  Explicitly setting the expand argument gets the expected plot
ggplot(dt, aes(x=x, y=y))+
  geom_point() +
  scale_x_continuous(
    limits = c(0, 7), 
    breaks = BreakFunc,
    expand = expand_scale(mult = 0, add = 0))
```

teunbrand:
This PR aims to fix #3445.

It simply adds a line to the function option for breaks that position scales get expanded limits. Since discrete scales document their own `breaks` param, this should only appear in continuous and binned scales.

### DIFF
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -19,6 +19,7 @@
 #'   - A numeric vector of positions
 #'   - A function that takes the limits as input and returns breaks
 #'     as output (e.g., a function returned by [scales::extended_breaks()]).
+#'     Note that for position scales, limits are provided after scale expansion.
 #'     Also accepts rlang [lambda][rlang::as_function()] function notation.
 #' @param minor_breaks One of:
 #'   - `NULL` for no minor breaks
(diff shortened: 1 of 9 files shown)


thomasp85 (approved):
LGTM

PR #5678, opened 2024-02-02
Crosslink documentation with book

### LINKED ISSUE #3748: Cross-link docs with ggplot2 chapters
As the ggplot2 book is now freely available it is a perfect end-point for "read more" linking from the docs.

teunbrand:
This PR aims to fix #3748.

Briefly, it sprinkels `@seealso` links to the book all over the place.
It currently links to the 3rd edition that is still WIP, and we'd probably be wise to wait merging this until &mdash;and double-check URLs when&mdash; that is polished/published.


teunbrand:
@hadley Is it fine to link to the WIP 3rd edition or would you prefer we link to the 2nd edition of the ggplot2 book?

hadley:
I think linking to the 3e is the right choice, and if any of the pages change, we'll get an error from the url checker as part of the release process (although that won't identify problems due to headings changing 🤔).

You might want to create a little helper like this one I made for httr2:

```R
rfc <- function(num, sec = NULL) {
  paste0(
    "[",
    if (!is.null(sec)) paste0("Section ", sec, " of "),
    "RFC ", num, "]",
    "(https://datatracker.ietf.org/doc/html/rfc", num,
    if (!is.null(sec)) paste0("#section-", sec),
    ")"
  )
}
```

That makes it easier to update if (say) you decide to change "online book" to something else.

teunbrand:
Thanks Hadley for that suggestion. If 3E is ok, I'll go ahead and undraft this PR.

hadley (approved):

teunbrand:
Thanks for the review!

PR #5679, opened 2024-02-06
Radial inner radius

teunbrand:
This PR aims to implement an internally discussed suggestion.

It renames the `donut` argument of `coord_radial()` to `inner.radius`.


thomasp85 (approved):
LGTM

PR #5681, opened 2024-02-06
Discard empty colourbars

teunbrand:
This PR to the RC solves a reverse dependency issue.

Briefly, if scale limits are set to e.g. `c(0, NA)`, but all values are `NA`, the limits never get updated to finite values, causing a problem for the colourbar. This PR intercepts such problems by discarding colourbars that have 0 breaks.


thomasp85 (approved):
LGTM

PR #5657, opened 2024-01-22
Don't strip classes in `Geom$use_defaults()`

teunbrand:
This PR to the RC aims to fix #5656.

Briefly, using ``base::`<-.data.frame` `` in `Geom$use_defaults()` causes classes to be stripped on the RHS of the assignment.
This PR uses `vec_cbind()` to do the same job without stripping classes.
I encountered some reordered rownames during a test, but as these are not important in ggplot2, I'm not concerned.

PR #5683, opened 2024-02-07
Revert #5657

teunbrand:
This PR to the RC aims to fix a revdep bug.

Briefly, in #5657, we used `vec_cbind()` to preserve AsIs vectors. However, {vctrs} does not treat expressions as vectors. Any geom declaring an expression outside of `aes()` would fail. This PR restores the old subassignment method, but instead explicitly restores AsIs vectors.


thomasp85 (approved):
LGTM

Issue #5609, opened 2023-12-22
Supporting unit()s comprehensively in ggplot2 aesthetics


mjskay:
### The problem

Currently, a big source of fiddlyness in ggplot2 comes down to problems when you need to make manual adjustments to the positions of objects in the final plot. Often (but not always) this happens when doing annotation, e.g. positioning labels such that they are nicely and consistently spaced or positioned relative to data, and/or relative to the plot itself. This can be frustrating because if the plot or data dimensions change, manually-positioned elements will also move, because the only way to position them is by supplying values in *data units*. However, annotation usually needs to be specified in plot units (e.g., points, npcs, etc)---or even worse---in a combination of data units and plot units.

That said, {grid} has robust support for unit conversion in the form of `grid::unit()`. If ggplot was able to specify positions (and even `size`s, `linewidth`s, etc) using `grid::unit()`s, it would be a lot easier to create charts that look good even when the limits of underlying x/y scales change or when the dimensions of the plot changes.

Even better would be able to combine data units and plot units: e.g., to be able to specify something like "put this text label exactly 5 points to the left of this data position". `grid::unit()` already has a data unit (the `"native"` unit) that could be used for this purpose, and it is somewhat unused in ggplot2 because it internally scales everything into `c(0,1)` anyway.

### A possible solution for positional aesthetics

I started playing with the above problem earlier this week, and I think I've come up with something that, with some polish, might be able to solve the problem, at least for positional aesthetics, without much surgery---and which would allow extension packages to more-or-less automatically take advantage of it.

The basic idea is to allow `unit()` vectors containing at most one `"native"` unit to be assigned to positional aesthetics, and for this native unit to represent the (transformed) data. Then you can do stuff like this:

```r
data.frame(var1 = 1:5, var2 = 1:5, name = letters[1:5]) |>
  ggplot(aes(var1, var2)) + 
  geom_point() +
  # a line exactly 5 points lower than `var2`
  geom_line(aes(y = unit(var2, "native") - unit(5, "pt"))) +
  # labels exactly 10 points left of their points, no matter how the
  # plot is resized
  geom_text(aes(label = name, x = unit(var1, "native") - unit(10, "pt"))) +
  # an annotation that is always 10 points inset from the lower right
  annotate("text", 
    x = unit(1, "npc") - unit(10, "pt"), 
    y = unit(10, "pt"), 
    label = "some label", vjust = 0, hjust = 1
  )
```
[...]

PR #5690, opened 2024-02-11
Minimal PR for unit support in positional aesthetics

mjskay:
Following discussion on #5609 and helpful feedback from @teunbrand, @pmur002, and @thomasp85 (thanks!), this is a second proof-of-concept PR for unit support in positional aesthetics in ggplot2. This PR focuses on being a minimal implementation *without* hacking on the unit type system.

The basic idea here is to instead add another aesthetic evaluation stage, `after_coord`, which is akin to `after_stat` or `after_scale`. It is implemented by passing aesthetic mappings with `stage()` or `after_coord()` through to `Coord$transform()` via the `panel_params` argument, which is already passed through to that function. `Coord$transform()` can then apply the `after_coord` mappings after it does its own `Coord`-specific transformations.

Some demos:

```r
data.frame(var1 = 1:5, var2 = 1:5, name = letters[1:5]) |>
  ggplot(aes(var1, var2)) + 
  geom_point() +
  geom_line() +
  # labels exactly 8 points left and 6 points above their points, no matter how 
  # the plot is resized
  geom_text(aes(
    label = name, 
    x = stage(var1, after_coord = unit(x, "native") - unit(8, "pt")),
    y = stage(var2, after_coord = unit(y, "native") + unit(6, "pt"))
  )) +
  # an annotation that is always 10 points inset from the lower right
  # N.B. can't use annotate() because it doesn't understand stages, but using a 
  # geom with data = data.frame() instead seems to work. If desired, I believe
  # annotate() could be made to understand stages by having it check for the
  # stage / after_stat / after_scale /after_coord functions in the bare
  # expressions supplied to it.
  geom_text(
    aes(
      x = after_coord(unit(1, "npc") - unit(10, "pt")),
      y = after_coord(unit(10, "pt"))
    ),
    label = "some label", vjust = 0, hjust = 1,
    data = data.frame()
  ) +
  # a horizontal line exactly 25 points from the bottom.
  # can't use after_coord() with the yintercept aesthetic because geom_hline()
  # translates that to `y` and `yend` under the hood before the after_coord()
  # stage, so must modify those values directly
  geom_hline(aes(
    yintercept = Inf,  # placeholder value to avoid "missing required aes" error
    y = after_coord(unit(25, "pt")),
    yend = after_coord(unit(25, "pt")),
  ))
```
image

With polar coordinates:

```r
data.frame(var1 = 1:5, var2 = 1:5, name = letters[1:5]) |>
  ggplot(aes(var1, var2)) + 
  geom_point() +
  geom_line() +
  geom_text(aes(
    label = name, 
    x = stage(var1, after_coord = unit(x, "native") + unit(10, "pt"))
  )) +
  geom_text(
    aes(
      x = after_coord(unit(1, "npc") - unit(10, "pt")),
      y = after_coord(unit(10, "pt"))
    ),
    label = "some label", vjust = 0, hjust = 1,
    data = data.frame()
  ) +
  coord_polar()
```
image

Some notes:

- The main workhorse here is `compute_staged_aes()`, which is based on a chunk of code I factored out of `Geom$use_defaults()` for computing `after_scale` aesthetics.
- `after_coord()` is a bit different from `after_scale()` in that it provides the placeholder value `Inf` to earlier parts of the pipeline, because it needed a non-`unit()` value that would not affect scales (see the comment on that function).
- the implementation moves current `Coord$transform()` implementations into `Coord$transform_numeric()`. This was done so that `Geom`s (which already call `Coord$transform()`) would not have to opt in to supporting `unit()`s; instead, `Coord`s in extension packages can opt-in simply by renaming their `transform()` methods to `transform_numeric()`. An alternative might be to leave `Coord$transform()` as-is but add a `Coord$transform_unit()` method, and then have `Geom`s opt-in to unit support instead of having `Coord`s opt-in to it. I think having `Coord`s opt-in is better since it requires fewer changes in extension packages and propagates support more easily, but I could be wrong.
- No hacking on `unit()` is required for this implementation, nor any upstream changes to {grid}. A small set of compatibility functions would need to be added to {vctrs}, akin to how {vctrs} supplies compatibility functions for other base-R types, like Dates. These are confined to `R/utilities-unit.R`, which could be turned into a PR on {vctrs}.
- In exchange for less "magic", this approach is a little more verbose than my original proposal and requires a little more understanding of the ggplot2 pipeline for users. Something like this in the current proposal:
   
   ```r
   x = stage(var1, after_coord = unit(x, "native") - unit(8, "pt"))
   ```

   could be written as this in the original proposal:

   ```r
   x = var1 - ggunit(8, "pt")
   ```

   However, this new proposal has more consistent semantics and no heuristic hacking of the `unit()` datatype :). A `ggunit()` subtype of `unit()` with consistent coercion rules for numerics could still be created to improve the syntax a bit (also without heuristic hacking of `unit()`); allowing something like this to work:

   ```r
   x = stage(var1, after_coord = x - ggunit(8, "pt"))
   ```

   Adding shortcuts like `as_pt()` could shorten this still, to:

   ```r
   x = stage(var1, after_coord = x - as_pt(8))
   ```

   The `ggunit()` subtype could be added to ggplot2, or if that is not desired, I'd be happy to create a small extension package with that capability.


teunbrand:
I'm happy to help think about the code and I still like the concept a lot. It is great that this PR is much less invasive than the previous one. 

However I feel like the sheer scope of this PR is outside my jurisdiction and the decision on whether to forge ahead with this is in @thomasp85's hands. If he doesn't reply here after invoking his name, I'll be sure to bring it up the next time Thomas and I have a chat (in ~2 weeks time).

Issue #5655, opened 2024-01-21
`binned_scale` defaults to `oob = squish` though the documentation suggests `oob = censor`


trekonom:
I was today years old when I learned that binned scales use `oob = squish` as default

https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/R/scale-.R#L269

though the documentation suggests that it is `oob = censor`

https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/man/binned_scale.Rd#L94-L95

The same holds for the documenation of "all" binned scales which inherit (the documentation of parameters) from `binned_scale`.

The issue seems to be that `binned_scale` itself inherits the documentation of `oob` from `continuous_scale`:

https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/R/scale-.R#L252

Feel free to ping me if I should prepare a PR though I guess this could be easily fixed.

teunbrand:
Considered fixed by #5691.

PR #5691, opened 2024-02-12
Fixing documentation in binned_scale suggesting it defaults to oob = censor

joshhwuu:
Attempting to resolve issue #5655


teunbrand:
Hi there, thanks for putting together a PR. The documentation is generated from the .R files, so the appropriate place to make edits is here: https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/R/scale-.R#L249-L266

After such edits have been made, you can use `devtools::document()` to regenerate the .Rd files.

joshhwuu:
When I was using roxygen2::roxygenise(), it automatically remade the .Rd files for a lot of the R files, including aes_colour_fill_alpha.Rd, position_dodge.Rd, etc. These changes seem to be minimal in every file except man/binned_scale.Rd,  man/scale_binned.Rd and man/scale_steps.Rd. 

For example, in man/aes_colour_fill_alpha.Rd:

image

It seems that document() simply reordered two of the items so all items are alphabetical, and the same goes for some of the other minimally-changed files.

Should I revert the changes to these minimally-changed documents and then push again or is it okay to leave them in?

teunbrand:
The unrelated changes are due to roxygen2 versions and is nothing to worry about. We'll merge in another PR soon that updates these files too, and they shouldn't show up as diffs anymore. I'm going to hold off on merging this PR until release, to avoid conflicts, but I think the PR looks great as-is.

teunbrand (approved):
Looks good to me!

PR #5696, opened 2024-02-18
For geom_label highlight passed through force and force_pull arguments 

julianbarg:
from ggrepel in man page

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -51,6 +51,7 @@ Suggests:
     covr,
     dplyr,
     ggplot2movies,
+    ggrepel,
     hexbin,
     Hmisc,
     knitr,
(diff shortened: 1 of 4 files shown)


teunbrand:
I apologise for any misconceptions on my part, but what issue is this PR aiming to fix?

julianbarg:
Better discoverability of probably the most used function argument of the function geom_label -- force and force_pull. Since it's such a simple issue I figured I would send a PR rather than opening a ticket.

teunbrand:
But `geom_label()` does not have force/force_pull arguments nor inner workings for these options. These seem arguments to ggrepel::geom_label_repel [see here](https://ggrepel.slowkow.com/reference/geom_text_repel). It's not really customary for ggplot2 do document arguments that extensions provide, they usually document these themselves, as ggrepel has.

PR #5699, opened 2024-02-22
Gradients in ribbons

### LINKED ISSUE #4690: `geom_ribbon`: aesthetics can not vary with a ribbon
Hi! I ran across the error in the title when mapping `alpha` to an aesthetic in `geom_ribbon`. ggplot2 is a mature package, so I suspect there are well-founded reasons this isn't supported, but on the off-chance, is this something that can be considered in the future?

In my use case, I was mapping confidence interval around an estimate to `geom_ribbon`, and wanted to have it "fade out" as it projected beyond the current date. Here's a toy reprex with the error:

``` r
library(tidyverse)

pi <- 3.1415926

tibble(x = seq(0, 2 * pi, length.out = 100)) %>%
  mutate(estimate = sin(x),
         ci_hi = estimate + 0.1,
         ci_lo = estimate - 0.1,
         
         # decrease alpha from pi -> 2*pi
         alpha = if_else(x >= pi, 0.25 * (1 - (x - pi)/pi), 0.25)) %>%
  ggplot(aes(x = x,
             y = estimate,
             ymin = ci_lo,
             ymax = ci_hi)) + 
  geom_line() +
  geom_ribbon(aes(alpha = alpha))
#> Error: Aesthetics can not vary with a ribbon
```

<sup>Created on 2021-12-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

teunbrand:
This PR aims to fix #4690.

Briefly, `geom_ribbon()` (and by extension `geom_area()`) didn't allow varying `fill` or `alpha` aesthetics within a group.
This PR encodes `fill` and/or `alpha` in a `grid::linearGradient()`, which allows these aesthetics to vary.

Demo for `fill`:

``` r
#> ℹ Loading ggplot2

ggplot(economics, aes(date, unemploy, fill = uempmed)) +
  geom_area(stat = "identity") +
  scale_fill_viridis_c()
```

Demo for `alpha`:

``` r
ggplot(economics) +
  aes(date, ymin = unemploy - 200, ymax = unemploy + 200, alpha = pop) +
  geom_ribbon()
```

<sup>Created on 2024-02-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

A few limitations:

* Using this in non-linear coordinates, say polar coordinates, is a bad idea as we cannot do gradient curves or conical gradients to morph the gradient appropriately. Therefore, varying `fill` or `alpha` in non-linear coordinates still gives the previous error.
* The default `stat = "align"` in `geom_area()` does not interpolate the `fill` or `alpha` at break points, which is then given a single flat fill/alpha.
* In theory, this will also form a gradient between discrete fills if user overrides the default `group` aesthetic. At the `GeomRibbon$draw_group()` method, it is no longer possible to discern whether `fill` was continuous or discrete, so there is no convenient way around this.


thomasp85 (approved):
LGTM

PR #5703, opened 2024-02-23
Contour breaks function works in absence of `bins` and `binwidth`

### LINKED ISSUE #5686: Passing a function to `breaks=` in `stat/geom_contour` does not work (as long as both `bins=` and `binwidth=` are `NULL`)
When answering [this](https://stackoverflow.com/questions/77967083/geom-contour-breaks-greater-than-a-value-instead-of-a-fixed-value/77968185#77968185) question on SO I stumbled over an issue when passing a function to the `breaks=` argument of `geom_contour`, i.e. passing a function to `breaks=` has no effect when both `bins=` and `binwidth=` are `NULL`.

As can be seen from the minimal reproducible example below, instead of four contour lines we get the default lines corresponding to the breaks returned by `pretty(z_range, 10)`:
``` r
library(ggplot2)

# Does not work: Expect 4 contour lines
v +
  geom_contour(
    breaks = \(z_range, binwidth) {
      seq(z_range[1], z_range[2], length.out = 4)
    }
  )
```

I had a look at the source and the issue seems to be that when a function is passed to `breaks` but both `bins=` and `binwidth=` are `NULL` (which is the default)  `contour_breaks()` will return the default breaks:

https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/R/stat-contour.R#L178-L181

A work-around would be to set one of the arguments to a non-`NULL` value, i.e. after adding `binwidth=1` the breaks are computed according to the function passed to `breaks=`:

``` r
# Works if one sets binwidth or bins to a non-NULL value
v +
  geom_contour(
    binwidth = 1,
    breaks = \(z_range, binwidth) {
      seq(z_range[1], z_range[2], length.out = 4)
    }
  )
```

<sup>Created on 2024-02-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5686.

Briefly, `contour_breaks()` will use a user-provided function even if `bins` and `binwidth` are `NULL`.
A missing `binwidth` argument is defaulted to 1/10<sup>th</sup> the available range.

Reprex from #5686:

``` r
#> ℹ Loading ggplot2

  geom_contour(
    breaks = \(x, y) seq(x[1], x[2], length.out = 4)
  )
```

<sup>Created on 2024-02-23 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
Are there other binning stats where the old behaviour is present? I think it is important that we ensure all behaves similarly

PR #5714, opened 2024-02-27
Don't attempt to calculate secondary `breaks = NULL`

### LINKED ISSUE #5713: How to avoid warnings when adding a secondary axis without breaks or labels in `ggplot2 3.5.0`?
Prior to this release, I could use the secondary axis only to display an additional label, and not include breaks or labels. To do so, I could do the following, but, as can be seen, warnings are generated, and I would like to avoid these warnings since the users of my package will also see these warnings.

---

``` r
library(ggplot2)

  geom_point() +
  scale_y_continuous(sec.axis = dup_axis(name = "my secondary axis", breaks = NULL)) 
```

``` r

  geom_point() +
  scale_y_continuous(sec.axis = sec_axis(transform = I, name = "my secondary axis", breaks = NULL)) 

#> Warning in min(x): no non-missing arguments to max; returning -Inf
```

<sup>Created on 2024-02-27 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5713.

Briefly, in #5569 we changed how out-of-bounds breaks were handled, that caused a subtle difference that in secondary axes `NULL` breaks became `numeric(0)` breaks. In this PR, we now skip break calculation when the length is 0, which includes both cases.


thomasp85 (approved):
LGTM

PR #5724, opened 2024-02-28
Remove `palette` from manual scale docs

### LINKED ISSUE #5625: scale_color_manual does not support palette even though documentation suggests otherwise
Hi,

I've just stumbled upon issue 3182, having encountered exactly the same issue (briefly, the documentation of `scale_color_manual` suggests it should work with a `palette` (i.e. a function that returns colors) but it does not.

The original issue was marked as completed and closed in 2019 but in my opinion the documentation is still inaccurate.

I'd be happy to supply a PR if that'd be appreciated.

teunbrand:
This PR aims to fix #5625.

It just removes `palette` from the documentation that `scale_{aes}_manual()` inherits from `discrete_scale()`.

### DIFF
--- a/R/scale-manual.R
+++ b/R/scale-manual.R
@@ -12,7 +12,7 @@
 #' of aesthetics provided via the `aesthetics` argument.
 #'
 #' @inheritParams scale_x_discrete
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5725, opened 2024-02-28
Add facet examples to programming vignette

### LINKED ISSUE #5181: Programming with facets
Extracted from R4DS because we're trying to get down to a page limit. I think this would be good content for https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html

---

### Faceting

Unfortunately, programming with faceting is a special challenge, because faceting was implemented before we understood what tidy evaluation was and how it should work.
So you have to learn a new syntax.
When programming with facets, instead of writing `~ x`, you need to write `vars(x)` and instead of `~ x + y` you need to write `vars(x, y)`.
The only advantage of this syntax is that `vars()` uses tidy evaluation so you can embrace within it:

```{r}
# https://twitter.com/sharoz/status/1574376332821204999
foo <- function(x) {
  ggplot(mtcars, aes(x = mpg, y = disp)) +
    geom_point() +
    facet_wrap(vars({{ x }}))
}

foo(cyl)
```

As with data frame functions, it can be useful to make your plotting functions tightly coupled to a specific dataset, or even a specific variable.
For example, the following function makes it particularly easy to interactively explore the conditional distribution of `carat` from the diamonds dataset.

```{r}
#| fig.show: hide

# https://twitter.com/yutannihilat_en/status/1574387230025875457
density <- function(color, facets, binwidth = 0.1) {
  diamonds |> 
    ggplot(aes(x = carat, y = after_stat(density), color = {{ color }})) +
    geom_freqpoly(binwidth = binwidth) +
    facet_wrap(vars({{ facets }}))
}

density()
density(cut)
density(cut, clarity)
```

teunbrand:
This PR aims to fix #5181.

The surrounding text in the current vignette already covers to the surrounding text proposed in #5181, so I interpreted the issue as a request to include examples for facets. That is what this PR adds.

### DIFF
--- a/vignettes/ggplot2-in-packages.Rmd
+++ b/vignettes/ggplot2-in-packages.Rmd
@@ -58,8 +58,8 @@ To create any graphic using ggplot2 you will probably need to use `aes()` at lea
 ```{r}
 mpg_drv_summary <- function() {
   ggplot(ggplot2::mpg) + 
-    geom_bar(aes(x = drv)) + 
-    coord_flip()
+    geom_bar(aes(y = drv)) + 
+    facet_wrap(vars(year))
 }
 ```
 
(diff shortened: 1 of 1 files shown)


vignettes/ggplot2-in-packages.Rmd
```diff
@@ -58,8 +58,8 @@ To create any graphic using ggplot2 you will probably need to use `aes()` at lea
 ```{r}
 mpg_drv_summary <- function() {
   ggplot(ggplot2::mpg) + 
-    geom_bar(aes(x = drv)) + 
-    coord_flip()
```
teunbrand, line 62: Using the bidirectional geom feature here instead of deprecated `coord_flip()`.

thomasp85 (approved):
LGTM

PR #5739, opened 2024-03-04
Deal with `NULL` margins in legend

### LINKED ISSUE #5738: Error when setting `theme(text = element_blank())`
This seems like something that should be handled more elegantly:

``` r
library(ggplot2)

  geom_point() +
  theme(text = element_blank())
#> Error in as.unit(e1): object is not coercible to a unit
```

<sup>Created on 2024-03-04 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The error occurs in `position_margin()` used in the legend.

teunbrand:
This PR aims to fix #5738.

Briefly, when `theme(text = element_blank())`, it has no margin. This messes up some assumptions made in the legend that it would have a margin. In this PR, we just use 0-margins when this occurs.

### DIFF
--- a/R/guide-legend.R
+++ b/R/guide-legend.R
@@ -652,7 +652,8 @@ keep_key_data <- function(key, data, aes, show) {
   keep
 }
 
-position_margin <- function(position, margin = margin(), gap = unit(0, "pt")) {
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

PR #5743, opened 2024-03-05
Unknown theme elements throw warnings instead of errors

### LINKED ISSUE #5719: Should undefined theme elements throw a warning instead of an error?
This issue is a follow-up to #5702.
Instead of throwing errors, I think we should throw a warning and remove the offending element from the theme.

``` r
library(ggplot2)

  geom_point() +
  theme(foobar = "qux")
#> Error in `plot_theme()` at ggplot2/R/plot-build.R:178:3:
#> ! The `foobar` theme element is not defined in the element hierarchy.
```

<sup>Created on 2024-02-28 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5719.

Briefly, when declaring an unknown theme element, we throw a warning instead of error. The offending theme element is then removed from the theme.

This is more gentle if you make a typo or for us if we want to deprecate theme arguments.

Reprex from the issue:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  theme(foobar = "qux")
#> Warning in plot_theme(plot): The `foobar` theme element is not defined in the
#> element hierarchy.
```

<sup>Created on 2024-03-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


tests/testthat/test-theme.R
```diff
@@ -293,7 +293,7 @@ test_that("incorrect theme specifications throw meaningful errors", {
 test_that("element tree can be modified", {
   # we cannot add a new theme element without modifying the element tree
   p <- ggplot() + theme(blablabla = element_text(colour = "red"))
-  expect_snapshot_error(print(p))
```
thomasp85, line 296: Wouldn't this require an update of the snapshot

thomasp85 (approved):
Apart from the comment LGTM

tests/testthat/test-theme.R
teunbrand: The message is exactly the same so the snapshot file hasn't changed

thomasp85: 👍

PR #5747, opened 2024-03-05
Redocument `stat_summary_bin()`'s  `breaks` argument.

### LINKED ISSUE #3480: Issue with documentation of stat_summary(_bin) and question about default x-axis position
I recently [tweeted](https://twitter.com/Japhir/status/1158758152113983488) about an issue with `stat_summary` documentation, and here's the issue for it. 

I intended to use `stat_summary_bin()` with custom breaks with gaps in between on the x-axis. I've tried to illustrate below. From the documentation it was not clear whether this was possible/desired, mostly because I didn't understand what kind of function to pass to `binwidth` from the docs. 

Below I'll explain my thinking so that it hopefully highlights the issue with the docs:

I was wondering if I should I try to write a function which is conditional on x-axis values? That sounds like it should be easier to specify with `breaks`. Also, it looks like the fun.data should operate only on the y-value, and doesn't get passed the x-value. I also tried to define a new group variable using `cut` or `findInterval` and set binwidth, but this still draws lines (most right one) that span the outside of the x-axis values. 

So then, when looking at `breaks`, the docs say that they overrules `center` and `boundary`, but neither of those are defined in these docs and don't point to references elsewhere.

In the end I created a new tibble with the summarized values for plotting.

```
data:            ...              . ... ...        ...               .... .......... . ...

desired summary: |*|              |___*___|        |*|               |_________*_________|

actual summary:  |---*---| |---------*--------| |-----*-----| |-----------*--------------|
```

I created below reprex.
- black dashed lines represent the desired breaks
- red dotted l ines highlight the min/max of each of the intervals
- black pointranges show the default method of doing this
- black vertical lines highlight the desired x-axis value to plot this pointrange at
- red pointranges highlight my grouping attempt, which still draws outside the actual x-axis limits.

``` r
library(ggplot2)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(tibble)

desired_breaks <- c(0, 40, 90)

dat <- tibble(x = c(1:3, 5:10, 15:23, 25:27, 50:86),
              y = rnorm(length(x), mean = seq(1, 10, length.out = length(x)), sd = 6),
              grp = findInterval(x, vec = desired_breaks))

ggplot(dat, aes(x = x, y = y)) +
  geom_point(alpha = .3) +
  stat_summary_bin(geom = "pointrange", fun.data = "mean_cl_normal",
                   breaks = desired_breaks) +
  stat_summary_bin(aes(group = grp), geom = "pointrange",
                   fun.data = "mean_cl_normal",
                   binwidth = 6,
                   colour = "red") +
  geom_vline(xintercept = desired_breaks, linetype = 2) +
  geom_vline(xintercept = c(1, 27, 50, 86),
             colour = "red",
             linetype = 3) +
  geom_vline(xintercept = c((27 - 1) / 2 + 1, (86 - 50) / 2 + 50))
```

<sup>Created on 2019-08-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This made me think that the default option of plotting these summary values between the specified breaks is actually undesired, since it should probably be drawn in the middle (or median) of the range of x-values. I.e. if you specify breaks like `c(0, 50, 100)` but your data only goes up to 60, your last summary point will be drawn all the way on the right of the data.

Thanks and kind regards

may be related to #3297

teunbrand:
This PR aims to fix #3480.

Briefly, `stat_summary_bin()` inherits documentation from `stat_bin()`, but doesn't have `boundary` or `center`.
This PR removes the mention of these arguments from the docs, preventing confusion demonstrated in #3480.

### DIFF
--- a/R/stat-summary-bin.R
+++ b/R/stat-summary-bin.R
@@ -1,5 +1,7 @@
 #' @rdname stat_summary
 #' @inheritParams stat_bin
+#' @param breaks Alternatively, you can supply a numeric vector giving the bin
+#'   boundaries.
 #' @export
 stat_summary_bin <- function(mapping = NULL, data = NULL,
                              geom = "pointrange", position = "identity",
--- a/R/stat-summary-bin.R
+++ b/R/stat-summary-bin.R
@@ -1,7 +1,7 @@
 #' @rdname stat_summary
 #' @inheritParams stat_bin
 #' @param breaks Alternatively, you can supply a numeric vector giving the bin
-#'   boundaries.
+#'   boundaries. Overrides `binwidth` and `bins`.
 #' @export
 stat_summary_bin <- function(mapping = NULL, data = NULL,
                              geom = "pointrange", position = "identity",
--- a/man/stat_summary.Rd
+++ b/man/stat_summary.Rd
@@ -145,9 +145,8 @@ stories in your data.
 The bin width of a date variable is the number of days in each time; the
 bin width of a time variable is the number of seconds.}
 
-\item{breaks}{Alternatively, you can supply a numeric vector giving
-the bin boundaries. Overrides \code{binwidth}, \code{bins}, \code{center},
-and \code{boundary}.}
+\item{breaks}{Alternatively, you can supply a numeric vector giving the bin
+boundaries.}
 
 \item{na.rm}{If \code{FALSE}, the default, missing values are removed with
 a warning. If \code{TRUE}, missing values are silently removed.}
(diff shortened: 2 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5754, opened 2024-03-06
Swap underscore arguments for dot-case arguments

### LINKED ISSUE #5751: Some arguments are in snakecase
I noticed a few arguments in the newer functions have words separated by _'s instead of .'s
https://ggplot2.tidyverse.org/reference/coord_polar.html
https://ggplot2.tidyverse.org/reference/guide_axis_logticks.html

Not sure if this intentional? It doesn't seem applied consistently.

I'd vote for big breaking changes to make all arguments with words separated by _'s. 

But if ggplot2 is not prepared to do this, then I think it is prob better to just be consistent in the package with words separated by .'s

teunbrand:
This PR aims to fix #5751.

Briefly, we (read: I) introduced new functions, `coord_radial()` and `guide_axis_logticks()`, where the arguments used underscores. In this PR we're deprecating the underscore arguments and replace them with dot-case arguments.

To prevent similar mistakes in the future, I added a snapshot test that lists all exported functions with underscore arguments. If we inadvertently introduce new ones, that test should fail, like a canary in the coalmine.


thomasp85 (approved):
LGTM

PR #5758, opened 2024-03-07
Only sort numeric breaks in binned guides

### LINKED ISSUE #5152: Unsorted breaks in `guide_coloursteps()`
`guide_coloursteps()` seems to give incorrect guide when breaks are not sorted in increasing direction.

In the example below, we can see the coloursteps aren't right. The same happens with continuous scales, but binned scales come with coloursteps by default, so that reprex was more minimal. This doesn't only happen when the breaks are sorted in descending order as in the example below, but this generally seems to happen when they're not increasing.

``` r
library(ggplot2)

p <- ggplot(mpg, aes(displ, hwy, colour = cty)) +
  geom_point()

p + scale_colour_viridis_b(breaks = seq(30, 10, by = -5))
```

You'd expect the guide to look like this:

``` r
p + scale_colour_viridis_b(breaks = seq(10, 30, by = 5))
```

<sup>Created on 2023-01-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I think the fix would be a simple sorting step in the guide training.

### LINKED ISSUE #5757: guide_colorsteps() from ne ggplot2 version
I found a problem with `guide_colorsteps` after installing the last ggplot2 version. 

I'm not sure why, but the order of classes is not correct. Some of the extreme classes are showing up in the middle. Using the function example data it seems to work, but here it doesn't. 

```r
library(terra)
library(giscoR)
library(elevatr)
library(tidyverse)

ger <- gisco_get_countries(country = "Germany")

r <- get_elev_raster(ger, 5) 
r <- rast(r) |> crop(ger) |> mask(ger) 

names(r) <- "elev"

df <- as.data.frame(r, xy = T)

ggplot(df) +
  geom_tile(aes(x, y, 
                fill = cut(elev, c(-50, 0, 25, 50, 100, 250, 500, 1000, 1500, 2000, 2330)))) +
  geom_sf(data = ger, fill = NA) +
  scale_fill_viridis_d() +
  guides(fill = guide_colorsteps()) +
  labs(fill = NULL) +
  theme_void() +
  theme(legend.position = "bottom",
        legend.key.height = unit(.3, "lines"),
        legend.key.width = unit(3, "lines"))

```

teunbrand:
This PR aims to fix #5757.

Briefly, guides try to parse bins from the break information in a scale. To fix #5152, we automatically sorted the breaks. However, that should only happen for numeric breaks. This PR moves the sorting step to apply to numeric breaks only.

Reprex from issue:

``` r
#> ℹ Loading ggplot2

ggplot(mtcars, aes(mpg, drat, colour = cut(disp, c(90, 200, 300, 400, 500)))) +
  geom_point() +
  scale_colour_viridis_d(guide = guide_colorsteps())
#> Warning: Removed 4 rows containing missing values or values outside the scale range
#> (`geom_point()`).
```

<sup>Created on 2024-03-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5763, opened 2024-03-07
Avoid documenting `scale_name` with `@inheritDotParams discrete_scale`

olivroy:
To avoid some clutter in the .Rd files.


teunbrand (approved):
Thanks @olivroy for the contribution. This looks good to me!

PR #5764, opened 2024-03-11
Resolution of `geom_tile()` per panel

### LINKED ISSUE #5740: geom_tile() not working with facet_wrap() for different ranges of y-axes.
I found a problem when combining `geom_tile()` with `facet_wrap()` when the y-axes of the different facets have fairly different ranges.

I expected that both facets exhibit the same kind of plot, but the one with the greatest range on the y-axis looks odd. Note that creating separate plots (i.e., without using facets but using a subset of the data) make the tiles work for both plots.

```r
library(ggplot2)

data <- data.frame(x = rep(seq(0, 1, length.out = 11), each = (11) * 2),
                   y = c(seq(100, 200, length.out = 11), seq(0, 1, length.out = 11)),
                   z = rnorm(((11) ** 2 * 2)),
                   var = rep(c("T1", "T2"), each = 11))

ggplot(data) +
  geom_tile(aes(x = x, y = y, fill = z)) +
  facet_wrap(vars(var), scales = "free_y")
```

<sup>Created on 2024-03-04 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5740.

Briefly, when `width` and `height` are missing, these are inferred from the data.
In #5125, resolutions were computed per panel instead of over the whole layer.
This PR follows that approach but for `geom_tile()`.

Reprex from issue:

``` r
#> ℹ Loading ggplot2

data <- data.frame(x = rep(seq(0, 1, length.out = 11), each = (11) * 2),
                   y = c(seq(100, 200, length.out = 11), seq(0, 1, length.out = 11)),
                   z = rnorm(((11) ** 2 * 2)),
                   var = rep(c("T1", "T2"), each = 11))

ggplot(data) +
  geom_tile(aes(x = x, y = y, fill = z)) +
  facet_wrap(vars(var), scales = "free_y")
```

<sup>Created on 2024-03-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


teunbrand:
Quick to note to self: should update to reflect #5765 if that gets merged first.

thomasp85 (approved):
LGTM once you've updated the PR as per your last comment

teunbrand:
The resolution bit was fixed in 80382d9 as part of resolving the merge conflicts

PR #5765, opened 2024-03-11
Revert default `resolution()`, add argument

### LINKED ISSUE #5709: resolution(<mapped_discrete>) should not always return 1 (ggplot2 3.5.0)
I noticed a minor bug in ggdist under ggplot2 3.5.0, which is caused by a change in behavior that I believe is a regression: `resolution(<mapped_discrete>)` now always returns 1.

E.g. in 3.4:

```r
resolution(ggplot2:::new_mapped_discrete(c(0.2,0.4,0.6,0.8)))
## [1] 0.2
```

And in 3.5:

```r
resolution(ggplot2:::new_mapped_discrete(c(0.2,0.4,0.6,0.8)))
## [1] 1
```

I believe the old behavior is correct. `mapped_discrete` is applied after discrete values are scaled into (0,1), and when asking the resolution one wants to know the resolution in plot units, not in original data units. This would also be consistent with what `resolution()` does on continuous variables, as it would be impossible to return the resolution in original data units when passed a `numeric`, as that information is not retained.

From my side it's an easy fix to just cast to `numeric()` before using `resolution()`, but I am fairly sure I should not have to do that.

teunbrand:
This PR aims to fix #5709.

Briefly, it adds a new argument `discrete`. If `discrete = TRUE`, then `<mapped_discrete>` are treated as having a resolution of 1. If `discrete = FALSE` then these are treated just as `<numeric>`.


thomasp85 (approved):
LGTM

PR #3919, opened 2020-03-31
Fix bug in discrete position scales with continuous limits (fixes #3918)

paleolimbot:
This PR fixes the case where only continuous values were provided to a discrete scale, which was also provided continuous limits. I introduced this bug in #3380, which changed the way that discrete scales were expanded into continuous limits ([old logic](https://github.com/tidyverse/ggplot2/blob/v3.2.1/R/scale-discrete-.r#L111), [new logic](https://github.com/tidyverse/ggplot2/blob/master/R/scale-expansion.r#L204))

``` r
library(ggplot2)

df <- data.frame(x = 1:14, y = -2:-15)

ggplot(df, aes(x, y)) +
  scale_y_discrete(limits = -1:-16, labels = LETTERS[1:16])
```

<sup>Created on 2020-03-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Note that `is.discrete()` in this PR is linked to `is.discrete()` in `ScaleDiscretePosition$train()`...I don't think that the discrete scale training logic is likely to change, but the fix in this PR depends on that behaviour. The other approach would be to check `is.null(scale$range$range)`, which is inconvenient the way the expansion code is currently set up (but totally possible if anyone thinks  it matters).

Issue #3918, opened 2020-03-31
Discrete scale broken with negative limits


dhyou21:
I recently reinstalled my R/Rstudio due to the issue with one of the packages, and during this process, I also (1) downloaded and installed R development tools and libraries (clang and gfortran) from https://cran.r-project.org/bin/macosx/tools/ (2) downloaded and installed gfortran compiler for R from https://github.com/fxcoudert/gfortran-for-macOS/releases/tag/6.3 (3) uninstalled Java/JDK 14 and installing Java/JDK 11 version; and (4) ran “R CMD javareconf” comman on the terminal. All of these actions were taken to resolve the error associated with one of the packages requiring Java connection.

However, during theses resolution steps, something else seemed to be messed up because the existing function/package/script started working inappropriately (it worked before I tried the resolution steps above). So the problem is (and this problem is SEPARATE from the package problem that I was trying to solve with the resolution steps mentioned above): I run this script that utilizes several packages including ggplot2, tibble, dplyr, etc. to generate my graphs. Just several days ago when I tried to generate these graphs, the script worked fine (see below picture, on the left labeled as a "good example"). 

![image](https://user-images.githubusercontent.com/62915812/78048637-0069e180-7348-11ea-988e-f743f2d1e092.png)

But after all the resolution steps that I took as discussed above, the exact same script no longer works properly (see above picture, on the right side labeled as a "bad example" that I'm getting now). I used the exact same script, same packages to generate these graphs but now the graphs are being generated strangely…It seems like the tickmarks on the y axis are now not getting evenly spaced out throughout the whole y-axis and the problem starts at the stage where I create "Plot."

Does anyone have any idea what might have gone wrong during the resolution that could have damaged R (or its associated packages) to cause this problem? I want to go back to the ggplot2 that used to create that "good example"

The data that I used are basically table of different cell numbers in each well ("Number_of_Objects) in 384-well plate and I tried to depict the data using the heatmap approach as shown in the figure. The figure shows the collection of small rectangles corresponding to each well in 384-well plate and they are colored on gray gradient based on the cell density (darker, the greater density). The code that I used to plot this is:

```r
library("tidyverse")
library("stringr")
library("ggplot2")
library("ggpubr")

#Table corresponds to the data table of 384-well plate that will be analyzed#

Plot=ggplot(Table, aes(x=Column, y=-Row)) + theme_classic()+
    scale_x_discrete(limits=c(1:24), name="") + 
[...]

PR #5771, opened 2024-03-12
Palettes for discrete scales

### LINKED ISSUE #5770: Feature request: expose `palette` in discrete position scales
The idea behind exposing a palette to discrete position scales is that the palette can be used to set custom spacings between discrete variables. Currently the 'palette' of discrete position scales is essentially `seq_along()`, and the ask here is to open this up to other functions that can translate discrete variables to continuous positions.

To give an example, let's say I have the following plot with a fake continuous scale separating groups of clarity:

``` r
library(ggplot2)

clarity_position <- function(x) {
  as.integer(x) + c(0, 1, 1, 2, 2, 3, 3, 4)[match(x, levels(x))] 
}
ggplot(diamonds, aes(clarity_position(clarity), price, group = clarity)) +
  geom_boxplot() +
  scale_x_continuous(
    breaks = c(1, 3, 4, 6, 7, 9, 10, 12),
    labels = levels(diamonds$clarity)
  )
```

<sup>Created on 2024-03-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I'd like to express a similar plot in the following way:

``` r
ggplot(diamonds, aes(clarity, price)) +
  geom_boxplot() +
  scale_x_discrete(
    palette = \(x) c(1, 3, 4, 6, 7, 9, 10, 12)
  )
```

I'd imagine it just uses the discrete expansion rules and not show minor breaks in the panel grid.

teunbrand:
This PR aims to fix #5770.

Briefly, it adds a `palette` argument to discrete scales that can be used to customise the position discrete variables are placed in.
Demonstrations perhaps show off the functionality better than text descriptions.

The reprex from the issue works:

``` r
#> ℹ Loading ggplot2

p <- ggplot(diamonds, aes(clarity, price)) +
  geom_boxplot()

p + scale_x_discrete(palette = \(x) c(1, 3, 4, 6, 7, 9, 10, 12))
```

You can use the palette to invert the order:

``` r
p + scale_x_discrete(palette = \(x) -seq_along(x))
```

If two limits share a position, position adjustments treat these like a single position (invoking dodging in the example) but axis guides show duplicated labels.

``` r
p + scale_x_discrete(palette = \(x) c(1, 2, 2, 3, 3, 4, 4, 5))
```

Invalid palette output is checked:

``` r
p + scale_x_discrete(palette = \(x) x)
#> Error in `scale_x_discrete()`:
#> ! The `palette` function must return a <numeric> vector.

p + scale_x_discrete(palette = \(x) seq_along(x)[-1])
#> Error in `scale_x_discrete()`:
#> ! The `palette` function must return at least 8 values.
```
When mixing the discrete scale with continuous values in a different layer;

```r
p2 <- p + annotate("line", x = c(1, 4.5, 8), y = c(5000, 10000, 5000))
p2
```

The continuous values stay as-is

``` r
p2 + scale_x_discrete(palette = \(x) c(1, 3, 4, 6, 7, 9, 10, 12))
```

<sup>Created on 2024-03-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

There are at this moment two outstanding issues:

* Discrete non-position scales take `palette` functions that take `n` as argument. In this PR, palette functions take the limits as arguments. Should I let palettes take `n` as an argument instead? I think the benefit of the current approach is that one could design a lookup table as `palette` argument.
* I did not understand one test, see comment below.


tests/testthat/test-scale-expansion.R
```diff
@@ -116,6 +116,6 @@ test_that("expand_limits_scale_discrete() begrudgingly handles numeric limits",
       coord_limits = c(NA, NA),
       range_continuous = c(-15, -2)
     ),
-    c(-15, -2)
+    c(-16, -1)
```
teunbrand, line 119: I didn't understand this test, even looking through #3918 and #3919 where this was introduced.
Why should scale expansion take the narrower range here?
In any case, this PR essentially takes the `range()` of both ranges.

thomasp85 (approved):
LGTM. I dislike the `palette` name for this functionality, but I guess this is in line with the terminology we have ended up with

teunbrand:
>  I dislike the palette name for this functionality

Yeah I also don't think it is a great name, but it is consistent with the rest of the scales api

PR #5773, opened 2024-03-13
Silence length 1 aesthetics warning for `geom_blank()`

### LINKED ISSUE #5772: `geom_blank()` should silence 1-row warning
Let's say I want to ensure I include the value 0 in the x-axis of the plot and I might be inclined to use `geom_blank()` to achieve this (rather than just setting the scales).
This throws a warning:

``` r
library(ggplot2)

ggplot(mpg) +
  geom_point(aes(displ, hwy)) +
  geom_blank(aes(x = 0))
#> Warning in geom_blank(aes(x = 0)): All aesthetics have length 1, but the data has 234 rows.
#> ℹ Did you mean to use `annotate()`?
```

<sup>Created on 2024-03-13 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

As `geom_blank()` is almost *exclusively* used for its side-effect of expanding scales, and the mapping to data is meaningless as it shows nothing, it should not throw this warning.

teunbrand:
This PR aims to fix #5772.

It uses the mechanism we included to opt-out of the warning.

### DIFF
--- a/R/geom-blank.R
+++ b/R/geom-blank.R
@@ -36,5 +36,6 @@ geom_blank <- function(mapping = NULL, data = NULL,
 GeomBlank <- ggproto("GeomBlank", Geom,
   default_aes = aes(),
   handle_na = function(data, params) data,
+  check_constant_aes = FALSE,
   draw_panel = function(...) nullGrob()
 )


thomasp85 (approved):
LGTM

PR #5776, opened 2024-03-15
Mention static aesthetic caveat

### LINKED ISSUE #4821: Mismatched fill colour with geom_dotplot
When using `geom_dotplot` and specifying a `fill` aesthetic, we get a different output depending on whether `fill` is specified inside the function `aes` or as a separate vector. In the example below, I would expect all points below x~12 to be the same colour in both graphs.

``` r
library(tidyverse)
library(patchwork)

# Use 'dose' inside aes
gg1 <- ggplot(ToothGrowth, aes(x = len)) +
    geom_dotplot(aes(fill = factor(dose)),
                 binwidth = .5) +
    guides(fill = "none") # Remove legend

# Create separate vector for fill
dot_cols <- as.integer(factor(ToothGrowth$dose))
gg2 <- ggplot(ToothGrowth, aes(x = len)) +
    geom_dotplot(fill = dot_cols,
                 binwidth = .5)

gg1 + gg2
```

<sup>Created on 2022-04-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

As we can see, on the right, there are about 5 points between x=10 and x=12 that should be black but are actually red. 

As far as I can tell, this could be a consequence of the data being sorted with respect to x, which means that the rows of the dataset no longer match the elements of the vector `dot_cols`. As a proof of concept, I recreated the second graph but I sorted the data first (I also used the `group` aesthetic to try to get as close as possible to the original graph):

``` r
library(tidyverse)

# Repeat last graph but reorder data first
ToothGrowth2 <- arrange(ToothGrowth, len)
dot_cols2 <- as.integer(factor(ToothGrowth2$dose))
ggplot(ToothGrowth, aes(x = len, group = factor(dose))) +
    geom_dotplot(fill = dot_cols2, 
                 binwidth = .5)
```

<sup>Created on 2022-04-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Now all points below x~12 are black, as expected.

teunbrand:
This PR aims to fix #4821.

Briefly, the discussion in the linked issue suggested either better warnings or better documentation for the problem of passing a length 2+ vector as static aesthetic outside of `aes()`. I argued in favour of keeping current behaviour without warnings, so this PR adjusts the documentation instead. It tries to make clear that there are no guarantees about vectors passed as static aesthetics.

### DIFF
--- a/R/geom-point.R
+++ b/R/geom-point.R
@@ -40,7 +40,9 @@
 #'     value and apply to the layer as a whole. For example, `colour = "red"`
 #'     or `linewidth = 3`. The geom's documentation has an **Aesthetics**
 #'     section that lists the available options. The 'required' aesthetics
-#'     cannot be passed on to the `params`.
+#'     cannot be passed on to the `params`. Please note that while passing
+#'     unmapped aesthetics as vectors is technically possible, the order and
+#'     required length is not guaranteed to be parallel to the input data.
 #'   * When constructing a layer using
 #'     a `stat_*()` function, the `...` argument can be used to pass on
 #'     parameters to the `geom` part of the layer. An example of this is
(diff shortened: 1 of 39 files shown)


thomasp85 (approved):
LGTM

PR #5781, opened 2024-03-18
Update to `update_*_defaults()`

### LINKED ISSUE #4993: Reset original defaults with `update_geom_defaults()`

teunbrand:
This PR aims to fix #4993 and supersedes #5098.

Briefly:
1. There is a reset mechanism by setting `new = NULL`.
2. The update functions return the previous defaults, as is customary.

In contrast to #5098:
1. There is now 1 mechanism, instead of two separate ones.
2. Cache uses snake_case of class as index, which prevents having to use two caches and should be somewhat easier to maintain.


teunbrand:
TODO: look into feasibility of the suggestion here: https://github.com/tidyverse/ggplot2/issues/4993#issuecomment-2019944681

R/geom-defaults.R
```diff
@@ -29,27 +31,45 @@
...
+    }
+    invisible(new)
+
+  } else { # Update default aesthetics
+
+    old <- obj$default_aes
+    # Only update cache the first time defaults are changed
+    if (!index %in% ls(cache_defaults)) {
+      cache_defaults[[index]] <- old
+    }
```
thomasp85, line 67: Maybe overengineering it, but this is a rather unperformant way to check for existence in an environment as it ignores the whole hash structure. `exists()` are there to check for bindings in an environment

thomasp85 (approved):
apart from comment lgtm

PR #5783, opened 2024-03-19
`guide_coloursteps(even.steps = FALSE)` works with discrete data

### LINKED ISSUE #3877: even.steps = FALSE does not work when binning is done outside of scale_binned
`even.steps = FALSE` should produce a stretched `guide_colorsteps()` bar when the variable has uneven breaks. This works when binning is done inside scale_binned.

``` r
library(tidyverse)

mtcars %>% 
  mutate(mpg_cut = cut(mpg, c(seq(10, 25, 5), 35))) %>% 
  ggplot(aes(hp, mpg, color = mpg_cut)) +
  geom_point() +
  guides(color = guide_colorsteps(show.limits = TRUE, even.steps = FALSE))
```

<sup>Created on 2020-03-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

teunbrand:
This PR aims to fix #3877.

Briefly, when binning is done outside the scale with `cut()`, the coloursteps guide understood how to display this, but couldn't display this with uneven steps. This PR makes the guide work with uneven steps.

Reprex from the linked issue:

``` r
#> ℹ Loading ggplot2

mtcars |>
  transform(mpg_cut = cut(mpg, c(seq(10, 25, 5), 35))) |>
  ggplot(aes(hp, mpg, color = mpg_cut)) +
  geom_point() +
  guides(color = guide_colorsteps(show.limits = TRUE, even.steps = FALSE))
```

<sup>Created on 2024-03-19 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

@thomasp85 mentioned why this was unsupported before [here](https://github.com/tidyverse/ggplot2/issues/3877#issuecomment-597047227), but I didn't figure out why it was unsupported.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,8 @@
 * When legends detect the presence of values in a layer, `NA` is now detected
   if the data contains values outside the given breaks (@teunbrand, #5749).
 * `annotate()` now warns about `stat` or `position` arguments (@teunbrand, #5151)
+* `guide_coloursteps(even.steps = FALSE)` now works with discrete data that has 
+  been formatted by `cut()` (@teunbrand, #3877).
 
 # ggplot2 3.5.0
 
--- a/R/guide-bins.R
+++ b/R/guide-bins.R
@@ -326,8 +326,7 @@ GuideBins <- ggproto(
   }
 )
 
-parse_binned_breaks = function(scale, breaks = scale$get_breaks(),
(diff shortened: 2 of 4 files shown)


thomasp85 (approved):
Well, tbh I can't remember the reason either but it appears I was right in that it had to wait for the guide rewrite 😄

PR #5543, opened 2023-11-28
`theme()` supports splicing arguments

teunbrand:
This PR aims to fix #5542.

Briefly, `...` is moved to the first argument of the theme (which is technically a breaking change, I suppose), so that it can catch splicing operators. Then `find_args()`, which does the argument collection, just uses `list2()` to support the splicing.

Reprex from #5542:

``` r
devtools::load_all("~/packages/ggplot2/")
#> ℹ Loading ggplot2

red_axis <- list(
  axis.line = element_line(colour = "red"),
  axis.text = element_text(colour = "red"),
  axis.ticks = element_line(colour = "red"),
  axis.title = element_text(colour = "red")
)

ggplot(mpg, aes(displ, hwy)) +
  geom_point() +
  theme(!!!red_axis)
```

<sup>Created on 2023-11-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #5790, opened 2024-03-21
Update theme.R

olivroy:
Attempt to document better https://github.com/tidyverse/ggplot2/pull/5543

### DIFF
--- a/R/theme.R
+++ b/R/theme.R
@@ -194,7 +194,7 @@
 #'   switched (`unit`)
 #'
 #' @param ... additional element specifications not part of base ggplot2. In general,
-#'   these should also be defined in the `element tree` argument.
+#'   these should also be defined in the `element tree` argument. A list is also supported.
 #' @param complete set this to `TRUE` if this is a complete theme, such as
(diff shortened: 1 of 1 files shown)


R/theme.R
```diff
@@ -194,7 +194,7 @@
 #'   switched (`unit`)
 #'
 #' @param ... additional element specifications not part of base ggplot2. In general,
-#'   these should also be defined in the `element tree` argument.
+#'   these should also be defined in the `element tree` argument. A list is also supported.
```
teunbrand, line 197:
```suggestion
#'   these should also be defined in the `element tree` argument. [Splicing][rlang::splice] a list is also supported.
```
A bare list is not supported, but splicing a list is.

teunbrand:
Thanks for the PR! Should we mention the splicing option for `aes()` too?

teunbrand:
> I don't think it is worth bothering with aes().

Fair enough. If you could hit `devtools::document()` to populate the .Rd file, I think this should be good to go.

teunbrand:
\document

teunbrand (approved):
Thanks for the contribution!

PR #5798, opened 2024-03-25
Mention `size.unit` option

### LINKED ISSUE #5792: Update vignette for `geom_text(size.unit)` setting
The vignette [here](https://ggplot2.tidyverse.org/articles/ggplot2-specs.html#font-size) points out the conversion factor for `geom_text()`/`geom_label()`'s  `size` aesthetic. We should probably mention the `size.unit` setting first.

teunbrand:
This PR aims to fix #5792.

Since the diff is hard to read, I've added this line:

> In `geom_text()` and `geom_label()`, you can set `size.unit = "pt"` to use points instead of millimeters. In addition,

### DIFF
--- a/vignettes/ggplot2-specs.Rmd
+++ b/vignettes/ggplot2-specs.Rmd
@@ -283,7 +283,8 @@ ggplot(df, aes(1, x)) +
 
 ### Font size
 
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

PR #5799, opened 2024-03-25
Update aesthetics of length 1 warning

### LINKED ISSUE #5762: Confusing warning in `geom_segment()` with length 1 aesthetics
Hi,

In the latest version, a warning is thrown if aesthetics are length 1 while the data is not, hinting to use `annotate()` instead of a geom.

Of note, I could have missed it but I couldn't find this change in NEWS.

While this warning should be legitimate in most cases, there are some where it does not make much sense and that `annotate()` is not usable, in my case when using `geom_segment()` with summary functions.

It is thrown in the documentation ([https://ggplot2.tidyverse.org/reference/aes_position.html](https://ggplot2.tidyverse.org/reference/aes_position.html#:~:text=%23%20Geom_segment%20examples)).

One could argue that the documentation example could indeed be replaced with `annotate()`, so here is an example that cannot:

``` r
library(ggplot2) 
ggplot(mtcars) +
  aes(x=mpg, y=disp) +
  geom_point() +
  geom_segment(aes(x=min(mpg), xend=max(mpg), y=min(disp), yend=max(disp)))
#> Warning in geom_segment(aes(x = min(mpg), xend = max(mpg), y = min(disp), : All aesthetics have length 1, but the data has 32 rows.
#> i Did you mean to use `annotate()`?
```

``` r

ggplot(mtcars) +
  aes(x=mpg, y=disp) +
  geom_point() +
  annotate("segment", x=min(mpg), xend=max(mpg), y=min(disp), yend=max(disp))
#> Error in FUN(X[[i]], ...): only defined on a data frame with all numeric-alike variables
```

<sup>Created on 2024-03-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5762.

Briefly, it updates a warning that lead to some confusion based on a suggestion by Claus.

### DIFF
--- a/R/layer.R
+++ b/R/layer.R
@@ -334,8 +334,8 @@ Layer <- ggproto("Layer", NULL,
         && length(aes_n) > 0 && all(aes_n == 1) && n > 1) {
       cli::cli_warn(c(
         "All aesthetics have length 1, but the data has {n} rows.",
-        i = "Please consider using {.fn annotate} or providing this layer \\
-        with data with a single row."
+        i = "Please consider using {.fn annotate} or provide this layer \\
+        with data containing a single row."
       ), call = self$constructor)
     }
     check_aesthetics(evaled, n)
--- a/R/layer.R
+++ b/R/layer.R
@@ -334,7 +334,8 @@ Layer <- ggproto("Layer", NULL,
         && length(aes_n) > 0 && all(aes_n == 1) && n > 1) {
       cli::cli_warn(c(
         "All aesthetics have length 1, but the data has {n} rows.",
-        i = "Did you mean to use {.fn annotate}?"
+        i = "Please consider using {.fn annotate} or providing this layer \\
+        with data with a single row."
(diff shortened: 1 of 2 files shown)


R/layer.R
```diff
@@ -334,7 +334,8 @@ Layer <- ggproto("Layer", NULL,
         && length(aes_n) > 0 && all(aes_n == 1) && n > 1) {
       cli::cli_warn(c(
         "All aesthetics have length 1, but the data has {n} rows.",
-        i = "Did you mean to use {.fn annotate}?"
+        i = "Please consider using {.fn annotate} or providing this layer \\
+        with data with a single row."
```
thomasp85, line 338:
```suggestion
        i = "Please consider using {.fn annotate} or provide this layer \\
        with data containing a single row."
```

I think this reads better

thomasp85 (approved):
apart from wording LGTM

PR #5803, opened 2024-03-25
Add `ggplot_build()` method for <ggplot_built> class

### LINKED ISSUE #5800: Feature request: implement `ggplot_build.ggplot_build` S3 method.
Several getter functions, like `layer_data()`, call `ggplot_build()` on a plot.
For testing purposes, it might be more convenient to build a plot once and then extract relevant data/grobs/scales without having to build the plot every time.
Therefore, it can be convenient to have a `ggplot_build.ggplot_build` S3 method  that just returns the input unaltered.
That way, it should be safe to use `layer_data()` or somesuch with a pre-build plot as input.

teunbrand:
This PR aims to fix #5800.

Briefly, it does exactly what the title says and the method just returns input.
Due to this, we were allowed a small simplification to `get_guide_data()`.


thomasp85 (approved):
LGTM

PR #5814, opened 2024-03-28
Add section about width/height

### LINKED ISSUE #5813: Documentation: relate `width`/`height` to position aesthetics
This issue is a proposal to include a little bit more details on how the `x` and `width` pairs and `y` and `height` pairs of aesthetics interact. I've seen enough confusion on this topic that I think this warrants a little section in `?aes_position`.
Should basically point out:

* `x` and `y` are point-based locations
* `width` and `height` are length-based dimensions
* location-based aesthetics have scales and respond to transformation
* length-based aesthetics do not have scales and are applied after transformation
* Give concrete example that `x = 10, width = 2` equates to `xmin = 9, xmax = 11` under identity scales, but to `xmin = 1, xmax = 100` under log10-scales.

teunbrand:
This PR aims to fix #5813.

Briefly, it does as the title indicates in the `?aes_position` documentation.
If there is a better wording for this, I'll be glad to adapt this PR.

### DIFF
--- a/R/aes-position.R
+++ b/R/aes-position.R
@@ -18,6 +18,18 @@
 #' `xmiddle`, `middle`, `xupper`, `upper`, `x0` and `y0`. Many of these are used
 #' and automatically computed in [`geom_boxplot()`].
 #'
+#' ## Relation to `width` and `height`
+#'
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5815, opened 2024-03-28
Fix `linewidth` of `draw_key_polygon()`

### LINKED ISSUE #4852: Legend for `geom_col()`'s size does not correctly depict change in size
I'm not sure if this is related to issue #3729 at all, but here it is: when using the `size` aesthetic for a bar/col geometry, the corresponding legend does not match differences in sizes in the plot like e.g. a point geometry would. This is especially true if the `colour` aesthetic is also mapped to a variable from the dataset, as the outline completely disappears from the legend and therefore the audience might think that it's the darker area that depicts the size:

``` r
library(ggplot2)
# coloured outline (continuous)
exp_df <- data.frame(x = letters[1:4], y = 1:4)
ggplot(exp_df, aes(x, y, size = y, colour = y)) +
  geom_col()
```

``` r
# coloured outline (discrete)
ggplot(exp_df, aes(x, y, size = y, colour = x)) +
  geom_col()
```

``` r
# transparent fill
ggplot(exp_df, aes(x, y, size = y, colour = y)) +
  geom_col(alpha = 0)
```

``` r
# single colour
ggplot(exp_df, aes(x, y, size = y)) +
  geom_col(colour = "red", alpha = 0)
```

<sup>Created on 2022-05-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

There is a very slight different in size (2 pixels) between the first square and the three others in the legend, but that's it:

image

teunbrand:
This PR aims to fix #4852.

Briefly, `draw_key_polygon()` used to cap the `linewidth` aesthetic to some portion of the key size. Now that we can let the grobs contribute to key sizes, it is more straightforward to use that mechanism than have the legend display incorrect linewidths.

A reprex similar to the one in the issue:

``` r
#> ℹ Loading ggplot2

exp_df <- data.frame(x = letters[1:3], size = c(1, 4, 16))
ggplot(exp_df, aes(x, 1L, linewidth = size)) +
  geom_col(colour = "black", fill = "transparent") +
  scale_linewidth_identity(guide = "legend")
```

<sup>Created on 2024-03-28 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5838, opened 2024-04-12
Restore axis line when `breaks = NULL`.

### LINKED ISSUE #5816: MIssing axis when breaks = NULL
This is a repeat of #2983. When breaks = NULL in `scale_*`, the axis disappears altogether, although I expect the line to remain. #2983 was solved back then by #3257, but now the same incorrect behavior happens.

``` r
library(dplyr)
library(ggplot2)
data.frame(x = rnorm(100), y = rnorm(100)) %>% 
  ggplot(aes(x,y)) +
  geom_point() +
  scale_x_continuous(breaks = NULL) +
  theme_classic()
```

<sup>Created on 2024-03-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5816.

Briefly, it restores the behaviour pre-3.5.0 that `breaks = NULL` would omit ticks and labels but include the axis line.
In addition, that behaviour is now passed on to `guide_axis_theta()` and should be dealt with correctly when using `guide_axis_stack()`.

``` r
#> ℹ Loading ggplot2

  geom_point() +
  coord_radial() +
  scale_x_continuous(breaks = NULL) +
  guides(
    theta = guide_axis_stack("axis_theta", "axis_theta", spacing = unit(5, "mm"))
  ) +
```

<sup>Created on 2024-04-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5842, opened 2024-04-15
Clarify when `dpi` applies in `ggsave()`

### LINKED ISSUE #5497: SVG saved with a different size to PNG
I found a problem with `ggsave` - when I save an SVG with an explicit height and width in px, it is saved with smaller dimensions than specified.

I expected the SVG save would take the same commands of the PNG save.

```r
ggsave("test.png", width = (1600), height = (900), units = "px", limitsize = FALSE)
ggsave("test.svg", width = (1600), height = (900), units = "px", limitsize = FALSE)
```

The SVG is saved with dimensions of 512x288.  The SVG header shows:
 
` width='384.00pt' height='216.00pt' viewBox='0 0 384.00 216.00'`

I don't understand why it is using pt rather than px, nor why they are 3.125x smaller. I assume that it's because something is treating the pixels as being printed at 96dpi and then trying to make them 300dpi?

The [documentation](https://ggplot2.tidyverse.org/reference/ggsave.html) says that I should be able to specify pixels and that the dpi setting only affects raster graphics.

Any idea what's going on here? I can manually multiply the dimension, but I'd like to understand if the bug's with ggplot2 or me.

teunbrand:
This PR aims to fix #5497.

Briefly, `dpi` is applied when units are specified in pixels, not *just* for raster output types.
This edits the description to clarify this.

### DIFF
--- a/R/save.R
+++ b/R/save.R
@@ -37,7 +37,8 @@
 #' @param units One of the following units in which the `width` and `height`
 #'   arguments are expressed: `"in"`, `"cm"`, `"mm"` or `"px"`.
 #' @param dpi Plot resolution. Also accepts a string input: "retina" (320),
-#'   "print" (300), or "screen" (72). Applies only to raster output types.
+#'   "print" (300), or "screen" (72). Only applies when converting pixel units,
+#'   as is typical for raster output types.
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5855, opened 2024-04-22
More sorting options for `facet_wrap()`

### LINKED ISSUE #5212: Feature request: allow blank panels of `facet_wrap()` to be at the beginning
# Existing behavior

When using `facet_wrap()` and there are fewer levels than will completely fill the created grid, the last panels will be left blank. Which panels are considered "last" depends on the values of the parameters `dir` and `as.table`.

# Requested behavior

The ability to have the first panels blank rather than the last. "End-justified" rather than "begin-justified" to put it another way.

# Examples

Drawing the base of the example from the examples in the `facet_wrap()` help page.

```r
library("ggplot2")

```

Since I'm only interested in the arrangement of the panels, I'm just going to summarize the output with a text grid showing where each panel is. `X` represents a blank spot.

```r
# Existing behavior
p + facet_wrap(vars(class))
# 123
# 456
# 7XX
p + facet_wrap(vars(class), dir = "v")
# 147
# 25X
# 36X
p + facet_wrap(vars(class), as.table = FALSE)
# 7XX
# 456
# 123
p + facet_wrap(vars(class), dir = "v", as.table = FALSE)
# 741
# X52
# X63
```

As a separate note, I would never have guessed what the combination of `dir = "v"` and `as.table = FALSE` would have given me. I would have assumed it would have been:

```r
# 36X
# 25X 
# 147
```

But that's not the point of this feature request.

```r
# Desired behavior
p + facet_wrap(vars(class), justify = "end")
# XX1
# 234
# 567
p + facet_wrap(vars(class), dir = "v", justify = "end")
# X25
# X36
# 147
p + facet_wrap(vars(class), as.table = FALSE, justify = "end")
# 567
# 234
# XX1
p + facet_wrap(vars(class), dir = "v", as.table = FALSE, justify = "end")
# 52X
# 63X
# 741
```

# Implementation suggestions

I have demonstrated these by adding a `justify` argument to `facet_wrap()`, but I am not recommending that as an implementation. Primarily, adding another argument just increases the number of combination of arguments that have to be accounted for and tested.

A suggestion which may be better is a scrapping of the `as.table` argument and replacing it with an argument which specifies which corner to begin placing panels in (values "ul" for upper left, "ll" for lower left, "ur" for upper right, "lr" for lower right). The existing `dir` argument would then specify which direction panels were to go next. For example:

```r
p + facet_wrap(vars(class), start = "ul", dir = "h") # the defaults
# 123
# 456
# 7XX
p + facet_wrap(vars(class), start = "ul", dir = "v")
# 147
# 25X
# 36X
p + facet_wrap(vars(class), start = "ll", dir = "h") 
# 7XX
# 456
# 123
p + facet_wrap(vars(class), start = "ll", dir = "v")
# 36X
# 25X
# 147
p + facet_wrap(vars(class), start = "ur", dir = "h") 
# 321
# 654
# XX7
p + facet_wrap(vars(class), start = "ur", dir = "v")
# 741
# X52
# X63
p + facet_wrap(vars(class), start = "lr", dir = "h") 
# XX7
# 654
# 321
p + facet_wrap(vars(class), start = "lr", dir = "v")
# X63
# X52
# 741
```

These 8 combinations map to the 8 combinations of `dir`/`as.table`/`justify`, but sometimes requiring the order of the faceting variable to be reversed.

teunbrand:
This PR aims to fix #5212.

Briefly, it adds more options for the `dir` argument to layout the panels in various ways. The `as.table` argument gets absorbed into `dir` when the previous way of declaring `dir = "v"` or `dir = "h"` is used.

The new options for `dir` are:
* `"lt"`: start in the top-left, start filling left-to-right.
* `"tl"`: start in the top-left, start filling top-to-bottom.
* `"lb"`: start in the bottom-left, start filling left-to-right.
* `"bl"`: start in the bottom-left, start filling bottom-to-top.
* `"rt"`: start in the top-right, start filling right-to-left.
* `"tr"`: start in the top-right, start filling top-to-bottom.
* `"rb"`: start in the bottom-right, start filling right-to-left.
* `"br"` start in the bottom-right, start filling bottom-to-top

Using `as.table` with any of the options above has no effect.

As an example where blank panels are set at the beginning:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  facet_wrap(~ class, dir = "rb")
```

<sup>Created on 2024-04-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # ggplot2 (development version)
 
+* `facet_wrap()` has new options for the `dir` argument to more precisely
+  control panel directions (@teunbrand, #5212)
 * When facets coerce the faceting variables to factors, the 'ordered' class
   is dropped (@teunbrand, #5666).
 * `coord_map()` and `coord_polar()` throw informative warnings when used
--- a/R/facet-wrap.R
+++ b/R/facet-wrap.R
@@ -23,7 +23,12 @@ NULL
 #'   either of the four sides by setting \code{strip.position = c("top",
 #'   "bottom", "left", "right")}
 #' @param dir Direction: either `"h"` for horizontal, the default, or `"v"`,
-#'   for vertical.
+#'   for vertical. When `"h"` or `"v"` will be combined with `as.table` to
+#'   set final layout. Alternatively, a combination of `"t"` (top) or
+#'   `"b"` (bottom) with `"l"` (left) or `"r"` (right) to set a layout directly.
+#'   These two letters give the starting position and the first letter gives
+#'   the growing direction. For example `"rt"` will place the first panel in
+#'   the top-right and starts filling in panels right-to-left.
 #' @param axes Determines which axes will be drawn in case of fixed scales.
 #'   When `"margins"` (default), axes will be drawn at the exterior margins.
 #'   `"all_x"` and `"all_y"` will draw the respective axes at the interior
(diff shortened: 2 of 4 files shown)


thomasp85 (approved):
LGTM - Maybe add an example with the new features

teunbrand:
Good idea thanks!

PR #5867, opened 2024-04-26
Interpreter/wrapper for `gpar()`

### LINKED ISSUE #5866: Standardising calls to `gpar()`
Throughout ggplot2's codebase, there are various calls to `grid::gpar()`.
There are several operations that happen almost ubiquitously, like the following `gpar(lwd = len0_null(linewidth * .pt))`.
The suggestion here is to write a wrapper for `grid::gpar()` that automatically applies such adjustments and ensures the absence of 0-length vectors, which I believe would result in cleaner code.

teunbrand:
This PR aims to fix #5866.

Briefly, a new helper function `ggpar()` is used to wrap `grid::gpar()` while applying ggplot's interpretation of graphical parameters.

It simplifies several things:
* We don't have to protect against 0-length vectors anymore with `len0_null()`.
* It automatically applies the `* .pt` modification to `lwd`.
* It applies the calculation for point `stroke` and `size` via `stroke` and `pointsize` arguments.

There are two cases where this isn't used:
1. When a default argument is an empty `gpar()`.
2. In `sf_grob()` which has a mixed interpretations of for example `lwd`, where in some cases it is `linewidth` and in other cases it can be `stroke`.

It is exported but marked as internal function so that extension developers can have access to it as well.


thomasp85 (approved):
LGTM

davidhodge931:
The `gg*` prefix is being used by a few extension packages to refer to a wrapper around a lot of ggplot2 code to print a plot as a whole. Such packages include {ggsstatsplot},  {ggpubr}, and {ggblanket}. 

I think it'd be cool if possible to keep this distinction as much as possible - so that a function prefixed as `gg` typically makes a plot (apart from the `ggplot_build` etc functions). This is good for the intuition of the user, and the use of autocomplete.

Don't know whether you'd consider renaming this function for that reason, @teunbrand ? E.g. `gpar2`

teunbrand:
We had already renamed it from `ggpar` to `gg_par` precisely because of a namespace clash with {ggpubr}. I'm only willing to consider changing the name if it becomes better, either more self-descriptive or shorter, whereas `gpar2()` is clearly a worse name for a function.

davidhodge931:
I think ggplot2::gpar2 is clearer that it is ggplot2's version of gpar. But this stuff is very subjective though

PR #5868, opened 2024-04-26
Be clearer when the `fill` aesthetic applies in points

### LINKED ISSUE #4358: Issue a message when fill aesthetic is used with GeomPoint and shape is not "filled"
I've spend an hour or so of `ggplot2` debugging to find out that `fill` aesthetics is simply ignored when `shape` of the point is not 21:25, i.e., not "filled". `grid::grid.point`, which is used, documents this case quite well, but `ggplot2` gives a hint only through the example that could be "hidden" from the user that struggles with, e.g., boxplot ouliers fill modification (as in my case).

I would hugely appreciate to get warning/message that inform the user he or she is demanding something that won't work as intended, like it is the case when I use `scale_alpha_*` with continuous variable.

teunbrand:
This PR aims to fix #4358.

It just rephrases a sentence in the vignette and adds a note to the `geom_point()` aesthetics section.

### DIFF
--- a/R/geom-point.R
+++ b/R/geom-point.R
@@ -27,7 +27,7 @@
 #' `geom_point(alpha = 0.05)`) or very small (e.g.
 #' `geom_point(shape = ".")`).
 #'
-#' @eval rd_aesthetics("geom", "point")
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #5871, opened 2024-04-29
List `"centre"` as option for `legend.box.just`

### LINKED ISSUE #4196: documentation: values list for legend.box.just appears incomplete
The current documentation for [`theme()`](https://ggplot2.tidyverse.org/reference/theme.html) states

**legend.box.just** justification of each legend within the overall bounding box, when there are multiple legends ("top", "bottom", "left", or "right")

I happened to need `theme(legend.box.just = "center")` so I tried it anyway and found it behaved as expected when centering two legends (an alpha legend and a merged color plus linetype legend) within a bottom legend box.

So life is good. But it would be useful to know if this just signals the documentation is a bit out of date or if I'm relying on a behavior that's formally unsupported.

Version; ggplot 3.3.2

teunbrand:
This PR aims to fix #4196.

It just lists `"centre"`/`"center"` as an option, which was already accepted, just not documented as such.

### DIFF
--- a/R/theme.R
+++ b/R/theme.R
@@ -110,8 +110,8 @@
 #' @param legend.box arrangement of multiple legends ("horizontal" or
 #'   "vertical")
 #' @param legend.box.just justification of each legend within the overall
-#'   bounding box, when there are multiple legends ("top", "bottom", "left", or
-#'   "right")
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #5878, opened 2024-05-02
Default labels from attributes (option 1)

### LINKED ISSUE #4631: Use label-attribute as default axis/legend title
Currently, axis and legend titles default to the column names of the data.frame. There are several packages that provide functions to attach more comprehensive variable labels to data.frame columns (e.g., `Hmsic`, `tinylabels`, `labelled`, and `sjlabelled`). While these packages differe with respect to the specific implementation, all implementations are compatible in the sense that they attach the variable labels to the data.frame colums via the `label` attribute. I think it would be useful to use such labels, if available, as they can probably be expected to be more informative than the bare column names.

Consider the following example. First, I'll create some labelled data (I think `tinylabels` provides the safest implementation, but any of the below will work).

~~~r
library("dplyr")
library("ggplot2")

# Set up data & labels using {tinylabels}
library("tinylabels")
mtcars2 <- mtcars |>
  mutate(gear = factor(gear)) |>
  label_variables(
    wt = "Weight (1000 lbs)",
    gear = "Gears",
    vs = "Engine"
  )

# # Set up data & labels using {Hmisc}
# library("Hmisc")
# mtcars2 <- mtcars |>
#   mutate(gear = factor(gear)) |>
#   within({
#     label(gear) <- "Gears"
#     label(wt) <- "Weight (1000 lbs)"
#     label(vs) <- "Engine"
#   })
# 
# # Set up data & labels using {sjlabelled}
# library("sjlabelled")
# mtcars2 <- mtcars |>
#   mutate(gear = factor(gear)) |>
#   var_labels(
#     wt = "Weight (1000 lbs)",
#     gear = "Gears",
#     vs = "Engine"
#   )
~~~

For demonstration purposes I slightly change `ggplot_add.labels` to use the `label` attribute, if available:

~~~r
ggplot_add.labels <- function (object, plot, object_name) {
  object <- add_variable_labels(object, plot) # Newly added code
  ggplot2::update_labels(plot, object)
}

add_variable_labels <- function(labels, plot) {
  vars <- sapply(plot$mapping, function(x) rlang::as_name(rlang::f_rhs(x)))
  
  if(length(vars) == 0) {
    return(labels)
  }
  
  variabel_labels <- lapply(plot$data, attr, "label")[vars]
  
  # Add variable labels
  to_add <- !names(vars) %in% names(labels)
  
  for(i in vars[to_add]) {
    if(!is.null(variabel_labels[[i]])) {
      labels[[names(vars[vars == i])]] <- variabel_labels[[i]]
    }
  }
  
  labels
}

assignInNamespace("ggplot_add.labels", ggplot_add.labels, "ggplot2")
~~~

With this change axis and legend labels will default to the `labels` attribute when `labs()` is called:

~~~r
p <- ggplot(mtcars2, aes(x = wt, y = mpg, colour = gear)) +
  geom_point()

# Standard behavior (uses column names)
p 

# Uses labels, where available
p + labs()

# Overwrite defaults
p + labs( 
  y = "Fuel economy (mpg)"
  , color = "Bears"
)
~~~

I think this would be very nice feature and I'd be happy to take a stab at it.

If this is of interest, I see three options for a proper implementation:

1. Always default to using labels (after some exploration, the required changes seem managable).
2. Only default to using labels when `labs()` is called (pretty much what I have implemented above).
3. Only default to using labels when `labs()` is called and labels are requested (e.g., `use_labels = TRUE`).

There are probably other sensible options that I'm not seeing. Either way, I'd be very interested to know if this would be a PR you would be willing to consider and in any thoughts you may have on this.

teunbrand:
This PR aims to fix #4631 and it competes with #5879.

Briefly, an attempt is made to derive default labels from the 'label' attribute.

Less brief,  `make_labels()` now accepts a `data` argument. When there are simple symbolic mappings, these are evaluated and searched for a label attribute. There are some limitations with this approach that are detailed below.

To give an example we can give the `mtcars` dataset some label attributes.
Notice in the plot below that all default labels are derived from the attribute, regarless of whether it comes from the global mapping, the layer mapping or data pronoun use.

``` r
#> ℹ Loading ggplot2

df <- mtcars
attr(df$mpg,  "label") <- "Miles per gallon"
attr(df$disp, "label") <- "Displacement"
attr(df$drat, "label") <- "Rear axle ratio"

ggplot(df, aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat))
```

When we use a standalone `+ aes(...)` it also yields the exact same plot.

``` r
ggplot(df) +
  aes(mpg, .data$disp) +
  geom_point(aes(colour = drat))
```

As for the limitations, here are some that I noticed.
When using layer data, global aesthetics do not get searched.

``` r
ggplot(mapping = aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat), data = df)
```

Also, when the data is a function, the labels cannot be retrieved.
The data also isn't `fortify()`'ed or in any way preprocessed.

``` r
ggplot(df, aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat), data = ~ subset(.x, mpg > 20))
```

<sup>Created on 2024-05-02 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The labels are also baked in at the time the layers are added, so doing the following:

```r
p <- ggplot(mtcars, aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat))
p %+% df
```

Does not give the attribute labels.

These limitations could be resolved if the labels were derived in the `ggplot_build()` step, but as that function is somewhat sacred, I tried this approach first. Also, as this does not have an on/off switch, this PR will break some plots (but hopefully for the better).

Despite these limitations, I think this is an improvement over the current situation. See #5879 for a second option with fewer limitations.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Default labels are now derived from a label attribute when available 
+  (@teunbrand, #4631).
 * (Internal) Applying defaults in `geom_sf()` has moved from the internal 
   `sf_grob()` to `GeomSf$use_defaults()` (@teunbrand).
 * `facet_wrap()` has new options for the `dir` argument to more precisely
--- a/R/aes-evaluation.R
+++ b/R/aes-evaluation.R
@@ -322,7 +322,9 @@ strip_stage <- function(expr) {
 
 # Convert aesthetic mapping into text labels
 make_labels <- function(mapping, data = NULL) {
-  data <- data %|W|% NULL
+  if (is.waive(data) || is.function(data)) {
+    data <- NULL
+  }
   default_label <- function(aesthetic, mapping) {
     # e.g., geom_smooth(aes(colour = "loess")) or aes(y = NULL)
     if (is.null(mapping) || is.atomic(mapping)) {
--- a/R/aes-evaluation.R
+++ b/R/aes-evaluation.R
@@ -321,7 +321,8 @@ strip_stage <- function(expr) {
 }
 
 # Convert aesthetic mapping into text labels
-make_labels <- function(mapping) {
+make_labels <- function(mapping, data = NULL) {
+  data <- data %|W|% NULL
   default_label <- function(aesthetic, mapping) {
     # e.g., geom_smooth(aes(colour = "loess")) or aes(y = NULL)
     if (is.null(mapping) || is.atomic(mapping)) {
@@ -331,6 +332,10 @@ make_labels <- function(mapping) {
     mapping <- strip_dots(mapping, strip_pronoun = TRUE)
     if (is_quosure(mapping) && quo_is_symbol(mapping)) {
       name <- as_string(quo_get_expr(mapping))
+      if (!is.null(data) && name %in% names(data)) {
+        value <- eval_tidy(mapping, data = data)
+        name <- attr(value, "label", exact = TRUE) %||% name
+      }
     } else {
       name <- quo_text(mapping)
       name <- gsub("\n.*$", "...", name)
--- a/R/labels.R
+++ b/R/labels.R
@@ -16,6 +16,24 @@ update_labels <- function(p, labels) {
   p
 }
 
+label_from_layer <- function(layer, plot) {
+  mapping <- make_labels(layer$mapping)
+  default <- lapply(
+    make_labels(layer$stat$default_aes),
+    function(l) {
+      attr(l, "fallback") <- TRUE
+      l
+  })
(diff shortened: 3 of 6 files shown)


olivroy:
I really like this! With the newest support for the [label attribute in RStudio](https://github.com/rstudio/rstudio/pull/14245), I am taking even more advantage of it. would be great if ggplot2 used it too!

teunbrand:
Closing this in favour of #5879

PR #5079, opened 2022-12-03
Alt text label can be function

teunbrand:
This PR aims to fix #4795, by allowing `alt` to be a function that takes the plot as input.

``` r
devtools::load_all("~/packages/ggplot2")
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(class)) +
  geom_bar() +
  labs(alt = ~ BrailleR::VI(.x)$text)

cat(get_alt_text(p), sep = "\n")
#> This is an untitled chart with no subtitle or caption.
#> It has x-axis 'class' with labels 2seater, compact, midsize, minivan, pickup, subcompact and suv.
#> It has y-axis 'count' with labels 0, 20, 40 and 60.
#> The chart is a bar chart with 7 vertical bars.
#> Bar 1 is centered horizontally at 2seater, and spans vertically from 0 to 5.
#> Bar 2 is centered horizontally at compact, and spans vertically from 0 to 47.
#> Bar 3 is centered horizontally at midsize, and spans vertically from 0 to 41.
#> Bar 4 is centered horizontally at minivan, and spans vertically from 0 to 11.
#> Bar 5 is centered horizontally at pickup, and spans vertically from 0 to 33.
#> Bar 6 is centered horizontally at subcompact, and spans vertically from 0 to 35.
#> Bar 7 is centered horizontally at suv, and spans vertically from 0 to 62.
```

<sup>Created on 2022-12-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #5879, opened 2024-05-02
Default labels from attributes (option 2)

### LINKED ISSUE #4631: Use label-attribute as default axis/legend title

### LINKED ISSUE #5894: Default labs where geom when the aesthetic is in both the ggplot and geom_* call
Default labs are not created based on aesthetics in the geom when the aesthetic is in both the ggplot and geom_* call 
``` r
library(tidyverse)
library(palmerpenguins)

penguins |> 
  ggplot(aes(x = flipper_length_mm, y = body_mass_g, colour = sex)) +
  geom_point(aes(x = bill_depth_mm, y = bill_length_mm, colour = species))
#> (`geom_point()`).
```

<sup>Created on 2024-05-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #4631 and fix #5894 and competes with #5878.

Briefly, this is also an attempt to derive default labels from the 'label' attribute.

In contrast to #5878, this resolves labels in the `ggplot_build()` stage, thereby circumventing some limitations of that approach discussed yonder. It also feels 'cleaner', as there is just a single place where default labels are determined and we have no need to moonlight in `ggplot_add()` methods (notable exception is  `ggplot_add.labels()` for obvious reason that these are user specified non-default labels). The drawback is that we desecrate `ggplot_build()`.

Let's demonstrate using plots that are limitations in the other PR.
Labels are not baked in at construction time, so `%+%` data replacement works as intended.

``` r
#> ℹ Loading ggplot2

df <- mtcars
attr(df$mpg,  "label") <- "Miles per gallon"
attr(df$disp, "label") <- "Displacement"
attr(df$drat, "label") <- "Rear axle ratio"

p <- ggplot(mtcars, aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat))
p %+% df
```

Naively, it doesn't appear to work when `data` is not a data.frame.

``` r
ggplot(df, aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat), data = ~ subset(.x, mpg > 20))
```

However, this is just because `subset()` drops attributes. If we use `dplyr::filter()` that doesn't drop attributes, it works just fine.

``` r
ggplot(df, aes(mpg, .data$disp)) +
  geom_point(aes(colour = drat), data = ~ dplyr::filter(.x, mpg > 20))
```

<sup>Created on 2024-05-02 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The choice is thus:
* #5878 with limitations, but not touching `ggplot_build()`
OR
* #5879 without (known) limitations, but touching `ggplot_build()`


teunbrand:
This PR had an unexpected interaction with `get_alt_text()` when the alt-text was a function (introduced in #5079).
The problem was that alt-text functions could no longer read out plot labels before the plot is build, so the solution here is to apply `get_alt_text()` to build plots.

thomasp85 (approved):
LGTM

PR #5887, opened 2024-05-10
Themes accept header font family

### LINKED ISSUE #5886: Have header font be part of theme specification
Currently the theme constructors takes a base font argument to set the font for the titles/headers.

This ties into #5833 as we should also keep at least two fonts in the geom defaults inside themes

teunbrand:
This PR aims to fix #5886.

Briefly, all themes gain a `header_family` argument that defaults to `base_family`. In addition, `plot.subtitle` and `plot.caption` no longer inherit from `title` but the root `text` element to follow https://github.com/tidyverse/ggplot2/issues/5886#issuecomment-2102552944.

A demonstration:

``` r
#> ℹ Loading ggplot2

  labs(
    title = "Fuel efficiency",
    subtitle = "Described for 234 cars from 1999 and 2008.",
    caption = "Source: U.S. Environmental Protection Agency",
    tag = "A"
  ) +
  theme_gray(header_family = "Impact")
```

<sup>Created on 2024-05-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85:
I kind of understand why you want to change the inheritance of subtitle and caption but I'm afraid it would throw users off (especially for subtitle). I now regret we didn't name it "description" or something like that

teunbrand:
So should we revert the change in inheritance and manually set the header font to the relevant elements in `theme_*()` functions?

thomasp85:
I'm not sure, but I think it would be best at least for subtitle. I'm fine with caption not inheriting from title, I think that was a wrong choice from the start

teunbrand:
After some thought, I think it would be best to have the `header_family = NULL` be the default.
That way, you don't bake in `header_family = ""`, which is convenient when you're looking to globally change the text font.

Example with the current PR, notice the titles remaining sans serif:

``` r
#> ℹ Loading ggplot2

  labs(
    title = "Fuel efficiency",
    subtitle = "Described for 234 cars from 1999 and 2008.",
    caption = "Source: U.S. Environmental Protection Agency",
    tag = "A"
  ) +
  theme_gray()

p + theme_gray() + 
  theme(text = element_text(family = "Times New Roman"))
```



However, this all resolves nicely when the header font is `NULL`:

``` r
p + theme_gray(header_family = NULL) +
  theme(text = element_text(family = "Times New Roman"))
```



<sup>Created on 2024-05-21 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

thomasp85:
I'm not sure tags should inherit from title... Usually you want these pretty unobtrusive and subdued

teunbrand:
It also isn't immediately obvious to me why tags should have title style. I'm happy to swap their inheritance to the root text element if you agree.

thomasp85:
yeah, let's change that

teunbrand:
Alright, it now looks like so:

``` r
#> ℹ Loading ggplot2

  labs(
    title = "Fuel efficiency",
    subtitle = "Described for 234 cars from 1999 and 2008.",
    caption = "Source: U.S. Environmental Protection Agency",
    tag = "A"
  ) +
  theme_gray(header_family = "Impact")
```



<sup>Created on 2024-07-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM

PR #5888, opened 2024-05-11
Stop CI on R 3.6

yutannihilation:
Now that `oldrel-4` refers to R 4.0, I think this dedicated case for R 3.6 can be removed.


teunbrand (approved):
Good call!

PR #5900, opened 2024-05-21
Restore `as.table`

### LINKED ISSUE #5898: Fallback for `facet_wrap(as.table)`
In #5855, we internally removed `as.table` and resolved it in the constructor.
Reverse dependency checks revealed that several facet extensions rely on `as.table` to be present.
It would be nice to have a fallbacks in place to not trigger errors in extensions.

teunbrand:
This PR aims to fix #5898.

Briefly, some extensions require `as.table` to be present, which this PR restores. Instead of incorporating `as.table` in `dir` in the constructor, we now do this in `wrap_layout()`.

It pains me a little bit, as I was quite pleased that we could get rid of `as.table` as a parameter before and now we have to bring back this vestigial parameter.

### DIFF
--- a/R/facet-wrap.R
+++ b/R/facet-wrap.R
@@ -115,13 +115,6 @@ facet_wrap <- function(facets, nrow = NULL, ncol = NULL, scales = "fixed",
                        axis.labels = "all") {
   scales <- arg_match0(scales %||% "fixed", c("fixed", "free_x", "free_y", "free"))
   dir <- arg_match0(dir, c("h", "v", "lt", "tl", "lb", "bl", "rt", "tr", "rb", "br"))
-  if (nchar(dir) == 1) {
-    dir <- base::switch(
-      dir,
-      h = if (as.table) "lt" else "lb",
-      v = if (as.table) "tl" else "tr"
-    )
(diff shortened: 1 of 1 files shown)


teunbrand:
A reprex with this PR:

``` r
#> ℹ Loading ggplot2
library(deeptime)

df <- data.frame(x = 1:10, y = 1:10, period = c("Permian", "Triassic"))
ggplot(df) +
  geom_point(aes(x, y)) +
  facet_wrap_color(vars(period), colors = periods)
```



<sup>Created on 2024-05-21 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Whereas on main, the same code gives:

```r
#> Error in `data_frame()` at ggplot2/R/utilities.R:351:16:
#> ! Can't recycle `ROW` (size 0) to size 2.
```

thomasp85:
could we deprecate it gracefully instead of committing to having it around for eternity?

teunbrand:
I'd like to but this is proving to be a bit more finnicky that I anticipated.

teunbrand:
closing this in favour of #5907

PR #5902, opened 2024-05-22
`after_*()` length checks

### LINKED ISSUE #5901: Length check for `after_scale()` aesthetics
I acknowledge that this is user-error, but the error message could be a little bit more informative here:

``` r
library(ggplot2)

ggplot(mpg[15:16, ], aes(displ, hwy)) +
  geom_point(
    aes(colour = after_scale(c("red", "green", "blue")))
  )
#> Error in `geom_point()`:
#> Caused by error in `data.frame()`:
#> ! arguments imply differing number of rows: 3, 2
```

<sup>Created on 2024-05-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

In particular, it could inform for which aesthetics the length was incorrect, to give a hint as to what to fix.
Moreover, I think this needn't be an error, but just a warning that applying these modifications failed.

teunbrand:
This PR aims to fix #5901.

Briefly, when `after_stat()` or `after_scale()` produce aesthetics of the wrong length, a warning is thrown that the function could not be applied.

The reprex from the issue now renders with a warning:

``` r
#> ℹ Loading ggplot2

ggplot(mpg[15:16, ], aes(displ, hwy)) +
  geom_point(
    aes(colour = after_scale(c("red", "green", "blue")))
  )
#> Warning: Failed to apply `after_scale()` for the following aesthetic: colour.
```

<sup>Created on 2024-05-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I have an alternate agenda with this PR.
Moving this check into `Geom$use_defaults()` frees us from having to do so in the legend code, allowing a simplification I ultimately need for #5833.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -88,6 +88,8 @@
   to different position aesthetics of the same axis (@teunbrand, #5818).
 * In `stat_bin()`, the default `boundary` is now chosen to better adhere to 
   the `nbin` argument (@teunbrand, #5882, #5036)
+* `after_stat()` and `after_scale()` throw warnings when the computed aesthetics
+  are not of the correct length (#5901).
 
 # ggplot2 3.5.1
 
--- a/R/geom-.R
+++ b/R/geom-.R
@@ -167,6 +167,9 @@ Geom <- ggproto("Geom",
       }
 
       names(modified_aes) <- names(rename_aes(modifiers))
+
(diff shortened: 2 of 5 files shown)


teunbrand:
Thanks for the review Thomas!

PR #5903, opened 2024-05-23
Let `Layer$compute_geom_2()` handle legend defaults

teunbrand:
This PR is an amendment to #5834, as it didn't handle the non-standard defaults for `geom_sf()` legends.

Briefly, this PR let's `Layer$compute_geom_2()` instead of `Geom$use_defaults()` handle populating legend keys with defaults.

The legend code was mirroring `Layer$compute_geom_2()` in many ways anyway, so this shouldn't be much of a stretch.
This ensures that the layer could impose some of its non-aesthetic parameters on the legend keys.
That is needed to resolve the defaults for the correct legend type in `geom_sf()`, which we'd need for #5833.

In addition the following changes are made:

* `Layer$compute_geom_2()` now accepts a custom `params` object, as legends only pass static aesthetics when they are of length 1.
* The sf legend detection mechanism has moved from `LayerSf$setup_layer()` to `LayerSf$compute_geom_1()`. This allows for more readable code and abolishes the stateful `computed_legend_key_type` field in `LayerSf`.
* There is now a visual test for `geom_sf()`'s legend types, because we had detected problems in revdepchecks that weren't covered in our own tests.
* There is now a test for legend keys when a geom does not having matching aesthetics, which we didn't cover before.

### DIFF
--- a/R/guide-legend.R
+++ b/R/guide-legend.R
@@ -225,35 +225,35 @@ GuideLegend <- ggproto(
 
   get_layer_key = function(params, layers, data) {
 
+    # Return empty guides as-is
     if (nrow(params$key) < 1) {
       return(params)
     }
 
     decor <- Map(layer = layers, df = data, f = function(layer, df) {
 
+      # Subset key to the column with aesthetic matching the layer
       matched_aes <- matched_aes(layer, params)
+      key <- params$key[matched_aes]
 
+      # Filter static aesthetics to those with single values
+      single_params <- lengths(layer$aes_params) == 1L
+      single_params <- layer$aes_params[single_params]
+
+      # Use layer to populate defaults
+      key <- layer$compute_geom_2(key, single_params)
+
+      # Filter non-existing levels
       if (length(matched_aes) > 0) {
-        # Filter out aesthetics that can't be applied to the legend
-        n <- lengths(layer$aes_params, use.names = FALSE)
-        layer_params <- layer$aes_params[n == 1]
-
-        aesthetics  <- layer$computed_mapping
-        is_modified <- is_scaled_aes(aesthetics) | is_staged_aes(aesthetics)
-        modifiers   <- aesthetics[is_modified]
-
-        data <- layer$geom$use_defaults(params$key[matched_aes], layer_params, modifiers)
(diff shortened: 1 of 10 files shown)


R/guide-legend.R
```diff
@@ -225,31 +225,36 @@ GuideLegend <- ggproto(
...
 
-        data <- layer$geom$use_defaults(params$key[matched_aes], layer_params, modifiers)
-        data$.draw <- keep_key_data(params$key, df, matched_aes, layer$show.legend)
-      } else {
-        reps <- rep(1, nrow(params$key))
-        data <- layer$geom$use_defaults(NULL, layer$aes_params)[reps, ]
+      # Filter non-existing levels
+      if (length(matched_aes) > 0) {
+        key$.draw <- keep_key_data(params$key, df, matched_aes, layer$show.legend)
       }
```
teunbrand, line 250: The visual test for non-sf legends is to cover this case, which was never triggered in the test base.

teunbrand, line 250: The visual test for the non-sf legend is to cover this case, which gets solved by `Layer$compute_geom_2()`, but was never triggered in the test base.

R/guide-legend.R
```diff
@@ -225,31 +225,36 @@ GuideLegend <- ggproto(
...
-        is_modified <- is_scaled_aes(aesthetics) | is_staged_aes(aesthetics)
-        modifiers   <- aesthetics[is_modified]
+      # Use layer to populate defaults
+      key <- layer$compute_geom_2(key, single_params)
 
-        data <- layer$geom$use_defaults(params$key[matched_aes], layer_params, modifiers)
-        data$.draw <- keep_key_data(params$key, df, matched_aes, layer$show.legend)
-      } else {
-        reps <- rep(1, nrow(params$key))
-        data <- layer$geom$use_defaults(NULL, layer$aes_params)[reps, ]
```
teunbrand, line 245: The visual test for the non-sf legend is to cover this case, which gets solved by Layer$compute_geom_2(), but was never triggered in the test base.

thomasp85 (approved):
LGTM

PR #5904, opened 2024-05-23
Encapsulate `sf_grob()` in `GeomSf$draw_panel()`

teunbrand:
This PR does not really solve an open issue, but does make `GeomSf` more consistent with other geoms.

Briefly, `sf_grob()` was a wrapper for `sf::st_as_grob()` that did two things that are usually handled by Geom classes.
It removed missing data, which is usually handled by `Geom$handle_na()`, and wrangles the graphical parameters for a grob, which is usually done in `Geom$draw_panel()`.
This PR divides these two responsibilities to the appropriate places in the <GeomSf> class.


thomasp85 (approved):
LGTM

arcresu:
@teunbrand after this refactor `lty = linetype` is no longer passed to `gpar` for `GeomSf`. Was this intentional?

teunbrand:
No that is not intentional, thanks for reporting on this!

Issue #5905, opened 2024-05-23
`geom_rug()` does not print a warning when `na.rm = FALSE`


pn317:
In the help file for `geom_rug()`, it states that missing values are removed with a warning if `na.rm = FALSE`.

This does not seem to be the case. 

Example

```r
library("ggplot2")

mtcars2 <- mtcars
mtcars2$mpg <- ifelse(runif(nrow(mtcars2)) < 0.2, NA, mtcars2$mpg)

ggplot(mtcars2, aes(x = wt, y = mpg)) +
  geom_point() +
  geom_rug()
```

In the example, a warning is generated from `geom_point()` but not from `geom_rug()`.

I *think* this is because `x` and `y` are both `optional_aes` in `GeomRug`, but the default `Geom$handle_na()` only passes `required_aes` and `non_missing_aes` to `remove_missing()`.

I will submit a pull request

teunbrand:
Considered fixed by #5906.

PR #5906, opened 2024-05-23
'geom_rug()' prints a warning when 'na.rm = FALSE'

pn317:
Fixes issue #5905.
When presented with missing values, 'geom_rug()' was not printing a warning message, contrary to the documentation.
A warning message is now printed when 'na.rm = FALSE' and suppressed when 'na.rm = TRUE', as expected.

### DIFF
--- a/R/geom-rug.R
+++ b/R/geom-rug.R
@@ -88,6 +88,20 @@ geom_rug <- function(mapping = NULL, data = NULL,
 GeomRug <- ggproto("GeomRug", Geom,
   optional_aes = c("x", "y"),
 
+  setup_params = function(self, data, params) {
+    self$required_aes <- character()
+
+    if (grepl("b|t", params$sides)) {
+      self$required_aes <- c(self$required_aes, "x")
+    }
+
+    if (grepl("l|r", params$sides)) {
+      self$required_aes <- c(self$required_aes, "y")
+    }
+
+    params
+  },
+
   draw_panel = function(self, data, panel_params, coord, lineend = "butt",
                         sides = "bl", outside = FALSE, length = unit(0.03, "npc")) {
     data <- check_linewidth(data, snake_class(self))
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ggplot2 (development version)
 
+* `geom_rug()` prints a warning when `na.rm = FALSE`, as per documentation (#)
 * The `arrow.fill` parameter is now applied to more line-based functions: 
   `geom_path()`, `geom_line()`, `geom_step()` `geom_function()`, line 
    geometries in `geom_sf()` and `element_line()`.


teunbrand (changes requested):
Hi thanks for preparing a PR to fix this!

There is a subtle problem with assigning variables to `self` with regards to statefulness. While it appears to work fine, like in the plot below;

``` r
devtools::load_all("~/packages/ggplot2/") # copied from your PR
#> ℹ Loading ggplot2

GeomRug$required_aes # empty
#> character(0)

  geom_rug()
```



What it actually does is to modify the global state of `GeomRug` to now have both `x` and `y` as required aesthetics. This will crash a subsequent plot wherein only one of the two is needed.

``` r
GeomRug$required_aes # populated
#> [1] "x" "y"

ggplot(mpg, aes(displ)) +
  geom_rug()
#> Error in `geom_rug()`:
#> ! Problem while setting up geom.
#> Caused by error in `compute_geom_1()` at ggplot2/R/ggproto.R:190:17:
#> ! `geom_rug()` requires the following missing aesthetics: y.
```

<sup>Created on 2024-05-23 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I think the solution to this will be to make a new `GeomRug$handle_na()` method, that computes the correct `required_aes` locally (without assigning it to `self`) and passes that on `remove_missing()`.

pn317:
@teunbrand thanks for pointing that out. I had missed that when testing because I was just alternating between

```r
ggplot(mtcars2, aes(x = wt, y = mpg)) +
  geom_point() +
  geom_rug(sides = "b")
```
which (rightly) doesn't warn, because there are no NAs in `wt`, and

```r
ggplot(mtcars2, aes(x = wt, y = mpg)) +
  geom_point() +
  geom_rug(sides = "l")
```
which does warn (with my fix) and seems to be happy by dumb luck because I am still providing both `x` and `y` aesthetics.

I will push a new commit to adapt `handle_na()` instead, as you suggest.

teunbrand:
Yeah this looks great, thanks! Is there any way I could convince you to add a small test that simply checks that the warning is emitted when there is missing data?

pn317:
@teunbrand -- sorry for the slow reply; I have been away from work for a bit. I'll have a go at doing something useful and shout if things are getting too complicated. Thanks for the support.

R/geom-rug.R
```diff
@@ -157,5 +157,44 @@ GeomRug <- ggproto("GeomRug", Geom,
...
+            c(axis, self$required_aes, self$non_missing_aes),
+            snake_class(self)
+          )
+        }
+      )
+      data <- switch(
+        paste0(sides_aes, collapse = ""),
+        "x" = ,
+        "y" = df_list[[1]],
+        "xy" = dplyr::union(df_list[[1]], df_list[[2]])
```
teunbrand, line 188: ggplot2 doesn't use dplyr (it is in suggests because for some cross-functionality).
I think a similar thing could be achieved by using `vec_set_union()` from {vctrs}.

tests/testthat/test-geom-rug.R
```diff
@@ -40,3 +40,21 @@ test_that("Rug lengths are correct", {
...
+
+    p1 <- ggplot(df2, aes(x = x)) + geom_rug()
+    p2 <- ggplot(df2, aes(x = x)) + geom_rug(na.rm = TRUE)
+
+    expect_warning(
+      ggplotGrob(p1),
+      paste0("Removed ", n_missing, " rows containing missing values or values outside the scale range")
+    )
+
+    expect_no_warning(ggplotGrob(p2))
```
teunbrand, line 58: I think this should be accompanied by a bump of the suggested {testthat} version to 3.1.5

R/geom-rug.R
pn317: Thanks - I have updated this as suggested

tests/testthat/test-geom-rug.R
pn317: Done!

teunbrand (approved):
Great, thank you the contribution and entertaining the suggestions!
I think this should be good to go, but if you want you can add your github handle to the news bullet.

pn317:
Thanks very much. I've updated the NEWS bullet as suggested.

PR #5907, opened 2024-05-24
Fallback for `dir = 'h'/'v'`

### LINKED ISSUE #5898: Fallback for `facet_wrap(as.table)`

teunbrand:
This PR aims to fix #5898 and replaces #5900.

Briefly, it is not `as.table` that we need to preserve, but old-style `dir = 'h'/'v'` need to be translated.
This PR has a fallback for these old options and soft-deprecates their use.

Developers might see the following in their testthat runs:
``` r
#> ℹ Loading ggplot2
library(deeptime)

df <- data.frame(x = 1:10, y = 1:10, period = c("Permian", "Triassic"))
ggplot(df) +
  geom_point(aes(x, y)) +
  facet_wrap_color(vars(period), colors = periods)
#> Warning: Internal use of `dir = "h"` and `dir = "v"` in `facet_wrap()` was deprecated in
#> ggplot2 3.5.2.
#> ℹ The `dir` argument should incorporate the `as.table` argument.
#> ℹ Falling back to `dir = "lt"`.
#> generated.
```

<sup>Created on 2024-05-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,7 +44,8 @@
 * (Internal) Applying defaults in `geom_sf()` has moved from the internal 
   `sf_grob()` to `GeomSf$use_defaults()` (@teunbrand).
 * `facet_wrap()` has new options for the `dir` argument to more precisely
-  control panel directions (@teunbrand, #5212)
+  control panel directions. Internally `dir = "h"` or `dir = "v"` is deprecated 
+  (@teunbrand, #5212).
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #5908, opened 2024-05-24
Custom inside-axis position for `coord_radial()`

### LINKED ISSUE #5805: Specify position of r axis for coord_radial when r_axis_inside = T
When using coord_radial we can move the `r axis` (aka y axis) inside using `r_axis_inside = T`.  However, when  `r_axis_inside = T` the position defaults to 0 degrees when plotting a full circle and this is confusing the `r axis` extends into the first value of the theta axis.

I would like to be able to control the position/angle of the `r axis` when it is inside the plot. For example below I would like to have the `r axis` at 45 degrees so it is not to be confused with the theta axis.

``` r
library(ggplot2)

# Generate data for each combination of circle (360 deg) and radius (0 to 90)
data <- expand.grid(
  angle = seq(0, 359, by = 1),
  radius = seq(0, 90, by = 1),
  value = 0.5
)

# Create the circular plot
ggplot(data, aes(x = angle, y = radius)) +
  geom_tile(alpha = 0.2) +
  coord_radial(
    r_axis_inside = T,
    expand = F
  ) +
  scale_x_continuous(breaks = seq(0, 270, 90)) +
  labs(title = "Circular Plot",
       x = "Angle (degrees)",
       y = "Radius")
```

<sup>Created on 2024-03-25 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Here is roughly what I am after:

Screenshot from 2024-03-25 17-27-57

teunbrand:
This PR aims to fix #5805.

Briefly, you can now use numeric values for `coord_radial(r.axis.inside)` that will place the radius axis at some specified position of the theta axis.

I think it is probably best illustrated how it works.
Let's make plot with clock-positions.

``` r
#> ℹ Loading ggplot2

set.seed(42)
df <- data.frame(x = runif(100, max = 12), y = rnorm(100))

p <- ggplot(df, aes(x, y)) +
  geom_point() +
  scale_x_continuous(limits = c(0, 12), breaks = 1:12, expand = c(0, 0)) +
```

This is how one would set the position:

``` r
p + coord_radial(r.axis.inside = 4.5)
```

If for some reason you have secondary axes, you can set a vector to control primary and secondary separately.

``` r
p + coord_radial(r.axis.inside = c(4.5, 7)) +
  guides(r.sec = "axis")
```

Out of bounds axes get squished to the nearest limit.

``` r
p + coord_radial(r.axis.inside = -10)
```

<sup>Created on 2024-05-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


BenVolpe94:
This is really nice! Am I correct that the radius axis angle provided to `r_axis_inside` is not calculated relative to the start point? I created a polar plot with a 360 degree x-axis limit. I chose a different start point and wanted to have the radial axis at 45 degrees, but I had to input 135 degrees instead, which is maybe counter intuitive. Also, I need to input the start value in radials, but the r_axis_inside value in degrees. For some reason the axis labels are rotated in my plot as well. 

See code below:

```R
set.seed(42)
df <- data.frame(x = runif(100, max = 360), y = rnorm(100))

p <- ggplot(df, aes(x, y)) +
  geom_point() +
  scale_x_continuous(limits = c(0, 360), breaks = seq(0, 360, by = 90)) +

p + coord_radial(theta = "x",
                 direction = -1,
                 start = -90 * pi / 180, 
                 clip = "off",
                 r_axis_inside = 135,
                 inner.radius = 0.075,
                 expand = FALSE)
```

image

teunbrand:
Thanks for raising these points!

> Also, I need to input the start value in radials, but the r_axis_inside value in degrees

The `r_axis_inside` value is in scale-units. If your scale indicates degrees, then `r_axis_inside` indicates degrees. If your scale indicates kilograms, then `r_axis_inside` indicates kilograms.

> For some reason the axis labels are rotated in my plot as well.

Yeah that shouldn't happen, I'll try to figure out what is going wrong here.

teunbrand:
Now correctly places axis at 135 degrees:
``` r
set.seed(42)
#> ℹ Loading ggplot2

df <- data.frame(x = runif(100, max = 360), y = rnorm(100))

p <- ggplot(df, aes(x, y)) +
  geom_point() +
  scale_x_continuous(limits = c(0, 360), breaks = seq(0, 360, by = 90)) +

p + coord_radial(theta = "x",
                 direction = -1,
                 start = -90 * pi / 180, 
                 clip = "off",
                 r.axis.inside = 135,
                 inner.radius = 0.075,
                 expand = FALSE)
```



<sup>Created on 2024-06-27 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM

PR #5912, opened 2024-05-28
Violin quantiles are based on observations

### LINKED ISSUE #4120: Feature request: quantiles based on observations instead of density estimates with geom_violin
geom_violin uses density estimate to plot quantiles, which doesn't align with box plot.

This has already been asked and is document: [https://stackoverflow.com/questions/36033341/differing-quantiles-boxplot-vs-violinplot](url)
issue: #2088

I think it would be useful to have an optional parameter to choose quantiles computed based on actual observations instead of density estimates. violin plots are useful to see rough distribution of the data but quantiles of actual observations also matter. Therefore, I think it would be great to have such an option without breaking existing behaviour.

PS: reprex didn't work for me but here is a simple reproducible code.
```r
library(ggplot2)
set.seed(5)
type1 = rnorm(n = 20, mean = 5)
type2 = c(10, 9, 9, 9, 9, 7, 7, 6, 5, 4, 1)
type3 = c(rnorm(n = 10, mean = 2.5), rnorm(n = 10, mean = 7.5, sd = 0.5))

#median(example1)
#median(example2)

df = data.frame(type = c(rep("Type 1", length(type1)), rep("Type 2", length(type2)), rep("Type 3", length(type3))), 
				val = c(type1, type2, type3))

ggplot(df, aes(x = type, y = val)) + theme_classic() + 
	geom_boxplot(alpha = 0.5) + 
	geom_violin(scale = "area", alpha = 0.5, draw_quantiles = c(0.25, 0.5, 0.75)) +
	geom_dotplot(binaxis = "y", stackdir = "center", alpha = 0.3, dotsize = 0.4)
```

teunbrand:
This PR aims to fix #4120.

Briefly, instead of GeomViolin estimating quantiles based on the density, StatYdensity includes `quantile` as a computed variable. GeomViolin is then able to draw the quantiles based on that computed variable.

I consider this a breaking change as StatYdensity how 'owns' the `draw_quantiles` parameter, rather than GeomViolin. For users, this should only matter when they are breaking the link between `geom_violin()` and `stat_ydensity()`. It is also a visual change, as the quantiles are now correct.

Reprex from the issue, notice how the red quantiles of the violin now aligns with the boxplot quantiles.

``` r
#> ℹ Loading ggplot2
set.seed(5)

types <- list(
  rnorm(n = 20, mean = 5),
  c(10, 9, 9, 9, 9, 7, 7, 6, 5, 4, 1),
  c(rnorm(n = 10, mean = 2.5), rnorm(n = 10, mean = 7.5, sd = 0.5))
)

df = data.frame(
  type = rep(paste0("Type ", seq_along(types)), lengths(types)),
  val  = unlist(types)
)

ggplot(df, aes(x = type, y = val)) + theme_classic() + 
  geom_boxplot(alpha = 0.5) + 
  geom_violin(scale = "area", alpha = 0.5, draw_quantiles = c(0.25, 0.5, 0.75), 
              colour = "red") +
  geom_dotplot(binaxis = "y", stackdir = "center", alpha = 0.3, dotsize = 0.4)
#> Bin width defaults to 1/30 of the range of the data. Pick better value with
#> `binwidth`.
```

<sup>Created on 2024-05-28 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85:
I'm considering whether `draw_quantiles` should still be a parameter of the geom. It seems wrong that drawing is controlled by a stat.

Would it make sense to have the stat always compute the quantile, and the geom drawing it depending on the value of `draw_quantile`?

teunbrand:
In principle I agree with you, though there are some practicalities that may get in the way.
The issue is that `draw_quantiles` currently is not a boolean and carries the numeric values for which quantiles, which the stat should know about, but the geom doesn't.
Another issue is what the default quantiles that are computed should be, but it'd probably should be `c(0.25, 0.5, 0.75)`.

thomasp85 (commented):
I still do not like how the stat flows into the how the geom behaves. Is it not possible to gracefully deprecate the old behaviour. 

So, `geom_violin()` will check the value of `draw_quantile`. If it is not a boolean it will emit a deprecation warning and forward the value to `quantiles` in the stat (which does what `draw_quantile` does now)

teunbrand:
We should discuss if we should give this the #5423 treatment. So instead of `draw_quantile = TRUE/FALSE` we use `quantiles.linetype = 0` and add `quantiles.colour = NULL`, `quantiles.linewidth = NULL`.

thomasp85 (approved):

PR #5917, opened 2024-05-30
Refactor facet panel drawing code

### LINKED ISSUE #5482: Feature Request: Break up `FacetX$draw_panels()` into smaller functions
The `draw_panels()` methods are quite beefy with various tasks. Extending facets with custom `draw_panels()` methods is a bit of a pain due to having to repeat/copy a lot of code. As I understand the code, there are three major steps that should be mostly independent of one another:

* Initialise the panels as a gtable.
* Attach axes to panels.
* Attach strips to panels.

There are more granular steps within each task, but I think the above tasks are good chunks to isolate into separate methods to make facets more extensible. My proposal is to keep `draw_panels()` for backward compatibility, but add 3 methods for the tasks above that are called in `draw_panels()`.

teunbrand:
This PR aims to fix #5482.

Briefly, the panel drawing code is now divided over 3 Facet ggproto methods: `init_gtable()`, `attach_axes()` and `attach_strips()`.

My main gripe with the previous code was that it was hard to follow along with what it was doing. I think reasons why it was hard for me to follow are the following:
* The code was alternating between the three tasks, for example `facet_grid()` was first setting up axes, then setting up strips, then initialising the table, then adding the axes and finally adding the strips.
* Some pieces of code were 4 `if` blocks deep which is hard to keep track of.

The primary benefit is that it should now be easier to 'just' tweak the axes, or 'just' tweak the strips in extensions.
Quick benchmark showed no speed difference between current main and this PR.

### DIFF
--- a/R/facet-.R
+++ b/R/facet-.R
@@ -133,8 +133,32 @@ Facet <- ggproto("Facet", NULL,
   draw_front = function(data, layout, x_scales, y_scales, theme, params) {
     rep(list(zeroGrob()), vec_unique_count(layout$PANEL))
   },
-  draw_panels = function(panels, layout, x_scales, y_scales, ranges, coord, data, theme, params) {
-    cli::cli_abort("Not implemented.")
+  draw_panels = function(self, panels, layout, x_scales = NULL, y_scales = NULL,
+                         ranges, coord, data = NULL, theme, params) {
+
+    free  <- params$free       %||% list(x = FALSE, y = FALSE)
+    space <- params$space_free %||% list(x = FALSE, y = FALSE)
+
+    if ((free$x || free$y) && !coord$is_free()) {
+      cli::cli_abort(
+        "{.fn {snake_class(self)}} can't use free scales with \\
+        {.fn {snake_class(coord)}}."
+      )
+    }
+
+    aspect_ratio <- theme$aspect.ratio
+    if (!is.null(aspect_ratio) && (space$x || space$y)) {
+      cli::cli_abort("Free scales cannot be mixed with a fixed aspect ratio.")
+    }
+
+    table <- self$init_gtable(
+      panels, layout, theme, ranges, params,
+      aspect_ratio = aspect_ratio %||% coord$aspect(ranges[[1]]),
+      clip = coord$clip
+    )
+
+    table <- self$attach_axes(table, layout, ranges, coord, theme, params)
+    self$attach_strips(table, layout, params, theme)
   },
   draw_labels = function(panels, layout, x_scales, y_scales, ranges, coord, data, theme, labels, params) {
     panel_dim <-  find_panel(panels)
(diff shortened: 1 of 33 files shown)


teunbrand:
TODO: try unifying `draw_panels()`

teunbrand:
The `draw_panels()` method is now part of the base `Facet` class, as well as the `init_gtable()` method.
`FacetWrap` still has to `ggproto_parent()` the panel drawing code due to interactions with `CoordFlip` that need to be adressed.
I've left `FacetNull` as-is, to keep using its highly simplified method.

R/facet-.R
```diff
@@ -133,8 +133,32 @@ Facet <- ggproto("Facet", NULL,
...
+      cli::cli_abort(
+        "{.fn {snake_class(self)}} can't use free scales with \\
+        {.fn {snake_class(coord)}}."
+      )
+    }
+
+    aspect_ratio <- theme$aspect.ratio
+    if (!is.null(aspect_ratio) && (space$x || space$y)) {
+      cli::cli_abort("Free scales cannot be mixed with a fixed aspect ratio.")
+    }
```
teunbrand, line 152: This won't be triggered for FacetWrap as it has no `params$space_free`, which defaults to `list(x = FALSE, y = FALSE)`.

R/facet-wrap.R
```diff
@@ -257,227 +257,203 @@ FacetWrap <- ggproto("FacetWrap", Facet,
...
-    strip_name <- paste0("strip-", substr(params$strip.position, 1, 1))
-    strip_mat <- empty_table
-    strip_mat[panel_pos] <- unlist(unname(strips), recursive = FALSE)[[params$strip.position]]
-    if (params$strip.position %in% c("top", "bottom")) {
-      inside_x <- (theme$strip.placement.x %||% theme$strip.placement %||% "inside") == "inside"
-      if (params$strip.position == "top") {
-        placement <- if (inside_x) -1 else -2
-        strip_pad <- axis_size$top
+  draw_panels = function(self, panels, layout, x_scales, y_scales, ranges, coord, data, theme, params) {
+    if (inherits(coord, "CoordFlip")) {
```
teunbrand, line 435: If failed to find a smarter solution to working out this interaction with CoordFlip, so FacetWrap does this first, then calls the parent method.

tests/testthat/_snaps/facet-layout.md
```diff
@@ -33,7 +33,7 @@
 
 # facet_grid throws errors at bad layout specs
 
-    `coord_fixed()` doesn't support free scales.
+    `facet_grid()` can't use free scales with `coord_fixed()`.
```
teunbrand, line 36: This wording changed slightly because FacetGrid and FacetWrap share the method that emits this warning

thomasp85 (approved):
LGTM

PR #5921, opened 2024-05-31
Sort scale and viewscale limits

### LINKED ISSUE #5918: `scale_*_continuous` fails to render breaks if axis is reversed
Where a scale is reversed in `scale_*_continuous`, it fails to render any breaks. 

For example, the following *does not* work:
`ggplot() + scale_x_continuous(limits = c(10,1))`

But the following *does* work:
`ggplot() + scale_x_continuous(limits = c(1,10))`

The bug developed somewhere between version 3.4.3 and 3.5.1.

teunbrand:
This PR aims to fix #5918.

Briefly, `continuous_scale()` now sorts the limits and position guides get passed sorted ranges.

The order of the limits should now only matter when choosing e.g. `limits = c(NA, 10)` to substitute the `NA` for the data limits. Examples below are from the issue.

Regular scale, flipped limits:
``` r
#> ℹ Loading ggplot2

ggplot() + scale_x_continuous(limits = c(10, 0))
```

Reverse scale, sorted limits:
``` r
ggplot() + scale_x_reverse(limits = c(0, 10))
```

<sup>Created on 2024-05-31 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ggplot2 (development version)
 
+* `continuous_scale()` passes sorted limits to scale (@teunbrand, #5918)
 * The `arrow.fill` parameter is now applied to more line-based functions: 
   `geom_path()`, `geom_line()`, `geom_step()` `geom_function()`, line 
    geometries in `geom_sf()` and `element_line()`.
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -127,11 +127,6 @@ continuous_scale <- function(aesthetics, scale_name = deprecated(), palette, nam
     guide <- "none"
   }
 
-  transform <- as.transform(transform)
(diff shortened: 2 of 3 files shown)


teunbrand:
This was discussed a little bit in the linked issue (TL;DR: isn't properly documented in ggplot2), but yes this would definitly break such thing if it were a feature.

teunbrand:
Yeah that's fair, but if we decide that is a proper feature, then it should be documented somewhere and implemented consistently.

thomasp85:
Yeah, I agree. We can discuss at a meeting where to go with this

teunbrand:
Closing  this in favour of #6070

PR #5923, opened 2024-05-31
Reverse dodging

### LINKED ISSUE #3610: reverse dodge order works differently in position_jitterdodge() vs., for example position_dodge2()
Reversing the order of dodged elements using position_dodge2(), one passes in `reverse=TRUE`. To reverse dodged element positions in position_jitterdodge(), it is necessary to use a negative value to `dodge.width=-.75`; there is no 'reverse' argument.  This is undocumented in position_jitterdodge and inconsistent with other dodge functions.

teunbrand:
This PR aims to fix #3610.

Briefly, it adds a `reverse` argument to `position_dodge()` and `position_jitterdodge()` that reverses the groups within a position in terms of dodge order.

As a detail, for consistency with `position_dodge2()`, we actually pass `!reverse` to `collide()`.

Adapted reprex from issue:

``` r
#> ℹ Loading ggplot2

p <- ggplot(mtcars) +
  aes(
    as.factor(cyl), mpg, 
    fill = as.factor(vs), 
    group = interaction(cyl, vs)
  )

p + 
  geom_boxplot(position = position_dodge(reverse = FALSE)) +
  geom_point(position = position_jitterdodge(reverse = FALSE), shape = 21)
```

``` r

p + 
  geom_boxplot(position = position_dodge(reverse = TRUE)) +
  geom_point(position = position_jitterdodge(reverse = TRUE), shape = 21)
```

<sup>Created on 2024-05-31 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5932, opened 2024-06-05
Margins can be units

teunbrand:
This PR is an amendement to  #5622.

Briefly, in #5622, we required that certain margin elements in the theme are of the class 'margin'.
However, reverse dependency checks shows that many people use e.g. `theme(plot.margin = unit(c(4, 4, 4, 4), "mm")`.
This isn't necessarily wrong, so this PR allows margin elements to accept units too.


thomasp85 (approved):
LGTM

PR #5933, opened 2024-06-05
Guard against old discrete scales with identity palettes

teunbrand:
This PR is an amendement to #5770.

Briefly, we introduced the `palette` argument, but revdep check show that some people contstruct their discrete position scales using `discrete_scale(..., palette = identity)`. The identity palette is problematic as it doesn't return the expected values. This PR replaces identity palettes with `seq_len` to keep consistent behaviour with older scales.

### DIFF
--- a/R/scale-discrete-.R
+++ b/R/scale-discrete-.R
@@ -140,6 +140,10 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
 
   map = function(self, x, limits = self$get_limits()) {
     if (is.discrete(x)) {
+      # Guard against old use of having identity palettes
+      if (identical(.subset2(self, "palette"), identity)) {
+        self$palette <- seq_len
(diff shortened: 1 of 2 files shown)


R/scale-discrete-.R
```diff
@@ -140,6 +140,10 @@ ScaleDiscretePosition <- ggproto("ScaleDiscretePosition", ScaleDiscrete,
 
   map = function(self, x, limits = self$get_limits()) {
     if (is.discrete(x)) {
+      # Guard against old use of having identity palettes
+      if (identical(.subset2(self, "palette"), identity)) {
+        self$palette <- seq_len
+      }
```
thomasp85, line 146: Could we fix this during construction? instead of checking at every map call?

teunbrand: Good call 👍

thomasp85 (approved):
LGTM

PR #5934, opened 2024-06-05
Update facet_wrap (adjsut the strip position for each panel when strip.placement is "outside")

junjunlab:
1. Making sure the **strip.switch.pad.wrap** works when **strip.placement=outside** and **strip.position="top"**.
2. Giving same gap between the strip and panel when **strip.position** is **"left"/"right"** or when **nrow>2**.


teunbrand:
Hi, thanks for putting together a PR!
If I read the code correctly, this change applies the padding regardless of whether an axis is placed between the panel and the strip, right?
If so, that breaks with what is documented for the `strip.switch.pad.*` arguments.
Would you mind opening an issue first describing the problem you are interested in solving?
This way, we may be able to discuss and better guide a contribution.

junjunlab:
Sure, I will open an issue now. Thanks for your replying.

PR #5937, opened 2024-06-07
Add `geom_step(orientation)` argument

### LINKED ISSUE #5936: `orientation` for `geom_step()`
_(First time posting, feedback is welcome!)_
When a vertical phenomenon must be represented (ex. depth or height profile) ,  `geom_line(orientation = "y")` could be use. I don't think that there is similar way to use `geom_step()`, since `orientation` is not an argument for that function.

I would like to be able to use the `orientation ="y"` in `geom_step()`. My actual workaround is to use the superseeded `coord_flip()`.  I completely aggree with its documentation...

> This function is superseded because in many cases, coord_flip() can easily be replaced by swapping the x and y aesthetics, or optionally setting the orientation argument in geom and stat layers.

... but it does not seem to be the case for `geom_step()`

The following example:
``` r
library(tidyverse)
#> Warning: le package 'ggplot2' a été compilé avec la version R 4.3.3
```

``` r
df = data.frame(height= c(0:10), var1= c(0,1,5,2,1,5,4,9,8,7,7))

# geom_line() way to represent a soil profil
ggplot(df, aes(x=var1, y = height)) + geom_line(orientation = "y")
```

``` r

# geom_step() doesn't take `orientation`
ggplot(df, aes(x=var1, y = height))  +geom_step(orientation = "y")
#> Warning in geom_step(orientation = "y"): Ignoring unknown parameters:
#> `orientation`
```

``` r

# The proper way seems to use the superseeded `coord_flip()`
ggplot(df, aes(x=height, y = var1))  +geom_point()+
  geom_step() +
  coord_flip()
```

<sup>Created on 2024-06-06 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5936.

Briefly, it adds the `orientation` argument to `geom_step()`.
If `orientation = "y"`, stairsteps are constructed vertically.

In the example below, recall that the steps go from top to bottom, so `direction = "hv"`  first does the horizontal step, then the vertical one (and vice versa for `direciton = "vh"`).

``` r
#> ℹ Loading ggplot2

recent <- economics[economics$date > as.Date("2013-01-01"), ]
ggplot(recent, aes(date, unemploy)) +
  geom_point(shape = 21, size = 3, fill = "white") +
  geom_step(aes(colour = "hv"), orientation = "y", direction = "hv") +
  geom_step(aes(colour = "vh"), orientation = "y", direction = "vh") +
  geom_step(aes(colour = "mid"),orientation = "y", direction = "mid")
```

<sup>Created on 2024-06-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5943, opened 2024-06-13
setup sides parameter in `geom_rug()`

### LINKED ISSUE #5940: layer code with geom = "rug" no longer works as expected in Development version
The below code works in 3.5.1, but now does not in the Development version.

I know `layer()` is not the conventional way to create a rug plot - but I am using this code under the hood in my ggplot2 extension package - and now it breaks with rug.

Maybe this has something to do with the `outside = FALSE, sides = bl, length = 0.03` defaults in geom_rug??

``` r
library(tidyverse)
library(palmerpenguins)

geom_rug()
#> geom_rug: outside = FALSE, sides = bl, length = 0.03, na.rm = FALSE
#> stat_identity: na.rm = FALSE
#> position_identity
```

``` r

palmerpenguins::penguins |>
  ggplot() +
  layer(
    geom = "rug",
    stat = "identity",
    position = "identity",
    mapping = aes(x = flipper_length_mm, y = body_mass_g, col = sex),
  )
#> Error:
#> Caused by error in `if (grepl("b|t", params$sides)) ...`:
#> ! argument is of length zero
```

<sup>Created on 2024-06-08 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5940.

Briefly, it uses the `setup_params` method to fill in a potentially missing `sides` argument.

### DIFF
--- a/R/geom-rug.R
+++ b/R/geom-rug.R
@@ -159,6 +159,11 @@ GeomRug <- ggproto("GeomRug", Geom,
 
   rename_size = TRUE,
 
+  setup_params = function(data, params) {
+    params$sides <- params$sides %||% "bl"
+    params
+  },
+
   handle_na = function(self, data, params) {
     sides_aes <- character()
 


thomasp85 (approved):
LGTM after news bullet

teunbrand:
It fixes a bug introduced in the dev version. This bug hasn't existed on CRAN, so this can fly under the news radar I think.

thomasp85:
ah, fair

PR #5944, opened 2024-06-14
`facet_grid()` evaluates facets before adding margins

### LINKED ISSUE #1864: facet_grid with an expression cannot have margins
A visual of the issue,

``` R
ggplot(mtcars, aes(wt, mpg)) + geom_point() + facet_grid(vs~I(am+1), margins=T)
```

grid-expression-margin

The column (expression variable) margin is missing data!

The problem arises because [these two lines](https://github.com/hadley/ggplot2/blob/170919680836605bdf01588727a79bf5ff5ddb55/R/facet-grid-.r#L231-L233) depend on each other. The margins must be added with all the facet variables present in the data, yet the facet values must be computed with margins already computed and present in the data.

teunbrand:
This PR aims to fix #1864.

Briefly, we evaluate facet variables before adding margins and keep track of margins via an index.

To explain a little; the issue results from (1) margins requiring all faceting variables to be present and (2) faceting values require all margins to be computed. We cannot satisfy both due to their interdependence. This PR solves the issue by evaluating all faceting variables before adding margins, thus satisfying (1) but not (2). The reason for requiring (2) is just so that data is recycled correctly along with the margins. In this PR, we use an index column to keep track of how the original data should be recycled to fit the margins.

Reprex from issue:

``` r
#> ℹ Loading ggplot2

  geom_point() + 
  facet_grid(vs~I(am+1), margins=T)
```

<sup>Created on 2024-06-14 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5946, opened 2024-06-18
Scale palettes from theme

### LINKED ISSUE #4696: Option to make default colour schemes accessible?
It would be a shortcut for changing all of the default colour/fill scales.

teunbrand:
This is a proof-of-concept PR for a part of #2239 and fix #4696.

It explores setting scale palettes from the theme. When a scale has a `NULL` palette, the palette will be retrieved from the theme. In the theme, palettes can be provided for different aesthetic/discreteness combinations. Currently, only aesthetics for which this is implemented are `colour` and `fill`. 

Some demos of this mechanism:

``` r
#> ℹ Loading ggplot2

  discrete_scale("colour", palette = NULL)

# Using a palette function
p + theme(palette.colour.discrete = pal_viridis(option = "H"))
```

``` r
# Using a palette name
p + theme(palette.colour.discrete = "hue")
```

``` r
# Using a vector of colours
colours <- palette.colors(7, palette = "Okabe-Ito")
p + theme(palette.colour.discrete = colours)
```

<sup>Created on 2024-06-18 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The idea is that we'd have the default scales, e.g. `scale_colour_discrete()`, `scale_colour_continuous()` etc. have `NULL` palettes so that this mechanism kicks in. I'd still have to figure out how to do this in a backwards compatible manner though.


teunbrand:
The default colour/fill scales like `scale_colour_discrete()`, `scale_fill_continuous()` etc. now all produce a scale where `palette = NULL` and thus theme palettes are invoked. For that reason, we wouldn't need to invoke a special scale here:

``` r
#> ℹ Loading ggplot2

  theme(
    palette.colour.discrete = pal_brewer("qual")
  )

p
```



The old way of specifying default scales through `options()` is reasonably preserved, though some error messages are somewhat less specific. I think if we're satisfied with supplying palettes through the `theme()`, we can start deprecating the `options()` method.

``` r
options("ggplot2.discrete.colour" = scale_colour_viridis_d)

p
```



teunbrand:
Now supports all vanilla non-position aesthetics.
Extension aesthetics can be added by using `register_theme_elements()`, where fallback palettes can be set as part of the specifications.
``` r
register_theme_elements(
  # Will be used as fallback palette
  palette.foobar.discrete = function(n) seq(0.1, 1, length.out = n),
  element_tree = list(
    palette.foobar.discrete   = el_def(c("character", "numeric", "integer", "function"))
  )
)
```
I'll bump this from POC to WIP.

teunbrand:
Should now work with the new `scales::as_discrete_pal()` and `scales::as_continuous_pal()`.
The named version of specifying palettes doesn't currently work until https://github.com/r-lib/scales/pull/448 gets merged.
Bumping this from WIP to ready for review

R/scales-.R
```diff
@@ -168,6 +168,39 @@ ScalesList <- ggproto("ScalesList", NULL,
...
+
+  set_palettes = function(self, theme) {
+    for (scale in self$scales) {
+      if (!is.null(scale$palette)) {
+        next
+      }
+
+      # Resolve palette theme setting for this scale
+      type <- if (scale$is_discrete()) "discrete" else "continuous"
+      elem <- paste0("palette.", scale$aesthetics[1], ".", type)
```
thomasp85, line 181: In the case the scale has 2 aesthetics, say `c("colour", "fill")`, it will only look for `palette.colour.<type>`. Shouldn't it look for the first match instead?

teunbrand: Yeah that is a good idea, thanks!

thomasp85 (approved):
LGTM - don't know if we should wait with this until the edition split? I think we discussed this and found that it was ok to merge, but please correct me if I'm wrong

PR #5948, opened 2024-06-19
Recognise empty scale with `limits = <function>` as empty

### LINKED ISSUE #5945: scale_y_discrete() fails without data to plot
Hitting some instances where `geom_col() + scale_y_discrete()` errors out instead of producing an empty plot when given no data.

I've encountered two different failure modes with ggplot 3.5.1 in R 4.4. repexes below simplified from actual use.
```r
theme_set(theme_bw()) # probably doesn't matter, but just in case

ggplot() + 
  geom_col(aes(x = meanX, y = level, fill = level), tibble(x = 1, level = factor(1)) %>% filter(x < 1) %>% group_by(level, x) %>% summarize(meanX = mean(x), .groups = "drop")) +
  labs(x = NULL, y = NULL, fill = NULL, title = "title") +
  scale_y_discrete()
Error in ans[ypos] <- rep(yes, length.out = len)[ypos] : 
  replacement has length zero
In addition: Warning message:
In rep(yes, length.out = len) : 'x' is NULL so the result will be NULL
```

```r
ggplot() + 
  geom_col(aes(x = meanX, y = level, fill = level), tibble(x = 1, level = factor(1)) %>% filter(x < 1) %>% group_by(level, x) %>% summarize(meanX = mean(x), .groups = "drop")) +
  labs(x = NULL, y = NULL, fill = NULL, title = "title") +
  scale_y_discrete(labels = NULL, limits = rev, drop = FALSE)
Error in `mapped_discrete()`:
! Can't convert `x` <list> to <double>.
Warning messages:
1: In structure(in_domain, pos = match(in_domain, breaks)) :
  Calling 'structure(NULL, *)' is deprecated, as NULL cannot have attributes.
  Consider 'structure(list(), *)' instead.
2: In structure(in_domain, pos = match(in_domain, breaks)) :
  Calling 'structure(NULL, *)' is deprecated, as NULL cannot have attributes.
  Consider 'structure(list(), *)' instead.
```
Workaround's to check for empty tibbles and skip ` + scale_y_discrete()` when they occur. Easy enough but it seems atypical for ggplot that callers need to check for and handle the no data case. Also seems curious explicitly specifying a default `scale_y_discrete()` fails but leaving it implicit succeeds.

teunbrand:
This PR aims to fix #5945.

Briefly, a discrete position scale was treated as empty when range data is absent and limits are `NULL`.
This PR also treats such scale as empty when range data is absent and limits is a function.
In such cases, functions don't recieve proper 'natural data limits' because the range data is absent.
Potentially, this would cause troubles downstream, so I thought it best to treat such cases as empty scales.

Reprex from issue, note the empty plot instead of errors:

``` r
#> ℹ Loading ggplot2

df <- data.frame(
  level = factor(character(), "1"),
  x = numeric(),
  meanX = numeric()
)

ggplot(df) + 
  geom_col(aes(x = meanX, y = level, fill = level)) +
  labs(x = NULL, y = NULL, fill = NULL, title = "title") +
  scale_y_discrete(labels = NULL, limits = rev, drop = FALSE)
```


thomasp85 (approved):
LGTM

PR #5949, opened 2024-06-19
Preserve `NA`s in discrete palettes

### LINKED ISSUE #5929: ggplot2 ignores the user's input for transparency via NA's in scale_fill_manual()
I encountered a problem with scale_fill_manual where ggplot was explicitly ignoring the NA (for transparency) I was attempting to use.

The problem appears to be this commit, where the default value for NA's is set to "gray50".   Instead of trusting that the user knows what they're doing when they set NA for transparency, it overrides them behind their back without letting them know.

https://github.com/tidyverse/ggplot2/commit/c9adeeddca2ecba9162f231a5c5d326f103e691b

It was quite maddening to try to figure out what was going on, since scale_fill_manual() changes the behavior silently.

It would be useful (and merciful on anyone trying to debug similar things) to have scale_fill_manual() spit out a message on NA's similar to 

```
Setting NA's to "gray50".    Please add 'na.value = NA' if you wish transparency
```

Code to demonstrate issue:
```
library(tidyverse)
library(sf)

map_wind <-
  read_sf("https://www.spc.noaa.gov/products/outlook/archive/2024/day1otlk_20240530_0100_hail.nolyr.geojson")

ggplot() +
    theme_bw() + 
    geom_sf(data = map_wind, aes(fill = LABEL), color = NA, show.legend = TRUE) +

    scale_fill_manual(
      limits = c("0.05",     "0.15",    "0.30",    "0.45",    "0.60",   "SIGN"),
      values = c("#c5a393", "#ffeb80", "#ff8080", "#ff00ff", "#912cee", NA)
    )
```

The image on the left shows the incorrect behavior.   scale_fill_manual() is silently replacing my NA (transparency) with 'gray50'.

The image on the right shows how it *should* look.  scale_fill_manual() does what I tell it to do, and renders transparency as actual transparency.

na_value

teunbrand:
This PR aims to fix #5929.

Briefly, there are 3 situations by which a value can become `NA` in a discrete scale.
1. The data value is `NA`.
2. The data value does not match the scale limits.
3. The data value is mapped to `NA` because a palette value is `NA`.

Currently, we cannot distinguish between situation (2) and situation (3).
This PR ensures that situation (1) and (2) are laundered through the `na.translate`/`na.value` mechanism, but (3) is not because we could assume (3) is intentional by users (see issue).

As a demo, notice that 'e' gets `NA` due to (3) and is subsequently removed, while 'r' gets `na.value` due to (2).

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(displ, hwy, colour = fl)) +
  geom_point() +
  scale_colour_manual(
    values = c("tomato", "dodgerblue", NA, "limegreen"),
    limits = function(x) setdiff(x, "r"),
    na.value = "black"
  )
#> Warning: Removed 8 rows containing missing values or values outside the scale range
#> (`geom_point()`).
```


R/scale-.R
```diff
@@ -969,23 +969,23 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
-      pal_match <- unname(pal_match)
-    } else {
-      # if pal is not named, limit the values directly
-      pal_match <- pal[match(as.character(x), limits)]
+      pal[is.na(match(pal_names, limits))] <- na_value
+      pal <- unname(pal)
+      limits <- pal_names
     }
+    pal <- c(pal, na_value)
+    pal_match <- pal[match(as.character(x), limits, nomatch = length(pal))]
```
teunbrand, line 983: This part is the trick to discrimating (2) and (3). In case (2) we get `nomatch`'ed to the last palette value, which we set to the `na.value`. This leaves (3) as-is.

R/scale-.R
```diff
@@ -969,23 +969,23 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
...
 
-    if (self$na.translate) {
-      ifelse(is.na(x) | is.na(pal_match), self$na.value, pal_match)
-    } else {
-      pal_match
+    if (!is.na(na_value)) {
+      pal_match[is.na(x)] <- na_value
```
teunbrand, line 986: We no longer need to translate `is.na(pal_match)` because we do so through the (lack of) matching above

thomasp85 (approved):
LGTM

PR #5953, opened 2024-06-20
Move coord clipping responsibility from facet to coord

### LINKED ISSUE #5951: Feature request: non-rectangular panel clipping
Currently, turning `clip = "on"` has little effect for `coord_radial()` with out-of-bounds values. In the example below, I'd like it to be possible to clip points to the pacman-shape and not display any data in the upper quarter of the plot.

``` r
library(ggplot2)

  geom_point() +
  scale_x_continuous(limits = c(3, 6), oob = scales::oob_keep) +
  coord_radial(
    start = 0.25 * pi, end = 1.75 * pi, 
    clip = "on"
  )
```

<sup>Created on 2024-06-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Clipping currently is a property of the grob added to the gtable, and thus the clipping is enforced by a rectangular gtable cell. While the clipping is setup by the coord, it ultimately under controlled by the facets, e.g. here:

https://github.com/tidyverse/ggplot2/blob/ba0b18ab7bcc37e86b971e382710ce2cac6ed60e/R/facet-null.R#L66

The coord currently has no option to intervene in the clipping, which would ideally occur before panels are handed off to the `Facet$draw_panels()` method. I think the best option will be to move the responsibility of the following piece of code from the `Layout` to the `Coord`, which can then intervene as it sees fit.

https://github.com/tidyverse/ggplot2/blob/ba0b18ab7bcc37e86b971e382710ce2cac6ed60e/R/layout.R#L79-L96

### LINKED ISSUE #5952: Inconsistent clipping of strip background depending on facets
When using `coord_*(clip = "on")` facet strips are clipped along with the panel.
When using `coord_*(clip = "off")`, `facet_wrap()` does not clip the strip, but `facet_grid()` does.

``` r
library(ggplot2)

# Plot with thick strip lines
  geom_point() +
  coord_cartesian(clip = "off") +
  theme(
    strip.background = element_rect(linewidth = 5, colour = "black")
  )

# Strip is clipped
p + facet_grid(~ year)
```

``` r
# Strip is not clipped
p + facet_wrap(~ year)
```

<sup>Created on 2024-06-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I'm unsure which is the correct behaviour, but they should be consistent regardless.
Perhaps this shouldn't be under control of the Coord's clipping setting at all, but rather `theme(strip.clip)`?

teunbrand:
This PR aims to fix #5951 and fix #5952.

Briefly, the solution here is that facets no longer clip panels to their rectangular cells in the gtable, but coords set the appropriate viewport on their panels. This also allows custom behaviour for `coord_radial()` to use a clipping path instead of rectangle.

Reprex from issue; note that the points in upper quarter are appropriately clipped.

``` r
#> ℹ Loading ggplot2

  geom_point() +
  scale_x_continuous(limits = c(3, 6), oob = scales::oob_keep) +
  coord_radial(
    start = 0.25 * pi, end = 1.75 * pi, 
    clip = "on"
  )
```

<sup>Created on 2024-06-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Of note, non-rectangular clipping paths are a newer feature in R graphics and are silently ignored prior to R4.1.0.


teunbrand:
As #5952 was code-adjacent to this PR, i've included a fix for that issue as well. The fix is to set `theme(strip.clip = "on")` by default and always use that instead of coord clip settings in the strips.

BenVolpe94 (approved):
Seems to work fine for me 👍

teunbrand:
I'm grateful for the thumbs up @BenVolpe94, but I'm afraid I need the approval of a maintainer to merge a PR 😅

thomasp85 (approved):
LGTM

thomasp85:
Can you add a TODO about adding a unit test for when the time comes and vdiffr supports clipping paths

PR #5954, opened 2024-06-21
Improve `pal_qualitative()`

### LINKED ISSUE #5013: unexpected behavior when setting named vector for default colour/fill
First, thanks for adding this functionality to ggplot2 - it has been enormously useful in my day-to-day coding. Now on to the issue:

When setting a default color palette using `ggplot2.options.discrete.colour` or `ggplot2.options.discrete.fill`, the plotting function will only pull a subset of the submitted vector of length equal to the number of unique colours needed. This works fine for unnamed color vectors, however it has undesirable behavior when the relevant named elements occur outside of this subset.

``` r
library(ggplot2)
options(ggplot2.discrete.fill=c("4"="red", "8"="blue", "6"="green"))
data(mpg)

# removing 5 here for a cleaner plot; works as expected
ggplot(mpg[mpg$cyl != 5,]) +
  aes(x=hwy, fill=factor(cyl)) +
  geom_density()
```

``` r
# note that even though we have a colour named "6" in the above vector, ggplot2
# does not find it and instead uses the NA fill
ggplot(mpg[mpg$cyl != 5 & mpg$cyl != 8,]) +
  aes(x=hwy, fill=factor(cyl)) +
  geom_density()
```

<sup>Created on 2022-10-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

teunbrand:
This PR aims to fix #5013.

Briefly, it implements the fix proposed [here](https://github.com/tidyverse/ggplot2/issues/5013#issuecomment-2069306370) and polishes `pal_qualitative()` a little bit.

Tangentially related, shouldn't `pal_qualitative()` be a function in {scales}?


tests/testthat/test-scale-hue.R
```diff
@@ -1,6 +1,5 @@
 test_that("scale_hue() checks the type input", {
-  pal <- pal_qualitative(type = 1:4)
-  expect_snapshot_error(pal(4))
+  expect_snapshot_error(pal_qualitative(type = 1:4))
```
teunbrand, line 2: error is now thrown in palette factory rather than the output function.

thomasp85 (approved):
LGTM

Issue #5628, opened 2024-01-04
Should the output gtable have a predictable shape?


teunbrand:
Currently, it is hard to guess which component of a plot goes into which cell in the gtable.
Some components, like axes, titles and (since recently) legends always get associated rows/columns in the gtable.
Other components, like facet strips, only get rows/columns in the gtable when they are present.
It might make it easier to perform operations at the gtable level if the structure was predictable, which may come up in packages like {patchwork}.

PR #5958, opened 2024-06-27
Add `strip.spacing` theme element

### LINKED ISSUE #5935: strip.placement = "outside" and strip.switch.pad.wrap don't work together perfectly for facet_wrap.
**strip.switch.pad.wrap** doesn't work when **strip.placement = "outside"** and **strip.position = "bottom"**:

```
library(ggplot2)
base <- ggplot(mtcars, aes(mpg, drat)) +
  geom_point()

base + facet_wrap(~cyl) +
  theme(strip.placement = "outside",
        strip.switch.pad.wrap = unit(0.1,"npc"))
```
image

strip will only be added on one panel when **strip.position = "bottom/left"**:

```
base + facet_wrap(~cyl,strip.position = "bottom",
                  ncol = 2) +
  theme(strip.placement = "outside",
        strip.switch.pad.wrap = unit(0.1,"npc"))
```
image

```
base + facet_wrap(~cyl,strip.position = "left") +
  theme(strip.placement = "outside",
        strip.switch.pad.wrap = unit(0.1,"npc"))
```
image

It doesn't work when **strip.position = "right"**:

```
base + facet_wrap(~cyl,strip.position = "right") +
  theme(strip.placement = "outside",
        strip.switch.pad.wrap = unit(0.1,"npc"))
```
image

teunbrand:
This PR aims to fix #5935.

Briefly, a new theme element `strip.spacing` is used to set the distance from the panel to the strip unconditionally.

Previously, we could only use `strip.switch.pad.{grid/wrap}` to set this distance, but only if there was an axis in between the panel and the strip. In case the conditions for the `strip.switch.pad.{grid/wrap}` are met, then `strip.spacing` is additive to that spacing. The default spacing is `rel(0)` inheriting from the root `theme(spacing)` argument, so it doesn't add space by default.

However, it does add a gtable cell even when there is 0-spacing, which helps towards #5628 as the gtable shape no longer depends on the relative placement of axes and strips.

To give a few examples. We weren't able to do this before:

``` r
#> ℹ Loading ggplot2

  geom_point()

p + facet_wrap(year ~ drv) +
  theme(strip.spacing = unit(1, "cm"))
```

Or this:

``` r
p + facet_grid(year ~ drv, switch = "both") +
  theme(strip.spacing = unit(1, "cm"))
```

<sup>Created on 2024-06-27 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* New `strip.spacing` theme element, determining the distance between panels
+  and strips (@teunbrand, #5935)
 * (internal) rearranged the code of `Facet$draw_paensl()` method (@teunbrand).
 * `geom_rug()` prints a warning when `na.rm = FALSE`, as per documentation (@pn317, #5905)
 * `position_dodge(preserve = "single")` now handles multi-row geoms better,
--- a/R/facet-grid-.R
+++ b/R/facet-grid-.R
@@ -390,6 +390,7 @@ FacetGrid <- ggproto("FacetGrid", Facet,
     attr(row_vars, "facet") <- "grid"
 
     strips  <- render_strips(col_vars, row_vars, params$labeller, theme)
+    spacing <- convertUnit(calc_element("strip.spacing", theme), "cm")
     padding <- convertUnit(calc_element("strip.switch.pad.grid", theme), "cm")
 
     switch_x <- !is.null(params$switch) && params$switch %in% c("both", "x")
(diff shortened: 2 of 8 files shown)


teunbrand:
Hmm you're right that this doesn't work with patchwork. I had thought that patchwork was able to deal with `strip.switch.{wrap/grid}.pad` and `strip.position = "outside"` but it does not (see https://github.com/thomasp85/patchwork/issues/361). Let's press the pause buttom on this PR for now until patchwork has fielded a solution for this.

PR #5961, opened 2024-06-28
Deprecate `geom_errorbarh()`

teunbrand:
This PR aims to fix https://github.com/tidyverse/ggplot2/pull/3506#issuecomment-525733450.

Briefly, `geom_errorbarh()` is now `geom_errorbar(orientation = "y")` + deprecation warning. The `<GeomErrorbarh>` ggproto class is kept as people have extended it (see e.g. [here](https://github.com/cran/ggiraph/blob/master/R/geom_errorbarh_interactive.R#L25)), but it is just a copy of `<GeomErrorbar>`. It also suffers from #3798 in that we don't really have a deprecation process for ggproto classes in place.


R/geom-errorbar.R
```diff
@@ -74,3 +102,9 @@ GeomErrorbar <- ggproto("GeomErrorbar", Geom,
 
   rename_size = TRUE
 )
+
+#' @rdname ggplot2-ggproto
+#' @format NULL
+#' @usage NULL
+#' @export
+GeomErrorbarh <- ggproto("GeomErrorbarh", GeomErrorbar)
```
thomasp85, line 110: Couldn't we throw in a deprecation warning in `setup_param()` and see how that works?

If we give it the same ID as the one in the constructor we can ensure it only fires once for those that uses the constructor

teunbrand: Yeah works great:

``` r
#> ℹ Loading ggplot2

df <- data.frame(
  trt = factor(c(1, 1, 2, 2)),
  resp = c(1, 5, 3, 4),
  group = factor(c(1, 2, 1, 2)),
  upper = c(1.1, 5.3, 3.3, 4.2),
  lower = c(0.8, 4.6, 2.4, 3.6)
)

ggplot(df, aes(resp, trt, colour = group)) +
  layer(
    geom = "errorbarh",
    stat = "identity",
    position = "identity",
    mapping = aes(xmin = lower, xmax = upper),
    params = list(width = 0.2, na.rm = FALSE, orientation = NA)
  )
#> Warning: `geom_errobarh()` was deprecated in ggplot2 3.5.2.
#> ℹ Please use the `orientation` argument of `geom_errorbar()` instead.
#> This warning is displayed once every 8 hours.
#> generated.
```



``` r

ggplot(df, aes(resp, trt, colour = group)) +
  geom_errorbarh(aes(xmin = lower, xmax = upper))
```



<sup>Created on 2024-09-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM

PR #5967, opened 2024-07-01
Layers have names

### LINKED ISSUE #4066: Add layer names
I don't know if there is much demand for such a feature, but I often find myself wanting to modify `ggplot` objects after they are constructed (zapping a layer, for example), but in order to do so, I need to know the index of the layer to modify.

Could there be a `name` parameter for layers?

```
df <-
  tibble(x = runif(1000, 0, 1)) %>% 
  mutate(y = x + runif(1000, x, 1)) 

g <-
  df %>% 
  ggplot() +
  geom_point(aes(x = x, y = y), alpha = 0.2) +
  geom_smooth(aes(x = x, y = y))

g$layers[[1]] <- NULL  

# would rather do something like 
# ... +
# geom_point(aes(x = x, y = y), alpha = 0.2, name = "scatter") +
# ...
# 
# g$layers[["scatter"]] <- NULL
```

teunbrand:
This PR aims to fix #4066.

Briefly, it sets names for the `layers` part of a plot that should be able to uniquely identify a layer. They're specified in the `params` argument to `layer()` and default to `NULL`.

As an example of using named layers, we see here that the `name`s provided to the layer end up as the `names` for the list of layers.

``` r
#> ℹ Loading ggplot2

  geom_point(name = "foo") +
  geom_point(name = "bar")

names(p$layers)
#> [1] "foo" "bar"
```
By default, `name = NULL`, in which case a name is chosen for a layer. This defaults to the constructor function and that is not unique, it gets a suffix through `vec_as_names()`.
```
  geom_point() +
  geom_point()

names(p$layers)
#> [1] "geom_point"     "geom_point...2"
```
Currently, you cannot set duplicated names. I was unsure whether the correct coarse here is to make names unique or throw an error. I decided it should throw an error on a whim, but I'd be happy to change this.
```r
  geom_point(name = "foo") +
  geom_point(name = "foo")
#> Error in `new_layer_names()` at ggplot2/R/plot-construction.R:169:3:
#> ! Names must be unique.
#> ✖ These names are duplicated:
#>   * "foo" at locations 1 and 2.
```
A quick demo that the names can be used to 'zap' layers.
``` r
  geom_point(colour = "blue", alpha = 0.5) +
  geom_point(colour = "red",  alpha = 0.5)

p$layers[["geom_point...2"]] <- NULL
p
```

<sup>Created on 2024-07-01 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

yutannihilation:
> Currently, you cannot set duplicated names. I was unsure whether the correct coarse here is to make names unique or throw an error. I decided it should throw an error on a whim, but I'd be happy to change this.

We might see a few revdep failures on those packages that clone layers. I can fix my package easily, so this isn't a big problem to me, though. Just for sharing.

teunbrand:
Thanks for the heads up! Do you think it makes more sense to throw a warning instead of an error, and automatically make the names unique?

yutannihilation:
Thanks. I think making it an error is a good idea. Imagining I would be a user of this feature, I think I would want to assume the names are unique. Let's revisit here if we find difficulty to enforce this change to the revdep packages.

PR #5968, opened 2024-07-02
Extra documentation for `ggplot_add()`

### LINKED ISSUE #4984: Document how to use `ggplot_add()`
It seems ggplot2 has no documentation that actually explains how to use `ggplot_add()`. Maybe it should be in [the Extending ggplot2 vignette](https://ggplot2.tidyverse.org/articles/extending-ggplot2.html)?

See https://github.com/tidyverse/ggplot2/issues/4983 for an example of confusion.

teunbrand:
This PR aims to fix #4984.

Briefly, it adds some details and an example to the documentation of `ggplot_add()`.
I'm not sure the example for `element_text()` is the most appropriate one, but I'd be happy to swap out that example for a better one.

### DIFF
--- a/R/plot-construction.R
+++ b/R/plot-construction.R
@@ -84,9 +84,32 @@ add_ggplot <- function(p, object, objectname) {
 #' @param object_name The name of the object to add
 #'
 #' @return A modified ggplot object
+#' @details
+#' Custom methods for `ggplot_add()` are intended to update the `plot` variable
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM - only fear is that someone would want to have this behaviour for `element_text()`🤣

teunbrand:
Well, now they know how to achieve it 😁

PR #5969, opened 2024-07-02
Don't require dplyr for `fortify.tbl()`

### LINKED ISSUE #4786: Requirement of dplyr for non-tibble tbls
Due to the fortify method for \<tbl> objects, {dplyr} is require for plotting with non-tibble \<tbl> classes: 
https://github.com/tidyverse/ggplot2/blob/0e64d9c56ccc8db31971723810c3c10f0a67d9e4/R/fortify.r#L19-L22 

I view {pillar} and the \<tbl> subclass as mainly for formatting so wondered if it would be possible to simply dispatch to the next data.frame method and have explicit methods for other \<tbl> sub-classes that need supporting separately.

I've also raised this in the [pillar repo](https://github.com/r-lib/pillar/issues/519) to get their opinion of how best to view the <tbl> class.

Hope this all makes sense. Example below (on install where dplyr is removed)

``` r
library(tibble)
library(ggplot2)

x <- y <- 1:2
tbl <- dat <- data.frame(x, y)
class(tbl) <- c("tbl", "data.frame")
tbl
#> # A data frame: 2 × 2
#>       x     y
#>   <int> <int>
#> 1     1     1
#> 2     2     2

# this errors
ggplot(tbl, aes(x,y)) + geom_col()
#> Error: dplyr must be installed to work with tbl objects

# this will print ok
ggplot(as_tibble(dat), aes(x,y)) + geom_col()
```

<sup>Created on 2022-03-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

teunbrand:
This PR aims to fix #4786.

Briefly, it attempts to exploit that some `as.data.frame()` methods will implicitly use `collect()` to realise a local data.frame.
For example:
``` r
dbplyr:::as.data.frame.tbl_sql
#> function (x, row.names = NULL, optional = NULL, ..., n = Inf) 
#> {
#>     as.data.frame(collect(x, n = n))
#> }
#> <bytecode: 0x000001b7f118b958>
#> <environment: namespace:dbplyr>
```
This should ensure that you can still work with e.g. `<tbl_sql>` classes that will be `collect()`'ed while not requiring {dplyr} for the `fortify.tbl()` method. Example demonstrating we can still use lazy data.frames:
``` r
#> ℹ Loading ggplot2
library(dbplyr)
con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
dplyr::copy_to(con, mtcars)
df <- dplyr::tbl(con, "mtcars")

ggplot(df, aes(mpg, disp)) +
  geom_point()
```

<sup>Created on 2024-07-02 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/fortify.R
+++ b/R/fortify.R
@@ -17,10 +17,7 @@ fortify.data.frame <- function(model, data, ...) model
 #' @export
 fortify.tbl_df <- function(model, data, ...) model
 #' @export
-fortify.tbl <- function(model, data, ...) {
-  check_installed("dplyr", reason = "to work with `tbl` objects.")
-  dplyr::collect(model)
-}
+fortify.tbl <- function(model, data, ...) as.data.frame(model)
 #' @export
 fortify.NULL <- function(model, data, ...) waiver()
 #' @export


yutannihilation (approved):
Looks good!

teunbrand:
Thanks for the review!

PR #5973, opened 2024-07-03
`n.breaks` propagate to `sec.axis`

### LINKED ISSUE #4483: n.breaks for sec_axis
Suggestion for improvement:

When using a continuous scale I find the `n.breaks` argument very useful and intuitive to use. However, when using the `sec.axis` argument to simply duplicate the existing axis I run into the issue that:

1. The labels/ticks of the original axis are not actually duplicated, rather the labels prior to any manipulations are duplicated.
2. The `sec_axis` function does not have a similar `n.breaks` argument.

I believe it would be very useful if the secondary axis could have the `n.breaks` argument and/or duplicate the primary axis that utilised this argument.

teunbrand:
This PR aims to fix #4483.

Briefly, it captures `n.breaks` in a closure when copying the default breaks from a primary scale during secondary scale initialisation.

Reprex from issue:

``` r
#> ℹ Loading ggplot2

dat <- data.frame(x = 1:10, y = sample(x = 1:1000, size = 50, replace = TRUE)) 

ggplot(data = dat, aes(x = x, y = y)) +
  geom_point() +
  scale_y_continuous(
    n.breaks = 20,
    sec.axis = sec_axis(~.)
  )
```

<sup>Created on 2024-07-03 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #5974, opened 2024-07-03
`n.breaks` are supplied to function-breaks.

### LINKED ISSUE #5972: `n.breaks` only supported when `breaks = waiver()`
This is a minor gripe, but in `scale_*_continuous(n.breaks = 20)`, the `n.breaks` only kicks in when `breaks = waiver()` (the default).
The ask here is to also have `n.breaks` apply when giving a custom `function(x, n) {...}` so that one can flexibly combine breaks functions with the desired number of breaks.

Many `scales::breaks_*()` function factories have `n` as an argument in the constructor, so it isn't a direct problem for this function family.
However, it'd be swell if this would work for user-defined functions too.

As an example, I'd like the following two plots to be identical:

``` r
library(ggplot2)

  geom_point()

p + scale_y_continuous(
    breaks = scales::breaks_extended(), # resulting function has `n` as argument
    n.breaks = 20
  )
```

``` r
p + scale_y_continuous(
    breaks = scales::breaks_extended(20),
  )
```

<sup>Created on 2024-07-03 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5972.

Briefly, it ensures that function-breaks are called with the `n = n.breaks` argument when supported, not just breaks derived from the transformation.

Example:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  scale_y_continuous(
    breaks = scales::breaks_extended(),
    n.breaks = 20
  )
```

<sup>Created on 2024-07-03 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I consider this PR a WIP because I'd also like to treat binned scales, e.g. https://github.com/tidyverse/ggplot2/issues/5972#issuecomment-2206917187. There are a few things that complicate this.

First, this has sort-of already been done in #3575, except instead of `n`, the `n.breaks` argument is used.
https://github.com/tidyverse/ggplot2/blob/2610840f7478af027294db0793df839199b8cb6b/R/scale-.R#L1287-L1291
I'm not sure how widespread an `n.breaks` argument is, but ideally we should only support `n` or `n.breaks` and not both. However, for backwards compatibility reasons, we might have to support both.

Secondly, `scale_x/y_binned()` already have a default `n.breaks` argument:

https://github.com/tidyverse/ggplot2/blob/2610840f7478af027294db0793df839199b8cb6b/R/scale-binned.R#L28

This would interfere with the strategy, because it would override `n` when providing `breaks = scales::extended_breaks(n = 20)` so that the internal function is called with `n = 10`. We could amend this by having `n.breaks = 10` be an internal default for `transformation$breaks()` only.

In any case, I'd be happy to be advised on the case of binned scales.


R/scale-.R
```diff
@@ -607,6 +607,12 @@ check_breaks_labels <- function(breaks, labels, call = NULL) {
   if (is.null(breaks)) {
     return(TRUE)
   }
+  if (identical(breaks, NA)) {
+    cli::cli_abort(
+      "Invalid {.arg breaks} specification. Use {.code NULL}, not {.code NA}.",
+      call = call
+    )
+  }
```
teunbrand, line 615: This is an adjacent change that seemed like a good idea because we can now omit this check in the `get_breaks()` calculation. It is the reason why the unit tests had to be adapted.

R/scale-.R
```diff
@@ -725,40 +737,27 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
...
-    } else if (is.waive(self$breaks)) {
-      if (!is.null(self$n.breaks) && trans_support_nbreaks(transformation)) {
-        breaks <- transformation$breaks(limits, self$n.breaks)
+    if (zero_range(as.numeric(limits))) {
+      return(limits[1])
+    }
+
+    if (is.function(breaks)) {
+      # Limits in transformed space need to be converted back to data space
+      limits <- transformation$inverse(limits)
```
teunbrand, line 748: realised we only needed the original limits for function-breaks, so zero-range can early-exit.

thomasp85 (commented):
You are deleting an awful lot of logic without adding in an equivalent amount... can you walk me through it?

R/scale-.R
```diff
@@ -725,40 +737,27 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
...
     # Compute `zero_range()` in transformed space in case `limits` in data space
     # don't support conversion to numeric (#5304)
-    if (zero_range(as.numeric(transformation$transform(limits)))) {
-      breaks <- limits[1]
-    } else if (is.waive(self$breaks)) {
-      if (!is.null(self$n.breaks) && trans_support_nbreaks(transformation)) {
-        breaks <- transformation$breaks(limits, self$n.breaks)
+    if (zero_range(as.numeric(limits))) {
+      return(limits[1])
+    }
```
teunbrand, line 744: We can early exit here. The transformation dance is not needed.

R/scale-.R
```diff
@@ -725,40 +737,27 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
...
-    if (is.null(self$breaks)) {
-      return(NULL)
-    }
-
-    if (identical(self$breaks, NA)) {
-      cli::cli_abort(
-        "Invalid {.arg breaks} specification. Use {.code NULL}, not {.code NA}.",
-        call = self$call
-      )
-    }
```
teunbrand, line 740: We catch this case early in the scale constructor

R/scale-.R
```diff
@@ -725,40 +737,27 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
...
+            "Ignoring {.arg n.breaks}. Use a {.cls transform} object or \\
+            {.arg breaks} function that supports setting number of breaks",
             call = self$call
           )
         }
-        breaks <- transformation$breaks(limits)
       }
-    } else if (is.function(self$breaks)) {
-      breaks <- self$breaks(limits)
-    } else {
```
teunbrand, line 760: We no longer need separate logic for the default breaks (breaks function(!) from transformation) and function-breaks, as both now access the `n.breaks` logic.

R/scale-.R
```diff
@@ -1005,13 +1004,6 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
       return(NULL)
     }
 
-    if (identical(self$breaks, NA)) {
-      cli::cli_abort(
-        "Invalid {.arg breaks} specification. Use {.code NULL}, not {.code NA}.",
-        call = self$call
-      )
-    }
-
```
teunbrand, line 1014: Also caught in constructor

R/scale-.R
```diff
@@ -1229,14 +1221,9 @@ ScaleBinned <- ggproto("ScaleBinned", Scale,
 
     if (is.null(self$breaks)) {
       return(NULL)
-    } else if (identical(self$breaks, NA)) {
-      cli::cli_abort(
-        "Invalid {.arg breaks} specification. Use {.code NULL}, not {.code NA}.",
-        call = self$call
-      )
```
teunbrand, line 1236: Also caught in constructor

R/scale-.R
```diff
@@ -1398,8 +1385,11 @@ check_transformation <- function(x, transformed, name, arg = NULL, call = NULL)
   cli::cli_warn(msg, call = call)
 }
 
-trans_support_nbreaks <- function(trans) {
-  "n" %in% names(formals(trans$breaks))
+support_nbreaks <- function(fun) {
+  if (inherits(fun, "ggproto_method")) {
+    fun <- environment(fun)$f
+  }
```
teunbrand, line 1391: Might need to deal with ggproto methods as `self$breaks` can be a function that gets wrapped by `ggproto()`

tests/testthat/test-scales-breaks-labels.R
```diff
@@ -170,38 +170,12 @@ test_that("suppressing breaks, minor_breask, and labels works", {
...
-  # Colour
-  scc <- scale_colour_continuous(breaks = NA)
-  scc$train(1:3)
-  expect_error(scc$get_breaks())
+  expect_error(scale_x_continuous(breaks = NA))
+  expect_error(scale_y_continuous(breaks = NA))
+  expect_error(scale_alpha_continuous(breaks = NA))
+  expect_error(scale_size_continuous(breaks = NA))
+  expect_error(scale_fill_continuous(breaks = NA))
+  expect_error(scale_colour_continuous(breaks = NA))
```
teunbrand, line 178: These tests can all be simplified as the constructor now calls out NA-breaks.

tests/testthat/test-scales.R
```diff
@@ -432,21 +432,18 @@ test_that("scales accept lambda notation for function input", {
...
   expect_snapshot_error(ggplotGrob(p))
-  p <- ggplot(mtcars) + geom_bar(aes(factor(gear))) + scale_x_discrete(breaks = NA)
-  expect_snapshot_error(ggplot_build(p))
+  expect_snapshot_error(scale_x_discrete(breaks = NA))
   p <- ggplot(mtcars) + geom_bar(aes(factor(gear))) + scale_x_discrete(labels = NA)
   expect_snapshot_error(ggplotGrob(p))
 
-  p <- ggplot(mtcars) + geom_bar(aes(mpg)) + scale_x_binned(breaks = NA)
-  expect_snapshot_error(ggplot_build(p))
+  expect_snapshot_error(scale_x_binned(breaks = NA))
```
teunbrand, line 446: Again simplification due to `breaks = NA` being caught in the constructor

R/scale-.R
```diff
@@ -717,6 +723,12 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
       return(numeric())
     }
     transformation <- self$get_transformation()
+    breaks <- self$breaks %|W|% transformation$breaks
```
teunbrand, line 726: I like to think of this line as the 'spirit of this PR'. We no longer need to treat function-breaks and default breaks any differently.

teunbrand:
I tried to elaborate somewhat more in the comments above. The TL;DR: we merge in the '`breaks` are a function'-branch with '`breaks` come from the transformation breaks function'-branch, simplifying the logic as we only need to deal with function-breaks once.

R/scale-.R
thomasp85: either the comment above needs updating or you are glossing over an issue :-)

teunbrand: Good catch thanks!

teunbrand:
> First, this has sort-of already been done in https://github.com/tidyverse/ggplot2/pull/3575, except instead of n, the n.breaks argument is used.

This is now solved by allowing both forms, which makes this PR ready for review.

thomasp85 (approved):
LGTM

PR #5976, opened 2024-07-05
Function to reset all aesthetics defaults

### LINKED ISSUE #5975: Could a function to reset all geom defaults back to the initial defaults be provided?
Thanks for the continual ggplot2 updates! 

I was wondering if there could be a function to reset all geom defaults back to the initial defaults? Otherwise, you have to know what the defaults are, or restart once you update something.

Something simple like below would be super helpful. 

```
reset_geom_defaults <- function() {
  ggplot2::update_geom_defaults("area", NULL)
  ggplot2::update_geom_defaults("bar", NULL)
  ggplot2::update_geom_defaults("boxplot", NULL)
  ggplot2::update_geom_defaults("col", NULL)
  ... # etc with all geom's there
}
```
Maybe this could be something for Tidy Dev Day, as hopefully pretty simple??
 
An idea (maybe for the future) is that it could also have an argument that would allow users to specify only certain geoms to reset

teunbrand:
This PR aims to fix #5975.

Briefly, it introduces `reset_geom_defaults()` and `reset_stat_defaults()` to undo changes introduced with `update_geom_defaults()` or `update_stat_defaults()` respectively.
Because the first time the `update_*_defaults()` functions are run, the then-current defaults are cached before being updated. This cache is used by the `reset_*_defaults()` functions as the source of original settings.


thomasp85 (approved):
LGTM

PR #4432, opened 2021-04-19
Throw error with illigal combination of space in facet_grid and aspect.ratio in theme

thomasp85:
Fix #3834 

This PR adds a check for when aspect ratio has been set by the theme and free space is requested in facet_grid and throws an error

PR #5977, opened 2024-07-05
`facet_grid(space = "free")` can work with `coord_fixed()`

### LINKED ISSUE #4584: allow `coord_fixed()` with `facet_grid(..., scales = "free", space = "free")` when x and y scales are discrete
It makes sense that it’s not possible to mix `facet_grid(..., scales = "free", space = "free")` with `coord_fixed()` in most situations. However, I think that it should be possible and makes sense to allow this when both, x and y, scales are discrete.

For example, a use case in this reprex with `ggplot2` version 3.3.4:

``` r
library(tidyverse) # with ggplot2 v3.3.4
set.seed(1)
df <- expand_grid(x = letters[1:6], y = LETTERS[1:2]) %>% 
  mutate(w = case_when(x %in% letters[1:2] ~ "1",
                       x %in% letters[3:6] ~ "2")) %>% 
  mutate(value = rnorm(6*2))

gg <- ggplot(df) +
    geom_point(aes(x, y, color = value)) +
    facet_grid(cols = vars(w), scales = "free", space = "free")

gg
```

Here, we have different numbers and levels of factors in x scale, so it’s necessary to use `scales = "free"` and `space = "free"` for efficient use of space.

When I make plots like this one, I like to make sure that the breaks on x and y axes are placed equidistantly (so the background grid is made of squares), which would generally be accomplished with `coord_fixed()` but not possible in this case since we are using free scales and space.

``` r
gg + coord_fixed()
#> Error: coord_fixed doesn't support free scales
```

As of version 3.3.3 of `ggplot2` it was possible to still accomplish this without `coord_fixed()` by the (perhaps hacky and off-label) use of `theme(aspect.ratio = 1)`, also pointed out in this [SO post](https://stackoverflow.com/questions/16258095/facet-with-free-scales-but-keep-aspect-ratio-fixed/44271207#44271207). See reprex below with version 3.3.3. This is however no longer possible since this bug was fixed (\#4432).

``` r
library(tidyverse) # with ggplot2 v3.3.3
set.seed(1)
df <- expand_grid(x = letters[1:6], y = LETTERS[1:2]) %>% 
  mutate(w = case_when(x %in% letters[1:2] ~ "1",
                       x %in% letters[3:6] ~ "2")) %>% 
  mutate(value = rnorm(6*2))

gg <- ggplot(df) +
  geom_point(aes(x, y, color = value)) +
  facet_grid(cols = vars(w), scales = "free", space = "free")

gg + theme(aspect.ratio = 1)
```

It would be great if this can be accomplished legitimately with `coord_fixed()` rather than by using hacks. Based on my admittedly superficial understanding, I think that the use of `coord_fixed()` when faceting with free scales and space should be legitimate as long as the x and y scales are discrete.

<sup>Created on 2021-08-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

teunbrand:
This PR aims to fix #4584.

Briefly, we set the ratio of a panel when `space = "free"` and using `coord_fixed()`.

Previously, in #4432 it was prohibited to set an aspect ratio in the theme and using free space.
However, there is a subtlety here. The *theme* aspect ratio controls the width/height of panels directly and you cannot reasonably set this ratio for multiple panels of disparate sizes. So theme aspect ratios still aren't allowed. The *coord* aspect ratio controls the <ins>data</ins> proportion that is consistent with free facet space, so this should be allowed.

``` r
#> ℹ Loading ggplot2

df <- data.frame(
  x = c("length a", "length b", "area", "mass", "density", "height"),
  y = c("a", "b", "c", "c", "b", "a"),
  f = c("A", LETTERS[1:5])
)

ggplot(df, aes(x, y)) +
  geom_tile() +
  facet_grid(~f, scales = "free", space = "free") +
  coord_fixed(1)
```

<sup>Created on 2024-07-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

smouksassi:
possible regression ? (CRAN version) 

``` r
library(ggplot2)
df <- data.frame(
  x = c("length a", "length b", "area", "mass", "density", "height"),
  y = c("a", "b", "c", "c", "b", "a"),
  f = c("A", LETTERS[1:5])
)

ggplot(df, aes(x, y)) +
  geom_tile() +
  facet_grid(~f, scales = "free", space = "free") +
  coord_fixed(1)
#> Error in `draw_panels()`:
#> ! `coord_fixed()` doesn't support free scales.
```

<sup>Created on 2025-05-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
The fix in this PR isn't on CRAN yet, but thanks for keeping an eye out!

PR #5983, opened 2024-07-08
Square line end for axis line children

### LINKED ISSUE #5978: Suggestion to change the default line end of axis lines in theme_classic
Inspired by the newly released option to cap the axis lines (a great addition btw), I would like to suggest a cosmetic improvement to theme_classic as I believe the current default of the line end of the axis lines is not the best fit.

Theme_classic is the only original complete theme that uses axis lines. The line end of the axis lines in this theme is actually set to NULL but practically it works like the "butt" line end. The "butt" line end cuts off the very end of the lines. This is an appropriate setting for things like axis ticks etc in all other themes (as axis ticks with the "butt" line-end will align perfectly with either the panel border or panel background of all the other themes), but I don't think it is the best choice for theme_classic. In theme_classic, when the axis lines are not capped, they join together at the origin with a very ugly notch in the left lower corner as the lines are simply too short to come together correctly (unlike for instance the lines of the panel border in the themes that use that). The notch is pretty noticeable and becomes very apparent when zooming in. It gets even worse when the axis lines are capped as the notch will then be present at the upper-outer corner of both the first and last breaks of both axes.

A solution would be to change the default of the line end for the axis lines to the "square" or "round" lineend. Hereby, the two axes would join nicely at the origin when not capped and when they are capped, they would line up perfectly with the axis ticks of the first and last breaks eliminating the ugly notches. It would probably be most appropriate to change the default line end to the "square" line end as this would keep the straight lines, thereby being consistent with the form of the axis ticks and the panel border of some of the other themes etc. If there is any way to make it a default of all axis lines in ggplot2, it would probably be beneficiary to do that, so customized themes built on, say, theme_gray with added axis lines will also use the "square" line end. I can't see any downsides to make this change, but I am not an expert in either R or ggplot2, so I don't know if there is a good explanation to the current settings. I just think it would improve the aesthetics of plots made with theme_classic.

teunbrand:
This PR also aims to fix #5978 alongside #5981 and is mutually exclusive with #5982.

Briefly, it edits the default theme to have `lineend = "square"` for the *children* of `axis.line`.
The good thing is that it doesn't require any 'hacks' to get it to work:

``` r
library(ggplot2)

  geom_point() +
  theme_classic(base_line_size = 5) +
  theme(axis.ticks.length = unit(5, "mm"))
p
```

The bad thing is you can't easily swap the lineend, i.e. the following does not work:

``` r
p + theme(axis.line = element_line(lineend = "butt"))
```

Because you'd have to set the lineend in the child elements:

``` r
p + theme(
  axis.line.x = element_line(lineend = "butt"),
  axis.line.y = element_line(lineend = "butt")
)
```

<sup>Created on 2024-07-08 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


teunbrand:
Closing this pending further discussion

PR #5985, opened 2024-07-09
Update minimum R version

rogerssam:
`ggplot2` will no longer install on R versions earlier than 4.0.0. This is due to a recursive dependency on lattice (via [mgcv](https://cran.r-project.org/web/packages/mgcv/index.html) -> [nlme](https://cran.r-project.org/web/packages/nlme/index.html)).
Lattice updated to require R >= 4.0.0 [last year](https://cran.r-project.org/web/packages/lattice/index.html).

PR #5987, opened 2024-07-09
Move {mgcv} to suggests

### LINKED ISSUE #5986: Review dependencies
This issue is prompted by #5985.

Briefly, {mgcv} depends on {nlme} which requires R >= 4.0.0.
We only use `mgcv::gam()` in `geom_smooth()` when there are >1000 points to fit, so it is a dependency with a very narrow application.
We should probably find a way to make this a suggested package rather than an imported package.

teunbrand:
This PR fixes #5986 and closes #5985.

Briefly, when `method = "gam"` and {mgcv} is not installed, we fallback to `method = "lm"` with a warning.
Needed a mock test to verify correct message is thrown and add one skip step.

Please note that downstream dependencies that use `geom_smooth(method = "gam")` will need to add {mgcv} to their dependencies.


clauswilke:
@teunbrand I believe this is a reasonable solution but just wanted to raise one potential issue: I believe once mgcv is in suggests it no longer gets installed automatically when people install ggplot2. This means most people will likely not have it installed going forward. I don't know what to do about this, just one of the annoyances of the R ecosystem that I've run into a couple of times. (I tell my students to install packages X, Y, and Z, and then key functionality is missing because some dependencies have been moved into suggests and no longer are installed automatically.)

yutannihilation:
> This means most people will likely not have it installed going forward.

I think it's a good point. I believe `rlang::check_installed()`, which interactively asks the user to install the package, should solve this as we do in several places.

https://github.com/search?q=repo%3Atidyverse%2Fggplot2%20check_installed&type=code

clauswilke:
@yutannihilation Ah, good point. Yes, this seems like the way to go.

teunbrand:
Yeah the lack of fallback is why I didn't use that initially, but giving it a 2nd thought we could probably wrap it in a `try_fetch()` block to set a fallback.

clauswilke:
Thinking some more about it, this can get rather cumbersome quickly if somebody can't install mgcv for some reason and gets prompted to do it every time they try to make a plot.

Here is an alternative to consider: ggplot could check on startup (library load) what packages are installed/missing and on occasion (once per session, once per week, whatever) print a message that lists suggested but not installed packages that would enable additional functionality.

Such a function could also not be run automatically but be available for manual checking.

teunbrand:
> this can get rather cumbersome quickly if somebody can't install mgcv for some reason and gets prompted to do it every time they try to make a plot.

This would only trigger when `method = "gam"`, either explicitly or by default when `nrow(data) > 1000`. If they don't want to be prompted about this, they could set the `method` argument to something else in order to opt-out these prompts.

clauswilke:
I see. Maybe mention this in the message that is generated?

teunbrand:
This is what happens now if {mgcv} is not installed in non-interactive versions.

``` r
#> ℹ Loading ggplot2

rlang::is_installed("mgcv")
#> [1] FALSE

  geom_smooth(method = "gam")
#> The `method` was set to "gam", but mgcv is not installed.
#> ! Falling back to `method = "lm"`.
#> ℹ Change the `method` argument to silence this message.
#> `geom_smooth()` using formula = 'y ~ x'
```



<sup>Created on 2024-07-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

In interactive versions, you'll be prompted  to install.
If you refuse, you get the same as the example above.
If you install {mgcv}, you continue with `method = "gam"` as per usual.

teunbrand:
Thanks for the suggestion Claus, it now reads:

> The `method` was set to "gam", but mgcv is not installed.
! Falling back to `method = "lm"`.
ℹ Install mgcv or change the `method` argument to resolve this issue.

thomasp85 (approved):
LGTM

PR #5988, opened 2024-07-09
Remove {glue} dependency

teunbrand:
This PR aims to fix part of https://github.com/tidyverse/ggplot2/issues/5986#issuecomment-2217476203.

It removes the {glue} dependency, which isn't technically required anymore since we do most text interpolation for messages now through the {cli} package. This PR mostly replaces residual {glue} functions with `paste0()`.
Tested with {glue} uninstalled.


yutannihilation (approved):
Looks good to me.

teunbrand:
Thanks for the review!

PR #5993, opened 2024-07-10
Move {MASS} to Suggests

teunbrand:
This PR aims to fix part of https://github.com/tidyverse/ggplot2/issues/5986#issuecomment-2217476203.

Briefly, MASS is required for 2D KDE and a particular ellipse type and not broadly all over ggplot2.
This PR moves it to Suggests, adds checks whether it is installed in the relevant places and skips tests when it is not installed.


R/stat-qq.R
```diff
@@ -32,8 +32,10 @@
 #' p <- ggplot(df, aes(sample = y))
 #' p + stat_qq() + stat_qq_line()
 #'
-#' # Use fitdistr from MASS to estimate distribution params
-#' params <- as.list(MASS::fitdistr(df$y, "t")$estimate)
+#' # Use fitdistr from MASS to estimate distribution params:
+#' # params <- as.list(MASS::fitdistr(df$y, "t")$estimate)
```
yutannihilation, line 36: Should we put this line behind `if (requireNamespace("MASS", quietly = TRUE)) { ... }`?

PietrH:
MASS is now dependent on R4.4, stopping me from installing ggplot2 (as a dep) on my friendly local cluster which tops out at R4.3. 

So, what's the status of this PR? :angel:

yutannihilation:
ggplot2's requirement is not that strict on the version of MASS, so it's not really a big deal in theory, while it feels tricky. You should be able to install some older version of MASS like this.

```r
remotes::install_version("MASS", "7.3-60.0.1")
```

R/stat-qq.R
thomasp85: yeah, I agree

thomasp85 (approved):
LGTM

Issue #2712, opened 2018-06-24
Aligning geoms with preserve = "single" appears impossible


Ax3man:
I want to plot a geom without with together with a geom that does have width, e.g. points and boxplots. 

Not all factor combinations are present.

The default of adjusting the width (`preserve = 'total'`) works fine. But when maintaining the box widths with `preserve = 'single'` I can no longer do the same for the points.

``` r
library(ggplot2)

ggplot(mtcars, aes(factor(gear), mpg, fill = factor(am))) + 
  geom_boxplot(position = position_dodge(preserve = 'total')) +
  geom_point(position = position_dodge(0.75, preserve = 'total'))
```

![](https://i.imgur.com/TJWY0Kl.png)

``` r

ggplot(mtcars, aes(factor(gear), mpg, fill = factor(am))) + 
  geom_boxplot(position = position_dodge(preserve = 'single')) +
  geom_point(position = position_dodge(0.75, preserve = 'total'))
```

![](https://i.imgur.com/WB3HoXJ.png)

``` r

ggplot(mtcars, aes(factor(gear), mpg, fill = factor(am))) + 
  geom_boxplot(position = position_dodge(preserve = 'single')) +
  geom_point(position = position_dodge(0.75, preserve = 'single'))
#> Warning in max(table(panel$xmin)): no non-missing arguments to max;
#> returning -Inf
```

![](https://i.imgur.com/ti0hIG5.png)

Created on 2018-06-24 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

batpigandme:
I don't think this (below) is what you're going for, but you might want to try playing around with [`position_dodge2()`](https://ggplot2.tidyverse.org/reference/position_dodge.html).
``` r
library(ggplot2)

ggplot(mtcars, aes(factor(gear), mpg, fill = factor(am))) + 
  geom_boxplot(position = position_dodge2(0.75, preserve = 'single')) +
  geom_point(position = position_dodge2(0.75, preserve = 'single'))
```

![](https://i.imgur.com/qLod72Y.png)

Created on 2018-06-24 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

Ax3man:
Yes ok, this solves it:
```
ggplot(mtcars, aes(factor(gear), mpg, fill = factor(am))) + 
  geom_boxplot(position = position_dodge2(0.75, preserve = 'single')) +
[...]

PR #6000, opened 2024-07-15
Fix for `position_dodge(preserve = "single")` with location based data

### LINKED ISSUE #3647: Aligning geoms with preserve = "single": appears not to work with several factors
Hi,

I discovered issue #2712 when looking for a solution to the misalignment of boxplots with overlaid points. While the solution shown works in the example, it does not seem to generalise to cases with a greater number of factors.

Ideally it would be possible to achieve the same thing with `position_jitterdodge()`, but I guess that is beyond the scope of this issue.

```r
library(ggplot2)

ggplot(mtcars, aes(factor(am), mpg, fill = factor(carb))) + 
  geom_boxplot(position = position_dodge2(0.75, preserve = 'single')) +
  geom_point(position = position_dodge(0.75, preserve = 'total'))
```

image

### LINKED ISSUE #5995: `position_dodge` not working well with `preserve = "single"` and `geom_text` (or `geom_point`)
In CRAN version, it doesn't manage to dodge the text correctly...

``` r
library(tidyverse, quietly = TRUE)
#> Warning: package 'ggplot2' was built under R version 4.4.1
```

``` r
library(palmerpenguins, quietly = TRUE)

penguins |>
  count(species, sex) |>
  ggplot(
    aes(x = sex,
        y = n,
        fill = species,
        label = n),
  ) +
  geom_col(
    position = position_dodge(preserve = "single"),
    width = 0.75,
    alpha = 0.9,
  ) +
  geom_text(
    colour = "black",
    position = position_dodge(width = 0.75, preserve = "single"),
    vjust = 1.33,
  )
#> Warning in max(table(panel$xmin)): no non-missing arguments to max; returning
#> -Inf
```

<sup>Created on 2024-07-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

With DEV version downloaded just now, it errors
image

teunbrand:
This PR aims to fix #5995 and fix #3647.

Briefly, when `xmin` is absent from the data, `x` is used to count the number of groups per location.
In addition, some guardrails are in place to not use non-existing columns in the calculation.

Reprex from issue:
``` r
#> ℹ Loading ggplot2

palmerpenguins::penguins |>
  dplyr::count(species, sex) |>
  ggplot(
    aes(x = sex,
        y = n,
        fill = species,
        label = n),
  ) +
  geom_col(
    position = position_dodge(preserve = "single"),
    width = 0.75,
    alpha = 0.9,
  ) +
  geom_text(
    colour = "black",
    position = position_dodge(width = 0.75, preserve = "single"),
    vjust = 1.33,
  )
```

<sup>Created on 2024-07-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/position-dodge.R
+++ b/R/position-dodge.R
@@ -106,6 +106,11 @@ PositionDodge <- ggproto("PositionDodge", Position,
   reverse = NULL,
   setup_params = function(self, data) {
     flipped_aes <- has_flipped_aes(data, default = self$orientation == "y")
+    check_required_aesthetics(
+      if (flipped_aes) "y|ymin" else "x|xmin",
+      names(data), snake_class(self)
+    )
+
     data <- flip_data(data, flipped_aes)
     if (is.null(data$xmin) && is.null(data$xmax) && is.null(self$width)) {
       cli::cli_warn(c(
--- a/R/position-dodge.R
+++ b/R/position-dodge.R
@@ -117,8 +117,10 @@ PositionDodge <- ggproto("PositionDodge", Position,
     if (identical(self$preserve, "total")) {
       n <- NULL
     } else {
-      n <- vec_unique(data[c("group", "PANEL", "xmin")])
-      n <- vec_group_id(n[c("PANEL", "xmin")])
+      data$xmin <- data$xmin %||% data$x
+      cols <- intersect(colnames(data), c("group", "PANEL", "xmin"))
+      n <- vec_unique(data[cols])
+      n <- vec_group_id(n[setdiff(cols, "group")])
       n <- max(tabulate(n, attr(n, "n")))
     }
 
--- a/tests/testthat/test-position_dodge.R
+++ b/tests/testthat/test-position_dodge.R
@@ -37,3 +37,19 @@ test_that("position_dodge() can reverse the dodge order", {
   ld <- get_layer_data(p + geom_col(position = position_dodge(reverse = FALSE)))
   expect_equal(ld$label[order(ld$x)], c("A", "A", "B", "B", "C"))
 })
+
+test_that("position_dodge warns about missing required aesthetics", {
+
+  # Bit of a contrived geom to not have a required 'x' aesthetic
+  GeomDummy <- ggproto(NULL, GeomPoint, required_aes = NULL, optional_aes = "x")
+
+  p <- ggplot(mtcars, aes(cyl, disp, colour = factor(vs))) +
+    layer(
+      geom = GeomDummy,
+      stat = "identity",
+      position = position_dodge(width = 0.5),
+      mapping = aes(x = NULL)
+    )
+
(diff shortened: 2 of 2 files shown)


teunbrand:
Accidentally also fixes #3647:

Note previously recommended as workaround in https://github.com/tidyverse/ggplot2/issues/2712#issuecomment-399759709 was to combine `position_dodge()` in point layer with `position_dodge2()` in boxplot layer, but this workaround is no longer necessary.


``` r
#> ℹ Loading ggplot2

dodge <- position_dodge(0.75, preserve = "single")

ggplot(mtcars, aes(factor(am), mpg, fill = factor(carb))) + 
  geom_boxplot(position = dodge) +
  geom_point(position = dodge, shape = 21)
```



<sup>Created on 2024-07-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/position-dodge.R
```diff
@@ -117,8 +117,10 @@ PositionDodge <- ggproto("PositionDodge", Position,
     if (identical(self$preserve, "total")) {
       n <- NULL
     } else {
-      n <- vec_unique(data[c("group", "PANEL", "xmin")])
-      n <- vec_group_id(n[c("PANEL", "xmin")])
+      data$xmin <- data$xmin %||% data$x
+      cols <- intersect(colnames(data), c("group", "PANEL", "xmin"))
```
thomasp85, line 121: Since "group" and "PANEL" are required data columns and you just defined the xmin column, is there any instances where this would yield anything different than `c("group", "PANEL", "xmin")`?

thomasp85: I guess if both "xmin" and "x" are initially `NULL` - is that what is the intend?

teunbrand: The column-name intersection is just to protect against any potentially missing columns. I imagine a determined extension could avoid creating `group` and `PANEL` columns.

thomasp85: If so, it would break in a multitude of other places as both columns are assumed to exist all over the place

thomasp85 (approved):
Apart from the ongoing discussion where I am still not convinced this LGTM

R/position-dodge.R
teunbrand: If we assume all well-behaved extensions, then still xmin/x can be `NULL` as you pointed out. But then there'd be multiple downstream assumptions that no longer hold too.

thomasp85: It is not necessary to assume well-behaved extensions. The extensions would never work in any settings if they removed these columns so they _cannot_ exist in any meaningful way. If there is a chance that min/x is NULL then this should be guarded by a proper error

teunbrand: Alright I've thrown in a `check_required_aesthetics()` in here

teunbrand:
We can have another stab at ironing out details once #6101 is fleshed out

Issue #6008, opened 2024-07-19
Regressions related to #5879


teunbrand:
This is an issue meant to track fallout from #5879 as determined through reverse dependency check in dbea0a6.
Issue #6003 was already identified and fixed in #6004.

Another issue is that `names(ggplot())` no longer contains the `labels` field. This trips up some tests in dependencies that expect that `length(ggplot())` is some fixed value. This should be easy to fix.

Another issue is that people sometimes test `expect_equal(plot$labels$var, "value")`. As `plot$labels` is no longer filled upon adding layers, `plot$labels` is empty in many such cases and thus the tests fail. This affects >30 packages and is not straightforward to fix, so we'd need to decide wether to revert #5879 or ask maintainers to adapt their tests.

olivroy:
I think #5879 is a great addition! If you decide to keep this, I'd be happy to help submit PRs to other packages. Maybe helping moving the tests from tests that are quite sensitive and slow down ggplot2's process.

```r
expect_length(ggplot(), <specific length>
# better test
expect_s3_class(ggplot(), "ggplot")

```

Maybe ggplot2 could benefit from having a `get_aes_labels()` which could return a named vector and be helpful for ggplot2 extensions. (Please disregard if it doesn't make sense).

teunbrand:
Thanks for the offer @olivroy! The `length(ggplot())` issue is very fixable, so we wouldn't need to bother other maintainers for this. I like the idea of writing a getter for the labels.

PR #6010, opened 2024-07-23
Fix missing `labels` slot

teunbrand:
This PR aims to fix part of #6008.

Briefly, reverse dependencies sometimes make assumptions about `length(ggplot())` or `names(ggplot())`. In #5879 we removed the `labels` field, causing these assumptions to no longer hold. This PR brings back the `labels` field to make these assumptions valid again.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -130,7 +130,8 @@ ggplot.default <- function(data = NULL, mapping = aes(), ...,
     coordinates = coord_cartesian(default = TRUE),
     facet = facet_null(),
     plot_env = environment,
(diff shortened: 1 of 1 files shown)


thomasp85 (commented):
Maybe add a comment on why we keep this vestigal slot there...

teunbrand:
I wouldn't mark this as vestigial. It doesn't track layers anymore, but it still keeps labels if users use `+ labs(...)`.

thomasp85 (approved):
LGTM

PR #6011, opened 2024-07-23
Ignore empty non-mapped aesthetics

### LINKED ISSUE #6009: Regression related to #5807
Using `width = NULL` used to work, but it no longer does since we changed it from a parameter to an aesthetic.
Should we just warn + ignore 0-length aesthetics instead of throwing errors?

``` r
library(ggplot2)

ggplot(mpg, aes(drv)) +
  geom_bar(width = NULL)
```

``` r
# Load current main branch
#> ℹ Loading ggplot2

ggplot(mpg, aes(drv)) +
  geom_bar(width = NULL)
#> Error in `geom_bar()`:
#> Caused by error in `check_aesthetics()` at ggplot2/R/geom-.R:181:5:
#> ! Aesthetics must be either length 1 or the same as the data (3).
#> ✖ Fix the following mappings: `width`.
```

<sup>Created on 2024-07-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6009.

Briefly, when passing `NULL` or 0-length aesthetics in `layer()` it will raise a warning. Previously, this would raise an error. 
As far as I know, there isn't any situation in which it is desirable to pass a 0-length aesthetic. Even if the data has 0-rows, nothing should be shown and no information is to be derived from a 0-length aesthetic. Note that this PR only applies to aesthetics passed outside the `aes()` function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Passing empty unmapped aesthetics to layers raises a warning instead of
+  throwing an error (@teunbrand, #6009).
 * Missing values from discrete palettes are no longer translated 
   (@teunbrand, #5929).
 * Fixed bug in `facet_grid(margins = TRUE)` when using expresssions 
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #6019, opened 2024-07-31
Update economics script & data

bwu62:
Update `economics.R` script to use updated FRED file links and process correctly, then ran script to generated updated `economics` and `economics_long` data frames.

I recently found myself wanting to use this data set and was disappoint to learn it hadn't been updated in nearly a decade.


teunbrand:
Thanks for the PR! I can see why it might be useful to have an updated dataset, but ultimately this dataset exists in ggplot2 to make it easier to show example timeseries plots rather than serving as a reputable source of this data. @thomasp85 could you advise on ggplot2's policy here (in lieu thereof, just your thoughts)?

clauswilke:
I don't think the dataset should be updated. Its main purpose is to be used for documentation and reproducible examples and nobody would benefit if those existing plots suddenly change. There is no reason for this dataset to be kept up-to-date. We're also routinely using the `mtcars` dataset which has data from 1974.

PR #6021, opened 2024-08-02
chore: merge isoband/isolines functions

teunbrand:
This PR doesn't address an issue, but I see it as a small chore.

Briefly, `iso_to_path()` and `iso_to_polygon()` were nearly identical functions. 
This PR merges these with a switch, which simplifies reasoning about the contours somewhat as you only have to understand 1 function instead of 2.

### DIFF
--- a/R/stat-contour.R
+++ b/R/stat-contour.R
@@ -142,7 +142,7 @@ StatContourFilled <- ggproto("StatContourFilled", Stat,
 
     isobands <- withr::with_options(list(OutDec = "."), xyz_to_isobands(data, breaks))
     names(isobands) <- pretty_isoband_levels(names(isobands))
-    path_df <- iso_to_geom(isobands, data$group[1], "polygon")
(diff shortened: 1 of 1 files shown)


R/stat-contour.R
```diff
@@ -142,7 +142,7 @@ StatContourFilled <- ggproto("StatContourFilled", Stat,
 
     isobands <- withr::with_options(list(OutDec = "."), xyz_to_isobands(data, breaks))
     names(isobands) <- pretty_isoband_levels(names(isobands))
-    path_df <- iso_to_polygon(isobands, data$group[1])
+    path_df <- iso_to_geom(isobands, data$group[1], "polygon")
```
thomasp85, line 145:
```suggestion
    path_df <- iso_to_geom(isobands, data$group[1], geom = "polygon")
```

Just for symmetry with the above change

thomasp85 (approved):
LGTM

PR #6022, opened 2024-08-02
Clean up orphan functions and tiny one-off helpers

teunbrand:
This PR doesn't adress an issue, but I think it is worth to consider anyway.

In essence, it tries to cleanup three types of functions:

1. Orphan functions that were used at some point, but have fallen in disuse because surrounding code has been refactored.
2. Tiny helper functions that are used only once and can easily be used written out in full.
3. Mirror functions where two versions of nearly identical code existed, like the `simplify()`/`simplify_formula()` or `as.quoted()`/`as_quoted()`.


R/layer.R
```diff
@@ -463,8 +463,6 @@ Layer <- ggproto("Layer", NULL,
   }
 )
 
-is.layer <- function(x) inherits(x, "Layer")
```
thomasp85, line 466: I'd actually say we should export this rather than remove it, since we do so for all the other main classes

R/margins.R
```diff
@@ -8,9 +8,6 @@ margin <- function(t = 0, r = 0, b = 0, l = 0, unit = "pt") {
   class(u) <- c("margin", class(u))
   u
 }
-is.margin <- function(x) {
-  inherits(x, "margin")
-}
```
thomasp85, line 13: don't touch :-)

R/plot-construction.R
```diff
@@ -126,7 +126,16 @@ ggplot_add.labels <- function(object, plot, object_name) {
 }
 #' @export
 ggplot_add.Guides <- function(object, plot, object_name) {
-  update_guides(plot, object)
+  if (inherits(plot$guides, "Guides")) {
```
thomasp85, line 129: `if (is_guides(plot$guides)) {...` is more readable

R/utilities.R
```diff
@@ -84,16 +84,6 @@ clist <- function(l) {
...
-# Return unique columns
-# This is used for figuring out which columns are constant within a group
-#
-# @keyword internal
-uniquecols <- function(df) {
-  df <- df[1, sapply(df, is_unique), drop = FALSE]
-  rownames(df) <- seq_len(nrow(df))
-  df
-}
-
```
thomasp85, line 96: What is the purpose of moving this function?

thomasp85 (commented):
Bunch of comments since this is a large PR. Overall good though I have one main objection as you can see in the comments

R/grob-dotstack.R
```diff
@@ -14,15 +14,13 @@ dotstackGrob <- function(
         y <- unit(y, default.units)
     if (!is.unit(dotdia))
         dotdia <- unit(dotdia, default.units)
-    if (!is_npc(dotdia))
+    if (!unitType(dotdia) == "npc")
```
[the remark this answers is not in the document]
teunbrand: We backport a version of this: https://github.com/tidyverse/ggplot2/blob/c108758cc8cfa0ee3dafa6b72c88e8b3207312cf/R/backports.R#L20-L38

R/utilities.R
teunbrand: This is a helper is only used in `stat_summary()`, so it feels more natural to be defined there rather than in the 'global' utilities. The only thing moving this does, is preventing me (possibly others too) from wondering 'where is this function used again' every time I search something in the global utilities.

teunbrand:
I'm not against all these class tests like `is.margin()`, but let's discuss whether we should be consistent about exporting them.
If it is only for internal use, and they are used only once, then I think `inherits()` prevents some clutter.

teunbrand:
TODO: keep class testers, make exports consistent

R/coord-sf.R
```diff
@@ -585,8 +581,13 @@ coord_sf <- function(xlim = NULL, ylim = NULL, expand = TRUE,
...
 parse_axes_labeling <- function(x) {
-  labs = unlist(strsplit(x, ""))
-  list(top = labs[1], right = labs[2], bottom = labs[3], left = labs[4])
+  if (is.character(x)) {
+    x <- unlist(strsplit(x, ""))
+    x <- list(top = x[1], right = x[2], bottom = x[3], left = x[4])
+  } else if (!is.list(x)) {
+    cli::cli_abort("Panel labeling format not recognized.")
+  }
+  x
```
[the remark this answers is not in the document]
teunbrand: Good idea, thanks!

R/aes-evaluation.R
```diff
@@ -207,10 +207,6 @@ stage_scaled <- function(start = NULL, after_stat = NULL, after_scale = NULL) {
 # Regex to determine if an identifier refers to a calculated aesthetic
```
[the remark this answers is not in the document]
teunbrand: Yeah might as well include that in this PR

R/utilities.R
thomasp85: fair enough. While you touch it can you make it so it turns off row names rather than set it to numbers (using `.set_row_names(now(df)`)

R/utilities-checks.R
```diff
@@ -1,3 +1,86 @@
...
+#' @rdname is_tests
+is.scale <- function(x) inherits(x, "Scale")
+
+#' @export
+#' @rdname is_tests
+is.guide <- function(x) inherits(x, "Guide")
+
+#' @export
+#' @rdname is_tests
+is.guides <- function(x) inherits(x, "Guides")
```
thomasp85, line 66: I don't think this needs exporting as it is an internal structure like Scales

thomasp85:
I think the testers needs to be below the class definition, not in their own file. This is our general MO

R/utilities-checks.R
teunbrand: `guides()` returns a `<Guides>` class object, so it isn't entirely internal

thomasp85: ah, I see - then ignore this

R/utilities.R
teunbrand: Fixed in 13b8d36

R/utilities-checks.R
```diff
@@ -1,3 +1,86 @@
...
+
+#' @export
+#' @rdname is_tests
+is.facet <- function(x) inherits(x, "Facet")
+
+#' @export
+#' @rdname is_tests
+#' @usage
+#' # Layer related tests
+#' is.layer(x)
```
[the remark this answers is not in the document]
teunbrand: Reverted this as order is scrambled by placing test function after constructors

thomasp85 (approved):
LGTM

PR #6027, opened 2024-08-05
Repurpose coord expansion

### LINKED ISSUE #6020: Feature request: alternate `expand` parameter
A common reason to muck around with a scale's `expand` argument is to ensure 0-anchored geometries like `geom_bar()` or `geom_area)` touch the x-axis without a gap.

``` r
library(ggplot2)

ggplot(mpg, aes(class)) +
  geom_bar() +
  scale_y_continuous(expand = c(0, 0, 0.05, 0))
```

The `expand` argument in scales is not very intuitive and I just happen to remember `c(0, 0, 0.05, 0)` as a rote incantation to get what I want. The helper function, `expansion()`, is also no help in this case, as I want to asymmetrically expand the scale. Adding to the confusion is that `coord_cartesian(expand)`, which has the same argument name and fusses over the same setting, takes a true/false value instead of the four values.

What I'm proposing is to rewire the `coord_*(expand)` argument to take top/right/bottom/left true/false values. It mirrors the setup we have for `margin()`, so it shouldn't be too unfamiliar to regular users.

``` r
ggplot(mpg, aes(class)) +
  geom_bar() +
  coord_cartesian(
    expand = c(TRUE, TRUE, FALSE, TRUE)
  )
#> Error in if (!expand) {: the condition has length > 1
```

<sup>Created on 2024-08-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6020.

Briefly, it allows `expand` to take a length-4 logical vector for each of the positions (top, right, bottom, left).
Alternatively, one could use a named vector too.

``` r
#> ℹ Loading ggplot2

# Showing logical vector
ggplot(mpg, aes(class)) +
  geom_bar() +
  coord_cartesian(expand = c(TRUE, TRUE, FALSE, TRUE))
```

``` r
# Showing named vector
ggplot(mpg, aes(y = class)) +
  geom_bar() +
  coord_cartesian(expand = c(left = FALSE))
```

<sup>Created on 2024-08-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/coord-flip.R
```diff
@@ -89,6 +89,7 @@ CoordFlip <- ggproto("CoordFlip", CoordCartesian,
   },
 
   setup_panel_params = function(self, scale_x, scale_y, params = list()) {
+    params$expand <- params$expand[c(2, 1, 4, 3)]
```
thomasp85, line 92: Not sure flipping the coord params is the logical solution here

thomasp85 (approved):
I like this idea

PR #6030, opened 2024-08-05
Update CONTRIBUTING.md

olivroy:
to update links. I couldn't find a replacement for the dead link docs.ggplot2.org/dev/vignettes/development.html so I deleted this. Happy to add back a link if you know what this was referring to!


teunbrand (approved):
The whole document feels like it could use an update at some point, but thanks for fixing the links!

Issue #3196, opened 2019-03-20
Checking `...`


ColinFay:
Following this [twitter thread](https://twitter.com/hadleywickham/status/1108363335395794949). 

`labs()` uses `...`, but the name of the arguments are not checked, so this can lead to errors when passing `tile = ` instead of `title`, or any other typo. 

It would be nice for the user to have at least a warning (or an error) here:

```r
library(ggplot2)
ggplot(iris, aes(Sepal.Length, Sepal.Width)) + 
  geom_point() + 
  labs(tile = "Title has a typo")
```

hadley:
And in all other functions that use `...`

yutannihilation:
Does this require us to replace all `list()` with `rlang::list2()`? Otherwise, `list()` actually uses `...`.

``` r
f1 <- function(..., title = "foo") {
  ellipsis::check_dots_used()
  
  list(..., title = title)
}

f2 <- function(..., title = "foo") {
  ellipsis::check_dots_used()
  
  rlang::list2(..., title = title)
}

f1(tile = "Title has a typo")
#> $tile
#> [1] "Title has a typo"
#> 
#> $title
#> [1] "foo"
f2(tile = "Title has a typo")
#> Warning: Some components of ... were not used: tile
#> $tile
#> [1] "Title has a typo"
#> 
#> $title
#> [1] "foo"
```

<sup>Created on 2019-03-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

yutannihilation:
Hmm, I come to think we can do nothing for `labs()`; there's a chance `tile` is a valid aesthetic, and `labs()` takes arbitrary aesthetics, which means we cannot verify the arguments without annoying false positive warnings. For example:

``` r
library(ggplot2)

labs <- function (..., title = waiver(), subtitle = waiver(), caption = waiver(), 
          tag = waiver()) {
  ellipsis::check_dots_used()
  
  args <- rlang::list2(..., title = title, subtitle = subtitle, caption = caption, tag = tag)
  
  is_waive <- vapply(args, ggplot2:::is.waive, logical(1))
  args <- args[!is_waive]
  args <- args[!duplicated(names(args))]
  args <- ggplot2:::rename_aes(args)
  
  structure(args, class = "labels")
}

labs(tile = "Title has a typo")
#> Warning: Some components of ... were not used: tile
#> $tile
#> [1] "Title has a typo"
#> 
#> attr(,"class")
[...]

PR #6032, opened 2024-08-07
Warn about ignored scales and guides

### LINKED ISSUE #3743: Feature request: warning for  fill/colour mismatch between aes() and scale...()
Custom legends are sensitive to whether it's a "fill" or "colour" aes. Switching between one and the other with essentially the same data is common. 

Currently, if you forget to change e.g. "colour" to "fill" for your custom legend / colour scheme, ggplot2 ignores your colour scale / custom legend (since the "colour" aes is not specified) and uses the default for the colours / legend (since no information is provided for the "fill" aes).

Below, p1 uses "colour". p2 uses "fill" but the user forgot to update "scale_colour...".  

```
library(ggplot2)
library(RColorBrewer)

p1 = ggplot(iris, aes(x = Species, y = Sepal.Length, colour = Species)) +
  geom_boxplot() +
  scale_colour_brewer(
    name = 'Dude which plant',
    palette = 'Set2'
    )

p2 = ggplot(iris, aes(x = Species, y = Sepal.Length, fill = Species)) +
  geom_boxplot() +
  scale_colour_brewer(
    name = 'Dude which plant',
    palette = 'Set2'
    )

```

p2 uses default legend and colour scheme without telling the user.

For p2, it might be useful to have a warning saying "you used scale_colour but didn't specify a colour aesthetic." or something like that.

May be related to #2691 / #2239.

teunbrand:
This PR aims to fix #3743 and guides indentified in https://github.com/tidyverse/ggplot2/issues/3196#issuecomment-2270818015.

Briefly, when using a scale or guide for an aesthetic that is not in use, you'll get a warning.

Implementation-wise, it wasn't too straightforward to allow guides and scales to assess their usage.  A few remarks on this in the code.

An example:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  scale_fill_discrete() +
  guides(shape = "bar")
#> Warning: Ignoring unknown guide: "shape".
#> Warning: Ignoring scale for unused aesthetics: "fill".
```

<sup>Created on 2024-08-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/layer.R
+++ b/R/layer.R
@@ -354,7 +354,7 @@ Layer <- ggproto("Layer", NULL,
 
   compute_statistic = function(self, data, layout) {
     if (empty(data))
-      return(data_frame0())
+      return(data)
 
     self$computed_stat_params <- self$stat$setup_params(data, self$stat_params)
     data <- self$stat$setup_data(data, self$computed_stat_params)
@@ -362,7 +362,7 @@ Layer <- ggproto("Layer", NULL,
   },
 
   map_statistic = function(self, data, plot) {
-    if (empty(data)) return(data_frame0())
+    if (empty(data)) return(data)
 
     # Make sure data columns are converted to correct names. If not done, a
     # column with e.g. a color name will not be found in an after_stat()
@@ -418,7 +418,7 @@ Layer <- ggproto("Layer", NULL,
   },
 
   compute_geom_1 = function(self, data) {
-    if (empty(data)) return(data_frame0())
+    if (empty(data)) return(data)
 
     check_required_aesthetics(
       self$geom$required_aes,
@@ -430,7 +430,7 @@ Layer <- ggproto("Layer", NULL,
   },
 
   compute_position = function(self, data, layout) {
-    if (empty(data)) return(data_frame0())
+    if (empty(data)) return(data)
 
     params <- self$position$setup_params(data)
     data <- self$position$setup_data(data, params)
--- a/R/plot-build.R
+++ b/R/plot-build.R
@@ -102,14 +102,9 @@ ggplot_build.ggplot <- function(plot) {
 
   # Train and map non-position scales and guides
   npscales <- scales$non_position_scales()
-  if (npscales$n() > 0) {
-    lapply(data, npscales$train_df)
-    plot$guides <- plot$guides$build(npscales, plot$layers, plot$labels, data)
-    data <- lapply(data, npscales$map_df)
-  } else {
-    # Only keep custom guides if there are no non-position scales
-    plot$guides <- plot$guides$get_custom()
-  }
+  npscales$train(data)
+  plot$guides <- plot$guides$build(npscales, plot$layers, plot$labels, data)
+  data <- npscales$map(data)
   data <- .expose_data(data)
 
   # Fill in defaults etc.
--- a/R/plot-build.R
+++ b/R/plot-build.R
@@ -104,7 +104,7 @@ ggplot_build.ggplot <- function(plot) {
   npscales <- scales$non_position_scales()
   npscales$train(data)
   plot$guides <- plot$guides$build(npscales, plot$layers, plot$labels, data)
-  data <- npscales$map(data)
+  data <- npscales$map(data, layers)
   data <- .expose_data(data)
 
   # Fill in defaults etc.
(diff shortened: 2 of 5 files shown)


R/layer.R
```diff
@@ -354,15 +354,15 @@ Layer <- ggproto("Layer", NULL,
 
   compute_statistic = function(self, data, layout) {
     if (empty(data))
-      return(data_frame0())
+      return(data)
```
teunbrand, line 357: 0-row data still contains relevant information in terms of column names.

R/plot-build.R
```diff
@@ -102,14 +102,9 @@ ggplot_build.ggplot <- function(plot) {
...
-    lapply(data, npscales$train_df)
-    plot$guides <- plot$guides$build(npscales, plot$layers, plot$labels, data)
-    data <- lapply(data, npscales$map_df)
-  } else {
-    # Only keep custom guides if there are no non-position scales
-    plot$guides <- plot$guides$get_custom()
-  }
+  npscales$train(data)
+  plot$guides <- plot$guides$build(npscales, plot$layers, plot$labels, data)
+  data <- npscales$map(data, layers)
```
teunbrand, line 107: In order to warn about guides when there are no non-position scales, the guide building needed to be executed unconditionally.

R/scales-.R
```diff
@@ -61,13 +61,44 @@ ScalesList <- ggproto("ScalesList", NULL,
...
+  },
+
   train_df = function(self, df, drop = FALSE) {
     if (empty(df) || length(self$scales) == 0) {
       return()
     }
     lapply(self$scales, function(scale) scale$train_df(df = df))
   },
 
+  map = function(self, data, layers) {
```
teunbrand, line 78: I'm not really a fan of this as layers and scales should ideally be separate on only interact through the data, but I haven't found a better way around this.

R/scales-.R
```diff
@@ -61,13 +61,44 @@ ScalesList <- ggproto("ScalesList", NULL,
...
+    if (length(self$scales) == 0) {
+      return(data)
+    }
+    aesthetics <- lapply(self$scales, `[[`, "aesthetics")
+
+    known <- unique(unlist(c(
+      lapply(data, colnames),
+      lapply(layers, function(x) names(x$computed_mapping))
```
teunbrand, line 86: Sometimes you can have a mapping like `colour = after_scale(alpha(fill, 0.3))` that isn't instantiated as a column in the data yet. In order to not warn about these, we need to access the layer's mapping.

tests/testthat/test-guides.R
```diff
@@ -137,16 +142,13 @@ test_that("guide_none() can be used in non-position scales", {
...
-    plot$layers,
-    plot$labels
-  )
+  expect_length(ggplot_build(p)$plot$guides$guides, 0)
+
+  p <- ggplot(mpg, aes(cty, hwy, colour = class)) +
+    geom_point() +
+    guides(colour = guide_none())
 
-  expect_length(guides$guides, 0)
```
teunbrand, line 149: This test wasn't testing what is was supposed to be testing.

tests/testthat/test-guides.R
```diff
@@ -944,8 +951,8 @@ test_that("guides title and text are positioned correctly", {
     geom_point() +
     # setting the order explicitly removes the risk for failed doppelgangers
     # due to legends switching order
-    guides(shape = guide_legend(order = 1),
-           color = guide_colorbar(order = 2)) +
+    guides(size  = guide_legend(order = 2),
+           color = guide_colorbar(order = 1)) +
```
teunbrand, line 955: This plot wasn't using the `shape` aesthetic, so the test was wrong. The checks in this PR correctly picked up on this.

thomasp85:
This seems surprisingly complicated. Isn't there a point in the build process where all mapped aesthetics are known and we can check that list against the provided scales and guides?

teunbrand:
Agree that it is more complicated than one'd hope. The reason that the approach you outline doesn't currently work is because `Stat$default_aes` does not get absorbed into the computed mapping. For example `StatSum$default_aes` has an after-stat default for `size`, a `size` scale is correctly initialised, but there is no `size` mapping, so we'd spuriously warn in this case.

teunbrand:
I'm giving up on this, this is too complicated to be worth it.

PR #6035, opened 2024-08-15
Adding the option `seed = 0` into the jitter argument

### LINKED ISSUE #5875: Reproducible example for `position_jitter()` not working
It seems to me, the last of the [examples](https://ggplot2.tidyverse.org/reference/position_jitter.html#ref-examples) of `position_jitter()` does not work as intended. In other words: The following code...

https://github.com/tidyverse/ggplot2/blob/8e63882c3c36259785baac0eed5d51525bc40f70/R/position-jitter.R#L43-L47

...does not seem to simply shift the jittered points:

position_jitter-5

But maybe I misunderstood what the example is supposed to show?

agmurray:
Fixes #5875

documentation update

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -78,7 +78,7 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.1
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #6037, opened 2024-08-15
`show.legend` for factor data for all levels, even when not observed

### LINKED ISSUE #5869: Fill and color guides for factor with `drop=FALSE` don't show color.
In `ggplot2` 3.5.1 the fill guide for a factor where not all level sare present but `drop=FALSE` is used shows the label for the missing level but not the color. A simple example:

```r
library(ggplot2)
d <-data.frame(x = factor(1:3, levels = 1:4), y = 1:3)
ggplot(d, aes(x, y, fill = x)) + 
    geom_col() +
    scale_fill_discrete(drop = FALSE)
```

This was different in the version available a year ago. Compare
[this](https://homepage.divms.uiowa.edu/~luke/classes/STAT4580-2024/slides/maps.html#217) from 2024 to [this](https://homepage.divms.uiowa.edu/~luke/classes/STAT4580-2023/slides/maps.html#218) from 2023.

agmurray:
Added text to show how to add color to legends for all levels, even when no data for some levels exists.

Fixes #5869

### DIFF
--- a/R/layer.R
+++ b/R/layer.R
@@ -58,7 +58,9 @@
 #'   `NA`, the default, includes if any aesthetics are mapped.
 #'   `FALSE` never includes, and `TRUE` always includes.
 #'   It can also be a named logical vector to finely select the aesthetics to
-#'   display.
+#'   display.  This option can be used to include color for all levels, even 
+#'   when no data exists, use `TRUE`.  If 'FALSE', all levels are shown in legend, 
+#'   but omitting colors for unobserved levels.
 #' @param inherit.aes If `FALSE`, overrides the default aesthetics,
 #'   rather than combining with them. This is most useful for helper functions
 #'   that define both data and aesthetics and shouldn't inherit behaviour from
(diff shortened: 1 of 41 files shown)


R/layer.R
```diff
@@ -58,7 +58,9 @@
 #'   `NA`, the default, includes if any aesthetics are mapped.
 #'   `FALSE` never includes, and `TRUE` always includes.
 #'   It can also be a named logical vector to finely select the aesthetics to
-#'   display.
+#'   display.  This option can be used to include color for all levels, even 
+#'   when no data exists, use `TRUE`.  If 'FALSE', all levels are shown in legend, 
+#'   but omitting colors for unobserved levels.
```
teunbrand, line 63:
```suggestion
#'   display. To include legend keys for all levels, even 
#'   when no data exists, use `TRUE`.  If `NA`, all levels are shown in legend, 
#'   but unobserved levels are omitted.
```

teunbrand (changes requested):
Thanks for the pull request, I think this is a great start!
There are a few things that ideally should be worded slightly differently, please see the suggestion below.
Lastly, would you mind running `devtools::document()` to update the corresponding .Rd file as well?
If you'd like any help, please don't hesitate to stop Thomas or me!

teunbrand (approved):
Looks great, thank you so much for the contribution!

PR #6038, opened 2024-08-15
Add labeller example for optional argument in label_both

### LINKED ISSUE #5916: labeller example that uses an optional argument
None of the examples here: https://ggplot2.tidyverse.org/reference/labellers.html
show how to use the optional arguments to the functions. The naive `p + facet_grid(. ~ cyl, labeller = label_both(sep=" = "))` doesn't work, so one example would help readers and point to the recommended usage.

sierrajohnson:
Fixes #5916


thomasp85 (approved):
LGTM

PR #6040, opened 2024-08-15
`geom_smooth()`: Updated documentation, "confidence intervals" to "confidence bands"

### LINKED ISSUE #6025: geom_smooth: Prediction interval and not confidence interval for se in the documentation
Hello,
if I use `geom_smooth()`, then an interval is shown in the standard `se = TRUE`, which does not have a constant width. Especially for linear estimates of slope parameters.
However, the documentation states that `se` is a confidence interval (https://ggplot2.tidyverse.org/reference/geom_smooth.html):

> Display confidence interval around smooth? (`TRUE` by default, see `level` to control.)

However, this cannot be the case, as a confidence interval is always a fixed interval around the slope parameter.
What `geom_smooth()` shows, however, is the prediction interval (https://en.wikipedia.org/wiki/Prediction_interval).

The same for `level`.

Example image (from the documentation), dark grey area:

Or have I made a mistake?

If not, this should be changed.

phispu:
Fixes #6025

### DIFF
--- a/R/geom-smooth.R
+++ b/R/geom-smooth.R
@@ -9,7 +9,7 @@
 #' `predictdf()` generic and its methods.  For most methods the standard
 #' error bounds are computed using the [predict()] method -- the
 #' exceptions are `loess()`, which uses a t-based approximation, and
-#' `glm()`, where the normal confidence interval is constructed on the link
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #6042, opened 2024-08-15
Fixes #3533 - Swap date(time) types in opposite scale

### LINKED ISSUE #3533: Suggestion: improve error message on wrong use of scale_x_datetime

laurabrianna:
In both scale_*_date and scale_*datetime, added helpful error to suggest using the other type when appropriate ( #3533 )


thomasp85:
Thank you for this. On closer inspection I think that we should coerce the data into the right format rather than throw an error

R/scale-date.R
```diff
@@ -369,6 +369,10 @@ ScaleContinuousDatetime <- ggproto("ScaleContinuousDatetime", ScaleContinuous,
         i = "The value was converted to {obj_type_friendly(x)}."
       ), call = self$call)
     }
+    if (inherits(x, "Date")) {
+     cli::cli_abort("You supplied a {.cls {class(x)}} field.
+                    Did you mean to use {.fn scale_*_date}?", call = self$call)
```
thomasp85, line 374:
```suggestion
     x <- as.POSIXct(x)
```

R/scale-date.R
```diff
@@ -416,6 +420,10 @@ ScaleContinuousDate <- ggproto("ScaleContinuousDate", ScaleContinuous,
         i = "The value was converted to {obj_type_friendly(x)}."
       ), call = self$call)
     }
+    if (inherits(x, "POSIXct")) {
+      cli::cli_abort("You supplied a {.cls {class(x)}} field.
+                    Did you mean to use {.fn scale_*_datetime}?", call = self$call)
```
thomasp85, line 425:
```suggestion
      x <- as.Date(x)
```

thomasp85 (changes requested):

thomasp85 (commented):
@teunbrand would you look into finishing this off?

thomasp85 (approved):
LGTM

PR #6043, opened 2024-08-15
correct misleading error

### LINKED ISSUE #4460: misleading error message thrown by `cut_interval()`
`cut_interval()` throws a misleading error message when the user does not specify exactly one of `n` or `length`. In the worst case, if the user supplies `width` instead of `length` (as they would for `cut_width()`, you get a very fun error 😀 

``` r
ggplot2::cut_interval(x = 1:10, width = 10)
#> Error: Specify exactly one of n and width
```

<sup>Created on 2021-05-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The error gets thrown by `breaks()` [here](https://github.com/tidyverse/ggplot2/blob/87e9b85dd9f2a294f339d88a353d0c11c851489d/R/utilities-break.r#L96). Looks like an easy fix. Happy to send a PR if you don't want to bother doing it yourself 🙂

catalamarti:
fixes #4460

### DIFF
--- a/tests/testthat/test-utilities-break.R
+++ b/tests/testthat/test-utilities-break.R
@@ -1,0 +1,3 @@
+test_that("cut_interval gives the correct", {
+  expect_snapshot_error(cut_interval(x = 1:10, width = 10))
+})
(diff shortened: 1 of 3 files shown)


tests/testthat/test-utilities-break.R
```diff
@@ -0,0 +1,3 @@
+test_that("cut_interval gives the correct", {
```
thomasp85, line 1:
```suggestion
test_that("cut_interval throws the correct error message", {
```

thomasp85 (changes requested):
Looks good. Just needs a small wording change in the test

thomasp85 (approved):
LGTM

PR #6048, opened 2024-08-15
Export GeomBin2d

### LINKED ISSUE #5791: Could a GeomBin2d ggproto object be created?
It would helpful to have a GeomBin2d ggproto object exported.

This is because for `ggplot2::update_geom_defaults`, it's useful to have different defaults for GeomTile and GeomBin2d.

lgibson7:
Fix #5791


teunbrand (approved):
Thank you for the contribution and apologies for the delay!

PR #6049, opened 2024-08-19
legend key examples

thomasp85:
closes #4875 

The changes in #4875 were good but too much has happened in the repo to fix up the PR (most notably renaming of all files from lower-case to upper-case .R file ending)


teunbrand (approved):
LGTM

PR #6050, opened 2024-08-20
Fix some lints in the `R` folder

etiennebacher:
Hello, I'm building a new package to find and automatically fix lints in R code: [`flint`](https://flint.etiennebacher.com/).

I'm using real-life, large packages to check its performance and `ggplot2` is one of them. Since I already test on this, there's no additional cost for me in proposing those changes.

FYI, those changes were all generated by `flint::fix_dir("R")`.

There are many changes but most of them are trivial:
- replace `any(is.na())` by `anyNA()`
- replace the `=` assignment by `<-` (since both are used across the codebase, I guess `<-` should be the default)
- add a leading 0 to decimal (e.g `.1` -> `0.1`)

and a few other things.

`flint` is quite new and linter rules are necessarily a bit opinionated, so some of those changes might not be to your taste, but I'd be happy to have some feedback on this.

### DIFF
--- a/R/annotation-logticks.R
+++ b/R/annotation-logticks.R
@@ -165,7 +165,7 @@ GeomLogticks <- ggproto("GeomLogticks", Geom,
 
       names(xticks)[names(xticks) == "value"] <- x_name   # Rename to 'x' for coordinates$transform
       xticks <- coord$transform(xticks, panel_params)
-      xticks = xticks[xticks$x <= 1 & xticks$x >= 0,]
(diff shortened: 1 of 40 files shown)


teunbrand (approved):
Thanks Etienne, these look all great to me.
I was in the progress of preparing a big linting PR as well so this is great to see.

An additional thing ggplot2 might still want to replace is `seq_len(nrow(x))` with `vctrs::vec_seq_along(x)`, but that perhaps not for this PR.

teunbrand:
I'm going to merge this quickly before I bestow any merge conflicts on this PR, thanks for this!

PR #6053, opened 2024-08-21
Fix typos layer and geom_text docs

collinberke:
I was reading the `geom_text()` docs and came across these minor typos. Feel free to reject or modify any change that goes against the original intent of what was originally written.

Thanks for maintaining this awesome package!

### DIFF
--- a/R/geom-text.R
+++ b/R/geom-text.R
@@ -49,7 +49,7 @@
 #'   can be used in various ways, including to prevent overplotting and
 #'   improving the display. The `position` argument accepts the following:
 #'   * The result of calling a position function, such as `position_jitter()`.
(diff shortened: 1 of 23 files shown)


collinberke:
I just realized I didn't include all the files from `devtools::document()`, since many functions use `@inheritParam layer`. If I need to back this out and resubmit, let me know.

teunbrand:
Hi there, thanks for the PR! It would be nice to regenerate all the relevant docs. I don't think there is a need to back out / resubmit. You could just add a second commit to the same branch and push again, at which point this PR would update with the latest changes.

thomasp85 (approved):
LGTM

thomasp85:
Thank you!

PR #6060, opened 2024-08-27
Skip viewscales if graticule is empty

### LINKED ISSUE #6052: Error: Invalid index: field name 'x_start' not found
options(repr.plot.width=10,repr.plot.height=10)
p1 <- ggplot(nc) +
geom_sf(aes(fill=meta[ as.character(nc$cell),'cell_type']))+
theme_void()
p1
image
I'm trying to find a picture of this.
but
Error: Invalid index: field name 'x_start' not found

teunbrand:
This PR aims to fix #6052.

Briefly, ggplot2 got stuck when graticules are empty. In this PR, we skip computation of axis ticks when the graticule is empty.

Reprex:

``` r
library(ggplot2)
library(sf)
#> Linking to GEOS 3.12.1, GDAL 3.8.4, PROJ 9.3.1; sf_use_s2() is TRUE

# Data with out-of-bounds values
df <- st_sf(
  g = st_sfc(st_point(c(-600, 1200))),
  crs = 4326
)

# Old behaviour
ggplot(df) + geom_sf()
#> Warning in st_is_longlat(x): bounding box has potentially an invalid value
#> range for longlat data
#> Error: Invalid index: field name 'x_start' not found

#> ℹ Loading ggplot2

# New behaviour
ggplot(df) + geom_sf()
#> Warning in st_is_longlat(x): bounding box has potentially an invalid value
#> range for longlat data
```

<sup>Created on 2024-08-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

ligeiaKie:
hi. sry to disturb you but I didn't get how to fix "Error: Invalid index: field name 'x_start' not found". what does "devtools::load_all("~/packages/ggplot2/")" do in this reprex?

teunbrand:
It loads a local development version of ggplot2, which is just easier for me as somebody editing code than fully installing the development version every time I switch git branches. For regular use, you can install the development version with `pak::pak("tidyverse/ggplot2")` and use `library(ggplot2)` instead.

ligeiaKie:
> It loads a local development version of ggplot2, which is just easier for me as somebody editing code than fully installing the development version every time I switch git branches. It also signals to readers of the reprex that I'm not using the CRAN version of the package. For regular use, you can install the development version with `pak::pak("tidyverse/ggplot2")` and use `library(ggplot2)` instead.



> It loads a local development version of ggplot2, which is just easier for me as somebody editing code than fully installing the development version every time I switch git branches. It also signals to readers of the reprex that I'm not using the CRAN version of the package. For regular use, you can install the development version with `pak::pak("tidyverse/ggplot2")` and use `library(ggplot2)` instead.

thx you!!! I see it. I used to upgrade package from CRAN and problem is solved when install development version.

Issue #1967, opened 2016-12-22
geom_hex: Add weight parameter [feature request]


sjackman:
Please consider adding a weight parameter to `geom_hex`, similar to `geom_bin2d`. Thanks!

hadley:
It's already supported?

sjackman:
Ah, I was confused. Thanks. `weight` is not listed in the aesthetics for neither `geom_bin2d` nor `geom_hex`. Should it be?
http://docs.ggplot2.org/current/geom_bin2d.html
http://docs.ggplot2.org/current/geom_hex.html

hadley:
Yes! Want to do a PR?

sjackman:
Sure. Can do.

PR #6062, opened 2024-08-27
Mention `weight` as a supported aes for `gem_hex`

jansim:
I wanted to use `weight` as an aes with `geom_hex()` and was surprised to not find it mentioned in the docs for the function.

After a quick search it seems to actually be supported and just missing in the documentation (see #1967). This PR explicitly adds `weight` as an optional aes to geom_hex therefore so that it'll show up in the docs.

Note: I didn't add an entry to NEWS for this, since it's so small, but can do so if preferred.


teunbrand:
Hi there, thank you for the PR! I'm not sure that using an optional aesthetic is the right approach for this. It is valid input for StatBinHex, which is linked to, but technically independent of GeomBinHex. I think this might not be the only case where stat input goes undocumented, so perhaps it might help to deliberate in an issue first before designing PRs to tackle the problem.

jansim:
Ah sorry, my bad then. I saw in the original issue that the suggestion was to fix this and didn't think it necessary to reopen a discussion.

Maybe the preferred solution then would be to list the aesthetics for StatBinHex as well in the documentation for geom_hex? This seems to be in line with the approach used for [geom_bar](https://ggplot2.tidyverse.org/reference/geom_bar.html), which also lists aesthetics for stat_cout.

If this works as a solution I could just force push to this PR, else I can also open an issue and close this one.

teunbrand:
> Maybe the preferred solution then would be to list the aesthetics for StatBinHex as well in the documentation for geom_hex?

I like this idea, because it keeps the geom and stat separate while still allowing for documentation of stat aesthetics.

> didn't think it necessary to reopen a discussion

You are right, because I didn't realise we already had a way to document this that you pointed out in the comment above 😅

jansim:
Perfect, I've just updated the PR 👍️

teunbrand (approved):
Looks great, thanks for the PR!

PR #6063, opened 2024-08-27
Ink and paper settings for built-in themes (re: #6059)

teunbrand:
This PR reimplements #6059 due to a gnarly git history that I couldn't manage to resolve.
The reasoning for this PR remains unchanged, but I'll post the examples again so that you can see that these are the same:

``` r
#> ℹ Loading ggplot2

  geom_point()

p + theme_bw(ink = "dodgerblue")
```

``` r

p + theme_gray(paper = "orchid")
```

``` r

p + theme_light(ink = "white", paper = "black")
```

<sup>Created on 2024-08-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/theme-defaults.R
```diff
@@ -154,14 +156,14 @@ theme_grey <- function(base_size = 11, base_family = "",
     axis.line =          element_blank(),
     axis.line.x =        NULL,
     axis.line.y =        NULL,
-    axis.text =          element_text(size = rel(0.8), colour = "grey30"),
+    axis.text =          element_text(size = rel(0.8), colour = col_mix(ink, paper, 0.305)),
```
thomasp85, line 159: I assume there is a numerical reason why you need the third digit in some but not all of these mix operations?

thomasp85 (approved):
LGTM

R/theme-defaults.R
teunbrand: Yeah this is just to land on the right hex code to preserve visual tests. I have no scruples to just take the prettier number here (like 0.3).

PR #6066, opened 2024-08-28
WIP: Group summary stat

### LINKED ISSUE #3501: Can ggplot2 have a Stat that simply summarises data by group?
Every time I encounter a question like #3497, I wonder why ggplot2 doesn't have a `Stat` that simply applies a function by group. Though, in terms of the computational efficiency, it's generally better to have a summarised version of the data before entering ggplot2, it would be handy if we can summarise in ggplot2 especially when we generate plots one after another with different groupings.

I believe `StatSummary` *could have* been implemented to be able to summarise data with other groupings than `c("group", "x")` because the code following seems very general one: 

https://github.com/tidyverse/ggplot2/blob/b8420241309c8eea00d7086002c01cdf38a50eac/R/stat-summary.r#L163-L169

But, as the current `make_summary_fun()` expects a function that takes a vector, not a data.frame, it would be difficult to expand `StatSummary` to accept a function that summarises both `x` and `y`. So, to satisfy the need, I feel it might be nice to have some simple geom like below.

I don't see reasons why we shouldn't implement such a Stat. Am I missing something...?

``` r
library(ggplot2)

stat_summary_by_group <- function(mapping = NULL, data = NULL,
                                  geom = "pointrange", position = "identity",
                                  ...,
                                  fun.data = NULL,
                                  na.rm = FALSE,
                                  show.legend = NA,
                                  inherit.aes = TRUE) {
  layer(
    data = data,
    mapping = mapping,
    stat = StatSummaryByGroup,
    geom = geom,
    position = position,
    show.legend = show.legend,
    inherit.aes = inherit.aes,
    params = list(
      fun.data = fun.data,
      na.rm = na.rm,
      ...
    )
  )
}

StatSummaryByGroup <- ggproto("StatSummaryByGroup", Stat,
  compute_group = function(data, scales, fun.data = NULL, na.rm = FALSE) {
    summary <- fun.data(data)
    unique <- ggplot2:::dapply(data, c("group"), ggplot2:::uniquecols)
    unique[names(summary)] <- summary
    unique
  }
)

d <- data.frame(x = c(1:5, 3:7), y = 1:10, g = rep(c("a", "b"), each = 5), stringsAsFactors = FALSE)
f <- function(d) {
  data.frame(x = min(d$x), xend = max(d$x), y = mean(d$y), yend = mean(d$y))
}

ggplot(d) +
  geom_point(aes(x, y, colour = g)) +
  stat_summary_by_group(fun.data = f, aes(x, y, xend = stat(xend), yend = stat(yend)), geom = "segment") +
  facet_grid(cols = vars(g))
```

<sup>Created on 2019-08-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

teunbrand:
This is my stab to fix #3501.

It is still WIP because we might exchange thoughts on the name of the stat, and/or how it is used, expand documentation and tests. Initially I'd welcome any feedback on the direction of the PR.

To recap #3501, we generally want a more flexible summary stat that does not (necessarily) group by `x`. This is `stat_summarise()` in this PR (name is up for discussion).
The 'twist' I've given this PR is that the default summary function is a bit dandy, which we'll go over as I narrate the examples.

By default, the stat does nothing to the input data and renders points.

``` r
#> ℹ Loading ggplot2

  stat_summarise()
```

In #3501, it was suggested that we can pass any function to the stat that returns a data.frame.
This is indeed how this stat works.

``` r
ggplot(mpg, aes(drv, hwy)) +
  geom_jitter(aes(colour = drv)) +
  stat_summarise(
    geom = "segment",
    fun = function(data) {
      data.frame(
        x = min(data$x) - 0.5, 
        xend = max(data$x) + 0.5, 
        y = mean(data$y))
    }
  )
```

However, the 'trick' in this PR is that the default summary function is an NSE function that evaluates expressions passed to `fun.args`. Hence, instead of being bothered to construct a separate function for every task, we can dynamically pass NSE expressions this way.

``` r
ggplot(mpg, aes(drv, hwy)) +
  geom_jitter(aes(colour = drv)) +
  stat_summarise(
    geom = "segment",
    fun.args = vars(
      xend = max(x) + 0.5, 
      x = min(x) - 0.5, 
      y = mean(y)
    )
  )
```

Like some other tidyverse functions, the expressions are evaluated sequentially, meaning that later expressions can use results of prior expressions. This allows us to, for example, compute a convex hull in a reasonably straightforward manner.

``` r
  geom_point() +
  stat_summarise(
    geom = "polygon",
    fun.args = vars(
      hull = chull(x, y),
      x = x[hull],
      y = y[hull]
    ),
    fill = NA
  )
```

The thing that works differently relative to similarly flavoured functions (e.g. summarise, reframe, mutate in {dplyr}) is that not every evaluated expression needs to result in column-compatible data right away. This allows us to declare temporary variables, as long as they're explicity removed.

``` r
  ggplot(mpg, aes(displ, colour = drv)) +
  stat_summarise(
    geom = "line",
    fun.args = vars(
      dens = density(x), # temporary variable
      x = dens$x, y = dens$y,
      dens = NULL # delete variable
    )
  )
```

<sup>Created on 2024-08-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85:
I'm not sure I think the very minor decrease in code length warrant the completely new mode of specifying a function, TBH

teunbrand:
The stat just takes a regular `fun` argument like any other summary stat. The stat itself doesn't really have any special tricks up its sleeve.
It is the default `fun` dat allows you to specify how variables should be manipulated using NSE.

One can achieve something similar using `stat_summary()` too:

``` r
library(ggplot2)
library(rlang)

fun <- function(data, ...) {
  if (!is.data.frame(data)) {
    data <- data.frame(y = data)
  }
  exprs <- list2(...)
  if (length(exprs) < 1) {
    return(data)
  }
  data <- unclass(data)
  nms <- names(exprs)
  for (i in seq_along(exprs)) {
    data[[nms[[i]]]] <- eval_tidy(exprs[[i]], data)
  }
  vctrs::data_frame(!!!data[unique(names(exprs))])
}

ggplot(mpg, aes(drv, displ)) +
  stat_summary(
    fun.data = fun,
    fun.args = vars(
      ymin = min(y),
      ymax = max(y),
      y    = mean(y)
    ),
    geom = "pointrange"
  )
```



<sup>Created on 2024-09-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
> I think it should be called stat_manual() and the default fun value should be identity. 

I like this better than the current form

teunbrand:
Closing in favour of #6103

PR #6070, opened 2024-09-02
Reversal of coordinates

### LINKED ISSUE #4021: scale_x_reverse( ) in renderPlot ( ) or renderPlotly( )
Since I updated R to version 4.0 the scale_x_reverse () function has stopped working together with renderPlot () or renderPlotly( ), which I use in a shiny app that I develop ...
However when I test scale_x_reverse () out of the render it works fine. It is worth mentioning that in all other versions of R it was working normally in both Windowns (7, 8 and 10) and Linux (Ubunto, mint, Debian, Suse). Now I am use Windowns 10, R (4.0), RStudio (1.2.5042).

Can someone help me?

---

Brief description of the problem

```r
# ` # Variables
  peran_multi <<- 0
  ysup_multi <- max(NMRData[1,])
  yinf_multi <- min(NMRData[1,])
  ysup_multi <- ysup_multi + ysup_multi*0.03
  chkzoom_multi <<- 1
  idb_multi <<- 0


  # Data.frame
  testy_multi <<- data.frame(Chemical_Shift=CS_values_real[1,],Spectrum=NMRData[1,])
  ranges_multi <- reactiveValues(x = c((min(testy_multi$Chemical_Shift)), max(testy_multi$Chemical_Shift)), y = c(-100,ysup_multi))
  # ranges_multi_plot <- reactiveValues(x = sort(c((min(testy_multi$Chemical_Shift)), max(testy_multi$Chemical_Shift)), decreasing = TRUE), y = c(yinf_multi,ysup_multi))
  spectrums_multi <- reactiveValues(dat = data.frame(Chemical_Shift=CS_values_real[1,],Spectrum=NMRData[1,]))

  # Plot
    output$plot_multi <- plotly::renderPlotly({

      ggplot2::ggplot(spectrums_multi$dat,ggplot2::aes(Chemical_Shift,Spectrum)) + ggplot2::geom_line(color='blue') + ggplot2::scale_x_reverse() +

        ggplot2::coord_cartesian(xlim = ranges_multi$x, ylim = ranges_multi$y, expand = FALSE) +

        ggplot2::theme(axis.text.x = ggplot2::element_text(size = 12, color = "#000000"),
                       axis.text.y = ggplot2::element_text(size = 12, color = "#000000"),
                       title = ggplot2::element_text(face = "bold", color = "#000000", size = 17),
                       axis.title = ggplot2::element_text(face = "bold", color = "#000000", size = 15)
                       ) +

        ggplot2::labs(x = "Chemical Shift", y = "Intensity")

    })
`
```

### LINKED ISSUE #4801: `scale_*_reverse()` should not be limited to continuous data
`scale_*_reverse()` should allow discrete data that we have comparators for (or anything we can reorder using `desc`). For instance, strings can be ordered alphabetically. `scale_*_reverse()` would be the the most intuitive choice one would think about when trying to reverse their axes. However, running it with discrete data yields the following error:

```r
baz <- tibble(foo = c(0, 1, 2), bar = c("b", "a", "c"))
ggplot(baz, aes(x = foo, y = bar)) +
  geom_point() +
  scale_y_reverse()
```

```Error in -x : invalid argument to unary operator```

Replacing the last line with `scale_y_reverse(NULL)` also returns the same error.

If `scale_*_reverse()` cannot be changed in such a manner as to be able to accept discrete data, perhaps a simple function `scale_*_reverse_discrete()` may be helpful, rather than the current workaround below:

```r
ggplot(baz, aes(x = foo, y = bar)) +
  geom_point() +
  scale_y_discrete(limits = rev)
```

(inspired by https://stackoverflow.com/questions/28391850/reverse-order-of-discrete-y-axis-in-ggplot2)

EDIT: Fixed formatting, removed extraneous NULL

### LINKED ISSUE #5918: `scale_*_continuous` fails to render breaks if axis is reversed

teunbrand:
This PR aims to fix #4021 and fix #4801 and fix #5918.

Briefly, several coords get a `reverse` argument that can reverse the x/y/theta/r direction of a coord.

Importantly, this PR gets rid of the behaviour that giving the limits in reverse order implies `transform = "reverse"` in the scale, for the following reasons;
* This PR gives  the `reverse` argument to get the same effect.
* The behaviour is undocumented.
* The behaviour is applies inconsistently.
* The behaviour is incompatible with other scale transformations, like date.
* The reversed limits had to be hardcoded, which makes the reversal gnarly to apply if not known a priori, especially for discrete or date scales.

The behaviour continues to work in `xlim()`/`ylim()` because that is the only place where it *is* documented.

Also worth noting:
* `coord_radial(direction)` was essentially doing the same thing, but just for 'theta'. This is now deprecated in favour of the `reverse` argument that works for both 'theta' and 'r'.
* `guide_grid()` has been rewritten to accommodate coord transformations/munching. This makes plots ~1ms slower to render.
    * As a consequence, `coord_radial()` now renders minor gridlines for the 'r' direction, which I didn't even realise were missing in the first place. This is the visual change you see in the snapshots.

Alright, some examples then. Reversing a discrete direction:

``` r
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(class, hwy)) +
  geom_boxplot()

p + coord_cartesian(reverse = "x")
```

Showing that we can reverse transformed directions:

``` r
p + coord_trans(y = "sqrt", reverse = "y")
```

Showing that we can reverse directions for `coord_sf()`:

``` r
world1 <- sf::st_as_sf(maps::map('world', plot = FALSE, fill = TRUE))
ggplot(world1) + 
  geom_sf() + 
  coord_sf(reverse = "xy")
```

Showing reversal for `coord_radial()`. Please note the anti-clockwise direction of 'theta' and outside-in direction of 'r', and preservation of the start/end angles.

``` r
  geom_point() +
  coord_radial(
    start = 0.25 * pi, end = 1.5 * pi,
    reverse = "thetar"
  )
```

<sup>Created on 2024-09-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #6072, opened 2024-09-03
Improve consistency of naming internal functions

### LINKED ISSUE #6024: Naming of internal helper functions
This dawned on me while preparing #6022, but the naming scheme for several (internal) functions is at times somewhat confusing.

Especially now that we're using the imported `check_*()` functions from {rlang}, it feels out of place to have ggplot2-native `check_*()` functions do something else than throwing errors/warnings. For example `check_linewidth()` returns data.

The inverse is also true, we have functions that are used solely for throwing errors/warnings, but don't have the `check_*()` prefix. An example of this is `validate_element()`.

I understand these names precede the rlang checks, but it probably won't hurt to make the names more consistent.

teunbrand:
This PR aims to fix #6024.

Briefly, it enforces some consistency on the various `check_*()` and `validate_*()` functions in ggplot2's unexported functions.

Check functions:
* Check input verbatim
* Their purpose is to throw conditions when input is unexpected
* Return nothing (that is, `invisible(NULL)`)

Validator functions:
* Can take different variants of input
* Their purpose is to converge output from input
* Return a single type of output

In addition, a smaller category are functions that take possibly deprecated values and restore it to valid values. I've named these `update_*()` functions here, but that might be confusing and perhaps we could choose a better name.


thomasp85:
An alternative name to `update_*()` would be `fix_*()` since it implies something is wrong and that this function will do something about it and return a usable version?

thomasp85 (approved):
LGTM

PR #6073, opened 2024-09-04
Fix mistake in alignment docs for `geom_text()`

### LINKED ISSUE #6033: `geom_text()` justification docs are wrong
The justification docs [on this page](https://ggplot2.tidyverse.org/reference/geom_text.html#alignment) (`?geom_text`) are wrong (or at least misleading). These state that a value of 0 is right/bottom and a value of 1 is top/left. Whereas the [specs here](https://ggplot2.tidyverse.org/articles/ggplot2-specs.html#justification) rightly state that 0 = bottom/left and 1 = top/right. I think the confusion comes from the fact that left justification moves items to the right and vice versa.

``` r
library(ggplot2)
    geom_point() +
    geom_text(hjust = 0)
```

``` r
    geom_point() +
    geom_text(hjust = 1)
```

<sup>Created on 2024-08-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR fixes #6033. The title pretty much covers the PR's content.

### DIFF
--- a/R/geom-text.R
+++ b/R/geom-text.R
@@ -31,8 +31,8 @@
 #'
 #' @section Alignment:
 #' You can modify text alignment with the `vjust` and `hjust`
-#' aesthetics. These can either be a number between 0 (right/bottom) and
-#' 1 (top/left) or a character (`"left"`, `"middle"`, `"right"`, `"bottom"`,
+#' aesthetics. These can either be a number between 0 (left/bottom) and
+#' 1 (top/right) or a character (`"left"`, `"middle"`, `"right"`, `"bottom"`,
 #' `"center"`, `"top"`). There are two special alignments: `"inward"` and
 #' `"outward"`. Inward always aligns text towards the center, and outward
 #' aligns it away from the center.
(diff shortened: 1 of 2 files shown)


R/geom-text.R
```diff
@@ -31,8 +31,8 @@
 #'
 #' @section Alignment:
 #' You can modify text alignment with the `vjust` and `hjust`
-#' aesthetics. These can either be a number between 0 (right/bottom) and
-#' 1 (top/left) or a character (`"left"`, `"middle"`, `"right"`, `"bottom"`,
```
yutannihilation, line 35: There's no problem with this fix, but I personally prefer "left/top" to match the order of "right/bottom". What do you think?

teunbrand: I like the order matching part, but it should be left/bottom & right/top then

yutannihilation: Ah, sorry, you are right. For some reason, I looked at the old description...

yutannihilation (approved):
Thanks for updating. Looks good!

teunbrand:
Thansk for the review!

PR #6075, opened 2024-09-04
Replace confusing wording for themes in vignette

### LINKED ISSUE #5927: Superfluous guidance in "Creating a new theme" section of ggplot2 packages vignette?
[Motivated by Posit Forum post [Creating a new ggplot2 theme](https://forum.posit.co/t/creating-a-new-ggplot2-theme/187065).]

The [Creating a new theme](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html#creating-a-new-theme) section of the [vignette](https://github.com/tidyverse/ggplot2/blob/main/vignettes/ggplot2-in-packages.Rmd) on using ggplot2 in packages creates a new theme called `theme_custom()` and advises using a wrapper to apply the new theme:
```
default_theme <- function() {
  theme_custom()
}

mpg_drv_summary2 <- function() {
  mpg_drv_summary() + default_theme()
}
```
However, since `theme_custom()` is already a function, the reason for the wrapper in not clear.   The motivation given is that:
> [...i]f not, the theme object is stored in the compiled bytecode of the built package, which may or may not align with the installed version of ggplot2! 

but my understanding is that this shouldn't happen unless, say, the package contained a call to `theme_set()`.

Is the wrapper superfluous, or am I missing something?

teunbrand:
This PR aims to fix #5927.

It does so by no longer advising to use a pointless theme wrapper.

### DIFF
--- a/vignettes/ggplot2-in-packages.Rmd
+++ b/vignettes/ggplot2-in-packages.Rmd
@@ -190,15 +190,11 @@ theme_custom <- function(...) {
 mpg_drv_summary() + theme_custom()
 ```
 
-It is important that the theme be calculated after the package is loaded. If not, the theme object is stored in the compiled bytecode of the built package, which may or may not align with the installed version of ggplot2! If your package has a default theme for its visualizations, the correct way to load it is to have a function that returns the default theme:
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

PR #6076, opened 2024-09-04
Enable 2D structures as aesthetics

### LINKED ISSUE #4189: Feature request: support for tibble aesthetics

### LINKED ISSUE #6090: Feature request: accept data.frame as scale `labels`
This is a very niche and selfish problem as I'm trying some stuff out in {gguidance}.
It somewhat relates to #6076 a bit, but I'd like the following code:

``` r
library(ggplot2)

  geom_point() +
  scale_x_continuous(
    breaks = 2:7,
    labels = data.frame(foo = LETTERS[1:6], bar = 1:6)
  )

get_guide_data(p, aesthetic = "x")
#> ! `breaks` and `labels` have different lengths.
```

To return the following data.frame (note that the `.label` column is a nested data.frame rather than two separate `.label.foo` and `.label.bar` columns):

```r
#>           x .value .label.foo .label.bar y
#> 1 0.1127946      2          A          1 0
#> 2 0.2811448      3          B          2 0
#> 3 0.4494949      4          C          3 0
#> 4 0.6178451      5          D          4 0
#> 5 0.7861953      6          E          5 0
#> 6 0.9545455      7          F          6 0
```

The justification I have for this is that {vctrs} treats data.frames as vectors where rows are treated as elements. We don't really have a good reason to deviate from this behaviour.

teunbrand:
This PR aims to fix #4189 and fix #6090.

Briefly, this PR would allow people to use 2D structures like matrices, data.frames and tibbles as aesthetics.
We just had to adopt the {vctrs} way of measure sizes and setting names at select places for this to.
Importantly, this does *not* implement appropriate scales for such aesthetics, but it removes a barrier for extension developers to implement them.

``` r
#> ℹ Loading ggplot2

df <- mtcars
df$tib <- tibble::tibble(mpg = df$mpg, disp = df$disp)
df$mtx <- cbind(mpg = df$mpg, disp = df$disp)

p <- ggplot(df, aes(cyl, disp, dummy = tib)) +
  geom_point()

layer_data(p)$dummy |> head()
#> Don't know how to automatically pick scale for object of type
#> <tbl_df/tbl/data.frame>. Defaulting to continuous.
#> # A tibble: 6 × 2
#>     mpg  disp
#>   <dbl> <dbl>
#> 1  21     160
#> 2  21     160
#> 3  22.8   108
#> 4  21.4   258
#> 5  18.7   360
#> 6  18.1   225

layer_data(p + aes(dummy = mtx))$dummy |> head()
#>       mpg disp
#> [1,] 21.0  160
#> [2,] 21.0  160
#> [3,] 22.8  108
#> [4,] 21.4  258
#> [5,] 18.7  360
#> [6,] 18.1  225
```

<sup>Created on 2024-09-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


teunbrand:
I'm unsure why the coverage CI is failing

teunbrand:
TODO: check errors in scales

teunbrand:
Also incorporated #6090 as it is thematically related in the sense of "ggplot2 should give vctrs treatment to input".

thomasp85 (approved):
LGTM

PR #6077, opened 2024-09-04
Label dictionary

### LINKED ISSUE #5178: Label dictionaries
I recently saw [this tweet](https://twitter.com/tjmahr/status/1621605306898780160?s=20&t=pXcytO_xalMj3OTNCLiWww), which I thought was a nice suggestion, because it allows you to reuse labels without falling victims to typos and such. It still required a lot of manual management though.

Then, I realised we have several related issues/feature requests already, for example #4631 or #4313. However, I think having a label dictionary is perhaps a more elegant approach than hardcoding a path to retrieve labels from attributes (see also https://github.com/tidyverse/ggplot2/issues/4631#issuecomment-1102602900).

For the vanilla case, I think it should work something like this:
``` r
library(ggplot2)

label_dictionary <- c(
  displ = "Engine displacement [L]",
  hwy   = "Highway miles per gallon",
  "factor(cyl)" = "Number of cylinders" # Maybe I should be able to omit the function call?
)

# What I would like to specify
  geom_point() +
  labs(dict = label_dictionary)

# What I would want it to look like
  geom_point() +
  labs(!!!setNames(label_dictionary, c("x", "y", "colour")))
```

<sup>Created on 2023-02-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

However, for the {Hmisc/tinylabels/labelled/sjlabelled} cases, we might be able to specify `dict = derive()` (or similar) to specifically instruct to lookup label names in the data.

teunbrand:
This PR aims to fix #5178.

Briefly, it adds `labs(dict)` that allows one to use a data-dictionary to label a plot based on variable names (rather than aesthetics).

Let's just jump into examples.
The premise of this idea is that somewhere in your analysis code, you have some nice labels about what variables in your dataset mean. For example, we could have the following for the `mpg` dataset.

``` r
#> ℹ Loading ggplot2

dict <- c(
  displ = "Engine Displacement",
  hwy   = "Highway miles per gallon",
  cty   = "City miles per gallon",
  drv   = "Drive train",
  manufacturer = "Manufacturer name",
  model = "Model name",
  year  = "Year of manufacture",
  cyl   = "Number of cylinders",
  trans = "Type of transmission",
  fl    = "Fuel type",
  class = "Type of car"
)
```

This PR lets you slap on such a dictionary to your labels, and all variable names will be translated. The benefit is that you only have to think about pretty lables for variables once and you needn't worry about them again.

```r
ggplot(mpg, aes(class, cty, fill = drv)) +
  geom_boxplot() +
  labs(dict = dict)
```

Noteably, this doesn't work when having more complex expressions, like `factor(cyl)` instead of `cyl`. In such case, you can fall back to labelling the aesthetic, or you can add an entry like `labs(dict = c(dict, `factor(cyl)` = dict[["cyl"]]))`. Also we can reuse the dictionary here because we're using the same dataset even though we're making a totally different plot.

``` r
  geom_point() +
  labs(dict = dict, colour = dict["cyl"])
```

<sup>Created on 2024-09-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


larmarange:
This is a nice idea. Would it be better to have a more explicit argument, i.e. `dictionary` instead of `dict`?

teunbrand:
I don't have too strong of an opinion on this, but I like that both `labs()` is terse and `dict` is terse.
Perhaps Thomas can render a decision on this one

teunbrand:
Double check: do we overwrite labs if their value is already inside?

shannonpileggi:
I love this idea! Thank you for initiating this work!

Just curious - have you given thought to workflows where the user doesn't actually specify a dictionary? For example, what if the data already has labels embedded, and the user wants those embedded labels to be displayed automatically? This would be a similar approach to the {gt} package.

``` r
library(gt)

# unlabelled data frame
df1 <- data.frame(x = 1)
gt(df1)
```

<div id="ffkqdeazma" style="padding-left:0px;padding-right:0px;padding-top:10px;padding-bottom:10px;overflow-x:auto;overflow-y:auto;width:auto;height:auto;">

<table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">
  <thead>
    <tr class="gt_col_headings">
      <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="x">x</th>
    </tr>
  </thead>
  <tbody class="gt_table_body">
    <tr><td headers="x" class="gt_row gt_right">1</td></tr>
  </tbody>
  &#10;  
</table>
</div>

``` r

# labelled data frame
df2 <- df1
attr(df2$x, "label") <- "variable x description"
gt(df2)
```

<div id="sikaogivnb" style="padding-left:0px;padding-right:0px;padding-top:10px;padding-bottom:10px;overflow-x:auto;overflow-y:auto;width:auto;height:auto;">

<table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">
  <thead>
    <tr class="gt_col_headings">
      <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="variable x description">variable x description</th>
    </tr>
  </thead>
  <tbody class="gt_table_body">
    <tr><td headers="x" class="gt_row gt_right">1</td></tr>
  </tbody>
  &#10;  
</table>
</div>

<sup>Created on 2024-10-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
Hi Shannon! We implemented something akin to what you describe in #5879 which is already merged in the dev version. You can give the dev version a spin if you like (and provide feedback if we can improve 😅 )

shannonpileggi:
Ohhhhh thank you for pointing that out! Will take a look!

thomasp85 (approved):
LGTM

R/labels.R
```diff
@@ -84,6 +84,15 @@ setup_plot_labels <- function(plot, layers, data) {
...
 
+  dict <- plot_labels$dictionary
+  if (length(dict) > 0) {
+    labels <- lapply(labels, function(x) {
+      dict <- dict[names(dict) %in% x]
+      x[match(names(dict), x)] <- dict
+      x
+    })
+  }
+
```
thomasp85, line 95: This is being called on all labels, right? So if a user has given a label name that just happens to be in the dictionary it will get translated

To me it should only apply to fallback labels, but I'm open to hearing against this

thomasp85: Can't roll back my approval but let's talk this one out before merging :-)

teunbrand: It applies to all 'derived' labels, not labels that users give verbatim.
Is this the kind of situation you mean?

``` r
#> ℹ Loading ggplot2

  geom_point() +
  labs(
    x = "Engine Displacement",
    dictionary = c(displ = "NO NOT THIS ONE", hwy = "new label")
  )
```



<sup>Created on 2024-12-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand: Also `labs(x = "foo", dictionary = c(foo = "bar"))` will give `"foo"` as label.

PR #6078, opened 2024-09-05
Label accessor

### LINKED ISSUE #6008: Regressions related to #5879

teunbrand:
This PR aims to fix #6008.

Briefly, this ameliorates the fallout of #5879 by providing a convenient accessor for completed labels that people can access.
The `get_labs()` function just returns all known labels for a plot.

``` r
#> ℹ Loading ggplot2

  geom_point(aes(colour = drv, shape = drv))

# Note that colour/shape share a legend
str(get_labs(p))
#> List of 7
#>  $ colour: chr "drv"
#>  $ shape : chr "drv"
#>  $ x.sec : NULL
#>  $ x     : chr "displ"
#>  $ y     : chr "hwy"
#>  $ y.sec : NULL
#>  $ alt   : chr ""
```

Reverse dependencies that test for the label can then declare the following in their test helpers, and use this instead of trying to access the values directly.

```
get_labs <- if ("get_labs" %in% getNamespaceExports("ggplot2")) {
  ggplot2::get_labs
} else {
  function(plot) plot$labels
}
```

<sup>Created on 2024-09-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #6079, opened 2024-09-05
Change `fill` of `GeomRibbon` to match with the previous default colour

yutannihilation:
This fixes an unintentional change in https://github.com/tidyverse/ggplot2/pull/5833.

---

**v3.5.1** (https://ggplot2.tidyverse.org/reference/geom_ribbon.html#ref-examples):
image

**dev** (https://ggplot2.tidyverse.org/dev/reference/geom_ribbon.html#ref-examples):
image


yutannihilation:
I just happened to catch this when I tried to understand how to determine these values... I think the other values are correct. Thanks for the prompt review!

Issue #5760, opened 2024-03-07
Switch to quarto for vignettes?


teunbrand:
Per release notes found [here](https://github.com/quarto-dev/quarto-r/releases/tag/v1.4), quarto is now available as vignette builder.
Should ggplot2 switch vignettes to quarto? Is pkgdown prepared to deal with .qmd vignettes?

PR #6083, opened 2024-09-06
Fix the YAML syntax of fig.alt chunk option

yutannihilation:
Positron warns me a lot about the YAML syntax of `fig.alt`. Interestingly, if I re-render the README, there's no difference on `alt` of the corresponding `` tag. So, it seems R Markdown is more permissive than Positron, and this is not a real problem. But, I think it's good to use valid YAML syntax anyway.

image


teunbrand (approved):
Looks good to me!
Probably would need these changes as well if we decide to implement #5760.

yutannihilation:
Ah, probably yes. I actually used Quarto for validating the syntax because Quarto doesn't allow `#| fig.alt =`. Thanks for reviewing!

PR #6097, opened 2024-09-11
Fix linetype regression

### LINKED ISSUE #6096: Regression in linetype scales
This is a report on a regression found in revdep checks, that presents as follows:

https://github.com/tidyverse/ggplot2/blob/d31c051bb87cd5d5f02e8d19b5c002ad2a129d6c/revdep/problems.md?plain=1#L374-L375

Briefly, setting a numeric linetype palette will throw error because `na.value = "blank"` which will cast the linetype as characters, which grid will not accept. I propose setting `na.value = NA`.

``` r
#> ℹ Loading ggplot2

p <- ggplot(economics_long, aes(date, value01)) +
  geom_line(aes(linetype = variable))

p + scale_linetype_manual(values = 1:5)
#> Error in grid.Call.graphics(C_lines, x$x, x$y, index, x$arrow): invalid line type: must be length 2, 4, 6 or 8

p + scale_linetype_manual(values = 1:5, na.value = NA)
```

<sup>Created on 2024-09-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6096.

Briefly, it replaces `na.value = "blank"`, which casts mapped value to characters, to `na.value = NA`, which will cast `NA` to whatever type mapped values have.

### DIFF
--- a/R/scale-linetype.R
+++ b/R/scale-linetype.R
@@ -35,7 +35,7 @@
 #'   scale_linetype_identity() +
 #'   facet_grid(linetype ~ .) +
 #'   theme_void(20)
-scale_linetype <- function(name = waiver(), ..., na.value = "blank") {
+scale_linetype <- function(name = waiver(), ..., na.value = NA) {
   discrete_scale(
     "linetype", name = name,
     palette = pal_linetype(),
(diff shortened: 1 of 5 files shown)


yutannihilation:
Looks good! I think this should be documented in `NEWS.md` as a (probably very minor) breaking change. IIUC, the following code was valid at v3.5.1. I think there are no or very little users who actually specify this, but just in case.

```r
p + scale_linetype_manual(values = 1:5, na.value = "dotted")
```

yutannihilation:
> if there are actual `NA`s to map

Aw, got it. Sorry, I didn't notice this difference and tried with a data without actual `NA`. Thanks for the explanation, that makes sense!

yutannihilation (approved):

teunbrand:
Thanks for the review!

PR #6098, opened 2024-09-11
Fix regression from #6027

teunbrand:
This PR fixes a regression introduced by #6027, that is detectable as:

https://github.com/tidyverse/ggplot2/blob/d31c051bb87cd5d5f02e8d19b5c002ad2a129d6c/revdep/problems.md?plain=1#L18570-L18572

The regression occurs when `expand = 0` or `expand = 1` is set in the coord, and this PR makes this valid input again.


thomasp85 (approved):
LGTM - maybe worth a unit test

PR #6099, opened 2024-09-12
Stats use `default_aes` prior to computation

### LINKED ISSUE #3860: default_aes not used in compute_*() methods
While writing about extensions I noticed that the content of `default_aes` in a stat is not added to the data before it is passed through the `compute_*()` mill. This effectively renders it useless for anything but defining mappings for the geom.

The current code reflects this as it has all sorts of code around extracting non-required aesthetics from the data, e.g. in StatBinHex

https://github.com/tidyverse/ggplot2/blob/1223de2d6841be4276b5b398aff747672797c1b9/R/stat-binhex.r#L45-L65

we set a default `weight`, but still has to guard it with `data$weight %||% rep(1L, nrow(data))` when using it in `compute_group()`

I cannot see any meaningful reason why this is so, and the fix seems obvious. Have I missed an underlying reason @hadley?

teunbrand:
This PR aims to fix #3860.

Briefly, it makes a `Stat$use_defaults()` method that evaluates `Stat$default_aes` prior to `Stat$setup_data()`.
This is convenient for filling in optional aesthetics, like `weight`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Default aesthetics in stats are now evaluated prior to computation 
+  (@teunbrand, #3860).
 * The `summary()` method for ggplots is now more terse about facets 
   (@teunbrand, #5989).
 * `guide_bins()`, `guide_colourbar()` and `guide_coloursteps()` gain an `angle`
--- a/R/layer.R
+++ b/R/layer.R
@@ -58,8 +58,8 @@
 #'   `NA`, the default, includes if any aesthetics are mapped.
 #'   `FALSE` never includes, and `TRUE` always includes.
 #'   It can also be a named logical vector to finely select the aesthetics to
-#'   display. To include legend keys for all levels, even 
-#'   when no data exists, use `TRUE`.  If `NA`, all levels are shown in legend, 
+#'   display. To include legend keys for all levels, even
+#'   when no data exists, use `TRUE`.  If `NA`, all levels are shown in legend,
 #'   but unobserved levels are omitted.
 #' @param inherit.aes If `FALSE`, overrides the default aesthetics,
 #'   rather than combining with them. This is most useful for helper functions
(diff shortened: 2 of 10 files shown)


teunbrand:
Also came across this:

https://github.com/tidyverse/ggplot2/blob/d31c051bb87cd5d5f02e8d19b5c002ad2a129d6c/R/stat-density-2d.R#L124

Where it seems weird to me that the stat declares aesthetics for the geom.

thomasp85:
I _think_ the contract is that `default_aes` in a Stat means "the default aesthetics to pass on to the geom if not provided by the user. So it is not exactly the same as `default_aes` in Geom

teunbrand:
I'm not sure that the contract as you describe it is honoured currently in the CRAN version. Note that the lines are blue, not red.

``` r
library(ggplot2)
packageVersion("ggplot2")
#> [1] '3.5.1'

update_stat_defaults(StatDensity2d, aes(colour = "red"))
StatDensity2d$default_aes
#> Aesthetic mapping: 
#> * `colour` -> "red"
#> * `size`   -> 0.5

ggplot(faithful, aes(eruptions, waiting)) +
  stat_density_2d()
```



With this PR, the updated aesthetic is enforced:

``` r
#> ℹ Loading ggplot2

update_stat_defaults(StatDensity2d, aes(colour = "red"))

ggplot(faithful, aes(eruptions, waiting)) +
  stat_density_2d()
```



<sup>Created on 2024-12-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thomasp85:
My guess is that it doesn't work because the geom "wins". This is also as it should be I think. The stats should not define "graphical" defaults such as colour and the defaults aesthetics for stats are primarily there for aesthetics such as width etc

This is also why they can't be resolved prior to computations as the defaults can rely on values computed by the stat

teunbrand:
Closing this per discussion that we don't really know what the desired outcome is here.

Issue #3345, opened 2019-05-30
Grouped boxplots that don't drop levels


ldecicco-USGS:
How do I  (or...can I?) create a grouped boxplot in ggplot2 that does not drop the levels?

Here's an example:

```r
df <- data.frame(x = rep(letters[1:2],6),
                 y = sample(100, size = 12),
                 type = rep(c("x","y","z"),4))

df2 <- df[!(df$x == "b" & df$type == "z"),]
df2$type <- factor(df2$type, levels = c("x","y","z"))

ggplot(data = df2) +
  geom_boxplot(aes(x=x,y=y, fill=type)) +
  scale_fill_discrete(drop=FALSE)

```

And what I get is:
![image](https://user-images.githubusercontent.com/1105215/58665116-c8dc7400-82f5-11e9-8e14-dc98c8d58dce.png)

What I would like is for "b" to have a 3rd empty grouping for type "z". I have tried adding an empty row like this:

```r
df3 <- rbind(df2, 
             data.frame(x = "b",
                        y = NA,
                        type = "z"))

df3$type <- factor(df3$type, levels = c("x","y","z"))

ggplot(data = df3) +
  geom_boxplot(aes(x=x,y=y, fill=type)) +
  scale_fill_discrete(drop=FALSE)
```

But I get the same result. 

There's a solution from [2013](https://stackoverflow.com/questions/15367762/include-space-for-missing-factor-level-used-in-fill-aesthetics-in-geom-boxplot?noredirect=1&lq=1) involving fake data and changing limits...but I was hoping to see if anything had improved since then that I've missed:

```r
df4 <- rbind(df2, 
             data.frame(x = "b",
                        y = 1000,
                        type = "z"))

df4$type <- factor(df4$type, levels = c("x","y","z"))

ggplot(data = df4) +
  geom_boxplot(aes(x=x,y=y, fill=type)) +
  scale_fill_discrete(drop=FALSE) +
  coord_cartesian(ylim = range(df2$y))
```
This is good, but seems hacky:
![image](https://user-images.githubusercontent.com/1105215/58665771-3fc63c80-82f7-11e9-9460-0856a65845e6.png)

paleolimbot:
I think you're looking for `position_dodge2(preserve = "single")`:

``` r
library(ggplot2)
df <- data.frame(x = rep(letters[1:2],6),
                 y = sample(100, size = 12),
                 type = rep(c("x","y","z"),4))

[...]

PR #6100, opened 2024-09-12
Aesthetics for position adjustments

### LINKED ISSUE #3022: text is misplaced with position_dodge()
In the example below, I would expect all of the text labels to be positioned perfectly on top of the data points. Instead, some of the text labels are not positioned correctly.

I think the issue is due to `position_dodge()`. I'm not sure exactly where to look to find the relevant code.

In the last example, I use ggrepel to help illustrate the problem more clearly. You can see the blue labels `34` and `290` are not pointing to the correct positions. It seems like they're pointing to the "undodged" positions instead of the "dodged" positions.

This issue was originally reported by @raviselker in ggrepel issues: https://github.com/slowkow/ggrepel/issues/122

``` r
library(tidyverse)
library(ggrepel)
# remotes::install_github("thomasp85/patchwork)
library(patchwork)

set.seed(1337)

df <- tibble(
  x = rnorm(500),
  g1 = factor(sample(c("A", "B"), 500, replace = TRUE)),
  g2 = factor(sample(c("A", "B"), 500, replace = TRUE)),
  rownames = 1:500
)

is_outlier <- function(x) {
  return(x < quantile(x, 0.25) - 1.5 * IQR(x) | x > quantile(x, 0.75) + 1.5 * IQR(x))
}

df_outliers <- df %>% group_by(g1, g2) %>% mutate(outlier = is_outlier(x))

p1 <- ggplot(df_outliers, aes(x = g1, y = x, fill = g2)) +
  geom_boxplot(width = 0.3, position = position_dodge(0.5))

p2 <- p1 +
  geom_text(
    data = . %>% filter(outlier),
    mapping = aes(label = rownames),
    position = position_dodge(0.5)
  )

p1 + p2
```

``` r

ggplot(df_outliers, aes(x = g1, y = x, fill = g2)) +
  geom_boxplot(width = 0.3, position = position_dodge(0.5)) +
  ggrepel::geom_label_repel(
    min.segment.length = 0,
    data = . %>% filter(outlier),
    mapping = aes(label = rownames),
    position = position_dodge(0.5)
  )
```

<sup>Created on 2018-12-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### LINKED ISSUE #3026: Consistency of handling a wrong length of parameter
During reviewing #3024, I'm a bit confused to see multiple values are handled differently. I'm yet to find which behaviour is the supposed one.

``` r
library(ggplot2)

df <- data.frame(
  x = c(1,3,2,5),
  y = c("a","c","d","c")
)
```

`nudge_x` and `nudge_y` accept a different length of parameter w/ a warning.

``` r
p <- ggplot(df, aes(x, y)) +
  geom_point() +
  geom_label(aes(label = y),
             nudge_y = c(-0.1, 0.1, -0.1))

invisible(ggplot_build(p))
#> Warning in y + params$y: longer object length is not a multiple of shorter
#> object length
```

If the parameter is an aesthetic variable, it raises an error.

``` r
p <- ggplot(df, aes(x, y)) +
  geom_point() +
  geom_label(aes(label = y),
             colour = c("red", "red", "green"))
invisible(ggplot_build(p))
#> Error: Aesthetics must be either length 1 or the same as the data (4): colour
```

Some parameters just ignore multiple values (this seems because `gpar()` does so).

``` r
ggplot(df, aes(x, y)) +
  geom_point() +
  geom_label(aes(label = y),
             label.size = c(1, 10, 100))
```

<sup>Created on 2018-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### LINKED ISSUE #5445: Feature Request: nudge_x and nudge_y as a true aesthetic
Sometimes I want more granular control over the position of specific text labels for `geom_text` and `geom_text_repel`, and I may want that to depend on some other aspect of the data (overlapping points, points that are above a certain threshold, specific problematic points that I know by name from previous versions, etc) using `mutate()`.

Since `nudge_x` and `nudge_y` *can* take vectors that are the same length as the data, it seems already acknowledged that people may want to sometimes have a different nudge value for each point, which could easily be a column of nudge values in the data frame or a function thereof, so I'm not sure why it works differently from other aesthetics which can be specified in `aes()` with a column name quosure or outside of `aes()` for all points.

Could we make `nudge_x` and `nudge_y` work like the `x` and `y` aesthetics themselves?

teunbrand:
This PR aims to fix #3026, fix #5445 and fix #3022.

Briefly, it adds the ability for position adjustments to declare their own aesthetics. It is currently only implemented for `position_nudge()`.

On any layer where `position = "nudge"`, one can use `nudge_x` and `nudge_y` either as a mapped aesthetic inside `aes()`, or static aesthetic passed through `layer(params)`.

A few examples:

``` r
#> ℹ Loading ggplot2

df <- data.frame(x = 1:2, y = 1:2, z = c(0.5, 1))

ggplot(df, aes(x, y)) +
  geom_point(position = "nudge", nudge_x = 0.5)
```

``` r

ggplot(df, aes(x, y, nudge_y = z)) +
  geom_point(position = "nudge")
```

<sup>Created on 2024-09-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


clauswilke:
@teunbrand As you're working on this, I'd suggest you also look at the long history of issues with misplaced boxplots or violins when categories are missing. I realized long ago that the only way to truly fix this is to somehow have position adjustments be aware of aesthetics. Not sure whether your current work can address this, but conceptually it seems related.

See for example the long discussion here: #3022, and various other issues quoted therein.

Also, here is a reprex that highlights the issue. There is no way I am aware of that can properly format and place boxplots when categories are missing.

``` r
library(ggplot2)

# boxes for water polo, netball, and gymnastics have the wrong size
ggplot(ggridges::Aus_athletes, aes(height, sport, fill = sex)) +
  geom_boxplot()
```



``` r

# boxes for netball and gymnastics are in the wrong location
ggplot(ggridges::Aus_athletes, aes(height, sport, fill = sex)) +
  geom_boxplot(position = position_dodge(preserve = "single"))
```



``` r

# boxes for water polo, netball, and gymnastics are in the wrong location
ggplot(ggridges::Aus_athletes, aes(height, sport, fill = sex)) +
  geom_boxplot(position = position_dodge2(preserve = "single"))
```



<sup>Created on 2024-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup> 

Let me know if you think I should open a new issue for this. I don't think we have one currently, but am not entirely certain.

teunbrand:
Hi @clauswilke, I think that the issue you're highlight is fundamentally the same as #3345. I have pondered them, but haven't been able to come up with a good solution for them (yet). In any case, that issue will not be fixed by this PR.

clauswilke:
Yes, agreed, it's the same as #3345.

teunbrand:
On second thought, this PR might be able to fix it if we have something like an `order` aesthetic for `position_dodge()`.

teunbrand:
I've added an `order` aesthetic to `position_dodge()` to fix #3022 and #3345.

As an example without `order`, consider the following case where boxes have partially arbitrary order within their `x` position:

``` r
#> ℹ Loading ggplot2

set.seed(0)
df <- data.frame(
  x = rep(rep(LETTERS[1:7], c(3, 2, 2, 2, 1, 1, 1)), each = 10),
  y = rnorm(120),
  group = rep(c("X", "Y", "Z")[c(1:3, 1:2, 1, 3, 2:3, 1:3)], each = 10)
)

ggplot(df, aes(x, y, fill = group)) +
  geom_boxplot(position = position_dodge(preserve = "single"))
```



Now *with* the `order` aesthetic, the boxes are always in the same order regardless of their `x` position.

``` r
last_plot() + aes(order = group)
```



<sup>Created on 2024-09-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
There isn't anything I'd like to add at this moment, so I'm marking this as ready for review

R/position-.R
```diff
@@ -66,6 +68,36 @@ Position <- ggproto("Position",
...
+      return(data)
+    }
+
+    new <- compact(lapply(defaults, eval_tidy, data = data))
+    new[names(params)] <- params
+    check_aesthetics(new, nrow(data))
+
+    data[names(new)] <- new
+    data
+
```
thomasp85, line 100: There are some restrictions here that don't apply to the classic aesthetics, right? You can't use `after_scale()` for instance. Is that somehow taken care of?

R/position-dodge.R
```diff
@@ -104,7 +106,10 @@ PositionDodge <- ggproto("PositionDodge", Position,
   preserve = "total",
   orientation = "x",
   reverse = NULL,
+  default_aes = aes(order = NULL),
```
thomasp85, line 109: I can be a bit worried that `order` is so general that it conflicts with aesthetics from extension packages. Have you looked into that?

teunbrand: It is hard to search for, but I can confidently tell you that no package on CRAN has a `scale_order_*()` function. I've personally never seen an extension package use an `order` aesthetic.

R/position-.R
teunbrand: Any surviving modifiers would be ignored. `after_stat()` is already evaluated, so that isn't really a concern. `after_scale()` doesn't really apply as position adjustments are computed before scales are applied. It will be interpreted as there is no such aesthetic, therefore the default value is substituted.

I'm not sure whether this was clear, but this does not replace `Geom$use_defaults()`, it just evaluates the `Position$default_aes` really.

thomasp85 (approved):
LGTM

PR #6101, opened 2024-09-12
POC: Alternative way of determining parameters

### LINKED ISSUE #1516: Better way of handling parameters
For example, when `setup_data()` (as in `GeomTile`) uses custom params

teunbrand:
This PR explores to fix #1516.

It follows the approach proposed in https://github.com/tidyverse/ggplot2/issues/1516#issuecomment-1405478064.
I've numbered the comments in the code below so it'd be easier to follow.

### DIFF
--- a/R/geom-.R
+++ b/R/geom-.R
@@ -63,6 +63,8 @@ Geom <- ggproto("Geom",
 
   default_aes = aes(),
 
+  default_params = NULL,
+
   draw_key = draw_key_point,
 
   handle_na = function(self, data, params) {
@@ -79,7 +81,7 @@ Geom <- ggproto("Geom",
     }
 
     # Trim off extra parameters
-    params <- params[intersect(names(params), self$parameters())]
+    params <- filter_args(params, self$draw_panel)
 
     if (nlevels(as.factor(data$PANEL)) > 1L) {
       data_panels <- split(data, data$PANEL)
@@ -96,8 +98,9 @@ Geom <- ggproto("Geom",
 
   draw_panel = function(self, data, panel_params, coord, ...) {
     groups <- split(data, factor(data$group))
+    params <- filter_args(list2(...), self$draw_group)
     grobs <- lapply(groups, function(group) {
-      self$draw_group(group, panel_params, coord, ...)
+      inject(self$draw_group(group, panel_params = panel_params, coord = coord, !!!params))
     })
 
     ggname(snake_class(self), gTree(
@@ -208,6 +211,9 @@ Geom <- ggproto("Geom",
   extra_params = c("na.rm"),
 
   parameters = function(self, extra = FALSE) {
+    if (!is.null(self$default_params)) {
+      return(names(self$default_params))
+    }
     # Look first in draw_panel. If it contains ... then look in draw groups
     panel_args <- names(ggproto_formals(self$draw_panel))
     group_args <- names(ggproto_formals(self$draw_group))
--- a/R/geom-boxplot.R
+++ b/R/geom-boxplot.R
@@ -178,7 +178,13 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
 
   # need to declare `width` here in case this geom is used with a stat that
   # doesn't have a `width` parameter (e.g., `stat_identity`).
-  extra_params = c("na.rm", "width", "orientation", "outliers"),
+  default_params = list(
+    na.rm = FALSE, width = NULL, orientation = NA, outliers = TRUE,
+    lineend = "butt", linejoin = "mitre", fatten = 2, outlier.colour = NULL,
+    outlier.fill = NULL, outlier.shape = NULL, outlier.size = NULL,
+    outlier.stroke = 0.5, outlier.alpha = NULL, notch = FALSE, notchwidth = 0.5,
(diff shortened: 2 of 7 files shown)


R/geom-.R
```diff
@@ -63,6 +63,8 @@ Geom <- ggproto("Geom",
 
   default_aes = aes(),
 
+  default_params = NULL,
+
```
teunbrand, line 67: 1.) The Geom and Stat classes have a new field that can hold a named list of default parameters.

R/geom-.R
```diff
@@ -208,6 +211,9 @@ Geom <- ggproto("Geom",
   extra_params = c("na.rm"),
 
   parameters = function(self, extra = FALSE) {
+    if (!is.null(self$default_params)) {
+      return(names(self$default_params))
+    }
```
teunbrand, line 216: 2) We simply return the name of the default parameter list for checking `layer()` input. For backward compatibility, if this list isn't present, we resort to the old way of determining input.

R/layer.R
```diff
@@ -360,8 +360,8 @@ Layer <- ggproto("Layer", NULL,
   compute_statistic = function(self, data, layout) {
     if (empty(data))
       return(data_frame0())
-
-    self$computed_stat_params <- self$stat$setup_params(data, self$stat_params)
+    params <- defaults(self$stat_params, self$stat$default_params)
+    self$computed_stat_params <- self$stat$setup_params(data, params)
```
teunbrand, line 364: 3. Before we add computed components to the parameters, we initialise the parameters from the new field

R/layer.R
```diff
@@ -430,7 +430,8 @@ Layer <- ggproto("Layer", NULL,
       c(names(data), names(self$aes_params)),
       snake_class(self$geom)
     )
-    self$computed_geom_params <- self$geom$setup_params(data, c(self$geom_params, self$aes_params))
+    params <- defaults(c(self$geom_params, self$aes_params), self$geom$default_params)
+    self$computed_geom_params <- self$geom$setup_params(data, params)
```
teunbrand, line 434: 3. Before we add computed components to the parameters, we initialise the parameters from the new field

R/geom-.R
```diff
@@ -79,7 +81,7 @@ Geom <- ggproto("Geom",
     }
 
     # Trim off extra parameters
-    params <- params[intersect(names(params), self$parameters())]
+    params <- filter_args(params, self$draw_panel)
```
teunbrand, line 84: 4. We then filter the parameters on the go in the `compute/draw_layer()` and `compute/draw_panel()` methods

R/stat-.R
```diff
@@ -73,6 +73,8 @@ Stat <- ggproto("Stat",
 
   optional_aes = character(),
 
+  default_params = NULL,
+
```
teunbrand, line 77: 5. We take the same approach with Stats

R/stat-summary.R
```diff
@@ -181,7 +181,11 @@ stat_summary <- function(mapping = NULL, data = NULL,
 StatSummary <- ggproto("StatSummary", Stat,
   required_aes = c("x", "y"),
 
-  extra_params = c("na.rm", "orientation", "fun.data", "fun.max", "fun.min", "fun.args"),
+  default_params = list(
+    na.rm = FALSE, orientation = NA,
+    fun.data = NULL, fun = NULL, fun.max = NULL, fun.min = NULL,
+    fun.args = list()
+  ),
```
teunbrand, line 188: 6. As an example stat, here is how we could implement the default parameters for `StatSummary`

R/geom-boxplot.R
```diff
@@ -178,7 +178,13 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
...
   # need to declare `width` here in case this geom is used with a stat that
   # doesn't have a `width` parameter (e.g., `stat_identity`).
-  extra_params = c("na.rm", "width", "orientation", "outliers"),
+  default_params = list(
+    na.rm = FALSE, width = NULL, orientation = NA, outliers = TRUE,
+    lineend = "butt", linejoin = "mitre", fatten = 2, outlier.colour = NULL,
+    outlier.fill = NULL, outlier.shape = NULL, outlier.size = NULL,
+    outlier.stroke = 0.5, outlier.alpha = NULL, notch = FALSE, notchwidth = 0.5,
+    staplewidth = 0, varwidth = FALSE, flipped_aes = FALSE
+  ),
```
teunbrand, line 187: 7. As a geom example, here is how we could implement the default parameter list for `GeomBoxplot`

R/geom-violin.R
```diff
@@ -183,15 +184,15 @@ GeomViolin <- ggproto("GeomViolin", Geom,
       quantile_grob <- if (nrow(both) == 0) {
         zeroGrob()
       } else {
-        GeomPath$draw_panel(both, ...)
+        inject(GeomPath$draw_panel(both, !!!params))
       }
 
       ggname("geom_violin", grobTree(
-        GeomPolygon$draw_panel(newdata, ...),
+        inject(GeomPolygon$draw_panel(newdata, !!!params)),
```
teunbrand, line 191: 8. Here is a downside of the method: we cannot simply pass dots here as GeomViolin recieves more paremters than GeomPolygon needs.

PR #6108, opened 2024-09-16
Un-namespace `stage()` calls

### LINKED ISSUE #6104: "stage()" does not work properly when prefixed with "ggplot2::"
When I use a function from ggplot2 in my own R package, I need to prefix it with "ggplot2::". I found that, in this case, "stage()" does not work properly!

This is the original example in ggplot2. I just changed "stage" to "ggplot2::stage".

```r
  geom_violin() +
  stat_summary(
    aes(
      y = ggplot2::stage(displ, after_stat = 8),
      label = after_stat(paste(mean, "±", sd))
    ),
    geom = "text",
    fun.data = ~ round(data.frame(mean = mean(.x), sd = sd(.x)), 2)
  )
```

Throws the following error：
```r
Error in `stat_summary()`:
! Problem while mapping stat to aesthetics.
ℹ Error occurred in the 2nd layer.
Caused by error:
! object 'displ' not found
```

teunbrand:
This PR also aims to fix #6104. It is mutually exclusive with #6107.

Briefly, this approach surgically tries to intercept `ggplot2::stage()` calls and change them to `stage()` calls, so that the function masking magic continues to work.

While more surgical than #6107 I feel this is more fussy.

### DIFF
--- a/R/aes.R
+++ b/R/aes.R
@@ -208,6 +208,9 @@ standardise_aes_symbols <- function(x) {
 
   # Don't walk through function heads
   x[-1] <- lapply(x[-1], standardise_aes_symbols)
+  if (is_call(x, "stage", ns = "ggplot2")) {
+    x[[1]] <- call("stage")[[1]]
+  }
 
   x
 }


teunbrand:
Closing in favour of #6110

PR #6110, opened 2024-09-17
Mask `stage()` at the expression level

### LINKED ISSUE #6104: "stage()" does not work properly when prefixed with "ggplot2::"

teunbrand:
This PR aims to fix #6104.
It is mutually exclusive with #6108 and #6107.

Briefly, this PR *always* masks the syntactic functions `after_stat()`, `after_scale()` and `stage()`.
That way, you needn't prefix it with `ggplot2::` if you don't have the namespace attached.

``` r
# Note: this PR is locally installed for this example
ggplot2::ggplot(ggplot2::mpg, ggplot2::aes(displ, hwy)) +
  ggplot2::geom_point(
    ggplot2::aes(fill = stage(drv, after_scale = scales::alpha(fill, 0.3))),
    shape = 21
  )
```

And users cannot override these functions, at least when used inside `aes()`.

``` r
library(ggplot2)

stage <- function(...) stop("error")

  geom_point(
    aes(fill = stage(drv, after_scale = alpha(fill, 0.3))),
    shape = 21
  )
```

<sup>Created on 2024-09-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


teunbrand:
Alright as a hail mary, I've changed from wrestling with data masks to changing the calls in the unevaluated aesthetics.
Thus, `stage()` now acts as any other function in that `stage()` doesn't work unless ggplot2 is loaded and `ggplot2::stage()` behaves as expected.
The only slightly insidious behaviour is that if another package, let's say {foobar} defines `foobar::stage()`, then this gets supplanted by ggplot2's functions.

``` r
# Note: this PR is locally installed for this example
ggplot2::ggplot(ggplot2::mpg, ggplot2::aes(displ, hwy)) +
  ggplot2::geom_point(
    ggplot2::aes(fill = ggplot2::stage(drv, after_scale = scales::alpha(fill, 0.3))),
    shape = 21
  )
```



``` r
# We cannot use naked `stage()` when we haven't loaded ggplot2
ggplot2::ggplot(ggplot2::mpg, ggplot2::aes(displ, hwy)) +
  ggplot2::geom_point(
    ggplot2::aes(fill = stage(drv, after_scale = scales::alpha(fill, 0.3))),
    shape = 21
  )
#> Error in `ggplot2::geom_point()`:
#> ! Problem while computing aesthetics.
#> Caused by error in `stage()`:
#> ! could not find function "stage"
```

<sup>Created on 2024-09-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thomasp85 (approved):
LGTM. I don't suppose there is an elegant way to test this?

teunbrand:
The masking mechanic is testable, using `ggplot2::stage()` is testable, but the 'if ggplot2 is not loaded' condition is hard to test.

PR #6114, opened 2024-09-17
Check continuous scale limits

### LINKED ISSUE #6092: Checking limits argument in scale_<x|y>_continuous
I added `limits = c(1, 2, 3)` to a scale_* and I didn't realize it (I though I used `breaks`). But was surprised by the resulting plot: it only uses the two first values and discards the rest (Totally logic and consistent).

But the documentation says: 

> A numeric vector of length two providing limits of the scale. Use NA to refer to the existing minimum or maximum

I would like to receive an error if the vector used in limits for scales is larger than 2. This would help to fail with an informative error. I got a warning of `Removed <x> rows containing missing values or values outside the scale range`:

```r
library("ggplot2")
  geom_point() +
  scale_x_continuous(limits = c(1, 5, 10))
## Warning message:
## Removed 36 rows containing missing values or values outside the scale range (`geom_point()`). 
```
Expected:

```r
library("ggplot2")
  geom_point() +
  scale_x_continuous(limits = c(1, 5, 10))
## Error in scale_x_continuous: 
## Provided limits for three dimensions when only two are used.  
```

This might happen to other limits from other scale_* functions or in other arguments

teunbrand:
This PR aims to fix #6092, but has escalated a bit.

I'd say the main thing is that this PR adds and uses a `check_length()` function that checks whether the length of a vector is of a particular length or within a range of lengths.
This is then used to check scale limits and replace ad hoc length checks.

Small demo:

``` r
#> ℹ Loading ggplot2
scale_x_continuous(limits = c("A", "B"))
#> ! `limits` must be a <numeric> vector, not a character vector.
scale_x_continuous(limits = c(1:3))
#> ! `limits` must be a vector of length 2, not length 3.
```

<sup>Created on 2024-09-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/stat-qq-line.R
```diff
@@ -71,9 +71,7 @@ StatQqLine <- ggproto("StatQqLine", Stat,
 
     theoretical <- inject(distribution(p = quantiles, !!!dparams))
 
-    if (length(line.p) != 2) {
-      cli::cli_abort("Cannot fit line quantiles {line.p}. {.arg line.p} must have length 2.")
-    }
+    check_length(line.p, 2L)
```
thomasp85, line 74: It seems this message is now less descriptive

thomasp85 (approved):
Apart from small comment LGTM

R/stat-qq-line.R
teunbrand: Yeah somewhat. Do you prefer to revert, ignore this, or wrap in `try_fetch()` block and appending the offending quantile line?

PR #6116, opened 2024-09-19
Partial margins

### LINKED ISSUE #6115: Consider handling/setting `margin()`'s default values based on the theme being used
Currently, the default margin values of `margin()` are set to 0. This is not ideal because if a user wants to change the margin only on one side of a plot, it forces the user to 1) find out what the default margins of the theme being used are and 2) redefine the default margin values on each other side of a plot.

It would be more convenient if `margin()` could handle the default values internally (or via a function passed as the default arguments) based on the theme being used so that e.g. one could write:

``` r
# ... +
  theme_classic() +
  theme(plot.margin = margin(r = 9.5))
```

Instead of:

``` r
# ... +
  theme_classic() +
  theme(plot.margin = margin(5.5, 9.5, 5.5, 5.5))
```

teunbrand:
This PR aims to fix #6115.

Briefly it allows NA-units in `margin()` that get back-filled from parent elements. The function `part_margin()` is a `margin()` wrapper that by default has `NA`-units.

A demonstration. Note no margins except bottom:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  theme(plot.background = element_rect(linetype = 2, colour = "black", fill = NA))

p + 
  theme(plot.margin = margin(b = 2, unit = "cm")) +
  ggtitle("`margin()`")
```

Note natural margins are preserved, except bottom:

``` r
p + 
  theme(plot.margin = part_margin(b = 2, unit = "cm")) +
  ggtitle("`part_margin()`")
```

<sup>Created on 2024-09-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


davidhodge931:
Great work, @teunbrand 

I think it might be helpful for users, if this was named with a margin prefix? This would organise the two functions, so that auto-complete would allow users to easily use the normal one - or switch to the partial one. 

So `margin_part` (or `margin_partial`)

teunbrand:
Thanks David, I like this suggestion :)

arnaudgallou:
Thanks for considering my FR. Looking forward to seeing this implemented.
Just wanted to let you know the NEWS file references the wrong issue ID (6155 when it should be 6115).

thomasp85 (commented):
While we are here, maybe add a `margin_auto()` that behaves like CSS, ie. `r` and `b` inherits from `t`, and `l` inherits from `r`

teunbrand:
> margin_auto() that behaves like CSS

I like this idea

thomasp85 (approved):
LGTM

Issue #5962, opened 2024-06-28
Deprecation for ggproto classes


teunbrand:
This now has come up in #3798 and #5961 and the gist of it is that we have a procedure for managing lifecycles of functions, but we don't have an equivalent for ggproto classes.

We don't use `GeomCol` and maybe soon `GeomErrorbarh` anymore in ggplot2, but they are still in use by extensions. We can see for example that [`GeomCol` is used](https://github.com/search?q=org%3Acran+GeomCol&type=code) in {ichimoku}, {entropart}, {ggiraph}, {ggside}. or that [`GeomErrorbarh` is used](https://github.com/search?q=org%3Acran+GeomErrorbarh&type=code) in {errors} and {ggiraph}. Having a deprecation mechanism in place allows us to discourage the use of these classes while not breaking other packages.

Generally speaking, I think there are 3 ways in which ggproto classes are used in other packages::

1. They may be subclassed in an extension class.
2. They may be provided as arguments in e.g. `layer(geom = GeomCol, ...)`.
3. They're used for accessing fields, .e.g. `GeomCol$default_aes`.

The question of this issue is whether we can find a satisfactory way to deprecate these old classes. I think it would be relatively straightforward to throw deprecation messages in cases (1) and (2) via either `ggproto()` or `check_subclass()`. I'm not sure (3) can be caught in an easy fashion.

teunbrand:
Just to jot down some notes regarding topics discussed off-github;

We should ensure symmetry at the `check_subclass()` (unsed in `layer()`) level, so that it can find classes based on constructors even if the class itself is absent. This would ensure that `layer(geom = "col")` will find `GeomBar` and `layer(geom = "errorbarh")` will find `GeomErrorbar`.

teunbrand:
Considering this closed by #6119 for now.

PR #6119, opened 2024-09-20
Fallback for `validate_subclass()`

teunbrand:
This PR aims to progress #5962 and #3798.

Taking the deprecation of GeomCol (#3798) as an example, we want users to be able to use `layer(geom = "col", ...)` even after the class is deprecated. This PR ensures that `layer(geom = "col")` finds the `geom_col()` constructor and returns the `geom` field from the resulting layer (even if we remove the GeomCol class).


R/layer.R
```diff
@@ -475,19 +475,42 @@ check_subclass <- function(x, subclass,
...
+  obj <- find_global(name, env = env)
+  if (inherits(obj, subclass)) {
+    return(obj)
+  }
+
+  # Try retrieving class via constructors
+  name <- snakeize(name)
+  obj <- find_global(name, env = env, mode = "function")
+  if (is.function(obj)) {
+    obj <- obj()
```
thomasp85, line 495: Maybe wrap in `try_fetch()` and rethrow the error with some additional context. We might end in situations where the constructor can't be called without arguments even though it generally should work

tests/testthat/_snaps/layer.md
```diff
@@ -25,7 +25,13 @@
 
 ---
 
-    `x` must be either a string or a <geom> object, not an environment.
+    `environment()` must be either a string or a <geom> object, not an environment.
+
```
thomasp85, line 29: This seems like a regression in the error message?

thomasp85 (approved):
LGTM apart from the small comment

tests/testthat/_snaps/layer.md
teunbrand: I'd argue the message has become more informative. The message comes from this test:

```r
expect_snapshot_error(validate_subclass(environment(), "geom"))
```

And we now use `caller_arg(x)` for messaging.

thomasp85: Oh - Ignore my pre-coffee eyes

PR #6127, opened 2024-10-01
A big lint

teunbrand:
This PR makes a lot of small tweaks in the code based on a select few linters from the {lintr} package.

I didn't blindly obey all the linters; in some cases, the linter has been selectively ignored. 
For example I didn't reindent/rewhitespace the `.element_tree` construction because the whitespace facilitates vertical alignment.
In other cases, I've chosen not to include some linters, like the cyclomatic complexity one, as ggproto classes were being counted as 1 big function.

I tried to group similar changes in the same commit so it is easy to undo some changes globally.


thomasp85 (commented):
Have you aligned this with the tidy style guide? I'm not sure lintr is fully reflective of that?

teunbrand:
No but I think lintr and tidyverse style guide share a lot of common ground.
I'm happy to either (1) remove commits from this PR that introduce changes you think are for the worse or (2) pass code through styler

thomasp85:
My guess is that this has now diverted too much from the main branch to be practical. I don't know if you wanna start afresh or update this PR

teunbrand:
Yeah it is probably best to do something like this just before a release. I'll close this for now and make a similar one later.

PR #6130, opened 2024-10-04
Restore rather than recalculate scales in `FacetWrap` with `CoordFlip`

mikmart:
Currently `FacetWrap$draw_panels()` recalculates `layout$SCALE_X` and `layout$SCALE_Y` if used with `coord_flip()`. This PR changes the behaviour to instead restore the scales as calculated by `$compute_layout()`.

This is relevant when a subclass of `FacetWrap` has a customised layout, but could still otherwise call `FacetWrap$draw_panels()`, like in [ggragged](https://github.com/mikmart/ggragged/blob/513525d100ed636e29bebadd7f29391193246ceb/R/facet_ragged_cols.R#L38).

Fixes mikmart/ggragged#2.

### DIFF
--- a/R/facet-wrap.R
+++ b/R/facet-wrap.R
@@ -447,16 +447,8 @@ FacetWrap <- ggproto("FacetWrap", Facet,
 
   draw_panels = function(self, panels, layout, x_scales, y_scales, ranges, coord, data, theme, params) {
     if (inherits(coord, "CoordFlip")) {
-      if (params$free$x) {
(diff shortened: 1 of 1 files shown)


teunbrand (approved):
I love this simplification, I've always hated the finnickyness of facet_wrap/coord_flip interaction. I'll approve this but will wait a few days before merging so that it can appear on Thomas' radar.

teunbrand:

mikmart:
Thanks!

PR #6131, opened 2024-10-07
Add test for using `stage()` with a mapping specified for `start` only

yjunechoe:
This PR adds tests to codify the behavior of `aes(stage(x))` and its variants working equivalently to `aes(x)`.

I had been under the impression that `stage(x)` will remain invalid since #4890 was closed without a follow-up (and with a general agreement that this is an odd thing to do), but that issue seems to have been resolved a while back as part of a larger PR in #5409. I assume that the currently working behavior of `stage(x)` is intended and desirable, and I would personally like for this to be guaranteed as a stable behavior (or at least, for there to be tests to detect its unintended regression).

Hence, the PR adds three tests of equivalence between `aes(x)` and:
1. `aes(stage(x))`
2. `aes(stage(start = x))`
3. `aes(stage(start = x, after_stat = NULL, after_scale = NULL))`

I saw some prior discussion on the significance of resolving an unnamed value passed to `start`, so the test includes both (1) and (2). I added (3) just for completeness, but can be removed if this is a weird thing to test (perhaps more meaningful is testing whether (3) works when other arguments are dynamically evaluated to `NULL`, but I digress).

---

In case `stage(x)` wasn't intended and the question of its desirability is still up in the air, I want to advocate for this with a pedagogical use-case: the practical equivalence between `aes(x)` and `aes(stage(x))` makes the introduction of more advanced strategies only one-step removed and thus easier to teach/learn, e.g., `aes(fill = stage(x, after_scale = darken(fill, .5)))`

### DIFF
--- a/R/aes-evaluation.R
+++ b/R/aes-evaluation.R
@@ -32,7 +32,7 @@
 #' Below follows an overview of the three stages of evaluation and how aesthetic
 #' evaluation can be controlled.
 #'
-#' ## Stage 1: direct input
+#' ## Stage 1: direct input at the start
 #' The default is to map at the beginning, using the layer data provided by
 #' the user. If you want to map directly from the layer data you should not do
 #' anything special. This is the only stage where the original layer data can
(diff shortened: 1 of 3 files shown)


teunbrand:
Thanks for the PR, June! I agree that `stage()` working with just the one argument should be equivalent to plain evaluation.

> the practical equivalence between aes(x) and aes(stage(x)) makes the introduction of more advanced strategies only one-step removed and thus easier to teach/learn

This might be nice to include in `?aes_eval` as well, but this is totally optional.

> I'd also like to clean up this comment

This makes sense to me.

yjunechoe:
> This might be nice to include in ?aes_eval as well, but this is totally optional.

Will think a bit about this!

yjunechoe:
I added a bit to the "Complex staging" section to spell out this equivalence (of ex: `x` <-> `stage(start = x)`). Plus some unsolicited changes to naming of headers (I thought it'd be nice to have the argument param name "start" be present in the "Stage 1" header) and intro paragraph about `stage()` (which reads to me a bit abrupt, but I recognize that this is function docs not a vignette, so I can roll it back if it's getting too long)

teunbrand:
I like it :) Do you think this is good to go or do you envision more edits?

yjunechoe:
That's it! Nothing else in my mind for the scope of this PR

teunbrand (approved):
Looks great, thanks June for the contribution!

PR #6138, opened 2024-10-15
Bump deprecations

teunbrand:
This PR does not fix an issue, but I see it as regular maintenance.

Briefly, it converts `deprecate_warn0()` to `lifecycle::deprecate_stop()` for functions/arguments that were deprecated prior to ggplot 3.3.0.
It converts `deprecate_soft0()` to `deprecate_warn0()` for functions/arguments that were deprecated up to ggplot2 3.2.0 and removes examples.

I thought 5 years to be an OK horizon for deprecation warnings and I felt it was time to stop allowing things from the previous major version. Happy to make these timelines shorter though :)


thomasp85 (commented):
Let's have a revdepcheck for this before we fully commit. I agree in principle about the timespan

thomasp85:
revdepchecks have been committed

teunbrand:
So about 30/347 packages show up with at least a deprecation warning. I think we might be slightly too aggressive about erroring on the theme settings, as giving a wrong argument usually gives a warning, not an error.

teunbrand:
Per https://github.com/tidyverse/ggplot2/pull/6281#issuecomment-2618221318:

> I think we are at the point were we can start a deprecation process. We already have deprecation warnings for `default_aes` to alert extension developers and 2 years in I think we can begin to make noise for the users

Regarding `size` to `linewidth` translation

thomasp85:
Do you want another revdepcheck for this?

teunbrand:
For this specifically, not really. I think we're the right amount of agressive with these now and if we aren't: it'll show up on future revdepchecks we might run for other issues too.

thomasp85 (approved):
LGTM

Issue #5246, opened 2023-03-23
Better documentation of `xseq` for `stat_smooth()`


larmarange:
The `xseq` argument is a sort of hidden option for `stat_smooth()` allowing to select where to draw the smooth line.

Would it be relevant to mention it in the documentation and eventually to provide an example?

teunbrand:
I'm not sure if it is hidden on purpose or not, but typically such options aren't advertised to users because they're considered internal.

hadley:
It's pretty esoteric, so I think we should only bother to document it if it's really required.

hadley:
If folks need that capability, these days I'd recommend that they fit the model themselves.

EvaMaeRey:
@teunbrand gave me a heads up about this issue, so dropping in another possible (bespoke?) pedagogical xseq use, of adding fitted values for of observed values of x as points along smooth.  And also segments between these fitted values and raw data, giving residuals:

mtcars |> 
  ggplot(aes(wt, mpg)) +
  geom_point() +
  geom_smooth() + 
  stat_smooth(geom = "point",                  # draw fitted values as points
                        color = "blue",                   
                        xseq = mtcars$wt) +
  stat_smooth(geom = "segment",            # draw residuals
                         color = "darkred",       
                         xseq = mtcars$wt,
                         xend = mtcars$wt,
                         yend = mtcars$mpg)

PR #6143, opened 2024-10-17
Boilerplate functions

### LINKED ISSUE #6142: Code duplication
A lot of ggplot2 code is boilerplate code, that is, repeated section of code with little to no variation.
Think of all the `geom_*()` or `stat_*()` or `scale_*()` constructors, which are all very similar.
It seems to me that this pattern of duplicated code is susceptible to accruing tiny inconsistencies.

Probably, a lot of this code could be generated (semi-)automatically. This would:
* Reduce duplications
* Decrease inconsistencies
* Be easier to maintain
* Be beneficial for extensions too

Notably #5484 started replacing a lot of duplicated `scales_*()` code, but I don't think this is systemic enough.

teunbrand:
This PR aims to (eventually) fix #6142.

Briefly, it introduces `boilerplate()` that can take in a ggproto class and spit out a constructor.

This PR is a proof-of-concept because it probably merits some discussion whether we would like to go down this road. As such, the PR only implements `boilerplate()` for Geoms, but can be expanded in the future for Stats and Scales as well, I think.
The `boilerplate()` function is used for amenable functions, but not more complex ones. For example, `geom_text()` or `geom_boxplot()` perform input checking that we cannot currently easily fit into the boilerplate code.

A few notes about the PR:
* Because we apply this function for suitable geoms, we had to swap the position of class definition and constructor, as `boilerplate()` takes the class as input. These are the largest diffs.
* Some order of arguments are changed, as they are now handled consistently.
* Many geoms had 'hidden' parameters, i.e. parameters like `lineend` that would be accepted by the `Geom$draw_panel/group()` method, but were set as arguments to the constructor. These are now all exposed.
* In visual tests, I had to accept the changes because `draw_key_polygon()` now actually gets the line settings (lineend, linejoin, linemitre) from the geom instead of having to default these.
* I don't have a lot of experience of programatically constructing function bodies. If there is a better way, I'd gladly accept pointers.


R/boilerplates.R
```diff
@@ -0,0 +1,129 @@
...
+    mapping  = args$mapping,
+    data     = args$data,
+    stat     = args$stat %||% "identity",
+    position = args$position %||% "identity",
+    `...` = quote(expr = ),
+    !!!args[extra_args],
+    na.rm    = args$na.rm %||% FALSE,
+    show.legend = args$show.legend %||% NA,
+    inherit.aes = args$inherit.aes %||% TRUE
+  )
```
yjunechoe, line 94: Using more idiomatic language

```suggestion
  # Build function formals
  fmls <- rlang::pairlist2(
    mapping  = args$mapping,
    data     = args$data,
    stat     = args$stat %||% "identity",
    position = args$position %||% "identity",
    `...` = rlang::missing_arg(),
    !!!args[extra_args],
    na.rm    = args$na.rm %||% FALSE,
    show.legend = args$show.legend %||% NA,
    inherit.aes = args$inherit.aes %||% TRUE
  )
```

R/boilerplates.R
```diff
@@ -0,0 +1,129 @@
...
+
+  if (length(extra_args) > 0) {
+    extra_args <- paste0(
+      "\n      ", extra_args, " = ", extra_args, ",", collapse = ""
+    )
```
yjunechoe, line 99: Could be resolved with ` rlang::exprs(!!!rlang::syms(extra_args), .named = TRUE)` instead; see below

R/boilerplates.R
```diff
@@ -0,0 +1,129 @@
...
+    show.legend = show.legend,
+    inherit.aes = inherit.aes,
+    params = list2(
+      na.rm = na.rm,",
+      extra_args, "
+      ...
+    )
+  )
+  ")
+  body <- str2lang(body)
```
yjunechoe, line 118: I would resolve the trivial `arg = arg` type arguments of `layer()` first:

```r
  # Resolve layer(...) arguments
  layer_args <- rlang::exprs(
    data, mapping, stat, position, show.legend, inherit.aes,
    .named = TRUE
  )
```

And then deal with `params` separately:

```r
  # Resolve layer(params = list2(...))
  layer_params_arg <- rlang::exprs(
    na.rm, !!!rlang::syms(extra_args),
    .named = TRUE
  )
  layer_params_arg <- c(layer_params_arg, quote(...))
  layer_args$params <- rlang::call2("list2", !!!layer_params_arg, .ns = "rlang")
```

Finally, put the arguments inside `layer()` call

```r
  # Complete layer() expression
  body <- rlang::call2("layer", !!!layer_args)
```

R/boilerplates.R
```diff
@@ -0,0 +1,129 @@
...
+
+  checks <- substitute(checks)
+  if (!missing(checks)) {
+    if (is_call(checks, "{")) {
+      checks[[1]] <- NULL
+    }
+    body <- inject(quote(`{`(!!!c(checks, body))))
+  }
```
yjunechoe, line 126: The API of `check` is interesting (e.g., accepting a `{`-block for multiple expressions), though I'm not sure about the audience. IIUC, I think the kind of advanced folks reaching for this feature would rather prefer to construct then pass in their own expressions, instead of writing code to be captured; ex: `checks = rlang::exprs(a, b)` vs. `checks = {a; b}`.

yjunechoe (commented):
I'm super excited to see this develop! (esp. for Stats!)

> programatically constructing function bodies

I don't have a perfect solution (not thoroughly tested beyond the example) but I thought I'd just throw in some ideas for that, in case the PR will eventually need to do this before it's ready to merge. (I only looked at constructing the `layer()` expression - `check` feels complicated at a more fundamental level?)

R/boilerplates.R
teunbrand: Yeah you're right, the target audience is other devs, so we can expect them to give correct input :)

R/boilerplates.R
teunbrand: Reminds me of `fortunes::fortune("parse()")` but for `str2lang()`

R/boilerplates.R
teunbrand: Good advice, thanks!

R/boilerplates.R
```diff
@@ -0,0 +1,122 @@
...
+  # Prepend any checks
+  if (!missing(checks)) {
+    lang <- vapply(checks, is_call, logical(1))
+    if (!all(lang)) {
+      cli::cli_abort(
+        "{.arg checks} must be a list of calls, such as one constructed \\
+        with {.fn rlang::exprs}."
+      )
+    }
+    body <- inject(quote(`{`(!!!c(checks, body))))
```
yjunechoe, line 118: Just to standardize this line with usage of `call2()` elsehere

```suggestion
    body <- call2("{", !!!checks, body)
```

R/boilerplates.R
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  new_function(fmls, body)
```
yjunechoe, line 121:
```suggestion
  new_function(fmls, body, env = caller_env())
```

I think this is necessary to prevent `boilerplate()`-internal variables from leaking. Currently:

```r
x <- boilerplate(GeomTest, checks = rlang::exprs(cat(fixed_fmls_names)))
invisible(x())
#> mapping data stat position ... na.rm show.legend inherit.aes
```

yjunechoe (commented):
Thanks for considering my suggestions! Following up w/ just 2 more things I noticed

R/boilerplates.R
teunbrand: Yes, much cleaner!

R/boilerplates.R
teunbrand: Good point!

mjskay:
FWIW, very excited to see something like this happen! {ggdist} has its own boilerplate generation for geoms and stats if you are looking for a comparison point: https://github.com/mjskay/ggdist/blob/master/R/abstract_geom.R

Besides reducing maintenance issues I have also moved a lot of documentation into it, so now comprehensive docs of aesthetics and computed variables are generated automatically. Has made my life quite a bit easier.

teunbrand:
Thanks Matthew, it is cool to see that other people have also run into (and solved!) 'there should be hidden parameters' problem I got stuck on as well ^_^ It is interesting (in the good way) to see you opted for a ggproto approach, but I guess this might help coordinate various tasks at once. 

I also recently was pointed to @corybrunson's https://github.com/corybrunson/ordr/blob/main/build-pre/build-layers.r, which performs a similar task as well.

teunbrand:
Alright, stats are covered too now. The method is similar to `make_constructor.Geom` but fixes the `layer(stat)` instead of `layer(geom)` and leaves `geom` is a user-specified argument rather than `stat`. Per discussion, the main function is now called `make_constructor()`. 

For Stats, I sometimes had to hide variables, which I did through the `make_constructor(omit)` argument. Mostly for two reasons.

1. It was a computed parameter in `Stat$setup_params()` or the parent `Stat$compute_*()` method. These should not be exposed to users, even though they technically could pass these arguments to `...`.
2. It was a parameter that a user could use, like `stat_bin_2d(breaks)`, but was not exposed, similar to #5246.

I hide both these cases, but for (1) we might consider a mechanism for explicitly prohibiting these arguments and for (2) whether such parameters should be exposed. This is outside the scope of this PR I think.

> I also see your other PR

This doesn't really narrow it down I'm afraid😅 

> how far that goes in the direction of breaking changes

I don't plan on breaking too much here. This is mostly a convenience function, so I'm looking for a standardised solution. Instead of packing *everything* in these `make_constructor()` methods, I aim to pack ~80% of functionality in these methods and keep the current constructors with exotic logic as-is. I don't think it is worth it to code the `geom_text()` nudge/position logic into the `make_constructor()` method for example.

teunbrand:
Marking this as ready for review as we decided we don't want a `make_constructor()` for scales

R/make-constructor.R
```diff
@@ -0,0 +1,217 @@
...
+#'
+#' The `make_constructor()` functions sets up a user-facing constructor for
+#' ggproto classes. Currently, `make_constructor()` is implemented for
+#' `Geom` classes.
+#'
+#' @param x An object to setup a constructor for.
+#' @param ... Name-value pairs to use as additional arguments in the
+#'   constructor. For layers, these are passed on to [`layer(params)`][layer()].
+#' @param checks A list of calls to be evaluated before construction of the
+#'   object, such as one constructed with [`exprs()`][rlang::exprs()].
```
thomasp85, line 15: Can you add an example of this to the @examples as I don't think it is immediately obvious to most users how to use it

thomasp85 (changes requested):

R/geom-curve.R
```diff
@@ -1,81 +1,50 @@
...
-        lwd = trans$linewidth,
-        lty = trans$linetype,
-        lineend = lineend),
-      arrow = arrow
-    )
-  }
+                       curveGrob(
+                         trans$x, trans$y, trans$xend, trans$yend,
+                         default.units = "native",
+                         curvature = curvature, angle = angle, ncp = ncp,
```
thomasp85, line 34: Are these whiteline changes intentional?

teunbrand:
I'm not considering this a carrot for editioning because this is developer-facing rather than user-facing

thomasp85 (commented):
Is it intentional that `geom_polygon()` now uses a different linejoin than before?

teunbrand:
If I'm not mistaken, `geom_polygon()` has the same linejoin as before. It is `draw_key_polygon()` that now absorbs the geom's linejoin setting. Is that what you were referring to?

thomasp85:
oh yeah, sorry

thomasp85 (approved):
LGTM

PR #6145, opened 2024-10-18
Fix binned scaled not accepting function-limits if there are transformations

### LINKED ISSUE #6144: Binned scales don't work with `limits = function()`
According to the docs, `scale_x_binned()` and friends can be given a function for `limits`. However, this fails when a transform is also set.

This example is supposed to make a bar plot counting days in each month (i.e. a binned date scale), and works fine until a limits function is provided:

```R
library(ggplot2)

df <- data.frame(x = seq(as.Date("2024-01-01"), as.Date("2024-06-30"), "day"))

ggplot(df) +
  geom_bar(aes(x)) +
  scale_x_binned(
    breaks = scales::breaks_width("1 month"),
    transform = scales::transform_date(),
    # works if the following line is commented out
    limits = function(x) { x }
  )

# Error in `scale_x_binned()`:
# ! Binned scales only support continuous data.
```

Looking through the code, it seems that `ScaleBinned` itself is fine (all access goes through `ScaleBinned$get_limits()` which handles function values properly), but the `binned_scale()` constructor causes the error here:

https://github.com/tidyverse/ggplot2/blob/ddd207e926cc1c1847dc661d9a099b8ec19c4010/R/scale-.R#L321-L324

In that snippet, `limits` is the bare value passed in by the user, so if that's a function then it's forwarded as-is to the transform. The date transform says that a function object is not numeric hence the error.

Found on 3.5.1 but the relevant code is the same on main.

teunbrand:
This PR aims to fix #6144.

It simply doesn't attempt to pre-transform the limits when the limits is a function.

Reprex from issue:

``` r
#> ℹ Loading ggplot2

df <- data.frame(x = seq(as.Date("2024-01-01"), as.Date("2024-06-30"), "day"))

ggplot(df) +
  geom_bar(aes(x)) +
  scale_x_binned(
    breaks = scales::breaks_width("1 month"),
    transform = scales::transform_date(),
    limits = function(x) { x }
  )
#> Warning in scale_x_binned(breaks = scales::breaks_width("1 month"), transform = scales::transform_date(), : Ignoring `n.breaks`. Use a breaks function that supports setting number of
#> breaks.
```

<sup>Created on 2024-10-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The spurious warning the scale throws is commented upon [here](https://github.com/tidyverse/ggplot2/pull/5974#issue-2389249603) as well and better fixed in that PR than this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* Fixed bug where binned scales wouldn't simultaneously accept transformations
+  and function-limits (@teunbrand, #6144).
 * Fixed bug where the `ggplot2::`-prefix did not work with `stage()` 
   (@teunbrand, #6104).
 * New `get_labs()` function for retrieving completed plot labels 
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -319,7 +319,7 @@ binned_scale <- function(aesthetics, scale_name = deprecated(), palette, name = 
   }
 
   transform <- as.transform(transform)
-  if (!is.null(limits)) {
+  if (!is.null(limits) && !is.function(limits)) {
     limits <- transform$transform(limits)
   }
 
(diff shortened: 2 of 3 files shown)


thomasp85 (approved):
LGTM

PR #6148, opened 2024-10-18
Discrete labels from break names

### LINKED ISSUE #6089: Discrete secondary axis labels & breaks arguments work in unexpected ways
I've been testing out the new discrete secondary axis functionality in the DEV version.

Firstly, awesome work and a great advance!

I felt the discrete secondary axis labels and breaks arguments were working in unexpected ways in that:

1. default sec.axis discrete labels are numeric - should they be the actual labels instead?
2. palette moves the breaks in scale_*_discrete, but breaks does this in sec.axis - should the sec.axis operate in the same way as the discrete scale?

I'm not familiar with the internals. So there be valid reasons for this that I'm unaware of. Anyway, great work on this. It'll be a super useful feature. Sorry to bombard with issues. As always, feel free to close, if it is working the way you want it to

``` r
library(tidyverse)

data <- data.frame(x = c("A", "B", "C"), y = c("D", "E", "F"))

#sec.axis discrete labels are numeric
ggplot(data, aes(x, y)) +
  geom_point() +
  scale_x_discrete(
    sec.axis = dup_axis()
  ) +
  labs(title = "default sec.axis discrete labels are numeric")
```

``` r

#palette moves the breaks in scale_*_discrete but breaks does this in sec.axis
ggplot(data, aes(x, y)) +
  geom_point() +
  scale_x_discrete(
    palette = \(x) c(0.5, 1.5, 3.5),
    sec.axis = dup_axis(
      breaks = c(1.3, 3, 4)
    )
  ) +
  labs(title = "palette moves the breaks in scale_*_discrete but breaks does in sec.axis")
```

### LINKED ISSUE #6147: Discrete scales don't use `names(breaks)` for default-labels
As the title implies, continuous scales use `names(breaks)` as their labels, whereas discrete scales do not:

``` r
library(ggplot2)

ggplot(mpg, aes(drv, hwy)) +
  geom_boxplot() +
  scale_y_continuous(
    breaks = c(twenty = 20, thirty = 30, fourty = 40)
  ) +
  scale_x_discrete(
    breaks = c("four-wheel" = 4, forward = "f", reverse = "r")
  )
```

<sup>Created on 2024-10-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I think we should probably mirror this mechanism in discrete scales, so that the x-axis above has the labels 'four-wheel', 'forward' and 'reverse'. 

The reason continuous scales have this mechanism is because `labels = waiver()` (the default) let's the transformation's format method label the breaks. Many such formatters return names first and only format when names are absent.

``` r
scales::transform_identity()$format
#> function (x) 
#> {
#>     if (!is.null(names(x))) {
#>         return(names(x))
#>     }
#>     ret <- format(x, ..., trim = TRUE, justify = "left")
#>     ret[is.na(x)] <- NA
#>     ret
#> }
#> <bytecode: 0x000002080e0c05d8>
#> <environment: 0x000002080e09fc70>
```

<sup>Created on 2024-10-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6147 and fix #6089.

Briefly, `names(breaks)` is used as default label for discrete breaks.
We also use this mechanism to populate discrete secondary labels (#6089).

Reprex from #6147 with a secondary axis:

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(drv, hwy)) +
  geom_boxplot() +
  scale_y_continuous(
    breaks = c(twenty = 20, thirty = 30, fourty = 40)
  ) +
  scale_x_discrete(
    breaks = c("four-wheel" = 4, forward = "f", reverse = "r"),
    sec.axis = dup_axis()
  )
```

<sup>Created on 2024-10-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

Issue #5895, opened 2024-05-20
`stat_bin()` should have the area (instead of height) represent the count.


mattansb:
Histograms convert counts withing bins into _areas_.

However, in `ggplot2`, the default behavior is to convert counts to bar _heights_. 

This discrepancy is typically not noticeable, because `stat_bin()` default to equi-width bins.
However it becomes apparent when using non-equi-width bins. 

Here is an example with equi-probable bins, in which each column should have the same area.

``` r
library(ggplot2)
#> Warning: package 'ggplot2' was built under R version 4.3.3
```

``` r

breaks <- quantile(mtcars$mpg, probs = seq(0, 1, len = 5))

cut(mtcars$mpg, 
    breaks = breaks, 
    include.lowest = TRUE) |> 
  table()
#> 
#> [10.4,15.4] (15.4,19.2] (19.2,22.8] (22.8,33.9] 
#>           8           9           8           7
```

``` r

ggplot(mtcars, aes(mpg)) +
  stat_bin(color = "black", breaks = breaks)
```

![](https://i.imgur.com/C7nl2RN.png)

``` r
# default: mapping = aes(y = after_stat(count))
```

But it should look like this:

``` r
ggplot(mtcars, aes(mpg)) +
  stat_bin(color = "black", breaks = breaks,
           mapping = aes(y = after_stat(sum(count) * density)))
```

![](https://i.imgur.com/7DOeMnY.png)

<sup>Created on 2024-05-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I suggest:
1. Adding a new computed variables (perhaps called `height`) that is equal to `sum(count) * density`
2. Defaulting to this variable for setting the bars' heights, at least when using this stat when `stat_bin()` with non-equi-width bins.

teunbrand:
Hi thanks for the suggestions!

Yes, using areas for histograms satisfies the [proportional ink](https://clauswilke.com/dataviz/proportional-ink.html) principle, but below are a few reasons I don't think we should do it.

* Users have come to expect counts by default. We have parted with defaults before, but I don't think we should depart a very clear and simple metric (counts) in favour of more complicated metrics.
* Counts and the proposed metric are only the same when the width of the bars are 1. If you replace the breaks by `binwidth = 0.01`, you see several values reach 200 with the proposed metric, whereas the data only has 32 observations in total.
[...]

PR #6151, opened 2024-10-21
Expand on docs/example for cases with non-equal-width bins in `stat_bin()`

mattansb:
This PR addresses the points brought up in #5895: 
- The default behavior of `stat_bin()` of mapping counts to _height_ instead of _area_ is suitable for most common applications.
- This is because most common applications use equal-width bins.
- When non-equal-width bins are used, this behavior produces histograms that are hard to interpret and violate the [proportional ink](https://clauswilke.com/dataviz/proportional-ink.html) principle. 
- Therefore, I suggested adding to the docs/examples something that discusses this / shows how to do it properly.

Which is what I did 

### 1. Slightly expanded the _Computed variables_ bullet for `width`.

Previously was:

> - `after_stat(width)`
widths of bins.

Now is:

> - `after_stat(width)`
widths of bins. Use with `after_stat(count / width)` to obtain bars with _areas_ representing counts (e.g., with non-equal-width bins). See example.

### 2. Added an example

``` r
# When using the non-equal-width bins, we need to set the area of the bars to
# represent the counts (not the height).
# Here we're using 10 equi-probable bins:
price_bins <- quantile(diamonds$price, probs = seq(0, 1, length = 11))

ggplot(diamonds, aes(price)) +
  geom_histogram(breaks = price_bins, color = "black") # wrong (height = count)
```

``` r

ggplot(diamonds, aes(price, after_stat(count / width))) +
  geom_histogram(breaks = price_bins, color = "black") # area = count
```

<sup>Created on 2024-10-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/geom-histogram.R
+++ b/R/geom-histogram.R
@@ -64,7 +64,7 @@
 #'   geom_freqpoly(binwidth = 500)
 #'
 #'
-#' # When using the non-equal-width bins, we need to set the area of the bars to
+#' # When using the non-equal-width bins, we should set the area of the bars to
 #' # represent the counts (not the height).
 #' # Here we're using 10 equi-probable bins:
 #' price_bins <- quantile(diamonds$price, probs = seq(0, 1, length = 11))
--- a/R/geom-histogram.R
+++ b/R/geom-histogram.R
@@ -63,6 +63,18 @@
 #' ggplot(diamonds, aes(price, after_stat(density), colour = cut)) +
 #'   geom_freqpoly(binwidth = 500)
 #'
+#'
+#' # When using the non-equal-width bins, we need to set the area of the bars to
+#' # represent the counts (not the height).
+#' # Here we're using 10 equi-probable bins:
+#' price_bins <- quantile(diamonds$price, probs = seq(0, 1, length = 11))
+#'
+#' ggplot(diamonds, aes(price)) +
+#'   geom_histogram(breaks = price_bins, color = "black") # wrong (height = count)
+#'
+#' ggplot(diamonds, aes(price, after_stat(count / width))) +
+#'   geom_histogram(breaks = price_bins, color = "black") # area = count
+#'
 #' if (require("ggplot2movies")) {
 #' # Often we don't want the height of the bar to represent the
 #' # count of observations, but the sum of some other variable.
--- a/R/geom-histogram.R
+++ b/R/geom-histogram.R
@@ -70,7 +70,7 @@
 #' price_bins <- quantile(diamonds$price, probs = seq(0, 1, length = 11))
 #'
 #' ggplot(diamonds, aes(price)) +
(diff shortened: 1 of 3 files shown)


R/geom-histogram.R
```diff
@@ -63,6 +63,18 @@
 #' ggplot(diamonds, aes(price, after_stat(density), colour = cut)) +
 #'   geom_freqpoly(binwidth = 500)
 #'
+#'
+#' # When using the non-equal-width bins, we need to set the area of the bars to
```
teunbrand, line 67:
```suggestion
#' # When using the non-equal-width bins, we should set the area of the bars to
```

R/geom-histogram.R
```diff
@@ -63,6 +63,18 @@
...
+#' # represent the counts (not the height).
+#' # Here we're using 10 equi-probable bins:
+#' price_bins <- quantile(diamonds$price, probs = seq(0, 1, length = 11))
+#'
+#' ggplot(diamonds, aes(price)) +
+#'   geom_histogram(breaks = price_bins, color = "black") # wrong (height = count)
```
teunbrand, line 73:
```suggestion
#'   geom_histogram(breaks = price_bins, color = "black") # misleading (height = count)
```

R/stat-bin.R
```diff
@@ -31,7 +31,7 @@
 #'   density  = "density of points in bin, scaled to integrate to 1.",
 #'   ncount   = "count, scaled to a maximum of 1.",
 #'   ndensity = "density, scaled to a maximum of 1.",
-#'   width    = "widths of bins."
+#'   width    = "widths of bins. Use with `after_stat(count / width)` to obtain bars with _areas_ representing counts (e.g., with non-equal-width bins). See example."
```
teunbrand, line 34: Could we move this addition to the 'details' or 'notes' section of the documentation? It feels a bit out of place as part of this list.

teunbrand (changes requested):
Thanks for submitting a PR to fix this! This mostly looks great, I'd just like to move a small part for clarity.

R/stat-bin.R
mattansb: Maybe we can add to `@details` instead: 

> By default, the _height_ of the bars represent the counts within each bin. However, the are situations where this behavior might produce misleading plots (e.g., when non-equal-width bins are used), in which case it might be preferable to have the _area_ of the bars represent the counts (by setting `aes(y = after_stat(count / width))`). See example below.

What do you think?

teunbrand: I like it! Just mind the typo in 'However, *there* are ...'

mattansb:
@teunbrand I think that's all of it!

teunbrand (approved):
Lovely, looks great!

mattansb:
Thank you!

PR #6158, opened 2024-10-24
Update `update_geom_defaults()` docs

### LINKED ISSUE #6085: The future of `update_geom_defaults()`
This issue came up in https://github.com/tidyverse/ggplot2/pull/6084#discussion_r1746598566.

Briefly, now that we can do `theme(geom = element_geom(...))`, there is much less need for `update_geom_defaults()`. Per https://github.com/tidyverse/ggplot2/pull/6084#discussion_r1746650639, we should ask ourselves if `update_geom_defaults()` should be deprecated in a future release.

In my opinion, there is still a niche for `update_geom_defaults()` as one might want to swap out defaults anyway, in particular when it should affect 1 particular geom and not all of them. I sometimes (ab)use it to get geoms to accept the aesthetics I want ([example](https://github.com/teunbrand/ggplot_tricks?tab=readme-ov-file#half-violin)). I'd therefore advise against future deprecation, but we should probably pivot from recommending `update_geom_defaults()` to `theme(geom)`.

teunbrand:
This PR aims to fix #6085.

Briefly, it explains the 'all geoms versus single geom' axis versus 'update once versus update always' axis of using the update functions, suggested here: https://github.com/tidyverse/ggplot2/issues/6085#issuecomment-2333594467.

### DIFF
--- a/R/geom-defaults.R
+++ b/R/geom-defaults.R
@@ -11,7 +11,17 @@
 #' @keywords internal
 #' @note
 #' Please note that geom defaults can be set *en masse* via the `theme(geom)`
-#' argument.
+#' argument. The guidelines for when to use which function are as follows:
+#'
+#' * If you want to change defaults for all geoms in all plots, use
+#'   `theme_update(geom = element_geom(...))`.
+#' * If you want to change defaults for all geoms in a single plot, use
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #6160, opened 2024-10-25
Rename `is.element()` to `is.theme_element()`

### LINKED ISSUE #6157: Conflict between `base::is.element()` and `ggplot2::is.element()`
In #6022, we introduced the `is.element()` function to test if some object is an `element_rect()`, `element_text()`, `element_blank()` etc.:

https://github.com/tidyverse/ggplot2/blob/ddd207e926cc1c1847dc661d9a099b8ec19c4010/R/theme-elements.R#L205-L207

However, there is also `base::is.element()`, which acts as `%in%`. Since it is unlikely that base will rename that function anytime soon, we might have to come up with an alternative name for testing if an object is any of the theme elements. Should we just use `is.theme_element()`?

teunbrand:
This PR aims to fix #6157.

Briefly, to test if something is a theme element, one now uses `is.theme_elment()` instead of `is.element()`.


thomasp85 (approved):
LGTM

PR #6163, opened 2024-10-26
Typo in geom-map.R

jonocarroll:
Minor typo in documentation.


teunbrand:
\document

jonocarroll:
Interesting; shouldn't the .Rd files be generated artefacts and thus whatever is committed is redundant? In any case, I've re-generated and committed the relevant file.

teunbrand:
Thanks for the update!
AFAIK we have no mechanism in place to automate regenerating the docs.
I think {gtable} has had a recent update that borks some tests that fail.
These are unrelated to your PR, so I'll approve this and merge when we've updated the tests.

teunbrand (approved):

teunbrand:

PR #6156, opened 2024-10-24
Upkeep 2024-10

teunbrand:
This PR aims to fix #6155.

Marked as a draft for now as things are checked off.

PR #6168, opened 2024-10-29
Checks if svglite is installed when saving to svg.

### LINKED ISSUE #6166: Saving to svg without svglite installed throws unhelpful error
Trying to save a plot to svg without svsglite installed (which is possible because svglite is a Suggest) throws this error:

```
> ggsave("poster.svg")
Saving 7 x 7 in imageError in loadNamespace(x) : there is no package called ‘svglite’
```

The error is not very helpful as it doesn't tell the user what the issue is and how to resolve it. 

The problem is that `plot_dev` doesn't check if svglite is installed here:

https://github.com/tidyverse/ggplot2/blob/b174986fbf9d6c2c7b0f0356a4dc98a21bf7de84/R/save.R#L280

so R errors out when trying to evaluate the `dev` function; probably here: 

https://github.com/tidyverse/ggplot2/blob/b174986fbf9d6c2c7b0f0356a4dc98a21bf7de84/R/save.R#L109

Maybe a better alternative would be to do something like this: 

```r
svg =  function(filename, ...) { 
   check_installed("svglite", reason = "to save as SVG.")
   svglite::svglite(file = filename, ...)
} 
```

eliocamp:
Closes #6166

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -232,6 +232,7 @@ documentation updates.
 * `annotate()` now warns about `stat` or `position` arguments (@teunbrand, #5151)
 * `guide_coloursteps(even.steps = FALSE)` now works with discrete data that has 
   been formatted by `cut()` (@teunbrand, #3877).
+* `ggsave()` now offers to install svglite if needed (@eliocamp, #6166).
 
 # ggplot2 3.5.0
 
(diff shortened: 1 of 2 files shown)


teunbrand (approved):
Looks great, thank you!
The failing tests are unrelated, but I'm going to wait to merge until these are resolved in #6156.

PR #6169, opened 2024-10-29
Checks for `register_theme_element(element_tree)`

### LINKED ISSUE #6162: Circular `register_theme_element()` bug
Accidentally, I found myself in a situation where I defined a theme element with itself as its parent.
I couldn't reprex this bug because it halts execution of `rmarkdown::render()`, but here are the steps to reproduce:

```r
library(ggplot2)
register_theme_elements(
  foobar = element_rect(),
  element_tree = list(foobar = el_def("element_rect", inherit = "foobar"))
)
calc_element("foobar", theme_get())
#> Error in match.fun(FUN) : node stack overflow
#> Error during wrapup: node stack overflow
#> Error: no more error handlers available (recursive errors?); invoking 'abort' restart
```

It is fine when I use `inherit = "rect"` or something that isn't the name of the element itself.
So it seems like defining an element with itself as its parent causes the bug.

This error was complicated to track down and the error message itself wasn't very helpful.
I propose we add a check to `register_theme_elements()` that throws an error when trying to define new elements that have themselves as parents.

teunbrand:
This PR aims to fix #6162.

Aside from throwing an error in de issue's case, I also included stricter checks on the structure of the `element_tree` argument.
There is the possibility this might break packages that define their own theme elements, but if so, they've been using it incorrectly.

Reprex from issue:
``` r
#> ℹ Loading ggplot2
register_theme_elements(
  foobar = element_rect(),
  element_tree = list(foobar = el_def("element_rect", inherit = "foobar"))
)
#> Error in `register_theme_elements()`:
#> ! Invalid parent in `element_tree`: "foobar".
```

<sup>Created on 2024-10-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #6170, opened 2024-10-29
Note `I()` behaviour may be unexpected

### LINKED ISSUE #5863: AsIs sometimes not preserved when computing geom parameters
At least in GeomBar (and possibly elsewhere), when an `<AsIs>` aesthetic for bar height is parameterized, the `<AsIs>` class is sometimes dropped.

```r
library(ggplot2)

p <- data.frame(x = 1:2, y = c(0.5, 2)) |>
  ggplot(aes(x, I(y))) +
  geom_col()
p
```

image

Here, `ymax` is stripped of `<AsIs>`:

```r
tibble::as_tibble(
  layer_data(p)[, c("ymin", "y", "ymax")]
)
#> # A tibble: 2 × 3
#>       ymin        y  ymax
#>   <I<dbl>> <I<dbl>> <dbl>
#> 1        0      0.5   0.5
#> 2        0      2     2
```

This also has consequences for training of the scales.[^1] In the plot, the height of the shortest bar is determined only by `expansion`

```r
ggplot_build(p)$layout$get_scales(1)$y
#> <ScaleContinuousPosition>
#>  Range:   0.5 --    2
#>  Limits:  0.5 --    2
```

---

Expected plot of bar heights rendered in npc:

image

[^1]: I'm actually torn on whether this part is also a bug. Technically, the bars *are* following the baseline-at-0 constraint (just that 0 is now interpreted in npc, which is meaningless). But maybe in this case GeomBar should override the baseline of the bars to always be on *data* scale (probably hard)? Or if users really want that for whatever reason, they could just add `ylim(0, NA)` I suppose

teunbrand:
This PR aims to fix #5863.

It makes a note pointing out using `I()` can lead to unexpected results


thomasp85 (approved):
LGTM

PR #6171, opened 2024-10-31
Use snapshot expectations for transformation breaks

teunbrand:
This PR fixes an issue discussed off-github.

Briefly, scales 1.4.0 was poised to break ggplot2 because `transform_exp()`'s breaks were improved and broke a test.
This PR makes all the transformation break tests snapshots, which means ggplot2 will not break on CRAN in the future if breaks are further improved.


thomasp85 (approved):
LGTM

PR #6173, opened 2024-11-01
Export `is.waive()` for extensions to use

arcresu:
`waiver()` is an internal detail of ggplot that end-users don't really need to be aware of, but extensions that define new `stat_*()` or `geom_*()` functions generally need to be able to check for waivers for the same reason that ggplot itself does.

I find that most extensions I write need to copy the definition of `is.waive()`, and I'm [not alone](https://github.com/search?q=function%28x%29+inherits%28x%2C+%22waiver%22%29+path%3A%2F%5ER%5C%2F%2F&type=code). Would you consider exporting it for extensions to use?


teunbrand (approved):
Agreed, this would be good to have available in extensions.
Would you mind adding a news bullet stating that `is.waiver()` is now exported?

arcresu:
Done. Though that does raise the question of whether this should be `is.waive()` or `is.waiver()`. I got used to the former, but since the class is `waiver` we could rename it? I'm happy either way.

arcresu:
No problem; rename done.

teunbrand:
Great, thank you very much for the contribution!

PR #6175, opened 2024-11-01
Fix ragg version check in test skip

MichaelChirico:
https://github.com/tidyverse/ggplot2/blob/b29b83151a8415093cadc8aead5024452ed7eae6/R/utilities-checks.R#L315-L318


teunbrand (approved):
LGTM, thanks!

PR #6176, opened 2024-11-01
Adjust version for svglite skip

MichaelChirico:
Should have been caught with #6175, sorry about that.

https://github.com/tidyverse/ggplot2/blob/cc329df041573c4ca0ce31b491de0144506c12f6/R/utilities-checks.R#L321-L324


teunbrand (approved):
Thank you!

PR #6186, opened 2024-11-15
Weighted ellipses

### LINKED ISSUE #5272: Weight parameter for stat_ellipse()
Geom_density, geom_barplot, geom_histogram and geom_barplot all support aes weight. Stat_ellipse() doesnt support.

I figured out that data[rep(data$rownumber,data$weight),] is a workaround, but that doesnt work too well(or at all) when weights are large. Scaling the weights could work, but then it would lose the accuracy of original data.

Would it be possible to add stat_ellipse() to suppport weight aesthetics?

teunbrand:
This PR aims to fix #5272.

Briefly, it adds a `weight` aesthetic to `stat_ellipse()`.

I don't know enough about data ellipses to verify the correctness of this approach.

In the example, we can see that the solid, weighted ellipses are shifted to the right because observations to the right have higher weights than those to the left.

``` r
#> ℹ Loading ggplot2

ggplot(faithful, aes(waiting, eruptions, color = eruptions > 3)) +
  geom_point() +
  stat_ellipse(alpha = 0.3) +
  stat_ellipse(aes(weight = rescale(waiting)))
```

<sup>Created on 2024-11-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #6187, opened 2024-11-15
Allow saving PDF pages

### LINKED ISSUE #5093: New feature to allow ggsave to save multiple plots to separate pages of pdf
It would be super great if `ggsave()` could save multiple plots to the same pdf (on different pages). 

I think ideally, the plot argument would be able to be passed a list. Same with the width and height, so that you can specify that for each plot.

teunbrand:
This PR aims to fix #5093.

Briefly, it allows `ggsave(plot)` to be a list and draws every element in that list.

While this works, it has a few drawbacks:
* No warning is thrown when `device != "pdf"`. This is because it is hard to detect whether the current device is a PDF device that supports the `onefile` argument.
* I don't think there is a way to set different `bg` for every page, so instead we don't use the trick where we set the background to the plot background colour.

Example usage:

``` r
#> ℹ Loading ggplot2

pl <- list(
  p + ggtitle("First"),
  p + ggtitle("Second"),
  p + ggtitle("Last")
)

# `onefile = TRUE` is default, but clarifies intent
ggsave("pages.pdf", plot = pl, device = "pdf", onefile = TRUE)
#> Saving 7 x 5 in image
```

<sup>Created on 2024-11-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/save.R
```diff
@@ -102,7 +102,7 @@ ggsave <- function(filename, plot = get_last_plot(),
   dim <- plot_dim(c(width, height), scale = scale, units = units,
     limitsize = limitsize, dpi = dpi)
 
-  if (is_null(bg)) {
+  if (is_null(bg) && is.ggplot(plot)) {
```
thomasp85, line 105: In the case a list is passed in, I think we should just grab the bg from the first plot in the list rather than completely ignore it

R/save.R
```diff
@@ -111,7 +111,10 @@ ggsave <- function(filename, plot = get_last_plot(),
     grDevices::dev.off()
     if (old_dev > 1) grDevices::dev.set(old_dev) # restore old device unless null device
   }))
-  grid.draw(plot)
+  if (!is_bare_list(plot)) {
+    plot <- list(plot)
+  }
+  lapply(plot, grid.draw)
```
thomasp85, line 117: There should be a `grid.newpage()` in between every draw call

teunbrand: Yes I would have thought so as well. This would make sense, but `grid.draw.ggplot()` calls `print.ggplot()`, which has the following clause:
https://github.com/tidyverse/ggplot2/blob/8efc700377c6bfb0583909846f7dc235b0122c34/R/plot.R#L196

which gets activated by default.

thomasp85: oh, yeah... forgot it was a ggplot object 🙈

R/save.R
teunbrand: This is sensible thanks!

thomasp85 (approved):
LGTM

PR #6193, opened 2024-11-20
Deprecate `fortify()` for models

teunbrand:
This PR aims to advance #3816.

It deprecates the `fortify()` methods for statistical models in favour of `broom::tidy()` and `broom::augment()`.
Examples now use broom outputs in plots, which requires adding broom to 'Suggests'.
Also the `fortify.lm()` examples have been condensed, because I couldn't figure out what they were showing aside from variable names.


teunbrand:
Thank you for the review!

PR #6194, opened 2024-11-21
Stability of layer data attributes

### LINKED ISSUE #3175: Access to layer information during build stages
I'm opening this issue as a more general discussion ground for the discussions going on in #3170 (which fixes #3116). The reason for this is that in my attempt to implement #3062 I've run into more or less the same issues but from the facet side, and I think the time is up for dealing without it in a more general way.

**The problem:** Throughout ggplot2's code base the layer data is the sole representative of the layer. It is not assumed that the layer have any other information of interest to share. This makes for clean code. We don't need to pass the layer objects around so we get nice separation of responsibility. But it also means that the layer cannot tell the outside world about any special property it has. Generally this has not been a problem as any specifics has been relevant to the geom and stat exclusively (part of the internal of the layer).

**One solution:** Take the approach given in #3170 and generalise it, so that any function receiving layer data also receives a `layer_param` object, as well as formalising setting up this object (probably within `setup_layer()`). This will require sweeping changes to many methods across many classes (internally and in extension packages) so It's certainly less attractive because of that. On the other hand it will result in an architecture that is consistent (using a params object). We may be able to modify the ggproto dispatch to quietly ignore unknown arguments in order to avoid breaking extension packages but it would remove the safety of early failure when misspelling arguments in code.

**Another solution:** Attenuate the data object with layer param attributes - this would allow everything to continue to work as it has done before, but will require internal changes to make sure the data keeps the information as it is passed around. A `layer_param` attribute could keep all relevant information that we would otherwise pass around as a `layer_param` argument. This will undoubtedly result in some more ugly code though helper functions could probably tidy most of it up. Worst is probably that we now divert into two different ways of passing around information between objects.

**A kind of middle ground:** Each layer data could also simply get an attribute (or an additional column) telling the index of the layer in the layer stack, and e.g. the facets could seep out layer information from all the layers during setup and then take care of matching that information with the data itself if needed. I'm not sure there's any benefit to this other than avoiding to put arbitrary amount of information in the data attributes. It would certainly result in even more housekeeping code than option two

I'm sure there are solutions I haven't considered so I'd be happy to take a broad discussion about how we want to solve it. I'm certain this is something we want to solve in general in order for development to move forward. (as an aside (can't find the issues right now), it will also allow for the layers to take additional params needed by extensions without triggering the `unknown argument` warning—I remember ggplotly requesting this at least.

@hadley, @karawoo, @clauswilke, @yutannihilation looking forward to your input

teunbrand:
This PR aims to fix #3175.

It avoids dropping attributes in some places and actively restores them in other places.
Attributes should now be stable from the moment aesthetics are computed up to the `Geom$draw_*()` methods.


yutannihilation:
I too cannot figure out, but I confirmed this error is reproducible on my local. As it looks too cryptic, maybe we can just ignore the failing tests for now? cf. https://covr.r-lib.org/#exclusions

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-layer.R:177:3'): attributes on layer data are preserved ──────
attr(ld, "test") (`actual`) not equal to "preserve me" (`expected`).

`actual` is NULL
`expected` is a character vector ('preserve me')
── Failure ('test-scales.R:71:3'): identity scale preserves input values ───────
`d1` (`actual`) not equal to `d2` (`expected`).

     names(actual) | names(expected)     
 [1] "shape"       - "size"          [1] 
 [2] "size"        - "alpha"         [2] 
 [3] "alpha"       - "x"             [3] 
 [4] "x"           - "y"             [4] 
 [5] "y"           - "colour"        [5] 
 [6] "colour"      - "fill"          [6] 
 [7] "fill"        - "shape"         [7] 
 [8] "PANEL"       | "PANEL"         [8] 
 [9] "group"       | "group"         [9] 
[10] "stroke"      | "stroke"        [10]

[ FAIL 2 | WARN 1 | SKIP 258 | PASS 1593 ]
```

teunbrand:
I think I've made some debugging process. I think for some reason, `ScalesList$transform_df()` correctly tells the plot to skip on identity transformations in most contexts, but we do not skip this in the coverage test. The clue for this comes from #6076, where identity scales protest in `check_transformation()`, though it has no business checking identity transformations.

```
(df = df) at ggplot2/R/scales-.R:102:5
 11.                 └─ggplot2 (local) transform_df(..., self = self) at ggplot2/R/ggproto.R:188:17
 12.                   └─base::lapply(df[aesthetics], self$transform) at ggplot2/R/scale-.R:520:5
 13.                     └─ggplot2 (local) FUN(X[[i]], ...)
 14.                       └─ggplot2 (local) transform(..., self = self) at ggplot2/R/ggproto.R:[188](https://github.com/tidyverse/ggplot2/actions/runs/11951275132/job/33694697269?pr=6194#step:5:189):17
 15.                         └─ggplot2:::check_transformation(...) at ggplot2/R/scale-.R:640:3
 ```

thomasp85 (changes requested):
Can you please add a test for being able to modify the attribute during computations

teunbrand:
Added a test. I should note that within the 'layer - panel - group' hierarchy attributes are unstable. So if you want to append some attributes, you should do it at the layer-level. This is because if you do it at e.g. the panel level and one panel returns a different attribute value than another, there is no principled way of merging attributes.

thomasp85 (approved):
LGTM

PR #6196, opened 2024-11-22
Scale parameters

### LINKED ISSUE #4269: Set common scale parameters outside of the scale function

teunbrand:
This PR follows up on #4271 and #5444, aiming to fix #4269.

Like before, it aims to provide the convenience of setting some scale parameters, without comitting to a full scale yet.
Unlike #5444, this PR is closer to Claus' approach in #4271.

This PR the `scale_params()` function that populates a new `ScalesList$params` field. When a new scale is added, it will get updated with the parameters in that field. The update mechanism required the most attention. Every `<Scale>` class now has a `Scale$new()` method that takes on most of the work that `continuous_scale()`, `discrete_scale()`, `binned_scale()` constructors were doing before. The `Scale$update()` method is a thin wrapper around this and should resolve gnarly situations (e.g. limit transformation).

In the demo below we can note:
* the y-limits are appropriately transformed by the sqrt transformation. If we were to use the transform first and then the limits, the result would be the same.
* We can set parameters for multiple scales at once, like colour/shape.

``` r
#> ℹ Loading ggplot2
p <- ggplot(mpg, aes(displ, hwy, colour = drv, shape = factor(year))) +
  geom_point()

p +
  scale_params("x", breaks = c(2, 4, 7)) +
  scale_params("y", limits = c(0, 50)) +
  scale_params("y", transform = "sqrt") +
  scale_params(c("colour", "shape"), guide = guide_legend(ncol = 2))
```

The update mechanism should also recognise invalid updates:

``` r
p + scale_params("colour", foobar = "i dont exist")
#> Error in `geom_point()`:
#> ! Problem while computing aesthetics.
#> Caused by error in `scale_colour_discrete()`:
#> ! Cannot update scale with the unknown argument:
#> • "foobar"
```

<sup>Created on 2024-11-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/scale-.R
```diff
@@ -633,6 +547,29 @@ check_breaks_labels <- function(breaks, labels, call = NULL) {
   TRUE
 }
 
+check_update_params <- function(scale, params) {
+  if (inherits(scale, "ScaleContinuous")) {
+    args <- fn_fmls_names(continuous_scale)
+  } else if (inherits(scale, "ScaleDiscrete")) {
+    args <- fn_fmls_names(discrete_scale)
+  } else if (inherits(scale, "ScaleBinned")) {
+    args <- fn_fmls_names(binned_scale)
```
thomasp85, line 556: Would be nice if this information was somehow incorporated into the class, rather than having to check an "unlinked" constructor function

teunbrand:
This could be a nice editioning carrot

jbengler:
Just giving this PR a little bit of love ❤️ — I’d be thrilled to see it get implemented!

PR #6200, opened 2024-11-25
Scale name function

### LINKED ISSUE #4313: Allow scale name to be a function.
I think it might be useful to have the possibility of passing a function to scale names like so:

``` r
library(ggplot2)

df <- expand.grid(X1 = 1:10, X2 = 1:10)
df$value <- df$X1 * df$X2
p1 <- ggplot(df, aes(X1, X2)) + geom_tile(aes(fill = value))

p1 + scale_fill_continuous(function(x) base::toupper(x))
```

``` r
# With function shorthand
p1 + scale_fill_continuous(~ base::toupper(.x))
```

<sup>Created on 2021-01-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Something like this would make code snippets more portable. In my case, I have a function that makes a plot and I had to define the title outside the plot depending on the function arguments. With this, I could've defined a global dictionary which translates column names to actual (human readable) variable names and re-use that all over my report. 

To produce the code above I just added this

```r
  if (is.formula(name)) {
    name <- as_function(name)
  }

  if (is.function(name)) {
    make_title <- name
    name <- waiver()
  } else {
    make_title <- super$make_title
  }
```

before this line: 

https://github.com/tidyverse/ggplot2/blob/master/R/scale-.r#L107

And then added `make_title = make_title` to the ggproto construction. 

All tests pass, so changing this should be safe.

teunbrand:
This PR aims to fix #4313.

Briefly, it enables users to set functions in `labs()`, `scale_*()` and `guide_*()` functions that are applied to the labels.

Here is an example:

``` r
devtools::load_all('~/packages/ggplot2/')
#> ℹ Loading ggplot2
df <- mtcars
attr(df$disp, "label") <- "engine displacment"

ggplot(df, aes(disp, mpg, colour = drat, shape = factor(cyl))) +
  geom_point() +
  labs(
    y = to_upper_ascii,
    shape = function(x) gsub("factor", "foo", x)
  ) +
  scale_shape_discrete(
    name = to_upper_ascii,
    guide = guide_legend(title = function(x) paste0(x, "!!!"))
  ) +
  scale_x_continuous(name = to_upper_ascii) +
  guides(colour = guide_colourbar(title = to_upper_ascii))
```

<sup>Created on 2024-11-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/guide-bins.R
```diff
@@ -212,7 +212,7 @@ GuideBins <- ggproto(
       key$.value <- 1 - key$.value
     }
 
-    params$title <- scale$make_title(params$title %|W|% scale$name %|W|% title)
+    params$title <- scale$make_title(params$title, scale$name, title)
```
teunbrand, line 215: Ideally, this would happen in the base `Guide` class, but axes don't have access to labels.

Issue #3994, opened 2020-05-13
Drop plot_env from ggplot2 objects?


clauswilke:
This comment https://github.com/tidyverse/ggplot2/issues/3619#issuecomment-628021555 prompted me to look into `plot_env`, and as far as I can see it's stored and handed around just so it can be eventually given to the function `combine_vars()`, which then doesn't use it:
https://github.com/tidyverse/ggplot2/blob/e9b9946786dae861dea1c352f2ac7b2a837b5f82/R/facet-.r#L544-L587
Is this a hold-over from the times before tidy eval? Can we remove this/or assign `NULL` here?
https://github.com/tidyverse/ggplot2/blob/5a686c34b304ddefeee16d44362f326cfa29634e/R/plot.r#L96

It is already marked as deprecated in the documentation:
https://github.com/tidyverse/ggplot2/blob/5a686c34b304ddefeee16d44362f326cfa29634e/R/plot.r#L35

clauswilke:
Setting `plot_env` to `NULL` doesn't work, which means it is still used somewhere as an environment. However, setting it to an empty environment seems to work without issues.

``` r
library(ggplot2)

p <- ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species)) +
  geom_point() +
  facet_wrap(~Species)

p$plot_env <- NULL
p
#> Error in exists(name, envir = env, mode = mode): use of NULL environment is defunct

p$plot_env <- rlang::new_environment()
p
```

![](https://i.imgur.com/3E9VJMJ.png)

<sup>Created on 2020-05-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It looks like the other place where the environment is used is for scale lookup. Maybe it would be better to use the global environment that is active when the plot is printed, rather than when the plot is created? That would be more predictable anyways, I assume.
https://github.com/tidyverse/ggplot2/blob/5a686c34b304ddefeee16d44362f326cfa29634e/R/scale-type.R#L28-L39

clauswilke:
On the last point, just setting `plot_env = rlang::new_environment()` means we cannot globally change color scales anymore. So we'd have to fix this appropriately.

``` r
library(ggplot2)

p <- ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species)) +
  geom_point() +
  facet_wrap(~Species)

scale_colour_discrete <- scale_colour_viridis_d
p
```

[...]

PR #6203, opened 2024-11-26
WIP: Initial infrastructure for editions

teunbrand:
This PR aims to adress a topic discussed by Thomas and me off-github.

Briefly, we'd like to bring edition structure to ggplot2 to opt-in otherwise backward incompatible changes. It is modelled after [testthat](https://testthat.r-lib.org/articles/third-edition.html), with a few adaptations. This PR just aims to introduce infrastructure for working with editions in ggplot2.

The most important difference with testthat is in the relation that it has to other packages. The testthat package has 1:1 relationships: the tests of one single package does not affect tests in a second package. In contrast, it is not unreasonable to assume that for example {extensionPkg} defines ggplot2 extensions, which are then subsequently used by wrappers in {wrapperPkg}. In this example {extensionPkg} and {wrapperPkg} might opt-in different editions, which needs to be facilitated. For this reason, this PR gives each package their own local edition.

This PR is currently marked WIP as some details need to be ironed out. Specifically:

* How do we expect maintainers of reverse dependencies to interact with the edition system?
* How do we expect users to interact with the edition system?
* What type of things do we want to edition?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,11 +33,13 @@ Depends:
     R (>= 4.0)
 Imports: 
     cli,
+    desc,
     grDevices,
     grid,
     gtable (>= 0.1.1),
     isoband,
     lifecycle (> 1.0.1),
+    pkgload,
     rlang (>= 1.1.0),
     scales (>= 1.3.0),
     stats,
@@ -122,6 +124,7 @@ Collate:
     'coord-transform.R'
     'data.R'
     'docs_layer.R'
+    'edition.R'
     'facet-.R'
     'facet-grid-.R'
     'facet-null.R'
--- a/R/edition.R
+++ b/R/edition.R
@@ -1,0 +1,72 @@
+
+ggplot_edition <- new.env(parent = emptyenv())
+
+local_edition <- function(edition, .env = parent.frame()) {
+  stopifnot(is_zap(edition) || (is.numeric(edition) && length(edition) == 1))
+  pkg <- get_pkg_name(.env)
+  local_bindings(!!pkg := edition, .env = ggplot_edition, .frame = .env)
+}
+
+edition_get <- function(.env = parent.frame(), default = 2024L) {
+  pkg <- get_pkg_name(.env)
+
+  # Try to query edition from cache
+  edition <- env_get(ggplot_edition, nm = pkg, default = NULL)
+  if (!is.null(edition)) {
+    return(edition)
+  }
+
+  # Try to query package description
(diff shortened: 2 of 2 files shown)


DESCRIPTION
```diff
@@ -33,11 +33,13 @@ Depends:
     R (>= 4.0)
 Imports: 
     cli,
+    desc,
     grDevices,
     grid,
     gtable (>= 0.1.1),
     isoband,
     lifecycle (> 1.0.1),
+    pkgload,
```
teunbrand, line 42: I don't like having to take on these dependencies.
Would it be bad to have a homebrew solution to parsing the description file?

teunbrand:
Add a few example cases to edition

R/edition.R
```diff
@@ -0,0 +1,72 @@
...
+  msg <- paste0(what, " requires the ", edition, " edition of {.pkg ggplot2}.")
+  cli::cli_abort(msg)
+}
+
+find_description <- function(path, package = NULL) {
+  if (!is.null(package)) {
+    return(desc::desc(package = package))
+  } else {
+    try_fetch(
+      pkgload::pkg_desc(path),
```
teunbrand, line 59: Avoid this block

teunbrand:
To answer some of the questions:

> How do we expect maintainers of reverse dependencies to interact with the edition system?

I think we expect maintainers to:
* Opt-in the 2025 edition by setting a 'Config/ggplot2/edition' value in their DESCRIPTION file.
* Use `edition_require()` whenever implementing features that rely on new editions.
* Use `edition_deprecate()` whenever they want to deprecate something along with ggplot2 editions.

How do we expect users to interact with the edition system?

I don't have a full answer yet. Currently they can use `local_edition(2025)` to opt-in newer features. I think it would be neat if we can have regular users opt-out of the 2025 edition while having packages opt-in, but this seems counterintuitive to implement.

teunbrand:
After discussion with Thomas, we want to use editioning only to deprecate mostly vestigial parts and swap defaults, not for opting in to newer features. Some examples of things we want to edition:

* default colour/fill scale palette
* default density bandwidth estimator (#5854)
* removing plot environments (#3994)
* deprecating ggproto classes (#5962)

teunbrand:
Putting this PR on hold per discussion that we don't want any edition infrastructure just yet in 4.0.0, just supersession of functions/arguments.

thomasp85:
We can close this, right?

teunbrand:
That's right, closing this in favour of #6756

PR #6206, opened 2024-11-27
Non-orthogonal coords preserve AsIs x/y variables

### LINKED ISSUE #6205: 'AsIs' position invalid when setting coord_sf: 'default_crs' parameter
The problem stems from an annotation about adding an out-of-panel position annotate. https://stackoverflow.com/questions/79222655/ggplot2-adds-a-second-image-title-on-the-right-side It has been suggested that ggplot 3.5.0 provides a more convenient location setup : `‘AsIs’ position aesthetics I() `. However, when mapping using `coords_sf`, there are some problems: everything works fine without asserting `default_crs` in it. Once this parameter is supplied, the labeled position changes. It should go back to using the x,y axis coordinates as defined. 
 
```r
library(tidyverse)
library(sf)
library(rnaturalearth)
ne_countries(continent = "africa") |> 
  # st_crs()
  ggplot()+
  geom_sf()+
  ggtitle('left')+
  annotate("text", label = "right", x = I(1), y = I(1), hjust = 1,
           vjust = -1, size = 4) +
  coord_sf(crs ="+proj=laea +y_0=0 +lon_0=155 +lat_0=-90 +ellps=WGS84 +no_defs",
           # default_crs = st_crs(4326),
           clip = "off")
```
Expected results:
Image

```r
library(tidyverse)
library(sf)
library(rnaturalearth)
ne_countries(continent = "africa") |> 
  # st_crs()
  ggplot()+
  geom_sf()+
  ggtitle('left')+
  annotate("text", label = "right", x = I(1), y = I(1), hjust = 1,
           vjust = -1, size = 4) +
  coord_sf(crs ="+proj=laea +y_0=0 +lon_0=155 +lat_0=-90 +ellps=WGS84 +no_defs",
           default_crs = st_crs(4326),
           clip = "off")
```
The position of the labeling has changed and seems to follow the definition of the axes: 
Image

teunbrand:
This PR aims to fix #6205.

Briefly in `coord_sf()`,  `coord_polar/radial()`, the x/y variables are preserved if they both have the 'AsIs' type. I don't think there is a reasonable interpretation if just one of the variables is 'AsIs' and the other is a plain numeric.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* In non-orthogonal coordinate systems (`coord_sf()`, `coord_polar()` and 
+  `coord_radial()`), using 'AsIs' variables escape transformation when
+  both `x` and `y` is an 'AsIs' variable (@teunbrand, #6205).
 * Custom and raster annotation now respond to scale transformations, and can
   use AsIs variables for relative placement (@teunbrand based on 
   @yutannihilation's prior work, #3120)
--- a/R/coord-polar.R
+++ b/R/coord-polar.R
@@ -180,6 +180,10 @@ CoordPolar <- ggproto("CoordPolar", Coord,
   },
 
   transform = function(self, data, panel_params) {
+    if (inherits(data$x, "AsIs") && inherits(data$y, "AsIs")) {
+      return(data)
+    }
+
     arc  <- self$start + c(0, 2 * pi)
     dir  <- self$direction
     data <- rename_data(self, data)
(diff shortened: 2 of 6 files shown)


yutannihilation:
Thanks for this fix. Generally looks good to me.

> I don't think there is a reasonable interpretation if just one of the variables is 'AsIs' and the other is a plain numeric.

Should this be ignored with a warning rather than with no warning? Would it be annoying? Probably it's a rare case anyway, so I think either is fine.

teunbrand:
> Should this be ignored with a warning

It probably can't hurt to include it, so now it throws a warning like:

```
`coord_radial()` cannot respect the <AsIs> class of `x` when `y` is not also <AsIs>.
```

yutannihilation (approved):
Looks good!

teunbrand:
Thank you for the review!

Issue #5712, opened 2024-02-26
Multiple legends with position="inside"


MauricioCely:
I have recently installed the `ggplot2` 3.5.0 release, and after reading the new features added to the Legends [Placement](https://www.tidyverse.org/blog/2024/02/ggplot2-3-5-0-legends/#placement), I was very curious to know if it is possible to have multiple inside legends (`position = "inside"`) at different positions by using the new `legend.position.inside` argument for each guide. For example, `cty` legend at bottom-right and `cyl` at top-right. However, it does not work.

``` r
library(ggplot2) # Create Elegant Data Visualisations Using the Grammar of Graphics CRAN v3.5.0

ggplot(mpg, aes(displ, hwy, shape = drv, colour = cty, size = year)) +
  geom_point(aes(alpha = cyl)) +
  guides(
    colour = guide_colourbar(position = "inside",
                             theme = theme(legend.position.inside = c(1, 0))),
    size   = guide_legend(position = "top"),
    alpha  = guide_legend(position = "inside", 
                          theme = theme(legend.position.inside = c(0, 1))),
    shape  = guide_legend(position = "left")
  )
```

![image](https://github.com/tidyverse/ggplot2/assets/26229943/80395629-15d8-4ca3-bb11-91609bc642d2)

It seems only works by using the global theme’s `legend.position.inside` arguments, grouping both legends at the same position, but for now is not possible to not place them separately.

thomasp85:
This is not possible at this time

teunbrand:
Should this be possible? I cannot think of a compelling use case on top of my head, or at least, more compelling than satisfying curiosity.

MauricioCely:
Some time ago, I saw this issue in Stackoverflow:

[https://stackoverflow.com/questions/21798364/independently-move-2-legends-ggplot2-on-a-map](https://stackoverflow.com/questions/21798364/independently-move-2-legends-ggplot2-on-a-map)

botanize:
I'm running into this as well. The documentation doesn't say anything about which components of theme are respected in `guide_legend`, at the moment there's no warning or error, the plot just doesn't respect the API as documented.

twest820:
[...]

PR #6210, opened 2024-11-28
multiple inside guide box with different position

Yunuuuu:
Try to fix https://github.com/tidyverse/ggplot2/issues/5712.

This enables users to set `legend.position.inside` and `legend.justification.inside` in `theme` of `guide_legend` and `guide_colourbar`.

Here is an example:
```r
ggplot(mpg, aes(displ, hwy, shape = drv, colour = cty, size = year)) +
    geom_point(aes(alpha = cyl)) +
    guides(
        colour = guide_colourbar(
            position = "inside",
            theme = theme(legend.position.inside = c(1, 0))
        ),
        size = guide_legend(position = "top"),
        alpha = guide_legend(
            position = "inside",
            theme = theme(legend.position.inside = c(0, 1))
        ),
        shape = guide_legend(position = "left")
    )
```
image


teunbrand (changes requested):
Hi Yunuuu, thanks for preparing a PR for this issue!

The PR appears to break a visual test for the alignment of legends with the inside position. It is important to us to preserve current behaviour without regressions.

Next, it is hard to tell from the PR what the assembly rules for inside legends are. When legend share `legend.position.inside` values, they merge (be placed in the same guide-box). It seems that justification also factors into this and I'm not convinced that it should. It'd be great if the logic for this is expressed more clearly, perhaps through code comments.

Lastly, there are a few fussy changes to the code I'd like to propose, but it'd make more sense to go over this after the PR adresses the previous points.

Yunuuuu:
@teunbrand Thank you for the reviewing. I have made the following changes based on your feedback:

1. Added more comments and explanations throughout the code to improve clarity.
2. Fixed the test error that was identified.
3. Merged the guide legends based solely on `legend.position.inside`, as it seems sufficient to identify the legend groups.

Yunuuuu:
I'll check it

Yunuuuu:
Should we always keep the inside guide box, even if there are no inside guide legends?

teunbrand:
> Should we always keep the inside guide box, even if there are no inside guide legends?

An empty inside guide box should give a `zeroGrob()` for consistency in gtable layout

Yunuuuu:
should fixed

R/guide-legend.R
```diff
@@ -559,9 +560,16 @@ GuideLegend <- ggproto(
...
+        t = 1, r = -1, b = -1, l = 1, z = -Inf
       )
     }
+
+    # for inside guide legends, we also save the position values 
+    # in this way, we can identify legends with same position
+    # and then merge them into same guide box in `Guides$draw()`
+    if (identical(.subset2(params, "position"), "inside")) {
+      attr(gt, "inside_position") <- .subset2(elements, "inside_position")
+    }
```
teunbrand, line 572: If think it should be possible to circumvent touching the legend/colourbar guides here if we move away from this attribute-based propagation of inside coordinates. If we can make this work without changing the guides, it will work for `guide_custom()` and other extension guides too.

R/guides-.R
```diff
@@ -529,11 +528,28 @@ Guides <- ggproto(
...
       )
+      # we'll merge inside legends with same coordinate into same guide box
+      # here, we define the groups of the inside legends
+      if (identical(positions[i], "inside")) {
+        positions[i] <- paste(
+          "inside",
+          paste(attr(.subset2(grobs, i), "inside_position"), collapse = "_"),
+          sep = "_"
+        )
+      }
```
teunbrand, line 542: Instead of reading out the attributes here, you can deduce what the inside position is by using outside this loop:

```
default_inside_position <- calc_element("legend.position.inside", theme) %||%
  valid.just(calc_element("legend.justification.inside"))
```
and this inside the loop:
```
calc_element("legend.position.inside", params[[i]]$theme) %||% default_inside_position
```

This would allow us to skip editing guide-colorbar.R and guide-legend.R.

R/guides-.R
```diff
@@ -546,8 +562,10 @@ Guides <- ggproto(
     # Determine default direction
     direction <- switch(
       position,
-      inside = , left = , right = "vertical",
-      top = , bottom = "horizontal"
+      left = , right = "vertical",
+      top = , bottom = "horizontal",
+      # for all inside guide legends
+      "vertical"
```
teunbrand, line 568:
```suggestion
      top = , bottom = "horizontal", "vertical"
```

R/guides-.R
```diff
@@ -529,11 +528,28 @@ Guides <- ggproto(
...
+          sep = "_"
+        )
+      }
     }
-    keep <- !vapply(grobs, is.zero, logical(1))
+
+    # move inside legends to the last
+    positions <- factor(positions, 
+      levels = c(.trbl, unique(positions[startsWith(positions, "inside")]))
+    )
```
teunbrand, line 548: Why do these need to be last?

R/guides-.R
```diff
@@ -566,19 +584,27 @@ Guides <- ggproto(
...
       global_yjust  <- global_just[2]
+      x <- inside_position[1]
+      y <- inside_position[2]
+      global_margin <- margin()
+    } else {
+      # Global justification of the complete legend box
+      global_just <- paste0("legend.justification.", position)
+      global_just <- valid.just(calc_element(global_just, theme))
+      x <- global_xjust  <- global_just[1]
+      y <- global_yjust  <- global_just[2]
```
teunbrand, line 607: I think we might not even have to alter the logic of `Guides$package_box()` if we just manage everything well in `Guides$assemble()`.

R/guides-.R
```diff
@@ -529,11 +528,28 @@ Guides <- ggproto(
...
+
+    # move inside legends to the last
+    positions <- factor(positions, 
+      levels = c(.trbl, unique(positions[startsWith(positions, "inside")]))
+    )
+    keep <- !vapply(grobs, is.zero, logical(1), USE.NAMES = FALSE)
+
+    # we grouped the legends by the positions
+    # for inside legends, they'll be splitted by the actual inside coordinate
     split(grobs[keep], positions[keep])
```
teunbrand, line 553: I think for this PR it might be easier to lift the reponsibilities of managing the guide position from the `Guides$draw()` method to the `Guides$assemble()` method. The assemble method can then properly forward the positions and 'position_inside' information to the `Guides$package_box()` method.

R/guides-.R
```diff
@@ -529,11 +528,28 @@ Guides <- ggproto(
...
-        theme = theme, position = as.character(positions[i]),
+        theme = theme, position = positions[i],
         direction = directions[i], params = params[[i]]
       )
+      # we'll merge inside legends with same coordinate into same guide box
+      # here, we define the groups of the inside legends
+      if (identical(positions[i], "inside")) {
+        positions[i] <- paste(
+          "inside",
+          paste(attr(.subset2(grobs, i), "inside_position"), collapse = "_"),
```
teunbrand, line 539: We shouldn't use `.subset2()` unless necessary, which applies to the rest of the PR too.

teunbrand (changes requested):
Thanks for getting the PR to work!
My overall feeling about this PR is that it is a little bit on the more invasive side, whereas I think it could be more surgical than it currently is.
If we lift the responsibility of managing the positions into `Guides$assemble()`, we should have to change less code elsewhere.

R/guides-.R
Yunuuuu: I found the original use levels of `c(.trbl, "inside")`, I don't know whether we must set the factor levels

teunbrand: Sorry this was just curiousity on my part, it isn't something that should be changed

R/guides-.R
Yunuuuu: Dose this means we should allow user to set numeric value in the `position` argument of `guide_legend` and `guide_colorbar`?

teunbrand: No I still think the position string left/right/top/bottom/inside is separate from the position location when legends are placed inside. I imagine you can set the position location with `guide_legend(theme = theme(legend.inside.position)` as in your example.

Yunuuuu: In this case, if we want users to set the inside legend position via the theme argument, we would need to modify the `GuideLegend` or `GuideColorbar` to ensure that the elements include the `legend.inside.position`.

Yunuuuu: The same with the `GuideCustom` class.

teunbrand: There was a reason I had to divorce numerical legend positions from string positions, although at the moment I can't quickly recall what that reason was. We do not need to guide to register the `legend.inside.position` as an element if it is not needed to render the guide itself. It is only needed to compose the guides into guide boxes. `position` is an exception, and implemented as a paramter, because axis guides absolutely need to know which side of the plotting panel they're on. Consistency then dictated that it should also be a parameter to legends, colourbars etc.

R/guide-legend.R
Yunuuuu: All modifications to guide_legend and guide_colorbar have now been reverted.

R/guides-.R
Yunuuuu: Thanks for the details, the position will now be managed in `Guides$package_box()`.

R/guides-.R
Yunuuuu: has been reverted to the original.

R/guides-.R
Yunuuuu: Solved

R/guides-.R
Yunuuuu: Solved

R/guides-.R
Yunuuuu: Solved.

Yunuuuu:
Thank you for your thorough review and all the valuable advice.

Yunuuuu:
After writing some examples, I’ve realized that it’s necessary to allow users to set the inside justification for each inside legend. This would be beneficial for splitting legends based on their justifications, as users may be confused about which justification will be applied if we only split legends with the same inside position into groups.

Yunuuuu:
I’m unable to run the tests on my local machine, as there are always some errors. I’m unable to update the test snapshots.

R/guides-.R
```diff
@@ -485,15 +485,76 @@ Guides <- ggproto(
...
+        # the actual inside position and justification can be set in each guide
+        # by `theme` argument, here, we won't use `calc_element()` which will
+        # use inherits from `legend.justification` or `legend.position`, we only
+        # follow the inside elements from the guide theme
+        inside_just <- params[[i]]$theme[["legend.justification.inside"]]
+        if (is.null(inside_just)) {
+          inside_justs[[i]] <- default_inside_just
+        } else {
+          inside_justs[[i]] <- valid.just(inside_just)
+        }
```
teunbrand, line 529:
```suggestion
        inside_justs[[i]] <- valid.just(inside_just %||% default_inside_just)
```

R/guides-.R
```diff
@@ -485,15 +485,76 @@ Guides <- ggproto(
...
+    default_inside_position <- calc_element(
+      "legend.position.inside", theme
+    )
+    inside_justs <- inside_positions <- vector("list", length(positions))
+
+    # we grouped the legends by the positions, for inside legends, they'll be
+    # splitted by the actual inside coordinate
+    groups <- positions
+    for (i in seq_along(positions)) {
+      if (identical(positions[i], "inside")) {
```
teunbrand, line 519:
```suggestion
    for (i in seq_along(positions)[positions == "inside"]) {
```

R/guides-.R
```diff
@@ -485,15 +485,76 @@ Guides <- ggproto(
...
+    indices <- locs$loc
+    grobs <- vec_chop(grobs, indices = indices)
+    names(grobs) <- locs$key
+
+    # for each group, they share the same locations,
+    # so we only extract the first one of `positions` and `inside_positions`
+    first_indice <- lapply(indices, `[[`, 1L)
+    positions <- vec_chop(positions, indices = first_indice)
+    inside_positions <- vec_chop(inside_positions, indices = first_indice)
+    inside_justs <- vec_chop(inside_justs, indices = first_indice)
```
teunbrand, line 557: I think this could be simplified by tracking positions/inside_positions/inside_justs in a data.frame.
`vec_group_loc()` treats rows that are the same as the same group, and the `locs$key` should allow you to skip having the subset the first index.

R/plot-build.R
```diff
@@ -508,19 +508,29 @@ table_add_legends <- function(table, legends, theme) {
...
+        name = paste("guide-box-inside", i, sep = "-")
+      )
+    }
+  } else { # to be consistent with original gtable layout
+    table <- gtable_add_grob(
+      table, zeroGrob(), clip = "off",
+      t = place$t, b = place$b, l = place$l, r = place$r,
+      name = "guide-box-inside"
+    )
+  }
```
teunbrand, line 533: I think it should be possible to avoid this by combining all packaged inside guides into a single grob.

R/guides-.R
```diff
@@ -502,43 +563,47 @@ Guides <- ggproto(
...
                   params = self$params,
                   guides = self$guides) {
-    positions <- vapply(
+    positions <- positions %||% vapply(
       params,
-      function(p) p$position[1] %||% default_position,
-      character(1)
+      function(p) p$position[1] %||% "right",
+      character(1), USE.NAMES = FALSE
     )
```
teunbrand, line 582: Can't we assume here that the `Guides$draw()` method receives a well-formed `positions` argument, so we don't have to fall back here? You could even change `positions = NULL` to `positions` in the arguments to signal that it is a required argument.

R/guides-.R
```diff
@@ -502,43 +563,47 @@ Guides <- ggproto(
...
+    grobs
   },
 
-  package_box = function(grobs, position, theme) {
-
+  # here, we put `inside_position` and `inside_just` in the last, so that it
+  # won't break current implement of patchwork, which depends on the top three
+  # arguments to collect guides
+  package_box = function(grobs, position, theme,
+                         inside_position = NULL, inside_just = NULL) {
```
teunbrand, line 606: I still think we could prevent having to edit this method.
We can just make a `theme` for every iteration that has the correct `inside_position` and `inside_just` baked in.

R/guides-.R
```diff
@@ -502,43 +563,47 @@ Guides <- ggproto(
...
-
-    directions <- rep(direction %||% "vertical", length(positions))
     if (is.null(direction)) {
-      directions[positions %in% c("top", "bottom")] <- "horizontal"
+      directions <- ifelse(
+        positions %in% c("top", "bottom"), 
+        "horizontal", "vertical"
+      )
+    } else {
+      directions <- rep(direction, length(positions))
```
teunbrand, line 589: I'm not understanding this change. The `else` branch repeats `NULL` some times, which remains `NULL` because it has 0-length?

teunbrand (commented):
Thanks for all the changes, Yunuuuu!
I apologise but I still have a few more to suggest. I'd be happy to help out trying to find a solution for not editing `Guides$package_box()` as much as is reasonable.

R/guides-.R
teunbrand: I'm happy to explore this if you feel this gets stuck

R/guides-.R
Yunuuuu: Here, in this loop, I tried to keep the the length of the `inside_justs` variable so that we can use `keep` variable to subset this.

R/guides-.R
Yunuuuu: Thanks for your suggestions! After applying them, the code looks much cleaner and more elegant.

R/guides-.R
Yunuuuu: Thanks for your suggestions! Should be fixed

R/plot-build.R
Yunuuuu: fixed! Thanks for your suggestions!

R/guides-.R
Yunuuuu: fixed! Thanks for your suggestions!

R/guides-.R
Yunuuuu: My apology, should be fixed

Yunuuuu:
Thanks for your suggestions! the code looks much cleaner.

> I'm happy to do this for you if you like.

Thank you! I need your help to update the test snapshots

R/guides-.R
teunbrand: That will continue to be the case with the proposed change.
If you find it easier to read, we can also use:

```r
for (i in which(positions == "inside")) {
```

Yunuuuu: Sorry for this miss

teunbrand (approved):
This looks a lot more self-contained than before so thank you for putting up with all the comments I had and putting all the work into this PR. In principle I think this PR is in the right shape but I have two brief points:

* If you like, you can add a news bullet to the NEWS.md file with a short description of what is new, what issue it solved and your github handle.
* Would you mind if I make a few small tweaks here and there, mostly for readability purposes?

Yunuuuu:
> Would you mind if I make a few small tweaks here and there, mostly for readability purposes?
Feel free to make any tweaks you think would improve readability. I appreciate your help!

Yunuuuu:
>If you like, you can add a news bullet to the NEWS.md file with a short description of what is new, what issue it solved and your github handle.

I'll add this

Yunuuuu:
Thanks for the thoughtful guidance and review process. Everything works perfectly on my end!

teunbrand:

neellab-umd:
This is ability to place legends separately inside a plot is an excellent improvement!  Thank you both so much for making it happen!  Do you know when it will go live?

teunbrand:
Well you can download the development version now to try it out, instructions are in the readme. It will probably be sent to CRAN a few months into 2025.

neellab-umd:
Thank you!  That is great.  This really helps for placing legends around data in graphs and especially placing them around geographic features on maps.

PR #6139, opened 2024-10-15
Treatment options for zeroes in histograms

teunbrand:
This PR aims to fix #3449.

Briefly, it adds a `keep.zeroes` argument to `stat_bin()` that controls what happens to zero-count bins.
Rather than `trim = TRUE/FALSE`, I think a third option to only preserve the middle zero-count bins is useful.
Reprex from issue, demonstrating all 3 options:

``` r
devtools::load_all("~/packages/ggplot2")
#> ℹ Loading ggplot2
set.seed(9416)
data <- data.frame(a = rchisq(1000, df = 3),
                   b = rchisq(1000, df = 1),
                   c = rchisq(1000, df = 10)) |>
  tidyr::gather()

p <- ggplot(data, aes(x = value)) +
  facet_wrap(~ key, ncol = 1)

p + geom_histogram(colour = 'red', keep.zeroes = "all") +
  ggtitle('keep.zeroes = "all"')
#> `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
```

``` r
p + geom_histogram(colour = 'red', keep.zeroes = "none") +
  ggtitle('keep.zeroes = "none"')
#> `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
```

Please note the absence of a gap in the 3rd panel at the bottom right.

``` r
p + geom_histogram(colour = 'red', keep.zeroes = "inner") +
  ggtitle('keep.zeroes = "inner"')
#> `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
```

<sup>Created on 2024-10-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #6212, opened 2024-12-04
Sanitise bin calculations

### LINKED ISSUE #6207: `geom_bin_2d()` should recognise 0-width data
In the example below, the computation goes wrong because in the y-direction we don't have any width.

``` r
library(ggplot2)
df <- data.frame(x = 1:5, y = 1)

ggplot(df, aes(x, y)) +
  geom_bin_2d()
#> Warning: Computation failed in `stat_bin2d()`.
#> Caused by error in `bin2d_breaks()` at ggplot2/R/stat-bin2d.R:64:5:
#> ! `origin` must be a number, not `NaN`.
```
<sup>Created on 2024-11-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I think the reasonable thing to do in such a case is to choose 1 bin for the y-direction with 1 width.

teunbrand:
This PR aims to fix #6207.

However, the linked issue is merely an occasion to sanitise bin calculations across ggplot2.
Most binning functions in ggplot2 acted similar, but not quite identical. An example of this is `stat_bin(boundary)` has replaced the `stat_bin(origin)` argument since ggplot2 2.1.0, but `stat_bin_2d(origin)` continued to exist.

The main changes in this PR:
*  We have a new binning orchestrator `compute_bins()`, which is now used by all binning functionality. It is modelled after the `stat_bin()` functionality, as this was the most polished version. `stat_bin()` is able to deal with 0-width data elegantly, so this fixed #6207.
* All binning stats now use `boundary`/`center` instead of `origin`. They were never formal arguments to most binning stats.
* I also realised that `stat_bin(keep.zeroes)` introduced in #6139 was fulfilling the same/very similar role as `stat_bin_2d(drop)`. However, `stat_bin(drop)` has been deprecated (also since 2.1.0). This PR renames `stat_bin(keep.zeroes)` to `stat_bin(drop)`, effectively resurrecting a formerly deprecated argument. There is no need for backward compatibility here since `keep.zeroes` was introduced in this cycle of development.
* `StatBin2d` is now a subclass of `StatSummary2d`, which prevents duplicated code.
* `stat_bin(bins)` can be a function now, purely for symmetry with the `stat_bin(breaks, binwidth)` arguments that can also be a function. You can use this to do `bins = nclass.Sturges` for example.

Reprex from the linked issue:

``` r
#> ℹ Loading ggplot2
df <- data.frame(x = 1:5, y = 1)

ggplot(df, aes(x, y)) +
  geom_bin_2d()
#> `stat_bin2d()` using `bins = 30`. Pick better value `binwidth`.
```

<sup>Created on 2024-12-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/stat-summary-2d.R
```diff
@@ -92,31 +92,50 @@ StatSummary2d <- ggproto("StatSummary2d", Stat,
...
+
   compute_group = function(data, scales, binwidth = NULL, bins = 30,
                            breaks = NULL, origin = NULL, drop = TRUE,
-                           fun = "mean", fun.args = list()) {
-    origin <- dual_param(origin, list(NULL, NULL))
-    binwidth <- dual_param(binwidth, list(NULL, NULL))
-    breaks <- dual_param(breaks, list(NULL, NULL))
+                           fun = "mean", fun.args = list(),
+                           boundary = 0, closed = NULL, center = NULL) {
     bins <- dual_param(bins, list(x = 30, y = 30))
```
teunbrand, line 113: Ideally this is also done in `setup_params()`, but then the default argument to `compute_group()` should be `bins = list(x = 30, y = 30)`, which causes a test to complain that there is a mismatch between `compute_group()` defaults and `stat_*()` constructor defaults.

tests/testthat/test-stat-bin.R
```diff
@@ -172,13 +172,10 @@ comp_bin <- function(df, ...) {
 
 test_that("inputs to binning are checked", {
   dat <- data_frame(x = c(0, 10))
-  expect_snapshot_error(comp_bin(dat, breaks = letters))
-  expect_snapshot_error(bin_breaks_width(3))
-  expect_snapshot_error(comp_bin(dat, binwidth = letters))
-  expect_snapshot_error(comp_bin(dat, binwidth = -4))
-
-  expect_snapshot_error(bin_breaks_bins(3))
```
teunbrand, line 180: No longer testing `bin_breaks_width(3)` and `bin_breaks_bins(3)` because these are always called from inside `compute_bins()`, which protects against improper ranges.

arcresu:
Thanks for this - it'll be great to have the bin handing unified!

I noticed that the documentation of the 2D binning layers has been a bit neglected and maybe this would be a good opportunity to bring them up to date? Specifically:

- The existing docs and examples don't really explain the API enabled behind the scenes by `dual_param()`. We could explain what happens when giving a scalar (effectively recycled for x and y) vs. named vector/list vs. unnamed vector of length 2. Possibly in the details section rather than repeating for each param?
- `StatBin2d$compute_group(breaks)` was already supported, but `stat_bin_2d(breaks)` wasn't documented. Especially now that you've brought parity to the accepted binning params, should they also be added to the docs and usage for the 2D layer constructors?

R/bin.R
```diff
@@ -128,6 +116,56 @@ bin_breaks_bins <- function(x_range, bins = 30, center = NULL,
...
+    breaks <- allow_lambda(breaks)
+    if (is.function(breaks)) {
+      breaks <- breaks(x)
+    }
+    if (is.scale(scale) && !scale$is_discrete()) {
+      breaks <- scale$transform(breaks)
+    }
+    check_numeric(breaks)
+    bins <- bin_breaks(breaks, closed)
+    return(bins)
```
thomasp85, line 136: Should we warn about ignoring the other arguments here?

thomasp85 (approved):
LGTM - great work

R/bin.R
teunbrand: `bins` usually gets prepopulated which results in false positive warnings. We can warn about the other arguments, but I think it'd be slightly inconsistent.

thomasp85: ok - let's leave it as is

PR #6214, opened 2024-12-04
delete test-data.R

teunbrand:
The test file only runs prior to R 4.0. Since ggplot2 now depends on R >= 4.0, this test has become obsolete.


yutannihilation (approved):

teunbrand:
Thanks for the review!

PR #6215, opened 2024-12-04
Vignette: discuss optional string input

### LINKED ISSUE #6208: Vignette ggplot2-in-packages does not cover programmatically passing NULL to aes()
I found a problem with the ggplot2-in-packages vignette.

I expected it to cover how to fully replace all behaviors from the now-deprecated aes_string() function. However, both of its recommended solutions (the .data pronoun and the embrace operator) have limitations that make them difficult to use to replace aes_string(), and the vignette does not cover how to work around these limitations.

Here is some test data for demonstrations:

```r
set.seed(1)
plotdat <- data.frame(height=rnorm(10,mean=160,sd=10),weight=rnorm(10,mean=70,sd=10))
plotdat$sex <- sample(c('F','M'),10,replace=T)
plotdat$age <- round(rnorm(10,mean=40,sd=20))
plotdat$employment <- sample(c('Employed','Unemployed','Retired'),10,replace=T)
```

The problem with using the .data pronoun is that the .data pronoun cannot handle situations where an aesthetic is set to NULL to indicate it should not be used:

```r
plot_data_pronoun <- function(dat,xvar,yvar,colorvar=NULL,shapevar=NULL) {
  myplot <- ggplot(dat,aes(x=.data[[xvar]],y=.data[[yvar]],color=.data[[colorvar]],shape=.data[[shapevar]])) +
    geom_point()
  return(myplot)
}

plot_data_pronoun(plotdat,'weight','height',colorvar='employment',shapevar=NULL)

# This fails with:
# Error in `geom_point()`:
#  ! Problem while computing aesthetics.
# Error occurred in the 1st layer.
# Caused by error in `.data[[NULL]]`:
#  ! Must subset the data pronoun with a string, not `NULL`.
```

The embrace operator can handle NULL, but the problem with using the embrace operator is that it does not work when the variable name (or NULL) that will be used for an aesthetic is determined when the code runs and cannot be written into the function call:

```r
plot_embrace <- function(dat,xvar,yvar,colorvar=NULL,shapevar=NULL) {
  myplot <- ggplot(dat,aes(x={{ xvar }},y={{ yvar }},color={{ colorvar }},shape={{ shapevar }})) +
    geom_point()
  return(myplot)
}

whichvar <- sample(c('sex','employment'),1)
plot_embrace(plotdat,weight,height,colorvar=age,shapevar=whichvar) # This produces an incorrect plot that does not use the column specified by "whichvar"
```

Both of these scenarios could be handled easily and seamlessly by aes_string():

```r
plot_aes_string <- function(dat,xvar,yvar,colorvar=NULL,shapevar=NULL) {
  myplot <- ggplot(dat,aes_string(x=xvar,y=yvar,color=colorvar,shape=shapevar)) + geom_point()
  return(myplot)
}

# Scenario 1: Passing NULL for an unused aesthetic
plot_aes_string(plotdat,'weight','height',colorvar='employment',shapevar=NULL) # This works

# Scenario 2: The variable used for shape isn't determined until the code runs
whichvar <- sample(c('sex','employment'),1)
plot_aes_string(plotdat,'weight','height',colorvar='age',shapevar=whichvar) # This works

# Scenario 3: Both things at once
whichvar <- NULL
plot_aes_string(plotdat,'weight','height',colorvar='age',shapevar=whichvar) # This works
```

Please consider updating the ggplot2-in-packages vignette to demonstrate how tidy evaluation idioms can be used to allow a function which uses ggplot() to handle both of these scenarios.

(I am submitting this as a bug report rather than a StackExchange question because, while I can figure out a duct-tape work around for these issues, I think it is important that the ggplot authors' intended/best practice solution to these issues be included in the official documentation. Or if there is currently no best practice solution that can handle both scenarios, there may be a need to create one to help users move away from aes_string().)

teunbrand:
This PR aims to fix #6208.

If somebody has better advice for dealing with optional character-input, I'd be glad to adjust.


yjunechoe:
Tangentially related, but I wonder whether `rlang::data_sym()` passing `NULL` through could resolve some of this headache:

```r
rlang::data_sym(NULL)
#> NULL
```

Instead of its current behavior:

```r
rlang::data_sym(NULL)
#> Error in `sym()`:
#> ! Can't convert `NULL` to a symbol.
```

Especially because I feel like this "optional character-input" problem has already been solved in tidyselect (with `all_of()`/`any_of()`), so it'd be nice if `data_sym()`/`data_syms()` could let us get something close to tidyselect semantics within tidyeval contexts (like `aes()`) by letting `NULL` pass through to mean an empty (NULL) selection. Ex:

```r
df <- data.frame(x = 1)

# Tidyselect behavior
tidyselect_get <- function(df, col) {
  tidyselect::eval_select(tidyselect::all_of(col), df)
}
tidyselect_get(df, "x")
#> x 
#> 1
tidyselect_get(df, NULL)
#> named integer(0)

# Tidyeval with `data_sym()`
tidyeval_get <- function(df, col) {
  rlang::eval_tidy(rlang::data_sym(col), df)
}
tidyeval_get(df, "x")
#> [1] 1
tidyeval_get(df, NULL)
#> Error in `sym()`:
#> ! Can't convert `NULL` to a symbol.

# Tidyeval with modified `data_sym2()` that passes NULL through
data_sym2 <- \(x) if (!is.null(x)) rlang::data_sym(x)
tidyeval_get2 <- function(df, col) {
  rlang::eval_tidy(data_sym2(col), df)
}
tidyeval_get2(df, "x")
#> [1] 1
tidyeval_get2(df, NULL)
#> NULL
```

teunbrand:
Thanks June, your comments make a lot of sense to me. I'd agree that this having `data_sym()` accept `NULL` would be the most elegant. I don't think ggplot2 should provide a user-exposed wrapper for this, as we'd have to maintain this into perpetuity. Rather, we could petition {rlang} if they'd be interested in this change. That allows us to include this PR as current advice and update the advice if and once we can use `data_sym(NULL)`.

yjunechoe:
> Rather, we could petition {rlang} if they'd be interested in this change.

Yeah, that's the intended build-up of my comment! Would be nice for the FR to be backed by a need in ggplot 😄

thomasp85:
I'm personally partial to advocating embrace for people wanting to build ggplot2 wrapper functions. The issue noted in the original issue (name of column coming as a string) is easily solved:

```r
plot_embrace <- function(dat,xvar,yvar,colorvar=NULL,shapevar=NULL) {
  myplot <- ggplot(dat,aes(x={{ xvar }},y={{ yvar }},color={{ colorvar }},shape={{ shapevar }})) +
    geom_point()
  return(myplot)
}

whichvar <- sample(c('sex','employment'),1)
plot_embrace(plotdat,weight,height,colorvar=age,shapevar=!!sym(whichvar))
```

teunbrand:
I'm happy to recommend the `!!sym()` trick, but it'd be a different approach than this PR, so I'll create a new one.

PR #6112, opened 2024-09-17
More `palette` flexibility

teunbrand:
This PR aims to fix #6064.
Additionally it would also fix #6068 and fix #5843 and #2851.

The PR is a draft, as the {scales} version that powers this PR is not released yet.
It adds a `palette` argument to the default discrete/binned/continuous colour/fill scales, which bypasses the .
This palette is then internally converted with `as_discrete_pal()` or `as_continuous_pal()`.
In the example below, we're using a single palette to populate all 3 types of scales.

``` r
devtools::load_all("~/packages/ggplot2")
#> ℹ Loading ggplot2
packageVersion("scales")
#> [1] '1.3.0.9000'

p <- ggplot(mpg, aes(displ, hwy, colour = cty)) +
  geom_point()

my_pal <- colour_ramp(c("cyan", "deepskyblue", "magenta"))

p + scale_colour_continuous(palette = my_pal)
```

``` r
p + scale_colour_binned(palette = my_pal)
```

``` r
p + scale_colour_discrete(palette = my_pal) +
  aes(colour = drv)
```

<sup>Created on 2024-09-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #6216, opened 2024-12-04
Flexible palettes

### LINKED ISSUE #4910: scale_colour_discrete not showing an example that uses the function
scale_colour_discrete documentation 

https://ggplot2.tidyverse.org/reference/scale_colour_discrete.html

does not show an actual example containing scale_colour discrete.

### LINKED ISSUE #5843: Could support the use of function factories like `scales::pal_hue()` and `scales::pal_viridis()`?
It'd be great if it was possible to do this for all of the non-positional scales, especially colour/fill
```
library(tidyverse)
library(palmerpenguins)

penguins |>
  ggplot() +
  geom_point(aes(x = flipper_length_mm,
                 y = body_mass_g,
                 col = species)) +
  scale_colour_manual(values = scales::pal_hue())

penguins |>
  ggplot() +
  geom_point(aes(x = flipper_length_mm,
                 y = body_mass_g,
                 col = bill_depth_mm)) +
  scale_colour_gradientn(colours = scales::pal_viridis())

penguins |>
  ggplot() +
  geom_point(aes(x = flipper_length_mm,
                 y = body_mass_g,
                 col = bill_depth_mm)) +
  scale_colour_stepsn(colours = scales::pal_viridis())
```

### LINKED ISSUE #6064: Feature request: improve palette interface
The current interface for interacting with palettes is by wrapping each one in a scale.
I dislike this for the following reasons:
* It expanded the number of scale functions in ggplot2 beyond modesty (do we really need continuous, discrete and binned variants of every palette?).
* It encourages developers of palette packages to model their interface after ggplot2's interface.

It is no good reverting the scales we already have, but we may yet prevent wildgrowth of future scale functions. The proposal is to make the `palette` argument in scales more accessible to users, by for example:
* Directly expose the `palette` argument in some scale functions
* This `palette` argument would then accept:
    * A function, as it does currently
    * A keyword, like "viridis" or "okabe-ito"
    * A vector of values that can be translated to a function
* Abstract away the distinction between discrete and continuous palettes.
    * A user shouldn't have to worry about this
    * Auto-translate discrete palettes to continuous ones
    * Auto-translate continuous palettes to discrete ones

### LINKED ISSUE #6289: Fallback for colour scales
This issue was identified in #6287.

Some packages use `scale_colour_continuous(low = "blue", high = "red")`. This is only valid when end-users haven't touched the default `"ggplot2.continuous.colour"` options. The pattern these packages use are thus fragile and instead they should use `scale_colour_gradient()`.
Likewise for `scale_colour_discrete()` and `scale_colour_hue()` and the `fill` aesthetic equivalents.

However, because the problem is quite ubiquitous (39 packages), we should have fallbacks in place.
I'll suggest that we use a grumpy fallback that always throws a warning, as the usage remains incorrect.

teunbrand:
This PR aims to fix #6064 and fix #4910 and fix #6289 and fix #5843 and replaces #6112.

Briefly, it attempts to cast non-null `palette` arguments with `as_discrete_pal()`/`as_continuous_pal()`. These are currently shims.

This PR replaces the older one because I could implemented more cleanly now that #5946 is merged.


teunbrand:
Includes fix for #6289, since it'd touch the same code.

R/scale-colour.R
```diff
@@ -57,99 +63,198 @@
...
 #' @export
-scale_colour_continuous <- function(..., aesthetics = "colour",
+scale_colour_continuous <- function(..., palette = NULL, aesthetics = "colour",
                                     guide = "colourbar", na.value = "grey50",
                                     type = getOption("ggplot2.continuous.colour")) {
 
-  if (!is.null(type)) {
+  has_old_args <- any(names(enexprs(...)) %in% c("low", "high"))
+
+  if (has_old_args || !is.null(type) && is.null(palette)) {
```
thomasp85, line 95:
```suggestion
  if (has_old_args || (!is.null(type) && is.null(palette))) {
```

R/scale-colour.R
```diff
@@ -57,99 +63,198 @@
...
 #' @export
-scale_fill_continuous <- function(..., aesthetics = "fill", guide = "colourbar",
+scale_fill_continuous <- function(..., palette = NULL, aesthetics = "fill", guide = "colourbar",
                                   na.value = "grey50",
                                   type = getOption("ggplot2.continuous.fill")) {
 
-  if (!is.null(type)) {
+  has_old_args <- any(names(enexprs(...)) %in% c("low", "high"))
+
+  if (has_old_args || !is.null(type) && is.null(palette)) {
```
thomasp85, line 117:
```suggestion
  if (has_old_args || (!is.null(type) && is.null(palette))) {
```

R/scale-colour.R
```diff
@@ -57,99 +63,198 @@
...
 #' @rdname scale_colour_continuous
-scale_colour_binned <- function(..., aesthetics = "colour", guide = "coloursteps",
+scale_colour_binned <- function(..., palette = NULL, aesthetics = "colour", guide = "coloursteps",
                                 na.value = "grey50",
                                 type = getOption("ggplot2.binned.colour")) {
-  if (!is.null(type)) {
+
+  has_old_args <- any(names(enexprs(...)) %in% c("low", "high"))
+
+  if (has_old_args || !is.null(type) && is.null(palette)) {
```
thomasp85, line 139:
```suggestion
  if (has_old_args || (!is.null(type) && is.null(palette))) {
```

R/scale-colour.R
```diff
@@ -57,99 +63,198 @@
...
 #' @export
 #' @rdname scale_colour_continuous
-scale_fill_binned <- function(..., aesthetics = "fill", guide = "coloursteps",
+scale_fill_binned <- function(..., palette = NULL, aesthetics = "fill", guide = "coloursteps",
                               na.value = "grey50",
                               type = getOption("ggplot2.binned.fill")) {
-  if (!is.null(type)) {
+  has_old_args <- any(names(enexprs(...)) %in% c("low", "high"))
+
+  if (has_old_args || !is.null(type) && is.null(palette)) {
```
thomasp85, line 160:
```suggestion
  if (has_old_args || (!is.null(type) && is.null(palette))) {
```

thomasp85 (changes requested):
I know the proposed changes don't practically affect the evaluation but I prefer to collect logical tests that define a specific condition (if that makes sense)

R/scale-colour.R
```diff
@@ -57,98 +63,198 @@
...
+#' p
+#'
+#' # Restoring the previous theme
+#' theme_set(old)
+scale_colour_discrete <- function(..., palette = NULL, aesthetics = "colour", na.value = "grey50",
+                                  type = getOption("ggplot2.discrete.colour")) {
+
+  has_old_args <- any(names(enexprs(...)) %in% c("h", "c", "l", "h.start", "direction"))
+
+  if (has_old_args || !is.null(type) && is.null(palette)) {
```
thomasp85, line 227: I know I'm being pedantic here but please make the same changes to the logic tests here and below :-)

teunbrand: Ah my bad, I should've spotted these!

thomasp85 (approved):
LGTM

PR #6218, opened 2024-12-05
Tweaks to `map_data()`/fortify.map()`

### LINKED ISSUE #3717: better documentation for map related functions?

### LINKED ISSUE #3816: Clarify the current status of fortify()

### LINKED ISSUE #5469: Putting up `fortify()` methods for adoption?
Currently, ggplot2 hosts some `fortify()` methods for objects from packages that do not come with the standard R distribution.
Examples of these include the fortify methods that live in `fortify-spatial.R`, `fortify-multcomp.R` and `fortify-map.R`.
To my understanding, these methods exist in ggplot2 because before the advent of extension packages, these *had* to live in ggplot2. However, ggplot2 has become highly extensible now, which means that specialised functionality could live elsewhere.
Therefore, it might be worth exploring whether we should put these methods up for adoption into extension packages.

It might be a good idea to include @terrytangyuan in the conversation for his thoughts and maybe probe if {ggfortify} has any interest in hosting these methods.

teunbrand:
This PR aims to fix #3816, close #5469 and fix #3717.

With regards to #3816, this PR checks off the last bullet that `fortify.map()` is deprecated.
WIth regards to #5469, this PR makes the issue obsolete, since we have deprecated all the functions we were considering putting up for adoption.
With regards to #3717 this PR fixes the mislinked `map_data(map)` argument.

### DIFF
--- a/R/fortify-map.R
+++ b/R/fortify-map.R
@@ -1,5 +1,8 @@
 #' Fortify method for map objects
 #'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
 #' This function turns a map into a data frame that can more easily be
 #' plotted with ggplot2.
 #'
--- a/R/fortify-map.R
+++ b/R/fortify-map.R
@@ -24,6 +24,9 @@
 #'   geom_polygon(aes(group = group), colour = "white")
 #' }
 fortify.map <- function(model, data, ...) {
+  lifecycle::deprecate_warn(
+    "3.6.0", I("`fortify(<map>)`"), "map_data()"
+  )
   df <- data_frame0(
     long = model$x,
     lat = model$y,
--- a/R/fortify-map.R
+++ b/R/fortify-map.R
@@ -46,10 +46,10 @@ fortify.map <- function(model, data, ...) {
 #' for plotting with ggplot2.
 #'
 #' @param map name of map provided by the \pkg{maps} package. These
-#'   include [maps::county()], [maps::france()],
-#'   [maps::italy()], [maps::nz()],
-#'   [maps::state()], [maps::usa()],
-#'   [maps::world()], [maps::world2()].
+#'   include [`"county"`][maps::county], [`"france"`][maps::france],
+#'   [`"italy"`][maps::italy], [`"nz"`][maps::nz],
+#'   [`"state"`][maps::state], [`"usa"`][maps::usa],
+#'   [`"world"`][maps::world], or [`"world2"`][maps::world2].
 #' @param region name(s) of subregion(s) to include. Defaults to `.` which
 #'   includes all subregions. See documentation for [maps::map()]
 #'   for more details.
--- a/R/fortify-map.R
+++ b/R/fortify-map.R
@@ -88,7 +88,11 @@ map_data <- function(map, region = ".", exact = FALSE, ...) {
   map_obj <- maps::map(map, region, exact = exact, plot = FALSE, fill = TRUE, ...)
 
   if (!inherits(map_obj, "map")) {
-    return(fortify(map_obj))
+    cli::cli_abort(c(
+      "{.fn maps::map} must return an object of type {.cls map}, not \\
+      {obj_type_friendly(map_obj)}.",
+      i = "Did you pass the right arguments?"
+    ))
   }
 
   df <- data_frame0(
--- a/R/fortify-map.R
+++ b/R/fortify-map.R
@@ -80,7 +80,23 @@ fortify.map <- function(model, data, ...) {
 map_data <- function(map, region = ".", exact = FALSE, ...) {
   check_installed("maps", reason = "for `map_data()`.")
   map_obj <- maps::map(map, region, exact = exact, plot = FALSE, fill = TRUE, ...)
-  fortify(map_obj)
+
+  if (!inherits(map_obj, "map")) {
+    return(fortify(map_obj))
+  }
+
+  df <- data_frame0(
+    long  = map_obj$x,
+    lat   = map_obj$y,
+    group = cumsum(is.na(map_obj$x) & is.na(map_obj$y)) + 1,
+    order = seq_along(map_obj$x),
+    .size = length(map_obj$x)
+  )
+
+  names <- lapply(strsplit(map_obj$names, "[:,]"), "[", 1:2)
+  names <- vec_rbind(!!!names, .name_repair = ~ c("region", "subregion"))
+  df[names(names)] <- vec_slice(names, df$group)
+  vec_slice(df, stats::complete.cases(df$lat, df$long))
 }
 
 #' Create a layer of map borders
--- a/R/fortify-spatial.R
+++ b/R/fortify-spatial.R
@@ -1,5 +1,8 @@
 #' Fortify method for classes from the sp package.
 #'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
 #' To figure out the correct variable name for region, inspect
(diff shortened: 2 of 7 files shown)


R/fortify-map.R
```diff
@@ -80,7 +86,23 @@ fortify.map <- function(model, data, ...) {
 map_data <- function(map, region = ".", exact = FALSE, ...) {
   check_installed("maps", reason = "for `map_data()`.")
   map_obj <- maps::map(map, region, exact = exact, plot = FALSE, fill = TRUE, ...)
-  fortify(map_obj)
+
+  if (!inherits(map_obj, "map")) {
+    return(fortify(map_obj))
+  }
```
yutannihilation, line 92: Do you know when `maps::map()` returns a non-`map` object? Maybe we should just ignore, or raise an error rather than forwarding to `fortify()`?

``` r
map <- c("county", "france", "italy", "nz", "state", "usa", "world", "world2")
vapply(
  map,
  \(x) class(maps::map(x, region = ".", exact = FALSE, plot = FALSE, fill = TRUE)),
  FUN.VALUE = character(1L)
)
#> county france  italy     nz  state    usa  world world2 
#>  "map"  "map"  "map"  "map"  "map"  "map"  "map"  "map"
```

<sup>Created on 2024-12-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

yutannihilation (commented):
Thanks for addressing this issue! Looks good in general.

R/fortify-map.R
teunbrand: If a user passes the `namesonly = TRUE` option, the `maps::map()` function does not return a non-map object.
Currently we throw the default error message, which will continue to be the case with this PR.

``` r
x <- maps::map(namesonly = TRUE, plot = FALSE)
str(x)
#>  chr [1:1627] "Aruba" "Afghanistan" "Angola" "Angola:Cabinda" "Anguilla" ...

ggplot2::map_data("world", namesonly = TRUE)
#> Error in `fortify()`:
#> ! `data` must be a <data.frame>, or an object coercible by `fortify()`,
#>   or a valid <data.frame>-like object coercible by `as.data.frame()`.
#> Caused by error in `.prevalidate_data_frame_like_object()`:
#> ! `dim(data)` must return an <integer> of length 2.
```

<sup>Created on 2024-12-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

yutannihilation: Oh, I didn't know this, thanks.

> Happy to throw a more custom error message if you think that is more appropriate!

IMHO, throwing an error is preferable because we can avoid unnecessary `fortify()`. 

How about something like `` `maps::map()` didn't return a map object. Did you pass some unusual options to `map_data()`?  ``? I think this is a corner case, so I don't think it needs to be very user-friendly.

teunbrand: Alright, now throwing the following error:

``` r
#> ℹ Loading ggplot2
map_data("world", namesonly = TRUE)
#> Error in `map_data()`:
#> ! `maps::map()` must return an object of type <map>, not a character
#>   vector.
#> ℹ Did you pass the right arguments?
```

<sup>Created on 2024-12-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

yutannihilation: The error looks good to me!

yutannihilation (approved):

teunbrand:
Thanks for the review!

PR #6220, opened 2024-12-06
Clip qq line

### LINKED ISSUE #6087: Could `geom_qq_line` always stop at the limits, regardless of other settings?
Following on from https://github.com/tidyverse/ggplot2/issues/6081

This would be helpful, as qq_line is always needed in the panel - and then you would have the flexibility to adjust the scale and coord settings without adversely affecting the abline. So, In essence, for the same reasons that it is useful for vline/hline/abline, it is also helpful for qq_line.

An example: say a user wants to have y limits on the range of the y data with zero expand - and also _not_ clip any points in half. They cannot do this as far as I am aware, as the qq_line will now extend past the panel.
``` r
library(tidyverse)
library(palmerpenguins)

# with default clip = off, points on the limits are cut in half  
p <-  penguins |> 
    ggplot(aes(sample = body_mass_g)) +
    geom_qq() +
    geom_qq_line(colour = "red") +
    scale_y_continuous(limits = range(penguins$body_mass_g, na.rm = T), oob = scales::oob_keep, expand = c(0, 0)) +
    theme(plot.margin = margin(50,50,50,50))

p
#> Warning: Removed 2 rows containing non-finite outside the scale range
#> (`stat_qq()`).
#> Warning: Removed 2 rows containing non-finite outside the scale range
#> (`stat_qq_line()`).
```

``` r

# but you cannot turn clip off, as this affects the qq_line
p +
  coord_cartesian(clip = "off")
#> Warning: Removed 2 rows containing non-finite outside the scale range (`stat_qq()`).
#> Removed 2 rows containing non-finite outside the scale range
#> (`stat_qq_line()`).
```

It would also help people new to ggplot2, as they wouldn't get confused when setting limits and the the qq_line disappearing. See example in https://github.com/tidyverse/ggplot2/issues/6081

teunbrand:
This PR aims to fix #6087.

It does so by preserving the `slope` and `intercept` as computed variables and using `geom = "abline"`.

``` r
#> ℹ Loading ggplot2
#> Warning: package 'testthat' was built under R version 4.4.2

ggplot(mpg, aes(sample = displ)) +
  geom_qq() +
  geom_qq_line() +
  ylim(range(mpg$displ)) +
  coord_cartesian(clip = "off")
```

<sup>Created on 2024-12-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85:
I can't believe you are still using the windows graphics device 😛

thomasp85 (approved):
LGTM

teunbrand:
> I can't believe you are still using the windows graphics device

Look man, it is the default GD when using {reprex}. I've my own session setup with ragg. I have an rstudio shortcut that is called 'raggprex' to the following code:

```r
#+ setup, include=FALSE
knitr::opts_chunk$set(dev = "ragg_png")

#+ main
```

Which I can then following with the actual code to reprex. *Sometimes* I forget to use the shortcut. If we could just permanently set the GD to ragg in {reprex}, I would be a happy camper, but until such a time, I'll throw around some pixellated lines every now and then :p

PR #6225, opened 2024-12-10
Limits of binned scales when data has zero width

### LINKED ISSUE #5066: scale_fill_binned throws error if all bins have the same value

teunbrand:
This PR aims to fix #5066.

Briefly, when the default limits of binned scales have zero-width, they are expanded to have 0.1 width (the same as `default_expansion()`).

Reprexes from the linked issue:

``` r
#> ℹ Loading ggplot2

data.frame(x = c("a", "b"), 
           y = c("A", "B")) |> 
  ggplot() + 
  geom_bin_2d(aes(x = x, y = y)) + 
  scale_fill_binned()
```

``` r
ggplot(data.frame(x=1), aes(x=x)) + geom_bar() + scale_x_binned()
```

<sup>Created on 2024-12-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #6227, opened 2024-12-10
Scales expose `aesthetics` argument

### LINKED ISSUE #5841: Should all scales expose the `aesthetics` parameter?
This issue is a follow-up on #2555, where all colour/fill scales exposed their `aesthetics` argument, to allow for non-standard aesthetics in extension packages. The question here is, should other scales, like `scale_linewidth_continuous()` or `scale_shape_manual()` similarly have `aesthetics = "linewidth"` and `aesthetics = "shape"` in their formals? 

The main reason we might want to is so that it is easier for extention packages to implement 'synonyms' of aesthetics (in the same way that `colour` and `fill` are synonymous aesthetics). It might also be beneficial for PRs like #5484 to generalise beyond colours.

The decision to include non-colour scales was deferred to the future in https://github.com/tidyverse/ggplot2/pull/2555#discussion_r187195653, as it would be an API change that requires some thought. I suppose that future has meanwhile arrived.

teunbrand:
This PR aims to fix #5841.

Briefly, following colour/fill scales, all scales now expose the `aesthetics` argument.


thomasp85 (approved):
LGTM

PR #6229, opened 2024-12-10
Explain how to fix fill guide for points with filled shapes

### LINKED ISSUE #6177: Default `pointshape` of 19 can cause confusion with fill guides
Occasionally it can be useful to use filled shapes (i.e. in the range `21:25`) to separately control the colour and fill aesthetics in `geom_point()`. Here's a contrived example using shape 21 that works:

``` r
library(ggplot2)
#> 
#> Attaching package: 'ggplot2'
#> 
#>     is.element

ggplot(head(diamonds, n = 20)) +
  geom_point(aes(carat, price, fill = cut), shape = 21, size = 5)
```

Now if we also map the shape aesthetic, the fill guide draws the keys using its default shape of 19 which doesn't support fill leading to the colours disappearing from the legend:

``` r

ggplot(head(diamonds, n = 20)) +
  geom_point(aes(carat, price, fill = cut, shape = color), size = 5) +
  scale_shape_manual(values = 21:25)
```

It can of course be fixed by overriding the shape to be 21 as in the first example:

``` r

ggplot(head(diamonds, n = 20)) +
  geom_point(aes(carat, price, fill = cut, shape = color), size = 5) +
  scale_fill_ordinal(guide = guide_legend(override.aes = list(shape = 21))) +
  scale_shape_manual(values = 21:25)
```

<sup>Created on 2024-11-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This isn't really a bug because it's behaving as intended, but it's still a bit surprising! Some possible solutions:

  1. Warn when constructing/drawing a guide that has `key_glyph = "point"`, a non-empty/non-constant fill, and an unfilled shape.
  2. Note the behaviour somewhere in documentation.
  3. Do nothing.

Note that even though `draw_key_point()` fills in a default of 19 that doesn't seem relevant in practice because the key is always draw with a non-`NULL` value determined by the theme's `pointshape` if shape isn't mapped for a guide.

<details><summary>Aside: using a key glyph outside shape scale</summary>
Even putting aside filled shapes, if my shape scale is only triangles and squares, should any of the guides really be using circles for the keys? I suspect that by now people are used to thinking of circles as the "neutral" point shape in ggplot.

``` r
library(ggplot2)
#> 
#> Attaching package: 'ggplot2'
#> 
#>     is.element

ggplot(head(diamonds, n = 20)) +
  geom_point(aes(carat, price, shape = cut, colour = color), size = 5) +
  scale_shape_manual(values = 2:6)
```

<sup>Created on 2024-11-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>
</details>

arcresu:
Adds an example for `geom_point` and a small note to the relevant part of the specs vignette so there's a better chance of people finding out how to deal with guides for filled shapes. Closes #6177.

### DIFF
--- a/R/geom-point.R
+++ b/R/geom-point.R
@@ -85,9 +85,9 @@
 #' ggplot(mtcars, aes(wt, mpg)) +
 #'   geom_point(shape = 21, colour = "black", fill = "white", size = 5, stroke = 5)
 #'
-#' # The default shape used in legends is not filled, so you will need to
-#' # override it in the guide in order to see the fill reflected in the legend
-#' ggplot(mtcars, aes(wt, mpg, fill = factor(carb), shape = factor(cyl), colour = factor(vs))) +
+#' # The default shape in legends is not filled, but you can override the shape
+#' # in the guide to reflect the fill in the legend
+#' ggplot(mtcars, aes(wt, mpg, fill = factor(carb), shape = factor(cyl))) +
 #'   geom_point(size = 5, stroke = 1) +
 #'   scale_shape_manual(values = 21:25) +
 #'   scale_fill_ordinal(guide = guide_legend(override.aes = list(shape = 21)))
--- a/R/geom-point.R
+++ b/R/geom-point.R
@@ -85,6 +85,13 @@
 #' ggplot(mtcars, aes(wt, mpg)) +
 #'   geom_point(shape = 21, colour = "black", fill = "white", size = 5, stroke = 5)
 #'
+#' # The default shape used in legends is not filled, so you will need to
+#' # override it in the guide in order to see the fill reflected in the legend
+#' ggplot(mtcars, aes(wt, mpg, fill = factor(carb), shape = factor(cyl), colour = factor(vs))) +
+#'   geom_point(size = 5, stroke = 1) +
+#'   scale_shape_manual(values = 21:25) +
+#'   scale_fill_ordinal(guide = guide_legend(override.aes = list(shape = 21)))
+#'
 #' \donttest{
 #' # You can create interesting shapes by layering multiple points of
 #' # different sizes
--- a/vignettes/ggplot2-specs.Rmd
+++ b/vignettes/ggplot2-specs.Rmd
@@ -251,7 +251,7 @@ ggplot(sizes, aes(size, stroke, size = size, stroke = stroke)) +
   scale_size_identity()
 ```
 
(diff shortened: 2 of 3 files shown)


R/geom-point.R
```diff
@@ -85,6 +85,13 @@
 #' ggplot(mtcars, aes(wt, mpg)) +
 #'   geom_point(shape = 21, colour = "black", fill = "white", size = 5, stroke = 5)
 #'
+#' # The default shape used in legends is not filled, so you will need to
+#' # override it in the guide in order to see the fill reflected in the legend
+#' ggplot(mtcars, aes(wt, mpg, fill = factor(carb), shape = factor(cyl), colour = factor(vs))) +
```
teunbrand, line 90:
```suggestion
#' ggplot(mtcars, aes(wt, mpg, fill = factor(carb), shape = factor(cyl))) +
```
Colour mapping is distracting from the point this example illustrates

R/geom-point.R
```diff
@@ -85,6 +85,13 @@
 #' ggplot(mtcars, aes(wt, mpg)) +
 #'   geom_point(shape = 21, colour = "black", fill = "white", size = 5, stroke = 5)
 #'
+#' # The default shape used in legends is not filled, so you will need to
+#' # override it in the guide in order to see the fill reflected in the legend
```
teunbrand, line 89:
```suggestion
#' # The default shape in legends is not filled, but you can
#' # override the shape in the guide to reflect the fill
```
Just slightly more terse

vignettes/ggplot2-specs.Rmd
```diff
@@ -251,6 +251,8 @@ ggplot(sizes, aes(size, stroke, size = size, stroke = stroke)) +
   scale_size_identity()
 ```
 
+Because most shapes are not filled, mapping the `fill` aesthetic of points is considered an advanced feature. Additional steps may be required to produce sensible results, such as overriding the shape used in the `fill` guide (refer to `geom_point()` for an example of this).
+
```
teunbrand, line 255: I don't think the first sentence clarifies anything.
I think it'd be better to just name the most common situation wherein the shape needs to be overriden, namely when you use a shape mapping with a discrete fill.

teunbrand (changes requested):
Thank your for the PR! I don't mean to be annoying, but I have a some nitpicks I'd like to see altered, see comments.

arcresu:
Thanks for the suggestions!

My thinking with the "advanced feature" warning was just to let people know that some code might assume points are not filled. Documenting that expectation probably would have stopped me from raising this in the first place. I rephrased that sentence; hopefully it's better now.

teunbrand (approved):
Looks great, thank you for the contribution!

PR #6231, opened 2024-12-11
New position adjustment: `position_connect()`

### LINKED ISSUE #6228: Feature request: `position_step()`
We now have `geom_step()`. The 'step' adjustment to the line could in theory work with other geometries as well, notably `geom_area()` or even `geom_ribbon()`. The proposal here is to make a `position_step()` and deprecate `GeomStep` in favour of `GeomLine` + `PositionStep`.

teunbrand:
This PR aims to fix #6228.

It introduces a new position adjustment that can be used to connect points via steps or other shapes.

It is probably best introduced with some examples:

``` r
#> ℹ Loading ggplot2

# Mirroring `geom_step()`
ggplot(head(economics, 20), aes(date, unemploy)) +
  geom_line(position = "connect")
```

``` r
# Making a histogram without bars
ggplot(faithful, aes(waiting)) +
  geom_area(
    stat = "bin", bins = 20, pad = TRUE,
    position = position_connect("mid")
  )
```

``` r

# Using custom connections with a matrix.
# Note that point A at (0, 0) is not included, but point B at (1, 1) is.
zigzag <- cbind(c(0.4, 0.6, 1), c(0.75, 0.25, 1))
x <- seq(0, 1, length.out = 20)[-1]
smooth <- cbind(x, scales::rescale(1 / (1 + exp(-(x * 10 - 5)))))

ggplot(head(economics, 10), aes(date, unemploy)) +
  geom_line(position = position_connect(zigzag), aes(colour = "zigzag")) +
  geom_line(position = position_connect(smooth), aes(colour = "smooth")) +
  geom_point()
```

<sup>Created on 2024-12-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


teunbrand:
Per discussion in the linked issue, here is an example of a KM curve next to ggsurvplot (who I trust have implemented this correctly for KM curves).

``` r
#> ℹ Loading ggplot2
library(survival)
model <- survfit(Surv(time, status) ~ x, data = aml)

df <- broom::tidy(model)

ggplot(df, aes(time, estimate, ymin = conf.low, ymax = conf.high)) +
  geom_step(aes(colour = strata)) +
  geom_ribbon(
    aes(fill = strata), alpha = 0.2,
    position = "connect"
  )
```



``` r
ggsurvfit::ggsurvfit(model) +
  ggsurvfit::add_confidence_interval() +
  theme_gray()
#> Warning: Ignoring unknown labels:
#> • `linetype = "NULL"`
```



<sup>Created on 2024-12-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thomasp85:
I'm a bit on the fence with this, tbh. There is obvious utility in its generality, **but** Position adjustments shouldn't make up new data - it is only for moving existing data around.

Now, you could argue that a straight line is an infinite set of points between two extremes so it is just moving those points around, but I think that would be stretching it a bit.

Conceptually this feels like it belongs in a stat, but that would preclude it from being used together with other stats like in the examples above. However, I feel we did touch on the idea of having a `stat_stack()`at some point which would remove this downside

teunbrand:
That makes sense to me. Once we can chain stats, we can reshape this into a `stat_connect()` or something, so I'll mark this as a draft for now.

teunbrand:
Closing this PR in favour of #6329.

PR #6232, opened 2024-12-11
Coord sf breaks fallback

### LINKED ISSUE #2985: coord_sf() drops axis labels if graticules don't extend all the way to the plot boundary
In ggplot2 3.1.0, `coord_sf()` drops axis labels if graticules don't extend all the way to the plot boundary. This is a regression from 3.0.0. A reprex follows.

``` r
library(ggplot2)

world <- sf::st_as_sf(rworldmap::getMap(resolution = "low"))
ggplot(world) + geom_sf() + theme_bw() + coord_sf()
```

``` r
ggplot(world) + geom_sf() + theme_bw() + coord_sf(expand = FALSE)
```

<sup>Created on 2018-11-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This bug was introduced by me when I wrote code such as the following, to calculate intersection points between the graticules and the plot boundaries:
https://github.com/tidyverse/ggplot2/blob/15ddc222d197f1d4fa596ddb930d46f4819d8db0/R/sf.R#L592

I did not consider that the plot range could extend beyond the physical limits of the earth. I'll have to ponder what the options are. One option might be to make the tolerance configurable. (I hardcoded 0.001.) Another option might be to base the tolerance on the expansion factors used for the x and y scales.

teunbrand:
This PR aims to fix #2985.

Briefly, if the graticule lines are horizontal or vertical, we could infer where the tickmark should be placed. This PR doesn't solve axis labels placement for all projections, but it does fix a common painpoint.

For WGS84, it solves the problem:

``` r
#> ℹ Loading ggplot2

world <- sf::st_as_sf(rworldmap::getMap(resolution = "low"))
p <- ggplot(world) + 
  geom_sf()
p
```

For the Robinson projection, it can place the y-axis labels and a single label at 0 degrees.

``` r
p + coord_sf(crs = "+proj=robin")
```

For this projection, no labels can be placed because all graticule lines are curved.

``` r
p + coord_sf(crs = "+proj=laea +lat_0=52 +lon_0=10 +x_0=4321000 +y_0=3210000")
```

The IGH projection only gets y-axis labels because all x-graticule lines are not straight.

``` r
p + coord_sf(crs = "+proj=igh")
```

<sup>Created on 2024-12-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM, though the situation with axes and sf continues to bring me pain

teunbrand:
> the situation with axes and sf continues to bring me pain

Me too, I wish there was an sf-based solution that would just work for all types of projections

edzer:
Looks great, thanks - I'll try to dig into this and see how we can be more helpful from the side of `sf` in this respect (and other packages plotting maps with R).

teunbrand:
That would be amazing to have @edzer! When I was trying to implement in-panel placement of axes, the problem I kept running into was not too much the ticks/label placement, but the axis line placement. We'd have to know the outline/extent of a map to gracefully place an axis. If that were easy to get from {sf} for most projections that'd really make a big dent in the larger issue.

PR #6237, opened 2024-12-12
Encourage specifying `breaks` when giving labels as a vector

### LINKED ISSUE #5208: scale_shape_manual mislabels if label argument is given and names in values are not alphabetical
I found a problem in scale_(shape)_manual, presumably others as well, where when the alphabetical sorting of values happens automatically the labels argument does not get sorted. See my example below, where A, B, C have the shapes i want but in the legend As get labeled as Cs and Cs get labeled as As. 

I expected that the values argument allow me to define which shapes i want for each of my string-values and the label argument would allow me to give more human readable versions of these strings and therefore any changes in the order of values gets applied to the labels as well.  I know i can fix this, by either respecting alphabetical order or defining the offending variables as factors/ordered but I'd prefer that i didn't have to if I don't care what order the legend is in.

Most importantly : This should certainly not happen without warning, as it makes the creation of wrongly labelled plots very easy.

``` r
library(tidyverse)
#> Warning: Paket 'tidyverse' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'ggplot2' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'tibble' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'tidyr' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'readr' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'purrr' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'dplyr' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'stringr' wurde unter R Version 4.2.2 erstellt
#> Warning: Paket 'forcats' wurde unter R Version 4.2.2 erstellt
n <- 10
df <- data.frame(x = rnorm(n), y = rnorm(n), z = sample(c("A", "B", "C"), replace = T, size = n))

ggplot(data = df, aes(x = x, y= y, shape = z)) +
  geom_point(size = 3) +
  geom_text(aes(label = z), hjust = 1, vjust = 1) +
  scale_shape_manual(values = c("C" = 1, "B" = 2, "A" = 3), 
                     label = c("long Text about C", "long Text about  B", "long Text about  A")
                     )
```

<sup>Created on 2023-02-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

```

teunbrand:
This PR aims to fix #5208.

Briefly, it implements the advice in https://github.com/tidyverse/ggplot2/issues/5208#issuecomment-1446935976 in the documentation:

>  I have long operated under the rule of thumb that one should never specify labels without specifying breaks. We could just formalize that rule.

### DIFF
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -34,7 +34,9 @@
 #'   may choose a slightly different number to ensure nice break labels. Will
 #'   only have an effect if `breaks = waiver()`. Use `NULL` to use the default
 #'   number of breaks given by the transformation.
-#' @param labels One of:
+#' @param labels One of the options below. Please note that when `labels` is a
+#'   vector, it is highly recommended to also set the `breaks` argument as a
+#'   vector to protect against unintended mismatches.
 #'   - `NULL` for no labels
 #'   - `waiver()` for the default labels computed by the
 #'     transformation object
--- a/man/binned_scale.Rd
+++ b/man/binned_scale.Rd
@@ -55,7 +55,9 @@ Note that for position scales, limits are provided after scale expansion.
 Also accepts rlang \link[rlang:as_function]{lambda} function notation.
 }}
 
-\item{labels}{One of:
+\item{labels}{One of the options below. Please note that when \code{labels} is a
+vector, it is highly recommended to also set the \code{breaks} argument as a
+vector to protect against unintended mismatches.
 \itemize{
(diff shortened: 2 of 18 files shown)


thomasp85 (approved):
LGTM

PR #6244, opened 2024-12-17
Fix `geom_ribbon(na.rm)`

### LINKED ISSUE #6243: Quirky NA-handling of `geom_ribbon()`
Now that we're allowing gradients in `geom_ribbon()` and derivatives, the following caught my attention.

In the example below, we simply want to vary the gradient along `x`, so top and bottom gradients should match.
However, the gradient is misplaced because the second group has a missing value (note it starts with dark blue in the middle).

``` r
#> ℹ Loading ggplot2

df <- data.frame(
  x = c(1:3, 1:3),
  ymin = c(1, 1, 1, NA, 2, 2),
  ymax = c(1.5, 2, 1.5, 2.5, 3, 2.5),
  group = c(1,1,1,2,2,2)
)

p <- ggplot(df, aes(x, ymin = ymin, ymax = ymax, group = group, fill = x)) +
  scale_fill_viridis_c()

# Wrong
p + geom_ribbon()
```

The gradients are drawn correctly if missing values are removed properly.

``` r
p + geom_ribbon(na.rm = TRUE)
```

<sup>Created on 2024-12-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I think there are two things wrong with this when `na.rm = FALSE`:
* There is no warning about missing values, as you would expect with any other geom
* It doesn't actually  remove the missing data, causing the misalignment of gradients.

I traced this decision to a remedy against #1549.

teunbrand:
This PR aims to fix #6243.

Briefly, missing values are thrown out and warn about when `na.rm = FALSE`.
The warning is skipped when `na.rm = TRUE` but missing values are thrown out.
When throwing out missing values, the `group` aesthetic is reformed so that groups with NA values break up into new uninterrupted groups.

Reprex from issue, note gradient is uniform now:

``` r
#> ℹ Loading ggplot2

df <- data.frame(
  x = c(1:3, 1:3),
  ymin = c(1, 1, 1, NA, 2, 2),
  ymax = c(1.5, 2, 1.5, 2.5, 3, 2.5),
  group = c(1,1,1,2,2,2)
)

ggplot(df, aes(x, ymin = ymin, ymax = ymax, group = group, fill = x)) +
  geom_ribbon() +
  scale_fill_viridis_c()
#> Warning: Removed 1 row containing missing values or values outside the scale range
#> (`geom_ribbon()`).
```

<sup>Created on 2024-12-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* `geom_ribbon()` now appropriately warns about, and removes, missing values 
+  (@teunbrand, #6243).
 * `guide_*()` can now accept two inside legend theme elements:
   `legend.position.inside` and `legend.justification.inside`, allowing inside
   legends to be placed at different positions. Only inside legends with the same
--- a/R/geom-ribbon.R
+++ b/R/geom-ribbon.R
@@ -126,7 +126,31 @@ GeomRibbon <- ggproto("GeomRibbon", Geom,
 
   draw_key = draw_key_polygon,
 
-  handle_na = function(data, params) {
+  handle_na = function(self, data, params) {
+
+    vars <- vapply(
+      strsplit(self$required_aes, "|", fixed = TRUE),
+      `[[`, i = 1, character(1)
+    )
+    if (params$flipped_aes || any(data$flipped_aes) %||% FALSE) {
+      vars <- switch_orientation(vars)
+    }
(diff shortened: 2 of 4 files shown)


thomasp85 (approved):
LGTM

PR #6245, opened 2024-12-17
:memo: Tweak docs for `lims()`

### LINKED ISSUE #5148: Datapoints disapearing when `lims()` conflicts with `position_dodge()`
Hi,

It is a bit unexpected that `lims()` consider the actual position of each point (after considering the dodge) and not its original position.

This makes some points disappear while the real value is within the limits.

For instance, consider:

``` r
library(tidyverse)


dat=tibble(x=rep(0:4, 2), y=rep(0:4, 2), gp=rep(c("A","B"), each=5))
p = ggplot(dat, aes(x,y, color=gp)) + geom_point(position=position_dodge(0.5))
p
```

``` r
p + xlim(0,NA)
#> Warning: Removed 1 rows containing missing values (`geom_point()`).
```

<sup>Created on 2023-01-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I cannot see how this can be a wanted behavior, as dodging is only about visualization and not about showing actual data. However, if it is, maybe this would be worth documenting, for instance in `?position_dodge` and/or in `?lims`.

teunbrand:
This PR aims to fix #5148.
It more clearly points out out-of-bounds handling options.

### DIFF
--- a/R/limits.R
+++ b/R/limits.R
@@ -4,7 +4,9 @@
 #' scales. By default, any values outside the limits specified are replaced with
 #' `NA`. Be warned that this will remove data outside the limits and this can
 #' produce unintended results. For changing x or y axis limits \strong{without}
-#' dropping data observations, see [coord_cartesian()].
+#' dropping data observations, see
+#' [`coord_cartesian(xlim, ylim)`][coord_cartesian], or use a full scale with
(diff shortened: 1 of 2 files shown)


thomasp85 (approved):
LGTM

PR #6249, opened 2024-12-17
✨ Point and polygon theme elements

### LINKED ISSUE #6248: Feature request: elements for other graphical primitives
The request here is to introduce theme elements that support drawing polygons and points.
There a no cases in ggplot2 itself where this is currently useful, but I think it might benefit the extension community.
Sharing a common definition of what a point or polygon theme element is like prevents mutually exclusive and confusing implementations.
Aside from constructors like `element_polygon()` and `element_point()`, I think it is also important to have a root theme element, like `theme(point, polygon)` that has sensible defaults.

teunbrand:
This PR aims to fix #6248.

It does the following:
* New `element_point()` and `element_polygon()` elements.
* New `theme(point, polygon)` arguments.
* Gave sensible defaults for these theme settings.

It does not do the following:
* Use these new elements in any meaningful way


teunbrand:
Good to know I'm not the only one with this need :) I think exposing elements to `alpha` is a related but separate discussion that merits its own issue.

R/geom-point.R
```diff
@@ -188,6 +186,9 @@ GeomPoint <- ggproto("GeomPoint", Geom,
 #' # Strings with 1 or less characters are interpreted as symbols
 #' translate_shape_string(c("a", "b", "?"))
 translate_shape_string <- function(shape_string) {
```
thomasp85, line 188: Shouldn't this be `validate_shape_string()` per the new focused vocabulary?

thomasp85 (approved):
LGTM

R/geom-point.R
teunbrand: Good point, yes I think it should in principle. However, it is exported so it isn't really an internal function.

thomasp85: ah - let's keep it asis

PR #6236, opened 2024-12-12
Include image of shape table

teunbrand:
This PR aims to fix #4034.

It adds an svg image of the shape table to html docs and a pdf image of the shape table in the pdf docs.
I've checked with `devtools::build_manual()` that also the pdf documentation looks correct.

The code used for creating the figures is:

```r
nms <- c(
  "square open", "circle open", "triangle open", "plus", "cross", 
  "diamond open", "triangle down open", "square cross", "asterisk", 
  "diamond plus", "circle plus", "star", "square plus", "circle cross", 
  "square triangle", "square", "circle small", 
  "triangle", "diamond", "circle", "bullet", "circle filled", "square filled", 
  "diamond filled", "triangle filled", "triangle down filled"
)
shape <- translate_shape_string(nms)

df <- data.frame(
  x = 0:25 %% 5,
  y = 0:25 %/% 5,
  name = nms,
  shape = shape
)

p <- ggplot(df, aes(x, y)) +
  geom_point(aes(shape = I(shape)), size = 4, fill = "red", stroke = 1) +
  geom_text(
    aes(label = paste0(shape, ":\n", name)),
    vjust = -0.5, size = 10, size.unit = "pt"
  ) +
  scale_y_continuous(transform = "reverse", limits = c(-0.5, 5.1)) +
  scale_x_continuous(limits = c(-0.5, 4.5)) +
  theme_void()

ggsave(
  file.path("man", "figures", "shape_table.svg"), 
  plot = p, device = svglite::svglite, fix_text_size = FALSE, 
  width = 6, height = 4, dpi = 72, bg = "white"
)
ggsave(
  file.path("man", "figures", "shape_table.pdf"), 
  plot = p,
  width = 6, height = 4, dpi = 72, bg = "white"
)
```

PR #6270, opened 2025-01-06
Convert linetype `NA`s to blank lines

### LINKED ISSUE #6269: `linetype = NA` should imply no line, not errors
When using it with lines, the observations get flagged as missing and get removed.

``` r
library(ggplot2)

p <- ggplot(data.frame(x = 1:2, y = 1:2), aes(x, y))
p + geom_line(linetype = NA)
#> (`geom_line()`).
```
However, if we don't throw away such observations, we invite errors from grid.

``` r
p + geom_tile(linetype = NA)
#> Error in grid.Call.graphics(C_rect, x$x, x$y, x$width, x$height, resolveHJust(x$just, : invalid line type
```

<sup>Created on 2025-01-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This can be painful when using scales, as in #6097 we've changed `scale_linetype()$na.value` to `NA`.
My suggestion for non-line geometries is to translate these `NA`s to `"blank"` or 0, depending on the type.

teunbrand:
This PR aims to fix #6269.

Reprex from the issue; note no errors and no red outline:

``` r
#> ℹ Loading ggplot2

ggplot(data.frame(x = 1:2, y = 1:2), aes(x, y)) +
  geom_tile(linetype = NA, colour = "red")
```

<sup>Created on 2025-01-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85:
Does this change mean that the linetype legend guide now has a blank slot for NA values? That is not very telling if so

teunbrand:
It has a blank slot for NA values when there are NA values included in the scale limits/breaks, which by default occurs when there are NAs in the data.

thomasp85:
I guess this is a rare occurrence but it may catch users off-guard. I know this isn't really a fault of this PR but more a consequence of changing `na.value` to `NA`. 

I think I'd feel better if `na.value` remained as `"blank"` but internally got translated to `NA` since this makes it clearer to the user that lines are expected to be blank

teunbrand:
The issue with `na.value = "blank"` is that you can't use numeric linetypes like `1` (solid), `2` (dashed) etc. Of course you could then specify `na.value = 0` but that is a bit of an unecessary inconvenience in my view.

I'm not sure I understand why having a blank linetype key would be undesirable? Is this not the correct behaviour?

``` r
#> ℹ Loading ggplot2

df <- data.frame(x = 1:3)

ggplot(df, aes(x, xend = x + 1, y = x)) +
  geom_segment(aes(linetype = c("A", "B", NA)))
#> Warning: Removed 1 row containing missing values or values outside the scale range
#> (`geom_segment()`).
```



<sup>Created on 2025-01-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thomasp85:
I just mean that the default in the scale constructor is `"blank"` but under the hood we convert it to an `NA` and then keep the changes in this PR. It is more about signalling to the user what the scale does as I think `na.value = NA` is very confusing and doesn't help the user

All that being said, there are actual issues to having it blank as it completely removes the data. So you cannot see where the missing data supposedly is. But that is a whole other discussion and the ship has sailed on it I think

teunbrand:
So if I interpret this right, this is just about making sure the usage section in the documentation gives an interpretable value?

If that is the case, I have a counterproposal. We remove `na.value = NA` from the linetype scales, as it is the default in `discrete_scale()`. However, we give linetypes the #6236 treatment and include a little table with character/numeric interpretation of linetypes.

teunbrand:
I went ahead and implemented the linetype table idea

thomasp85 (commented):
LGTM

thomasp85 (approved):

Issue #3441, opened 2019-07-15
Feature request: enable making copies of trained position scales


clauswilke:
To create plots with marginals, it would be helpful if there was a straightforward method to reliably create copies of trained position scales. I'm wondering whether this is possible or can be made possible after @paleolimbot's refactoring of the position scales.

For example, consider this code, which uses the function `cowplot::axis_canvas()`:
``` r
library(tidyverse)
library(cowplot)

p1 <- ggplot(iris, aes(Sepal.Length, Sepal.Width, color = Species)) +
  geom_point(alpha = 0.5)

p1
```

![](https://i.imgur.com/PxeDPqU.png)

``` r

p2 <- axis_canvas(p1, "x", iris, aes(x = Sepal.Length, fill = Species)) +
  geom_density()

p2 + theme_grey()
```

![](https://i.imgur.com/zLMauY1.png)

``` r

p1 %>% insert_xaxis_grob(p2) %>% ggdraw()
```

![](https://i.imgur.com/q5K33aH.png)

I'd want to be able to generate the plot `p2` by writing something like the following:
``` r
p2 <- ggplot(iris, aes(x = Sepal.Length, fill = Species)) +
  geom_density() +
  scale_x_copy(p1)
```
or maybe 
``` r
p2 <- ggplot(iris, aes(x = Sepal.Length, fill = Species)) +
  geom_density() +
  scale_x_copy(ggplot_build(p1))
```
The function `scale_x_copy()` would take all information (type, limits, expansion, labels) about the x position scale from the built object of `p1` and then faithfully reproduce it. Importantly, the scale should not change range or other properties based on the data provided in the second plot.

I'd be happy to implement such a scale myself if somebody can give me some hints about how to best go about it.

paleolimbot:
For `CoordCartesian()`, `ggplot_build(p1)$layout$panel_params[[1]]$x` is scale-like (i.e., has `get_limits()`, `get_breaks()` and `get_labels()`...it is technically a `ViewScale`). You might have some trouble with expansion...I forget whether `get_limits()` returns the expanded range or the original range. I *think* I thought ahead on this and had it return the unexpanded range and had `$dimension()` return the expanded range...

clauswilke:
[...]

PR #6273, opened 2025-01-07
Continuous limits for discrete scales

### LINKED ISSUE #4174: Harmonize width and location of horizontal bars across facets with different number of bars
First, many thanks again to everyone involved with this wonderful package.

I regularly make plots like the one below. While it conveys all the necessary info, its format is (imho) not really fit for publication.

What bothers me is that 
a) bar widths are not the same (=should all have the width of the bars in the facet with most bars, i.e. europe), 
b) that the bars in the different facets are not horizontally aligned (i.e. the bar BR in the facet south america should have the same vertical position as the bar ES in the facet europe). I would have hoped that facet_row would do the trick, but it seems not to be the case.

I might be mistaken, but it seems to me that the plot below is a fairly regularly required format. It would be great if there is a convenient way to produce it.

Many thanks!

``` r
library(tidyverse)
df <- data.frame(
  stringsAsFactors = FALSE,
         continent = c("europe","europe","europe",
                       "europe","asia","asia","asia","south america"),
           country = c("AT", "DE", "DK", "ES", "PK", "IN", "TW", "BR"),
             value = c(10L, 15L, 20L, 25L, 17L, 17L, 13L, 5L)
)

df %>% 
  ggplot()+
  geom_bar(aes(y=country,
               x=value),
           orientation = "y",
           stat="identity")+
  facet_wrap(vars(continent),
             scales = "free_y")
```

<sup>Created on 2020-08-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
```

### LINKED ISSUE #6259: Feature request: set limits for `coord_radial()`
is it possible to set limits in `coord_radial()`? Just like what we do in `coord_cartesian(xlim = *, ylim = *)`

### PR DESCRIPTION
**Pull request by teunbrand**
This PR aims to fix #6259 and fix #4174.

Briefly, the discrete position scales get a new argument `continuous.limits` which controls what range is displayed.
It always acts as 'zoom limits', similar to limits set in the coord or continuous limits with `oob = scales::oob_keep()`.

Reprex from #4174:

``` r
#> ℹ Loading ggplot2
df <- data.frame(
  stringsAsFactors = FALSE,
  continent = c("europe","europe","europe",
                "europe","asia","asia","asia","south america"),
  country = c("AT", "DE", "DK", "ES", "PK", "IN", "TW", "BR"),
  value = c(10L, 15L, 20L, 25L, 17L, 17L, 13L, 5L)
)

p <- df |>
  ggplot() +
  geom_col(
    aes(y = country, x = value),
    orientation = "y"
  ) +
  facet_wrap(vars(continent), scales = "free_y")

p + scale_y_discrete(continuous.limits = c(1, 4))
```

<sup>Created on 2025-01-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* `scale_{x/y}_discrete(continuous.limits)` is a new argument to control the
+  display range of discrete scales (@teunbrand, #4174, #6259).
 * `geom_ribbon()` now appropriately warns about, and removes, missing values 
   (@teunbrand, #6243).
 * `guide_*()` can now accept two inside legend theme elements:
--- a/R/scale-discrete-.R
+++ b/R/scale-discrete-.R
@@ -69,7 +69,8 @@
 #' }
 scale_x_discrete <- function(name = waiver(), ..., palette = seq_len,
                              expand = waiver(), guide = waiver(),
-                             position = "bottom", sec.axis = waiver()) {
+                             position = "bottom", sec.axis = waiver(),
+                             continuous.limits = NULL) {
   sc <- discrete_scale(
     aesthetics = ggplot_global$x_aes, name = name,
     palette = palette, ...,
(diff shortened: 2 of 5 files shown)


teunbrand:
I now also believe this would be needed for #3441 to make copies of discrete scales with immutable ranges.

thomasp85 (approved):
LGTM

werkstattcodes:
many thanks for implementing the feature i had raised in [#4174.](https://github.com/tidyverse/ggplot2/issues/4174)!!

I was just giving it a try, and fail to understand how I can "top-align" the bars within a facet. 
Currently, they are on the bottom of the facet. 

Thanks again for all your contributions to ggplot2 and beyond!

teunbrand:
You can use `coord_cartesian(reverse = "y")`

PR #6281, opened 2025-01-13
🩹 Turn off `size` fallback from `GeomBar`

### LINKED ISSUE #4848: `n` legend with single value `1` when using `geom_bar(stat = "sum")`
Using `stat = "sum"` in `geom_bar()` displays a count legend labelled "n" with a grey square as a default. It is particularly out of place when the bars are filled, resulting in what seems to be two `fill` legends, the grey not matching anything on the plot.

The issue has been brought up on StackOverflow, with a workaround: https://stackoverflow.com/questions/50378718/what-is-the-n-1-box-in-my-r-geom-bar-legend-and-how-do-i-remove
``` r
library(ggplot2)
exp_df <- data.frame(x = c("A", "B", "B", "C"),
           value = 1:4,
           group = c("Z", "Z", "Y", "Y"))
ggplot(exp_df, aes(x, value)) +
  geom_bar(stat = "sum")
```

``` r
# with fill
ggplot(exp_df, aes(x, value, fill = group)) +
  geom_bar(stat = "sum")
```

<sup>Created on 2022-05-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Should `geom_bar()` not display this legend at all?

teunbrand:
This PR aims to fix #4848.

Briefly, `geom_bar()` does not have the `size` aesthetic, but the 3.4.0 fallback mechanism generated an awkward scale when using `stat = "sum"`. This turns off the fallback mechanism for this geom, but perhaps a larger conversation to be had is when the fallback should be turned off in general. Turning off the fallback also turns off the awkward legend that the issue was about in the first place.


thomasp85:
I think we are at the point were we can start a deprecation process. We already have deprecation warnings for `default_aes` to alert extension developers and 2 years in I think we can begin to make noise for the users

thomasp85 (approved):
LGTM

PR #6282, opened 2025-01-13
✨ Using date breaks/minor breaks/labels in time scales.

### LINKED ISSUE #4335: Adding date_breaks to `scale_*_time`
I am working with a dataset where there are huge time differences. 
If I use `scale_*_time` I cannot convert those labels to a more meaningful label such as 1 week, 2 weeks... as it can be done with `scale_*_date*`. Could it be possible to add the `date_breaks` argument to `scale_*_time` (or something similar)?

```r
library("hms")
library("ggplot2")
data.frame(diff = as_hms(c(1850000, 800)), time = rep(Sys.time(), 2))
ggplot(df) +
    geom_point(aes(time, diff)) 
```

This would make it easier to interpret the plots. 

----

Couldn't find any issue related to this

teunbrand:
This PR aims to fix #4335.

Briefly, it pulls `scale_{x/y}_time()` into the `datetime_scale()`-fold of functions.
I'm at a slight 'Chesterton's fence'-risk because I don't know why `scale_{x/y}_time()` was excluded in the first place, but I'll take the risk.


thomasp85 (approved):
I don't like the name `date_breaks` for this because a date has a fixed position in time whereas the time scale is relative. But for symmetry with the other scales it is probably worth this minor gripe

teunbrand:
I also don't think the names are great, but I'm not a big fan of the date break mechanism to begin with.

PR #6285, opened 2025-01-21
Inheritance of theme-based aesthetics

### LINKED ISSUE #6347: Could `element_geom` use aesthetic argument names?
`element_geom` looks like a cool new feature. 

Was wondering if there is a reason why colour/fill etc are not used - and instead ink/paper etc are?

It'd be cool if the arguments could align with the aesthetic names.

So currently like this...
```
penguins |> 
  ggplot() +
  geom_boxplot(aes(x = species, y = flipper_length_mm)) +
  theme(geom.boxplot = element_geom(ink = "blue", paper = "red")) 
```

This would seem much more intuitive...
```
penguins |> 
  ggplot() +
  geom_boxplot(aes(x = species, y = flipper_length_mm)) +
  theme(geom.boxplot = element_geom(colour = "blue", fill = "red")) 
```

Sorry if I'm missing something

teunbrand:
This PR is an amendment to #5833 and closes #6347.

Briefly, we want to deprecate `update_geom_defaults()` and to do that we need to increase the flexibility of the theme based styling of geom defaults. In this PR, we include inheritance of theme-based aesthetics to allow finer customisation.

In order to make this not too restrictive, we're allowing any `theme(geom.* = element_geom())`, which takes same logic in terms of theme validation.

``` r
#> ℹ Loading ggplot2

class(GeomDensity)
#> [1] "GeomDensity" "GeomArea"    "GeomRibbon"  "Geom"        "ggproto"    
#> [6] "gg"

ggplot(mpg, aes(displ)) +
  geom_density() +
  theme(
    geom.ribbon = element_geom(borderwidth = 2),
    geom.area = element_geom(bordertype = 2),
    geom.density = element_geom(ink = "red")
  )
```

<sup>Created on 2025-01-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85:
@clauswilke and @yutannihilation I'd like your eyes on this and especially our bid to fully make `update_geom_defaults()` disposable. 

I always really disliked that global state change that `update_geom_defaults()` incur but can also see why people have used it - I think the above change strikes a nice middle ground but as someone who never uses `update_geom_defaults()` I might have blind spots

teunbrand:
Just to give away one weakness of the approach; people can change specifically the colours of `geom_point()` using this mechanism, but not the `fill` because `GeomPoint$default_aes` gives `fill = NA` instead of `fill = from_theme(...))`.

yutannihilation:
I agree this is 100% the right direction!

One thing I'm still unsure is how long it takes until the "ink" mechanism gets understood by the majority of ggplot2 users and the extension developers. I personally feel the trickiness @teunbrand pointed out is understandable, but we might hit more cases like that, especially on Geoms provided by the extension package. It's not visible to users which aesthetic of a Geom supports `from_theme()` unless they read the source code, right? Regarding this point, an ordinary user might feel `update_geom_defaults()` is simpler.

teunbrand:
Thanks for taking a look Claus and Hiroaki! 

> Don't issues such as the fill one you point out just mean that geom defaults would have to be updated accordingly?

Yes they can be, it just would make me slightly grumpy that there then would be 3 places where defaults are defined, which is hard to track/explain. (i. the `default_aes` field, ii. the `geom` theme setting and iii. `geom.point` theme setting). The thing I like about the current solution is that there is only (i) and (ii).

> One thing I'm still unsure is how long it takes until the "ink" mechanism gets understood by the majority of ggplot2 users and the extension developers.

Thomas and I have discussed putting together a deep-dive blogpost on how to theme/style your plots, so this is definitely something that I planned to include in such post. 

> It's not visible to users which aesthetic of a Geom supports from_theme() unless they read the source code, right?

You can also print e.g. `geom_point()$geom$default_aes`. Alternatively, we could do something along the lines of #5461 to make this more clear.

yutannihilation:
Thanks, I'm looking forward to the blog post! https://github.com/tidyverse/ggplot2/pull/5461 looks a good solution if it's possible.

teunbrand:
Important point from discussion in https://github.com/tidyverse/ggplot2/issues/6347#issuecomment-2691461667:

> the only thing that I'd like to ask is that I can override any default color mixing in the theme.

teunbrand:
It now works as follows. The `ink`/`paper` setttings work as before:
``` r
#> ℹ Loading ggplot2

p <- ggplot(economics, aes(date, unemploy)) +
  geom_area()

p + theme(geom.area = element_geom(ink = "red", paper = "white"))
```



But a `colour`/`fill` setting can override the `ink`/`paper` setting. The idea is that you only use this in geom-specific theme elements, and leave colour/fill as `NULL` in the global `theme(geom)`.

``` r
p + theme(geom.area = element_geom(
  colour = "red", fill = "white",
  ink = "blue", paper = "black"
))
```



<sup>Created on 2025-03-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In addition documentation now looks like so:

image

thomasp85 (approved):
LGTM

PR #6296, opened 2025-01-22
🚑 Fix for revdep problem with adding `NULL` to theme

### LINKED ISSUE #6291: Adding `NULL` to a theme
This was identified in #6287 and affects 16 packages.

You can add `NULL` to a regular theme and it will be a no-op.
However, some packages don't also have the `"gg"` class for their themes, which causes problems when adding `NULL`.

``` r
library(ggplot2)

class(theme())
#> [1] "theme" "gg"
class(survminer::theme_survminer())
#> [1] "theme"

theme() + NULL
#>  Named list()
#>  - attr(*, "class")= chr [1:2] "theme" "gg"
#>  - attr(*, "complete")= logi FALSE
#>  - attr(*, "validate")= logi TRUE
survminer::theme_survminer() + NULL
#> Error in survminer::theme_survminer() + NULL: non-numeric argument to binary operator
```

<sup>Created on 2025-01-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This can become a problem in the chunk of code below:

https://github.com/tidyverse/ggplot2/blob/6a61201d7e8da66579bfdcfa1b39786c5a52f089/R/guides-.R#L541-L549

teunbrand:
This PR aims to fix #6291.

Briefly it circumvents the `+` dispatch by using `add_theme()` directly.
In addition, contains a small failsafe for when `theme(legend.position)` is not recognised.

### DIFF
--- a/R/guides-.R
+++ b/R/guides-.R
@@ -546,7 +546,8 @@ Guides <- ggproto(
           legend.justification.inside = groups$key$justs[[i]]
         )
       }
-      grobs[[i]] <- self$package_box(grobs[[i]], position, theme + adjust)
+      adjust <- add_theme(theme, adjust, "internal theme settings")
+      grobs[[i]] <- self$package_box(grobs[[i]], position, adjust)
     }
 
     # merge inside grobs into single gtable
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

PR #6297, opened 2025-01-22
🚑 Fix for using `vec_slice()` before naturalising data

### LINKED ISSUE #6292: Mapping facets with `vec_slice()` is too stringent
This problem was identified in #6287 and affects 3 packages.

Essentially, when you're dealing with a `data.frame` superclass (rather than a subclass), `vec_slice()` complains.

``` r
vctrs::vec_slice(mtcars, 1)
#>           mpg cyl disp  hp drat   wt  qsec vs am gear carb
#> Mazda RX4  21   6  160 110  3.9 2.62 16.46  0  1    4    4

df <- mtcars
class(df) <- union(class(df), "foo")

vctrs::vec_slice(df, 1)
#> Error in `vctrs::vec_slice()`:
#> ! `x` must be a vector, not a <data.frame/foo> object.
```

<sup>Created on 2025-01-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This normally doesn't happen because we rebuild our own plain data.frame after evaluating aesthetics, but it becomes a problem before that when mapping facets:

https://github.com/tidyverse/ggplot2/blob/97edd6260a44c71a8ce7954ae4ad6be29224bda3/R/facet-grid-.R#L312

teunbrand:
This PR aims to fix #6292.

It replaces the vctrs subsetting by base subsetting.

### DIFF
--- a/R/facet-grid-.R
+++ b/R/facet-grid-.R
@@ -309,7 +309,9 @@ FacetGrid <- ggproto("FacetGrid", Facet,
         params$margins
       )
       # Apply recycling on original data to fit margins
-      data <- vec_slice(data, facet_vals$.index)
+      # We're using base subsetting here because `data` might have a superclass
+      # that isn't handled well by vctrs::vec_slice
+      data <- data[facet_vals$.index, , drop = FALSE]
       facet_vals$.index <- NULL
     }
 


thomasp85 (approved):
LGTM

PR #6302, opened 2025-01-23
🚑 Amend clause in #6244

teunbrand:
This PR aims to fix an issue uncovered in #6287.

Briefly, in some extension packages an if-statement might have length other than 1. This PR asks for exactly length 1.


thomasp85 (approved):
LGTM

PR #6303, opened 2025-01-23
🚑 Fallback for boolean `space_free`

teunbrand:
This PR aims to fix an issue incovered in #6287.

Briefly, some extensions pass the `space_free` parameter as a boolean instead of a named list with `x` and `y` elements.
This PR ensures the boolean gets translated.


teunbrand:
Closing this as I feel it is better handled at the extension's end of things.

PR #6304, opened 2025-01-23
🚑 Give default value to `Guides$build(theme)`

teunbrand:
This PR aims to fix an issue identified in #6287.

Briefly, some packages implement `ggplot_build()` methods. We added a `theme` argument to `Guides$build()`, which is now causing complaints about missing arguments in reverse dependencies. This PR provides a fallback for when the extension's `ggplot_build()` method does not support the new guide building yet.


thomasp85 (approved):
LGTM

PR #6238, opened 2024-12-12
Deprecate `fatten`

teunbrand:
This PR aims to fix #4881.

Briefly it deprecates `fatten` in `geom_pointrange()` (per linked issue), but also `geom_boxplot()` and `geom_crossbar()`.
The parameter still works as it used to, but this at least starts the deprecation process.
There is a slight hiccup in `geom_boxplot()` as it isn't a named argument, but we check for if the argument has been picked up in the parameters.

``` r
devtools::load_all("~/packages/ggplot2/")
#> ℹ Loading ggplot2

df <- data.frame(
  trt = factor(c(1, 1, 2, 2)),
  resp = c(1, 5, 3, 4),
  group = factor(c(1, 2, 1, 2)),
  upper = c(1.1, 5.3, 3.3, 4.2),
  lower = c(0.8, 4.6, 2.4, 3.6)
)

p <- ggplot(df, aes(x = trt, y = resp, col = group, ymin = lower, ymax = upper))

ggplotGrob(
  p + geom_pointrange() + 
    geom_crossbar() + 
    geom_boxplot()
) |> invisible()

ggplotGrob(
  p + geom_pointrange(fatten = 1) + 
    geom_crossbar(fatten = 1) + 
    geom_boxplot(fatten = 1)
) |> invisible()
#> Warning: The `fatten` argument of `geom_pointrange()` is deprecated as of ggplot2 3.6.0.
#> ℹ Please use the `size` aesthetic instead.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> Warning: The `fatten` argument of `geom_crossbar()` is deprecated as of ggplot2 3.6.0.
#> ℹ Please use the `middle.linewidth` argument instead.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> Warning: The `fatten` argument of `geom_boxplot()` is deprecated as of ggplot2 3.6.0.
#> ℹ Please use the `median.linewidth` argument instead.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
```

<sup>Created on 2024-12-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #6310, opened 2025-01-29
🩹 Don't warn when fatten is absent

teunbrand:
This PR amends #6238.

Briefly: it prevents a bunch of false positive warnings about the `fatten` parameters.

In some details; `lifecycle::is_present(params$fatten)` was evaluated as `TRUE` when fatten was absent and `params$fatten` returns `NULL`. This caused a bunch of false-positive deprecation warnings in stats that has `geom = "crossbar"` or `geom = "pointrange"`, like `stat_summary_bin()`.


thomasp85 (approved):
LGTM

PR #6322, opened 2025-02-05
Skip annotation labels

### LINKED ISSUE #6316: Deprioritise annotations in deriving labels
This was first suggested in https://github.com/tidyverts/feasts/pull/166#discussion_r1936807002.

When you use an annotation you want to use as the background, you use that annotation before the real 'data' layers. For example, to highlight the 70s you can use:

``` r
#> ℹ Loading ggplot2

  annotate(
    "polygon",
    x = as.Date(c("1970-01-01", "1970-01-01", "1980-01-01", "1980-01-01")),
    y = c(Inf, -Inf, -Inf, Inf),
    alpha = 0.25
  ) +
  geom_line()
```

<sup>Created on 2025-01-31 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

However the default label is now `x` and not `date` and the question is whether we should deprioritise annotations in deriving automatic labels? Currently, there is nothing distinguishing an annotation layer from a regular layer beyond the constructor.

teunbrand:
This PR aims to fix #6316.

Briefly, it skips label extraction when the mapping was constructed using the `aes_all()` function. This is communicated through a subclass. The layers from `annotation()` and `expand_limits()` functions are thereby not contributing to labels.

### DIFF
--- a/R/aes.R
+++ b/R/aes.R
@@ -360,7 +360,7 @@ aes_all <- function(vars) {
   # refer to the data mask
   structure(
     lapply(vars, function(x) new_quosure(as.name(x), emptyenv())),
-    class = "uneval"
+    class = c("unlabelled_uneval", "uneval")
   )
 }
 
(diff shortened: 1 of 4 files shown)


thomasp85 (approved):
LGTM

PR #6326, opened 2025-02-07
Chain stat

### LINKED ISSUE #6325: Feature request: stat executing multiple other stats
From Thomas in [the discussion of #6231](https://github.com/tidyverse/ggplot2/pull/6231#issuecomment-2618134236):

> Conceptually this feels like it belongs in a stat, but that would preclude it from being used together with other stats like in the examples above. However, I feel we did touch on the idea of having a stat_stack() at some point which would remove this downside

I'm lifting this out from that discussion to a separate issue as it blocks progress for #6231. Briefly, it would be good to have a stat, whose job it is to direct the serial computation of other stats. It would make re-implementing the `position_connect()` to a `stat_connect()` worthwhile. In addition, I think some other 'chains' of stats can be neat too, such as a typical stat (e.g. `stat_bin`) combined with more of a 'geometric' stat like `ggforce::stat_link()`, `ggforce::stat_bezier()` etc.

teunbrand:
This PR aims to fix #6325.

Briefly, it introduces a new stat `stat_chain()` that can combine other stats. It uses `link_stat()` as a helper to parameterise the individual stats that it combines.

I like the 'links in a chain' analogy for this stat better than a 'stack of stats', because while they both indicate a serial nature, for chains it matters how they are interlinked, whereas a stack isn't connected properly. The only downside is that 'link' these days evoke HTML hyperlinks more than chain links, which is a potential source of confusion.

The example section rendered:

``` r
#> ℹ Loading ggplot2

p <- ggplot(mpg, aes(displ, colour = drv))
# Binning unique observations
p + stat_chain(stats = c("unique", "bin"))
#> `stat_bin()` using `bins = 30`. Pick better value `binwidth`.
```

``` r
# Controlling parameters
p + stat_chain(
  stats = list("unique", link_stat("bin", bins = 10))
)
```

``` r
# Evaluate expressions after computing stats
p + stat_chain(
  stats = list("unique", link_stat("bin", mapping = aes(y = density)))
)
#> `stat_bin()` using `bins = 30`. Pick better value `binwidth`.
```

<sup>Created on 2025-02-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I've tried to put some thought into the API and I can't really come up with a better solution than the `link_stat()` helper function. I'd happily consider other suggestions.

### DIFF
--- a/R/stat-chain.R
+++ b/R/stat-chain.R
@@ -37,13 +37,19 @@
 #' p + stat_chain(stats = c("unique", "bin"))
 #' # Controlling parameters
 #' p + stat_chain(
-#'   stats = c("unique", "bin"),
-#'   stat.params = list(NULL, list(bins = 10))
+#'   stats = list("unique", link_stat("bin", bins = 10))
 #' )
 #' # Evaluate expressions after computing stats
+#' p + stat_chain(stats = list(
+#'   link_stat("unique",  after.stat = aes(x = x + 1)),
+#'   link_stat("density", after.stat = aes(y = density))
+#' ))
+#' # Note that the last `after.stat` argument serves the same role as the
+#' # `after_stat()` function in the layer mapping, so the following is
+#' # equivalent to the previous plot
 #' p + stat_chain(
-#'   stats = c("unique", "bin"),
-#'   redirect = list(aes(x = x + 1), aes(y = density))
+#'   mapping = aes(y = after_stat(density)),
+#'   stats = list(link_stat("unique", after.stat = aes(x = x + 1)), "density")
 #' )
 stat_chain <- function(
     mapping = NULL,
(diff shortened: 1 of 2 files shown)


teunbrand:
Consider:
* Pass dots to first stat?

teunbrand:
> Pass dots to first stat?

This will be more complicated than anticipated. 
If we try to resolve this in the stat constructor before `layer()`, this will not port well when using `stat = "chain"` in a geom constructor.
We cannot resolve this during `layer()` because `StatChain$parameters()` is unaware of the stats passed as the `stats` argument due to statelessness. Without editing `layer()`, there is also no way within `layer()` to have a stat accept all, arbitrary parameters (to filter later).

PR #6327, opened 2025-02-10
Repurpose rect key

### LINKED ISSUE #5385: `draw_key_rect` fails to get color from outline when `fill` is `NA`
#### Problem:
When using `draw_key_rect` with a `geom` that has `fill = NA`, it fails to catch the outline color and returns an invisible (`NA`) rectangle.

#### Minimal reprex:
``` r
library(ggplot2)
ggplot(mtcars, aes(disp, color = factor(cyl))) + geom_density(key_glyph = draw_key_rect)
```

<sup>Created on 2023-08-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5385.

Briefly, the `draw_key_rect()` function defaults NA-fill to the colour aesthetic.
To course-correct `geom_rect()` is now using `draw_key_polygon()` instead, which has been given 0-width linewidth by default.


teunbrand:
I have nu clue why the checks pass locally and most CI checks fail

thomasp85 (approved):
LGTM

PR #6328, opened 2025-02-11
Dodge stack for ggplot

szkabel:
I have been using this extension of mine for a while for dodging and stacking together. Even though it is a functionality only useful for column geoms, it still seems to be a topic of interest based on a recently opened feature request. I myself implemented it because of a need in a research project.

The idea of the functionality is that after dodging the rest of the values are stacked. For us this was really beneficial. Also the addition is written so that it doesn't change any previous behavior. 

I have made a description on why I think this is useful and why previously this may have caused confusion here: https://rpubs.com/szkabel/dodgeStackDemo


teunbrand:
Hi there, thanks for putting together this PR. Unfortunately, we don't think this functionality is a good fit for ggplot2 and would instead recommend implementing this functionality in an extension package. We nonetheless appreciate the motivation and time invested in preparing this PR and hope this doesn't impede any future contributions!

clauswilke:
@szkabel To follow up on Teun's point: I also think that this should not be part of the standard `position_dodge()` position adjustment but instead go into its own. And once you do that you don't have to add the code to ggplot2 main. Something like ggforce might be more appropriate, if you don't want to write your own extension package for just one single function.

PR #6330, opened 2025-02-12
Compute jitter width by panel

### LINKED ISSUE #3722: Misleading noise when using geom_jitter in combination with facet_wrap
Hello,

I like to use geom_jitter mostly to plot discrete data and avoid overplotting.

When using a combination of `geom_jitter` and `facet_wrap` with free scales I came to the issue that due to the free scales, jittering appeared misleading. 

In the example below, I would like to see approximately equal noise, as I use jittering purely for visibility of data points. Apparently though, due to the free scales in `geom_wrap` the jittering looks different in the two facets. 
How could this be approached? Could this be an issue inside `face_wrap`?

``` r
library(ggplot2)

set.seed(123)
df <- data.frame(
  y = rnorm(100, 50, 10),
  x = c(rep(1:5, 10), rep(seq(0, 100, length.out = 5), 10)),
  group = c(rep("a", times = 50), rep("b", times = 50))
)

ggplot(df, aes(x = x, y = y)) +
  geom_point(color = "orange", alpha = 0.7) +
  geom_point(position = "jitter", color = "blue", alpha = 0.7) +
  facet_wrap( ~ group, scales = "free")
```

<sup>Created on 2020-01-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

teunbrand:
This PR aims to fix #3722.

It ensures that the default, resolution-based jitter width is computed per panel instead of per layer.

Reprex from  the issue, notice that the jitter has about equal visual spread across panels:

``` r
#> ℹ Loading ggplot2

set.seed(123)
df <- data.frame(
  y = rnorm(100, 50, 10),
  x = c(rep(1:5, 10), rep(seq(0, 100, length.out = 5), 10)),
  group = c(rep("a", times = 50), rep("b", times = 50))
)

ggplot(df, aes(x = x, y = y)) +
  geom_point(color = "orange", alpha = 0.7) +
  geom_point(position = "jitter", color = "blue", alpha = 0.7) +
  facet_wrap( ~ group, scales = "free")
```

<sup>Created on 2025-02-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

An open question is whether we should do this for `position_jitterdodge()` too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -239,8 +239,8 @@
   and (non-text) margins inherit from (@teunbrand, #5622).
 * `geom_ribbon()` can have varying `fill` or `alpha` in linear coordinate
   systems (@teunbrand, #4690).
-* `geom_tile()` computes default widths and heights per panel instead of
-  per layer (@teunbrand, #5740).
+* `geom_tile()` and `position_jitter()` computes default widths and heights 
+  per panel instead of per layer (@teunbrand, #5740, #3722).
 * The `fill` of the `panel.border` theme setting is ignored and forced to be
   transparent (#5782).
 * `stat_align()` skips computation when there is only 1 group and therefore
--- a/R/position-jitter.R
+++ b/R/position-jitter.R
@@ -68,8 +68,8 @@ PositionJitter <- ggproto("PositionJitter", Position,
       seed <- self$seed
     }
     list(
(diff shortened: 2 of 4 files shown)


thomasp85 (approved):
LGTM

Issue #6293, opened 2025-01-22
Coord_flip() has some issues when interacting with panel.grid.major and panel.grid.minor


oskard95:
I found a problem with panel.grid and coord_flip in the 3.5 update. It seems that the panel grids only produce on observed values but not on axis values after coord_flipping. 

In the code below, when you leave out coord_flip - the panel.grid lines will be normal - when you include coord_flip() the X axis lines (flipped to Y axis) only show on values of X. 

If you include scale_y_contininious - the lines appear again. 

Below, leave is some simple code to reproduce the bug:

```r
library(tidyverse)
x <- -1:10
y <- -1:10
dati <- as.data.frame(x,y)


dati %>% 
  ggplot(aes(x = x, y = y)) +
  geom_point() +
  scale_x_continuous(breaks = seq(-15, 15, by = 3), 
                     limits = c(-15,15)) +
  # scale_y_continuous(breaks = seq(-15, 15, by = 3), 
  #                     limits = c(-15,15)) +
  # coord_flip()+
  theme(panel.grid.major = element_line(color = "blue"),
        panel.grid.minor = element_line(color = "red"))


```

![Image](https://github.com/user-attachments/assets/7df9feaf-a777-47d7-a1f8-6d005189bc0c)
This is without coord_flip - as is in the code example. 

![Image](https://github.com/user-attachments/assets/cef3ee7a-c2d4-4558-a411-edbdc3d7fd87)

And this is when i include coord_flip exact same code - just removed the comment in front of coord_flip. If the scale_y_contininious is left in there, it fixes itself but I think this is not supposed to be needed.

I think this is related to the 3.5 update as its been confirmed by others to work as intended in previous versions. I understand the function was superseded but I think this is unintended behaviour

teunbrand:
Thanks for the report! I can reproduce the issue and agree that it looks like a bug.

PR #6332, opened 2025-02-14
Flip `geom_curve()` direction in some circumstances

### LINKED ISSUE #5069: `geom_curve` curvature direction is inconsistent on flipped axes
The direction of the curvature in `geom_curve` is inconsistent with respect to other plot elements. This is easy to work around, but it would be nice if it were consistent.

See the following examples:

``` r
library(ggplot2)

p <- 
  ggplot(mtcars, aes(x = drat, y = mpg)) +
  geom_point() +
  annotate("curve", x = 4.5, xend = 4.5, y = 20, yend = 15, curvature = -1)

p + ggtitle("In the base plot, the curve arcs away from the points...")
```

``` r

p + coord_flip() + 
  ggtitle("Flipped, the curve arcs *towards* from the points...")
```

``` r

p + coord_flip() + scale_y_reverse()+ 
  ggtitle("Flipped reversed, the curve arcs away from the points again!")
```

<sup>Created on 2022-11-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #5069.

Briefly, when reversing or flipping scales, we need to adjust the curves in `geom_curve()` to be consistent. This PR flips curves by swapping the start- and end coordinates.

The approach harcodes some conditions wherein the swap needs to happen as there isn't a sensible way to derive that from the data itself. The downside of this is that a user can implement a reversing transformation with a different name and we wouldn't detect this.

Also `geom_curve()` wasn't covered by any tests, so I added some visual ones.

Based on the reprexe in the linked issue; note that the curve consistently curves 'outward' (away from the panel centre).
The missing panel grid is due to #6293.

``` r
#> ℹ Loading ggplot2

p <- 
  ggplot(mtcars, aes(x = drat, y = mpg)) +
  geom_point() +
  annotate("curve", x = 4.5, xend = 4.5, y = 20, yend = 15, curvature = -1, arrow = arrow())

p
```

``` r

p + coord_flip()
```

``` r

p + coord_flip() + scale_y_reverse()
```

<sup>Created on 2025-02-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM, though the flip_curve logic will be a nightmare to maintain :-)

Can't see any other way except for ignoring the issue though

PR #6334, opened 2025-02-17
Theme palettes warn about options

### LINKED ISSUE #6333: Should `theme(palette.* = )` raise a warning when `options(ggplot2.* = )` is set?
I read the comments on https://github.com/tidyverse/ggplot2/pull/5946 and it's for the sake of compatibility that the new `theme(palette.* = )` is enabled only when the corresponding `palette` is `NULL`. But, I feel it's a bit confusing to see `theme(palette.* = )` is silently ignored.

Especially, if I were a theme developer, I will be afraid of inquiries like "Hey, your theme doesn't work on my machine, Why?" from those who set `options(ggplot2.*)`. I think it's a rare case and not a serious issue, but I'm wondering if it's friendlier to show a warning.

``` r
library(ggplot2)

  theme(
    palette.colour.discrete = scales::get_palette("mint")
  )

# theme is applied
p
```

``` r
options(ggplot2.discrete.colour = "viridis")

# ignored
p
```

<sup>Created on 2025-02-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6333.

Briefly, `theme()` checks whether colour palettes have been set simultaneously in the options and emit a warning when they are.

RE: my earlier comment:
> Mechanically, would you propose to detect this situation whenever `theme()` is called, or when `scale_colour_*()` is called?

I think `theme()` is the more appropriate place because:
* `theme()` has access to the theme palette *and* the options
* `scale_colour_*()` never gets access to the theme, and the alternative would be `ScalesList$set_palettes()`, at which point it would be too late the change the entire scale rather than just the palette.
* I imagine `options()` is more used in a global fashion than `theme()`. The case that we'd be missing now is if we do a `theme_set()` with a palette before `options()`.
* An alternative is to do this in `check_theme()`, which runs every render.

Using theme without options is silent:
``` r
#> ℹ Loading ggplot2

  geom_point()

p + theme(palette.colour.discrete = c("dodgerblue", "orange", "orchid"))
```

Using options without theme settings is silent:
``` r
options(ggplot2.discrete.colour = c("red", "yellow", "limegreen"))
p
```

But using both emits a warning.
``` r
p + theme(palette.colour.discrete = c("dodgerblue", "orange", "orchid"))
#> Warning: The `options('ggplot2.discrete.colour')` setting is incompatible with the
#> `palette.colour.discrete` theme setting.
#> ℹ You can set `options(ggplot2.discrete.colour = NULL)`.
```

<sup>Created on 2025-02-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/theme.R
```diff
@@ -469,6 +469,27 @@ theme <- function(...,
                   validate = TRUE) {
   elements <- find_args(..., complete = NULL, validate = NULL)
 
+  elements <- fix_theme_deprecations(elements)
```
teunbrand, line 472: I've separated this in its own function as all these deprecations became somewhat distracting in ascertaining what `theme()` actually does.

thomasp85 (approved):
LGTM

PR #6336, opened 2025-02-18
Layer parameter controlling facet layout

### LINKED ISSUE #3062: Consider adding "static" parameter to geom/stat functions

teunbrand:
This PR aims to (eventually) fix #3062.

Briefly, it adds a `layer(layout)` argument to control how data is assigned to panels. Currently only `layout = "fixed"` is implemented, which repeats data over every panel. This PR is considered WIP because there might be other strategies that might also be interesting to discuss.

``` r
#> ℹ Loading ggplot2

  geom_point(colour = "grey", layout = "fixed") +
  geom_point() +
  facet_grid(year ~ drv)
```

<sup>Created on 2025-02-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/facet-.R
```diff
@@ -88,7 +88,95 @@ Facet <- ggproto("Facet", NULL,
...
+    # If we need to fix rows or columns, we make the corresponding faceting
+    # variables missing on purpose
+    if (grid_layout) {
+      if (identical(layer_layout, "fixed_rows")) {
+        facet_vals <- facet_vals[setdiff(names(facet_vals), names(params$cols))]
+      }
+      if (identical(layer_layout, "fixed_cols")) {
+        facet_vals <- facet_vals[setdiff(names(facet_vals), names(params$rows))]
+      }
+    }
```
teunbrand, line 141: This part is new

R/facet-.R
```diff
@@ -88,7 +88,95 @@ Facet <- ggproto("Facet", NULL,
...
+    layout[] <- lapply(layout, as_unordered_factor)
+
+    # Add PANEL variable
+    keys <- join_keys(facet_vals, layout, by = names(vars))
+    data$PANEL <- layout$PANEL[match(keys$x, keys$y)]
+
+    # Filter panels when layer_layout is an integer
+    if (is_integerish(layer_layout)) {
+      data <- vec_slice(data, data$PANEL %in% layer_layout)
+    }
```
teunbrand, line 177: This part is new

R/facet-.R
```diff
@@ -88,7 +88,95 @@ Facet <- ggproto("Facet", NULL,
...
+      return(data)
+    }
+
+    # Compute faceting values
+    facet_vals <- eval_facets(vars, data, params$.possible_columns)
+
+    include_margins <- !isFALSE(params$margin %||% FALSE) &&
+      nrow(facet_vals) == nrow(data) && grid_layout
+    if (include_margins) {
+      # Margins are computed on evaluated faceting values (#1864).
```
teunbrand, line 118: I merged the wrap/grid layout mapping because it reduces duplication and the only tangible difference was dealing with margins.

R/facet-.R
```diff
@@ -88,7 +88,95 @@ Facet <- ggproto("Facet", NULL,
...
+    }
+
+    grid_layout <- all(c("rows", "cols") %in% names(params))
+    layer_layout <- attr(data, "layout")
+    if (identical(layer_layout, "fixed")) {
+      n <- vec_size(data)
+      data <- vec_rep(data, vec_size(layout))
+      data$PANEL <- vec_rep_each(layout$PANEL, n)
+      return(data)
+    }
```
teunbrand, line 110: This is new

thomasp85 (approved):
I do like that logic for grid and wrap has been merged but I don't think it should live as a method in the virtual Facet class.

It assumes existence of properties in params that is not "global"

Can we move this to a free function that both `facet_grid()` and `facet_wrap()` calls?

PR #6340, opened 2025-02-19
Don't prepopulate legend key sizes

### LINKED ISSUE #6339: Legend size adjustment within guides
Hi!

When I tried to adjust the `legend.key.size` in guides it doesn't seem to work as expected. But `legend.key.spacing.y` works fine.
When I include it in the theme(adjusting the whole legend?) , the code works very well.

I'm not sure if this is similar to the question below? Or is it a bug? If this behavior is indeed due to an underlying mechanism, wouldn’t it be preferable for the documentation to include a note about it, as is done for this issue?
[https://github.com/tidyverse/ggplot2/issues/6263](https://github.com/tidyverse/ggplot2/issues/6263)

Here is the code to reproduce the issue:

```r
library(tidyverse)

data = tibble(x = rep(c('a','b'),each = 10),
              y = runif(20,min = 0, max=10))

ggplot(data = data, aes(x = x, y = y, fill = x))+
  geom_boxplot()+
  guides(fill = guide_legend(theme = theme(legend.key.spacing.y = unit(5,'lines'),
                                           legend.key.size = unit(3, 'lines')))
         )

```

Image

```r
ggplot(data = data, aes(x = x, y = y, fill = x))+
  geom_boxplot()+
  guides(fill = guide_legend(theme = theme(legend.key.spacing.y = unit(5,'lines')))
  )+
  theme(legend.key.size = unit(3, 'lines'))
```

Image

teunbrand:
This PR aims to fix #6339.

Briefly, it no longer prepopulates `legend.key.width` and `legend.key.height` from  the global theme and instead uses the default path of computing these, which includes the local theme.

``` r
#> ℹ Loading ggplot2
big_keys <- theme(legend.key.size = unit(3, "cm"))

  geom_point() +
  guides(colour = guide_legend(theme = big_keys))
```

<sup>Created on 2025-02-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
LGTM

PR #6343, opened 2025-02-21
POC: Position guides have acces to layer data

### LINKED ISSUE #6247: Expose position guides to titles
It would be nice if axes also can read layer data for some purposes.
The non-position guides already have this access.

teunbrand:
This PR aims to fix #6247.

Briefly, it adds the plumbing to let panel data reach the `get_layer_key()` method of position guides.

I've put this in as a draft, because I'm not entirely sure that we should take this approach.
It has some overhead because it needs to split up data by panel, making code slower.
We can illustrate with this benchmark:

``` r
#> ℹ Loading ggplot2

  geom_point() +
  facet_grid(year ~ cyl)

bench::mark(ggplotGrob(p), min_iterations = 10)
#> disabled.
#> # A tibble: 1 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 ggplotGrob(p)    101ms    105ms      6.51    10.8MB     11.7
```

<sup>Created on 2025-02-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The exact same code on the main branch gives this result:

```r
#> # A tibble: 1 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 ggplotGrob(p)   81.3ms   82.3ms      7.71    10.5MB     12.3
```

To illustrate that it works, we can assemble a quick-and-dirty guide extension that displays a density line for the data it sees, instead of the actual axis line.

``` r
#> ℹ Loading ggplot2

GuideDensity <- ggproto(
  "GuideDensity", GuideAxis,
  get_layer_key = function(params, layers, data = NULL, theme = NULL) {
    vals <- unlist(lapply(data, function(d) d[[params$aesthetic]]))
    dens <- density(vals, from = min(vals), to = max(vals))
    params$decor <- data.frame(x = dens$x, z = dens$y / max(dens$y))
    params
  },
  build_decor = function(decor, grobs, elements, params) {
    element_grob(
      elements$line,
      x = unit(decor$x, "npc"),
      y = unit(1 - decor$z, "npc")
    )
  }
)

  geom_point() +
  guides(x = new_guide(super = GuideDensity, available_aes = "x")) +
```

<sup>Created on 2025-02-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/coord-.R
+++ b/R/coord-.R
@@ -163,8 +163,8 @@ Coord <- ggproto("Coord",
     guide_params[!empty] <- Map(
       function(guide, guide_param, scale) {
         guide_param <- guide$train(guide_param, scale)
+        guide_param <- guide$get_layer_key(guide_param, layers, data)
         guide_param <- guide$transform(guide_param, self, panel_params)
-        guide_param <- guide$get_layer_key(guide_param, layers)
         guide_param
       },
       guide = guides[!empty],
--- a/R/coord-.R
+++ b/R/coord-.R
@@ -147,7 +147,7 @@ Coord <- ggproto("Coord",
     panel_params
   },
 
-  train_panel_guides = function(self, panel_params, layers, params = list()) {
+  train_panel_guides = function(self, panel_params, layers, params = list(), data = NULL) {
 
     aesthetics <- c("x", "y", "x.sec", "y.sec")
 
--- a/R/coord-map.R
+++ b/R/coord-map.R
@@ -249,7 +249,7 @@ CoordMap <- ggproto("CoordMap", Coord,
     panel_params
   },
 
-  train_panel_guides = function(self, panel_params, layers, params = list()) {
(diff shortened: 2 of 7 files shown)


teunbrand:
In addition, it is not entirely clear what data the guide should recieve in terms of facetting and free scales. For example, in grid facets, an x-axis guide for the first column should perhaps see all the data in that column, but that doesn't quite hold for `facet_wrap()`. All in all I'm not convinced by the goal here.

teunbrand:
I don't think it is possible to satisfactorily allow x/y position guides access to the data.

PR #6345, opened 2025-02-24
Background for `theme_void()` and `theme_minimal()`

### LINKED ISSUE #6251: Should theme_void() include paper argument? ink argument not working with geoms
Thank you for the excellent work in the theming arena! 🚀

Just opening an issue to note that having the 'paper' argument for `theme_void()` might be misleading. 

It also seems that the ink declaration doesn't seem to propagate to the geoms.

``` r
library(ggplot2)
p <- ggplot(mtcars, 
            aes(x = factor(cyl), 
                fill = factor(cyl))) +
  geom_bar() + 
  geom_point(stat = StatCount, size = 7) + 
  geom_tile(aes(y = -2, fill = NULL))

p +
  theme_void(ink = "plum", paper = "cornsilk", base_size = 18)
```

``` r

last_plot() + 
  theme(geom = element_geom(ink = "plum"))
```

<sup>Created on 2024-12-18 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6251.

Briefly, some themes had `plot.background = element_blank()` (explicitly or implicitly), which led to counterintuitive colouring.
By providing this element we surprise less.

Examples:

``` r
#> ℹ Loading ggplot2

  geom_point()

p + theme_minimal(paper = "cornsilk")
```

``` r
p + theme_void(paper = "cornsilk")
```

<sup>Created on 2025-02-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/theme-defaults.R
+++ b/R/theme-defaults.R
@@ -559,7 +559,10 @@ theme_void <- function(base_size = 11, base_family = "",
   # Only keep indispensable text: legend and plot titles
   t <- theme(
     line =               element_blank(),
-    rect =               element_blank(),
+    rect =               element_rect(
+                           fill = paper, colour = NA, linewidth = 0, linetype = 1,
+                           inherit.blank = FALSE
+                         ),
     polygon =            element_blank(),
     point =              element_blank(),
     text =               element_text(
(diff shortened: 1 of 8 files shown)


R/theme-defaults.R
```diff
@@ -553,13 +553,16 @@ theme_void <- function(base_size = 11, base_family = "",
...
   half_line <- base_size / 2
 
   # Only keep indispensable text: legend and plot titles
   t <- theme(
     line =               element_blank(),
-    rect =               element_blank(),
+    rect =               element_rect(
+                           fill = paper, colour = NA, linewidth = 0, linetype = 1,
+                           inherit.blank = FALSE
+                         ),
```
thomasp85, line 565: Curious why you are not just setting `plot.background` to a non-blank element?

teunbrand: You're right to wonder about this! Some test was protesting that we couldn't inherit from a blank root element. I then tried `plot.background = element_rect(..., inherit.blank = FALSE)` , but that gets overruled by `theme(..., complete = TRUE)`. I then tried using `+theme(...)` at the end of the function to force `inherit.blank = FALSE` without being bound to `complete = TRUE`, but this felt like a dirty solution. I've landed on  the current solution because at least it is straightforward.

thomasp85: that makes sense (in its own little way)

thomasp85 (approved):
LGTM

Issue #6352, opened 2025-03-04
Adopting S7 OOP


teunbrand:
This issue is to put to writing an idea that Thomas and I have discussed off-GitHub.

There are some issues with the S3 OOP system that cannot be resolved in a satisfying way, notably #3986, #5536 and some additional plans that'd require double dispatch. The solution I'd like to propose is to use S7 OOP instead. This has been attempted before in #5537, which we put on hold until S7 was more mature. I think that might be the case.

We can expect some reverse dependency failures when we adopt S7 and it may make sense to bump the major version due to this. Possibly, we can mitigate a series of failures by implementing `$`-methods that are list-like.

teunbrand:
Listing what systems/classes might adopt S7:

- [ ] The ggplot class
- [ ] The theme class
- [ ] The element classes
- [ ] The uneval class (from `aes()`)
- [ ] The labels class (from `labs()`)

Please feel free to comment more that I might've forgotten.

teunbrand:
So I'm running into a crinkle with regards to naming conventions. Essentially, the guidance from S7 is that classes should have the same name as the object that stores the class e.g.:

```r
foo <- S7::new_class("foo")
```

However we have several instances where we'd clash between a generic constructor (think `ggplot()` that has a 'default' and 'function' method) and a class object. My proposal is that we use a `class_`-prefix to clearly denote when using the S7 class. As an example, we'd use `class_ggplot <- S7::new_class("ggplot", ...)` and let `ggplot()` remain a function that constructs on object using `class_ggplot()`.

teunbrand:
Considered fixed by #6364

PR #6355, opened 2025-03-07
S7 theme elements

teunbrand:
This PR aims to fix part of #6352.

In essence it converts all S3 `element_*()` functions and `margin()` to S7 classes.

As it stands, there are some backwards compatibility issues for extensions:

* Testing wether an object is a particular type element with e.g. `inherits(x, "element_blank")` will no longer work. From 4.3.0 onwards, using `inherits(x, element_blank)` will work, and `S7::S7_inherits(x, element_blank)` should work. Alternatively, one could in theory use `inherits(x, c("element_blank", "ggplot2::element_blank"))` for backward compatibility, but it'd require a change in the extension.
* Typical extraction with `$`, `[` or `[[` of properties will no longer work and extensions will have to use `@`. On ggplot2's end, we can provide methods for the typical generics, but we'd run into https://github.com/RConsortium/S7/issues/390.
* Extensions implementing subclasses like `marquee::element_marquee()` or `ggtext::element_markdown()` will also have to write their elements as S7 classes. I cannot see a good solution to have this be backwards compatible.


teunbrand:
A few things to consider:

* For the `inherit(x, "element_blank")` problem, we could consider tagging on an additional class to the S7 objects, without the `ggplot2::` prefix.
* If it turns out there is too much broken in reverse dependencies, we could retain S3 classes for a cycle, implement `@.element` so that extensions can adapt before we commit to S7 element classes.

teunbrand:
An update:
* With the 3.5.2. release we now have `is_theme_element()` that will allow graceful transition for `inherit(x, "element_blank")` and such.
* I've introduced the `[`, `[[` and `$` operators for elements, with the plan that they'll be deprecated the subsequent release cycle. This is also to allow for graceful transition.

With that, I think 90% of the worries are gone. Packages that implement their own elements will still need to update, but that should be a manageable number of packages. As such, this PR is ready for review.

R/geom-label.R
```diff
@@ -88,7 +88,7 @@ GeomLabel <- ggproto("GeomLabel", Geom,
     data <- coord$transform(data, panel_params)
     data$vjust <- compute_just(data$vjust, data$y, data$x, data$angle)
     data$hjust <- compute_just(data$hjust, data$x, data$y, data$angle)
-    if (!is_margin("margin")) {
```
thomasp85, line 91: What on earth was this supposed to do?

R/theme-elements.R
```diff
@@ -81,55 +81,86 @@ NULL
...
+  hjust      = property_nullable(S7::class_numeric),
+  vjust      = property_nullable(S7::class_numeric),
+  angle      = property_nullable(S7::class_numeric),
+  size       = property_nullable(S7::class_numeric),
+  lineheight = property_nullable(S7::class_numeric),
+  margin     = property_nullable(margin),
+  face       = property_choice(c("plain", "bold", "italic", "oblique", "bold.italic"), allow_null = TRUE),
+  linewidth  = property_nullable(S7::class_numeric),
+  linetype   = property_nullable(S7::class_numeric | S7::class_character),
+  lineend    = property_choice(c("round", "butt", "square"), allow_null = TRUE),
```
thomasp85, line 107: We should probably add linejoin while we are doing the rewrite

R/theme-elements.R
```diff
@@ -276,6 +329,45 @@ rel <- function(x) {
   structure(x, class = "rel")
 }
 
+#' @export
+`$.ggplot2::element` <- function(x, i) {
+  # deprecate_soft0("4.0.0", I("`<ggplot2::element>$i`"), I("`<ggplot2::element>@i`"))
```
thomasp85, line 334: We would wait for the 4.1.0 release with this, right?

thomasp85 (approved):
LGTM

R/geom-label.R
teunbrand: If I'm tracing this correctly, this is a merge scar from 3.5.2.

R/theme-elements.R
teunbrand: That's right, let me put the right version in these staged deprecations

R/theme-elements.R
teunbrand: I'll happily add that

teunbrand:
Closing, because merged with #6364

PR #6364, opened 2025-03-17
Reimplement classes in S7

teunbrand:
This PR aims to fix a large part of #6352.

Briefly it implements the following S7 classes:
- class_ggplot
- class_ggplot_built
- class_mapping
- class_labels
- class_theme

I've put in the usual extractors and replacers (`[`, `[[`, `$` and their `<-` methods) for the ggplot/ggplot_built classes for backwards compatibility. The code internally now uses `@` though, so that we can deprecate the usual extractors/replacers later.


R/aes.R
```diff
@@ -130,14 +129,9 @@ new_aesthetic <- function(x, env = globalenv()) {
...
 }
-new_aes <- function(x, env = globalenv()) {
-  check_object(x, is.list, "a {.cls list}")
-  x <- lapply(x, new_aesthetic, env = env)
-  structure(x, class = "uneval")
-}
 
 #' @export
-print.uneval <- function(x, ...) {
+`print.ggplot2::mapping` <- function(x, ...) {
```
teunbrand, line 134: When we have an S3 generic we need for both S3 and S7 objects, we need to implement methods for the S7 objects in the S3 fashion, because using `S7::method()` will invalidate the S3 methods.

teunbrand:
Older versions of R seem to get stuck on dispatch with `+`. We can get around this by implementing this `+` as S7 methods only (i.e. converting the `+.gg` method), but it makes error messages less informative. For example: `Cannot use + with a single argument. i Did you accidentally put + on a new line?` will become `invalid argument to unary operator` and also `Cannot add <ggproto> objects together.  i Did you forget to add this object to a <ggplot> object?` becomes `non-numeric argument to binary operator`.

R/all-classes.R
```diff
@@ -0,0 +1,177 @@
+
+class_gg     <- S7::new_class("gg", abstract = TRUE)
+class_S3_gg  <- S7::new_S3_class("gg")
```
teunbrand, line 3: Add note to remove at some point

teunbrand:
TODO: Remove class gymnastics

R/layer-sf.R
```diff
@@ -45,7 +45,7 @@ LayerSf <- ggproto("LayerSf", Layer,
     # automatically determine the name of the geometry column
     # and add the mapping if it doesn't exist
     if ((isTRUE(self$inherit.aes) && is.null(self$computed_mapping$geometry) &&
-         is.null(plot$computed_mapping$geometry)) ||
+         is.null(self$computed_mapping$geometry)) ||
```
thomasp85, line 48: nope - read the whole statement again :-)

R/layer.R
```diff
@@ -304,7 +302,7 @@ Layer <- ggproto("Layer", NULL,
 
     # Evaluate aesthetics
     evaled <- eval_aesthetics(aesthetics, data)
-    plot$scales$add_defaults(evaled, plot$plot_env)
+    plot@scales$add_defaults(evaled, plot@plot_env)
```
thomasp85, line 305: no fault of yours, but this mix of `@` and `$` indexing is really aesthetically unpleasant

R/plot-construction.R
```diff
@@ -61,10 +60,30 @@
...
+  e2name <- deparse(substitute(e2, env = caller_env(2)))
+  add_theme(e1, e2, e2name)
+}
+
 
 #' @rdname gg-add
 #' @export
-"%+%" <- `+.gg`
+"%+%" <- function(e1, e2) {
+  if (getRversion() < "4.3.0") {
```
thomasp85, line 81: maybe add a note about deprecation?

thomasp85 (changes requested):

R/layer-sf.R
teunbrand: A right, now we're testing this twice. Though now `plot$computed_mapping` cannot reasonably exist, so we can best skip this.

R/layer.R
teunbrand: Agreed. It wouldn't be that much of a problem to implement `@.ggproto` I think, so we can eliminate one source of annoyance.

teunbrand:
#6355 is now merged in via https://github.com/teunbrand/ggplot2/pull/16

R/plot-construction.R
```diff
@@ -97,109 +115,100 @@ add_ggplot <- function(p, object, objectname) {
...
+    for (o in object) {
+      plot <- ggplot_add(o, plot, object_name)
+    }
+    plot
+  }
+
+S7::method(ggplot_add, list(S7::new_S3_class("by"), class_ggplot)) <-
+  function(object, plot, object_name, ...) {
+    ggplot_add(unclass(object), plot, object_name)
+  }
```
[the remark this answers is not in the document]
teunbrand: There is an example in `?ggplot_add` for adding new classes of objects. Do you mean for using subclasses of <ggplot> objects?

teunbrand:
Re https://github.com/tidyverse/ggplot2/pull/6364#discussion_r2088298746 (can't reply there for some reason):

> Do you know the timeline for resolving this?

It's been tagged with the S7 v0.3.0 milestone here: https://github.com/RConsortium/S7/issues/390. I don't know anything about the specifics of the timeline though.

R/aes.R
teunbrand: It's been tagged with the S7 v0.3.0 milestone here: https://github.com/RConsortium/S7/issues/390. I don't know anything about the specifics of the timeline though.

R/layer-sf.R
```diff
@@ -44,13 +44,9 @@ LayerSf <- ggproto("LayerSf", Layer,
 
     # automatically determine the name of the geometry column
     # and add the mapping if it doesn't exist
-    if ((isTRUE(self$inherit.aes) && is.null(self$computed_mapping$geometry) &&
-         is.null(plot$computed_mapping$geometry)) ||
-        (!isTRUE(self$inherit.aes) && is.null(self$computed_mapping$geometry))) {
-      if (is_sf(data)) {
-        geometry_col <- attr(data, "sf_column")
-        self$computed_mapping$geometry <- sym(geometry_col)
-      }
```
teunbrand, line 53: If we define `A` as `self$inherit.aes` and `B` as `self$computed_mapping$geometry`, which the new class slots we'd be testing the following:

```r
if ((isTRUE(A) && is.null(B) && is.null(B)) || (!isTRUE(A) && is.null(B)))
```
which simplifies to:
```r
if ((isTRUE(A) && is.null(B)) || (!isTRUE(A) && is.null(B)))
```
which simplifies to:
```r
if ((isTRUE(A) || !isTRUE(A)) && is.null(B))
```
Which cancels out the OR operation because it will always return `TRUE`, thus we're left with:
```r
is.null(B)
```
I don't quite understand why the logic was hard to follow in the first place.

R/plot-construction.R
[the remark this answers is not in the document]
thomasp85: ah, maybe I missed it then...

R/layer.R
teunbrand: I hear you

R/layer-sf.R
thomasp85: oh, so I think actually it should have been `plot$mapping$geometry` and not `plot$computed_mapping$geometry`... but the global mapping has already conditionally been merged into `self$computed_mapping` so none of this even matters

you are right that it can all be simplified to `is.null(self$computed_mapping$geometry)`

teunbrand:
I'm slightly annoyed that we have a `class_`-prefix for most classes now, e.g. `class_ggplot`, but not for the theme elements which are still `element_line` etc.

R/layer-sf.R
teunbrand: Yeah thanks for pointing out something was amiss in the logic here, this cleaned up really nicely ☺️

thomasp85:
I'm open to suggestions, but can also live a long and happy life with the discrepancy

teunbrand:
Alright let's not worry about it too much then :)

thomasp85 (approved):
LGTM - do you want a revdepcheck before merging?

teunbrand:
I'm having a hard time getting this to go over smoothly with some widely used packages, or even my own extensions for that matter. Perhaps we need some aggressive backward compatibility for `ggplot_build()`, `ggplot_gtable()`, `ggplot_add()` and `element_grob()`.

teunbrand:
I think it is necessary to make `ggplot()` a S3 generic again. I have the following reasons:

1. Extension have S3 methods for the S3 generic. 
    * You cannot write S3 methods for an S7 generic.
    * There are too many extensions to force them all to use S7 right now.
1. Unlike S3 `ggplot_build()`, which we plan to replace with S7 `build_ggplot()`, the `ggplot()` function is much more user-facing. I don't think we want to phase-out `ggplot()` for other function names.
1. If people want to use S7 methods, they can register them for the S3 generic without any issues.

R/plot-construction.R
```diff
@@ -61,10 +61,29 @@
...
+
 
 #' @rdname gg-add
 #' @export
-"%+%" <- `+.gg`
+"%+%" <- function(e1, e2) {
+  if (getRversion() >= "4.3.0") {
+    deprecate_soft0("4.0.0", I("<ggplot> %+% x"), I("<ggplot> + x"))
+  }
+  agg_gg(e1, e2)
```
schloerke, line 85: Maybe??

```suggestion
  add_gg(e1, e2)
```

teunbrand:
This is my take on the revdepcheck as it stands.

Of the 648 packages that I've annotated:

* 91 fail because they have tests that pertain to classes in ggplot2. Most of these tests are of poor quality, like testing the `class()` vector itself instead of using `inherits()` or `expect_s3_class()` or `is_ggplot()`. This is very much an S7-transition related failure.
* 69 fail due to errors in vignettes rather than tests or examples. I've not looked into these in much detail.
* 64 are due to plotly, which is not due to the S7-transition. Relevant changes were addressed here:
https://github.com/plotly/plotly.R/pull/2442
* 63 are due to GGally, which was due to S7-transition, but I've sent out a PR: 
https://github.com/ggobi/ggally/pull/528
* 43 failures are unknown, which just means that I don't understand them fully
* 32 packages do invalid stuff with labels, like `labs(y = element_blank())`. This is related to the S7 transition as we're now more strict about such things.
* 31 packages test for labels that are no longer there. This is not related to the S7 transition.
* 25 packages are due to patchwork's `&` operator that doesn't quite work well out of the box.
* 23 packages have namespace conflicts, like `sjPlot::set_theme()` or `e1071::element()`. Some of them are due to S7 transition, some of them aren't. In any case, these packages should probably be more careful with their imports to avoid conflicts.
* 23 packages are probably false positives that get picked up because they emit warnings during package building.
* 15 packages set properties incorrectly, like `element_text(family = TRUE, colour = 12)`. Technically using an integer for colour should index the base graphics palette, but I don't think this should work. This is due to the S7 transition because we are more strict.
* 12 packages use functions that we've fully deprecated now. This isn't due to the S7 transition.
* 11 packages use internals that have changed, which isn't related to the transition.
* 10 packages construct incorrect margins, like `margin(c(1, 2, 3, 4))` instead of `margin(1, 2, 3, 4)`. This is an S7-transition thing, as we're now more strict about this.
* 8 packages have `na.value` that cannot be combined with palette values, which isn't related to S7 transition.
* 5 packages are due to qqplotr, haven't dig too deeply yet, but I'll send a PR.
* 5 packages fail because they manually construct `structure(..., class = "uneval")` instead of `aes()`. This is due to S7 transition, because we now have a `<mapping>`  class.
* 5 packages fail because their test don't expect `plot$layers` to be named. This isn't related to the S7-transition.
* 5 packages fail because they use ggtern.
* 4 packages don't have a tibble import, whereas they should.
* 4 packages use ggplot2 in very non-standard ways
* 4 packages expect to be able to `as.list(<ggplot>)`, whereas they couldn't at the time of the check.
* 4 packages fail due to ggiraph, for which I've made a PR:
https://github.com/davidgohel/ggiraph/pull/319
* 4 packages use `aes(label = <expression>)`, which they shouldn't.
* 4 packages had problems that I couldn't locally reproduce.
* 4 packages were in trouble to to ggtext.
* 3 packages do `ifelse(test, <plot>, ...)` istead of `if (test) {<plot>} else {...}`. This worked because ggplot objects were lists, i.e. vectors, but they aren't anymore.
* 3 packages due to ggparty.
* 3 packages due to gganimate.
* 3 packages improperly access geom defaults, they should use `get_geom_defaults()`.
* Various other problems that are infrequent or specific to single packages that I won't all discuss here.

cpsievert:
Please bump the version to something like `3.5.2.9001`, so we can switch logic depending on the version

teunbrand:
I don't see how we can further minimise the impact of this PR without abandoning the S7 transition altogether. I'll merge this into main so I can send out some PRs and instruct to check with the dev version, which is easier than check with a PR.

PR #6371, opened 2025-03-20
Use innate plot size in `ggsave()`

### LINKED ISSUE #6370: Easy way to use absolute plot size in `ggsave()`
This is a sub-issue of #6367.

Briefly, when using `theme(panel.widths, panel.heights)`, the plot has a fixed dimension as the panels no longer expand to fill the available space. It would be neat if `ggsave()` would understand a plot with fixed dimensions and make it easier to inform the output file dimensions.

teunbrand:
This PR aims  to fix #6370.

Briefly, `plot_dim()` receives `width` and `height` separately in addition to `plot`, to be able to use absolute plot dimensions for plot output. This is achieved by using the `derive()` flag for `width` and/or `height`.

In plot below, the `width` is absolute, whereas `height` is not. By using `derive()` we set the absolute width of the plot, but as the height is not absolute, we fall back to the device size.

``` r
#> ℹ Loading ggplot2

  geom_point() +
  theme(panel.widths = unit(10, "in"))

file <- tempfile(fileext = ".png")
ggsave(file, plot = p, width = derive(), height = derive())
#> Saving 10.6 x 5 in image

knitr::include_graphics(file)
```

<sup>Created on 2025-03-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/save.R
+++ b/R/save.R
@@ -34,6 +34,9 @@
 #' @param scale Multiplicative scaling factor.
 #' @param width,height Plot size in units expressed by the `units` argument.
 #'   If not supplied, uses the size of the current graphics device.
+#'   Alternatively, these can be set to `derived()` in order to use innate
+#'   plot dimensions for output. This is useful when the
+#'   `theme(panel.widths, panel.heights)` options are set to absolute units.
 #' @param units One of the following units in which the `width` and `height`
 #'   arguments are expressed: `"in"`, `"cm"`, `"mm"` or `"px"`.
 #' @param dpi Plot resolution. Also accepts a string input: "retina" (320),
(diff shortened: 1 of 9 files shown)


R/save.R
```diff
@@ -189,12 +196,48 @@ parse_dpi <- function(dpi, call = caller_env()) {
...
+        {.obj_type_friendly {plot}}.",
+        call = call
+      )
+    }
+    width  <- if (is.derived(width))  gtable_width(plot)  else width
+    height <- if (is.derived(height)) gtable_height(plot) else height
+  }
+
+  if (is.unit(width)) {
+    if (has_null_unit(width)) {
```
teunbrand, line 223: The assumption here is that if the plot dimensions contain `"null"` units, the dimension is relative. If these are absent, the plot is assumed to have absolute dimensions. 

It is possible for facet or axis guide extensions to break this assumption (as did `facet_null()`).

thomasp85:
I feel like we have been here before? You cannot know how much space a graphics device needs for text so you cannot apriori know the size needed for a plot with fixed panel size. Also, width and height might be linked, either because of a fixed aspect ratio or because one depend on the either due to e.g. text wrapping.

Unless this has magically been fixed I strongly feel ggplot2 shouldn't make promises about being able to do this

PR #6375, opened 2025-03-21
Checks for bandwidth in `stat_density_2d()`

### LINKED ISSUE #6374: stat_density_2d error messages uninformative
I found an error thrown by stat_density_2d not very informative. It computes an illegal bandwidth for me behind the scene, which causes an internal error that is not explained in the messages. Specifying the bandwidth explicitly can fix the problem. However, I expect stat_density_2d can either handle these edge cases or point out that the default value given the data is illegal and manual input is required.

The following example works fine:
```r
library(ggplot2)
df <- data.frame(x=sample(0:10, 100, replace=T), y=rep(0:10, 100, replace=T))
ggplot(df) + stat_density_2d(geom='density_2d', mapping=aes(x,y))
```
but the next one will throw an error:
```r
df <- data.frame(x=sample(0:10, 100, replace=T), y=c(rep(5, 80), sample(0:10, 20, replace=T)))
ggplot(df) + stat_density_2d(geom='density_2d', mapping=aes(x,y))
```

> Error in `stat_density_2d()`:
> ! Problem while computing stat.
> ℹ Error occurred in the 1st layer.
> Caused by error in `seq_len()`:
> ! argument must be coercible to non-negative integer

The error messages is quite confusing.  By digging into the warnings, I found the root cause of the problem:

> 1: Computation failed in `stat_density2d()`
> Caused by error in `MASS::kde2d()`:
> ! bandwidths must be strictly positive 

In stat_density_2d, h is automatically computed before calling kde2d, if not given
```r
if (is.null(h)) {
  h <- c(MASS::bandwidth.nrd(data$x), MASS::bandwidth.nrd(data$y))
  h <- h * adjust
}

# calculate density
dens <- MASS::kde2d(
  data$x, data$y, h = h, n = n,
  lims = c(scales$x$dimension(), scales$y$dimension())
)
```
and bandwidth.nrd uses the following formula by default
```r
function(x)
{
    r <- quantile(x, c(0.25, 0.75))
    h <- (r[2] - r[1])/1.34
    4 * 1.06 * min(sqrt(var(x)), h) * length(x)^(-1/5)
}
```
So if one of data$x and data$y has more than 75% of identical values, defualt bandwidth will become 0 without warning, and it will immediately be considered as illegal by kde2d.

teunbrand:
This PR aims to fix #6374

Briefly, it introduces some fallbacks and more informative error messages for calculating the bandwidth in `stat_density_2d()`.

The reprex from the issue now runs without errors, due to fallbacks when `h = 0` to use `stats::bw.nrd0()` instead of `MASS::bandwidth.nrd()`.

``` r
#> ℹ Loading ggplot2
set.seed(42)
df <- data.frame(
  x = sample(0:10, 100, replace = T), 
  y = c(rep(5, 80), sample(0:10, 20, replace = T))
)
ggplot(df, aes(x, y)) + stat_density_2d()
```

Moreover, if the bandwidth is nonsense it will throw an error (demoted to warning by stats).

``` r
ggplot(df, aes(x, y)) + stat_density_2d(h = c(0, NA))
#> Warning: Computation failed in `stat_density2d()`.
#> Caused by error in `precompute_2d_bw()` at ggplot2/R/stat-density-2d.R:185:5:
#> ! The bandwidth argument `h` must contain numbers larger than 0.
#> ℹ Please set the `h` argument to stricly positive numbers manually.
```

<sup>Created on 2025-03-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/stat-density-2d.R
+++ b/R/stat-density-2d.R
@@ -141,6 +141,9 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
     check_installed("MASS", reason = "for calculating 2D density.")
     # first run the regular layer calculation to infer densities
     data <- ggproto_parent(Stat, self)$compute_layer(data, params, layout)
+    if (empty(data)) {
+      return(data_frame0())
+    }
 
     # if we're not contouring we're done
     if (!isTRUE(params$contour %||% TRUE)) return(data)
--- a/R/stat-density-2d.R
+++ b/R/stat-density-2d.R
@@ -182,24 +182,7 @@ StatDensity2d <- ggproto("StatDensity2d", Stat,
   compute_group = function(data, scales, na.rm = FALSE, h = NULL, adjust = c(1, 1),
                            n = 100, ...) {
 
-    if (is.null(h)) {
-      # Note: MASS::bandwidth.nrd is equivalent to stats::bw.nrd * 4
-      h <- c(MASS::bandwidth.nrd(data$x), MASS::bandwidth.nrd(data$y))
-      # Handle case when when IQR == 0 and thus regular nrd bandwidth fails
-      if (h[1] == 0) {
-        h[1] <- bw.nrd0(data$x) * 4
-      }
-      if (h[2] == 0) {
-        h[2] <- bw.nrd0(data$y) * 4
-      }
-      h <- h * adjust
-    }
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #6383, opened 2025-03-27
`geom_raster()` fallback for non-linear Cartesian coordinates

### LINKED ISSUE #6382: No error reported but not displayed when using crs and default_crs parameter in coord_sf
Hi, I'm using ggplot to process some geospatial data and I'm finding that setting `crs` and `default_crs` doesn't produce the expected results and no errors are reported.
I'm not sure if it's because I don't fully understand these two parameters or if it's something else. 
- `crs` is supposed to determine the reference system for the final plot; 
- `default_crs` is used to specify the original reference system for the data.

In the example below, the raw data is in the `epsg:4326` coordinate system, and I'm trying to show it using a different projected coordinate system, but the plotting results are blank.

I provided data that could be reproduced: [example data](https://drive.google.com/file/d/1QQDhM83WhkAAO4cA8xvhk_QRqW6QiQ2l/view?usp=share_link)

```r
library(ggplot2)
library(sf)

data <- read.csv('./example.csv', row.names = 1)

targetcrs <- paste(
  "+proj=lcc", 
  "+lat_1=25", 
  "+lat_2=47", 
  "+lon_0=105", 
  "+datum=WGS84", 
  "+units=m", 
  sep = " "
)

ggplot(data = data, aes(x = x, y = y))+
  geom_raster(aes(fill = value))+
  facet_wrap(~class)+
  coord_sf(
    crs = st_crs('epsg:32649'), # or crs = targetcrs,
    default_crs = st_crs('epsg:4326'))
```

Image

teunbrand:
This PR aims to fix #6382 and amends #5627.

It kicks off the fallback mechanism too if the coord is non-linear. I hadn't foreseen at the time that `coord_sf()` is technically a non-linear child-coordinate of `coord_cartesian()`.

Reprex from the issue:

``` r
#> ℹ Loading ggplot2
library(sf)
#> Linking to GEOS 3.12.1, GDAL 3.8.4, PROJ 9.3.1; sf_use_s2() is TRUE

file <- file.choose()
data <- read.csv(file, row.names = 1)

targetcrs <- paste(
  "+proj=lcc", 
  "+lat_1=25", 
  "+lat_2=47", 
  "+lon_0=105", 
  "+datum=WGS84", 
  "+units=m", 
  sep = " "
)

ggplot(data = data, aes(x = x, y = y))+
  geom_raster(aes(fill = value))+
  facet_wrap(~class)+
  coord_sf(
    crs = st_crs('epsg:32649'),
    default_crs = st_crs('epsg:4326')
  )
#> `geom_raster()` only works with `coord_cartesian()`.
#> ℹ Falling back to drawing as `geom_rect()`.
#> `geom_raster()` only works with `coord_cartesian()`.
#> ℹ Falling back to drawing as `geom_rect()`.
#> `geom_raster()` only works with `coord_cartesian()`.
#> ℹ Falling back to drawing as `geom_rect()`.
```

<sup>Created on 2025-03-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/geom-raster.R
+++ b/R/geom-raster.R
@@ -90,7 +90,7 @@ GeomRaster <- ggproto("GeomRaster", Geom,
 
   draw_panel = function(self, data, panel_params, coord, interpolate = FALSE,
                         hjust = 0.5, vjust = 0.5) {
-    if (!inherits(coord, "CoordCartesian")) {
+    if (!inherits(coord, "CoordCartesian") || !coord$is_linear()) {
       cli::cli_inform(c(
         "{.fn {snake_class(self)}} only works with {.fn coord_cartesian}.",
         i = "Falling back to drawing as {.fn {snake_class(GeomRect)}}."


R/geom-raster.R
```diff
@@ -90,7 +90,7 @@ GeomRaster <- ggproto("GeomRaster", Geom,
 
   draw_panel = function(self, data, panel_params, coord, interpolate = FALSE,
                         hjust = 0.5, vjust = 0.5) {
-    if (!inherits(coord, "CoordCartesian")) {
+    if (!inherits(coord, "CoordCartesian") || !coord$is_linear()) {
```
thomasp85, line 93: Shouldn't it be enough to test `coord$is_linear()` and forget about the special case for `CoordCartesian`?

The message below also needs an update

teunbrand: Yeah good points, I agree

thomasp85 (approved):
LGTM

Breeze-Hu:
Great! 
Thanks : )

PR #6388, opened 2025-04-01
Forward compatibility: `is_*()` functions

teunbrand:
This PR aims to improve forwards compatibility.

Briefly, it re-implements parts of #6022 specifically the test functions. We deviate from that PR through deprecating dot.case functions and introducing snake_case function names are used instead.

Notably, `is_element(x, type = ....)` allows one to test for specific subtypes of elements as well.


R/theme-elements.R
```diff
@@ -144,6 +146,22 @@ element_text <- function(family = NULL, face = NULL, colour = NULL,
...
+    rect  = inherits(x, "element_rect"),
+    line  = inherits(x, "element_line"),
+    text  = inherits(x, "element_text"),
+    blank = inherits(x, "element_blank"),
+    # TODO: ideally we accept more elements from extensions. We need to
+    # consider how this will work with S7 classes, where ggplot2 doesn't know
+    # about the extension's class objects.
+    FALSE
+  )
+}
```
teunbrand, line 164: Thomas, I'd like to draw your eye to this bit with regard to the comment here

thomasp85: I don't think we _need_ to support anything but our own classes here tbh

thomasp85 (approved):

PR #6391, opened 2025-04-02
Simplify `zeroGrob()`

### LINKED ISSUE #6390: Replace `zeroGrob()` with `grid::nullGrob()`
To my understanding, ggplot2 use to have a `nullGrob()` function itself. At some point, {grid} also introduced a homonymous function, forcing ggplot2 to introduce `zeroGrob()` to avoid a namespace conflict (7f12f21f89ba51793bea9223c19e2a4fb02bd7b8). However, I think `grid::nullGrob()` does the same thing as `ggplot2::zeroGrob()`, namely: draw nothing. In principle, we could just use `grid::nullGrob()` instead.

teunbrand:
This PR aim to fix #6390.

Briefly, instead of implementing a full `<zeroGrob>` class, we piggyback on the `grid::nullGrob()`.
Simplification is the goal of this PR.


R/utilities-grid.R
```diff
@@ -48,6 +48,17 @@ gg_par <- function(..., stroke = NULL, pointsize = NULL) {
...
 }
 
+#' The zero grob draws nothing and has zero size.
+#'
+#' @keywords internal
+#' @export
+zeroGrob <- function() .zeroGrob
+
+.zeroGrob <- NULL
+on_load(.zeroGrob <- add_class(nullGrob(), "zeroGrob"))
```
teunbrand, line 58: The zeroGrob class is appended here, because `inherits(x, "zeroGrob")` is sometimes used instead of `is.zero()`. I expect the situation to be the same in extension packages.

R/utilities-grid.R
```diff
@@ -48,6 +48,17 @@ gg_par <- function(..., stroke = NULL, pointsize = NULL) {
   inject(gpar(!!!args))
 }
 
+#' The zero grob draws nothing and has zero size.
+#'
+#' @keywords internal
+#' @export
+zeroGrob <- function() .zeroGrob
```
teunbrand, line 55: I've not removed this function altogether because there is some optimisation value in only constructing the `.zeroGrob` object once per session instead of all over the place. Moreover, I expect it might be prevalent in extensions.

Issue #6396, opened 2025-04-05
Feature request: drawing ordering from layer to layer instead of panel to panel


Yunuuuu:
Currently, ggplot2 draws each facet panel sequentially, which means the complete drawing of one panel is finished before the next begins. This causes issues when a geom spans across multiple facet panels—for example, in cases like tree structures (e.g., dendrograms) that connect elements across panels.

![Image](https://github.com/user-attachments/assets/d02422aa-2377-4cc8-9898-c773fcb83e00)

Under the current rendering logic, parts of such layers can be hidden or clipped because layers are drawn panel-by-panel rather than layer-by-layer across panels. It's hard for us to create layer like above dendrogram. 

To address this, I’ve experimented with creating a custom grob that allows communication between different viewports. Specifically, I built a `channelGrob()` object that emits "signals" from different panels and connects them with a line once all signals are available:

```r
# pak::pak("Yunuuuu/ggalign")
library(ggalign)
# here we create a new channel, we will emit two singals
# when drawing: we just add a line between the two signals
channel <- channelGrob(function(locations) {
    loc1 <- .subset2(locations, 1L)
    loc2 <- .subset2(locations, 2L)
    grid::segmentsGrob(loc1$x, loc1$y, loc2$x, loc2$y)
})
gt <- gtable::gtable(unit(1:2, c("cm")), unit(5, "cm"))
gt <- gtable::gtable_add_grob(
    gt,
    list(
        grid::rectGrob(gp = gpar(color = "black", fill = NA)),
        channel$signal(0.5, 0.5, "npc")
    ),
    t = 1, l = 1, name = c("rect1", "signal1")
)
gt <- gtable::gtable_add_grob(
    gt,
    list(
        grid::rectGrob(gp = gpar(color = "red", fill = NA)),
        channel$signal(0.5, 0.5, "npc")
    ),
    t = 1, l = 2, name = c("rect2", "signal2")
)
grid::grid.newpage()
grid::grid.draw(gt)
```

![Image](https://github.com/user-attachments/assets/eabe5cec-eeeb-4f50-9d3f-6f4a4f8559f4)

This works in isolation, but integrating it into ggplot2’s layer system is quite difficult. The issue is that this kind of grob needs to wait for multiple signals before drawing, and it must be drawn after all contributing panels have emitted their signal. This behavior conflicts with ggplot2’s panel drawing order since it will override other geom layers that were already drawn in those panels

Yunuuuu:
[...]

PR #6398, opened 2025-04-05
theme gains a new element `panel.merge` to control whether to merge anel area into single gTree

Yunuuuu:
Try to fix https://github.com/tidyverse/ggplot2/issues/6396


Yunuuuu:
I'm not sure whether the current implementation will be accepted. I've reviewed the tests—while some changes are present, there don’t appear to be any visual differences. The current error is due to the vignette, which I plan to tweak once the implementation itself is accepted.

teunbrand:
Hi Yunuuuu,

Thanks for putting together the PR to fix the issue, however we feel that the added complexity of the code does not result in sufficient utility.
Moreover, several reverse dependencies (like patchwork) may also rely on the current panel structure. While I understand why you'd want a panel spanning dendrogram, I do not think viewport-based tricks will be the cleanest path to achieve what you want. Instead, I'll suggest that you implement a Facet subclass where you have control over placing panels and annotations. Moreover, because facets have access to scales, you should be able to derive the break locations of discrete scales, so you don't have to rely on these viewports.
Because we feel a superior alternative is available and we're reluctant to add this level of complexity for a niche feature, we're closing this PR.

Yunuuuu:
After testing, I found it difficult to handle this using a single Facet object alone, since the panels input to Facet$draw_panels() is unpredictable. It’s challenging to break down panels into individual layer grobs.

Several issues contribute to this unpredictability:

- The Layer$draw_geom method doesn't always return a grob if the developer doesn’t follow the expected conventions. No error or warning is issued in such cases, and `gList(NULL)` or `gList(gList())` simply works silently. 
- `Coord$draw_panel` will sometimes wrap all layer grobs into single gTree in `CoordRadial`, sometimes it won't

Solutions:
- Ensure that `Layer$draw_geom()` always returns a single `grob` or a `gTree`.
- always wrap the all layer grobs into a single `gTree` before passing to ``Coord$draw_panel`

PR #6399, opened 2025-04-07
Rename `borders()` to `annotation_borders()`

### LINKED ISSUE #6392: The grammarlessness of `borders()`
There is a layer constructor function called `borders()`. It draws geographical borders. While it is not a bad name for a function on itself, it doesn't really "fit" in the established grammar of ggplot2. For example, it is not immediately clear that `borders()` is dissimilar to a helper like `margins()` or is unrelated to `gt::cell_borders()`. It isn't on face-value related to any of the other grammar components, so it is not intuitive that it can be `+`-ed to a plot.

I think it should become something like `annotation_borders()`, where it is at least clear that it falls with the 'annotation'-flavour of layers.

teunbrand:
This PR aims to fix #6392.

In addition to renaming I:
* Removed one example that wasn't using the function
* Use `annotate("polygon")` instead of `geom_polygon()` to avoid propagating x/y labels


thomasp85 (approved):
LGTM

PR #6400, opened 2025-04-07
Deprecate the `optional_aes` field in Geom/Stat

### LINKED ISSUE #6393: Discourage use of `optional_aes` field
The `Geom$optional_aes` and `Stat$optional_aes` fields seem redundant to me.
Instead of having these fields, we can declare `default_aes = aes(optional_aesthetic = NULL)`.
While I would like to see these go in their entirety, that seems like backward incompatible, so instead I propose to discourage these fields. We can (i) remove any references to that field in the extension vignette (ii) ensure we adopt the NULL-aesthetic approach within ggplot2 and (iii) check if the ggplot book would need to be adapted.

teunbrand:
This PR aims to fix #6393.

It takes the 'violent' path of deprecating the field. 
We can also choose to delay deprecating the field.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* The `optional_aes` field in `Geom` and `Stat` classes is now deprecated in
+  favour of using `default_aes = aes(foo = NULL)` to mark the 'foo' aesthetic
+  as optional (@teunbrand, #6393).
 * `position_fill()` avoids stacking observations of zero (@teunbrand, #6338)
 * New `layer(layout)` argument to interact with facets (@teunbrand, #3062)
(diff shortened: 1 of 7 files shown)


thomasp85:
I'm slightly unsure why we need this. Yes, it "simplifies" the class definition, but on the other hand it also removes some self-documenting aspects of it. 

Do you have other reasons for this beside "we can do without"?

teunbrand:
The main reason I'd say is that it is a cause of confusion when learning how to build ggplot2 extensions because it isn't really clear what should go into `optional_aes`. It is better now that #6426 describes the field, but it remains somewhat redundant and befuddles a clear path for how to implement optional aesthetics.

Issue #6406, opened 2025-04-09
Feature request: make the input panels of `Facet$draw_panels()` predictable


Yunuuuu:
Related to https://github.com/tidyverse/ggplot2/pull/6398 and https://github.com/tidyverse/ggplot2/issues/6396.

After testing, I found it difficult to handle the original feature request using a single Facet object alone, since the `panels` input to `Facet$draw_panels()` is unpredictable. It’s challenging to break down panels into individual layer grobs.

Several issues contribute to this unpredictability:

- The `Layer$draw_geom` method doesn't always return a grob if the developer doesn’t follow the expected conventions. No error or warning is issued in such cases, and `gList(NULL)` or `gList(gList())` simply works silently. 
- `Coord$draw_panel` will sometimes wrap all layer grobs into single gTree in `CoordRadial`, sometimes it won't

Solutions:
- Ensure that `Layer$draw_geom()` always returns a single `grob` or a `gTree`.
- always wrap the all layer grobs into a single `gTree` before passing to `Coord$draw_panel`

In this way, the input `panels` will always be a list of `gTree` object for each panel.
each `gTree` object contains 3 children for `Coord$render_fg`, `Coord$render_bg` and another `gTree` of `Facet$draw_back`,  all layer grobs, and `Facet$draw_front`

teunbrand:
On what part do you get stuck if you try to implement panel spanning dendrograms? I'm unsure the premise of #6396 (using communicating grobs) is the right way to go in this case. If we don't need communicating grobs, then the current situation with how `Facet$draw_panels()` handles empty/collapsed grobs doesn't matter.

Yunuuuu:
Thanks, and apologies for the frequent PRs recently. I'm a big fan of circle plots and currently working on the circular design.

The main issue I'm encountering relates to drawing order: ggplot2 renders each facet panel sequentially, meaning the entire drawing process for one panel is completed before the next begins.

a segment drawn in one panel can be overridden by the background of the panel to its right:
```
library(ggplot2)
ggplot(data.frame(x = c(1, NA), facet = c("a", "b"))) +
  annotation_custom(
    grob = grid::roundrectGrob(
      gp = grid::gpar(fill = "red", col = "red"),
      width = 0.5, height =   0.5
    ),
    xmin = -Inf, xmax = Inf, ymin = -Inf, ymax = Inf
  ) +
  geom_segment(aes(x, y = 1L), yend = 1, xend = 2L) +
[...]

PR #6407, opened 2025-04-09
Make the input panels for `Facet$draw_panesl` predicatable

Yunuuuu:
fix https://github.com/tidyverse/ggplot2/issues/6406

This PR just make sure the `panels` input for `Facet$draw_panel` follows the shape:

It'll be a list of gTree object for each panel: each gTree object contains 3 children for Coord$render_fg, Coord$render_bg and another gTree of Facet$draw_back, all layer grobs, and Facet$draw_front

### DIFF
--- a/R/coord-.R
+++ b/R/coord-.R
@@ -211,17 +211,14 @@ Coord <- ggproto("Coord",
   },
 
   draw_panel = function(self, panel, params, theme) {
-    fg <- self$render_fg(params, theme)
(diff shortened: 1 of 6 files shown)


Yunuuuu:
I'm curious: why doesn't get_layer_grob() return the grob directly, but instead returns a list with just one element?

teunbrand:
Closing this in favour of #6421.

Issue #6408, opened 2025-04-09
Wrong parameters due to partial matching


Yunuuuu:
Currently, the `layer()` function does not initialize all parameters for the associated `Geom` or `Stat`. This can lead to issues when multiple parameters share a common prefix. A typical example is `StatSummary`, which includes parameters like `fun`, `fun.data`, `fun.max`, and `fun.min`. If a `Stat` is invoked within a `geom_*` function using the `stat` argument, problems can arise in `StatSummary$setup_params` where `params$fun` may be incorrectly matched to fun.data, fun.max, or fun.min due to R's partial matching behavior.

Please see the warning message.
``` r
options(warnPartialMatchDollar = TRUE)
df <- data.frame(
  x = c(1, 1, 2, 2, 1, 1, 2, 2),
  y = c(1, 2, 2, 3, 2, 3, 1, 2),
  fill = c(rep("A", 4), rep("B", 4))
)
library(ggplot2)
# I won't draw anything, please see the warning message
p <- ggplot(df, aes(x, y, color = fill, fill = fill)) +
  geom_smooth(stat = "summary", fun.data = mean_se)
pdf()
print(p)
#> Warning in params$fun: partial match of 'fun' to 'fun.data'
dev.off()
#> png 
#>   2
```

<sup>Created on 2025-04-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.4.2 (2024-10-31)
#>  os       Ubuntu 24.04.1 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language en
#>  collate  C.UTF-8
#>  ctype    C.UTF-8
#>  tz       Asia/Shanghai
#>  date     2025-04-09
#>  pandoc   3.1.3 @ /usr/bin/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date (UTC) lib source
#>  cli           3.6.3      2024-06-21 [1] CRAN (R 4.4.0)
#>  colorspace    2.1-1      2024-07-26 [1] CRAN (R 4.4.2)
#>  digest        0.6.37     2024-08-19 [1] CRAN (R 4.4.2)
[...]

PR #6409, opened 2025-04-09
initialize all parameters required by `Stat` and `Geom`

Yunuuuu:
Try to fix https://github.com/tidyverse/ggplot2/issues/6408


Yunuuuu:
I've reviewed the failed cases — while there are some differences, they don't result in any visible changes to the output.

Yunuuuu:
Please see https://github.com/tidyverse/ggplot2/pull/6410 instead

PR #6410, opened 2025-04-09
fix wrong parameters due to partial matching

Yunuuuu:
fix https://github.com/tidyverse/ggplot2/issues/6408


Yunuuuu:
It seems difficult to fix all of these by simply initilizing all required parameters, as some Geom and Stat classes rely on this behavior to dispatch default arguments. This PR wrap the params into a new class to avoid partial matching

teunbrand:
Thanks for the PR, but I don't think making a new class just to disable partial matching is the right solution for this problem. I'd rather just fix instances of partially matched parameters and/or have a unit test that scans all Geom/Stat class parameters for potential issues (to a similar tune of `tests/testthat/test-function-args.R`).

Yunuuuu:
Thanks! I'm totally fine with whichever method works best—as long as the issue gets resolved, that’s what matters most.

teunbrand:
I'm closing this PR in favour of #6428

PR #6421, opened 2025-04-13
Facet: gain draw_facet_panels method to draw grobs for each panel

Yunuuuu:
fix https://github.com/tidyverse/ggplot2/issues/6406

Following the advice of @teunbrand

### DIFF
--- a/R/facet-.R
+++ b/R/facet-.R
@@ -142,8 +142,8 @@ Facet <- ggproto("Facet", NULL,
   draw_front = function(data, layout, x_scales, y_scales, theme, params) {
     rep(list(zeroGrob()), vec_unique_count(layout$PANEL))
   },
-  draw_facet_panels = function(self, panels, layout, x_scales, y_scales,
-                               ranges, coord, data, theme, params, ...) {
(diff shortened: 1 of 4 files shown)


teunbrand:
Thanks for the PR! I'll get around to reviewing it at some point after discussing with Thomas a bit

Yunuuuu:
Thanks for reviewing—I’ve renamed the method accordingly.

R/facet-.R
```diff
@@ -68,6 +68,10 @@ NULL
 #'   between the layer stack and the foreground defined by the Coord object
 #'   (usually empty). The default is, as above, to return an empty grob.
 #'
+#'   - `draw_facet_panels`: Draws each panel for the facet. Should return a list
```
teunbrand, line 71:
```suggestion
#'   - `draw_panel_content`: Draws each panel for the facet. Should return a list
```

teunbrand (approved):
Looks great, thank you for the contribution! I think the failing tests are due to the new scales release, so we might have to wait for #6441 to get merged in first.

PR #6422, opened 2025-04-14
Don't link {interp} and {akima} packages

teunbrand:
While switching to R 4.5.0, I hadn't installed {interp} and {akima} yet. I noticed roxygen complains when these aren't present due to some links in the doc pages. I don't think it is worth it to place them in suggested packages just for the purpose of linking docs. So instead I propose to remove the links.


thomasp85 (approved):
LGTM

Issue #6423, opened 2025-04-14
coord_radial: Missing labels when only one unique value in `theta` axis


Yunuuuu:
When only one value in `theta` axis, no axis label will be displayed

Here is the example for the issue (note the theta axis label):
```r
library(ggplot2)
ggplot(data.frame(x = "a", y = 1:3), aes(x, y)) +
    geom_point() +
    coord_radial(start = 0, end = pi / 2, inner.radius = 0.5) +
    guides(theta = guide_axis_theta(angle = 0))
```

![Image](https://github.com/user-attachments/assets/7eb5389c-bd61-4618-b329-b4c1570b70c8)

```r
ggplot(mpg, aes(class, displ)) +
    geom_boxplot() +
    coord_radial(start = 0.25 * pi, end = 1.75 * pi) +
    guides(theta = guide_axis_theta(angle = 0))
```

![Image](https://github.com/user-attachments/assets/502653fc-c748-4ec6-9ca9-adf2a7a571bf)

PR #6424, opened 2025-04-14
guide_axis_theta: fix missing label for one unique key

Yunuuuu:
fix https://github.com/tidyverse/ggplot2/issues/6423

I don't understand why the `vdiff` test gives different results when I run the same code interactively."

```r
ggplot(data.frame(x = "a", y = 1:3), aes(x, y)) +
  geom_point() +
  coord_radial(start = 0, end = pi / 2, inner.radius = 0.5) +
  guides(theta = guide_axis_theta(angle = 0))
```
image


tests/testthat/test-guide-axis.R
```diff
@@ -382,6 +382,14 @@ test_that("guide_axis_theta sets relative angle", {
...
 })
 
+test_that("guide_axis_theta with only one axis key", {
+  p <- ggplot(data.frame(x = "a", y = 1:3), aes(x, y)) +
+    geom_point() +
+    coord_radial(start = 0, end = pi / 2, inner.radius = 0.5) +
+    guides(theta = guide_axis_theta(angle = 0))
+  expect_doppelganger("guide_axis_theta with only one axis key", p)
+})
+
```
teunbrand, line 392: I don't think this needs a visual snapshot, we can `get_guide_data(p, "theta")` to test the representation of the guide. The rest of the plot can also be simplified by leaving out the `guides()` and the arguments to `coord_radial()`.

teunbrand (commented):
Thanks for catching this! Just a small comment on simplifying the test

Yunuuuu:
Thanks for the clarification—I've simplified the test accordingly.

teunbrand (approved):
This looks good to me, thank you for the contribution!

PR #6437, opened 2025-04-22
New roxygen tag: `@aesthetics`

### LINKED ISSUE #6436: Feature request: expose aesthetic section helper
Whenever one is writing a new Geom/Stat, it is always a bit of an inconvenience to manually type out the aesthetics section. With the automatic links and recently showing the default value, writing such a section manually is a pain. One could use `#' @eval ggplot2:::rd_aesthetics(...)` to generate the section, but you're not supposed to use unexported functions.

teunbrand:
This PR aims to fix #6436.

Briefly, it introduces a new `@aesthetics` roxygen tag that takes a ggproto object (Stat/Geom/Position) and makes a little section declaring the required/optional aesthetics. 

I've tested it a little bit, but extensions should be able to use the tag for their objects too. For reasons I don't quite understand, the new section appears after the examples in the `.Rd` files, but I don't think that does any harm.


thomasp85 (approved):
LGTM

Issue #6414, opened 2025-04-09
Should `accent` argument be accessible from theme_grey(), theme_classic() etc along with `paper` and `ink`


EvaMaeRey:
The new theming system means that the look and feel of plots can be dramatically changed - even just by using base ggplot2 convenience functions `theme_grey()`, `theme_classic()` etc.  🎉

Currently, `paper` and `ink` are arguments in these functions that let users change the background and forground colors - including layer colors!  (amazing!) 

However, the user cannot change the accent color (blue in geom_smooth for example) as theme_grey and friends are currently written.  This feels a bit strange to me, because you can affect the colors (as well as size) for all grobs from the convience `theme_*` functions -- except if the grob happens to be defined by accent.  But the geom_smooth's blue (and a handful of other grobs') is untouchable unless you learn or find some more involved syntax.  

Just knowing geom_smooth's popularity and use in a lot of ggplot2 education to demonstrate building up plot layers, I wonder if making `accent` available in theme functions could be considered. 🙏

teunbrand:
Hi Gina, sorry it took a while to get back to you! Do you happed to be interested in putting together a PR for this?

PR #6438, opened 2025-04-22
add `accent` argument to theme_grey, theme_minimal, etc

EvaMaeRey:
This PR is aimed making `accent` (e.g. geom_smooth's blue) accessible from the built-in convenience theme functions like `theme_grey`, `theme_minimal` etc.  Most of the changes are pretty minimal, but for `theme_void`, it seems that `geom = element_geom(...` should be added so that that the new geom dimension of theme can be manipulated from `theme_void()`


teunbrand (changes requested):
Hi Gina, thank you so much for the PR! There are a few things that I think should be tweaked. One is in the comment, the other is whether you can run `devtools::document()` to propagate changes from the `@param` description to the `.Rd` files. In other aspects this looks great!

teunbrand:
Just a comment for linking purposes, that this PR would fix #6414.

teunbrand (approved):
Hi Gina,

Per email exchange, I think this can now be merged in.
I reverted some commits that I thought were unrelated to the goal of the PR.
Other than that, I think this is good to go, thank you so much for the contribution!

PR #6441, opened 2025-04-25
Adapt to scales 1.4.0

### LINKED ISSUE #6172: Upcoming scales 1.4.0 release
The upcoming release of scales (https://github.com/r-lib/scales/issues/474) prompts a few tasks on ggplot2's side.

- [x] Prepare for future incorporation of `breaks_exp()` into `transform_exp()` (#6171)
- [ ] Remove shims like the following https://github.com/tidyverse/ggplot2/blob/b29b83151a8415093cadc8aead5024452ed7eae6/R/utilities.R#L792-L808
- [ ] Tweak `col_mix()` calls to continue to produce the same hex codes, because the shim has a `grDevices::rgb()` backend while `scales::col_mix()` is based on {farver}, producing tiny differences.
- [x] Rework default colour/fill scales to accept new palette specifications (#6112, #6216)
- [x] Ensure that default palettes can be retrieved from the theme (#5946)

teunbrand:
This PR aims to fix #6172.

It implements the two remaining items on the list in the issue.
Most of this PR are just tiny changes in colour mixing proportions to preserve the hex code.
One visual test changed, but that is fine as that theme option wasn't available in previous versions, so there isn't anything to preserve really.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,7 +39,7 @@ Imports:
     isoband,
     lifecycle (> 1.0.1),
     rlang (>= 1.1.0),
-    scales (>= 1.3.0),
+    scales (>= 1.4.0),
     stats,
     vctrs (>= 0.6.0),
     withr (>= 2.5.0)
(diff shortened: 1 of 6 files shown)


thomasp85 (approved):
LGTM

PR #6443, opened 2025-04-30
Simplify scale training

### LINKED ISSUE #6184: implementation strategy: robust `ggplot2` functionality for visualizing `vctrs`-objects
posting here in the hopes of getting developer advice:

TL;DR: 

`vctrs`-objects implemented as list-type vectors fail `ggplot2`'s type checking for (continuous) scales, because scales are determined in `train_position` before `stats` are applied, and our `vctrs`- objects are only converted to x-y-tuples by the specific `Stat` we implement for our `vctrs`-class. 

we are uncertain what a feasible and sensible strategy is:

- go the whole `sf`route and implement entire custom coordinate systems etc (seems like it might be overkill in our case)
- just implement new primary scale types specifically for our `vctrs`-objects (how?)
- something else entirely?

we would appreciate any guidance or pointers to existing implementations solving similar issues.

more details: 

Our package [`tf`](https://github.com/tidyfun/tf) defines a `vctrs`-class `tf` for vectors of functions --basically: lists of 2-column matrices containing $(t, f_i(t))$-tuples.  
We were able to define a custom `StatTf` and some custom `geom`s based on it (e.g. [geom-spaghetti](https://github.com/tidyfun/tidyfun/blob/dev/R/geom-spaghetti.R)) to enable quick plots like this:
```r
> library(tidyfun)
> library(ggplot2)
> d <- data.frame(
+     x = tf_rgp(3), #3 random functions
+     f = factor(c(1,1,2))
+ )    
> ggplot(d) + geom_spaghetti(aes(y = x, col = f))
```
Image

This works for many different kinds of `geom`s and facetting etc etc but breaks down currently as soon as we try to use for example `scale`-transformations, see also https://github.com/tidyfun/tidyfun/issues/135:

```r
> ggplot(d) + geom_spaghetti(aes(y = x, col = f)) + scale_y_reverse()
Error in `as.character()`:
! Can't convert `x` <tfd_reg> to <character>.
> rlang::last_trace()
<error/vctrs_error_cast>
Error in `as.character()`:
! Can't convert `x` <tfd_reg> to <character>.
---
Backtrace:
     ▆
  1. ├─base (local) `<fn>`(x)
  2. ├─ggplot2:::print.ggplot(x)
  3. │ ├─ggplot2::ggplot_build(x)
  4. │ └─ggplot2:::ggplot_build.ggplot(x)
  5. │   └─layout$train_position(data, scale_x(), scale_y())
  6. │     └─ggplot2 (local) train_position(..., self = self)
  7. │       └─self$facet$train_scales(...)
  8. │         └─ggplot2 (local) train_scales(...)
  9. │           └─ggplot2:::scale_apply(layer_data, y_vars, "train", SCALE_Y, y_scales)
[...]
````

### PR DESCRIPTION
**Pull request by teunbrand**
This PR aims to fix #6184.

Briefly, the actual solution is already implemented in the scales package in https://github.com/r-lib/scales/pull/481. We just adapt ggplot2 to this solution by abdicating responsibility to scales.

### DIFF
--- a/R/scale-.R
+++ b/R/scale-.R
@@ -684,19 +684,7 @@ ScaleContinuous <- ggproto("ScaleContinuous", Scale,
     if (length(x) == 0) {
       return()
     }
-    # Intercept error here to give examples and mention scale in call
-    if (is.factor(x) || !typeof(x) %in% c("integer", "double")) {
-      # These assumptions only hold for standard ContinuousRange class, so
-      # we skip the error if another range class is used
-      if (inherits(self$range, "ContinuousRange")) {
-        cli::cli_abort(
-          c("Discrete values supplied to continuous scale.",
-            i = "Example values: {.and {.val {head(x, 5)}}}"),
-          call = self$call
-        )
-      }
-    }
-    self$range$train(x)
+    self$range$train(x, call = self$call)
   },
 
   is_empty = function(self) {
@@ -964,19 +952,12 @@ ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
     if (length(x) == 0) {
       return()
     }
-    # Intercept error here to give examples and mention scale in call
-    if (!is.discrete(x)) {
-      # These assumptions only hold for standard DiscreteRange class, so
-      # we skip the error if another range class is used
-      if (inherits(self$range, "DiscreteRange")) {
-        cli::cli_abort(
-          c("Continuous values supplied to discrete scale.",
-            i = "Example values: {.and {.val {head(x, 5)}}}"),
-          call = self$call
-        )
-      }
-    }
-    self$range$train(x, drop = self$drop, na.rm = !self$na.translate)
+    self$range$train(
+      x,
+      drop  = self$drop,
+      na.rm = !self$na.translate,
+      call  = self$call
+    )
   },
 
   transform = identity,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/testthat/_snaps/scales.md [hunk 1]
```diff
@@ -198,13 +198,13 @@
...
-    i Example values: 1, 2, 3, 4, and 5
+    Continuous value supplied to a discrete scale.
+    i Example values: 1, 2, 3, 4, and 5.
 
 ---
 
-    Discrete values supplied to continuous scale.
-    i Example values: "A", "B", "C", "D", and "E"
+    Discrete value supplied to a continuous scale.
+    i Example values: "A" and "E".
```
teunbrand, line 207: Arguably slightly worse, but if we want to improve this, this should happen in the scales package.

R/scale-.R
```diff
@@ -1196,17 +1177,16 @@ ScaleBinned <- ggproto("ScaleBinned", Scale,
   is_discrete = function() FALSE,
 
   train = function(self, x) {
+    if (length(x) == 0) {
+      return()
+    }
```
teunbrand, line 1182: For consistency's sake early exit first

thomasp85 (approved):
LGTM

PR #6448, opened 2025-05-06
Prevent instantiating NULL palettes

### LINKED ISSUE #6447: setting continuous colour not working for geom_point when followed by a continuous fill colour set
``` r
library(tidyverse)
library(palmerpenguins)
#> 
#> Attaching package: 'palmerpenguins'
#> The following objects are masked from 'package:datasets':
#> 
#>     penguins, penguins_raw

#works
ggplot(mpg, aes(displ, hwy, colour = displ)) +
  geom_point() +
  theme(palette.fill.continuous = viridis::rocket(9)) +
  theme(palette.colour.continuous = viridis::rocket(9))  
```

``` r

#doesn't work
ggplot(mpg, aes(displ, hwy, colour = displ)) +
  geom_point() +
  theme(palette.colour.continuous = viridis::rocket(9)) + 
  theme(palette.fill.continuous = viridis::rocket(9))
```

<sup>Created on 2025-05-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6447.

Briefly, using the `replace_null()` function will set `NULL` elements when replacements are not present. Such `NULL` elements would then override proper palettes causing the bug. This PR prevents those palettes from being initiated to `NULL` elements. This still allows an explicit `theme(palette.colour.continuous = NULL)` to replace an older palette.

### DIFF
--- a/R/theme.R
+++ b/R/theme.R
@@ -545,13 +545,14 @@ validate_theme_palettes <- function(elements) {
   }
 
   # Standardise spelling
-  elements <- replace_null(
-    elements,
-    palette.colour.discrete   = elements$palette.color.discrete,
-    palette.colour.continuous = elements$palette.color.continuous
-  )
-  elements$palette.color.discrete   <- NULL
-  elements$palette.color.continuous <- NULL
+  if ("palette.color.continuous" %in% names(elements)) {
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

Issue #6389, opened 2025-04-01
Review supersession


teunbrand:
It'd be helpful to review which functions have preferred alternatives and mark them as superseded. 
As an example, `coord_polar()` can be replaced with `coord_radial()`, but isn't documented like that.
I'd expect some helper functions to be functionally superseded as well.

teunbrand:
Let's create a list. It's open for additional suggestions.

* `coord_polar()` superseded by `coord_radial()`
* `coord_equal()` superseded by `coord_fixed()`
    * Perhaps these could even be incorporated in `coord_cartesian()`, they are easy to implement.

teunbrand:
Behaviour to supersede:

* `options("ggplot2.continuous.colour")`: use lifecycle signal superseded. Put superseded badges in documentation. Review documentation and books.

* supersede `plot_env` or soft-deprecate. Disentangle facet evaluation wrt plot_env.

TODO: review arguments specifically.

davidhodge931:
Could `expand_limits` not be superseded, please? I know you can do it in the scale via the limits argument with a function, but that is quite complex for a lot of users. It is a common need (e.g. adding a zero), so it's great to have a quick, easy method to do this

davidhodge931:
Also, the new approach to setting colour palettes does not do entirely what the old options approach does, as you cannot set the NA value with the new approach. 

Great to start the process of getting rid of the redundant functions!

davidhodge931:
Not everyone needs to learn ggplot2 to an advanced level, @teunbrand

clauswilke:
> Not everyone needs to learn ggplot2 to an advanced level, @teunbrand 

I think discoverability and ability to memorize patterns are both very important for regular users. I don't even know what `limits = ~ range(.x, value)` does.

In general, I have found that while it's great that there are all these nifty tricks that people can in principle use to get advanced outputs in an elegant way, in practice very few people have the mental capacity to memorize all these tricks and use them in the appropriate circumstances.

Not sure how useful this is in the present context, just pointing this out as something to keep in mind.

teunbrand:
Separate from this PR, my idea is that there will be limit helper functions in scales, e.g. https://github.com/r-lib/scales/issues/490. Which has the benefits of not being circuitous and being discoverable.

clauswilke:
I was debating whether I should bring up the scales package as an example of a repository of great helper functions that can in principle solve all sorts of issues I have and that I can never remember how to use... ;-)

[...]

PR #6450, opened 2025-05-07
New `coord_cartesian(ratio)` argument

teunbrand:
This PR implements a proposal in https://github.com/tidyverse/ggplot2/issues/6389#issuecomment-2772817978.

Briefly, we incorporate the `ratio` logic in `coord_cartesian()` so that we can supersede `coord_fixed()` and `coord_equal()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* New `coord_cartesian(ratio)` argument that supersedes `coord_fixed()` and 
+  `coord_equal()`.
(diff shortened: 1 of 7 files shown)


thomasp85 (commented):
I'm not sure we should supersede these functions, but perhaps they should just be shortcut constructors for CoordCartesian?

thomasp85:
That, and remove the superseded badges etc

thomasp85 (approved):
LGTM

PR #6451, opened 2025-05-13
Extra information for the 'position' parameter.

williamlai2:
Extra information on the possible values for 'position'.

### DIFF
--- a/R/guide-legend.R
+++ b/R/guide-legend.R
@@ -18,6 +18,7 @@
 #'   guide overrides, and is combined with, the plot's theme.
 #' @param position A character string indicating where the legend should be
 #'   placed relative to the plot panels.
+     One of "top", "right", "bottom", "left", or "inside".
 #' @param direction  A character string indicating the direction of the guide.
 #'   One of "horizontal" or "vertical."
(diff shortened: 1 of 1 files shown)


R/guide-legend.R
```diff
@@ -18,6 +18,7 @@
 #'   guide overrides, and is combined with, the plot's theme.
 #' @param position A character string indicating where the legend should be
 #'   placed relative to the plot panels.
+     One of "top", "right", "bottom", "left", or "inside".
```
teunbrand, line 21: This line needs a roxygen comment `#'` in front.

teunbrand (changes requested):
Thanks for the PR! There are a few issues with the proposed changes.

* This seems to be based on the 3.5.2 version, can you merge in the main branch?
* The documentation needs to be regenerated with `devtools::document()`
* See comment below

R/guide-legend.R
williamlai2: Sure. I'll make the changes. I might close this one though and put in a new pull request though. I ran `devtools::document()` and there are are a lot of minor changes to the documentation for other functions, though I'll just add the relevant ones.

PR #6453, opened 2025-05-13
Update svglite-based snapshots

teunbrand:
For a few visual tests, we use a custom svglite writer instead of vdiffr's writer to incorporate patterns.
The latest release of svglite appears to write text slightly differently, which this PR updates.


teunbrand:
I've piggybacked this into #5974, so this can be closed

Issue #6459, opened 2025-05-15
Chore: standardize one-liner ggproto methods to use `{`


yjunechoe:
There are some inconsistencies in how one-liner ggproto methods are written, with and without `{`. As far as I can tell, `{` for one-liners seems like an arbitrary stylistic choice in the source code

```r
library(ggplot2)

Geom$setup_data
#> <ggproto method>
#>   <Wrapper function>
#>     function (...) 
#> setup_data(...)
#> 
#>   <Inner function (f)>
#>     function (data, params) 
#> data

Stat$setup_data
#> <ggproto method>
#>   <Wrapper function>
#>     function (...) 
#> setup_data(...)
#> 
#>   <Inner function (f)>
#>     function (data, params) 
#> {
#>     data
#> }
```

It's my understanding that starting a function body with `{` is more standard. It lets methods show up on the RStudio outline pane and also some debugging functions require it (ex: `trace()` with an `at` argument).

A quick search finds a few hits for one-liners without `{` like `Geom$setup_data`:

```r
library(ggplot2)
ggprotos <- mget(ls("package:ggplot2"), asNamespace("ggplot2"), ifnotfound = list(NULL)) |>
  Filter(\(x) inherits(x, "ggproto"), x = _)
ggproto_methods_nobrackets <- lapply(ggprotos, \(x) {
  Filter(
    \(m) inherits(x[[m]], "ggproto_method") &&
      length(body(get(m, x))) == 1,
    ls(envir = x)
  )
})
sum(lengths(ggproto_methods_nobrackets))
#> [1] 19
Filter(length, ggproto_methods_nobrackets) |>
  head()
#> $Coord
#> [1] "is_free"   "is_linear"
#> 
#> $CoordCartesian
#> [1] "is_free"   "is_linear"
#> 
#> $CoordFixed
#> [1] "is_free"
#> 
#> $CoordPolar
#> [1] "aspect"  "is_free"
#> 
#> $CoordRadial
#> [1] "is_free"
#> 
#> $CoordSf
#> [1] "is_free" "labels"
```

This should be a trivial, non-intrusive chore that's easy to unit test. I'm happy to PR this if it seems within scope!

teunbrand:
Thanks for the issue June! I believe from our prior discussion that one-liners are also un`trace()`able, right? In any case, if it is harmless and `{` is preferred, I see no obstacles for a PR :)

PR #6460, opened 2025-05-15
Ensure ggproto methods start with `{`

yjunechoe:
As discussed in #6459, this PR changes instances of `function() ...` into `function() { ... }` for ggproto methods.

There was one not-so-trivial case of `ScaleDiscrete$transform`:

https://github.com/tidyverse/ggplot2/blob/86b32ce1b2e4b0f7eddb44115be342a4e32a5919/R/scale-.R#L981-L983

Because `base::identity()` is a one-liner without `{`, this fails the test:

```r
identity
#> function (x) 
#> x
#> <bytecode: 0x000001b6a4393740>
#> <environment: namespace:base>
```

In this PR, `ScaleDiscrete$transform` is now the spelled-out version:

```r
  transform = function(self, x) {
    x
  },
```


teunbrand:
Sorry for assuming a little bit of control over the PR, but some gnarly merge conflicts had emerged that were my fault. I also tweaked the test a little bit so that now it'll report which ggprotos and which method is failing.

PR #6470, opened 2025-05-21
Fix some lints

etiennebacher:


R/theme.R
```diff
@@ -990,7 +990,7 @@ combine_elements <- function(e1, e2) {
 
   # If e2 is 'richer' than e1, fill e2 with e1 parameters
   is_subclass <- !any(inherits(e2, class(e1), which = TRUE) == 0)
-  is_subclass <- is_subclass && length(setdiff(class(e2), class(e1)) > 0)
+  is_subclass <- is_subclass && length(setdiff(class(e2), class(e1))) > 0
```
etiennebacher, line 993: This was likely a bug but it was apparently never triggered.

etiennebacher:
I forgot that the `import-standalone` files are automatically generated, I reverted the change in this file.

R/theme.R
teunbrand: Good catch! It is for an interaction with extensions, which is hard to properly test

teunbrand (approved):
Looks good to me, thanks for the contribution!

PR #6469, opened 2025-05-21
revdepcheck 2025-05-20

thomasp85:

PR #6471, opened 2025-05-21
Fix reverse dependency issue with facet labelling

teunbrand:
This PR aims to fix a problem identified in #6469.

Briefly, some packages assumed that `facet$params$labeller` is a function.
In #6138 we stopped matching character input to functions, so that no longer holds true.
This PR matches character input to functions again.


thomasp85 (approved):
LGTM

PR #6118, opened 2024-09-20
Make `ScaleDiscrete$map()` compatible with vctrs-palettes

teunbrand:
This PR aims to fix #6117.

Briefly, `ScaleDiscrete$map()` has been made vctrs-aware and now works with 2D structures too.

A demonstration:

``` r
devtools::load_all("~/packages/ggplot2")
#> ℹ Loading ggplot2

rcrd <- vctrs::new_rcrd(list(foo = LETTERS[1:3], bar = 1:3), class = "foo")

# Because `alpha()` casts to character, we can show the 'rcrd' palette works
as.character.foo <- function(x) {
  x <- vec_match(x, rcrd)
  c("dodgerblue", "orchid", "limegreen")[x]
}

# For demonstration purposes
format.foo <- function(x, ...) format(vec_proxy(x), ...)

p <- ggplot(mpg, aes(displ, hwy, colour = drv)) +
  geom_point() +
  scale_colour_manual(
    values = rcrd,
    na.value = NA
  )
p
```

``` r

head(layer_data(p)$colour)
#> <foo[6]>
#>      
#> 1 B 2
#> 2 B 2
#> 3 B 2
#> 4 B 2
#> 5 B 2
#> 6 B 2
```

<sup>Created on 2024-09-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #6474, opened 2025-05-22
Fix reverse dependency issue with scale `na.value`

teunbrand:
This aims to fix a problem identified in #6469.

Briefly, using `vec_c()`  to concatenate the palette with `na.value` is  too strict in some cases. It was introduced in #6118.
Herein, we use regular `c()`, relying on `vctrs:::c.vctrs_vctr` to handle to vctrs case correctly.


thomasp85 (approved):
LGTM

PR #6475, opened 2025-05-22
Fix reverse dependency issue with `scale_backward_compatibility()`

teunbrand:
This PR aims to fix a problem identified in #6469.

Briefly, the `scale_backward_compatibility()` fallback we used sometimes got handed `scale = NULL`, which I didn't anticipate. That change was introduced in #5946. Here, we inspect `type` to inform us what to do in that case.


thomasp85 (approved):
LGTM

PR #6490, opened 2025-05-30
Add kapa.ai integration to website

hadley:


_pkgdown.yml
```diff
@@ -299,7 +313,13 @@ news:
 navbar:
   structure:
     left: [intro, reference, news, articles, extensions]
+    right: [search, github, ai]
   components:
     extensions:
-      text: Extensions
```
teunbrand, line 304: image

There appears to be sufficient space to write this out in full on my end

teunbrand (commented):
Using `pkgdown::build_site()`, I get the following:

image

Is it supposed to work with the locally built version or should it only work once deployed?

_pkgdown.yml
hadley: You need to make the browser a bit narrower

teunbrand (approved):
Thanks for the PR Hadley! I trust it'll work once deployed

yutannihilation:
Confirmed it works on the dev version of website. Looks cool!

image

hadley:
@yutannihilation thanks for checking!

PR #6494, opened 2025-06-05
Support saving objects from `grDevices::recordPlot()`

ppoyk:
Add support for saving plot objects created by the `recordPlot` function from the base-R-bundled `grDevices` package. The dedicated function `replayPlot` is used in a new `grid.draw` method to display a `recordedplot` in a device.

 This enables `ggsave` to be used more flexibly with base R code.

### DIFF
--- a/R/save.R
+++ b/R/save.R
@@ -329,6 +329,10 @@ validate_device <- function(device, filename = NULL, dpi = 300, call = caller_en
 grid.draw.ggplot <- function(x, recording = TRUE) {
   print(x)
 }
+#' @export
+grid.draw.recordedplot <- function(x) {
+  grDevices::replayPlot(x)
+}
 
 absorb_grdevice_args <- function(f) {
   function(..., type, antialias) {


teunbrand:
Hi there, thanks for the PR! 
Just for clarification because there is no associated issue: what is the problem that this PR aims to solve?
From your description it seems you want `ggsave()` to work with base R plots, is that correct?
Can you give an example how this would be used?

thomasp85:
Thanks for this contribution. However, it is not the objective of `ggsave()` to be used for anything other than saving ggplots and as such we are not going to merge it

ppoyk:
That's unfortunate.

As ggsave is very user-friendly and intuitive function to use, it is ubiquitous for saving plots.
It would be handy to be able to save all major plot types (`grid`|`base R`) with just one function.

For posterity:
You can add the capability for `ggsave()` to save plot objects from `recordPlot()`. Globally define a new S3 method for `grid.draw()`, which is used internally by `ggsave()`.
```
grid.draw.recordedplot <- function(x) { replayPlot(x) }
```
The class produced by `recordPlot()` is "`recordedplot`".

PR #6506, opened 2025-06-13
Fix revdep failure: don't extract <element_blank>@hjust

teunbrand:
This affects only 2 packages (smallsets and utile.visuals), but it is our faux pas.
Essentially, because `element_blank` does not have the `hjust` property, trying to extract it will error.
In this PR, we are more careful in trying to extract this property.

### DIFF
--- a/R/theme.R
+++ b/R/theme.R
@@ -518,9 +518,9 @@ fix_theme_deprecations <- function(elements) {
     )
     if (is.null(elements[["legend.title"]])) {
       elements$legend.title <- element_text(hjust = elements$legend.title.align)
-    } else {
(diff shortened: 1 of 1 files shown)


teunbrand:
Also snuck in another tiny issue I ran into

teunbrand:
When we use `labs(shape = NULL)`, we want to render no title for the `shape` guide. Throwing warnings to the tune of 'We're ignoring your empty labels', is irksome because the natural response to this is 'Yes, I know, that is exactly what I want you to do'.

thomasp85 (approved):
LGTM

PR #6508, opened 2025-06-15
A really minor document tweak

yutannihilation:
The document of `element` says the return value is "S3", but `element` and `margin` are already S7. `rel` is still S3, so this pull request just removes `S3` for accuracy.


teunbrand (approved):
Well spotted!

Issue #6504, opened 2025-06-11
Revdep: bridging `+` methods


teunbrand:
In some reverse dependency checks, it has become clear that some `+` methods no longer work as intended. I had already struggled to find a solution for this problem in GGally. It has come up in these packages, as far as I recall:

* GGally
* ggside
* feasts

And is also reported for nlmixr2plot (https://github.com/tidyverse/ggplot2/issues/6498#issuecomment-2963423787).

I'm not sure what the optimal solution is for this case. I don't think using S7's double dispatch is the immediate solution, in particular because it has trouble dispatching on S3 classes in some cases (see also https://github.com/RConsortium/S7/issues/544). One option could be to also implement the S3 class as an S7 class, in which case the S7 double dispatch becomes viable. 

In GGally's case, the custom LHS S3 class included `"gg"` and the RHS in e.g. `labs()` also has an S3 `"gg"` class for legacy reasons. In this case, overriding `+.gg` worked but that may not be the case for every type of class.

billdenney:
@teunbrand, the issue you raised in the linked S7 repo seems to be exactly the one that I'm experiencing with `nlmixr2plot`. If the package can use `chooseOpsMethod`, then it works for my `ggtibble` package. That may be a general solution.

teunbrand:
Yeah I'm aware of that function ever since you mentioned it in https://github.com/tidyverse/ggplot2/issues/5536#issuecomment-2526350823, but I can't seem to figure out how to make this work with the S7 methods.

teunbrand:
@billdenney My list of packages to inform arrived at ggtibble. I saw that there was a recent update to CRAN with a news entry that it was compatible with ggplot2 4.0.0. Thanks for taking such a proactive stance! I'm guessing having to inform you is now a moot point?

billdenney:
Yeah, `ggtibble` should be taken care of.

teunbrand:
Notable workaround by @schloerke in https://github.com/RConsortium/S7/issues/544#issuecomment-3028371548

PR #6516, opened 2025-06-18
docs(update_plot): Remove `@keywords internal`

schloerke:
If this method is to be leveraged, it should be public. 

`{GGally}` is looking forward to this method

### DIFF
--- a/R/plot-construction.R
+++ b/R/plot-construction.R
@@ -113,7 +113,6 @@ add_ggplot <- function(p, object, objectname) {
 #' exposed at this point, which comes with the responsibility of returning
 #' the plot intact.
 #'
-#' @keywords internal
 #' @export
 #' @examples
 #' # making a new method for the generic


teunbrand:
The `update_plot()` generic is a workaround for an S7 issue (https://github.com/RConsortium/S7/issues/543).
I'm not sure we'd want to advertise its usage if S7 allows S3 methods for S7 generics as mentioned in the issue.
If S7 solves the issue, we can all stick to `ggplot_add()` and no migration is necessary.

schloerke:
Ah! I had missed that we can no override the `+` with S7 double dispatch. 🎉🎉

https://github.com/tidyverse/ggplot2/blob/a2d0181adc10de7589bce50b13910a97e6fca756/R/plot-construction.R#L68-L76 

Related: Double dispatch with an S3 object in the first slot isn't dispatching properly: https://github.com/tidyverse/ggplot2/issues/6504

-------------------------------------------

Slightly related: 
GGally would no longer need `+.gg` method now that ggplot2 v4 leverages S7. However, GGally only needs ggplot2 >= v3.5.

To get around this, GGally can conditionally export it's `+.gg` method depending on the ggplot2 version. Similar to https://github.com/RConsortium/S7/blob/db5964265a88d114fb6de97361060979ad63b3d6/vignettes/packages.Rmd#L62-L66 , but with `#' @rawNamespace if (packageVersion("ggplot2") < "3.5.2.9001") S3method("+",gg)`

schloerke:
Closing as PR isn't needed

PR #6522, opened 2025-06-19
bug(S7): Use `local()` to avoid `rm()` call

schloerke:
If/when S7 decides to fix the bug, `rm("non-existant")` will produce a warning message.

By using `local()`, it will work for the current version and if/when S7 fixes it.


hadley:
Nice work around!

Issue #5998, opened 2024-07-12
Feature request: `geom_elbow`


fmarotta:
Hi! Currently geom_segment() supports one variant, geom_curve(), to connect two points with a curved line. I would find it useful to have in base ggplot an "elbow" geom, that connects points with a, well, elbow-shaped line, similar to geom_step() but with the same aesthetics as geom_segment() and geom_curve(). I am aware of geom_edge_elbow() in ggraph, but I believe this geom can be used more generally.

teunbrand:
Hi there, thanks for your suggestion!
Generally we recommend using the extensions system for innovation of new stats and geoms and keep ggplot2 contained to core functionality.
The elbow geom does not seem unreasonably hard to implement, nor does it appear to require any change in ggplot2 to be implemented, so this seems to fit the purpose of the extension system like a glove.
For this reason, I suggest proposing this elsewhere in an extension package (or start one of your own).

fmarotta:
Ahh, I see, thanks for your reply. It seems a pity to scatter such simple and basic functionality to yet another package, I'll try to move this issue to ggforce, which already has some nice geoms for connecting points. I'm also happy to prepare a MR.

ssp3nc3r:
If I understand your description, It seems that `ggalt::geom_xspline` may already provide something similar.

fmarotta:
I was more looking for something with `x, y, xend, yend` aesthetics, but I'll look into `geom_xspline`, thanks for the suggestion!

teunbrand:
I realised that this doesn't necessarily needs to be a new geom, it essentially can be done by exposing the `shape` argument via `geom_curve()`.

``` r
library(grid)
grob <- curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.2, y2 = 0.8, shape = 0, curvature = -1)
grid.newpage()
grid.draw(grob)
```

![](https://i.imgur.com/nIGASo8.png)

<sup>Created on 2025-06-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Breeze-Hu:
If `geom_curve` is made optimised, is there any chance of implementing something like `geom_xspline` natively in ggplot2: https://yonicd.github.io/ggalt/reference/geom_xspline.html

It may be possible to plot curves and points more elegantly (with some loss of precision, of course) in dealing with some time series. Instead it's all _**spikes**_.

teunbrand:
[...]

PR #6523, opened 2025-06-19
Expose `shape` argument in `geom_curve()`

fmarotta:
Hi! there was some discussion in issue #5998 about exposing the `shape` argument in geom_curve(). This PR attempts to do that.

I've run some tests to understand the effect of the `curveGrob()` arguments, and it seems that elbow-style lines can indeed be achieved with the following combination: `curvature = 1, ncp = 1, angle = 90, square = TRUE` (see the reprex below).

``` r
library(grid)
library(gridExtra)
grobs <- list(
  `curvature=0.1` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 0.1),
  `curvature=1` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1),
  `ncp=5` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1, ncp = 5),
  `ncp=1` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1, ncp = 1),
  `angle=60` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1, ncp = 1, angle = 60),
  `angle=90` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1, ncp = 1, angle = 90),
  `square=FALSE` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1, ncp = 1, angle = 90, square = FALSE),
  `square=TRUE` = curveGrob(x1 = 0.2, x2 = 0.8, y1 = 0.1, y2 = 0.6, shape = 0, curvature = 1, ncp = 1, angle = 90, square = TRUE)
)
grobs_with_titles <- mapply(function(grob, title) {
  grobTree(
    grob,
    textGrob(title, x=0.5, y=0.7, gp=gpar(fontsize=10))
  )
}, grobs, names(grobs), SIMPLIFY = FALSE)
grid.newpage()
grid.arrange(grobs = grobs_with_titles, ncol = 2, nrow = 4)
```

<sup>Created on 2025-06-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The PR adds the `shape` argument, but I also had to change the `square` argument. `square` is not exposed in `geom_curve()` but it is hard-coded to `FALSE`. The `grid.curve()` [documentation](https://www.rdocumentation.org/packages/grid/versions/3.6.2/topics/grid.curve) says that
> When ncp is 1 and angle is 90, this is typically TRUE, otherwise this should probably be set to FALSE.

so in this PR I follow such prescription.

Using `square = FALSE` results in "tilted" elbows, whereas when `square = TRUE` the lines are horizontal and vertical (see reprex above). The tilted effect can still be achieved by setting `angle` to something different than 90.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * `get_layer_data()` and `get_layer_grob()` now accept layer names as index 
   (@lgaborini, #6724)
+* Added new argument `geom_curve(shape)` that will be passed down to 
+  `grid::curveGrob()` (@fmarotta, #5998).
 
 # ggplot2 4.0.1
 


tests/testthat/test-geom-curve.R
```diff
@@ -9,3 +9,27 @@ test_that("geom_curve flipping works", {
...
+  p_1 <- ggplot(df, aes(x, y, xend = xend, yend = yend)) +
+    geom_curve(arrow = arrow(), shape = 1)
+
+  p_m1 <- ggplot(df, aes(x, y, xend = xend, yend = yend)) +
+    geom_curve(arrow = arrow(), shape = -1, angle = 40)
+
+  expect_doppelganger("shape=0 geom_curve", p_0)
+  expect_doppelganger("shape=0 geom_curve", p_0_not_square)
+  expect_doppelganger("shape=1 geom_curve", p_1)
+  expect_doppelganger("shape=-1 geom_curve", p_m1)
```
teunbrand, line 33: Can we make this 1 plot with 4 `geom_curve()` layers, each with something like `aes(colour = "not square")` that is indicative of the setting, so that it sort of becomes self-descriptive?

teunbrand (commented):
This looks good to me in principle. Would you mind adding an example to the docs using the new parameter (or integrating it in some example)?

We're currently in the process of preparing a release, so we've temporily halted adding new features. If all is well we can consider merging this after the release (and perhaps a hotfix) is out.

fmarotta:
Thanks for the review! The example is included.

teunbrand (approved):
This is great, thank you for this! I apologise for taking a long time to circle back to this.

PR #6527, opened 2025-06-22
Additional notes on legend.ticks.length

Breeze-Hu:
Hi, Teun and Thomas!

When adjusting the length of ticks in the legend of ggplot2, many users are not aware of adjusting the length of a single side, e.g. to generate a legend with only one side of ticks. I would like to add to the documentation accordingly.
If you all think it's not appropriate, clear feel free to close this PR.

(This is my first PR, if there are problems please help to modify or close it.🤣)

### DIFF
--- a/R/theme.R
+++ b/R/theme.R
@@ -92,10 +92,7 @@
 #'   or `"top"`, or a two-element numeric vector.
 #' @param legend.frame frame drawn around the bar ([element_rect()]).
 #' @param legend.ticks tick marks shown along bars or axes ([element_line()])
-#' @param legend.ticks.length length of tick marks in legend
-#'   ([`unit()`][grid::unit]); By default, tick marks are positioned symmetrically.  
-#'   One-sided tick lengths can be controlled using absolute (`unit()`) or relative (`rel()`) values.  
-#'   For example, `unit(c(-0.15, 0), "cm")` or `rel(c(0.175, 0))` produces ticks on one side only. inherits from `legend.key.size`.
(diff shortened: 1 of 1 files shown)


teunbrand:
Hi Breeze-Hu, thank you for the pull request!
The PR is great and has no technical problems. However, I worry that adding this description will over-specify the theme element. For example, `guide_bins()` also uses the `legend.ticks.length` theme setting, but it only has a one-sided axis instead of a two-sided axes. As a consequence, setting a length 2 unit/rel vector applies alternatingly to ticks, instead to different sides. I'd be fine to add in the description of `guide_colourbar()` that this behaviour exists, but if we do so, we should also ensure we test the behaviour so we don't accidentally break it in the future.

Breeze-Hu:
Oh, yeah! Sorry, I did overlook some cases involving single-sided ticks. I'm afraid I don't have the capacity for deeper optimizations and adjustments at the moment. I hope this PR can be kept as a reminder, and you or someone else can address it when a better solution becomes available.

davidhodge931:
+1 for supporting ticks on a colourbar on a single side, if possible :)

teunbrand:
@davidhodge931 do you mean keeping the current behaviour?

davidhodge931:
Do ggplot2 officially support ticks on one side of colour bars at the mo? Or is this off-label use? It's be cool if it was officially supported

teunbrand:
> It's be cool if it was officially supported

I've made separate issue #6640 to discuss this. For now, I'll close this PR as I imagine the proposed fix has no code overlap with this PR and Breeze-Hu mentioned there is no capacity to take this on at the moment.

Issue #6537, opened 2025-07-02
warning message for @aesthetics roxygen2 tag


Yunuuuu:
I found a problem with `#' @aesthetics *****`

![Image](https://github.com/user-attachments/assets/ba50d215-5ada-45ce-a5b5-bd07dfa1efa7)

PR #6538, opened 2025-07-02
fix: correct Rd link syntax for aesthetics roxygen2 tag

Yunuuuu:
fix https://github.com/tidyverse/ggplot2/issues/6537

### DIFF
--- a/R/utilities-help.R
+++ b/R/utilities-help.R
@@ -78,7 +78,7 @@ rd_match_docpage <- function(aes) {
   )[index + 1]
   no_match <- index == 0
   docpage[!no_match] <- paste0(
-    "\\link[=ggplot2::", docpage[!no_match],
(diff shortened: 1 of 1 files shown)


Yunuuuu:
Everything works well for me in this PR: 
image

teunbrand:
Does this look fine both when ggplot2 uses the `@aesthetics` tag and extensions use the `@aesthetics` tag?

Yunuuuu:
Yeah, both work fine regarding the docs! I did notice a few errors in the ggplot2 R CMD check (not related to the docs):
image

Here is the error:
image

teunbrand:
Thanks for confirming! Are the failures you see on your end just due to a testthat version <3.2.2? Because I don't see any of those failures in the GHA checks

Yunuuuu:
Everyhing works well after updating testthat.

PR #6540, opened 2025-07-03
Export class definitions

### LINKED ISSUE #6517: feat: Export all `class_*` objects
When working with S7 dispatch for `+`, I would like for the `class_FOO` objects to be exported. These methods do not need a public docs page (e.g. `@keywords internal`) but could display an example of how to make your own `+` method. 

`{GGally}` would benefit having access to `class_theme`, `class_labels`, `class_ggproto` , etc.

Prior art: S7 package exports all `class_FOO` objects for us to build upon. It'd be nice to have access to all ggplot2 `class_FOO` objects to build upon them. 

Thank you!

teunbrand:
This PR aims to fix #6517.

In addition to exporting classes, I've made a docpage where you can see all class definitions in one spot.
All S7 constructors also combine the `...` with a warning if `...` is actually used.


thomasp85 (approved):
LGTM

PR #6542, opened 2025-07-03
Show `make_constructor()` a few times in extension vignette

### LINKED ISSUE #6524: Showcase `make_constructor()` in extension vignette
Spend all this time making it, it'd be a shame if nobody knew about it!

teunbrand:
This PR aims to fix #6524.

It replaces some constructor code with the `make_constructor()` shortcut. This also leaves 1 instance of when not to use a constructor when you want to bake-in the 'stat' for example.

### DIFF
--- a/vignettes/extending-ggplot2.Rmd
+++ b/vignettes/extending-ggplot2.Rmd
@@ -493,17 +493,7 @@ GeomPolygonHollow <- ggproto("GeomPolygonHollow", GeomPolygon,
     alpha = NA)
   )
 
-# We're baking in StatChull instead of exposing it via an argument, so we
-# need a custom constructor
-geom_chull <- function(mapping = NULL, data = NULL, 
-                       position = "identity", na.rm = FALSE, show.legend = NA, 
-                       inherit.aes = TRUE, ...) {
-  layer(
-    stat = StatChull, geom = GeomPolygonHollow, data = data, mapping = mapping,
-    position = position, show.legend = show.legend, inherit.aes = inherit.aes,
-    params = list(na.rm = na.rm, ...)
-  )
-}
+geom_chull <- make_constructor(GeomPolygonHollow, stat = "chull")
 
 ggplot(mpg, aes(displ, hwy)) + 
   geom_point() + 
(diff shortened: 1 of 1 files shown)


teunbrand:
Suggestion from @EvaMaeRey: should we change `make_constructor()` to `write_constructor()`? I kind of like it because it writes a function for you so that you don't have to.

EvaMaeRey:
Hi Teun, Exciting to see the new function in action in the vignette!

I think the geom_chull case would *not* be one where I'd opt to  write out the constructor instead of using make_constructor/write_constructor.  I think the convention is arguably to expose the stat (even though it would be very strange to see a user switch this out).

Actually, in the ggplot2 book's example of geom_chull the `stat` argument is exposed, https://ggplot2-book.org/extensions.html#sec-modifying-geom-defaults, and I think this represents convention more than a locked-in stat (and will enable the super terse make_constructor move! ); maybe more representative of conventions?

And in the book case study, where geom_spring is first defined as a near alias to stat_spring, `stat` argument is also exposed https://ggplot2-book.org/ext-springs.html#constructors    21.2.4 Constructors   And writing these two functions is a place where make_constructor/write_constructor really shines (otherwise scaffolding code 2x). 

stat_spring <- make_constructor(StatSpring, geom = "path") # write a constructor instead of geom_path(stat = StatSpring)
geom_spring <- write_constructor(GeomPath, stat = "spring")  # and just because 'ggplot2 users are accustomed to adding geoms, not stats, when building up a plot'

From base ggplot2, geom_count is another example where the stat is exposed, even though the function name kind of refers to the stat (StatSum) 

``` r
ggplot2::geom_count
```

function (mapping = NULL, data = NULL, stat = "sum", position = "identity", 
    ..., na.rm = FALSE, show.legend = NA, inherit.aes = TRUE) 
{
    layer(mapping = mapping, data = data, geom = "point", stat = stat, 
        position = position, show.legend = show.legend, inherit.aes = inherit.aes, 
        params = list2(na.rm = na.rm, ...))
}


While, I think there probably should be a couple of examples of write-it-out construction, I don't think geom_chull is the right spot to do it (so the vignette should probably follow the convention that geom_*() functions in general expose stat argument to users - as weird it would be for users to actually switch the stat out).

teunbrand:
Thanks Gina, I'm now also using `make_constructor()` for `geom_chull()`.

EvaMaeRey:
Okay!!  🙏 In a hypothetical further reworking, I think it would be really nice to newcomers for geom_chull be written with ggplot2::GeomPolygon X StatChull and included in the section 'creating a new `Stat`'.  Right now, `geom_chull`'s currently exists to demonstrate 1) the creation of GeomPolygonHollow and 2) combining new Stats with new Geoms, but maybe creating geom_polygon_hollow <- make_constructor(GeomPolygonHallow, stat = "identity") would feel somewhat more straightforward to digest (not sure of easy data to show it's use), and then a separately demonstration of GeomNew X StatNew combos thereafter (say 'actually, we might like our geom_chull (see 'creating a new `Stat`) to be defined with our newly defined GeomPolygonHollow.

PR #6546, opened 2025-07-07
Streamline S7 parts

teunbrand:
This PR amends #6364.

There are two workarounds we've used to make S7 work that, after some consideration, may not be necessary.

1. To work around `S7::method()` returning a function in the namespace that then clashed with S3 functions, we used `function_name.ggplot2::ggplot` and its ilk. Since then we learned of using `local()` to avoid this problem, and it is arguably a better solution than hacking away at S3 methods. Except for replacement functions, this seems to work great.
2. We introduced 4 S7 'adaptor generics' to facilitate transition between S3 and S7 as we cannot transfer the S3 methods written by extension packages to S7 generics. However, I've since learned that it is S7's intent to have S3 fallback for S7 generics in the future. At that point, the S7 adaptor generics would become vestigial. To prevent having to instigate a future deprecation cycle, I think it is best to keep these S3 generics with S7 methods for now. It concerns the following pairs:
    * `build_ggplot()` for `ggplot_build()`
    * ~`update_ggplot()` for `ggplot_add()`~ This uses double dispatch that cannot be S3.
    * `draw_element()` for `element_grob()`
    * `gtable_ggplot()` for `ggplot_gtable()`


teunbrand:
Had to do a small consession because I already used `build_ggplot()` in https://github.com/rstudio/thematic/pull/156 which is already on CRAN.

R/aes.R
```diff
@@ -131,40 +131,40 @@ new_aesthetic <- function(x, env = globalenv()) {
...
   }
+})
 
-  invisible(x)
-}
+local({
+  S7::method(`[`, class_mapping) <- function(x, i, ...) {
+    class_mapping(`[`(S7::S7_data(x), i, ...))
+  }
+})
```
thomasp85, line 155: Do we expect this to be necessary in perpetuity? If not, maybe add a note about fixing it once the bug in S7 has been fixed

R/plot-build.R
```diff
@@ -25,19 +25,21 @@
...
-build_ggplot <- S7::new_generic("build_ggplot", "plot", fun = function(plot, ...) {
+ggplot_build <- function(plot, ...) {
   env <- try_prop(plot, "plot_env")
   if (!is.null(env)) {
     attach_plot_env(env)
   }
-  S7::S7_dispatch()
-})
+  UseMethod("ggplot_build")
+}
```
thomasp85, line 34: Maybe add a note about fixing this once possible

R/plot-build.R
```diff
@@ -208,12 +198,12 @@ layer_grob <- get_layer_grob
 #' @keywords internal
 #' @param data plot data generated by [ggplot_build()]
 #' @export
-gtable_ggplot <- S7::new_generic("gtable_ggplot", "data", function(data) {
+ggplot_gtable <- function(data) {
   attach_plot_env(data@plot@plot_env)
-  S7::S7_dispatch()
-})
+  UseMethod("ggplot_gtable")
+}
```
thomasp85, line 204: Similar. Add note about updating in the future

thomasp85 (approved):
LGTM but a few comments

Issue #6544, opened 2025-07-04
S7 class compatibility


Yunuuuu:
It seems we cannot use S7 class to extend the new class system:

``` r
class_new_theme <- S7::new_class("new_theme", ggplot2::class_theme)
S7::convert(ggplot2::theme(), class_new_theme)
#> Error in (structure(function (elements, complete, validate) : argument "elements" is missing, with no default
```

<sup>Created on 2025-07-04 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>
~                                                                           
~                                                                           
~

Yunuuuu:
This works, we must pass the property manually:

``` r
class_new_theme <- S7::new_class("new_theme", ggplot2::class_theme)
old_theme <- ggplot2::theme(plot.title = ggplot2::element_text())
class_new_theme(old_theme, old_theme@complete, old_theme@validate)
#> <new_theme> List of 1
#>  $ plot.title: <ggplot2::element_text>
#>   ..@ family       : NULL
#>   ..@ face         : NULL
#>   ..@ italic       : chr NA
#>   ..@ fontweight   : num NA
#>   ..@ fontwidth    : num NA
#>   ..@ colour       : NULL
#>   ..@ size         : NULL
#>   ..@ hjust        : NULL
#>   ..@ vjust        : NULL
#>   ..@ angle        : NULL
#>   ..@ lineheight   : NULL
#>   ..@ margin       : NULL
#>   ..@ debug        : NULL
#>   ..@ inherit.blank: logi FALSE
#>  @ complete: logi FALSE
#>  @ validate: logi TRUE
```

<sup>Created on 2025-07-04 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>
~                                                                           
~                                                                           
~

teunbrand:
It is unclear to me what you're trying to achieve here. Would you mind explaining a bit more?

Yunuuuu:
Apologies, I missed the message. The main issue we're facing is that it's not possible to directly extend or attach an extensible class to an existing ggplot2 object. If a user provides a `ggplot` object, our goal is to convert it into our own custom class for further extension. However, the standard S7 conversion `S7::convert()` method doesn't behave as expected in this case, which limits our ability to build on top of ggplot2 objects in a flexible way.

Would you have any suggestions or workarounds for this limitation?

teunbrand:
[...]

PR #6557, opened 2025-07-18
fix: ensure S7 constructor matches class name to avoid identity check…

Yunuuuu:
Fix partial: https://github.com/tidyverse/ggplot2/issues/6544


Yunuuuu:
image

teunbrand:
I don't like changing the class names. They're used in error messages and it'll expose implementation details that are confusing to users.

teunbrand:
Thank you for the PR and the willingness to contribute. However, I'm going to close this, because it'd be confusing to change the class names.

PR #6597, opened 2025-09-08
Reformat code with air

gaborcsardi:
Had to fix one test case, ggsave:144.


thomasp85:
Thanks - we have to be very mindful with when we merge in something like that as it will create conflict in every open PR

PR #6609, opened 2025-09-12
Fix draw_key_rect() not using data$fill for glyph fill

### LINKED ISSUE #6608: Fill colours are ignored in draw_key_rect
A regression introduced in ggplot2 v4.0.0 resulting from #6327 results in `grey20` fills when a valid fill is mapped.

**ggplot2 v.3.5.2**

``` r
library(ggplot2)

# Column chart
mtcars |> 
  ggplot(aes(x = factor(cyl), fill = factor(cyl))) +
  geom_col(aes(y = mpg), key_glyph = draw_key_rect)
```

<sup>Created on 2025-09-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

**ggplot2 v4.0.0**

``` r
library(ggplot2)

# Column chart
mtcars |> 
  ggplot(aes(x = factor(cyl), fill = factor(cyl))) +
  geom_col(aes(y = mpg), key_glyph = draw_key_rect)
```

<sup>Created on 2025-09-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

mitchelloharawild:
This PR resolves #6608:

``` r
library(ggplot2)

# Column chart
mtcars |> 
  ggplot(aes(x = factor(cyl), fill = factor(cyl))) +
  geom_col(aes(y = mpg), key_glyph = draw_key_rect)
```

<sup>Created on 2025-09-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

As an aside, the pattern of checking if a value is NULL or NA seems to be a common one in defaulting aesthetics that could be wrapped up into a helpful infix function, e.g. `x %NA0% y` to use `y` if `x` is `NA` or empty. Looking at the code here, it seems that if both fill and colour are NA, then `"grey20"` may not be used?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,8 @@
 
 ### Bug fixes
 
-* Fixed regression where `draw_key_rect()` stopped using `fill` colours (#6609).
+* Fixed regression where `draw_key_rect()` stopped using `fill` colours 
+  (@mitchelloharawild, #6609).
 
 # ggplot2 4.0.0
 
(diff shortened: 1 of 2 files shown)


mitchelloharawild:
Thanks, I've updated the PR.

teunbrand:

PR #6612, opened 2025-09-14
suggested small tweaks to README

smach:
ggplot2 is now a lot older than 10 years, so suggest updating that. In addition, maybe README should reference the major new 4.0 version after saying 'ggplot2 itself changes relatively little', since there's been a significant change?


thomasp85:
From a user perspective there is still relatively little change and 4.0.0 is in that regard not more different than 3.5 or any of the other minor releases which also had a few new features. The reason for the major release is the adoption of the S7 system which is rather invisible to the user

smach:
Good point. Done (I think)

teunbrand (approved):
This looks good to me, thank you for the contribution!

PR #6629, opened 2025-09-19
set_theme() resets theme if called with no arguments.

### LINKED ISSUE #6598: Resetting the default theme
As far as I know, there is no way to restore the global theme to the original defaults.
I'm proposing we can use `set_theme()` without any arguments to restore the original defaults.
The original default theme is stored under `ggplot_global$theme_default` internal variable.

VisruthSK:
`set_theme()` called without any arguments resets the theme to the default theme.

Fixes #6598.

### DIFF
--- a/R/theme-current.R
+++ b/R/theme-current.R
@@ -91,12 +91,10 @@ get_theme <- function() {
 theme_get <- get_theme
 
 #' @rdname get_theme
-#' @param new new theme (a list of theme elements). Sets theme to the default (theme_grey) if `new` isn't supplied.
+#' @param new new theme (a list of theme elements). Sets theme to the default ([theme_grey()]) if `new` isn't supplied.
 #' @export
 set_theme <- function(new) {
-  if (missing(new)) {
-    new <- ggplot_global$theme_default
-  }
+  new <- new %||% ggplot_global$theme_default
   check_object(new, is_theme, "a {.cls theme} object")
   old <- ggplot_global$theme_current
   ggplot_global$theme_current <- new
--- a/R/theme-current.R
+++ b/R/theme-current.R
@@ -93,7 +93,7 @@ theme_get <- get_theme
 #' @rdname get_theme
 #' @param new new theme (a list of theme elements). Sets theme to the default ([theme_grey()]) if `new` isn't supplied.
 #' @export
-set_theme <- function(new) {
+set_theme <- function(new = null) {
   new <- new %||% ggplot_global$theme_default
   check_object(new, is_theme, "a {.cls theme} object")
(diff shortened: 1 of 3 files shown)


R/theme-current.R
```diff
@@ -91,9 +91,12 @@ get_theme <- function() {
 theme_get <- get_theme
 
 #' @rdname get_theme
-#' @param new new theme (a list of theme elements)
+#' @param new new theme (a list of theme elements). Sets theme to the default (theme_grey) if `new` isn't supplied.
```
lionel-, line 94: This should link to the default theme:

```suggestion
#' @param new new theme (a list of theme elements). Sets theme to the default ([theme_grey()]) if `new` isn't supplied.
```

R/theme-current.R
```diff
@@ -91,9 +91,12 @@ get_theme <- function() {
...
 #' @export
 set_theme <- function(new) {
+  if (missing(new)) {
```
lionel-, line 97: We generally prefer to take `NULL` as default argument, then you can use the `%||%` operator to assign the default:

```r
new <- new %||% ggplot_global$theme_default
```

tests/testthat/test-theme.R
```diff
@@ -56,6 +56,12 @@ test_that("modifying theme element properties with + operator works", {
   )
 })
 
+test_that("set_theme() resets theme to default when called with no arguments", {
```
lionel-, line 59: 👍

R/theme-current.R
```diff
@@ -91,12 +91,10 @@ get_theme <- function() {
 theme_get <- get_theme
 
 #' @rdname get_theme
-#' @param new new theme (a list of theme elements). Sets theme to the default (theme_grey) if `new` isn't supplied.
+#' @param new new theme (a list of theme elements). Sets theme to the default ([theme_grey()]) if `new` isn't supplied.
 #' @export
 set_theme <- function(new) {
```
lionel-, line 96: This needs to be `new = NULL`

R/theme-current.R
```diff
@@ -91,9 +91,10 @@ get_theme <- function() {
 theme_get <- get_theme
 
 #' @rdname get_theme
-#' @param new new theme (a list of theme elements)
+#' @param new new theme (a list of theme elements). Sets theme to the default ([theme_grey()]) if `new` isn't supplied.
 #' @export
-set_theme <- function(new) {
+set_theme <- function(new = null) {
```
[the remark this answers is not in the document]
VisruthSK: Thanks for correcting it for me! I was so sure I wrote it in all caps :(

lionel-:
Thanks!

PR #6630, opened 2025-09-19
Fix for resolution - if all differences less than tolerance, return 1

### LINKED ISSUE #6562: geom_jitter fails when coordinate values have very small spread
When geom_jitter is given some data where the y coordinates are all the same, a default jitter height is chosen. But if the y coordinates differ only by a small magnitude, the selection of the jitter height seems to fail and nothing is plotted. 

Minimal working example:

``` r
library(ggplot2)

set.seed(123)
ggplot(data.frame(x = 1:50, y = 0)) +
    geom_jitter(aes(x, y))
```

``` r
set.seed(123)
ggplot(data.frame(x = 1:50, y = 1e-8 * rnorm(50))) +
    geom_jitter(aes(x, y))
#> Warning in min(d[d > tolerance]): no non-missing arguments to min; returning
#> Inf
#> Warning in stats::runif(length(x), -amount, amount): NAs produced
#> Warning: Removed 50 rows containing missing values or values outside the scale range
#> (`geom_point()`).
```

<sup>Created on 2025-07-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I would have expected both plots to be more or less the same.

laderast:
Fixes #6562

### DIFF
--- a/R/utilities-resolution.R
+++ b/R/utilities-resolution.R
@@ -39,6 +39,6 @@ resolution <- function(x, zero = TRUE, discrete = FALSE) {
     return(1)
   }
 
-  min(d[d > tolerance], 1)
+  min(d[d > tolerance])
 
 }


R/utilities-resolution.R
```diff
@@ -33,5 +34,11 @@ resolution <- function(x, zero = TRUE, discrete = FALSE) {
   }
   d <- diff(sort(x))
   tolerance <- sqrt(.Machine$double.eps)
+
+  if(all(d < tolerance)){
+    return(1)
+  }
+
   min(d[d > tolerance])
+
```
thomasp85, line 43: Thanks for this. I think this can be solved with fewer computations by doing something `min(d[d > tolerance], 1)`

thomasp85: oh, I completely messed up in that suggestion. Sorry about that. 

You can revert to your old approach

laderast: Ah, will fix

thomasp85:
Thanks!

PR #6631, opened 2025-09-19
Add warning when boxplot orientation is ambiguous fixes #4250 

### LINKED ISSUE #4250: Issue warning when geom orientation is ambiguous
I'm using the new-ish bidirectional geom support to make boxplots stretching along the x-axis. If I place a continuous aesthetic on the `y` axis, I believe the warning should instead say `Continuous y aesthetic`. Since I'm not sure how ggplot2 could know my intentions (i.e. whether I want x or y to be considered as the grouping variable), maybe it would be less confusing if the warning said something along the lines of "One of x or y aesthetic should be discrete or you can use `aes(group = ...)` to specify the grouping variable." Not sure if that's too wordy, but something along those lines would be an improvement on the current warning.

``` r
library(tidyverse)
ggplot(mpg, aes(y = displ, x = cty)) +
  geom_boxplot()
#> Warning: Continuous x aesthetic -- did you forget aes(group=...)?
```

<sup>Created on 2020-10-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

sierrajohnson:
fixes #4250

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -562,7 +562,7 @@ has_flipped_aes <- function(data, params = list(), main_is_orthogonal = NA,
     }
   }
 
-  default
+  as.logical(default)[1]
 }
 #' @rdname bidirection
 #' @export


thomasp85:
Thanks for this. Just a small tweak, but the core of the fix is good

thomasp85 (approved):

thomasp85:
Thanks!!

PR #6632, opened 2025-09-19
Fixes Delete code for old versions of R

### LINKED ISSUE #6589: Delete code for old versions of R
There are a few places in the code base that only run if `getRversion() < "3.3"` (or similar). Since ggplot2 now requires R 4.1, many of these can be deleted.

cpenaloza:
Deleted all code snippets that call `getRversion() <= "4.1"`. Fixes #6589


thomasp85 (approved):
Thanks!!

PR #6633, opened 2025-09-19
add stat to `hline` `vline` and `abline`

### LINKED ISSUE #6559: expose `stat` argument in geom_vline and geom_hline?
I think it would be nice to expose the stat argument in geom_vline and geom_hline.  These are rare exceptions where you aren't be able to access the Geom from the analogous user-facing function geom_*(), like geom_segment(stat = StatManual), geom_point(stat = StatNewCustom)...

The following modification might be in the right direction.  They do probably also raise some more questions of exposing position, and other arguments additionally.

```r
geom_vline <- function(mapping = NULL, data = NULL,
                       ...,
                       xintercept,
                       na.rm = FALSE,
                       show.legend = NA,
                       stat = "identity") {                   #<<<< new exposed argument

  # Act like an annotation
  if (!missing(xintercept)) {
    # Warn if supplied mapping and/or data is going to be overwritten
    if (!is.null(mapping)) {
      cli::cli_warn("{.fn geom_vline}: Ignoring {.arg mapping} because {.arg xintercept} was provided.")
    }
    if (!is.null(data)) {
      cli::cli_warn("{.fn geom_vline}: Ignoring {.arg data} because {.arg xintercept} was provided.")

  # add a warning about stat being ignored if xintercept is provided ?
  
    }

    data <- data_frame0(xintercept = xintercept)
    mapping <- aes(xintercept = xintercept)
    show.legend <- FALSE
  }

  layer(
    data = data,
    mapping = mapping,
    stat = stat,                   #<<<< not fixed
    geom = GeomVline,
    position = PositionIdentity,
    show.legend = show.legend,
    inherit.aes = missing(xintercept),       #<<<< This change or similar
    params = rlang::list2(
      na.rm = na.rm,
      ...
    )
  )
```

sierrajohnson:
Fixes #6559

### DIFF
--- a/man/geom_abline.Rd
+++ b/man/geom_abline.Rd
@@ -9,31 +9,37 @@
 geom_abline(
   mapping = NULL,
   data = NULL,
+  stat = "identity",
   ...,
   slope,
   intercept,
   na.rm = FALSE,
-  show.legend = NA
+  show.legend = NA,
+  inherit.aes = FALSE
 )
 
 geom_hline(
   mapping = NULL,
   data = NULL,
+  stat = "identity",
   position = "identity",
   ...,
   yintercept,
   na.rm = FALSE,
-  show.legend = NA
+  show.legend = NA,
+  inherit.aes = FALSE
 )
 
 geom_vline(
   mapping = NULL,
   data = NULL,
+  stat = "identity",
   position = "identity",
   ...,
   xintercept,
   na.rm = FALSE,
-  show.legend = NA
+  show.legend = NA,
+  inherit.aes = FALSE
 )
 }
 \arguments{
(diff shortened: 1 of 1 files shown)


R/geom-abline.R
```diff
@@ -73,7 +73,9 @@ geom_abline <- function(mapping = NULL, data = NULL,
                         slope,
                         intercept,
                         na.rm = FALSE,
-                        show.legend = NA) {
+                        show.legend = NA,
+                        stat = "identity",
+                        inherit.aes = FALSE) {
```
thomasp85, line 78: can you match the argument order with the other geoms. Especially `stat` tend to come earlier

thomasp85 (commented):
Thanks! There are a few comments. Lastly you need to run `document()` to update the documentation (this is the reason for the failing tests)

thomasp85:
You still need to run `devtools::document()` to update the documentation :-)

PR #6651, opened 2025-09-26
Graphical parameters for `geom_smooth()` band

### LINKED ISSUE #6551: Feature Request：Add some parameters to `geom_smooth` to adjust the properties of the lines at the edges of the confidence intervals.
Hi, Teun!
In the process of using ggplot for plotting fitted lines, I came across a scenario of **_trying to show confidence intervals using line segments rather than shaded space_**. I asked a question on stackoverflow, But I don't think it's a very elegant or user-friendly solution.

https://stackoverflow.com/questions/79668693/plotting-the-95-error-margin-segments-of-the-fitted-line-and-controlling-the-li

I used the following code approach to achieve the results I wanted to show:
``` r
library(ggplot2)

ggplot(data = mpg, aes(x = displ,y = hwy))+
  geom_point()+
  geom_smooth(method = "lm", se = FALSE, color = "black") +
  stat_smooth(method = "lm", 
              geom = 'ribbon',
              fill = NA,
              linetype = 2,
              se = TRUE, color = "blue")
#> `geom_smooth()` using formula = 'y ~ x'
#> `geom_smooth()` using formula = 'y ~ x'
```

<sup>Created on 2025-07-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I know that the code actually does the same calculation twice for the same process. 

Is there any chance that similar to the previous additions to the whiskers, staples and median line related properties of `geom_boxplot`, `geom_smooth` could be adjusted appropriately as well? It's not even about aesthetics, it's about simply adjusting **basic properties like colour/color, linewidth and linetype**.

I hope I'm not disturbing you and interfering with the release of a new version of ggplot.

Best wishes,
Hu

teunbrand:
This PR aims to fix #6551.

Briefly, we allow setting the colour/linetype/linewidth of the confidence bands in `geom_smooth()`.
A demonstration:

``` r
#> ℹ Loading ggplot2

ggplot(mpg, aes(displ, hwy, fill = drv)) +
  geom_smooth(
    band.colour = "black",
    band.linetype = 3,
    band.linewidth = 0.5
  )
#> `geom_smooth()` using method = 'loess' and formula = 'y ~ x'
```

<sup>Created on 2025-09-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/man/geom_smooth.Rd
+++ b/man/geom_smooth.Rd
@@ -16,7 +16,7 @@ geom_smooth(
   se = TRUE,
   band.colour = NULL,
   band.color = NULL,
-  band.linetype = NULL,
+  band.linetype = "blank",
   band.linewidth = NULL,
   na.rm = FALSE,
   orientation = NA,


teunbrand:
That's right. Do you think missing parameters should be drawn from the central line?

thomasp85:
That is how we do it for the others, right? `NULL` means don't draw, `NA` means inherit..?

teunbrand:
I think the meaning of `NULL` and `NA` are swapped. The main reason I initially didn't take properties from the central line was because it should be hidden by default. Of course we might achieve that by setting transparent colours, 0 linewidth or blank linetypes.

thomasp85:
oh, yeah you are right... Then just default the arguments to `NA` to ensure default is not to draw, and let `NULL` mean inherit

teunbrand:
I'm now hiding it with `band.linetype = "blank"`. Activating the band lines requires setting a non-blank linetype.

``` r
#> ℹ Loading ggplot2

  geom_smooth(band.linetype = 1, linetype = 2)
#> `geom_smooth()` using method = 'loess' and formula = 'y ~ x'
```



`band.linetype = NULL` inherits from center line.

``` r
  geom_smooth(band.linetype = NULL, linetype = 2)
#> `geom_smooth()` using method = 'loess' and formula = 'y ~ x'
```



<sup>Created on 2025-12-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #6652, opened 2025-09-26
Rename `justification` settings to `just` settings

### LINKED ISSUE #6607: Could the `legend.box.just` be named `legend.box.justification` for consistency?
Sorry to be pedantic. Feel free to close, if no bandwidth. Maybe a tidydev day thing? Well done on the v4 release!
```
theme(
  legend.justification = "top",
  legend.key.justification = "right", 
  legend.box.just = "right",
)
```

teunbrand:
This PR aims to fix #6607.

I'll reiterate that this is just for consistency purposes.
I know from previous revdepchecks that some packages save their themes to objects instead of regenerating them by calling `theme()`.
Other packages have `expect_silent()` in their test, which is incompatible with lifecycle changes.
Such packages are at risk of being negatively affected.

### DIFF
--- a/man/guide_bins.Rd
+++ b/man/guide_bins.Rd
@@ -29,7 +29,7 @@ guide partially overrides, and is combined with, the plot's theme.
 Arguments that apply to a single legend are respected, most of which have
 the \code{legend}-prefix. Arguments that apply to combined legends
 (the legend box) are ignored, including \code{legend.position},
-\verb{legend.justification.*}, \code{legend.location} and \verb{legend.box.*}.}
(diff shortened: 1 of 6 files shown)


teunbrand:
If the code is building a `<theme>` object, no. If they're reading out a `<theme>` object, it might. I could be persuaded that this is a risky move that is not worth the intended benefit of consistency.

thomasp85:
Ah - I see now how you deal with the old spelling inside the code... I'm fine with putting this in the code base - if you want we can run a revdepcheck on it before merging?

thomasp85 (approved):

teunbrand:
Yeah I think a revdepcheck would be wise for this.

PR #6655, opened 2025-09-26
Use ellipsis argument in `theme_sub_*()` family

### LINKED ISSUE #6610: Compatibility between theme_sub_* functions and RStudio’s "Check arguments to R function calls" option
When the "Check arguments to R function calls" option in RStudio is enabled, warnings appear as shown in the image. Disabling this option removes the warnings, but since I would like to use it for other parts of my work, it would be preferable for this feature to be compatible.

Image

Image

teunbrand:
This PR aims to fix #6610.

It does as the title indicates.
In addition to soothing the RStudio code diagnostics, it also ensures we can add or remove arguments in the future.
Adding the ellipsis as first argument also ensures position-based arguments no longer work, which ---as discussed before--- is a good thing.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -5,7 +5,7 @@ about: Submit a bug report to help us improve ggplot2
 
 ### Tips for a helpful bug report:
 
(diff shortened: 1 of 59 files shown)


thomasp85 (approved):
LGTM

PR #6660, opened 2025-09-29
Document `guide_colourbar()` ticks behaviour

### LINKED ISSUE #6640: Make `guide_colourbar()` tick behaviour 'official'
This issue flows forth from discussion at #6527.

Briefly, `guide_colourbar()` interprets a length 2 `legend.ticks.length` setting as setting for the right/bottom and the left/top separately. Note the asymmetric ticks in the example below.

``` r
library(ggplot2)

ggplot(mpg, aes(displ, hwy, colour = cty)) +
  geom_point() +
  theme(
    legend.ticks.length = rel(c(0.5, 0))
  )
```

<sup>Created on 2025-09-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This is an undocumented trick I added in the guide rewrite, but the question is whether we should 'cement' this behaviour to make it official. It'd be a matter of writing down this behaviour in the documentation of `guide_colourbar()` and adding/amending a test to use this behaviour. We shouldn't document this under the `theme(legend.ticks.length)` behaviour, because `guide_bins()` (and most of the legendry package) don't interpret this setting in the described fashion.

teunbrand:
This PR aims to fix #6640.

Briefly, it adds a line of documentation and includes the behaviour in a test.


thomasp85 (approved):
LGTM

PR #6665, opened 2025-09-30
Reorganise plot rendering

teunbrand:
This PR fixes a pet peeve of mine: the build method(s) don't read very nicely.

First off, `ggplot_build()` and `ggplot_gtable()` are in the same file, which is messy because they do things independently and are also documented separately.

Secondly, I understand why `ggplot_build()` is hard to read: it is complicated.
However, `ggplot_gtable()` has a lot of gtable operation complexity, which can be easily hidden away in some choice helper functions. It becomes a lot simpler to understand what `ggplot_gtable()` is doing if the steps are separate functions.


teunbrand:
That's what tests are for!

Issue #6681, opened 2025-10-03
update mentions of RStudio Community?


luisDVA:
Hi,
I noticed that "RStudio community" is still mentioned in several places across ggplot2. Most of the links are updated and point to https://forum.posit.co/, and the redirect still works for the few instances of https://community.rstudio.com/. 

The forum has been rebranded for a few years now so I wonder if it's worth updating the name to Posit Community and when adequate, "Posit Community (formerly RStudio Community)". I'm willing to make a PR for this if the dev team agrees.
Thanks!

teunbrand:
Hi Luis, yes I agree that is should point to the Posit forum instead. We'd be happy to review a PR for this!

teunbrand:
Considered fixed by #6683.

PR #6683, opened 2025-10-06
update mentions of RStudio Community

luisDVA:
discussed in #6681
I've updated all mentions and urls of RStudio Community to posit Community, but added "(formerly RStudio community)" when relevant for context and disambiguation


teunbrand (approved):

PR #6685, opened 2025-10-07
Expression aesthetics

### LINKED ISSUE #6622: expressions as annotations broken
I think this is due to the recent update. It used to work. 

``` r
library(ggplot2)
ggplot(mtcars, aes(x = wt, y = mpg)) + geom_point() +
  annotate("text", x = 3, y = 30, label = latex2exp::TeX("$\\alpha + \\beta x$"))
#> Error in `annotate()`:
#> ℹ Error occurred in the 2nd layer.
#> Caused by error in `list_sizes()` at ggplot2/R/geom-.R:518:3:
#> ! `x$label` must be a vector, not a <latexexpression/expression> object.
```

<sup>Created on 2025-09-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #6638: expressions as tick label broken
This issue was first posted as a reply to [this issue](https://github.com/tidyverse/ggplot2/issues/6622#issuecomment-3321896346). I was asked to make a new issue.
This code worked perfectly a while ago.

```
library(ggplot2)
library(latex2exp)
ggplot(mtcars, aes(x = wt, y = mpg)) +
    geom_point() +
    scale_x_continuous(breaks = 3, labels = TeX("$\\alpha$"))
```
yields

> Error in `vec_size()`:
> ! `x` must be a vector, not a <latexexpression/expression> object.

In a  [previous issue](https://github.com/tidyverse/ggplot2/issues/6622#issuecomment-3321896346) on a similar problem for annotations in a ggplot, specifying `output = character`in the TeX function and `parse = TRUE` in the annotate function works. However, this cannot solve the problem at hand here:

```
ggplot(mtcars, aes(x = wt, y = mpg)) +
    geom_point() +
    scale_x_continuous(breaks = 3, labels = TeX("$\\alpha$", output = character), parse = TRUE)
```
yields:

> Error in scale_x_continuous(breaks = 3, labels = TeX("$\\alpha$", output = character),  : 
>   unused argument (parse = TRUE)

I found a workaround since labelling tick marks with a vector still seemed to work. I add a second label to a tick that I do not show on the graph by setting axis limits:

```
ggplot(mtcars, aes(x = wt, y = mpg)) +
    geom_point() +
    scale_x_continuous(breaks = c(0, 3),
                       labels = TeX(c("",  "$\\alpha$")),
                       limits = c(1.5, 5.5))
```
Another workaround, as suggested by @yutannihilation is using `unclass` or use `parse(text =)` instead of latex2exp.

```
ggplot(mtcars, aes(x = wt, y = mpg)) +
    geom_point() +
    scale_x_continuous(breaks = 3, labels = unclass(latex2exp::TeX("$\\alpha$")))
```

teunbrand:
This PR aims to fix #6622 and fix #6638.

Essentially, whenever an expression `label` comes in, we pack it as a list. In the drawing methods of `GeomText` and `GeomLabel` we unpack the list back to expressions. This way, expressions survive internals.

### DIFF
--- a/R/layer.R
+++ b/R/layer.R
@@ -176,6 +176,7 @@ layer <- function(geom = NULL, stat = NULL,
   if (check.aes && length(extra_aes) > 0) {
     cli::cli_warn("Ignoring unknown aesthetics: {.field {extra_aes}}", call = call_env)
   }
+  aes_params$label <- normalise_label(aes_params$label)
 
   # adjust the legend draw key if requested
   geom <- set_draw_key(geom, key_glyph %||% params$key_glyph)
@@ -552,6 +553,7 @@ Layer <- ggproto("Layer", NULL,
 
     # Evaluate aesthetics
     evaled <- eval_aesthetics(aesthetics, data)
+    evaled$label <- normalise_label(evaled$label)
     plot@scales$add_defaults(evaled, plot@plot_env)
 
     # Check for discouraged usage in mapping
@@ -962,4 +964,18 @@ cleanup_mismatched_data <- function(data, n, fun) {
 
   data[failed] <- NULL
   data
+}
+
+normalise_label <- function(label) {
+  if (is.null(label)) {
+    return(NULL)
+  }
+  if (obj_is_list(label)) {
+    label[lengths(label) == 0] <- ""
+    labels <- lapply(labels, `[`, i)
+  }
+  if (is.expression(label)) {
+    label <- unclass(as.list(label))
+  }
+  label
 }
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ### Bug fixes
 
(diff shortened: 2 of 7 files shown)


R/layer.R
```diff
@@ -963,3 +965,17 @@ cleanup_mismatched_data <- function(data, n, fun) {
...
+normalise_label <- function(label) {
+  if (is.null(label)) {
+    return(NULL)
+  }
+  if (obj_is_list(label)) {
+    label[lengths(label) == 0] <- ""
+    labels <- lapply(labels, `[`, i)
+  }
+  if (is.expression(label)) {
+    label <- unclass(as.list(label))
```
teunbrand, line 978: The `unclass()` idea is from #6639.
It counters length 1 expression subclasses, which is admittedly a niche problem.

thomasp85:
I think there is a small bug but otherwise LGTM

PR #6690, opened 2025-10-09
Only warn about empty aesthetics when data size is known

### LINKED ISSUE #6532: Ignoring empty aesthetics when data has 0 rows should be less intrusive
This was suggested here: https://github.com/cmu-delphi/covidcast/issues/669#issuecomment-3009656882

We can either check later in the plot building progress whether this is the case, e.g. in `Geom$use_defaults()` or demote from warning to message.

``` r
library(ggplot2)

ggplot(mpg[0,], aes(displ, hwy)) +
  geom_point(colour = NULL)
#> Warning in geom_point(colour = NULL): Ignoring empty aesthetic: `colour`.
```

<sup>Created on 2025-06-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

teunbrand:
This PR aims to fix #6532.

In essence we don't want to throw a warning when e.g. `colour = NULL` and also `nrow(data) == 0` because the length matches.
For this, we need to know `data`, which isn't available during the layer construction. For this reason, we move this warning during `compute_geom_2()` when the unmapped aesthetics are applied.

### DIFF
--- a/tests/testthat/test-layer.R
+++ b/tests/testthat/test-layer.R
@@ -29,7 +29,8 @@ test_that("unknown aesthetics create warning", {
 })
 
 test_that("empty aesthetics create warning", {
-  expect_snapshot_warning(geom_point(fill = NULL, shape = character()))
+  p <- ggplot(mtcars) + geom_point(aes(disp, mpg), fill = NULL, shape = character())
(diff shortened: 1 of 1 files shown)


thomasp85 (approved):
LGTM

PR #6715, opened 2025-10-28
Revdepcheck 28-10-2025

thomasp85:

PR #6717, opened 2025-10-30
Fixes for revdep

teunbrand:
This PR aims to fix #6715.

It has two easy fixes and a more complicated one.
The more complicated one has to do with fallback palettes, which affected 2 packages.
The idea is that we have fallbacks for the `fallback.palette` argument in scales.


teunbrand:
>  do you want another revdepcheck after this?

Yes please! To confirm that these problems are now fixed and which vignette errors remain.

PR #6730, opened 2025-11-05
Centralise documentation for common layer arguments

teunbrand:
This PR is upkeep and aims to fix a pet peeve.

The pet peeve is that there were multiple documentation entries for common parameters and tracking the inheritance of some layer parameters is a bit of a pain. This puts all the common parameters in one place. I also replaced the `rd_orientation()` helper, as it essentially pasted a section that can also be just inherited.


teunbrand:
> If `@noDoc` still allows inheritance I think it is preferred over `@keyword` internal

`@noDoc` isn't recognised by roxygen2, but putting `@noRd` there causes roxygen2 to not find the inherited parameters.
The `@aliases NULL` already 'hides' the page from autocompleting when typing `?` and I don't know of any other further steps to obfuscate this from users. I'll merge this and if any bright ideas pop up later, we can always adapt.

PR #6733, opened 2025-11-06
Synchronise lifecycle management

### LINKED ISSUE #6731: Lifecycle management
Whenever we bump deprecations, like in #6138, we still need to manually edit all the functions to swap out soft-deprecations for full deprecations and swap out full deprecations for defunct declarations. This is a bit of of an annoyance to do and we might be happier in the long term if we manage this is a more systematic way.

teunbrand:
This PR aims to fix #6731.

It replaces the `deprecate_warn0()` and `deprecate_soft0()` internal functions with a new function: `deprecate()`.
If we want to advance the lifecycle of deprecated functions, we just need to edit the versions in `deprecate()` instead of finding an replacing functions all over the place. There is also an option to escalate a deprecation to defunct or warning ahead of the cycle.

### DIFF
--- a/R/aes-variants.R
+++ b/R/aes-variants.R
@@ -127,6 +127,6 @@ aes_all <- function(vars) {
 #' @keywords internal
 #' @export
 aes_auto <- function(data = NULL, ...) {
-  lifecycle::deprecate_stop("2.0.0", "aes_auto()")
+  deprecate("2.0.0", "aes_auto()")
 }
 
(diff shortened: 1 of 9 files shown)


teunbrand:
> It would be nice if deescalation was possible as wel

I agree that this would be nice to have, but doing it should be discouraged. I've added an option to postpone a cycle.

teunbrand:
Found a few cases of `lifecycle::deprecate_stop()` that could be folded into this mechanism, so I did.

PR #6738, opened 2025-11-10
Parity between folders in `R/` and in `tests/testthat/`

teunbrand:
This PR is upkeep and has no related issue.

It brings ggplot2 into the convention that there are pairs of code and test files, making it easy to spot what some test file is testing and where to find tests associated with a code file. To do this, I often:

* Renamed test files, to match the `R/` file.
* Renamed some `R/` files to better its intent, for example:
    * `aes-evaluation.R` was about the delayed evaluation mechanism, not evaluation in general.
    * `performance.R` contained utilities, so it seemed more clear to me to name this `utilities-performance.R` instead.
 * Merged files when testing was also tightly intertwined.
     * `geom-col.R` and `geom-bar.R` are essentially the same thing
     * The 'abline', 'hline' and 'vline' geoms now live in a single `R/geom-abline-hline-vline.R` file.
 * Split up test files when implementation was spread out over multiple files.
     * `test-theme.R` contained all tests for themes, whereas there are 5 `theme-{subject}.R` files.
     * Facet testing was along procedural lines, which is now split between `facet_grid()` and `facet_wrap()`.
There are a few notable exceptions to pairing:

* Several `R/` files, lack a corresponding test file. Typically this occurs when functionality is already implicitly covered in another test, like `geom_density()` being covered by tests for `stat_density()`.
* There is a 'meta' test in `tests/testthat/test-prohibited-functions.R` that tests the repository as a whole.

Moreover there may be a few surprises:

* I've merged some files with related utility when it just made sense


teunbrand:
Had to wrestle in the mud with one particular test that locally passing but failing on CI, but I think it won't misbehave now.

PR #6745, opened 2025-11-12
Don't bother subsetting labels

teunbrand:
The behaviour removed here was discussed in https://github.com/tidyverse/ggplot2/pull/3740#discussion_r367380217.
It was added just to be sure specifically for scale labels. Howveer, it is a hindrance as a global behaviour because people have wild inputs for the `label` aesthetic. I think the best solution here is to just drop this defensive behaviour and let people deal with bad input themselves.

### DIFF
--- a/R/layer.R
+++ b/R/layer.R
@@ -972,10 +972,8 @@ normalise_label <- function(label) {
     return(NULL)
   }
   if (obj_is_list(label)) {
-    # Ensure that each element in the list has length 1
+    # Ensure no elements are empty
     label[lengths(label) == 0] <- ""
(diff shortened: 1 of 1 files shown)


aphalo:
@teunbrand Hi Teun, I should have known better long ago when I wrote 'ggpp' than to piggy-back on the `label` aesthetic using it with lists. Sorry! Let me know if a code breaking renaming of the aesthetic in 'ggpp' is necessary.

aphalo:
Some visual tests in 'ggpp' were failing, but the OLD references were wrong, not the current output with the 'ggplot2' rc/4.0.1. So, the failures were incomsequential. I committed the new snaps and pushed them to GitHub.

PR #6747, opened 2025-11-17
Use namespace prefixes in `find_scale()`/`validate_guide()`

### LINKED ISSUE #4705: Support scale functions in packages not attached via scale_type()

teunbrand:
This PR aims to fix #4705 and replaces #4799.

Essentially it allows you to specifiy a scale/guide using the `pkgname::`-prefix.
By declaring the namespace we only need to search one namespace and we can avoid searching in *all* loaded namespaces.
Example guides:
``` r
#> ℹ Loading ggplot2

x <- validate_guide("legendry::colring")
is_guide(x)
#> [1] TRUE

x <- validate_guide("ggraph::edge_direction")
is_guide(x)
#> [1] TRUE
```
Example scale:
```r
registerS3method(
  "scale_type", "pillar_num", 
  method = function(x) c("pillar::num")
)

is_attached("pillar")
#> [1] FALSE

# Note the class name ends with 'Num'
find_scale("x", pillar::num(1))
#> <ScaleContinuousPositionNum>
#>  Range:  
#>  Limits:    0 --    1
```

<sup>Created on 2025-11-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Practically, packages that implement this may want to declare both the unprefixed and prefixed versions while transitioning. 
For example, the definition of the pillar method ([here](https://github.com/r-lib/pillar/blob/30bbeb2d98e2e196d1a80ad23bc398f3d99136c8/R/ggplot2.R#L94)) may look like this:

``` r 
scale_type.pillar_num <- function(x) c("num", "pillar::num", "continuous")`
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # ggplot2 (development version)
 
+* Scales and guides specified by a string can now use package name prefixes to
+  indicate a namespace wherein to look for the scale/guide. For example, one can
+  use `scale_x_continuous(guide = "legendry::axis_base")` (@teunbrand, #4705).
 * `get_layer_data()` and `get_layer_grob()` now accept layer names as index 
   (@lgaborini, #6724)
 * Added new argument `geom_curve(shape)` that will be passed down to 


teunbrand:
Setting this as a draft until I found a way to test this

teunbrand:
TODO: use pillar for test along with `skip_on_cran()`

teunbrand:
Instead of adopting earlier advice on using external packages to test this, I'm now using a mocked `asNamespace()` for testing to return an environment of my choosing.

R/guides-.R
```diff
@@ -915,8 +915,19 @@ include_layer_in_guide <- function(layer, matched) {
...
-    fun <- find_global(paste0("guide_", guide), env = global_env(),
-                       mode = "function")
+    check_string(guide, allow_empty = FALSE)
+    search_env <- list(global_env())
+    if (isTRUE(grepl("::", guide))) {
+      # Append prefix as namespaces to search environments
+      prefix <- sub("::.*", "", guide)
+      search_env <- c(search_env, list(as_namespace(prefix)))
+      # Remove prefix from guide name
+      guide <- sub(".*::", "", guide)
```
thomasp85: Might be cleaner to do `strsplit(guide, "::", fixed = TRUE)` instead of two separate regexes

PR #6756, opened 2025-11-25
Initial infrastructure for editions: attempt 2

teunbrand:
This PR aims to adress a topic discussed off-github and replaces #6203.

Compared to #6203, this edition infrastructure is simpler. In this PR, we're not tracking parallel editions set by packages; it is only controlled by the user. It has two arms: the carrot and the stick.

The carrot is `edition_require()`, which we can use to gatekeep new features tempting users to opt-in the newer editions.
The stick is build into the `deprecate()` function from #6733 and the new `supersede()` function, meaning that superseded and deprecated functions become unavailable when the new edition is set.

Broadly, I'm still skeptical of the edition concept. What we really want is to clean up the codebase and remove code we no longer recommend and make maintaining ggplot2 easier. The main thing preventing this goal is that the code in question is still in use by packages and in people's scripts. However, what editioning really seems to do is put additional maintenance burden around lifecycle management without achieving the main benefit of removing the code.


thomasp85 (approved):
LGTM

teunbrand:
There was some useful feedback from the chat I had with the team on Monday.
Essentially `set_edition()` should become `set_ggplot2_edition()` to avoid namespace clashes, and we should have a `local_ggplot2_edition()` function.

teunbrand:
@thomasp85 Do you think we need to export *just* the setter, or also other helpers?

teunbrand:
I've opted for also exporting some other utilities

PR #6766, opened 2025-12-05
Add check function for named-ness

teunbrand:
This PR does not fix a posted issue.

I'd like this check function to use in #6196. I'm filing this PR separately, because it has the potentially breaking change of disallowing duplicate arguments to `ggproto()`.

### DIFF
--- a/R/stat-bindot.R
+++ b/R/stat-bindot.R
@@ -6,7 +6,9 @@ StatBindot <- ggproto("StatBindot", Stat,
   required_aes = "x",
   non_missing_aes = "weight",
   default_aes = aes(y = after_stat(count)),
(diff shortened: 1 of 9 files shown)


R/stat-bindot.R
```diff
@@ -122,9 +124,7 @@ StatBindot <- ggproto("StatBindot", Stat,
       data$x <- midline
     }
     return(data)
-  },
-
-  dropped_aes = c("weight", "bin", "bincenter")
```
teunbrand, line 127: By adding this check to `ggproto()`, I found that this was a duplicated entry.

thomasp85 (approved):
LGTM

PR #6772, opened 2025-12-10
Track ggplot2 version in plot

### LINKED ISSUE #6714: Version info stored and can be queries in ggplot object
Similar issue to 6637 where old versions of ggplot object does not display correctly.

Please add an item to the ggplot objects with the version of ggplot2 that created them than can be queried by the user. This way, I can alert the user that the saved ggplot object may have to regenerated.

The issue is that the code that generated the data used for the object is long running MCMC program that saves the output  to a directory for later usage. These include basic ggplot plots  saved for the user to annotate when the program is done using quarto etc.

If the version is saved with the the ggplot object, I can query the plot object and warn the user if a breaking change has occurred.

Thanks
Carl Schwarz
cschwarz.stat.sfu.ca@gmail.com

teunbrand:
This PR aims to fix #6714.

Whenever a plot is created, the new `version` property is populated with the current version.
If a plot is to be rendered, an error is thrown if the plot's version is not the same as the current version.
This should typically occur when a plot has been saved to disk with a prior version of ggplot2, and now no longer works.

There are a few open questions:
* Should we do this at all?
* Should we throw a warning instead of an error?

We can also do a revdepcheck to weed out all the packages that misbehave.

### DIFF
--- a/R/all-classes.R
+++ b/R/all-classes.R
@@ -337,6 +337,7 @@ class_ggplot <- S7::new_class(
     layout  = class_layout,
     labels  = class_labels,
     meta    = S7::class_list,
+    version = S7::new_S3_class("package_version"),
     plot_env = S7::class_environment
   ),
   constructor = function(
@@ -368,6 +369,7 @@ class_ggplot <- S7::new_class(
       layout      = layout %||% ggproto(NULL, Layout),
       labels      = labels,
       meta        = meta,
+      version     = current_version,
       plot_env    = plot_env
     )
   }
--- a/R/plot-build.R
+++ b/R/plot-build.R
@@ -31,6 +31,14 @@ ggplot_build <- function(plot, ...) {
   if (!is.null(env)) {
     attach_plot_env(env)
   }
+  if (!identical(try_prop(plot, "version"), current_version)) {
+    cli::cli_abort(c(
+      "The plot object was created with a different version of ggplot2. \\
+      Please reconstruct the plot.",
+      i = "Did you save the plot object to disk?"
+    ))
+  }
+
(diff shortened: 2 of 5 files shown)


aphalo:
@teunbrand Hi Teun! This is an interesting question. If you decide to implement this, I suggest using a warning, unless you are sure all old plots will not render at all. Sometimes, it can be useful to get some idea of what a plot is about, even if inperfectly rendered. Another thought is that it would be better to use "object versions" than 'ggplot2' versions. Why issue an error for a change from version 4.0.1 to 4.0.2? I would change the label only when there is a significant compatibility break. In such a case, the warning (or error), could give the 'ggplot2' versions that can render the object as a range, which would be a lot friendlier to users in the case they cannot rebuild the object.
I (almost) never save ggplot objects to disk. So, as you mentioned in #6714, the main solution is to discourage users from saving these objects to disk. However, nothing prevents users from saving objects or expecting this to work, say sharing a ggplot object with somone who does not have the code to regenearte it. Another possible situation is a user working interactively at the command prompt, saving the R workspace, and expecting to be able to render the objects in the workspace at a later time, possibly after an update to 'ggplot2'. So I think that a coarser grained test, triggering an error only when rendering is not possible would be more helpful. (I am assuming rendering is possible at least across minor versions.)

teunbrand:
Thanks for your thoughts Pedro! I get that this would be easier from a user perspective, but from a maintenance point of view it'd become tedious to test and verify if there is between-version compatibility and doing the bookkeeping. It is not even guaranteed for minor version updates. People can share plots with themselves and oneanother as data + code, or as graphics, but shouldn't do this as ggplot objects.

PR #6779, opened 2025-12-18
Violin quantiles

### LINKED ISSUE #6757: geom_violin(): draw_quantiles deprecation message is incomplete and has a typo
Repex for this is actually [in the docs](https://ggplot2.tidyverse.org/reference/geom_violin.html), which include an example of the deprecation warning.
```r
geom_violin(..., draw_quantiles = c(0.25, 0.5, 0.75))
Warning message:
The `draw_quantiles` argument of `geom_violin()` is deprecated as of ggplot2 4.0.0.
ℹ Please use the `quantiles.linetype` argument instead.
```
Looks like the warning should probably read
```R
ℹ Please use the `quantiles` argument with the `quantile.linetype` argument instead.
```
for clarity as the replacing syntax is apparently
```R
geom_violin(..., quantiles = c(0.25, 0.5, 0.75), quantile.linetype = "solid")
```
and there is no `quantiles.linetype` argument.

Some confusion over needing to move `quantile.linetype` off its default as well as rename `draw_quantiles` to `quantiles` seems likely, so likely it'd be good to update the docs to show an example of the new syntax in addition to `draw_quantiles`' deprecation.

BrentWHR:
This PR should resolve #6757

Provides better examples on working with quantiles in geom_violin and gives more detailed information in the deprecation message when using argument draw_quantiles

### DIFF
--- a/R/geom-violin.R
+++ b/R/geom-violin.R
@@ -67,6 +67,9 @@
 #' # Show quartiles
 #' p + geom_violin(quantile.linetype = 'solid')
 #'
+#' # Show different quantiles
+#' p + geom_violin(quantiles = c(0.2, 0.4, 0.6, 0.8), quantile.linetype = 'solid')
+#'
 #' # Scales vs. coordinate transforms -------
 #' if (require("ggplot2movies")) {
 #' # Scale transformations occur before the density statistics are computed.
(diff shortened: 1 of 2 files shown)


teunbrand (approved):
Lovely, this looks good to me: thank you for the contribution!

PR #6781, opened 2025-12-19
Add argument for minimal group size in boxplot

### LINKED ISSUE #6776: Request change to boxplot code to handle tiny data sets better
In situations when we want to make side-by-side boxplots, ideally if there are less than 5 points that group is shown as dots. 

Here is the current behaviour:

Image

and this is the desirable behaviour:

Image

Notice the samples at NT and ACT only have 3 and 2 points respectively, so showing them as boxplots is misleading.

A diff of the code change is at:

[https://github.com/dicook/ggplot2/commit/7dbc03e52dfb6c19696dfddd6adeee5a33b2981b](https://github.com/dicook/ggplot2/commit/7dbc03e52dfb6c19696dfddd6adeee5a33b2981b)

teunbrand:
This PR aims to fix #6776.

Instead of mirroring https://github.com/dicook/ggplot2/commit/7dbc03e52dfb6c19696dfddd6adeee5a33b2981b, the threshold for what constitutes a small group is set via an argument. This preserves the current behaviour, while allowing for the requested behaviour. I'm still open for better argument names than `min.group.n`.

Note in example below, the first two groups are rendered as points without box and whiskers.

``` r
#> ℹ Loading ggplot2
df <- data.frame(
  x = rep(LETTERS[1:4], c(1, 5, 10, 50)),
  y = rnorm(66)
)

ggplot(df, aes(x, y)) +
  geom_boxplot(min.group.n = 6)
```

<sup>Created on 2025-12-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # ggplot2 (development version)
 
+* New `stat_boxplot(min.group.n)` argument, which skips drawing the box and 
+  whiskers for small groups (@teunbrand based on code by @dicook, #6776)
 * The `arrow` and `arrow.fill` arguments are now available in 
   `geom_linerange()` and `geom_pointrange()` layers (@teunbrand, #6481).
 * (internal) `zeroGrob()` now returns a `grid::nullGrob()` (#6390).
--- a/R/geom-boxplot.R
+++ b/R/geom-boxplot.R
@@ -304,6 +304,27 @@ GeomBoxplot <- ggproto("GeomBoxplot", Geom,
       ))
     }
 
+    outliers_grob <- NULL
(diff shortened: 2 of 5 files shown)


jospueyo:
I was about to issue exactly this. This would be very helpful to me at this moment. May I do something to speed up the process?

teunbrand:
This PR would need a review from another maintainer and then the repo as a whole would need to wait until we do a new release. If you need t his *today*, I think you could indicate to download a PR via `pak::pak()`, but you may have to puzzle the syntax together a bit.

PR #5514, opened 2023-11-09
Stats imports

klin333:
Add utils:: and stats:: in front of functions not explicitly imported.

Fixes #5186

Issue #5186, opened 2023-02-13
Several occurrences of unimported `stats` and `utils` function calls


JorisChau:
There appear to be several occurrences of `stats` and `utils` function calls that are not imported and not declared explicitly (as e.g. `stats::approx()` or `utils::getFromNamespace()`):

```{r}
* Source code references:
  approx at axis-secondary.R#239:
    239: old_val <- approx(full_range, old_range, range_info$major_source)$y
  approx at axis-secondary.R#255:
    255: old_val_minor <- approx(full_range, old_range, range_info$minor_source)$y
  approxfun at stat-align.R#72:
    72: y_val <- approxfun(data$x, data$y)(unique_loc)
  ecdf at stat-ecdf.r#102:
    102: data_ecdf <- ecdf(data$x)(x)
  getFromNamespace at backports.R#4:
    4: absolute.units <- getFromNamespace("absolute.units", "grid")
  getFromNamespace at backports.R#5:
    5: absolute.units.unit <- getFromNamespace("absolute.units.unit", "grid")
  getFromNamespace at backports.R#6:
    6: absolute.units.unit.list <- getFromNamespace("absolute.units.unit.list", "grid")
  getFromNamespace at backports.R#7:
    7: absolute.units.unit.arithmetic <- getFromNamespace("absolute.units.unit.arithmetic", "grid")
  quantile at stat-qq-line.R#79:
    79: y_coords <- quantile(sample, line.p)
```

klin333:
agree this gets very annoying for people that set up R session to NOT have stats package in namespace by default. i suppose this is a wider problem not restricted to ggplot2. some might say why on earth would anyone not load stats by default, well, usually it's to prevent namespace clutter. 

is there a way to systematically check this package for usage of stats functions not explicitly imported? maybe run the unit tests of this package without stats loaded by default?

klin333:
FYI quick hack code to search for instances of naked library calls

```r
library(tidyverse)

pkg <- 'stats'
pkg_exports <- ls(str_c('package:', pkg))

src_files <- list.files('R', '*.R', recursive = TRUE, full.names = TRUE)
src_file_lines <- set_names(map(src_files, read_lines), src_files)

result_df <- expand_grid(file = src_files, export = pkg_exports) %>%
  mutate(found_naked = map2_lgl(file, export, function(f, i) {
    lines <- src_file_lines[[f]]
    # may miss usage as variables but otherwise too many false positives
[...]

PR #6793, opened 2026-01-20
add stats:: in front of stats functions that were not explicitly imported

klin333:
This problem of not explicitly putting `stats::` in front of stats package functions was previously documented in #5186. Previously the fix was done in eg #5514

This PR is similar, cleaning up more calls to stats functions that do not have `stats::` in front of them.

### DIFF
--- a/R/position-dodge.R
+++ b/R/position-dodge.R
@@ -150,13 +150,13 @@ PositionDodge <- ggproto("PositionDodge", Position,
     }
 
     data$order <- xtfrm( # xtfrm makes anything 'sortable'
-      data$order %||% stats::ave(data$group, data$x, data$PANEL, FUN = match_sorted)
+      data$order %||% vec_ave(data$group, data[c("x", "PANEL")], fn = match_sorted)
     )
     if (isTRUE(params$reverse)) {
       data$order <- -data$order
     }
     if (is.null(params$n)) { # preserve = "total"
-      data$order <- stats::ave(data$order, data$x, data$PANEL, FUN = match_sorted)
+      data$order <- vec_ave(data$order, data[c("x", "PANEL")], fn = match_sorted)
     } else { # preserve = "single"
       data$order <- match_sorted(data$order)
     }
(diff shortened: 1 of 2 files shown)


klin333:
hi maintainers, maybe @teunbrand? Could you help allow this merge. the check failures seems unrelated to this PR, though i'm not sure... thank you.

Breeze-Hu:
Did the previous PR replace `stats::ave` with `vec_ave`?  https://github.com/tidyverse/ggplot2/pull/6698
Is this missing? Rather than using `stats::ave`?

teunbrand:
@Breeze-Hu is right. Instead of prefixing the `ave()` function, can you replace it with `vec_ave()` instead?

klin333:
thanks guys, but i don't think it's a simple replace for those remaining stats::ave to vec_ave

These remaining stats::ave has multiple grouping variables, but the vec_ave only supports 1 grouping variable 
https://github.com/tidyverse/ggplot2/blob/1052371e856b2739c1efa4f8ad2fac8e171d218d/R/position-dodge.R#L152-L154
https://github.com/tidyverse/ggplot2/blob/1052371e856b2739c1efa4f8ad2fac8e171d218d/R/utilities.R#L938-L947

I think this PR is unrelated the to stats::ave -> vec_ave change from #6698. Any way we can allow this PR through without being blocked by these unrelated changes? Ideally there should be a continuation of #6698 separately, assuming that PR simply missed these stats::ave, again probably because these weren't prefixed with stats::. 

This forgetting of prefixing stats:: just keeps creeping up year after year, to the point where perhaps it's better to just import all of stats, or configure testthat to run all test suite without stats imported at R startup. But that's above my pay grade, probably @teunbrand 

https://github.com/tidyverse/ggplot2/blob/1052371e856b2739c1efa4f8ad2fac8e171d218d/R/ggplot2-package.R#L4-L10

teunbrand:
I've added the `vec_ave()` myself.

> This forgetting of prefixing stats:: just keeps creeping up year after year

Yeah it just part of the general fight against entropy. I'd not be happy to wholesale import stats, but it might be fine if somebody wants to contribute a separate PR that creates a unit test to guard against this.

teunbrand (approved):

PR #6797, opened 2026-01-26
Wrap `list2` usage in `make_contructor()`

### LINKED ISSUE #6796: object 'ffi_list2' not found: build time capture of `rlang::list2()` has resulted in unfortunate errors for users
TLDR for users hitting

> object 'ffi_list2' not found

I believe if you do the following, you can get back up and running:

- `remove.packages("rlang")`
- Restart R to be safe
- `install.packages("rlang")`
- Restart R to be safe
- Try your code again

That should put you on rlang 1.1.7, which has `ffi_list2`.

---

ggplot2 seems to capture `rlang::list2()` at _package build time_, meaning that whatever definition of `rlang::list2()` is available at build time is the one that all users of ggplot2 must have.

https://github.com/tidyverse/ggplot2/blob/4ea78a42bd37ef36111f1f098502ed03be262fde/R/make-constructor.R#L142-L143

This constructor seems to then be called at package build time in many places, like

https://github.com/tidyverse/ggplot2/blob/4ea78a42bd37ef36111f1f098502ed03be262fde/R/geom-contour.R#L84

In rlang 1.1.7, we changed the internals of `list2()` to directly drop into a new C function called `ffi_list2`.

Unfortunately I believe this can cause a large variety of errors.

See https://github.com/r-lib/rlang/issues/1866

And https://stackoverflow.com/questions/79869479/cannot-plot-geom-point-in-r-ggplot2

Based on looking at these problems, my diagnosis is:

- CRAN builds ggplot2 4.0.1 binaries against rlang 1.1.7 with the `ffi_list2` definition. This captures that definition at _ggplot2 binary build time_ for _all_ users who download ggplot2 from CRAN.
- A user downloads ggplot2 4.0.1. This does _not_ necessarily prompt them to update rlang to 1.1.7.
- They have an older rlang and try and plot, this results in the "`ffi_list2` not found" error seen in all of the reports.

I believe there is nothing to be done on the rlang side of things.

If ggplot2 could please not capture `rlang::list2()` at package build time, and could send out a patch release that fixes this, then that would be great!

teunbrand:
This PR aims to fix #6796.

It prevents attach `rlang::list2()` as a variable direclty. I've tested the built time dependency by building ggplot2 with rlang 1.1.6, upgrading to rlang 1.1.7 and seeing if there are warnings when using the fruits of `make_contructor()`. There were no longer such warnings. Note PR targets the rc/4.0.2 branch, not the main branch.

### DIFF
--- a/R/make-constructor.R
+++ b/R/make-constructor.R
@@ -140,7 +140,7 @@ make_constructor.Geom <- function(x, ..., checks = exprs(), omit = character(),
   body <- call2("{", !!!checks, body)
 
   # We encapsulate rlang::list2
-  new_env <- new_environment(list(list2 = list2), env)
+  new_env <- new_environment(list(list2 = list2_wrapper), env)
 
   new_function(fmls, body, new_env)
 }
@@ -225,8 +225,11 @@ make_constructor.Stat <- function(x, ..., checks = exprs(), omit = character(),
   body <- call2("{", !!!checks, body)
 
   # We encapsulate rlang::list2
-  new_env <- new_environment(list(list2 = list2), env)
+  new_env <- new_environment(list(list2 = list2_wrapper), env)
 
   new_function(fmls, body, new_env)
 }
 
+list2_wrapper = function(...) {
+  rlang::list2(...)
+}


thomasp85 (approved):
LGTM

PR #6801, opened 2026-02-03
Update startup tip to extension gallery

iamYannC:
I replaced the Stack Overflow tip with a pointer to the ggplot2 extension gallery.

I think it could get a larger shakedown with tips on using your favorite LLM but idk, didnt want to get over my head

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -13,7 +13,7 @@ random_tip <- function() {
     "Learn more about the underlying theory at https://ggplot2-book.org/",
     "Keep up to date with changes at https://tidyverse.org/blog/",
     "Use suppressPackageStartupMessages() to eliminate package startup messages",
-    "Need help? Try Stackoverflow: https://stackoverflow.com/tags/ggplot2",
+    "Explore the ggplot2 extension gallery to extend your plotting capabilities: https://exts.ggplot2.tidyverse.org/",
     "Need help getting started? Try the R Graphics Cookbook: https://r-graphics.org",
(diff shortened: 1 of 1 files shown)


iamYannC:
I have no idea what made some of the workflows to fail but i can guarantee it has nothing to do with my actual work ;)

teunbrand:
I'm not against the new tip, but why should the stackoverflow tip be removed? Sure it is not as active as it once was, but ggplot2 questions still get answered regularly.

iamYannC:
> I'm not against the new tip, but why should the stackoverflow tip be removed? Sure it is not as active as it once was, but ggplot2 questions still get answered regularly.

Realistically, SO is past it's glory. When this message popped up I felt nostalgic even 😀
I'm willing to just add my new suggestion if this feels crucial for the team, but personally (as a user i mean), suggesting stack overflow these days is just not as great of a tip

teunbrand:
Past it's glory or not, I still think this is the right place to point to if people want human help, alongside the posit community fora.

thomasp85:
I agree - no reason to remove that

iamYannC:
> I agree - no reason to remove that

Should I modify (add back SO)  or just close the PR and leave it for now?

teunbrand:
Yes please resurrect the SO tip. The git history within a PR doesn't matter too much because we squash-merge them anyway.

iamYannC:
> Yes please resurrect the SO tip. The git history within a PR doesn't matter too much because we squash-merge them anyway.

done. thanks for the quick loop

teunbrand (approved):
Perfect, thank you for the contribution!

PR #6808, opened 2026-02-20
`GeomMap$draw_panel()` uses `ggname()`

### LINKED ISSUE #6807: `GeomMap$draw_panel()` does not use `ggname()` to edit prefix when returning its `polygonGrob()`
* Most Geom's `draw_panel()` methods use `ggname()` to edit the name of the returned grob but for some reason `GeomMap` does not: https://github.com/tidyverse/ggplot2/blob/28ec18f804053130d9a10bc737d7f93fe3cbdb4d/R/geom-map.R#L145
* I wasn't sure if this is deliberate or an oversight but when I briefly searched I couldn't see a discussion for why this Geom should be an exception.

trevorld:
closes #6807


teunbrand (approved):
Looks good to me, thank you for the contribution!

PR #6834, opened 2026-04-07
Update broken blogpost links

arcresu:
The URL for a few older blogposts changed. `https://www.tidyverse.org` redirects to `https://tidyverse.org`, so I removed the subdomain also from more recent entries.


teunbrand (approved):
Great, thanks for keeping an eye out for this and the contribution!

PR #6835, opened 2026-04-09
Update Plausible analytics snippet

jeroenjanssens:
## Summary

- Replace the old Plausible `<script defer data-domain=...>` tag with the new async snippet that includes `plausible.init()`

## Context

Plausible has updated their recommended tracking snippet. The new format uses `async` loading with an inline init script, and no longer requires the `data-domain` attribute.


teunbrand (approved):
Thanks Jeroen, I'm trusting you on this one

Issue #6831, opened 2026-04-03
`geom_vline()` and friends cannot use `mapping` when `xintercept` is set


arcresu:
I wanted to make `geom_vline()` adopt the theme's accent colour. Below was my sequence of attempts:

``` r
library(ggplot2)

ggplot() + geom_vline(xintercept = 0, colour = from_theme(accent))
#> Error:
#> ! object 'accent' not found
ggplot() + geom_vline(aes(colour = from_theme(accent)), xintercept = 0)
#> Warning: `geom_vline()`: Ignoring `mapping` because `xintercept` was provided.
```

![](https://i.imgur.com/zXZf9cH.png)

``` r
ggplot() + geom_vline(aes(xintercept = 0, colour = from_theme(accent)))
```

![](https://i.imgur.com/2lYRHzq.png)

<sup>Created on 2026-04-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The first is obviously wrong because `from_theme()` only works inside `aes()`. The second seems like it should work, but the geoms are being too strict.

In the past I understand why it might have been discouraged to use `mapping`, but I think with the newer features like `from_theme()` the situation changed a bit and the limitation seems unnecessary.

Would it be worth making `geom_vline(aes(colour = from_theme(accent)), xintercept = 0)` work?

EvaMaeRey:
I think there is a bit more complexity in this case as well as with geom_hline and geom_abline because they were created to act as an annotation layers when xintercept is provided, https://github.com/tidyverse/ggplot2/issues/6559.  

I think that given use of unmapped xintercept triggers the annotation behavior, perhaps there shouldn't be the expectation to be able to do more than can be done with annotate(), where mapping isn't in play?  e.g.  something like the following? 

```r
get_accent <- function(){ theme_get()$geom$accent } 

ggplot() +
     annotate(geom = "point", x = 0, y = 0, color = get_accent())
```

But, only works if theme that you are using is set, not applied with `+`.

arcresu:
The issue is that when any intercept/slope is provided, the provided mapping is discarded completely and replaced with one that only uses the internally constructed data frame. My proposal would be to still create that data frame but to update the existing mapping rather than replacing it. I can't see an inherent problem with that approach but might be overlooking some way it would create confusion. On the contrary it feels like it would improve consistency with other geoms where constant aesthetics can be passed as params instead of mappings.

[...]

Issue #6833, opened 2026-04-07
variable inherit.aes behavior for geom_vline and friends


EvaMaeRey:
In #6633 the inherit.aes behavior is exposed to the user, but the default is set to FALSE, which means that that layer is not in sync with most user-facing geom_* layers.  I think setting this to NULL (so still exposing to user), and turning inheritance off if xintercept is non-null would feel more consistent when this geom_* family is used in a non-annotative way.  Thoughts?

teunbrand:
It is not that fresh in memory anymore, but I think the default is set to FALSE to preserve backward compatibility. As long as that is preserved, I have no qualms with the proposed behaviour.

EvaMaeRey:
Great!  I've sometimes thought that x could/should be an alternate for the xintercept aesthetic for geom_vline(), and y for yintercept... (but leaving it so setting xintercept and yintercept triggers annotation behavior).   But maybe there is not analogue for geom_abline. Still I wonder if there are reasons for not pursuing those changes.   I'll put aside for now to focus on the inherit.aes questions!

EvaMaeRey:
It feels like the geom_abline is a little different in kind because of the slope, intercept defaults  (1 and 0 if not provided) whereas this doesn't happen for geom_hline and geom_vline with  xintercept and yintercept.  Perhaps inherit.aes needs to stay FALSE given these internally provided defaults for the geom_abline case - will give it some more thought!

PR #6782, opened 2025-12-19
docs: clarify how vline, hline, and abline affect position scales

tanmaydimriGSOC:
This PR updates the documentation for geom_vline(), geom_hline(), and geom_abline() to clarify that these geoms can affect position scales due to their position aesthetics (xintercept, yintercept, slope, intercept).

This reflects the current behavior introduced when position aesthetics were consolidated (#3342), and avoids suggesting that these geoms never affect scales.

Closes #6768.

PR #6836, opened 2026-04-10
change geom_vline and geom_hline inherit.aes parameter default to NULL

EvaMaeRey:


R/geom-abline-hline-vline.R
```diff
@@ -129,7 +129,10 @@ geom_hline <- function(mapping = NULL, data = NULL,
                        yintercept,
                        na.rm = FALSE,
                        show.legend = NA,
-                       inherit.aes = FALSE) {
+                       inherit.aes = NULL) {
+
+    # if NJLL default is based on if xintercept is missing
```
arcresu, line 134:
```suggestion
    # if NULL, default is based on whether xintercept is missing 
```
(And same below).

arcresu:
I wonder whether the docs should also be updated to explain how the behaviour changes based on `xintercept`'s presence (see also #6782).

teunbrand:
Gina does this PR aim to fix #6831?

arcresu:
I think this is for #6833, which is separate. If this is merged, the `mapping` would still be thrown out when `xintercept` is specified, which is what #6831 is about. I updated the title there to hopefully be clearer.

EvaMaeRey:
That's right, @arcresu.  This is about #6833 ...

PR #6839, opened 2026-04-11
Correct links for ggplot2 version history

JosephBARBIERDARNAL:
Hi! Just a small PR to correct invalid urls in the news section. For example, it currently has: https://tidyverse.org/articles/2019/06/ggplot2-3-2-0/, while the actual url should be https://tidyverse.org/blog/2019/06/ggplot2-3-2-0/

Other urls in the news section are valid.


arcresu:
See also #6834

JosephBARBIERDARNAL:
Oops! didn't see it, closing as dupplicat

PR #6851, opened 2026-04-16
Drop `Suggests: RColorBrewer`

arcresu:
This does not correspond to an issue, but by chance I noticed that the only place `RColorBrewer` was used was in the script for generating icons. I replaced it with the equivalent call to `pal_brewer()` and dropped the dependency. The only remaining reference to `RColorBrewer` is in `man/scale_brewer.Rd`, pulled in indirectly by `@inheritParams scales::pal_brewer`.

The unrelated bugfix is just because the icon script wouldn't run as `g` was undefined at that line.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -65,7 +65,6 @@ Suggests:
     quantreg,
     quarto,
     ragg (>= 1.2.6),
-    RColorBrewer,
     roxygen2,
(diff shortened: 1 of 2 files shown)


teunbrand (approved):
Neat simplification, it looks good to me. Thank you for the contribution!

PR #6854, opened 2026-04-22
docs: replace stateful ggproto example with stateless one

### LINKED ISSUE #6582: Example in `?ggproto` counterproducive
The example in `?ggproto` gives an example of a *stateful* class (.e.g `Adder$x`) is a stateful field.
It advertises a property of ggproto classes that should be used sparingly and ideally not at all if you're developing extensions.

CuiweiG:
Closes #6582.

The current `?ggproto` example demonstrates a stateful class (`Adder$x`) that, per the issue, "advertises a property of ggproto classes that should be used sparingly and ideally not at all if you're developing extensions."

The replacement builds a stateless `Math` class with two static methods plus a `Mather` child overriding one method via `ggproto_parent()`. Construction, method dispatch, and parent-method invocation are all still covered; no `self$*` field access appears anywhere. The child method carries `self` in its signature solely to thread identity into `ggproto_parent(Math, self)`.

The "Working with ggproto classes" section gains a short paragraph stating that most ggplot2 layer classes (Geoms, Stats, Scales) are stateless, and encouraging extension authors to follow that convention.

Local verification: all example code paths run under `devtools::load_all()` with the expected values.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,9 @@
 * `theme(strip.placement.x)` and `theme(strip.placement.y)` can be used for more
   granular control of strip placement when facetting. These have existed for some
   time but were not previously documented (@arcresu, #6827).
+* The example in `?ggproto` now demonstrates a stateless class,
+  with additional guidance noting that most ggplot2 layer classes
+  are stateless (@CuiweiG, #6582).
 
 # ggplot2 4.0.2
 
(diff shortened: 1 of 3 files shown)


teunbrand:
Thanks for preparing a PR!
I think the proposed change is not exactly what we want users to learn about ggproto classes. It may have very well be my insufficient explanation in #6582 that had set this on the wrong track.
To explain a bit more; it is fine to *read* from a ggproto object's fields, but to *write* to that field is discouraged.
In the old example, `self$x <- self$x + n` is the problematic piece.

CuiweiG:
Thanks for the clarification — the write-vs-read distinction wasn't clear to me from the issue alone. Will revise the example to keep the field-based inheritance pattern but drop the mutation. Back to you shortly.

PR #6866, opened 2026-06-09
Revise as_cli to return a single string without \n characters

### LINKED ISSUE #6789: Error messages using cli may wrap incorrectly for S7 property validators
In an S7 property validator, the error message is supposed to be "a single string" according to the docs, but I noticed that ggplot2 sometimes violates this. When you use {cli} to format the error message, it may cause a long message to wrap across multiple lines. Each line is then a separate string in a multi-string vector.

The consequence of this is that each line of the error message is prefaced with `- @<property_name>`, making it look perhaps like several error messages instead of just one long one. Here's an example (note that the line wrapping depends on the console width):

``` r
ggplot2::element_text(face = "foo")
#> Error: <ggplot2::element_text> object properties are invalid:
#> - @face must be one of "plain", "bold", "italic", "oblique", or
#> - @face "bold.italic".
```

I think this would be easier to read as:

```r
#> Error: <ggplot2::element_text> object properties are invalid:
#> - @face must be one of "plain", "bold", "italic", "oblique", or
#>   "bold.italic".
```

If you agree that this is a (admittedly very small) problem, I think it could be fixed by revising your internal `as_cli` function to use `paste(collapse)`, so the output would be a single string. Something like this:

```r
as_cli <- function(..., env = parent.frame(), collapse = NULL) {
  cli::cli_text(..., .envir = env) |>
    cli::cli_fmt() |>
    paste(collapse = collapse)
}

long_string <- "I am a very {.emph long} string and I just keep going and going but now I'm done"

as_cli(long_string)
#> [1] "I am a very \033[3mlong\033[23m string and I just keep going and going"
#> [2] "but now I'm done"                                                      

as_cli(long_string, collapse = "\n")
#> [1] "I am a very \033[3mlong\033[23m string and I just keep going and going\nbut now I'm done"
```

Note that this preserves `as_cli`'s current behavior if you leave `collapse = NULL` . When using this for S7 property validators, you could perhaps do `as_cli(long_string, collapse = "\n  ")` to indent the wrapped line a bit.

Anyway, this is me being super picky about a very minor issue. Thanks for your consideration.

dmuenz:
Attempts to fix #6789. This PR revises the internal helper function `as_cli`, which is used to richly format text for printing to the console.

Previously, `as_cli(...)` could return a character vector of length > 1 if the text in `...` was longer than the console width. That's a problem for S7 property validators which expect a string (i.e. a length-1 character vector) for signaling an error. So this PR aims to fix that by

1) using the `cli::cli_fmt(collapse = TRUE, strip_newline = TRUE)` options which ensure we get a string but which may have internal `"\n"` characters, and

2) locally setting the `cli.width` global option to an enormous number so that `cli::cli_fmt` thinks the console is enormously wide and thus does not insert any `"\n"`s.

The reason I don't want `cli::cli_fmt` to insert newline characters is that the function generally doesn't have access to the complete text to be printed, and thus it will insert newlines in the wrong places. E.g., with S7 property validators, the text `"- @prop_name "` is prepended to the error message, but cli doesn't account for the length of that text which can lead to strange-looking line wrapping.

Regarding testing, I added a new direct test of the revised `as_cli` behaviour in `test-utilities.R`. I also revised a couple tests of S7 property validators in `test-properties.R` which previously passed but only because they weren't testing the error messages with a narrow console width. Now the tests simulate a narrow console and they still pass but only because of the revision to `as_cli`.

I ran the whole suite of package tests and they all passed (except for a few that were skipped because they're OS-specific), including tests of the behavior of other functions that use `as_cli`, including `check_required_aesthetics` and `validate_subclass`.

I think this is ready for review at your convenience.

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -786,7 +786,12 @@ attach_plot_env <- function(env) {
 }
 
 as_cli <- function(..., env = caller_env()) {
-  cli::cli_fmt(cli::cli_text(..., .envir = env))
+  withr::local_options(cli.width = 1e9)
+  cli::cli_fmt(
+    cli::cli_text(..., .envir = env),
+    collapse = TRUE,
+    strip_newline = TRUE
+  )
 }
 
 as_unordered_factor <- function(x) {
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -786,12 +786,10 @@ attach_plot_env <- function(env) {
 }
 
 as_cli <- function(..., env = caller_env()) {
-  withr::local_options(cli.width = 1e9)
-  cli::cli_fmt(
+  inject(paste(!!!cli::cli_fmt(
     cli::cli_text(..., .envir = env),
-    collapse = TRUE,
     strip_newline = TRUE
-  )
+  ), collapse = ""))
 }
 
 as_unordered_factor <- function(x) {
--- a/tests/testthat/test-properties.R
+++ b/tests/testthat/test-properties.R
@@ -21,7 +21,6 @@ test_that("property_choice works as intended", {
   expect_length(choice$validator(NULL), 0)
   expect_length(choice$validator("B"), 0)
   # Bad input
-  withr::local_options(cli.width = 20)
   expect_length(choice$validator("X"), 1)
   expect_length(choice$validator(12), 1)
 })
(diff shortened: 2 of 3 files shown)


teunbrand (approved):
Thanks this mostly looks good to me. I made a few tweaks to rely a little less on the cli width option, and moved about the test for narrow console size. Thank you for the contribution!

PR #3663, opened 2019-12-12
Fix theme inheritance when not inheriting from blank

clauswilke:
I think this fixes #3662. I still want to add a few additional unit tests to prevent future such regressions.

I had to update one vdiffr reference image because it wasn't drawn correctly previously. (Line styling wasn't properly inherited across a blank element.)

``` r
library(ggplot2)

ggplot(mtcars) + 
  geom_point(aes(disp, mpg)) + 
  facet_wrap(~gear) + 
  theme(strip.text = element_blank(), strip.text.x = element_text())
```

<sup>Created on 2019-12-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #6878, opened 2026-07-09
Avoid ':::' for exported name in tests

MichaelChirico:
Came across this while testing `lintr::namespace_linter()`.

This is exported since 14 years ago (6661a5b312810f719556cc422db4d0d29e2ad243):

https://github.com/tidyverse/ggplot2/blob/6870419aa6e106c3580c45c81d5b688cb31758bd/NAMESPACE#L275

This appears to have been an oversight in #3663


teunbrand (approved):
Nice catch! LGTM, thanks Michael for the contribution

PR #6880, opened 2026-07-12
Fix plot tag rendering at horizontal positions

sims1253:
`vjust` is assigned only inside `if (top || bottom)` but used in the matching `else`:

Found with https://github.com/sims1253/ry :)


sims1253:
I think the rlang bump from 1.2.0 to 1.3.0 is causing the reordered legends which cause the snapshot tests to fail.

teunbrand (approved):
Nice find, this looks good to me. Thank you for the contribution!

PR #6885, opened 2026-07-29
Export `combine_elements()`

### LINKED ISSUE #6882: Public API for resolving a theme element against a parent
I'm developing a ggplot2 extension that needs to resolve a user-supplied `element_*()` object as if it were a child of a specific theme element. I'd like to do this without permanently registering the element in the global element tree.

### Use case:

My extension, let's call it supertable, adds tables to plots. Users can customize their appearance both via the theme (through custom registered elements like `supertable.value`) and via inline parameters:

```r
supertable(
  value_style = element_text(size = rel(2)),
  title_style = element_text(margin = margin_part(b = 6),
  line_style = element_line(color = "red")
)
```

The inline parameter should sit "above" the theme element in the inheritance chain — i.e., `rel()` and `margin_part()` should resolve against the fully inherited theme element value, exactly as they would for any normal parent-child relationship in the element tree.

### The problem:

The only way to get full resolution (including `rel()`, `margin_part()`, and any future deferred-resolution mechanisms) is through `calc_element()`. `merge_element()` seems like it should help here, but it simply overwrites fields from the new element onto the old one without resolving deferred values like `rel()`. So `calc_element()` appears to be the only option, but it requires the element to be registered in the global tree via `register_theme_elements()`. So my current workaround is:

```r
calc_element_with_override <- function(override, element, theme, el_class) {
  if (is.null(override)) override <- el_class()
  secret_name <- paste0(element, ".__override__")

  register_theme_elements(
    !!secret_name := el_class(),
    element_tree = rlang::list2(
      !!secret_name := el_def(el_class, inherit = element)
    )
  )

  theme[[secret_name]] <- override
  calc_element(secret_name, theme)
}
```

This works, but it permanently alters the global element tree as a side effect, which feels wrong.

### What would help:

An exported function that resolves an element object against a parent element object. This would apply the same resolution logic as `calc_element` (resolving `rel()`, `margin_part()`, etc.) but without requiring the element to be registered globally. It could be called `resolve_element`.

I suspect this could also be useful for other extension authors who need to compute derived styles at render time. Thanks for your consideration.

dmuenz:
This PR aims to fix #6882.

ggplot2 does not have an exported function that resolves a theme element object against a parent element object, applying the same resolution logic as `calc_element()`. In particular, such a function should correctly resolve deferred values like `rel()` and `margin_part()`. However, an unexported function named `combine_elements()` does exactly this -- it is the function that `calc_element()` uses under the hood.

So this PR simply (1) exports `combine_elements`, (2) improves its documentation since it will now be user-facing, and (3) for the same reason changes the function's parameter names from `e1` and `e2` to `child` and `parent`. As a bonus, (4) I added some new unit tests to check that deferred value inheritance works.

Note that an exported function named `merge_element()` exists and appears very similar to `combine_elements()`. But `merge_element()` does not resolve deferred values between a child and parent element object. I address this in the revised documentation for `combine_elements()`.

BTW, I'm open to other name ideas for what I'm calling "deferred values", i.e., the return values of the `rel()` and `margin_part()` helpers that don't fully specify a value. I use this term in the documentation.


teunbrand:
Thanks for this, I certainly think `child` and `parent` are clearer variable names. I'm bemusing if we should expose this functionality via an S7 generic, and have the ANY-ANY signature run the current `combine_elements()`. It'd allow people to override specialty inheritance in case they develop a richer variant of base elements, like `ggtext::element_markdown()` and `element_text()`. Any thoughts?

R/theme.R
```diff
@@ -944,98 +944,136 @@ S7::method(merge_element, list(S7::new_S3_class("element"), S7::class_any)) <-
...
+#' [margin()] object. Can also be `NULL` to inherit everything from `parent`.
+#' @param parent A parent element object or property value from which `child`
+#' inherits. Can be `NULL` if there is nothing to inherit from.
 #'
-combine_elements <- function(e1, e2) {
-
-  # If e2 is NULL, nothing to inherit
-  if (is.null(e2) || is_theme_element(e1, "blank")) {
-    return(e1)
+#' @keywords internal
```
llrs-roche, line 961: If it is exported it shouldn't be documented as internal

llrs-roche (commented):
While I'd prefer to keep things simple if this would help to existing established packages as ggtext it might be worth the extra work. 

Looking forward to this thanks @dmuenz for issue and PR

R/theme.R
teunbrand: `@keywords internal` is appropriate here, this is extension developer oriented, not end-user oriented

PR #6886, opened 2026-08-01
Correct secondary axis test assertions

sims1253:
Found this during my most recent [ry](https://github.com/sims1253/ry) audit

A number of tests had parantheses issues or were self-comparisons.


teunbrand (approved):
This looks good to me. Perhaps these tests ought to use the tolerance option in `expect_equal()`, but that is a problem for another day. Thanks for the contribution!

PR #6893, opened 2026-09-02
Snapshot upkeep

teunbrand:
Added two things to upkeep snapshots:

* Fixed legend order, to prevent frivolous placement of legends depending on hash algorithm. Needed because of a recent change in rlang hashing.
* Announcing skippable snapshot files to prevent deleting the snapshots by just running the tests on non-windows machines. First time I'm personally running the tests on macos, which prevented me from spotting the issue before.


thomasp85 (approved):
LGTM

PR #6901, opened 2026-09-17
Documents scale_*_date NA behavior #6868

### LINKED ISSUE #6868: scale_x_date does not allow NA as minimum limit
I found a problem with the minimum `limits` of `scale_x_date` and `scale_x_datetime`. According to the documentation, `NA` could be use.

> A numeric vector of length two providing limits of the scale. Use NA to refer to the existing minimum or maximum.
But a  `POSXIct` seems to be required.  

I expected that `NA` could be use, without aving to use `as.POSIXct(NA)`. The upper limit (second element of `limit`) seems OK.

```r
library(ggplot2)
library(lubridate)

# Example adapted from ?scale_x_date() 
last_month <- Sys.Date() - 0:29
set.seed(1)
       
df <- data.frame(
  date = last_month,
  datetime = last_month-hours(0:29), # testing with datetime
  price = runif(30)
)  

base_datetime <- ggplot(df, aes(datetime, price)) +
  geom_line()

# NA as maximum works
base_datetime + scale_x_date(limits = c(Sys.Date() - 7, NA))

```

``` r

#NA as minimum don't works: 
base_datetime + scale_x_datetime(limits = c(NA,  Sys.Date() - 7) )
#> Error in `transform$transform()`:
#> ! `transform_time()` works with objects of class <POSIXct> only

# Work-around: as.POSIXct(NA)
base_datetime + scale_x_datetime(limits = c(as.POSIXct(NA), Sys.Date() - 7)   )
#> Warning: Removed 7 rows containing missing values or values outside the scale range
#> (`geom_line()`).
```

``` r

# Same behaviour with date (instead of datetime)
base <- ggplot(df, aes(date, price)) +
  geom_line()
base + scale_x_date(limits = c( NA,Sys.Date() - 7 ))
#> Error in `transform$transform()`:
#> ! `transform_date()` works with objects of class <Date> only
```
<sup>Created on 2026-06-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I'm a new GitHub contributor, feedback is welcome.

melissa-gaughan:
Adds clearer documentation on how to use NA as existing minimum or maximum in limits for datetimes.

Fixes #6868


thomasp85 (changes requested):
Thanks!

Can I get you to rerun documentation so that the Rd file is updated as well

R/scale-date.R
```diff
@@ -28,6 +28,16 @@
 #'   - A `Date`/`POSIXct` vector giving positions of minor breaks
 #'   - A function that takes the limits as input and returns minor breaks as
 #'     output
+#' @param limits One of:
+#'   - `NULL` to use the default scale range
+#'   - A numeric vector of length two providing limits of the scale.
```
teunbrand: If this is date-specific documentation, maybe we should change this recommendation from 'A numeric vector' to 'A date vector' or 'A date(time) vector' as well.

thomasp85: oh yeah, good call

thomasp85 (approved):
LGTM

thomasp85:
Thanks!

